Issue #3050, opened 2017-12-13
After first building, add new codes to Program.cs of .net core project and build again does not work well.


Xiu-Li:
- Create a new .net core console app project, first build, run and debug succeeded.
- The Output window of Build shows as below:
`========== Build: 0 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========`
- Then edit Program.cs file of the project, add some new codes "`Console.WriteLine("Hi Xiu!");`" and save.
- Build the project again, the Output window of Build still shows " `========== Build: 0 succeeded, 0 failed, 1 up-to-date, 0 skipped ========== `".
![image](https://user-images.githubusercontent.com/24784908/33932122-11d4f816-e02d-11e7-897e-092f37de37ea.png)
- Run the project again, the new codes added can't be outputted.
- Put a breakpoint in the code. F5, the breakpoint is not hit.
![image](https://user-images.githubusercontent.com/24784908/33932181-37a63d20-e02d-11e7-9b4f-4f77d2fce0b5.png)

VS Build: **D15.6/27212.00 (15.6.0 Preview 2.0)**
dotnet --info:
> .NET Command Line Tools (2.1.4-preview-007237)
> 
> Product Information:
>  Version:            2.1.4-preview-007237
>  Commit SHA-1 hash:  9c5843d071
> 
> Runtime Environment:
>  OS Name:     Windows
>  OS Version:  10.0.10586
>  OS Platform: Windows
>  RID:         win10-x64
>  Base Path:   C:\Program Files\dotnet\sdk\2.1.4-preview-007237\
> 
> Microsoft .NET Core Shared Framework Host
> 
>   Version  : 2.0.4
>   Build    : 7f262f453d8c8479b9af91d34c013b3aa05bc1ff
>

panopticoncentral:
@davkean If I turn on up to date logging, it looks like we are getting nothing from the output group dataflow, so there's no output to compare against. The code hasn't changed, are there any known issues in the dataflow?

davkean:
Yes the dataflow block has changed - will forward you the email.

panopticoncentral:
Actually, this issue is not fixed by the PR. The issue seems to be that CPS sometimes reports output groups, sometimes doesn't. That made it seem like this fixed the problem but it really didn't.

PR #3196, opened 2018-01-29
Fix XAML rule setup for up to date check

panopticoncentral:
Customer scenario

The customer makes a change to their .NET Core project and runs. The project system thinks the project is up to date and doesn't rebuild, so the customer doesn't see their changes.

Bugs this fixes:

#3050

Workarounds, if any

Customer has to do a rebuild.

Risk

Low, this is installing a file that is missing.

Performance impact

None, makes things work the way they're supposed to.

Is this a regression from a previous update?

Yes.

Root cause analysis:

A fix for up to date checks introduced a new XAML file which needed to be added to setup to function in setup scenarios. This didn't show up in F5 testing of the project system.

How was the bug found?

Customer report


jmarolf:
@panopticoncentral can you create a bug so we add infrastructure to verify the swr files are up-to-date with the xaml files?

natidea:
Approved to merge via https://devdiv.visualstudio.com/DevDiv/_workitems/edit/559198

Pilchie:
I happened to be online, so I merged this so that we could get a signed build going for insertion a bit earlier.

panopticoncentral:
Thanks, I was just checking in.

PR #3198, opened 2018-01-30
Use binpath for communication with IDE

KevinRansom:
Multi-targeting doesn't work correctly in the F# IDE.

The reason is the IDE doesn't know how to identify the correct cached references and source files.  Currently it uses the project file name, this is insufficient because in a multi-targeting project there is in fact an in memory project for each target framework.

The project tracker can identify the correct in memory project using the binpath of the project output file.  In the project system the context object also identifies the binpath for the project being "design-time" built.

This change modifies the CommandLine notification to send the binpath rather than the project path.

/cc @Pilchie ,  can we see if this makes the bar for preview 4?

It also will require an F# update.

Thanks

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.FSharp/ProjectSystem/LanguageServices/FSharpParseBuildOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.FSharp/ProjectSystem/LanguageServices/FSharpParseBuildOptions.cs
@@ -17,17 +17,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         private const string SlashReferencePrefix  = "/r:";
         private const string LongReferencePrefix   = "--reference:";
 
-        private readonly UnconfiguredProject _project;
-
         [ImportMany]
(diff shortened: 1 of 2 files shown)


brettfo:
What's the danger if this change gets inserted, but the corresponding F# change doesn't make it?  On a related note, can we get pre-approval for both this change _and_ the F# change to ensure they make it in together?

Pilchie:
Yes, shiproom would approve the scenario as a whole, and we should insert both changes into VS in the same PR.

KevinRansom:
@brettfo the F# PR takes care of that.

KevinRansom:
@brettfo but yes, this will actually break, netsdk if the F# is not built.  So the F# change is vital, the project system change necessary for it to work.

I could make an additional export that would cause both to work, do you think I should, or can we get the coordination correct?

jmarolf:
@KevinRansom can you link to the F# PR please?

ctaggart:
> can you link to the F# PR please?

@jmarolf PR is https://github.com/Microsoft/visualfsharp/pull/4286

PR #3205, opened 2018-01-31
Set up Sdk Dependency Model Ids to ignore version information

### LINKED ISSUE #3184: Create a new .net core/standard project, the dependencies node occurs warning sign.
## **Description:**
Create a new .net core/standard project, restore, build and run succeeded.
The Output window shows restore finished, no any warnings.
(I try a restore in the command line, restore completed, no any warnings.)
But the dependencies node occurs warning sign.
image
Solution: Close the solution and reopen
## **Test Environment:**
d15svc/27130.2024 (15.5.4) + master/SDK 2.1.300-preview2-008017:
(Add dotnet-core package source: https://dotnet.myget.org/F/dotnet-core/api/v3/index.json)
dotnet --info:
```
.NET Command Line Tools (2.1.300-preview2-008017)

Product Information:
 Version:            2.1.300-preview2-008017
 Commit SHA-1 hash:  f4a0d2db0e

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10240
 OS Platform: Windows
 RID:         win10-x86
 Base Path:   C:\Program Files\dotnet\sdk\2.1.300-preview2-008017\

Microsoft .NET Core Shared Framework Host

  Version  : 2.1.0-preview1-26116-04
  Build    : eeb3a84e2142aba791fe003564b8afd048c45ff9

```

natidea:
Version was unset for earlier releases, but is set for the first time in .NET Core 2.1 (by https://github.com/dotnet/sdk/pull/1572) creating problems matching SdkDependencyModel to NuGet Package references. See https://github.com/dotnet/project-system/issues/3184#issuecomment-361791926 for more details

This fix overrides base class implementation and ignores version information. It turns out this approach is used in other derived classes as well. I also considered a fix in which the broken filter checks for SDK using an Id without version and one with version, but this approach seems more consistent with other models in this project. 

At this point, it is unclear why the base class implementation needs to include version in its ID since many of its derived classes ignore version or leave it empty. But since this is an escrow fix, I'll avoid making a sweeping change.

**Customer scenario**

Customers using .NET Core 2.1 SDK see unresolved warning in dependency node, and cannot expand SDK to see its packages

**Bugs this fixes:** 

Fixes #3184 
Fixes [556559](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/556559)

**Workarounds, if any**

Close and reopen solution often resolves this, but this depends on the order of dataflows

**Risk**

Low, previous SDKs did not include version information either

**Performance impact**

Low, this just overrides a base class implementation

Regression from .NET Core 2.0 to .NET Core 2.1 SDK

**Root cause analysis:**

.NET Core SDK team probably noticed version was missing and added it. 2.1 is still in early preview so problems like this are being weeded out.

**How was the bug found?**

Test team

/cc @dotnet/project-system

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/SdkDependencyModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/SdkDependencyModel.cs
@@ -43,5 +43,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Models
                                 .FirstOrDefault();
             Caption = string.IsNullOrEmpty(Version) ? baseCaption : $"{baseCaption} ({Version})";
         }
+
+        private string _id;
+        public override string Id
+        {
+            get
+            {
+                if (_id == null)
+                {
+                    _id = OriginalItemSpec;
+                }
+
+                return _id;
+            }
+        }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/SdkDependencyModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/SdkDependencyModel.cs
@@ -44,18 +44,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Models
             Caption = string.IsNullOrEmpty(Version) ? baseCaption : $"{baseCaption} ({Version})";
         }
 
-        private string _id;
-        public override string Id
-        {
-            get
-            {
-                if (_id == null)
-                {
-                    _id = OriginalItemSpec;
-                }
-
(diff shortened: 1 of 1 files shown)


natidea:
Shiproom Approved to merge via [556559](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/556559)
/cc @Pilchie @dotnet/project-system for more signoffs

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/SdkDependencyModel.cs
```diff
@@ -43,5 +43,19 @@ internal class SdkDependencyModel : DependencyModel
                                 .FirstOrDefault();
             Caption = string.IsNullOrEmpty(Version) ? baseCaption : $"{baseCaption} ({Version})";
         }
+
+        private string _id;
+        public override string Id
+        {
+            get
+            {
+                if (_id == null)
```
Pilchie: Is this really worth it?  It looks like `OriginalItemSpec` is passed to the base already constructed.  Does the property do anything other than just return it?

Issue #2887, opened 2017-10-12
Add/move file gestures in file tree


cartermp:
Ref: https://github.com/Microsoft/visualfsharp/issues/3740

Add new file above or below selected item will have to be implemented to get us to feature parity for F# projects.

forki:
This is very important for languages that have explicit file order (that one at least). We need a way to add files below or above other files and a "move file up" / down context menu

davkean:
As a workaround right now, you'll be able to edit the project live and move things around.

forki:
Yes that's what I already do. Thank you for making the fsproj directly
editable from vs btw

Am 13.10.2017 03:35 schrieb "David Kean" <notifications@github.com>:

Pilchie:
Still can't assign issues to @Tihan for some reason :(

cartermp:
Just as an update for community folks - @TIHan has made good progress on this after he sorted out the remaining file ordering issues.

TIHan:
This has been resolved.

Pilchie:
We have the Move, but not the Add, right?

TIHan:
Ah, gotcha. I thought this was only for the move. Re-opening.

forki:
Cool. So that implies ordering is resolved as well?

jmarolf:
@forki yes ordering has been fixed.  These should all be available in the next Visual Studio Preview

Pilchie:
Ordering should *work* in 15.5, though the UI doesn't update live if it's changed in 15.5.  That has been fixed for 15.6 as well.  Also, technically, I think this will show up in the *second* preview from now (15.6 Preview 4), since we've already forked for Preview 3.

TIHan:
Now this is resolved. :) 

We still have more work to do with folders and paste/drop, but we will create separate issues for them.

PR #3206, opened 2018-01-31
Added Add Above and Add Below commands for files

TIHan:
**Customer scenario**

Users need to be able to add files above and below existing files in solution explorer for languages that need file ordering, such as F#.

**Bugs this fixes:** 

https://github.com/dotnet/project-system/issues/2887

**Workarounds, if any**

Add a file and move it using solution explorer or editing the proj file.

**Risk**

Adds some changes to existing ordering for moving up and down, but nothing significant. Ordering work is self contained.

**Performance impact**

No performance impact.

No

**Root cause analysis:**

Feature was not implemented to begin with.

**How was the bug found?**

Feature was not implemented to begin with.

---

A few comments:

These changes are only just for adding files above and below other files or folders. It also overrides the behavior for the standard add new item/add existing item for languages that need file ordering, F#.

The commands for ordering are in their own folder to keep everything organized.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
@@ -1,0 +1,89 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.ProjectSystem.Input;
+
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
+{
+    internal abstract class AbstractAddItemCommand : AbstractSingleNodeProjectCommand
+    {
+        private readonly IPhysicalProjectTree _projectTree;
+        private readonly IUnconfiguredProjectVsServices _projectVsServices;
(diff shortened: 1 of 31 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        protected abstract bool CanAdd(IProjectTree target);
+
+        protected abstract IProjectTree GetNodeToAddTo(IProjectTree target);
+
+        protected abstract System.Threading.Tasks.Task OnAddingNodesAsync(IProjectTree nodeToAddTo);
+
+        protected abstract System.Threading.Tasks.Task OnAddedNodesAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> addedNodes, IProjectTree target);
+
+        protected System.Threading.Tasks.Task ShowAddNewFileDialogAsync(IProjectTree target)
```
Pilchie: Nit: Does this need to be qualified?  Consider a using alias for `Task` if so.

Pilchie:
Well need the ask mode template filled out on the VSTS bug here.

TIHan:
Template has been filled out. This should be ready to go. :)

TIHan:
@jmarolf what bit are you referring to?

jmarolf:
@TIHan we have tests for move but not add item

TIHan:
@jmarolf I don't think we can test add item as it is a dialog.

TIHan:
There are tests for seeing if moving nodes above and below other nodes is correct. This happens after a file(s) as been added after the dialog.

Pilchie:
Submitted the [VSTS Bug](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/506515) for @MattGertz

natidea:
Approved to merge via https://devdiv.visualstudio.com/DevDiv/_workitems/edit/506515

Issue #3207, opened 2018-02-01
Newly added F# signature files are not part of compilation by default, ItemType is None.


TIHan:
When adding a new ".fsi", F# signature file", to a F# project, that file will not be part of compilation by default because the item type is None.

**Current**
```<None Include="File1.fsi">```

**What We Expect**
```<Compile Include="File1.fsi">```

PR #3208, opened 2018-02-01
Added .fsi, .ml, .mli file extensions to F# schema

TIHan:
**Customer scenario**

Customer adds a ".fsi", F# signature file, to their F# project and it by default will not be part of compilation.

**Bugs this fixes:** 

https://github.com/dotnet/project-system/issues/3207

**Workarounds, if any**

Edit the "fsproj" file to change the item type.

**Risk**

No risk

**Performance impact**

No performance impact

No

**Root cause analysis:**

F# devs typically don't create ".fsi" a lot.

**How was the bug found?**

Testing adding files.

---

I haven't been able to get this to work, or be able to test this. I'm certain this change is what associates file extensions with item types, according to here: https://github.com/Microsoft/VSProjectSystem/blob/master/doc/extensibility/custom_item_types.md

I also followed the instructions for VS to respect the rules: https://github.com/dotnet/project-system/blob/master/docs/repo/getting-started.md#command-line-1

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/FSharp.ProjectItemsSchema.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/FSharp.ProjectItemsSchema.xaml
@@ -12,5 +12,8 @@
     <ItemType Name="Compile" DisplayName="F# compiler"/>
     
     <FileExtension Name=".fs" ContentType="FSharpFile"/>
+    <FileExtension Name=".fsi" ContentType="FSharpFile"/>
+    <FileExtension Name=".ml" ContentType="FSharpFile"/>
+    <FileExtension Name=".mli" ContentType="FSharpFile"/>
 </ProjectSchemaDefinitions>
 


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/FSharp.ProjectItemsSchema.xaml
```diff
@@ -12,5 +12,8 @@
     <ItemType Name="Compile" DisplayName="F# compiler"/>
     
     <FileExtension Name=".fs" ContentType="FSharpFile"/>
+    <FileExtension Name=".fsi" ContentType="FSharpFile"/>
```
Pilchie: Is the right default action to include these as part of the compile step?

TIHan: Yes, most definitely.

Pilchie:
Let's move this to 15.7 instead of taking it through escrow - there is a pretty easy workaround.

davkean:
Can you try manually copying over these files into program files and see if my instructions are wrong?

TIHan:
I did the manual copying in program files and it worked! :)

davkean:
@Pilchie Can we get approval for this?

Pilchie:
Approved - thanks @TIHan

PR #3217, opened 2018-02-02
Add dot net core2.1 detection

BillHiebert:
This address Bug 557147: [E2EBlocking] Visual Studio and .NET Core 2.1 Preview changes for 15.6
Specifically,
1. Added a new DotNetCoreProjectCompatibilityDetector which is import by the managed project package during its async initialization. Since this part is in the global project scope the package needed to obtain the CPS ExportProvider and use that to get the component.
2. The detector has 3 levels of compatibility: Supported, Partial, Not Supported. The difference in levels defines the message the users sees and whether the message can be suppressed when opening new projects or solutions.
3. Behavior is:
-  If visual studio is a preview release the detection is disabled. 
- Listens to solution load events and detects when background solution load completes.  For each project in the solution is obtains its TFP and if targeting <= 2.0 it is supported, if = 2.1 is partially supported, > 2.1 it is not supported.
- Handles projects added to the solution or unload\reloaded into the solution
- The most severe message is the one which is displayed to the user.
- A user will only be warned once per solution
- If the project being validated is newly created no message is displayed
4. Added support for INewProjectInitializationProvider in order to detect that a project was newly created. CPS will call this method for new projects and it captures this state and make it available to the detector
5. Added XAML for a DontShowAgainMessageBox to handle the checkbox and learn more hyperlinks in the messages
6. Added interface definition for IVsAddId to determine release\preview builds


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    /// <summary>
+    /// <see cref="IDotNetCoreProjectCompatibilityDetector"/>
+    /// </summary>
+    [Export(typeof(IDotNetCoreProjectCompatibilityDetector))]
+    internal sealed class DotNetCoreProjectCompatibilityDetector : IDotNetCoreProjectCompatibilityDetector, IVsSolutionEvents, IVsSolutionLoadEvents, IDisposable
+    {
+        // The version's below this are compatible, versions above it are unsupported (stronger warning message and no don't show again option), equal
```
Pilchie: Nit: Not a possessive.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+        // to are "partial" so tooling should generally work, new features may not have tooling.
+        static Version s_partialSupportedVersion = new Version(2, 1);
```
Pilchie: Nit: explicit "private"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+
+        const string LearnMoreFwlink = "https://go.microsoft.com/fwlink/?linkid=867264";
```
Pilchie: Nit: explicit private here too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+            DialogServices = dialogServices;
+            ThreadHandling = threadHandling;
+        }
+
+        public async Task InitializeAsync()
+        {
+            await ThreadHandling.Value.SwitchToUIThread();
+
+            // Do nothing, don't hook into events if not the release channel
+            if(IsReleaseChannel())
```
Pilchie: I thought Andrew had mentioned that we didn't want to have different behavior between Previews and RTWs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+                }
+            
+                if (finalCompatLevel != CompatibilityLevel.Supported)
+                {
+                    // Warn the user.
+                    await WarnUserOfiIncompatibleProjectAsync(finalCompatLevel).ConfigureAwait(false);
+               }
+
+               // Used to know when to process newly added projects
+               _solutionOpened = true;
```
Pilchie: What happens if there is a solution loaded when the package loads (can happen if loaded for Tools\Options, etc)?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+        public int OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)
+        {
+            // Only check this project if the solution is opened and we haven't already warned at the maximum level. Note that fAdded
+            // is true for both add and a reload of an unloaded project
+            if(_solutionOpened && fAdded == 1 && _compatibilityLevelWarnedForThisSolution != CompatibilityLevel.NotSupported)
+            {
+                UnconfiguredProject project = pHierarchy.AsUnconfiguredProject();
+                if (project != null)
+                {
+                    ThreadHandling.Value.JoinableTaskFactory.StartOnIdle(async () =>
```
Pilchie: Why double-hop to idle and then to the threadpool?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// Fired when the solution load process is fully complete, including all background loading 
+        /// of projects. This event always fires after the initial opening of a solution 
+        /// </summary>
+        public int OnAfterBackgroundSolutionLoadComplete()
+        {
+            // Schedule this to run on idle
+            ThreadHandling.Value.JoinableTaskFactory.StartOnIdle(async () =>
+            {
+                // Run on the background
+                await TaskScheduler.Default;
```
Pilchie: Same question about the double-hop.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+               }
+
+               // Used to know when to process newly added projects
+               _solutionOpened = true;
+            });
+
+            return VSConstants.S_OK;
+        }
+
+        private async Task WarnUserOfiIncompatibleProjectAsync(CompatibilityLevel compatLevel)
```
Pilchie: extra "i"?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+        }
+
+        private async Task<CompatibilityLevel> GetProjectCompatibilityAsync(UnconfiguredProject project)
+        {
+            if (project.Capabilities.AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharp))
+            {
+                IProjectProperties properties = project.Services.ActiveConfiguredProjectProvider.ActiveConfiguredProject.Services.ProjectPropertiesProvider.GetCommonProperties();
+                string tfm = await properties.GetEvaluatedPropertyValueAsync("TargetFrameworkMoniker").ConfigureAwait(false);
+                var fw = new FrameworkName(tfm);
+                if ((fw.Identifier.Equals(".NETCore", StringComparison.OrdinalIgnoreCase) || fw.Identifier.Equals(".NETCoreApp", StringComparison.OrdinalIgnoreCase)))
```
Pilchie: Isn't `.NETCore` here for UWP apps, and isn't it already at 5 or something?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+                        return CompatibilityLevel.NotSupported;
+                    }
+                }
+            }
+
+            return CompatibilityLevel.Supported;
+        }
+
+        /// <summary>
+        /// Query the ChannelSuffix property. Will be empty if this is the release channel
```
Pilchie: Is this a documented and supported way of determining this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+        }
+
+        #region Unused
+        
+        /// <summary>
+        /// Unused IVsSolutionEvents
+        /// </summary>
+        public int OnAfterOpenSolution(object pUnkReserved, int fNewSolution)
+        {
+            return VSConstants.S_OK;
```
Pilchie: Is it better to return E_NOTIMPL to avoid repeated calls?  (Not sure if IVsSolutionEvents does that trick or not).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -0,0 +1,317 @@
...
+                uiShell.GetAppName(out string caption);
+
+                if(compatLevel == CompatibilityLevel.Partial)
+                {
+                    // Get current dontShowAgain value
+                    var settingsManager = (ISettingsManager)_serviceProvider.GetService(typeof(SVsSettingsPersistenceManager));
+                    bool suppressPrompt = false;
+                    if( settingsManager != null)
+                    {
+                        suppressPrompt = settingsManager.GetValueOrDefault(SuppressDotNewCoreWarningKey, defaultValue:false);
```
Pilchie: Is there a way for the user to reset the value of this if they change their mind?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Interop/IVsAppId.cs
```diff
@@ -0,0 +1,289 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
Pilchie: Do we have do this?  There isn't a PIA available?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/DontShowAgainMessageBox.xaml
```diff
@@ -0,0 +1,60 @@
+﻿<ui:DialogWindow
```
Pilchie: This handles all the `EnableModeless` stuff VS requires, right?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -282,4 +282,19 @@ In order to debug this project, add an executable project to this solution which
   <data name="OutputWindowPaneTitle" xml:space="preserve">
     <value>Project</value>
   </data>
+  <data name="PartialSupportedDotNetCoreProject" xml:space="preserve">
+    <value>Visual Studio 2017 15.7 or newer is recommended for .NET Core 2.1 projects.</value>
```
Pilchie: I thinking branding recommendations are to put "version" between 2017 and 15.7.

Pilchie: Tagging @tmeschter about additional localized strings too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: No.  Here's an excerpt from the spec:
(note: this will only appear in 15.6 RTW since 15.7 will be available at the time and we want to encourage people to stay on the edge where they’ll have the best experience)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: That's a hold over from when my early version of the code where it used IVsHiearcny to get the TFM. I've removed it in both places.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
tmeschter: If we take this we'll need to ask the Loc team to make another pass, but that can be done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: The code will be initialized and hook into events but it won't do anything until the solution closes and a new one is opened. Generally this is OK, because it means the solution doesn't contain any .netcore projects. However. it would miss the case of a customer adding an existing .NET core project to the already open solution. I'll address that case

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: Not easily since this goes into the settings store. They would likely need to delete the files under local appdata  or wherever VS stores the settings

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: Ah yes. I removed that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: I don't know if it will handle E_NOTIMPL - it is possible it treats it as an error. I'd have to check the sources. I think S_OK is find here since this is just a notification.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Interop/IVsAppId.cs
BillHiebert: Yep. I have thread with the shell team on this and this was the recommendation.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
BillHiebert: Good point

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/DontShowAgainMessageBox.xaml
BillHiebert: Yes. Handled in the base class in from the IDE

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Interop/IVsAppId.cs
BillHiebert: Note that we could remove all the enums and just leave the one constant we are using.  It would certainly reduce the size of the file.

Pilchie: The size of the file doesn't bother me if it's just a COM import definition.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
Pilchie: Yeah, but I think this may have changed yesterday - best to follow up and be sure.

Pilchie: I just synced up with Andrew - we think we should change it to just always show the warning, even if it's preview.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
Pilchie: I thought it was a UX requirement to have a way to reset the state of options like this (usually from Tools\Options).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -211,23 +220,65 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+                    if (fw.Identifier.Equals(".NETCoreApp", StringComparison.OrdinalIgnoreCase))
                     {
-                        return CompatibilityLevel.Partial;
+                        return GetCompatibilityLevelFromVersion(fw.Version);
                     }
-                    else if(fw.Version.Major > s_partialSupportedVersion.Major || (fw.Version.Major == s_partialSupportedVersion.Major && fw.Version.Minor > s_partialSupportedVersion.Minor))
+                    else if (fw.Identifier.Equals(".NETFramework", StringComparison.OrdinalIgnoreCase))
                     {
-                        return CompatibilityLevel.NotSupported;
+                        // Teh interesting case here is Asp.Net Core on full framework
```
Pilchie: Nit: Spelling

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -211,23 +220,65 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+                        IImmutableSet<IUnresolvedPackageReference> pkgReferences = await project.Services.ActiveConfiguredProjectProvider.ActiveConfiguredProject.Services.PackageReferences.GetUnresolvedReferencesAsync().ConfigureAwait(false);
```
Pilchie: Is this a race?  Are they guaranteed not to be resolved?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -211,23 +220,65 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+
+                        // Look through the package references
+                        foreach (var pkgRef in pkgReferences)
+                        {
+                            if (string.Equals(pkgRef.EvaluatedInclude, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase) ||
```
Pilchie: Are package names supposed to ignore case?

Pilchie:
Thanks @BillHiebert.  Two remaining questions:

1. Are package name comparisons supposed to be case insensitive?
2. Is checking the unresolved packages a race?  What happens if background restore happens somewhere in the middle there?

BillHiebert:
1. Package names case insensitive (at least on windows) and I verified this is the case. 
2. There is no race with restore since the code is checking the set of PackageReferences, not the results of restore.

Pilchie:
@BillHiebert I meant [here](https://github.com/dotnet/project-system/pull/3217/files#diff-a93a9f671a3b8255a40c09c60091739dR228) - it is looking at `GetUnresolvedReferencesAsync`.  What happens if they aren't unresolved, but resolved?  Or am I misunderstanding what that method does?

BillHiebert:
My understanding is that it is the list of package references defined in the project file as compared to the set of packages and its dependencies. It doesn't change whether they are resolved by nuget or not.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -28,60 +31,85 @@ internal sealed class DotNetCoreProjectCompatibilityDetector : IDotNetCoreProjec
...
+            ShellUtilitiesHelper = vsShellUtilitiesHelper;
+            FileSystem = fileSystem;
         }
 
+        private readonly IServiceProvider _serviceProvider;
+        private Lazy<IProjectServiceAccessor> ProjectServiceAccessor { get; set; }
+        private Lazy<IDialogServices> DialogServices { get; set; }
+        private Lazy<IProjectThreadingService> ThreadHandling { get; set; }
+        private Lazy<IFileSystem> FileSystem { get; set; }
+        private Lazy<IVsShellUtilitiesHelper> ShellUtilitiesHelper { get; set; }
```
Pilchie: Why private properties instead of fields?  Do they all need "get" or can they be readonly?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -28,60 +31,85 @@ internal sealed class DotNetCoreProjectCompatibilityDetector : IDotNetCoreProjec
...
+
+        private string VersionDataCacheFile { get; set; }
+        private Version OurVSVersion { get; set; }
+
+        private uint _solutionCookie = VSConstants.VSCOOKIE_NIL;
+        private bool _solutionOpened;
+        private CompatibilityLevel _compatibilityLevelWarnedForThisSolution = CompatibilityLevel.Recommended;
+
+        // Tracks how often we meed to look for new data
```
Pilchie: Spelling "meed"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -254,28 +286,150 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
 
         /// <summary>
-        /// Compares the psased in version to the partially supported version and returns the compatability level
+        /// Compares the passed in version to the compatibility data to determine the compat level
         /// </summary>
-        private CompatibilityLevel GetCompatibilityLevelFromVersion(Version version)
+        private CompatibilityLevel GetCompatibilityLevelFromVersion(Version version, VersionCompatibilityData compatData)
         {
-            if(version.Major == s_partialSupportedVersion.Major && version.Minor == s_partialSupportedVersion.Minor)
+            // Omly compare major, minor. The presence of build with change the comparison. ie: 2.0 != 2.0.0
```
Pilchie: Spelling "Omly", and "with" instead of "will"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -254,28 +286,150 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+                    if (versionCompatData != null)
+                    {
+                        // Cache the data locally
+                        if (downLoadedVersionData != null)
+                        {
+                            FileSystem.Value.WriteAllText(VersionDataCacheFile, downLoadedVersionData);
+                        }
+
+                        // First try to match exatly on our VS version and if that fails, match on just major, minor
+                        if (versionCompatData.TryGetValue(OurVSVersion, out VersionCompatibilityData compatData) || versionCompatData.TryGetValue(new Version(OurVSVersion.Major, OurVSVersion.Minor), out compatData))
```
Pilchie: nit: wrap second call to newline.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -254,28 +286,150 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+                        {
+
```
Pilchie: nit: extra blank line.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -254,28 +286,150 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+                                compatData.OpenUnsupportedMessage = string.Format(VSResources.NotSupportedDotNetCoreProject, s_partialSupportedVersion.Major, s_partialSupportedVersion.Minor);
+                            }
+                            _curVersionCompatibilityData = compatData;
+                        }
+                    }
+                }
+            }
+            catch
+            {
+
```
Pilchie: Consider: Logging telemetry so that we know how often/why this ends up happening.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/VsShellUtilitiesHelper.cs
```diff
@@ -37,5 +38,44 @@ public VsShellUtilitiesHelper(IProjectThreadingService threadingService)
...
+
+        /// <summary>
+        /// <see cref="IVsShellUtilitiesHelper.GetVSVersionAsync"/>
+        /// </summary>
+        public async Task<Version> GetVSVersionAsync(IServiceProvider serviceProvider)
+        {
+            await _threadingService.SwitchToUIThread();
+
+            IVsAppId vsAppId = serviceProvider.GetService<IVsAppId, SVsAppId>();
+            if (ErrorHandler.Succeeded(vsAppId.GetProperty((int)VSAPropID.VSAPROPID_ProductSemanticVersion, out object oVersion)) && oVersion is string semVersion)
```
Pilchie: Nit: Wrap after `&&`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -254,28 +286,150 @@ public int OnAfterBackgroundSolutionLoadComplete()
...
+                            versionCompatData = GetCompabilityDataFromCacheFile(checkTimeStamp: false);
+                        }
+                    }
+
+                    if (versionCompatData != null)
+                    {
+                        // Cache the data locally
+                        if (downLoadedVersionData != null)
+                        {
+                            FileSystem.Value.WriteAllText(VersionDataCacheFile, downLoadedVersionData);
```
Pilchie: `VersionDataCacheFile` is only set if `GetLocalAppDataFolderAsync` works.  Do we need to handle the case where it doesn't?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
Pilchie: Also, what if `OurVSVersion` is null here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/VersionCompatibilityData.cs
```diff
@@ -0,0 +1,52 @@
...
+        public string OpenSupportedMessage { get; set; }
+
+        [JsonProperty(PropertyName = "unsupportedVersion")]
+        public Version UnsupportedVersion { get; set; }
+
+        [JsonProperty(PropertyName = "openUnsupportedMessage")]
+        public string OpenUnsupportedMessage { get; set; }
+
+        /// <summary>
+        ///
```
Pilchie: Either add something or remove.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/VersionCompatibilityData.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
+{
+
+    /// <summary>
+    /// Data is looks like the following, Support version is the version at which VS will warn the user and the warning will have a 
+    /// don't show again checkbox. The error version (if specified) indicates the version at which the customer is given a stronger 
+    /// message about support and does not have the don't show again dialog box option. The two versions extablish a range of 
+    /// "supported" versions.
+    /// eg:
```
Pilchie: Consider wrapping in `<code>`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/VersionCompatibilityData.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
+{
+
+    /// <summary>
+    /// Data is looks like the following, Support version is the version at which VS will warn the user and the warning will have a
```
Pilchie: "is looks like"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/VersionCompatibilityData.cs
```diff
@@ -0,0 +1,52 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
+{
+
+    /// <summary>
+    /// Data is looks like the following, Support version is the version at which VS will warn the user and the warning will have a 
+    /// don't show again checkbox. The error version (if specified) indicates the version at which the customer is given a stronger 
+    /// message about support and does not have the don't show again dialog box option. The two versions extablish a range of
```
Pilchie: spelling: "extablish"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: Just style, though I tend to use fields more now or properties with just a get. I'll fix this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: If that doesn't work there are serious problems already in VS, but I'll add a null check.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: Given the whole thing is wrapped in a try catch, it will fallback to what we shipped.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: The missing strings is a supported scenario where we will just use the strings (localized) that are in the product. No reason to add telemetry for this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/VersionCompatibilityData.cs
BillHiebert: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
Pilchie: Still "meed".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
Pilchie: I guess i meant if the download fails, or if we can't read the file from disk for some reason, so maybe the try/catch around line 352 would be a better place for telemetry.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -334,57 +344,48 @@ private CompatibilityLevel GetCompatibilityLevelFromVersion(Version version, Ver
         /// less than 24 hours old, it uses that data. Otherwise it downloads from the server. If the download fails it will use the previously cached
         ///  file, or if that file doesn't not exist, it uses the data baked into this class
         /// </summary>
-        private async Task<VersionCompatibilityData> GetVersionCmpatibilityDataAsync()
+        private VersionCompatibilityData GetVersionCmpatibilityData()
```
Pilchie: Just noticed a missing "o" in Compatibility...

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -334,57 +344,48 @@ private CompatibilityLevel GetCompatibilityLevelFromVersion(Version version, Ver
...
         {
+                // Do we need to update our cached data? Note that since the download could take a long time like tens of seconds we don't really want to
```
Pilchie: Nit: Indented too far.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: Rats, I fixed a different one

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: thanks

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
BillHiebert: fixed

tmeschter:
@Pilchie I'll ask the Loc team to make another pass.

davkean:
Is there a reason this didn't go through the IVsRetargetProjectAsync infrastructure for the block? It has better semantics around prompting/unloading/grouping projects.

davkean:
Why isn't the package reference check stored in the file that is downloaded? Why does the project system know about specific frameworks/packages?

PR #3228, opened 2018-02-08
fix default target DebuggerFlavor property

efvink:
This change fixes an issue when trying to add a custom debugger to an extension - the DebuggerFlavor property gets overridden by the default target.


Pilchie:
Can you retarget this to either `dev15.6.x` or `dev15.7.x` branch depending on the email discussion we're having?

efvink:
Will create a new one

PR #3239, opened 2018-02-11
fix default target DebuggerFlavor property

efvink:
This change fixes an issue when trying to add a custom debugger to an extension - the DebuggerFlavor property gets overridden by the default target.


davkean:
@Pilchie Do 15.7 changes need to be approved?

Pilchie:
@DavKean - yes, unless they are directly .NET Core 2.1 related.

PR #3243, opened 2018-02-12
Fix missing "Clean" command in new projects

Pilchie:
It seems that Clean is hidden if the project isn't ready to build, but VS only
asks once. Having switched to using dataflow, we would return false on the very
first call.

Reverts the changes in 36e5c6c2.

**Customer scenario**

Create a new ASP.NET Core project.  The "Clean" button will be missing until the solution is reloaded.

**Bugs this fixes:** 

[VSTS Bug 550183](https://devdiv.visualstudio.com/DevDiv/Managed%20Project%20System/_workitems/edit/550183)

**Workarounds, if any**

Reload the solution.

**Risk**

Low - reverts to previous implementation used for 15.5.

**Performance impact**

Low - just tracking configuration changed events.

Yes, it's a regression from 15.5, related to moving the service to use DataFlow.

**Root cause analysis:**

It seems that Clean is hidden if the project isn't ready to build, but VS only
asks once. Having switched to using dataflow, we would return false on the very
first call.

**How was the bug found?**

Internal automated tests.


Pilchie:
Approved to merge via [550183](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/550183)

PR #3244, opened 2018-02-12
Allow Project settings AppDesigner tabs to be activated via UI automation

etbyrd:
**Customer scenario**

@Pilchie 
"Pulling back into 15.6 based on Accessibility team's recommendation."

**Bugs this fixes:** 

https://devdiv.visualstudio.com/DevDiv/Managed%20Project%20System/_queries?id=433488&_a=edit&triage=true

**Workarounds, if any**

None

**Risk**

Low

**Performance impact**

None

No

**Root cause analysis:**

We missed a scenario for Accessibility  

**How was the bug found?**

Originally found through Integration Tests

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/DesignerTabButton.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/DesignerTabButton.vb
@@ -280,6 +280,20 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                 _button.PerformClick()
             End Sub
 
+            ''' <summary>
+            ''' Calls default action when select is called. Checks for correct flag
+            ''' </summary>
+            Public Overrides Sub [Select](flags As AccessibleSelection)
+                '''AccessibleSelection.TakeSelection is the flag set when an object is selected,
+                '''specfically from SelectionItem.Select If this flag is set, we call the default
+                '''action on the PageTab, which gives the desired behavior of selecting the item.
+                If (flags And AccessibleSelection.TakeSelection) = AccessibleSelection.TakeSelection Then
+                    DoDefaultAction()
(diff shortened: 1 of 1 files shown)


tannergooding:
Also CC. @dpoeschl

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/DesignerTabButton.vb
```diff
@@ -280,6 +280,20 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
...
             End Sub
 
+            ''' <summary>
+            ''' Calls default action when select is called. Checks for correct flag
+            ''' </summary>
+            Public Overrides Sub [Select](flags As AccessibleSelection)
+                '''AccessibleSelection.TakeSelection is the flag set when an object is selected,
+                '''specfically from SelectionItem.Select If this flag is set, we call the default
+                '''action on the PageTab, which gives the desired behavior of selecting the item.
+                If (flags And AccessibleSelection.TakeSelection) = AccessibleSelection.TakeSelection Then
```
tannergooding: It would probably be good to determine that only respecting `TakeSelection` is the correct behavior.

My own intuition says it is, but we don't want to accidentally miss some edge accessibility scenario where some different behavior is expected for `AddSelection`, for example.

davkean:
@etbyrd Thanks for the bug fix, can you just squash this into a single commit with title/message that describes what you are fixing and why? If you need help with that, give me a ping.

etbyrd:
@davkean No problem - I got that done.

Pilchie:
Approved to merge via [433488](https://devdiv.visualstudio.com/web/wi.aspx?pcguid=011b8bdf-6d56-4f87-be0d-0092136884d9&id=433488)

PR #3227, opened 2018-02-07
Changing the debug dropdown does not run the correct debug target

BillHiebert:
**Customer scenario**
Changing the debug dropdown to run a different debug target does not work reliably. Even though the UI updates, the actual debug target doesn't change. This is a regression caused by this [PR](https://github.com/dotnet/project-system/pull/3174)

**Bugs this fixes:** 
https://devdiv.visualstudio.com/DevDiv/_workitems?id=560770

(either VSO or GitHub links)

**Workarounds, if any**

No real workaround other than to close and reopen the project after making the change. 

**Risk**

Low risk. The change is to link to the Unconfigured capabilities rather than the Configured capabilities. The previous PR mixed Unconfigured project level dataflow with Configured project level dataflow.

**Performance impact**

Low. 

**Is this a regression from a previous update?**
Yes.

**Root cause analysis:**

**How was the bug found?**
Test team manual walk throughs

(E.g. customer reported it vs. ad hoc testing)

PR #3174, opened 2018-01-23
Include capabilities in project subscription dataflow

BillHiebert:
This addresses the issue described in DEVDIV [553720](https://devdiv.visualstudio.com/web/wi.aspx?pcguid=011b8bdf-6d56-4f87-be0d-0092136884d9&id=553720)

Basically, when first reading the launch settings files in response to dataflow, the AspNetCoreWeb is not present which breaks the launch settings provider which imports custom serializers. This is similar to the other race conditions that been reported recently with capabilities not being present when responding to dataflow. The fix is to include the capabilities dataflow in the project subscription in a manner similar to this [PR](https://github.com/dotnet/project-system/pull/3167).

In addition to the above, I was concerned about capabilities changing even when nothing in the project subscription has changed.  The project rule dataflow change handler only looks at the ActiveDebugProfile property and would miss the capabilities only change. To address this I created a separate dataflow block to handle that.

**Risk**
The risk should be low as other than adding the new capability handler the core code didn't change at all. 

**Performance impact**
Shouldn't be any.

**Is this a regression from a previous update?**
Yes.

**Root cause analysis:**
Appears to be related to a change in when capabilities evaluated from msbuild are completed in CPS

**How was the bug found?**
AdHoc and signoff testing

PR #3246, opened 2018-02-13
Fix for bug 565888 - [E2EBlocking] Create ASP.NET Core 2.0 Web Applic…

BillHiebert:
**Customer scenario**
Create a new ASP.NET Core web application with individual authentication.  The sslport is set to 0 instead of the actual port value. When you F5, you'll see a page not found in the browser as it is configured to launch on the https url.l

**Bugs this fixes:** 
[565888](https://devdiv.visualstudio.com/DevDiv/_workitems?id=565888): Create ASP.NET Core 2.0 Web Application with IndAuth, the "ssIport" is 0 in launchSettings.json.

**Workarounds, if any**
Hand edit the launchsettings.json file to add the correct ssl port

**Risk**
Should be low. Removed the extra capabilities dataflow that was added by PR 3174 as it is completely unnecessary. The first snapshot from this dataflow does not contain the proper set of capabilities and unfortunately, the first launch settings snapshot is created from this data. When the web project system validates IIS Express settings later, it doesn't see any because the snapshot was created without the correct capabilities and overwrites the actual settings with its own.

**Performance impact**
Shouldn't be any perf impact

Yes. A regression caused by [this ](https://github.com/dotnet/project-system/pull/3174) change and [this ](https://github.com/dotnet/project-system/pull/3227)one which incorrectly fixed the first one.

**Root cause analysis:**

Poor testing of related scenarios. There is pretty good unit test coverage, but no integration tests.

**How was the bug found?**
CTI team manual test walk throughs.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
@@ -101,7 +101,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
         private TaskCompletionSource<bool> _firstSnapshotCompletionSource = new TaskCompletionSource<bool>();
 
         protected IDisposable ProjectRuleSubscriptionLink { get; set; }
-        protected IDisposable CapabilitiesSubscriptionLink { get; set; }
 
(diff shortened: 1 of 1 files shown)


Pilchie:
@lifengl - can you take a look at this too?  @BillHiebert given where we are, can we do some buddy testing here?

BillHiebert:
@Pilchie -definitely on the buddy testing.

PR #3250, opened 2018-02-13
Add net21 detection to15.5

BillHiebert:
This is a port of this [PR](https://github.com/dotnet/project-system/pull/3217) which adds dotnet 2.1 (and later) detection to VS 15.5.x.  The change is much simpler because it doesn't have the ability to download a compatibility file to change the behavior, and only has two compatibility states: supported and unsupported. Everything above 2.0 is unsupported. The detection logic is all the same with the exception that it does not need to check for newly created projects as VS 15.5 does not support creating projects targeting .NET Core > 2.0.

Note the fwlink is different for 15.5
The strings are identical to what was in the 15.6 PR (even though one of them is not used in this version.
The don't show again message box is identical


Pilchie:
Pro-actively merging to get a signed build and val build for buddy testing.

PR #3252, opened 2018-02-14
Fixed an ordering bug dealing with empty folders and adding multiple items

TIHan:
**Customer scenario**

Adding items to empty folders. Adding multiple items to a folder. Currently, doing these actions can cause invalid ordering of files.

**Bugs this fixes:** 

(n/a)

**Workarounds, if any**

Modifying the .proj file directly.

**Risk**

Low risk. Modified the ordering code slightly.

**Performance impact**

None

No

**Root cause analysis:**

Missed it due to lack of specific testing; the specifics were not defined in the file ordering spec.
We should add these scenarios in the file ordering spec.

**How was the bug found?**

When trying to implement paste ordering.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
@@ -35,7 +35,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
 
         protected abstract Task OnAddingNodesAsync(IProjectTree nodeToAddTo);
 
-        protected abstract Task OnAddedNodesAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> addedNodes, IProjectTree target);
+        protected virtual Task OnAddedNodesAsync(ConfiguredProject configuredProject, IProjectTree target, IEnumerable<IProjectTree> addedNodes, IProjectTree updatedTarget)
+        {
+            // We are not using the updated target so we don't step over the added nodes.
+            return OrderingHelper.TryMoveNodesToTopAsync(configuredProject, addedNodes, target);
+        }
 
         protected Task ShowAddNewFileDialogAsync(IProjectTree target)
         {
@@ -63,25 +67,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
         {
             var nodeToAddTo = GetNodeToAddTo(node);
 
-            // Publish any existing changes that could potentially be here.
-            // Then call OnAddingNodesAsync.
+            // Call OnAddingNodesAsync.
             // Then publish changes that could have taken place in OnAddingNodesAsync.
-            await _projectTree.TreeService.PublishLatestTreeAsync(waitForFileSystemUpdates: true).ConfigureAwait(true);
             await OnAddingNodesAsync(nodeToAddTo).ConfigureAwait(true);
             await _projectTree.TreeService.PublishLatestTreeAsync(waitForFileSystemUpdates: true).ConfigureAwait(true);
 
-            // Get the difference to see what was added.
-            // We do a sanity check to make sure they have a valid display order.
-            // We also order the added nodes by their display order.
             var updatedNode = _projectTree.CurrentTree.Find(node.Identity);
             var updatedNodeToAddTo = _projectTree.CurrentTree.Find(nodeToAddTo.Identity);
-            var addedNodes =
-                updatedNodeToAddTo.Children.Where(x => !nodeToAddTo.TryFind(x.Identity, out var subtree) && OrderingHelper.HasValidDisplayOrder(x))
-                .OrderBy(OrderingHelper.GetDisplayOrder).ToList();
+            var addedNodes = OrderingHelper.GetAddedNodes(nodeToAddTo, updatedNodeToAddTo);
 
             if (addedNodes.Any())
             {
-                await OnAddedNodesAsync(_projectVsServices.ActiveConfiguredProject, addedNodes, updatedNode).ConfigureAwait(true);
+                // Make sure to change ConfigureAwait to "true" if we need switch back.
+                await OnAddedNodesAsync(_projectVsServices.ActiveConfiguredProject, node, addedNodes, updatedNode).ConfigureAwait(false);
             }
 
             return true;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
@@ -156,6 +157,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
         /// </summary>
         public static IProjectTree GetLastChild(IProjectTree projectTree)
         {
+            Requires.NotNull(projectTree, nameof(projectTree));
+
             return GetChildren(projectTree).LastOrDefault();
         }
 
@@ -166,7 +169,79 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
         /// </summary>
         public static IProjectTree GetFirstChild(IProjectTree projectTree)
         {
+            Requires.NotNull(projectTree, nameof(projectTree));
+
             return GetChildren(projectTree).FirstOrDefault();
+        }
+
+        /// <summary>
+        /// Get the difference to see what was added.
+        /// We do a sanity check to make sure they have a valid display order.
+        /// We also order the added nodes by their display order.
+        /// </summary>
+        public static ImmutableArray<IProjectTree> GetAddedNodes(IProjectTree previousTarget, IProjectTree currentTarget)
+        {
+            Requires.NotNull(previousTarget, nameof(previousTarget));
+            Requires.NotNull(currentTarget, nameof(currentTarget));
+
+            return currentTarget.Children
+                .Where(updatedChild => IsAddedNode(previousTarget, updatedChild))
+                .OrderBy(GetDisplayOrder)
+                .ToImmutableArray();
+        }
+
+        /// <summary>
+        /// This moves the nodes above the target's first child. 
+        /// If there are no children with a valid display order of the target or the target is an empty folder, it will try the parent target's first child.
+        /// </summary>
+        public static bool TryMoveNodesToTop(Project project, IEnumerable<IProjectTree> nodes, IProjectTree target)
+        {
+            Requires.NotNull(project, nameof(project));
+            Requires.NotNull(nodes, nameof(nodes));
+            Requires.NotNull(target, nameof(target));
+
+            // Move added nodes to the top. 
+            var nodeToAddAbove = GetFirstChild(target);
+
+            // This could have been an empty folder and nodeToAddAbove is null, therefore
+            //     get the first child of the target's parent.
+            // Empty folders do not have a valid display order.
+            // In the future, if we want to order empty folders, meaning they have a valid display order, 
+            //     this will need to be extended to handle that behavior.
+
+            // This recursively tries to find a node to work with. If there is none at all, this will break on project root.
+            var currentTarget = target;
+            while(nodeToAddAbove == null && currentTarget.IsFolder && !GetChildren(currentTarget).Any())
+            {
+                currentTarget = currentTarget.Parent;
+                nodeToAddAbove = GetFirstChild(currentTarget);
+            }
+
+            if (nodeToAddAbove != null && nodes.Any())
+            {
+                var didAllMoveSuccessfully = true;
+
+                foreach (var node in nodes)
+                {
+                    if (!TryMoveAbove(project, node, nodeToAddAbove))
+                    {
+                        didAllMoveSuccessfully = false;
+                    }
+                }
+
+                return didAllMoveSuccessfully;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// This moves the nodes above the target's first child. 
+        /// If there are no children with a valid display order of the target or the target is an empty folder, it will try the parent target's first child.
+        /// </summary>
+        public static Task<bool> TryMoveNodesToTopAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> nodes, IProjectTree target)
+        {
+            return ModifyProjectAsync(configuredProject, project => TryMoveNodesToTop(project, nodes, target));
         }
 
         /// <summary>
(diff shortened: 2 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -168,6 +168,63 @@ public static IProjectTree GetFirstChild(IProjectTree projectTree)
...
+        /// <summary>
+        /// This moves the nodes above the target's first child. 
+        /// If there are no children of the target and the target is an empty folder, it will try the parent target's first child.
+        /// </summary>
+        public static async Task MoveNodesToTopAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> nodes, IProjectTree target)
+        {
+            await ModifyProjectAsync(configuredProject, project => {
+                MoveNodesToTop(project, nodes, target);
+                return true;
+            }).ConfigureAwait(true);
```
davkean: Why "true" here? Do you need to switch back to original thread if you ended up switching threads?

TIHan: I'm going to rename this to TryMoveNodes*. I will also make it return true if all nodes were successfully moved to the right place. If one node was not moved successfully, then it will return false.

davkean: I mean, why `ConfigureAwait(true)`? We use `ConfigureAwait(false)` to avoid unnecessarily thread switches unless absolutely needed.

jmarolf: why is this method `async` at all?  Why not just return the task to the caller?
```C#
public static Task MoveNodesToTopAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> nodes, IProjectTree target)
{
    return ModifyProjectAsync(configuredProject, project =>
    {
        MoveNodesToTop(project, nodes, target);
        return true;
    });
}
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -168,6 +168,63 @@ public static IProjectTree GetFirstChild(IProjectTree projectTree)
...
 
+        /// <summary>
+        /// Get the difference to see what was added.
+        /// We do a sanity check to make sure they have a valid display order.
+        /// We also order the added nodes by their display order.
+        /// </summary>
+        public static IEnumerable<IProjectTree> GetAddedNodes(IProjectTree previousTarget, IProjectTree currentTarget)
+        {
+            return
+                currentTarget.Children.Where(updatedChild => {
```
jmarolf: please put the `{` on the next line

TIHan:
I'm quite happy with the changes now that I've thoroughly vetted some scenarios. The last thing I want to do is add more specific tests if I can. I will do that tomorrow.

jmarolf:
>             }

Add newline

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs:38 in f3e4aba. [](commit_id = f3e4aba76943784f53171f1e323911712657c251, deletion_comment = False)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -165,9 +171,87 @@ public static IProjectTree GetLastChild(IProjectTree projectTree)
...
+
+            return
+                currentTarget.Children.Where(updatedChild => 
+                {
+                    var hasChild = previousTarget.TryFind(updatedChild.Identity, out var child);
+                    var hasValidUpdatedChild = HasValidDisplayOrder(updatedChild);
+
+                    return (!hasChild || (hasChild && !HasValidDisplayOrder(child))) && hasValidUpdatedChild;
+                })
+                .OrderBy(GetDisplayOrder).ToList();
```
jmarolf: Why the `ToList`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: The changes just return the task.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -165,9 +171,87 @@ public static IProjectTree GetLastChild(IProjectTree projectTree)
...
+                nodeToAddAbove = GetFirstChild(currentTarget);
+            }
+
+            if (nodeToAddAbove != null && nodes.Any())
+            {
+                var didAllMoveSuccessfully = true;
+
+                foreach (var node in nodes)
+                {
+                    var didMove = TryMoveAbove(project, node, nodeToAddAbove);
```
jmarolf: You could inline this method call to just be `if (!TryMoveAbove(project, node, nodeToAddAbove))`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: I want to force execution and not make this lazy.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: Yes. I actually had this and it didn't get included.

TIHan: Done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
jmarolf: What happens if its lazy? Is there a threading concern? A comment here would be useful so we know not to change this.

jmarolf: Also this logic is complex enough that you may want to create an `IsAddedNode` method so the code just becomes
```C#
return currentTarget.Children
    .Where(updatedChild => IsAddedNode(previousTarget, updatedChild))
    .OrderBy(GetDisplayOrder)
    .ToList();
```

TIHan: I've changed this to return a ReadOnlyCollection which is really what I wanted in the first place. I don't think a comment is really needed at that point.

TIHan: I added `IsAddedNode`.

TIHan:
This is finished.

davkean:
Is this waiting on another sigh off?

TIHan:
Do I need to fill out a template for this one?

Pilchie:
Can we get another review of it before merging?  Otherwise I'm fine.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -166,9 +168,82 @@ public static IProjectTree GetLastChild(IProjectTree projectTree)
...
+        public static ReadOnlyCollection<IProjectTree> GetAddedNodes(IProjectTree previousTarget, IProjectTree currentTarget)
+        {
+            Requires.NotNull(previousTarget, nameof(previousTarget));
+            Requires.NotNull(currentTarget, nameof(currentTarget));
+
+            return currentTarget.Children
+                .Where(updatedChild => IsAddedNode(previousTarget, updatedChild))
+                .OrderBy(GetDisplayOrder)
+                .ToList()
+                .AsReadOnly();
```
davkean: This is very inefficient allocation-wise, it creates a `ReadOnlyCollection<T>` that wraps a `List<T>` that wraps an `T[]`, plus the other 3 allocations for the Linq query. Can we just put this straight into an ImmutableList<T> via `ToImmutableList()/ToImmutableArray()` to reduce some of that?

TIHan: Yes, `ToImmutableArray` should be fine.

TIHan: Done, also changed a similar method to do the same.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
```diff
@@ -35,7 +35,11 @@ public AbstractAddItemCommand(IPhysicalProjectTree projectTree, IUnconfiguredPro
 
         protected abstract Task OnAddingNodesAsync(IProjectTree nodeToAddTo);
 
-        protected abstract Task OnAddedNodesAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> addedNodes, IProjectTree target);
+        protected virtual Task OnAddedNodesAsync(ConfiguredProject configuredProject, IProjectTree target, IEnumerable<IProjectTree> addedNodes, IProjectTree updatedTarget)
+        {
+            // We are not using the updated target so we don't step over the added nodes.
+            return OrderingHelper.TryMoveNodesToTopAsync(configuredProject, addedNodes, target);
```
davkean: I'm unsure about this behavior; If 10 node were successfully handled, but 1 failed to move - you say that the command wasn't handled and someone else gets to handle it. Why is that?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
```diff
@@ -63,25 +67,18 @@ protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree
...
             var updatedNodeToAddTo = _projectTree.CurrentTree.Find(nodeToAddTo.Identity);
-            var addedNodes =
-                updatedNodeToAddTo.Children.Where(x => !nodeToAddTo.TryFind(x.Identity, out var subtree) && OrderingHelper.HasValidDisplayOrder(x))
-                .OrderBy(OrderingHelper.GetDisplayOrder).ToList();
+            var addedNodes = OrderingHelper.GetAddedNodes(nodeToAddTo, updatedNodeToAddTo);
 
             if (addedNodes.Any())
             {
-                await OnAddedNodesAsync(_projectVsServices.ActiveConfiguredProject, addedNodes, updatedNode).ConfigureAwait(true);
+                await OnAddedNodesAsync(_projectVsServices.ActiveConfiguredProject, node, addedNodes, updatedNode).ConfigureAwait(true);
```
davkean: Why are we switching back to the original threads in this method? It's very unclear - and should have a comment. Normally we only switch back if 1) we need to stay on the UI thread, or 2) we took a lock and need to get back on the same thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
TIHan: You are referring to that we are ignoring the result of the `Try`?

davkean: Oh, I was reading this wrong - I assumed it was returning the `Task<bool>` here but its not, its throwing it away.

In that case, why we do spent a bunch of effort capturing whether we successfully added all nodes if we ignore the result?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
TIHan: There isn't a reason why we are switching back. I'm happy to change it to "false".

davkean: Make note, you'll need "true" if you are within the callback that you call within the write lock, but it doesn't appear you are.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
TIHan: Part of the effort is for tests, where we do look at the result. In this case, we can't test the async version, only the non-async one, `TryMoveNodesToTop`.

davkean: Because of the lock? Look at the new IProjectAccessor interface I added for this purpose.

davkean: Unsure if the right stuff in is 15.7 or master.

Pilchie:
Approved for 15.7 - thanks @TIHan

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
TIHan: I believe it was for the locks and the instance of the configured project. A plus side to this is separated concerns.

`IProjectAccessor` is identical to what `OrderingHelper.ModifyAsync` is doing, except that I want to read/write the `Project`, which `IProjectAccessor` doesn't have.

Also, how would `IProjectAccessor` be discoverable?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
TIHan: Yes. I think the reasoning behind setting it to "true" before was to accidentally prevent bad things from ever happening.

TIHan:
No problem, thank you all for reviewing. :)

Issue #3253, opened 2018-02-14
Crash when browsing for executable in custom profile


mattscheffer:
Repro steps:
Latest D15.6 preview 5 build
Install .NET 2.1 latest preview

Copy solution from [internalshare]\public\v-masche\launchsettings
Open the solution, open the properties page and navigate to the debug tab.
Change the profile to newProfile1
Click the browse button for the executable.
At this point VS crashes.

I've not been able to repro this outside of this solution. The crash dump is in the same folder as the solution.

davkean:
``` C#
                           var file = ExecutablePath;
                            if ((file.IndexOfAny(Path.GetInvalidPathChars()) == -1) && Path.IsPathRooted(file)) // NullReferenceException because file is null
                            {
                                dialog.InitialDirectory = Path.GetDirectoryName(file);
                                dialog.FileName = file;
                            }
                            dialog.Multiselect = false;
                            dialog.Filter = _executableFilter;
                            var result = dialog.ShowDialog();
                            if (result == System.Windows.Forms.DialogResult.OK)
                            {
                                ExecutablePath = dialog.FileName.ToString();
                            }
```

davkean:
Fix in review.

PR #3254, opened 2018-02-14
Avoid crash when loading profile without executable

davkean:
Fixes: #3253

Avoids crash when browse for an executable when we load a profile without an executable path. This can occur when the user didn't set an initial one, or removed it explicitly from the file.

All launch profile properties can be null - searched the code base for other similar usages, and this was the only one that didn't handle null.


Pilchie:
@dotnet/roslyn-infrastructure - windows_debug timed out, and apparently we don't capture dmp files when that happens?

Pilchie:

PR #3268, opened 2018-02-15
Port version fixes from RepoToolset to 15.5

natidea:
Fixing the official build version to not longer go over 65535

Infrastructure only change.
Confirmed this change fixes build break: [20180215.7](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=1390168&_a=summary)

/cc @Pilchie @tmat @tannergooding

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -20,6 +20,26 @@
     <RemoveIntegerChecks>true</RemoveIntegerChecks>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


Pilchie:
And confirmed that it generates the right build version?

natidea:
Yes, the versions appear to match values from other branches:

image

tannergooding:
This change should be dropped when it is merged to master, as newer versions of repo-toolset contains this logic itself.

FYI. @natidea, @Pilchie

PR #3270, opened 2018-02-16
Replace usage of IProjectXmlAccessor with IProjectAccessor

davkean:
- Replace usage of IProjectXmlAccessor with IProjectAccessor in configuration providers
- Delete IProjectXmlAccessor because it now unused

Also ported 8175e46c which this changed needed.


davkean:
Tested this out, add/removing/renaming configurations works like a charm.

davkean:
Rebased this against 15.7.x.

PR #3278, opened 2018-02-19
Apply formatting rules across the tree

davkean:
Applied the formatting rules across the tree. Kept running into these as I was making changes, rather than individually apply them I've just applied them against the entire codebase.

Note, I've turned on the new enforcement of a blank line between using groups - I left it as `true` because that's the default, but I'm okay either way if people feel strongly one way or the other (left it as a separate commit that I can undo).


davkean:
I've tried to not touch any tables, I avoided formatting the test code because of it - please tell me if you see a table that I've broken.

Pilchie:
Consider targeting 15.7.x, so this doesn't conflict with everything we do for 15.7.

davkean:
I'll retarget this one against 15.7.x - but I gave up on the others, too many dependencies already in master.

davkean:
I retargeted and ended up going and fixing all the tests code undoing any formatting that screwed up tables. I'll also port this to master and reapply to catch any missed changes.

PR #3284, opened 2018-02-20
Stop checking for null for MEF-provided components

davkean:
Consistently stop checking that MEF-provided components are not-null. Very few of the team actually did it - resulting in a confusing mix of both styles. Remove all checks/tests that covered said checks.


src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/CSharpProjectGuidProvider.cs
```diff
@@ -19,7 +19,6 @@ internal class CSharpProjectGuidProvider : IItemTypeGuidProvider
         [ImportingConstructor]
         public CSharpProjectGuidProvider(UnconfiguredProject unconfiguredProject)
```
sharwell: :bulb: You can improve the resistance of this type to incorrect use by applying one of the following approaches in addition to the change already made:

1. Mark the constructor `private` instead of `public`
2. Use `[Obsolete("This type may only be obtained through the MEF container", error: true)]`

davkean: We use the constructors in tests.

davkean:
These are internal types - they are not a public API contract.

sharwell:
As a style decision I wanted the chance to explain my view more fully. This is a situation I encountered in the past, which influences my perspective. In that other project, every once in a while test interactions with exported components would deviate from production patterns just enough to let a bug ship. We ended up moving to a pattern that originally seemed impossible:

1. Only one container is allowed to exist for the duration of a single test
2. An object can either be obtained from the container, or not. Under no circumstances will a type ever deviate from this (including for testing).

It turned out the pattern was not only easy to follow in the end, but it was exceptionally reliable. We solved the "constructors used in tests" issue by controlling the container configuration prior to first use. The pattern was remarkably successful, and caught fascinating bugs where a test unintentionally used two different instances of the same type which was reasoned to be a singleton - one constructed by the test and the other obtained by some other code via the container. In some cases the test was just broken (not testing what we thought we were testing), while in other cases correcting the situation fixed the test and revealed a production bug.

Obviously this example is a different problem than the one addressed by this pull request. However, a move in the direction of what I describe naturally resolves the code style issue by making it impossible to invoke the constructor of an exported component will null arguments. This static guarantee is what I like to see when argument validation is removed from APIs, including internal-only APIs like this one. In the absence of statically-known behavior, I tend to prefer validation remain. Eventually the code will be a very interesting case study into C# 8 null analysis, and it might make sense to provide analyzers that detect validation of signature contracts on internal API boundaries, and offer to remove the validation.

As an observer on this repository (as opposed to a core developer or reviewer), I'm dismissing my Request Changes review on the basis that I was given the opportunity to state my view. This view is an opinion which may or may not be appropriate for this repository, and it doesn't make sense for me to claim a stakeholder position in the final outcome. I do appreciate the chance to voice an alternate viewpoint though. 😄

davkean:
Since the project started, I've been a big proponent of adding these checks. However:

- I've yet to see them catch any real issue
- They aren't enforcing any public APIs (I only removed from MEF constructors, nothing else)
- Almost all contributors (even me in recent changes) skip adding these checks

At the moment they are just noise. When C# 8.0 comes along and adds non-null reference types, we'll add the appropriate annotations - but until then I've just consistently applied what appears to be the consensus based on the changes I've seen contributed over the past three years.

Issue #3286, opened 2018-02-20
Integration tests are failing with "Message: System.ArgumentException : The parameter is incorrect. (Exception from HRESULT: 0x80070057 (E_INVALIDARG))"


davkean:
All the currently enabled integrations tests are failing with the following both on dev boxes and Jenkins:

```
Test Name:	AddPackageReference
Test FullName:	Microsoft.VisualStudio.ProjectSystem.IntegrationTests.CSharpProjectFileEditTest.AddPackageReference
Test Source:	E:\project-system2\src\Microsoft.VisualStudio.ProjectSystem.IntegrationTests\CSharp\CSharpProjectFileEditTest.cs : line 65
Test Outcome:	Failed
Test Duration:	0:00:00.001

Result StackTrace:	
at EnvDTE.Commands.Item(Object index, Int32 ID)
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstance.<>c__DisplayClass85_0.<StartRemoteIntegrationService>b__0()
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstance.<>c__DisplayClass88_0`1.<RetryRpcCall>b__0()
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstance.RetryRpcCall(Action action)
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstance.RetryRpcCall[T](Func`1 action)
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstance.StartRemoteIntegrationService(DTE dte)
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstance..ctor(Process hostProcess, DTE dte, ImmutableHashSet`1 supportedPackageIds, String installationPath)
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstanceFactory.UpdateCurrentlyRunningInstance(ImmutableHashSet`1 requiredPackageIds, Boolean shouldStartNewInstance)
   at Microsoft.VisualStudio.IntegrationTest.Utilities.VisualStudioInstanceFactory.GetNewOrUsedInstance(ImmutableHashSet`1 requiredPackageIds)
   at Microsoft.VisualStudio.ProjectSystem.IntegrationTests.AbstractIntegrationTest..ctor(String solutionName, String projectTemplate, VisualStudioInstanceFactory instanceFactory)
[...]

PR #3287, opened 2018-02-20
Disable integration tests

davkean:
We've never enabled the integration tests as a check-in blocker - nor have they ever passed as far as I can see on the jenkins runs - every single build that it has history on has failed. Disable them to avoid reduced confidence by those that are new to the tree.

Filed https://github.com/dotnet/project-system/issues/3286 to track reenabling them.


davkean:
This lets someone choose "Run All" and have confidence that any failed test is caused by their changes.

PR #3289, opened 2018-02-20
Suppress usage of JoinableTaskFactory.Run

davkean:
This is a valid usage - only wrapped for test purposes.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/DebugFrameworksDynamicMenuCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/DebugFrameworksDynamicMenuCommand.cs
@@ -106,7 +106,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
         /// </summary>
         protected virtual void ExecuteSynchronously(Func<Task> asyncFunction)
         {
+#pragma warning disable VSTHRD102 // Only wrapped for test purposes
             ThreadHelper.JoinableTaskFactory.Run(async () =>
+#pragma warning restore VSTHRD102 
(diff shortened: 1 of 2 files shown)


sharwell:
@davkean Is there a reason why this *cannot* be corrected? Differences in threading patterns between test and production code runs a high risk for failing to detect bugs in production before they ship.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/DebugFrameworksDynamicMenuCommand.cs
```diff
@@ -106,7 +106,9 @@ public override bool QueryStatusCommand(int cmdIndex, EventArgs e)
         /// </summary>
         protected virtual void ExecuteSynchronously(Func<Task> asyncFunction)
```
sharwell: ❓ Why can this not be changed to this?

```csharp
protected virtual Task ExecuteAsync(Func<Task> asyncFunction)
{
  return await asyncFunction().ConfigureAwait(false);
}
```

This change would allow tests to override the implementation and also correct the underlying warning without a suppression. The use of `JoinableTaskFactory.Run` in the caller of this method should not be an impediment to testing, especially if test code overrides this method to return the result synchronously.

davkean: ThreadHelper expects to be initialized on a pumping UI thread, this avoids that.

sharwell: :bulb: This would be good to put in the documentation for the method.

PR #3290, opened 2018-02-20
Fix problem with up-to-date check and custom items

panopticoncentral:
**Customer scenario**

The customer adds a XAML file to their SDK project. From that point on, the up to date check believes the project is never up to date.

(This also improves the up to date check logging so it is more obvious why a check failed.)

**Bugs this fixes:** 

#2763

**Workarounds, if any**

Remove the XAML file.

**Risk**

Low. We're avoiding a previously unhandled exception.

**Performance impact**

Improves performance, avoids unhandled exception.

No

**How was the bug found?**

Customer report

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -344,28 +344,28 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
 
             if (!_tasksService.IsTaskQueueEmpty(ProjectCriticalOperation.Build))
             {
-                return Fail(logger, "Critical build tasks are running, failed.", "CriticalTasks");
+                return Fail(logger, "Critical build tasks are running, not up to date.", "CriticalTasks");
             }
 
             if (_lastVersionSeen == null || _configuredProject.ProjectVersion.CompareTo(_lastVersionSeen) > 0)
             {
-                return Fail(logger, "Project information is older than current project version, failed.", "ProjectInfoOutOfDate");
+                return Fail(logger, "Project information is older than current project version, not up to date.", "ProjectInfoOutOfDate");
             }
 
             if (itemsChangedSinceLastCheck)
             {
-                return Fail(logger, "The list of source items has changed since the last build, failed.", "ItemInfoOutOfDate");
+                return Fail(logger, "The list of source items has changed since the last build, not up to date.", "ItemInfoOutOfDate");
             }
 
             if (_isDisabled)
             {
-                return Fail(logger, "The 'DisableFastUpToDateCheckProperty' property is true, failed.", "Disabled");
+                return Fail(logger, "The 'DisableFastUpToDateCheckProperty' property is true, not up to date.", "Disabled");
             }
 
             var copyAlwaysItem = _items.SelectMany(kvp => kvp.Value).FirstOrDefault(item => item.CopyType == CopyToOutputDirectoryType.CopyAlways);
             if (copyAlwaysItem.Path != null)
             {
-                logger.Info("Item '{0}' has CopyToOutputDirectory set to 'Always', failed.", copyAlwaysItem.Path);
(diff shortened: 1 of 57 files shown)


Pilchie:
Is the issue here that `FullName` isn't in the rule for Xaml items?  Also, should we just be looking at `Identity` instead of `FullName`?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -339,28 +344,28 @@ private bool CheckGlobalConditions(BuildAction buildAction, BuildUpToDateCheckLo
...
             {
-                return Fail(logger, "The 'DisableFastUpToDateCheckProperty' property is true, skipping check.", "Disabled");
+                return Fail(logger, "The 'DisableFastUpToDateCheckProperty' property is true, failed.", "Disabled");
             }
 
             var copyAlwaysItem = _items.SelectMany(kvp => kvp.Value).FirstOrDefault(item => item.CopyType == CopyToOutputDirectoryType.CopyAlways);
             if (copyAlwaysItem.Path != null)
             {
-                logger.Info("Item '{0}' has CopyToOutputDirectory set to 'Always', skipping check.", copyAlwaysItem.Path);
+                logger.Info("Item '{0}' has CopyToOutputDirectory set to 'Always', failed.", copyAlwaysItem.Path);
```
sharwell: :bulb: ~~failed~~ treating as not up-to-date

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -236,7 +239,9 @@ private void OnSourceItemChanged(IProjectSubscriptionUpdate e, IProjectItemSchem
 
             foreach (var itemType in e.ProjectChanges.Where(changes => (itemTypesChanged || changes.Value.Difference.AnyChanges) && _itemTypes.Contains(changes.Key)))
             {
-                var items = itemType.Value.After.Items.Select(item => (item.Value[FullPath], GetLink(item.Value), GetCopyType(item.Value)));
+                var items = itemType.Value.After.Items
+                    .Select(item => (GetFullPath(item.Value), GetLink(item.Value), GetCopyType(item.Value)))
+                    .Where(tuple => tuple.Item1 != null);
```
sharwell: ❔ Can you briefly explain how this was interfering with the XAML check, and why it's better than the workaround I was using previously?

panopticoncentral:
@Pilchie @sharwell Sorry, I wasn't clear. The repro project has a XAML rule file declared in its project file as a XamlPropertyRule item. It's showing up in the list of source items, but we don't have a rule for it. So the up to date check is crashing. This change basically hardens us against source items that we don't know anything about.

panopticoncentral:
Yeah, it's been long enough that I can't remember why I used FullPath instead of Identity, I do believe I considered it. Off the top of my head, maybe identity doesn't always include the full path? It's a little hazy. I can make a note to go back and look at it.

sharwell:
@panopticoncentral By the title, I was expecting this pull request to correct this: dotnet/roslyn-sdk#99

Is that not the case?

panopticoncentral:
Sorry, my title was misleading. It does not fix those issues.

panopticoncentral:
@dotnet/project-system Can someone review?

Pilchie:
I'd like to know if we could just switch to `Identity` instead of `FullPath` all together.  @davkean will a dataflow rule always have `Identity`?

lifengl:
Can you use item.Key, when item.Value[FullPath] does not exist.

davkean:
You'll need to resolve it via the project `UnconfiguredProject.MakeRooted`.

lifengl:
That will be the evaluated path of each item in the msbuild.  Frankly, I didn't realize you are using FullPath here, and I am thinking whether we can get rid of it in the rule subscription.  That property is lots of long strings in the memory, and msbuild evaluates it in a lazy way, so the property in the rule means we recreate the string on every reevaluation.

panopticoncentral:
@lifengl Does this look better? I moved all uses of FullPath to just use item.Key + MakeRooted. I also removed the properties on items that I had added it to since we don't need it.

panopticoncentral:

PR #3292, opened 2018-02-21
Target .NET 4.6.1 instead of 4.6.0.

Pilchie:
The Roslyn Workspaces and above layers all depend on 4.6.1 already, and all
workloads that include the project system also require it.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -27,13 +27,7 @@
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
 
     <!-- All projects in this repository target the same framework by default -->
-    <TargetFramework Condition="'$(IsTestProject)' != 'true'">net46</TargetFramework>
-
-    <!--
-      All tests projects in this repository target another. This includes the DeployTestDependencies
-      and DeployIntegrationDependencies projects which must be updated independently.
-    -->
-    <TargetFramework Condition="$(MSBuildProjectName.EndsWith('.UnitTests')) or $(MSBuildProjectName.EndsWith('.IntegrationTests'))">net461</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
 
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
 
--- a/src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
+++ b/src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
@@ -5,7 +5,7 @@
     <!-- TODO: Function doesn't return a value on all code paths (https://github.com/dotnet/project-system/issues/2592) -->
     <NoWarn>$(NoWarn);42353</NoWarn>
     <StrongNameKeyId>Microsoft</StrongNameKeyId>
-
+    <TargetFramework>net46</TargetFramework>
     <!-- Nuget -->
     <IsPackable>true</IsPackable>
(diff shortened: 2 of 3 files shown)


Pilchie:
@jmarolf - can you verify that I'm correct in my claims about workloads?

davkean:
How do we figure that out? I was under the impression that minimum was still 4.6.

Pilchie:
We figure it out by looking at the setup authoring of the workloads.  Roslyn Workspaces and above, and all workloads that include them target 4.6.1 or higher.  Core Editor and others like Python, etc, only require 4.6.

Pilchie:
Yep.  Agree that this might not be right.  @jaredpar also just corrected a misunderstanding I had above :-/

Pilchie:
Alternate in https://github.com/dotnet/project-system/pull/3293

jaredpar:
Here is the state of the world:

- The "Managed Desktop" workload forces 4.6.1 to be installed. What forced this change is unclear to me. @tmat may know since it appears to be driven by interactive.
- dotnet/roslyn targets the following TFs 
    - net461: the interactive layer 
    - net46: everything else


Logically the dotnet/roslyn repo seems to be it the wrong here. VS setup authoring guarantees we are on 4.6.1 for most of these components (sans MSBuild workspace). Hence targeting 4.6.1 should be just fine.

davkean:
Do Editors/AppDesigner get installed elsewhere other than the Managed Desktop workload? Teach me to fish - how would I figure that out?

jaredpar:
> Do Editors/AppDesigner get installed elsewhere other than the Managed Desktop workload? Teach me to fish - how would I figure that out?

Don't forget that you asked for this ...  

I will send you the fishing instructions I found.

davkean:
I'm figuring out if all the places we install pre-req 4.6.1...

davkean:
This isn't a safe change to make without a little more investigation: https://gist.github.com/davkean/b50cd3ac282f6d5f42bcd5d79355df43.

Microsoft.VisualStudio.Editors is pulled in by almost all workloads.
Microsoft.VisualStudio.ProjectSystem.Managed is pulled in by other workloads, including Microsoft.VisualStudio.Workload.Azure.

@jaredpar do you know if Microsoft.Net.PackageGroup.4.6.1.Redist representing the 4.6.1 redist getting installed?

davkean:
I'm made the assumption that `Microsoft.Net.PackageGroup.4.6.1.Redist` represents the 4.6.1 redist, if so, the following workloads install Editors, but do you not install 4.6.1 redist:
```
Microsoft.VisualStudio.Workload.CoreEditor
Microsoft.VisualStudio.Workload.FeedbackClient
Microsoft.VisualStudio.Workload.NativeCrossPlat
Microsoft.VisualStudio.Workload.NativeMobile
Microsoft.VisualStudio.Workload.NetCoreBuildTools
Microsoft.VisualStudio.Workload.Node
Microsoft.VisualStudio.Workload.NodeBuildTools
Microsoft.VisualStudio.Workload.TestAgent
Microsoft.VisualStudio.Workload.TestController
Microsoft.VisualStudio.Workload.TestProfessional
Microsoft.VisualStudio.Workload.WebCrossPlat
```

I've confirmed that anywhere that installs Microsoft.VisualStudio.ProjectSystem.Managed also installs the redist. It sounds like the Editors/AppDesigner layer cannot target 4.6.1.

jaredpar:
> I'm made the assumption that Microsoft.Net.PackageGroup.4.6.1.Redist 

This is the same assumption that I've been operating under.

davkean:
@Pilchie I've submitted a change to this PR that changes Editors/AppDesigner to 4.6.

Pilchie:
Thanks @davkean - @jmarolf any objections?

PR #3293, opened 2018-02-21
Target .NET 4.6 instead of a mix 4.6 and 4.6.1.

Pilchie:
The Roslyn Workspaces and above layers all depend on 4.6.1 already, and all
workloads that include the project system also require it.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -27,13 +27,7 @@
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
 
     <!-- All projects in this repository target the same framework by default -->
-    <TargetFramework Condition="'$(IsTestProject)' != 'true'">net46</TargetFramework>
-
-    <!--
-      All tests projects in this repository target another. This includes the DeployTestDependencies
-      and DeployIntegrationDependencies projects which must be updated independently.
-    -->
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -59,4 +53,4 @@
     product artifacts into wrong obj directories. Remove once we can use a custom SDK (https://github.com/Microsoft/msbuild/issues/1493)
   -->
   <Import Project="build\RepoToolset\ProjectLayout.props"/>
-</Project>
\ No newline at end of file
+</Project>
```
davkean: Nit: Trailing new line was removed.

davkean:
Ugh:

```
error NU1202: Package Microsoft.VisualStudio.IntegrationTest.Utilities 2.6.0-beta1-62113-02 is not compatible with net46 (.NETFramework,Version=v4.6). Package Microsoft.VisualStudio.IntegrationTest.Utilities 2.6.0-beta1-62113-02 supports: net461 (.NETFramework,Version=v4.6.1) [D:\j\workspace\windows_debug---a4dca817\ProjectSystem.sln]
```

Pilchie:
Closing, since this seems harder than first thought, and I'd prefer #3292 if it is safe.

PR #3299, opened 2018-02-22
Replace "CSharp | VB | FSharp" usage with ".NET"

davkean:
This replaces all usage of specific languages in our components to the new ".NET" capability. This enables other .NET languages to opt into our feature set.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectCapability.cs
```diff
@@ -36,5 +36,6 @@ internal static class ProjectCapability
         public const string ProjectConfigurationsDeclaredDimensions = ProjectCapabilities.ProjectConfigurationsDeclaredDimensions;
         public const string LanguageService = ProjectCapabilities.LanguageService;
         public const string SortByDisplayOrder = ProjectCapabilities.SortByDisplayOrder;
+        public const string DotNet = ".NET";
```
Pilchie: Don't you need to remove the one on line 14?

davkean: yeah that would help.

davkean:
Just testing this.

davkean:
This breaks shared projects, which don't pull use our factory registration.

davkean:
We need to make https://github.com/dotnet/project-system/issues/3311 before we check this in.

davkean:
Ready for review, Shared Projects is fixing in: https://github.com/dotnet/project-system/pull/3521

Issue #2940, opened 2017-11-06
Perf: extra overhead during set project Guid


lifengl:
In a standard perf box, it wasted 500-800ms UI thread time in this code path to do nothing.

The overhead is mostly to get a read lock and switch to a thread pool to scan the project and check whether the GUID is ever persisted.  It was scanned once in CPS to retrieve the GUID during the earlier phase of loading the project.

CPS defined a new ProjectGuidStorageProvider contract to optimize that in 15.6.  Please work with the contract to eliminate this wasted time.

davkean:
- Copy the default provider from CPS
- Consume this provider inside our in-memory properties provider instead of reading project file

Pilchie:
Moving to 15.7 - I took a quick look at this, but it wasn't totally clear to me how to do this.

davkean:
Just removing our usage of the rule before removing it in lieu of the IProjectGuidProvider/IProjectGuidStorageProvider; https://github.com/dotnet/project-system/pull/3303, https://github.com/dotnet/project-system/pull/3314 and https://github.com/dotnet/project-system/pull/3313.

Pilchie:
Fixed now, right @davkean?

Pilchie:
Fixed now, right @davkean ?

davkean:
Ah yes.

PR #3303, opened 2018-02-23
Simplify StartupProjectRegistrar and remove ProjectGuid property usage

davkean:
- Removed need to keep state if the project is debuggable or not, IVsStartupProjectsListService is designed to no op if the project has already been added or removed.
- Removed need to switch to UI thread. IVsStartupProjectsListService is free-threaded.
- Stop listening to project GUID changes - the project GUID rule is being removed to fix https://github.com/dotnet/project-system/issues/2940.
- Listen to all properties not just output type. If not using launch profiles, output type doesn't actually influence whether you are debuggable or not. It's based on the debugger properties such as executable, etc.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
@@ -89,7 +89,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             return Task.CompletedTask;
         }
 
-        internal async Task OnProjectChangedAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> e)
+        internal async Task OnProjectChangedAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> e = null)
         {
             bool isDebuggable = await _launchProviders.Value.IsDebuggableAsync()
                                                             .ConfigureAwait(false);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
@@ -56,7 +56,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
 
         [ProjectAutoLoad(startAfter: ProjectLoadCheckpoint.ProjectFactoryCompleted)]
         [AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharp)]
-        internal Task Load()
+        public Task OnProjectFactoryCompleted()
         {
             return InitializeAsync();
         }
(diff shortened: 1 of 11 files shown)


davkean:
Need to fix tests.

davkean:
There's a little more diff in this than I'd like because I rewrote it and then replaced the existing one. I'll reduce the diff there's concerns.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -2,150 +2,135 @@
...
     /// </summary>
-    internal class StartupProjectRegistrar : OnceInitializedOnceDisposed
+    internal class StartupProjectRegistrar : OnceInitializedOnceDisposedAsync
     {
+        private readonly IAsyncServiceProvider _serviceProvider;
         private readonly IProjectThreadingService _threadingService;
-        private readonly IActiveConfiguredProjectSubscriptionService _activeConfiguredProjectSubscriptionService;
+        private readonly IActiveConfiguredProjectSubscriptionService _projectSubscriptionService;
         private readonly ActiveConfiguredProject<DebuggerLaunchProviders> _launchProviders;
-        private readonly IVsService<IVsStartupProjectsListService> _startupProjectsListService;
```
davkean: IVsService enforces the service be accessed on the UI thread - which is not want we want here. EIther I need to remove it and in lieu of IAsyncServiceProvider long term, or change its pattern to "async".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -2,150 +2,135 @@
...
 
-        protected override void Initialize()
+        protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
         {
-            var watchedEvaluationRules = Empty.OrdinalIgnoreCaseStringSet.Add(ConfigurationGeneral.SchemaName);
-            var evaluationBlock = new ActionBlock<IProjectVersionedValue<IProjectSubscriptionUpdate>>(ConfigurationGeneralRuleBlock_ChangedAsync);
-            _evaluationSubscriptionLink = _activeConfiguredProjectSubscriptionService.ProjectRuleSource.SourceBlock
-                                            .LinkTo(target: evaluationBlock, ruleNames: watchedEvaluationRules, suppressVersionOnlyUpdates: true);
-        }
+            _projectGuid = await _projectGuidService.GetProjectGuidAsync()
```
davkean: We can assume the GUID never changes over the lifetime of a project one the factory has completed.

davkean:
Fixed tests, ready for review.

davkean:
Hit: https://github.com/dotnet/project-system/pull/3308.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -2,150 +2,138 @@
...
-                    await AddOrRemoveProjectFromStartupProjectListAsync(initialize: true).ConfigureAwait(false);
-                }
+            Assumes.Present(_startupProjectsListService);
 
-                return;
-            }
+            _subscription = _projectSubscriptionService.ProjectRuleSource.SourceBlock.LinkTo(
+                target: new ActionBlock<IProjectVersionedValue<IProjectSubscriptionUpdate>>(OnProjectChangedAsync),
+                suppressVersionOnlyUpdates: true,
+                linkOptions: new DataflowLinkOptions() { PropagateCompletion = true });
```
Pilchie: Do you need to capture capabilities here?

davkean: Good point. @lifengl thoughts? We read from a OrderPrecedenceImportCollection from the active configuration in the handler.

davkean: I filed https://github.com/dotnet/project-system/issues/3322 to track this; I didn't change when we got initialized or when we walk the debuggers, so if the capabilities were wrong, they were always wrong. I'll do a pass over every dataflow usage and figure out if we need start a new context.

Issue #3279, opened 2018-02-19
Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests.dll are timing out during builds


davkean:
These tests are timing out during builds: 

https://ci.dot.net/job/dotnet_project-system/job/master/job/windows_release_prtest/885
https://ci.dot.net/job/dotnet_project-system/job/master/job/windows_debug/379

davkean:
I suspect the new tests I added to test dataflow, or the tests that fire up a pumping thread.

Pilchie:
Quoting @jaredpar from an email about how `roslyn` captures dumps:

>That’s correct. We essentially spawn of xunit as a process and if it doesn’t complete in ~50 minutes >then we use procdump to kill it and grab a dump.  The 50 minute timeout is roughly 4X the longest >time that xunit should take for our DLLs.
 
>Short term this is the code we use
 
>https://github.com/dotnet/roslyn/blob/master/src/Tools/Source/RunTests/Program.cs#L157

davkean:
The tests I added to test data flow aren't hitting this, I rewrote them and it still deadlocked.

davkean:
This appears to actually be Microsoft.VisualStudio.CSharp.VS.UnitTests.dll on my box, in particular:

```
Rename_Symbol_Should_TriggerUserConfirmationAsync
```

davkean:
I'm wondering aloud if this correlates with increase in rename failures that I'm seeing.

davkean:
Also seeing Microsoft.VisualStudio.Managed.VS.UnitTests hang.

333fred:
I believe https://ci.dot.net/job/dotnet_project-system/view/Official%20Builds/job/dev15.7.x/job/windows_release/49/ is this as well, likely in Microsoft.VisualStudio.Managed.VS.UnitTests

davkean:
I've checked in https://github.com/dotnet/project-system/pull/3718 which seems to have removed the hangs from my box. I also added time outs to the other tests that I thought might have been the cause here so that we can get more info: https://github.com/dotnet/project-system/pull/3714.

I'll consider this bug done and open new bugs for any new failures.

PR #3308, opened 2018-02-26
Prevent ActiveConfiguredProjectsLoader tests from deadlocking

davkean:
Fixes: #3279

These tests were using dataflow messages to test the service which I believe were deadlocking due to blocks not completing. I've not been able to reproduce locally, but we've hitting a lot since 15.6 - and these are the only thread-related tests that we've added.

Move to a simpler model of just calling the event handler.


davkean:
Looks like these are not the cause...

Issue #3187, opened 2018-01-25
build.cmd fails with spaces in the path


loop-evgeny:
```C:\W ork\project-system>build.cmd```

>C:\W : The term 'C:\W' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
> At line:1 char:1
> + C:\W ork\project-system\build\Build.ps1 -configuration Debug -restore ...
> + ~~~~
>     + CategoryInfo          : ObjectNotFound: (C:\W:String) [], CommandNotFoundException
>     + FullyQualifiedErrorId : CommandNotFoundException

**Comment by davkean** [comment]
I spent a non-trivial amount of time attempting to fix this - and gave up. Here's what I've attempted so that others can learn from my adventure:

- Tried using `-File`. [this breaks the `-FooBar $False` parsing](https://stackoverflow.com/questions/5079413/how-to-pass-boolean-values-to-a-powershell-script-from-a-command-prompt/7302626)
- Tried using `-Command` with a bunch of quotes and ampersand to work cmd and powershell parsing `powershell -ExecutionPolicy ByPass -Command "& """E:\New Folder\PathWithSpace.ps1""" -Build:$False"` 
   - This broke later, when we tried to call MSBuild with the wrong command-line

At this point I'd suggest just removing the space from your repo name.

**Comment by tannergooding** [comment]
@davkean, the second issue (breaking in the call to Msbuild) is an issue in repo toolset, where it doesn't properly quote all the paths (cc. @tmat)

**Comment by tmat** [comment]
@tannergooding Can you point to the place where the paths are not quoted properly?

**Comment by tannergooding** [comment]
Most of these properties need to be quoted: https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/tools/Build.proj#L98

**Comment by tannergooding** [comment]
There may be more, but I didn't finish investigating last night.

**Comment by tannergooding** [comment]
Hmmm, actually, this may be specific to something project-system is doing. I don't get the same for another repo-toolset based project (such as dotnet/sdk)

**Comment by tannergooding** [comment]
PR is here: https://github.com/dotnet/project-system/pull/3318.

There is something that breaks if you specify `/p:QuietRestore=true`, remove that and everything works as expected.

**Comment by tannergooding** [comment]
Looks like when `/p:QuietRestore=true` is specified, we call `Exec` rather than the `MSBuild` task:

```
[...]

PR #3318, opened 2018-02-27
Updating the build scripts to supppot paths with spaces in the name

tannergooding:
This resolves https://github.com/dotnet/project-system/issues/3187.

FYI. @tmat, @davkean, @jaredpar

### DIFF
--- a/Test.cmd
+++ b/Test.cmd
@@ -1,3 +1,3 @@
 @echo off
-powershell -ExecutionPolicy ByPass %~dp0build\Build.ps1 -test %*
+powershell -ExecutionPolicy ByPass -Command "& """%~dp0build\Build.ps1""" -test %*"
 exit /b %ErrorLevel%
--- a/build.cmd
+++ b/build.cmd
@@ -30,7 +30,7 @@ if /I "%1" == "/rootsuffix" set PropRootSuffix=/p:RootSuffix=%2&&shift&&shift&& 
 call :Usage && exit /b 1
 :DoneParsing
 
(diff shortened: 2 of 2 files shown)


tannergooding:
It appears there is something terribly broken with `/p:QuietRestore=true`, as including that parameter causes:
```
MSBUILD : error MSB1008: Only one project can be specified. [C:\Users\tagoo\.nuget\packages\RoslynTools.RepoToolset\1.0.0-beta-62615-02\tools\Build.proj]
C:\Users\tagoo\.nuget\packages\RoslynTools.RepoToolset\1.0.0-beta-62615-02\tools\Build.proj(152,5): error MSB3073: The command ""C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MS
Build.exe" /nodeReuse:false "C:\Users\tagoo\.nuget\packages\RoslynTools.RepoToolset\1.0.0-beta-62615-02\tools\Tools.proj" /bl:"D:\project system\artifacts\Debug\log\RestoreRepoTools.binlog" /nologo /m /v:quiet
/t:Restore /p:Configuration=Debug /p:CIBuild=False /p:RepoRoot=D:\project system\ /p:VersionsPropsPath=D:\project system\build\Versions.props /p:DotNetPackageVersionPropsPath= /p:BaseIntermediateOutputPath=D:\p
roject system\artifacts\toolset\Common /p:ExcludeRestorePackageImports=true /p:UsingPipeBuildPublishing=false" exited with code 1.
```

tannergooding:
Also FYI. @Pilchie, @jasonmalinowski

davkean:
The change looks good. Please separate commits next time, these are two very different things that you are fixing.

tannergooding:
@davkean. See https://github.com/dotnet/roslyn-tools/issues/181

Quoting the args is fairly trivial, but there is another "blocker" which is that any path with a trailing slash is misinterpreted. For example: `/p:RepoRoot="D:\project system\"` gets weirdly interpreted by MSBuild and breaks the rest of the argument parsing.

The "easiest" fix from there is likely to ensure the trailing slash is removed from the `Exec` command and is instead appended by the props/targets files, where required.

PR #3326, opened 2018-02-28
Avoid some async overhead in DependenciesProjectTreeProvider.OnDependenciesSnapshotChanged

### LINKED ISSUE #3319: Minor: unnecessary task chains + JTF in OnDependenciesSnapshotChanged
The code inside this method is mostly unnecessary.  BuildTreeForSnapshotAsync is not an async method (it returns immediately in all code path), so there is no need to chain a task continuation chain.  Also it is strange to wrap the code inside a JTF for the first call, but didn't do for the rest.  The JTF seems to be unnecessary here as well.

Pilchie:
Fixes #3319.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -36,7 +36,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
     {
         private readonly IDependencyTreeTelemetryService _treeTelemetryService;
         private readonly object _treeUpdateLock = new object();
-        private Task _treeUpdateQueueTask = null;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DependenciesProjectTreeProvider"/> class.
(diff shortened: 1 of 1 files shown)


Pilchie:
Also tagging @lifengl

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -366,32 +361,16 @@ private void OnDependenciesSnapshotChanged(object sender, SnapshotChangedEventAr
...
-
-                        await BuildTreeForSnapshotAsync(snapshot).ConfigureAwait(false);
-                    }).Task;
-                }
-                else
-                {
-                    _treeUpdateQueueTask = _treeUpdateQueueTask.ContinueWith(
-                        t => BuildTreeForSnapshotAsync(snapshot), TaskScheduler.Default);
-                }
+                BuildTreeForSnapshot(snapshot);
```
abpiskunov: >BuildTreeForSnapshot [](start = 16, length = 20)

Did you realize what old code was doing? It was ensuring that all snapshot updates were applied to the tree in the order they were received.. Since snapshot is updated by changes in multiple DT builds associated with unresolved and then resolved rules for dependencies. Since everything is async older snapshots could be applied to the tree before later ones potentially, which could result in weird tree state.

abpiskunov: If implementation could be changed, but the overall behavior should remain the same , i would not do this change.

---
In reply to: [171345584](https://github.com/dotnet/project-system/pull/3326#discussion_r171345584) [](ancestors = 171345584)

abpiskunov: I chatted with Lifeng offline and we agree that queue logic is needed there. The only thing that might be changed if desired is how queue is implemented - avoid async tasks there if possible. I will leave it to you to decide if you want to change that.

---
In reply to: [171345891](https://github.com/dotnet/project-system/pull/3326#discussion_r171345891) [](ancestors = 171345891,171345584)

Pilchie:
Closing based on @abpiskunov's feedback.  Thanks for the review!

PR #3329, opened 2018-03-01
Avoid reading project GUID from evaluation and taking unneeded locks

### LINKED ISSUE #2940: Perf: extra overhead during set project Guid

davkean:
Fixes #2940

An implement of IProjectGuidStorageProvider that replaces our project rule for <ProjectGuid /> property. This avoids us needing to evaluate to figuring out the project's GUID - and also avoids us from needed to take a read-lock when we are subsequentially told to set the project GUID.

The actual rule for ProjectGuid will be removed in a later commit - it's currently unused after this change.


davkean:
Verified the following scenarios:

- New project with GUID inside the project (not told to update)
- New project with no GUID (solution tells project to update it, we ignore it)
- Existing project with GUID inside the project (not told to update)
- Existing project with no GUID inside the project (solution tells project to update it, we ignore it)
- Existing project with different GUID to what the solution thinks it is (solution tells project to update it)

davkean:
Hmm, looks like ProjectGuid is still implicitly being pulled by configuration general listeners - so removing the rule will actually fix the issue.

PR #3331, opened 2018-03-01
Added fix for FSharp hotkeys in legacy projects

TIHan:
**Customer scenario**

This fixes an issue where alt+up/down hotkeys for legacy F# projects wasn't working anymore.

**Bugs this fixes:** 

https://github.com/Microsoft/visualfsharp/issues/4405

**Workarounds, if any**

Don't use hotkeys, or use Tools\Options\Environment\Keyboard and remap them (but this means you need to use different keybinding for non-Core and Core F#.

**Risk**

Low risk.

**Performance impact**

None.

Yes.

**Root cause analysis:**

How did we miss it?
Didn't check legacy projects for F#.

**How was the bug found?**

https://developercommunity.visualstudio.com/content/problem/204016/vs-156-preview6-alt-updown-arrows-do-not-repositio.html

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
@@ -97,30 +97,6 @@
     </Groups>
 
     <Buttons>
-      <Button guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveUp" priority="0x100">
-        <Parent guid="guidManagedProjectSystemOrderCommandSet" id="OrderingGroup" />
-        <CommandFlag>TextChanges</CommandFlag>
-        <CommandFlag>DynamicVisibility</CommandFlag>
-        <CommandFlag>DefaultInvisible</CommandFlag>
-        <CommandFlag>DefaultDisabled</CommandFlag>
-        <Strings>
-          <ButtonText>&amp;Move Up</ButtonText>
-          <CommandName>MoveUp</CommandName>
-        </Strings>
-      </Button>
-
-      <Button guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveDown" priority="0x101">
-        <Parent guid="guidManagedProjectSystemOrderCommandSet" id="OrderingGroup" />
-        <CommandFlag>TextChanges</CommandFlag>
-        <CommandFlag>DynamicVisibility</CommandFlag>
-        <CommandFlag>DefaultInvisible</CommandFlag>
-        <CommandFlag>DefaultDisabled</CommandFlag>
-        <Strings>
-          <ButtonText>&amp;Move Down</ButtonText>
-          <CommandName>MoveDown</CommandName>
-        </Strings>
-      </Button>
-
       <Button guid="guidManagedProjectSystemOrderCommandSet" id="cmdidAddNewItemAbove" priority="0x102">
         <Parent guid="guidManagedProjectSystemOrderCommandSet" id="AddAboveGroup"/>
         <Icon guid="guidSHLMainMenu" id="1"/>
@@ -212,11 +188,6 @@
       </Button>
     </Buttons>
   </Commands>
-
-  <KeyBindings>
-    <KeyBinding guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveUp" editor="SlnExplorerGuid" key1="VK_UP" mod1="Alt" />
-    <KeyBinding guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveDown" editor="SlnExplorerGuid" key1="VK_DOWN" mod1="Alt" />    
-  </KeyBindings>
 
   <Symbols>
     <!-- This is out managed package guid. -->
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -35,12 +35,15 @@ namespace Microsoft.VisualStudio.Packaging
         public const int DebugFrameworksCmdId = 0x3050;
 
         public const string ManagedProjectSystemOrderCommandSet = "{6C4806E9-034E-4B64-99DE-29A6F837B993}";
-        public const int MoveUpCmdId = 0x2000;
-        public const int MoveDownCmdId = 0x2001;
         public const int AddNewItemAboveCmdId = 0x2002;
         public const int AddExistingItemAboveCmdId = 0x2003;
         public const int AddNewItemBelowCmdId = 0x2004;
         public const int AddExistingItemBelowCmdId = 0x2005;
+
+        // The guid and values are from the old F# project system, don't change them.
+        public const string FSharpProjectCmdSet = "{75AC5611-A912-4195-8A65-457AE17416FB}";
+        public const int MoveUpCmdId = 0x3002;
+        public const int MoveDownCmdId = 0x3007;
 
         public const string SolutionExplorerGuid = "{3AE79031-E1BC-11D0-8F78-00A0C9110057}";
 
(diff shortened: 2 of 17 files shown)


TIHan:
I will fill out the template soon.

TIHan:
To note, for dev15.7.x, we will re-add the move/down buttons and remove the one for visualfsharp. We do this so we don't break nightlies for F#.

davkean:
Can you expand on the issue? I'm a little confused why new MoveUp/MoveDown commands would conflict with legacy F#? I can see why Keybindings would be an issue - just confused on the commands.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
```diff
@@ -214,8 +216,8 @@
   </Commands>
 
   <KeyBindings>
-    <KeyBinding guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveUp" editor="SlnExplorerGuid" key1="VK_UP" mod1="Alt" />
-    <KeyBinding guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveDown" editor="SlnExplorerGuid" key1="VK_DOWN" mod1="Alt" />    
+    <KeyBinding guid="guidManagedProjectSystemMoveOrderCommandSet" id="cmdidMoveUp" editor="SlnExplorerGuid" key1="VK_UP" mod1="Alt" />
```
davkean: If these key bindings already exist, why do you need to redefine them?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
```diff
@@ -97,7 +97,9 @@
     </Groups>
 
     <Buttons>
-      <Button guid="guidManagedProjectSystemOrderCommandSet" id="cmdidMoveUp" priority="0x100">
+      <!-- These buttons are temporarily commented out as we would have duplicate context menu items for move up/down in F# projects.
```
davkean: Please remove the code instead of commenting it out. Source control gives history.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
TIHan: Good point.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
TIHan: ok

TIHan:
@davkean There isn't issues with the commands themselves. Just the context menu items becoming duplicates when we define buttons with the same guids and ids used in legacy F# and CPS.

davkean:
I'm okay with merging the commands - they should be anyway (you don't want the same effective command shipping with different ids).  In the future though, commands like this should be invisible  by default, and only show up if the project "handles" the command - that avoids the commands getting duplicated.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
```diff
@@ -251,6 +222,14 @@
       <IDSymbol name="cmdidAddNewItemBelow" value="0x2004" />
       <IDSymbol name="cmdidAddExistingItemBelow" value="0x2005" />
     </GuidSymbol>
+
+    <!-- The guids and ids below are the same as F#'s legacy project values.
+         We do this to allow hotkey alt+up/down to work in legacy and CPS projects.
+         When F# fully switches over to CPS for older projects, we can get rid of this and use guidManagedProjectSystemOrderCommandSet instead. -->
```
Pilchie: I don't understand this comment?  `guidManagedProjectSystemOrderCommandSet` is what you are defining.

Pilchie: Nevermind I see now.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -35,13 +35,16 @@ internal partial class ManagedProjectSystemPackage : AsyncPackage
...
         public const string ManagedProjectSystemOrderCommandSet = "{6C4806E9-034E-4B64-99DE-29A6F837B993}";
-        public const int MoveUpCmdId = 0x2000;
-        public const int MoveDownCmdId = 0x2001;
         public const int AddNewItemAboveCmdId = 0x2002;
         public const int AddExistingItemAboveCmdId = 0x2003;
         public const int AddNewItemBelowCmdId = 0x2004;
         public const int AddExistingItemBelowCmdId = 0x2005;
 
+        // These guids and ids represent F# legacy project ones. This is temporary until F# projects switch over to CPS.
+        public const string ManagedProjectSystemMoveOrderCommandSet = "{75AC5611-A912-4195-8A65-457AE17416FB}";
```
Pilchie: As I said before, I don't think these should be temporary - once all F# projects switch over, we should still use the old command so we don't break any customers that have customized keybindings, menus, etc.

TIHan: Ok. I see what you mean. I'll update the comments.

TIHan:
@dotnet-bot test this please

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -35,13 +35,16 @@ internal partial class ManagedProjectSystemPackage : AsyncPackage
...
         public const string ManagedProjectSystemOrderCommandSet = "{6C4806E9-034E-4B64-99DE-29A6F837B993}";
-        public const int MoveUpCmdId = 0x2000;
-        public const int MoveDownCmdId = 0x2001;
         public const int AddNewItemAboveCmdId = 0x2002;
         public const int AddExistingItemAboveCmdId = 0x2003;
         public const int AddNewItemBelowCmdId = 0x2004;
         public const int AddExistingItemBelowCmdId = 0x2005;
 
+        // These guids and ids represent F# legacy project ones too.
+        public const string ManagedProjectSystemMoveOrderCommandSet = "{75AC5611-A912-4195-8A65-457AE17416FB}";
```
davkean: What does F# call this? Call it exactly.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
```diff
@@ -251,6 +222,12 @@
       <IDSymbol name="cmdidAddNewItemBelow" value="0x2004" />
       <IDSymbol name="cmdidAddExistingItemBelow" value="0x2005" />
     </GuidSymbol>
+
+    <!-- The guids and ids below are the same as F#'s legacy project values. -->
+    <GuidSymbol name="guidManagedProjectSystemMoveOrderCommandSet" value="{75AC5611-A912-4195-8A65-457AE17416FB}">
```
davkean: Remove unreferenced symbols.

TIHan:
Changes are done.

davkean:
Can you file a bug to track merging all the commands? I suspect MoveUp/MoveDown is only an issue here because they are the only ones with bindings, rigth?

davkean:
@Pilchie Can you revisit the review?

TIHan:
Made an issue: https://github.com/dotnet/project-system/issues/3344

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -35,13 +35,15 @@ internal partial class ManagedProjectSystemPackage : AsyncPackage
...
 
         public const string ManagedProjectSystemOrderCommandSet = "{6C4806E9-034E-4B64-99DE-29A6F837B993}";
-        public const int MoveUpCmdId = 0x2000;
-        public const int MoveDownCmdId = 0x2001;
         public const int AddNewItemAboveCmdId = 0x2002;
         public const int AddExistingItemAboveCmdId = 0x2003;
         public const int AddNewItemBelowCmdId = 0x2004;
         public const int AddExistingItemBelowCmdId = 0x2005;
 
+        public const string FSharpProjectCmdSet = "{75AC5611-A912-4195-8A65-457AE17416FB}";
```
Pilchie: I'd suggest a comment here saying these are duplicated from the old project system, so that people don't change them.

TIHan: done

Pilchie:
Approved to merge via internal bug [574576](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/574576)

PR #3346, opened 2018-03-06
Add tests for AbstractEvaluationCommandHandler

davkean:
About to make changes to this class to fix a hang, and it has no tests. Add some to cover some scenarios.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandlerTests.ConcreteAbstractEvaluationCommandLineHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandlerTests.ConcreteAbstractEvaluationCommandLineHandler.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandlerTests.ConcreteAbstractEvaluationCommandLineHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.Collections.Immutable;
+using System.Collections.ObjectModel;
+using Microsoft.VisualStudio.ProjectSystem.Logging;
+
+namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
+{
+    partial class AbstractEvaluationCommandLineHandlerTests
+    {
+        private class ConcreteAbstractEvaluationCommandLineHandler : AbstractEvaluationCommandLineHandler
```
Pilchie: Worst name ever?

davkean: Ha, that made me lol - didn't think about the name just appended concrete to original name without thinking. Still getting used to the "Abstract" naming which came from Roslyn.

davkean: Fixed.

Issue #3218, opened 2018-02-02
Opening an F# project containing two items that refer to the same file deadlocks VS


KevinRansom:
1. `git clone https://github.com/btrepp/vspreviewcrash`
2. Download https://www.nuget.org/api/v2/package/Paket/5.148.0, rename to zip and open
2. Copy contents of tools folder to vspreviewcrash\dotnet\.paket
3. `cd vspreviewcrash\dotnet`
4. `.paket\paket install`
5. `dotnet restore`
6. `VSPreviewCrash.sln`

Original Issue: __https://github.com/Microsoft/visualfsharp/issues/4300__
I guess this is CPS or Project System, at least looking at the stacks involved.

Pilchie:
From the F# issue:

```
=====================
2/02/2018 10:52:10 AM
Recoverable
System.AggregateException: One or more errors occurred. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.ProjectSystem.UpToDate.BuildUpToDateCheck.Initialize()
   at Microsoft.VisualStudio.ProjectSystem.OnceInitializedOnceDisposed.EnsureInitialized(Boolean allowInitializing)
   at Microsoft.VisualStudio.ProjectSystem.UpToDate.BuildUpToDateCheck.Load()
   at Microsoft.VisualStudio.ProjectSystem.ConfiguredProjectImpl.<>c__DisplayClass108_2.<<InitializeAsync>b__6>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.ProjectSystem.ProjectMultiThreadedService.<>c__DisplayClass28_1.<<Fork>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.ProjectSystem.UpToDate.BuildUpToDateCheck.Initialize()
[...]

PR #3350, opened 2018-03-06
Fixed duplicate items for FSharp

TIHan:
**Customer scenario**

Customers sometimes reference the same file in projects.

**Bugs this fixes:** 

https://github.com/dotnet/project-system/issues/3218

**Workarounds, if any**

Don't include duplicate files in the same project.

**Risk**

Very Low

**Performance impact**

None

Yes.

**Root cause analysis:**

We missed it due to no testing duplicate files. In the future, we should try to get more F# users on the preview builds.

**How was the bug found?**
Customer report

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Order/TreeItemOrderPropertyProviderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Order/TreeItemOrderPropertyProviderTests.cs
@@ -42,6 +42,39 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Order
             Assert.Equal(expectedOrder, values.DisplayOrder);
         }
 
+        private List<(string type, string include)> _simpleOrderFileDuplicate = new List<(string type, string include)>
+        {
+            ("Compile", "Order.fs"),
+            ("Compile", "Customer.fs"),
+            ("Compile", "Program.fs"),
+            ("Compile", "Program.fs")
+        };
+
+        [Theory]
+        [InlineData("Customer.fs", "Compile", false, "X:\\Project\\Customer.fs", 2)]
+        [InlineData("order.fs", "Compile", false, "X:\\Project\\order.fs", 1)] // case insensitive
+        [InlineData("Program.fs", "Compile", false, "X:\\Project\\Program.fs", 3)]
+        [InlineData("Misc.txt", "Content", false, "X:\\Project\\Misc.txt", int.MaxValue)] // unknown type
+        [InlineData("ordered.fsproj", null, false, "X:\\Project\\ordered.fsproj", int.MaxValue)] // hidden file
+        [InlineData("Debug", null, true, null, 0)] // unknown folder
+        public void VerifySimpleOrderedUnderProjectRootDuplicate(string itemName, string itemType, bool isFolder, string rootedPath, int expectedOrder)
+        {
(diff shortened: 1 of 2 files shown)


davkean:
Given the `<None Include="*.*"/>` that is including fs files - doesn't this mean that file ordering is completely based on the order of which these files are found on disk?

davkean:
Tests to cover this case?

TIHan:
Working on a test case.

TIHan:
@davkean If you use `<None Include="*.*"/>`, then ordering of the `.fs` files will be entirely based on alphabetical order. I'm assuming we want to change that.

davkean:
It will be file ordering - ie whatever the disk returns, which may be alphabetical.

What I mean is that, is this a complete fix? The file order for projects like this is going to be broken.

TIHan:
I see what you are saying. Let me look at this.

davkean:
I think we should opt F# projects out of all globs, or stop *.fs from being included in <None/>.

TIHan:
We opt out of globbing. The issue most likely is happening in FSharp.NET.Sdk, which is not supported.

Pilchie:
@TIHan isn't the issue that F# opts out of default `Compile` globbing, but not default `None` globbing?

@KevinRansom ?

TIHan:
@Pilchie EnableDefaultNoneItems is false in FSharp.Net.SDK.props.

davkean:
I just did a /pp on this project with and without the reference to FSharp.NET.SDK.props, nothing opts out EnableDefaultNoneItems.

KevinRansom:
David, that project is an FSharp.Net.SDK project, we don't own or ship 'FSharp.Net.Sdk'.  It is owned by @enricosada .  It ships on nuget and can be found here:  https://www.nuget.org/packages/FSharp.NET.Sdk/

I can easily believe that they don't opt out of DefaultNoneItems.

I believe we had to opt out as a bug fix.  The Microsoft FSharp.Net targets and propts do opt out
they can be found here:   https://github.com/Microsoft/visualfsharp/blob/master/src/fsharp/FSharp.Build/Microsoft.FSharp.NetSdk.props#L34

Kevin

KevinRansom:
@davkean 
Here is their props file:  https://github.com/dotnet/netcorecli-fsc/blob/master/src/FSharp.NET.Sdk/build/FSharp.NET.Sdk.props

As you can see no opt out of DefaultNone.

If you want, we can update our targets, to opt out for them, if we see that they are running, but I'd rather  not.

Kevin

davkean:
I did a /pp on this project and it does not import Microsoft.CSharp.NetSdk.props: https://gist.github.com/davkean/82433b1dc39b500e222a592180618290.

davkean:
Interestingly enough, a new project does: https://gist.github.com/davkean/a8a2c348d72696fc70f8401f29532ad0.

KevinRansom:
@davkean The new project created with vs or the dotnet cli, uses our targets which explicitly opt out.  The repro uses the FSharp.Net.SDK targets which don't opt out.  When our targets see that the project is an fsharp.netsdk project they get out of the way and do nothing.

davkean:
@KevinRansom As I mentioned, I removed FSharp.NET.SDK from the project before I ran /pp.

KevinRansom:
@davkeen,  did you clean the obj and bin directory, because the nugget package adds targets and props that are autogenerated on install of the nugget package.

davkean:
Yep that was the cause, cleaning the project resulted in the None glob being opt'd out.

Remaining questions; 

- Is this SDK even relevant anymore?
- Should this SDK prevent us from including our targets?

I'm okay with the change based on above.

Pilchie:
Approved to merge via internal bug 577380.

PR #3334, opened 2018-03-01
Remove SDK references from tests

davkean:
Works around https://github.com/Microsoft/msbuild/issues/2997.

Back ports this from master, and fixes the additional ones that were added this week.

PR #3353, opened 2018-03-07
Remove SDK references from tests

davkean:
Works around https://github.com/Microsoft/msbuild/issues/2997.

This was supposed to be this: https://github.com/dotnet/project-system/pull/3334 - but accidentally merged into master instead of dev15.7.x.


jeffkl:
The stack trace shows a class that does not exist anymore.  Are you sure the tests are using the correct MSBuild assemblies?

davkean:
@jeffkl Regardless of whether we're using older or newer assemblies, why is the VS installation breaking the tests?

jeffkl:
I don’t know. I need more info. I can’t repro it

davkean:

PR #3354, opened 2018-03-07
Try to be free threaded

Pilchie:
For @jasonmalinowski - I've done zero validation whatsoever, but it builds on my machine.

<details>**Customer scenario**

common a scenario is)

**Bugs this fixes:** 

(either VSO or GitHub links)

**Workarounds, if any**

**Risk**

**Performance impact**

**Root cause analysis:**

**How was the bug found?**

</details>

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
@@ -39,14 +39,28 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             Requires.NotNull(update, nameof(update));
             Requires.NotNull(context, nameof(context));
 
-            if (handlerType == RuleHandlerType.Evaluation)
+            try
             {
-                HandleEvaluation(update, context, isActiveContext);
+                InvokeMethodIfFound(context, "StartBatch");
+                if (handlerType == RuleHandlerType.Evaluation)
+                {
+                    HandleEvaluation(update, context, isActiveContext);
+                }
+                else
+                {
+                    HandleDesignTime(update, context, isActiveContext);
+                }
             }
-            else
+            finally
             {
-                HandleDesignTime(update, context, isActiveContext);
+                InvokeMethodIfFound(context, "EndBatch");
             }
+        }
+
+        private static void InvokeMethodIfFound(object o, string methodName)
+        {
+            var method = o.GetType().GetMethod(methodName);
+            method?.Invoke(o, null);
         }
 
         private void HandleEvaluation(IProjectVersionedValue<IProjectSubscriptionUpdate> update, IWorkspaceProjectContext context, bool isActiveContext)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
@@ -173,7 +173,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
 
         private static void ProcessDesignTimeBuildFailure(IProjectChangeDescription projectChange, IWorkspaceProjectContext context, IProjectLogger logger)
         {
-            // If 'CompileDesignTime' didn't run due to a preceeding failed target, or a failure in itself, CPS will send us an empty IProjectChangeDescription
+            // If 'CompileDesignTime' didn't run due to a preceding failed target, or a failure in itself, CPS will send us an empty IProjectChangeDescription
             // that represents as if 'CompileDesignTime' ran but returned zero results.
             //
             // We still forward those 'removes' of references, sources, etc onto Roslyn to avoid duplicate/incorrect results when the next
@@ -254,7 +254,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         {
             if (succeeded)
             {
-                logger.WriteLine("Last design-time build suceeeded, turning semantic errors back on.");
+                logger.WriteLine("Last design-time build succeeded, turning semantic errors back on.");
             }
             else
             {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
@@ -85,11 +85,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
                     IProjectChangeDescription projectChange = update.Value.ProjectChanges[ruleName];
                     if (projectChange.Difference.AnyChanges)
                     {
-                        _currentHandlerTask = _currentHandlerTask.ContinueWith(
-                            _ => handler.handler.Handle(version, projectChange, isActiveContext, logger),
-                            CancellationToken.None,
-                            TaskContinuationOptions.None,
-                            TaskScheduler.Default);
+                        EnqueueWork(_ => handler.handler.Handle(version, projectChange, isActiveContext, logger));
                     }
                     else
                     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
```diff
@@ -44,16 +44,31 @@ public void Handle(IProjectVersionedValue<IProjectSubscriptionUpdate> update, Ru
             Requires.NotNull(update, nameof(update));
             Requires.NotNull(context, nameof(context));
 
-            if (handlerType == RuleHandlerType.Evaluation)
+            dynamic dynamicContext = context;
```
jasonmalinowski: Why the dynamic, given you're not using it in the end and using classic reflection?

Pilchie:
Because I forgot to remove it :-/

—
Kevin Pilch
https://twitter.com/Pilchie

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
```diff
@@ -73,7 +90,11 @@ private void HandleEvaluation(IProjectVersionedValue<IProjectSubscriptionUpdate>
                     IProjectChangeDescription projectChange = update.Value.ProjectChanges[ruleName];
                     if (projectChange.Difference.AnyChanges)
                     {
-                        handler.handler.Handle(version, projectChange, isActiveContext, logger);
+                        _currentHandlerTask = _currentHandlerTask.ContinueWith(
```
davkean: Is there a race here? What if both threads are in ContinueWith at same time?

davkean:
There's explicit switches to UI thread elsewhere in language service; both in creation of the IWorkspaceProjectContext and before LanguageServiceHandlerManager.Handle is called I think.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/SemaphoreSlimExtensions.cs
```diff
@@ -57,5 +57,19 @@ await joinableTask.Task
                 }
             }
         }
+
+        public static async Task ExecuteWithinLockAsync(this SemaphoreSlim semaphore, JoinableTaskCollection collection, JoinableTaskFactory factory, Action action)
+        {
+            // Join the caller to our collection, so that if the lock is already held by another task that needs UI 
+            // thread access we don't deadlock if we're also being waited on by the UI thread. For example, when CPS
+            // is draining critical tasks and is waiting us.
```
jmarolf: do you mean `is awaiting us`?

Pilchie: Not sure - copied from the `Task<T>` version.  @davkean should this be "is awaiting us", or "is waiting for us"?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -273,6 +280,18 @@ private Task ExecuteWithinLockAsync(Func<Task> task)
...
 
+        /// <summary>
+        /// Roslyn is working on supporting free-threaded initialization.
+        /// in order to support a smooth transition, we decide whether to serialize
+        /// based on whether a particular member that was added at the same time exists.
+        /// </summary>
+        /// <returns><see langword="true"/> if the workspace supports batching.</returns>
+        internal static bool WorkspaceSupportsBatchingAndFreeThreadedInitialization()
+        {
+            return typeof(IWorkspaceProjectContext).GetMethod("StartBatch") != null &&
```
jasonmalinowski: Is it worth caching this? Not sure how expensive this is for repeated lookups.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
```diff
@@ -175,12 +192,30 @@ private void ProcessItems(IComparable version, IProjectChangeDescription project
...
+            {
+                _currentHandlerTask = _currentHandlerTask.ContinueWith(
+                    action,
+                    CancellationToken.None,
+                    TaskContinuationOptions.None,
+                    TaskScheduler.Default);
+            }
+            else
+            {
+                // If not, we assume that the LanguageServiceHost has serialized us onto the UI
```
jasonmalinowski: Is there an assert you can write there?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
```diff
@@ -175,12 +192,30 @@ private void ProcessItems(IComparable version, IProjectChangeDescription project
...
             }
 
             CommandLineNotifications.FirstOrDefault()?.Value.Invoke(context.BinOutputPath, addedItems, removedItems);
         }
 
+        private void EnqueueWork(Action<Task> action)
+        {
+            if (LanguageServiceHost.WorkspaceSupportsBatchingAndFreeThreadedInitialization())
+            {
+                _currentHandlerTask = _currentHandlerTask.ContinueWith(
```
jasonmalinowski: Why form a queue here vs. just let the calls still go on? Given there's a lack of locking on _currentHandlerTask itself clearly this is already being serialized somewhere higher up.

jasonmalinowski: (unless you're implementing the "let's let this run later" proposal, but not sure if it's better to keep behavior more similar for now?)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
Pilchie: I think the only race is in `HandleDesignTime`, that another project could queue something in between the 3 tasks that happen there.  Otherwise, isn't reference assignment atomic on x86?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
Pilchie: I was using this to serialize, but maybe it's better just to lock while making the calls.  The purpose was to have just the events related to the current batch happen within the batch, but I think a better approach for that might be to hold a lock for the length of the batch.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
Pilchie: Removed in latest.

Pilchie: Cached in latest.  Not sure whether worth it or not either.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
Pilchie: Removed in latest.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
Pilchie: These are actually already serialized in LanguageServiceHost.HandleAsync.

Pilchie: (so I removed the queuing entirely).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
Pilchie: There is no race because these were already protected by the lock in `LanguageServiceHost.HandleAsync`.

Pilchie:
@davkean said:
> There's explicit switches to UI thread elsewhere in language service; both in creation of the IWorkspaceProjectContext and before LanguageServiceHandlerManager.Handle is called I think.

There are before creating and disposing of IWorkspaceProjectContext.  I'm not terribly worried about those for the purposes of this change, though I agree they would be good to address as well.

The one before LanguageServiceHandlerManager.Handle is moved in this PR to be conditional on whether or not the Roslyn support is there.

Pilchie:
@jasonmalinowski what are your thoughts on the switch before creating `IWorkspaceProjectContext`.  Do you want to merge this and then follow up with that, or wait until I get that done?

jasonmalinowski:
@Pilchie Assuming haste doesn't make waste, I wouldn't block on that. It would be good to exercise that those can be called free-threaded too, but having at least the bulk updated would be fantastic.

Pilchie:
@davkean - if you're okay with it, let's get this merged.

Issue #3338, opened 2018-03-02
Close solution and reopen, an internal error occurred.


Xiu-Li:
## **Repro steps:**
1. Create a new .net core/standard project.
2. Close the solution and reopen or unload the project and reload.
**Actual results:**
Prompt for the following information:
![image](https://user-images.githubusercontent.com/24784908/36889508-41df95de-1e35-11e8-8ef2-925b759c994f.png)
The diagnostic log:
\\bvtsrv2\team\v-lixiu\DevOpsCTI\E2E Validation\Diagnostic Logs\VsProjectFault_40e71a4f-17db-42fd-82ca-75a31ff5ee0d.failure.txt
**Expected results:**
No any errors and warnings
## **Related information:**
VS build: D15.7stg/27501.3002 (15.7 Preview 1) 
dotnet --info:
```
.NET Command Line Tools (2.1.100)

Product Information:
 Version:            2.1.100
 Commit SHA-1 hash:  b9e74c6520

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.100\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.5
  Build    : 17373eb129b3b05aa18ece963f8795d65ef8ea54
```

davkean:
Looks like a regression that I introduced: 
```
Microsoft.Assumes+InternalErrorException: An internal error occurred. Please contact Microsoft Product Support Services.
   at Microsoft.Assumes.Fail(String message, Boolean showAssert)
   at Microsoft.VisualStudio.ProjectSystem.VS.Debug.StartupProjectRegistrar.<InitializeCoreAsync>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.ProjectSystem.OnceInitializedOnceDisposedAsync.<<-ctor>b__4_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #3364, opened 2018-03-08
Add a safe way to grab the project GUID

davkean:
The project GUID can be set by the solution right up to the point it fires IVsSolutionEvents.OnAfterOpenProject (IProjectAsyncLoadDashboard.ProjectLoadedInHost completes) in reaction to a clash with another project. Add a singleton API that is intended as a wrapper around IProjectGuidService/IProjectGuidService2 that waits for the project to be loaded before retrieving the GUID..

Fixes: #3338

Updated services to consume that.

Still testing this, but code is complete.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/VsSafeProjectGuidServiceTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/VsSafeProjectGuidServiceTests.cs
@@ -8,6 +8,7 @@ using Xunit;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS
 {
+    [Trait("UnitTest", "ProjectSystem")]
     public class VsSafeProjectGuidServiceTests
     {
         [Fact]
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/VsSafeProjectGuidServiceTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/VsSafeProjectGuidServiceTests.cs
@@ -15,7 +15,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         public void GetProjectGuidAsync_WhenProjectAlreadyUnloaded_ReturnsCancelledTask()
         {
             var tasksService = IProjectAsynchronousTasksServiceFactory.ImplementUnloadCancellationToken(new CancellationToken(canceled: true));
-            var loadDashboard = IProjectAsyncLoadDashboardFactory.ImplementProjectLoadedInHost(() => Task.Delay(-1));
+            var loadDashboard = IProjectAsyncLoadDashboardFactory.ImplementProjectLoadedInHost(() => Task.Delay(Timeout.Infinite));
 
             var accessor = CreateInstance(tasksService, loadDashboard);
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSafeProjectGuidService.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    /// <summary>
+    ///     An implementation of <see cref="IProjectGuidSafeAccessor"/> that waits until the project
```
davkean: Ugh, old name need to fix that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/VsSafeProjectGuidServiceTests.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    [Trait("UnitTest", "ProjectSystem")]
+    public class VsSafeProjectGuidServiceTests
+    {
+        [Fact]
+        public void GetProjectGuidAsync_WhenProjectAlreadyUnloaded_ReturnsCancelledTask()
+        {
+            var tasksService = IProjectAsynchronousTasksServiceFactory.ImplementUnloadCancellationToken(new CancellationToken(canceled: true));
+            var loadDashboard = IProjectAsyncLoadDashboardFactory.ImplementProjectLoadedInHost(() => Task.Delay(-1));
```
Pilchie: What does `-1` mean here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -69,12 +73,8 @@ public Task InitializeAsync()
 
         protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
         {
-            IProjectGuidService2 projectGuidService = ProjectGuidServices.FirstOrDefault()?.Value as IProjectGuidService2;
```
Pilchie: Should `ProjectGuidServices` be removed?

davkean: Yes it should.

davkean:
> Is there any concern about these async calls being delayed until the project is loaded? 
No. Language service already waited for ProjectLoadedInHost. 

>Will this significantly affect PPL or introduce more UI delays immediately after the solution is loaded?
No, Language service behavior is exactly the same. The StartupProjectRegistrar behavior has slightly changed in that it runs after open - but I removed it switch to the UI thread a while back.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/VsSafeProjectGuidServiceTests.cs
davkean: Infinite.

davkean: Will fix.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
davkean: Fixed.

davkean:
This looks good.

Issue #3273, opened 2018-02-16
Remove check for imported targets from up to date check


panopticoncentral:
The fast up to date check checks to see if imported targets have been modified when determining up to date-ness. Unfortunately, MSBuild does not do that, so you can end up in the situation where an updated target file causes a project to be always up to date. We should remove this check until such time as MSBuild starts to check it too.

davkean:
Ìt does via the `<MSBuildAllProjects>` property which becomes an input for certain targets.

panopticoncentral:
Right, I meant to say that -- we either need to remove the check (which is what the csproj project system does) or look at MSBuildAllProjects instead.

PR #3375, opened 2018-03-13
Fix up to date check of imports

panopticoncentral:
**Customer scenario**

The up to date check for .NET Core projects is currently too aggressive -- we currently check the project against all imported target files. MSBuild, however, only checks the projects listed in $(MSBuildAllProjects). So you can get into a situation where a targets file has been touched and the project will never be considered up to date because the build won't regenerate the output (since MSBuild thinks it's up to date).

**Bugs this fixes:** 

#3273

**Workarounds, if any**

User would have to do a manual rebuild.

**Risk**

Low to moderate. This should be more precise, but it's possible there is some quirk about $(MSBuildAllProjects) that might cause the up to date behavior to be wrong.

**Performance impact**

Low, we'll be checking fewer targets files than before.

No

**How was the bug found?**

Customer reported

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -102,10 +102,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
         {
             _link = ProjectDataSources.SyncLinkTo(
                 _configuredProject.Services.ProjectSubscription.JointRuleSource.SourceBlock.SyncLinkOptions(new StandardRuleDataflowLinkOptions { RuleNames = ProjectPropertiesSchemas }),
-                _configuredProject.Services.ProjectSubscription.ImportTreeSource.SourceBlock.SyncLinkOptions(),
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -117,6 +116,12 @@ private void OnProjectChanged(IProjectSubscriptionUpdate e)
             _msBuildProjectDirectory = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectDirectoryProperty, _msBuildProjectDirectory);
             _outputRelativeOrFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.OutputPathProperty, _outputRelativeOrFullPath);
 
+            var allProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, string.Empty)
+                .Split(';')
+                .Where(path => !string.IsNullOrWhiteSpace(path));
```
Pilchie: I think `Split` should never give null, and `StringSplitOptions` should let you skip empties without the Linq?

panopticoncentral:

PR #3376, opened 2018-03-13
Changed how getting item elements for ordering work

TIHan:
This changes the way of determining which elements in a xml project file are to be moved. 

Before the change, we would wait for a new published tree, then look at the tree's elements to determine which nodes were added. With those nodes, we would get the corresponding xml elements.

Since we cannot wait for a published tree in write locks in some cases (e.g. pasting items), we have to change the way we get the added xml elements. With this PR, we look at what elements were added using the MSBuild Project class itself instead of a project tree; and we do this for everything so we do not have duplicate logic for determining added elements.


TIHan:
@dotnet-bot test windows_debug please

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
```diff
@@ -67,19 +68,21 @@ protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree
...
-            var updatedNodeToAddTo = _projectTree.CurrentTree.Find(nodeToAddTo.Identity);
-            var addedNodes = OrderingHelper.GetAddedNodes(nodeToAddTo, updatedNodeToAddTo);
+            var addedElements = await OrderingHelper.AddItems(_projectVsServices.ActiveConfiguredProject, _accessor, () => OnAddingNodesAsync(nodeToAddTo)).ConfigureAwait(true);
 
-            if (addedNodes.Any())
+            if (addedElements.Any())
             {
-                // Make sure to change ConfigureAwait to "true" if we need switch back.
-                await OnAddedNodesAsync(_projectVsServices.ActiveConfiguredProject, node, addedNodes, updatedNode).ConfigureAwait(false);
+                await _accessor.OpenProjectForWriteAsync(_projectVsServices.ActiveConfiguredProject, project => OnAddedElements(project, addedElements, node, nodeToAddTo)).ConfigureAwait(true);
```
davkean: Why ConfigureAwait(true)? Please add a comment indicating why, we usually don't want to switch back to the same thread, or back to threadpool if started on it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractMoveCommand.cs
```diff
@@ -43,74 +44,19 @@ public AbstractMoveCommand(IPhysicalProjectTree projectTree, SVsServiceProvider
 
         protected override async Task<bool> TryHandleCommandAsync(IProjectTree node, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut)
         {
-            var didMove = await TryMoveAsync(_configuredProject, node).ConfigureAwait(true);
+            var didMove = false;
+                
+            await _accessor.OpenProjectForWriteAsync(_configuredProject, project => didMove = TryMove(project, node)).ConfigureAwait(true);
```
davkean: Why ConfigureAwait(true)?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
+using Microsoft.VisualStudio.Packaging;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
+{
+    /// <summary>
+    /// These are helper functions to select items, expand and collapse folders in a IVsUIHierarchy.
+    /// </summary>
+    internal static class HACK_NodeHelper
```
davkean: We cannot check in this name. Can you expand what is causing this to be a hack?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /// <summary>
+        /// Callbacks with a hierarchy and hierarchy window for use.
+        /// </summary>
+        private static void UseWindow(ConfiguredProject configuredProject, IServiceProvider serviceProvider, Action<IVsUIHierarchy, IVsUIHierarchyWindow> callback)
+        {
+            if (configuredProject.UnconfiguredProject.Services.HostObject is IVsUIHierarchy hierarchy)
+            {
+                callback(hierarchy, GetUIHierarchyWindow(serviceProvider, Guid.Parse(ManagedProjectSystemPackage.SolutionExplorerGuid)));
```
davkean: Use VsConstants.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// Get reference to IVsUIHierarchyWindow interface from guid persistence slot.
+        /// </summary>
+        /// <param name="serviceProvider">The service provider.</param>
+        /// <param name="persistenceSlot">Unique identifier for a tool window created using IVsUIShell::CreateToolWindow.
+        /// The caller of this method can use predefined identifiers that map to tool windows if those tool windows
+        /// are known to the caller. </param>
+        /// <returns>A reference to an IVsUIHierarchyWindow interface, or <c>null</c> if the window isn't available, such as command line mode.</returns>
+        private static IVsUIHierarchyWindow GetUIHierarchyWindow(IServiceProvider serviceProvider, Guid persistenceSlot)
+        {
+            if (serviceProvider == null)
```
davkean: Requires.NotNull

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+            IVsUIHierarchyWindow uiHierarchyWindow = null;
+
+            try
+            {
+                if (ErrorHandler.Succeeded(shell.FindToolWindow(0, ref persistenceSlot, out var frame)) && frame != null)
+                {
+                    ErrorHandler.ThrowOnFailure(frame.GetProperty((int)__VSFPROPID.VSFPROPID_DocView, out pvar));
+                }
+            }
+            finally
```
davkean: Why is this in a failure block? Why do you ignore FindToolWindow, but throw on frame.GetProperty?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -112,142 +75,104 @@ public static bool TryMoveDown(Project project, IProjectTree projectTree)
...
-        /// <summary>
-        /// Get the difference to see what was added.
-        /// We do a sanity check to make sure they have a valid display order.
-        /// We also order the added nodes by their display order.
-        /// </summary>
-        public static ImmutableArray<IProjectTree> GetAddedNodes(IProjectTree previousTarget, IProjectTree currentTarget)
-        {
-            Requires.NotNull(previousTarget, nameof(previousTarget));
-            Requires.NotNull(currentTarget, nameof(currentTarget));
+            var targetChild = GetChildren(target).FirstOrDefault();
```
davkean: This needs a comment, I don't understand why you look only look two levels?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -112,142 +75,104 @@ public static bool TryMoveDown(Project project, IProjectTree projectTree)
...
+            var includes = GetEvaluatedIncludes(projectTree);
 
-            // This recursively tries to find a node to work with. If there is none at all, this will break on project root.
-            var currentTarget = target;
-            while(nodeToAddAbove == null && currentTarget.IsFolder && !GetChildren(currentTarget).Any())
-            {
-                currentTarget = currentTarget.Parent;
-                nodeToAddAbove = GetFirstChild(currentTarget);
-            }
+            var elements = new List<ProjectItemElement>();
```
davkean: Use a ImmutableArray.Builder.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -112,142 +75,104 @@ public static bool TryMoveDown(Project project, IProjectTree projectTree)
...
+        /// If the project gets any additional items from the callback, this will return those added items.
         /// </summary>
-        public static Task<bool> TryMoveNodesToTopAsync(ConfiguredProject configuredProject, IEnumerable<IProjectTree> nodes, IProjectTree target)
+        public static async Task<ImmutableArray<ProjectItemElement>> AddItems(ConfiguredProject configuredProject, IProjectAccessor accessor, Func<Task> addItems)
         {
-            return ModifyProjectAsync(configuredProject, project => TryMoveNodesToTop(project, nodes, target));
+            var previousIncludes = await accessor.OpenProjectForReadAsync(configuredProject, project =>
+            {
+                return project.AllEvaluatedItems.Select(x => x.EvaluatedInclude).ToImmutableHashSet(StringComparer.OrdinalIgnoreCase);
+            }).ConfigureAwait(true);
```
davkean: Why ConfigureAwait(true)?

The project could change between these two OpenProjectForReadAsync drastically, what happens if it does?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <returns>A reference to an IVsUIHierarchyWindow interface, or <c>null</c> if the window isn't available, such as command line mode.</returns>
+        private static IVsUIHierarchyWindow GetUIHierarchyWindow(IServiceProvider serviceProvider, Guid persistenceSlot)
+        {
+            if (serviceProvider == null)
+            {
+                throw new ArgumentNullException(nameof(serviceProvider));
+            }
+
+
+            if (!(serviceProvider.GetService(typeof(SVsUIShell)) is IVsUIShell shell))
```
davkean: Just do the direct cast and let it fail. If this fails (which it won't), an InvalidCastException is easier to debug and gives more info than an InvalidOperationException.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
davkean: You're trying to handle command-line mode? Under what circumstances would this code be called in command-line mode?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+                window.ExpandItem(hierarchy, itemId, EXPANDFLAGS.EXPF_CollapseFolder);
+            });
+        }
+
+        /// <summary>
+        /// Callbacks with a hierarchy and hierarchy window for use.
+        /// </summary>
+        private static void UseWindow(ConfiguredProject configuredProject, IServiceProvider serviceProvider, Action<IVsUIHierarchy, IVsUIHierarchyWindow> callback)
+        {
+            if (configuredProject.UnconfiguredProject.Services.HostObject is IVsUIHierarchy hierarchy)
```
davkean: Just do the cast, if this fails (which it won't) this will be easier to debug than silent failure.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
```diff
@@ -27,6 +28,7 @@ public AbstractAddItemCommand(IPhysicalProjectTree projectTree, IUnconfiguredPro
             _projectTree = projectTree;
             _projectVsServices = projectVsServices;
             _serviceProvider = serviceProvider;
+            _accessor = projectVsServices.ActiveConfiguredProject.UnconfiguredProject.Services.ExportProvider.GetService<IProjectAccessor>();
```
davkean: Just import IProjectAccessor.

davkean:
Can you review all calls to ConfigureAwait(true)? By default we ConfigureAwait(false) which causes us to stay on the same thread when the continuation is run (ie the code after the await). We typically only ConfigureAwait(true) for very specific reasons; ie we started on a UI thread and need to switch back, or we've taken a project lock and need to get back to the threadpool.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
davkean: I'd like this to be pulled into a separate instance class/interface and for us to either:

1) Make these methods Async and switch to the UI thread at the start of them
2) Make these methods sync and enforce that we're on the UI thread - and every caller to deliberately switch to the UI thread.

davkean: The reason for this - is it makes it very clear when/when we're not on the UI thread. At the moment, a simple refactoring/ConfigureAwait change could result in this being called on a background thread and introduce a COM RPC.

davkean:
I like this - much more focused PR.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractMoveCommand.cs
TIHan: It shouldn't. since later I do a ConfigureAwait(true) for the publish before I call into Select.

davkean: Expand on that?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: I will add a comment describing this.

Basically, it's a hack because CPS should give us ways to do this. They have implementations for this functionality but do not expose it. But I need it in this case, and I have no choice but to implement it.

I can make them async and switch to the ui thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: Ok

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
davkean: Don't call it hack, just give it a real name - and check it in. File a bug against CPS to open this up, and if they fund/get around to it, we'll remove our version.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: Ok, I found it. VSConstants.StandardToolWindows.SolutionExplorer. Didn't know it existed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: Ok

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: Ok

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: I have no idea, this whole function was a copy and paste from CPS to get the hierarchy window.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: Yes, great, that is what I definitely want. :)

TIHan:
@davkean I went through the awaits, made them all false except the ones I want to come back from a UI thread, I commented those.

This should be done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/Build/ProjectRootElementFactory.cs
```diff
@@ -7,8 +7,13 @@ namespace Microsoft.Build.Construction
 {
     internal static class ProjectRootElementFactory
     {
-        public static ProjectRootElement Create(string xml = "<Project/>")
+        public static ProjectRootElement Create(string xml = null)
```
jmarolf: Why this change?

TIHan: When I use the IProjectAccessorFactory, its default value is null and that gets passed to the ProjectRootElementFactory. Because my tests use IProjectAccessFactory and I don't pass an argument, it's null therefore ProjectRootElementFactory needs to gracefully handle it. Technically I could pass "<Project/>", but I chose not to. I can change if need be.

TIHan:
I have made my final changes to this PR with @jviau 's suggestions using the hint receiver. I can safely say this is ready.

TIHan:
@dotnet-bot test windows_release please

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
+{
+    /// <summary>
+    /// These are helper functions to select items, expand and collapse folders in a IVsUIHierarchy.
+    /// This is hack due to CPS not exposing fuctionality to do this. They have it internally though.
+    /// </summary>
+    internal static class HACK_NodeHelper
```
Pilchie: I would the HACK part of of this.  Instead, mention it as you do in the doc comment, and link to the bug where you asked for it to be made accessible (which may not get fixed, but would have discussion about why not).

TIHan: So, this might be able to go away soon. At this point in time, I just need Selecting a node. The reason is due to moving an item that you selected, it wouldn't remain selected.  No guarantees, but this https://devdiv.visualstudio.com/DevDiv/_workitems/edit/585954 may fix issue as CPS isn't doing the right thing when items were re-positioned.  // cc @jviau

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderAddItemHintReceiver.cs
```diff
@@ -0,0 +1,103 @@
...
+                                break;
+                        }
+                    }).ConfigureAwait(false);
+                }
+            }
+        }
+
+        public async Task HintingAsync(IProjectChangeHint hint)
+        {
+            if (_action != OrderAddItemHintReceiverAction.NoOp)
```
Pilchie: Can we tighten the conditions here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderAddItemHintReceiver.cs
```diff
@@ -0,0 +1,103 @@
...
+                                break;
+                            default:
+                                break;
+                        }
+                    }).ConfigureAwait(false);
+                }
+            }
+        }
+
+        public async Task HintingAsync(IProjectChangeHint hint)
```
Pilchie: The fact that this takes a single hint, but `HintedAsync` takes a set makes me thing that this might get called many times if several files are added.  Should we ensure that `_previousIncludes` is only set once?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
Pilchie: Let's make sure to rename before merging and add a comment with the issue link.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -112,142 +74,99 @@ public static bool TryMoveDown(Project project, IProjectTree projectTree)
...
-        /// </summary>
-        public static IProjectTree GetFirstChild(IProjectTree projectTree)
-        {
-            Requires.NotNull(projectTree, nameof(projectTree));
-
-            return GetChildren(projectTree).FirstOrDefault();
-        }
+            // If we didn't find a child and our target is an empty folder and not the project root, let's walk up the tree to find a new target child.
+            // Empty folders do not have a valid display order currently in CPS. If they ever do, we have to make changes to this.
+            if (targetChild == null && target.IsFolder && !target.Flags.Contains(ProjectTreeFlags.ProjectRoot))
```
Pilchie: This seems strange - if we're told to move the item to the top of an empty folder, we instead move it to the top of the parent folder?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -453,27 +335,30 @@ private static bool TryMoveElements(ReadOnlyCollection<ProjectItemElement> eleme
             }
 
             // Sanity check
-            var atLeastOneElementMoved = false;
+            var didAllElementsMove = true;
```
Pilchie: Why the change in meaning here?  Can you document the semantics of the return value in the doc comment?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderAddItemHintReceiver.cs
TIHan: Makes sense. I can add logic to make it only set once when the action is not a NoOp.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderAddItemHintReceiver.cs
TIHan: What do you mean by 'tighten' in this case?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: I wanted it to be for all elements moved. It helps ensure the tests absolutely are correct. Yes, I can add a doc comment for it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderAddItemHintReceiver.cs
Pilchie: I meant, "make it evaluate to true less often" so that we aren't doing this work when we don't need to.  In the other method there were several conditions, but here there is only the single one for the action.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: That's correct. The reason is I need choose the correct XML element to move the added item(s) above it. Empty folders do have an XML element, but they are not part of the display order. In the future, if we do consider empty folders to have a display order, this logic will be changed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/HACK_NodeHelper.cs
TIHan: I'm going to find out if we can make a slight change to CPS where we don't need this class anymore. I found where CPS does the logic for previously selected nodes.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderAddItemHintReceiver.cs
TIHan: I see. I will put the same conditions as I did in the order method.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: I completely agree it will be surprising for users. Today is has worse behavior, this will make it a little less surprising.

If we are going to be able to order folders, we need to make a change to CPS. Once I get paste handling in, I can look at the folder work.

TIHan: `TryMoveElementsToTop` is used in the hint receiver and tests.

TIHan:
Ok, @Pilchie , I've made the changes and added a link to the bug for `NodeHelper`. I'm hopeful `NodeHelper` will be removed possibly in 15.8.

TIHan:
Ok. Will do in the future.

TIHan:
@dotnet-bot test windows_debug please

davkean:
@dotnet-bot test windows_release please

PR #3382, opened 2018-03-14
Integrate OpenCover and codecov.io for code coverage

sharwell:


src/Directory.Build.targets
```diff
@@ -26,4 +26,99 @@
...
+      <_TestArchitecture>%(_TestArchitectureItems.Identity)</_TestArchitecture>
+      <_TestEnvironment>$(TargetFramework)|$(_TestArchitecture)</_TestEnvironment>
+      <_TestOutPathNoExt>$(ArtifactsTestResultsDir)$(MSBuildProjectName)_$(TargetFramework)_$(_TestArchitecture)</_TestOutPathNoExt>
+      <_TestStdOutPath>$(_TestOutPathNoExt).log</_TestStdOutPath>
+      <_TestResultsXmlPath>$(_TestOutPathNoExt).xml</_TestResultsXmlPath>
+      <_TestResultsHtmlPath>$(_TestOutPathNoExt).html</_TestResultsHtmlPath>
+      <_TestResultsOpenCoverPath>$(_TestOutPathNoExt).coverage</_TestResultsOpenCoverPath>
+      <_RunOnCore>false</_RunOnCore>
+      <_RunOnCore Condition="$(TargetFramework.StartsWith('netcoreapp'))">true</_RunOnCore>
+      <_UseOpenCover>$(UseOpenCover)</_UseOpenCover>
```
davkean: Who/where is this set?

sharwell: It's passed in using `/p:UseOpenCover` in **build.ps1**.

build/build.ps1
```diff
@@ -111,8 +111,15 @@ function Build {
   }
 
   $nodeReuse = !$ci
+  $useCodecov = $ci -and $env:CODECOV_TOKEN -and ($configuration -eq 'Debug')
```
davkean: This should be passed to Build.ps1 as a switch, build.ps1 takes arguments that dictate this sort of the "like --cover".

src/Directory.Build.targets
```diff
@@ -26,4 +26,99 @@
...
+      <_OpenCoverConsoleExe>OpenCover.Console.exe</_OpenCoverConsoleExe>
+      <_OpenCoverCommand>$(NuGetPackageRoot)opencover\$(OpenCoverVersion)\tools\$(_OpenCoverConsoleExe)</_OpenCoverCommand>
+      <_OpenCoverCommandArgs>-register:user -returntargetcode -hideskipped:All -filter:"+[*]*" -excludebyattribute:*.ExcludeFromCodeCoverage* -excludebyfile:*\*Designer.cs -output:"$(_TestResultsOpenCoverPath)"</_OpenCoverCommandArgs>
+      <_TestRunnerCommand>"$(_OpenCoverCommand)" $(_OpenCoverCommandArgs) -target:"$(_XUnitRunnerCommand)" -targetargs:"$(_XUnitRunnerCommandArgs.Replace(`"`, `\"`))"</_TestRunnerCommand>
+    </PropertyGroup>
+
+    <MakeDir Directories="$(ArtifactsTestResultsDir)"/>
+    <Delete Files="$(_TestOutPathNoExt).*" />
+
+    <Message Text="Running tests: $(TargetPath) [$(_TestEnvironment)]" Importance="high"/>
```
davkean: Why do you redefine the test call instead say just running before and after the Test target?

src/Directory.Build.targets
davkean: Nevermind, figured it out.

src/Directory.Build.targets
sharwell: It wraps (replaces) the xunit executable used for running tests. When RepoToolset is updated to support code coverage, this entire target can be removed.

build/build.ps1
sharwell: Coverage data for local test runs is handled differently from coverage data used by the CI builds.

1. Rather than report the results to codecov.io, the results need to run through ReportGenerator to produce a local HTML view.
2. The OpenCover step extracts PDBs from binaries, leaving a bunch of .pdb and .ppdb files on the file system which are not cleaned in a normal manner. This is fine for the CI since the entire directory is cleared before each build, but would be a nuisance for persistent directories such as the ones on developer machines.

I was hoping to resolve these items separately once we have a chance to review the overall coverage functionality and have a discussion about the additional features we want to see available.

src/Directory.Build.targets
```diff
@@ -26,4 +26,119 @@
...
+      <_TargetFileNameNoExt>$([System.IO.Path]::GetFileNameWithoutExtension('$(TargetFileName)'))</_TargetFileNameNoExt>
+      <_TargetExtractedPdb>$(TargetDir)$(_TargetFileNameNoExt).ppdb</_TargetExtractedPdb>
+      <_SkipPdbConversion>false</_SkipPdbConversion>
+      <_SkipPdbConversion Condition="'$(_UseOpenCover)' != 'true'">true</_SkipPdbConversion>
+      <_SkipPdbConversion Condition="Exists($(_TargetExtractedPdb))">true</_SkipPdbConversion>
+    </PropertyGroup>
+
+    <!-- Extract PDBs from binaries -->
+    <ItemGroup Condition="'$(_SkipPdbConversion)' != 'true'">
+      <_BinariesInTestDirectory Include="$(TargetDir)*.dll" />
```
sharwell: :memo: As a follow-up issue, this step needs to be changed to ensure each binary is only processed once in a parallel build where all projects are targeting the same output directory.

PR #3389, opened 2018-03-15
Delete IPageMetadata

davkean:
This was obsoleted in 15.7 and is not used anywhere. It was accidently made public when the Debug property page was ported.


Pilchie:

PR #3395, opened 2018-03-16
Changed NeutralLanguageCombobox.AutoCompleteMode from Append to SuggestAppend

### LINKED ISSUE #3128: Assembly neutral language not sorted
When trying to select the neutral language for an assembly it was found that the languages are not sorted. This makes it very hard to find the appropriate language.

Additionally it would be very nice to be able to type in this field and the list of best matches be presented and options to pick. sort of like predictive text.

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/180611/assembly-neutral-language-not-sorted.html
VSTS ticketId: 550178_
(no comments)
(no solutions)

sputier:
Fixes #3128 
The combobox can now display a list of suggestions for the language based on the user input


Pilchie:
Thanks @sputier !

PR #3402, opened 2018-03-19
Remove code names from sln

davkean:


davkean:
Non-product change, merging in.

PR #3404, opened 2018-03-19
Remove tests/types to correct namespace

davkean:
These tests live in VS assembly and hence namespace.


davkean:
Namespace only change, merged in.

PR #3413, opened 2018-03-21
Add URL property to GeneralBrowseObject

jviau:
**Customer scenario**

Add the URL property to GeneralBrowseObject. This is to support in-memory editor. XML language service queries this property to find the canonical name of the project.

**Bugs this fixes:** 

Not a bug fix.

**Workarounds, if any**

None

**Risk**

Low

**Performance impact**

None. Adds support for another property via already used method.

No


Pilchie:
@sharwell - this code coverage report doesn't look right to me...

Pilchie:
Merged this - @jviau do you need this for 15.7P3?

jviau:
@Pilchie yes I do

PR #3415, opened 2018-03-22
First stab at feature comparison between project systems

davkean:


docs/feature-comparison.md
```diff
@@ -0,0 +1,59 @@
+# Feature Comparison
+
+The following is an incomplete list of features differences between the legacy project system and the new project system. 
+
+**Feature**|**Legacy**|**New**|**Notes**
+---|---|---|---
+**Platforms**                                                               |
+.NET Standard                                                               | No       | Yes
```
sharwell: .NET Standard is supported by csproj (at least through 1.6) ([example](https://github.com/sharwell/antlr4cs/blob/942fb4b7dfd53b78db523234c5526bf432c95cee/runtime/CSharp/Antlr4.Runtime/Antlr4.Runtime.netstandard.csproj))

docs/feature-comparison.md
```diff
@@ -0,0 +1,59 @@
...
+Universal Windows Platform (UWP)                                            | Yes      | No
+Windows Presentation Framework (WPF)                                        | Yes      | No
+Windows Forms                                                               | Yes      | No
+**Build**|
+Target multiple frameworks (multi-target) from single project               | No       | Yes
+Show build (design-time) errors & warnings in Error List as you make them   | No       | Yes
+**Debug**|
+Debug multiple frameworks from single project                               | No       | Yes
+Debug with multiple environments from single project ("launch profiles")    | No       | Yes
+Debug settings persistence                                                  |project.csproj.user (per user, per machine)|launchsettings.json (source control)
```
sharwell: Settings persistence in project file (source control) is supported by csproj ([example](https://github.com/tunnelvisionlabs/InheritanceMargin/blob/1e6a7ef714cf9295c77cb1ab76b962d98d777160/Tvl.VisualStudio.InheritanceMargin/Tvl.VisualStudio.InheritanceMargin.csproj#L26-L31))

docs/feature-comparison.md
```diff
@@ -0,0 +1,59 @@
...
+Modify environment variables on debug                                       | No       | Yes
+Launch with native debugging                                                | Yes      | Partial | Need to put `"nativeDebugging": true` in launchsettings.json for new project system
+Launch with SQL Server debugging                                            | Yes      | No
+Launch with remote debugging                                                | Yes      | No
+Launch with Azure Snapshot Debugger                                         | No       | Yes
+**Publish**                                                                 |
+Publish to Azure                                                            | No       | Yes
+ClickOnce Publish                                                           | Yes      | No
+**Project**                                                                 |
+Globbing support                                                            | No       | Yes    | `<Compile Include="*.cs" />`
```
sharwell: 💭 I would put globbing support in csproj as **Partial**.

davkean: It's not partial at all, its very limited. It doesn't refresh, doesn't play nicely with up-to-date and is expanded on any manipulation.

docs/feature-comparison.md
davkean: These are not standard usages, and the feature has been turned off in newer versions.

docs/feature-comparison.md
davkean: Also not standard usages - want supported features listed here, not things that people can hack together.

PR #3419, opened 2018-03-23
Adding telemetry to track when the language service is finished initializing

jmarolf:
simple change in order to track language service initialization

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/DesignTimeBuildTelemetryService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/DesignTimeBuildTelemetryService.cs
@@ -1,0 +1,66 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Threading;
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    [Export(typeof(IDesignTimeBuildTelemetryService))]
+    internal class DesignTimeBuildTelemetryService : IDesignTimeBuildTelemetryService
+    {
+        private static volatile int s_numberOfProjectsInDesignTimeBuildQueue = 0;
+        private static volatile int s_numberOfProjectsInLanguageServiceQueue = 0;
+        private static readonly object s_seenProjectLock = new object();
+        private static List<string> s_seenProjects = new List<string>();
+
+        private readonly ITelemetryService _telemetryService;
+
+        [ImportingConstructor]
+        public DesignTimeBuildTelemetryService(ITelemetryService telemetryService)
+        {
+            _telemetryService = telemetryService;
+        }
+
+        /// <summary>
+        /// When a new project is created by the project system this method is called.
+        /// This indicates that a design-time build has been queued.
+        /// </summary>
+        public void OnDesignTimeBuildQueued()
+        {
+            if (Interlocked.Increment(ref s_numberOfProjectsInDesignTimeBuildQueue) == 1)
+            {
+                _telemetryService.PostEvent("Load/DesignTimeBuild/Start");
+            }
+        }
+
+        /// <summary>
+        /// A design-time build has completed.
+        /// </summary>
+        public void OnDesignTimeBuildCompleted(string fullPathToProject)
+        {
+            lock (s_seenProjectLock)
+            {
+                if (!s_seenProjects.Contains(fullPathToProject))
+                {
+                    s_numberOfProjectsInLanguageServiceQueue++;
+                    s_seenProjects.Add(fullPathToProject);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Project system has finished notifying the language service of the project items.
+        /// </summary>
+        public void OnLanguageServicePopulated()
+        {
+            var numberOfProjectsInLanguageServiceQueue = Interlocked.Decrement(ref s_numberOfProjectsInLanguageServiceQueue);
+
+            // Every design time build that has been queued has been handed off to the
+            // Language service.
+            if (s_seenProjects.Count == s_numberOfProjectsInDesignTimeBuildQueue &&
+                numberOfProjectsInLanguageServiceQueue == 0)
+            {
+                _telemetryService.PostEvent("Load/DesignTimeBuild/End");
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -94,6 +99,45 @@ private Task OnProjectFactoryCompletedAsync()
...
+        private static volatile int s_numberOfTimesInitialized = 0;
+        private static object s_telemetryLock = new object();
+
+        /// <summary>
+        /// Only posts an event the first time this method is called in the VS process
+        /// Also updates a global count so we know how many project hosts have been initialized
+        /// </summary>
+        private void LogLanguageServiceHostInitializeStart()
+        {
+            if (s_isFirstStartEvent)
```
Pilchie: Couldn't you avoid the use of the lock and bool by using the return of the Interlocked.Increment being 1?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -94,6 +99,45 @@ private Task OnProjectFactoryCompletedAsync()
...
+        }
+
+        /// <summary>
+        /// Decrements the global count and reports that all hosts have been initialized
+        /// if the count is zero.
+        /// </summary>
+        private void LogLanguageServiceHostInitializeStop()
+        {
+            Interlocked.Decrement(ref s_numberOfTimesInitialized);
+            if (s_numberOfTimesInitialized == 0)
```
Pilchie: You should be checking the result of `Interlocked.Decrement`, not the field.  Otherwise, it might have already changed again by the time you compare it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -94,6 +99,45 @@ private Task OnProjectFactoryCompletedAsync()
...
+            {
+                _telemetryService.PostEvent("LanguageServiceHost/Stop");
```
Pilchie: I'm not *sure* that this sequence is guaranteed to work.  I could imagine it being possible that we could start these and stop them in an interleaved fashion allowing things to get to zero multiple times?

@davkean @lifengl - can you comment on if there are any guarantees here?

davkean: No guarantees that these don't interleave.

davkean:
What _exactly_ is this trying to track?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
davkean: I'd prefer; 1 not including the type name in the event. This entire class has been removed in my current change there is no LanguageServiceHost.

davkean:
Internally it sounds like you want to track:
> language service is fully initialized for intellisense

This is not tracking that. This looks like it is attempting to track once we've created all the abstract projects, but not populated them.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
jmarolf: @Pilchie for solution open I am not seeing any interleaving, though I agree it is possible.  If we get multiple stop events in traces that would invalidate the data.

jmarolf:
@davkean yeah, I'll need to add more telemetry to track when the projects have been populated.  Do you have any suggestion about a good place to start?

jmarolf:
@davkean @panopticoncentral In testing this new approach worked for determining when all design time builds are done, but I would love some feedback here.

@Pilchie this approach waits for *all* design-time builds to be done.  The IDE may be in a usable state before this for some set of features since solution crawler optimizes for the project and files you have open.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHandlerManager.cs
```diff
@@ -91,6 +99,7 @@ private void HandleDesignTime(IProjectVersionedValue<IProjectSubscriptionUpdate>
         {
             Assumes.False(update.Value.ProjectChanges.Count == 0, "CPS should never send us an empty design-time build data.");
```
davkean: We kick off design-time builds for all "buildable" configs in a multi-targeted project.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/DesignTimeBuildTelemetryService.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Threading;
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    [Export(typeof(IDesignTimeBuildTelemetryService))]
+    internal class DesignTimeBuildTelemetryService : IDesignTimeBuildTelemetryService
+    {
+        private static volatile int s_numberOfProjectsInDesignTimeBuildQueue = 0;
```
davkean: Why are these static fields?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -81,6 +85,7 @@ internal partial class LanguageServiceHost : OnceInitializedOnceDisposedAsync, I
         [AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharpLanguageService)]
         private Task OnProjectFactoryCompletedAsync()
         {
+            _designBuildTelemetryService.OnDesignTimeBuildQueued();
```
davkean: This is only called once per project. The other methods are called multiple times during a project's lifetime.

davkean:
@jmarolf Can we sync up so that I can figure out what you're trying to do so I can advise on an approach for this? I'm missing context.

jmarolf:
@Pilchie I have a plan that I've yet to test for 15.7.  Doing this on top of @davkean's work in `master` would be significantly easier.

Pilchie:
The one thing that is preventing me from just saying to base this on @davkean's work in master is that I would love to have a perf test in place to measure the gains of that work...

jmarolf:
yeah, let me test my approach and we can make the call on monday

Issue #2733, opened 2017-08-25
Editor doesn't handle conditional compilation symbol with netstandard or .net core


natidea:
From [VSFeedback](https://developercommunity.visualstudio.com/content/problem/97422/editor-doesnt-handle-conditional-compilation-symbo.html) | [484140](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/484140)

> Create either a Netstandard or Net Core project and put this code in the class:
> 
> #if DEBUG
>         public static int I = 5;
> #else
>         public static int I = 6;
> #endif
> 
> Second statement (#else) will be correctly grayed out.
> Change configuration to Release.
>
> **Expected:** first statement is grayed out.
>
> **Actual:** second is still grayed out.
>
> Note: Sometimes it picks the change but then gets stuck again to that. Compiler works properly. 

In general these changes are not reflected in the editor until you reload the project.

EPinci:
Guy, this is a serious nightmare to work on with for everybody who's relying on compilation symbols.
It probably won't warrant a QFE for 15.3 but can you, please, reconsider it and bump it up to 15.4? Thank you.

matiii:
In 15.4 still doesn't work

MihaMarkic:
Milestone 15.6? Seriously? :(

andreujuanc:
This should have gotten bigger priority. Unloading and Reloading a project just to change what compilation symbols are picked up is, as @EPinci  said, a nightmare.

ericbrunner:
Just added my own depug symbols DEV, STAGE and PROD. What is that f..... automatically added DEBUG_* and RELEASE_* and how to turn that of? Thanks in advance.
```

<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <DefineConstants>TRACE;DEBUG;DEV;NETSTANDARD1_4</DefineConstants>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug-PROD|AnyCPU'">
    <DefineConstants>TRACE;PROD;DEBUG_PROD;NETSTANDARD1_4</DefineConstants>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug-STAGE|AnyCPU'">
    <DefineConstants>TRACE;STAGE;DEBUG_STAGE;NETSTANDARD1_4</DefineConstants>
  </PropertyGroup>

[...]

PR #3423, opened 2018-03-26
[WIP] Language Service hookup rewrite

davkean:
This is a rewrite of the language service hookup that moves from an all knowing single object `LanguageServiceHost` that is aware of all configurations and related IWorkspaceProjectContext, to instead individual configured-project-based services `WorkspaceContextHostInstance` that get fired up and torn down when a configuration becomes "implicitly" active. This naturally supports active config switches from debug -> release, and multi-targeting projects.

This also fixes: https://github.com/dotnet/project-system/issues/2733.

Currently this lives alongside the existing language service hookup, but is intended to replace it. 

Lots of TODOs:

- [ ] Feature flag to turn it on but turn off the other one
- [ ] Multi-targeting display names
- [ ] All IVsHierarchy interaction with Roslyn is through the "real" hierarchy and we handle any of the context properties
- [ ] "Active Context"
- [ ] All external iteration; Container Language, Error List, Edit & Continue, Code Model
- [ ] Need to push to UI thread
- [ ] Need to avoid overlapping evalution/design-time builds
- [ ] Prevent unloading project while updating context - but also bail early

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
@@ -9,14 +9,14 @@ namespace Microsoft.VisualStudio.ProjectSystem
 {
     /// <summary>
     ///     An <see langword="abstract"/> base class that loads, or unloads its inner instance when a 
-    ///     <see cref="ConfiguredProject"/> becomes implicitly activated, or deactivated, respectively.
+    ///     <see cref="ConfiguredProject"/> becomes implicitly activated, or deactivated.
     /// </summary>
     internal abstract class AbstractImplicitlyActiveComponent : AbstractProjectDynamicLoadComponent
     {
         private readonly IConfiguredProjectImplicitActivationTracking _activationTracking;
 
         protected AbstractImplicitlyActiveComponent(IConfiguredProjectImplicitActivationTracking activationTracking, 
-                                                 JoinableTaskContextNode joinableTaskContextNode)
+                                                    JoinableTaskContextNode joinableTaskContextNode)
             : base(joinableTaskContextNode)
         {
             Requires.NotNull(activationTracking, nameof(activationTracking));
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    /// <summary>
+    ///     An <see langword="abstract"/> base class that loads, or unloads its inner instance when a 
+    ///     <see cref="ConfiguredProject"/> becomes implicitly activated, or deactivated, respectively.
+    /// </summary>
+    internal abstract class AbstractImplicitlyActiveComponent : AbstractProjectDynamicLoadComponent
+    {
+        private readonly IConfiguredProjectImplicitActivationTracking _activationTracking;
+
+        protected AbstractImplicitlyActiveComponent(IConfiguredProjectImplicitActivationTracking activationTracking, 
+                                                 JoinableTaskContextNode joinableTaskContextNode)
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+            _activationTracking = activationTracking;
+        }
+
+        protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
+        {
+            _activationTracking.ImplicitlyActivated += OnImplicitlyActivated;
+            _activationTracking.ImplicitlyDeactivated += OnImplicitlyDeactivated;
+
+            if (_activationTracking.IsImplicitlyActive)
+                return LoadAsync();
```
Pilchie: Is `LoadAsync` meant to be idempotent?  Isn't there a race here if the event fires between subscription and here?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerItemHandler.cs
```diff
@@ -13,26 +13,37 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
...
+        private IWorkspaceProjectContext _context;
         private readonly HashSet<string> _paths = new HashSet<string>(StringComparers.Paths);
 
+        [ImportingConstructor]
+        public AnalyzerItemHandler(UnconfiguredProject project)
+            : this(project, null)
+        {
+        }
+
         [ImportingConstructor]
```
Pilchie: Should be removed?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectPropertiesItemHandler.cs
```diff
@@ -21,6 +21,16 @@ public ProjectPropertiesItemHandler(IWorkspaceProjectContext context)
...
         }
 
+        public string EvaluationRule
+        {
+            get { return ConfigurationGeneral.SchemaName; }
+        }
+
+        public void Initialize(IWorkspaceProjectContext context)
+        {
+            _context = context;
```
Pilchie: It's expected that we can overwrite the once given in the constructor now?

davkean: Ah, problem is supporting both language services hosts; one grabs them via MEF, the other manually creates them. I'll add an assert.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
davkean: Yes it is.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerItemHandler.cs
davkean: Yes it should.

davkean:
I had to rebase this, and will need to rebase in the future, so I'd hold off reviewing at the moment.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
davkean: There is a race here; if we deactivate after we check this property - we'll Load but never unload until project tear down.

davkean:
Please still hold off reviewing this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractImplicitlyActiveComponent.cs
davkean: Okay, I've figured out a better way to do this - I'm going to drop the events.

davkean: https://github.com/dotnet/project-system/pull/3689

davkean:
This has been closed in lieu of the features/LanguageService branch.

timbecn:
Hey @davkean - what does the closure of this ticket mean for #2733?

Pilchie:
It's still being worked on, just one commit at a time in the `features/languageservice` branch instead of a single PR.

PR #3424, opened 2018-03-26
Bump the version in dev15.7.x branch

tmeschter:
This change should have been about revving the
`$(PreReleaseVersionLabel)` from `beta3` to `beta4`, but it turns out
that we haven't updated version numbers in a while. So instead
`$(PreReleaseVersionLabel)` is already at `beta4` and staying there,
while `$(VersionBase)` is being revved to 15.7.0/2.7.0.


jasonmalinowski:
(Tagging @sharwell as an interesting FYI that codecov is thinking coverage is changing from this pull request...)

Issue #3414, opened 2018-03-22
In recent builds, files are opening in Misc IntelliSense before switching to project-IntelliSense


davkean:
1. Open ProjectSystem.sln
2. Open src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\Configuration\BaseProjectConfigurationDimensionProvider.cs
3. Close the solution
4. Reopen the solution

![documents](https://user-images.githubusercontent.com/1103906/37745399-a63b2186-2dc8-11e8-9c17-dd120d08e99d.gif)

Pilchie:
The fix in #3447 was inserted for 15.7 Preview 6.

PR #3443, opened 2018-04-04
Add ProjectLoadedInHost and PrioritizedProjectLoadedInHost

davkean:
This is the first part of fixing https://github.com/dotnet/project-system/issues/3414 by moving language service initialization to PrioritizedOnAfterOpenProject instead of OnAfterOpenProject.

Add two properties intended as a replacement for IProjectAsyncLoadDashboard usage. This has an added advantage of:

- Being able to hook onto both OnAfterOpenProject *and* PrioritizedOnAfterOpenProject solution events, instead of just OnAfterOpenProject.

- Avoid needing to combine IProjectAsyncLoadDashboard.ProjectLoadedInHost with IProjectAsynchronousTasksService.UnloadCancellationToken to avoid hanging when the project fails to open.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
@@ -51,8 +51,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
                 {
                     await _threadingService.SwitchToUIThread();
 
-                    // Due to synchronous disposable, we should be on UI thread,
-                    // accessing IVsService<T>.Value will enforce that.
                     HResult result = _solution.Value.UnadviseSolutionEvents(_cookie);
                     if (result.Failed)
                         throw result.Exception;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
@@ -62,10 +62,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         public int OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)
         {
             UnconfiguredProjectTasksService tasksService = GetUnconfiguredProjectTasksServiceIfApplicable(pHierarchy);
-            if (tasksService != null)
-            {
-                tasksService.OnProjectLoadedInHost();
-            }
+            tasksService?.OnProjectLoadedInHost();
 
             return HResult.OK;
         }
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UnconfiguredProjectTasksService.cs
```diff
@@ -3,20 +3,42 @@
...
+        [ProjectAutoLoad(completeBy:ProjectLoadCheckpoint.ProjectFactoryCompleted)]
+        [AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharp)]
+        public Task OnProjectFactoryCompleted()
+        {
+            return _loadedInHostListener.StartListeningAsync();
+        }
+
+        public Task ProjectLoadedInHost
+        {
+            get { return _projectLoadedInHost.Task.WithCancellation(_tasksService.UnloadCancellationToken); }
```
davkean: Undecided whether to do this - *or* CancellationToken.Register. Was worried that the latter would get out of hand for large solutions as you can never unregister.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            return HResult.OK;
+        }
+
+        public int PrioritizedOnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)
+        {
+            UnconfiguredProjectTasksService tasksService = GetUnconfiguredProjectTasksServiceIfApplicable(pHierarchy);
+            if (tasksService != null)
+            {
+                tasksService.OnPrioritizedProjectLoadedInHost();
```
Pilchie: Consider: `taskService?.OnPrioritizedProjectLoadedInHost();` (and above)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
```diff
@@ -0,0 +1,214 @@
...
+            await _threadingService.SwitchToUIThread();
+
+            HResult result = _solution.Value.AdviseSolutionEvents(this, out _cookie);
+            if (result.Failed)
+                throw result.Exception;
+        }
+
+        protected override async Task DisposeCoreAsync(bool initialized)
+        {
+            await _threadingService.SwitchToUIThread();
```
jviau: Wrap this in `if (initialized) { ... }`. I think you will currently throw if disposed before initialization has finished.

davkean: Throw how?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
davkean: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
davkean: I don't think we'll throw or can't think of a circumstance that we'll throw. We'll unnecessarily switch to the UI thread, so I've added a check.

jviau: Yeah I didn't notice the cookie is nil check. Thought you would unadvise without ever having advised in the first place.

davkean:
Ended up merging this with #3447.

PR #3445, opened 2018-04-04
[WIP] Disable integration tests

sharwell:


sharwell:
test ci please

sharwell:
I don't have a way to see if this worked (the CI test disables all jobs anyway), but I confirmed that it doesn't break the other settings in the same block.

PR #3456, opened 2018-04-09
Fix the remove implementation of DelegatedProjectPropertiesProviderBase

### LINKED ISSUE #3455: Events on DelegatedProjectPropertiesProviderBase do not properly implement remove
Upon reviewing the code I noticed that the code that implements `Microsoft.VisualStudio.ProjectSystem.Properties.IProjectPropertiesProvider` in `\src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\Properties\DelegatedProjectPropertiesProviderBase.cs` do not properly implement the remove operator.

```cs
public event AsyncEventHandler<ProjectPropertyChangedEventArgs> ProjectPropertyChanged
{
    add { DelegatedProvider.ProjectPropertyChanged += value; }
    remove { DelegatedProvider.ProjectPropertyChanged += value; }
}
```

The `+=` should be `-=` for the remove.

dannyvv:
Fix #3455: Update the remove implementation of DelegatedProjectPropertiesProviderBase to properly remove the handler instead of accidentally adding it

**Customer scenario**

Not sure in which scenario's CPS calls this code, so not sure how user can hit it. I observed this during reviewing of the code

**Bugs this fixes:** 

#3455

**Risk**

Low: CPS's backing class implements remove properly.

**Performance impact**

**Root cause analysis:**

Was introduced like this since initial commit. I don't know when CPS will call remove on the event, CPS itself only seems to be calling the event, not even adding.

**How was the bug found?**

Reviewing of the codebase.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/DelegatedProjectPropertiesProviderBase.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/DelegatedProjectPropertiesProviderBase.cs
@@ -47,19 +47,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         public event AsyncEventHandler<ProjectPropertyChangedEventArgs> ProjectPropertyChanged
         {
             add { DelegatedProvider.ProjectPropertyChanged += value; }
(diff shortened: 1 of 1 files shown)


davkean:
Thanks again for the contribution @dannyvv.

PR #3436, opened 2018-03-30
Added Paste Ordering

TIHan:
**Customer scenario**

When pasting copied, cut, or dropped F# files into a project or folder, those files need to be moved to the top of it's parent; parent being a project or folder.

**Bugs this fixes:** 

(n/a)

**Workarounds, if any**

Modify fsproj

**Risk**

Low

**Performance impact**

None

**Is this a regression from a previous update?**

No

**Root cause analysis:**

(n/a)

**How was the bug found?**

(n/a)

PR #3459, opened 2018-04-10
Enable NuGet package signing

tmat:
**Customer scenario**

Infrastructure only change. NuGet packages are required to be signed.

**Bugs this fixes:** 

**Workarounds, if any**

**Risk**

**Performance impact**

**Root cause analysis:**

**How was the bug found?**

### DIFF
--- a/build/SignToolData.json
+++ b/build/SignToolData.json
@@ -62,7 +62,7 @@
       ]
     },
     {
-      "certificate": null,
+      "certificate": "NuGet",
       "strongName": null,
       "values": [
         "packages/*.nupkg"


Pilchie:
Approved - but keep in mind that if we don't take any more escrow changes for 15.7, then this change won't apply to packages that match the last inserted build.

tmat:
@Pilchie All packages published to NuGet after 5/15 will have to be signed. If we publish unsigned 15.7 packages before that we should be fine.

tmat:
Test failure due to networking issue.

tmat:
test windows_release please

tmat:
test windows_debug  please

jasonmalinowski:
> Approved - but keep in mind that if we don't take any more escrow changes for 15.7, then this change won't apply to packages that match the last inserted build.

@Pilchie and @tmat, to this point: what would you like to do? We're seeing insertions triggered today because of this.

Pilchie:
I expect will have an escrow insertion to do for https://github.com/dotnet/project-system/pull/3436 soon, so it will probably work itself out.

Issue #2248, opened 2017-05-21
Show the version of the SDK in the solution explorer


davidfowl:
![image](https://cloud.githubusercontent.com/assets/95136/26286148/c15e603c-3e12-11e7-9858-e2b9bdb01365.png)

Visual Studio for MAC does this and it's super useful:

![image](https://cloud.githubusercontent.com/assets/95136/26286157/ef087df6-3e12-11e7-8d0f-8540ce3b0e1e.png)

DamianEdwards:
Why are we calling it The SDK though, it isn't. Those versions are different, but I'd  like to see them all.

davkean:
Agreed @davidfowl I found this missing too.

@DamianEdwards What do you mean?

DamianEdwards:
Maybe it's fine now that we're attempting to unify version numbers better, but for a long time the actual SDK version was different to the shared framework version (which is what Microsoft.NETCore.App actually is). The SDK is actually Microsoft.Net.SDK

MattGertz:
Post-15.3, please.  (15.4 would be OK.)  It's a great idea, but this is borderline feature work and you already have great quantities of bugs to push through for 15.3 shutdown.

tmeschter:
@davidfowl @DamianEdwards @davkean So what should we show next to "Microsoft.NETCore.App", since the version of the SDK and the version of Microsoft.NETCore.App could be different?

I'm inclined to show the actual version of Microsoft.NETCore.App, and show the actual SDK version in the SDK node (e.g. "SDK (1.1.1)") or perhaps just show it in the property grid when you select the SDK node.

tmeschter:
PR #2807 will add the version to the captions of items under the SDK node.

Is it useful to show the SDK version on the SDK node itself?

davidfowl:
> Is it useful to show the SDK version on the SDK node itself?

@tmeschter yes. There are actually 2 versions:

- The SDK version
- The runtime version

Both of these today are nearly impossible to figure out what VS chose to use without looking at an msbuild diagnostics log. We need to show both of these versions in the IDE.

tmeschter:
@davidfowl OK. PR #1572 will address the runtime version (well, that PR plus a small change to the SDK).

Now I'm trying to figure out if the SDK version is available in an MSBuild property.

tmeschter:
Sounds like further work on this is blocked by https://github.com/dotnet/cli/issues/7051.

dsplaisted:
[...]

PR #3461, opened 2018-04-11
[WIP] SDK node rename to Framework

etbyrd:
**Customer scenario**

N/A

**Bugs this fixes:** 

Part of https://github.com/dotnet/project-system/issues/2248

**Workarounds, if any**

N/A

**Risk**

Low, just renaming of files and the SDK node

**Performance impact**

None, just renaming

No

**Root cause analysis:**

N/A

**How was the bug found?**

Requests from team members


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -179,7 +179,7 @@ Press Ignore to ignore the external changes. Your changes may be lost if you clo
     <value>COM</value>
   </data>
   <data name="SdkNodeName" xml:space="preserve">
```
Pilchie: Should probably rename this as well.

etbyrd:
retest windows_debug please

Nirmal4G:
Why rename the Sdk node, the previous issue comments show that we need both SDK node and the Framework Node, so Why not keep the SDK node and duplicate the SDK node into Framework node and put the SDK node under a feature flag or something so that we can enable them when the SDK resolvers implement them?!

It's just a waste to throw all of the work away!

jmarolf:
@Nirmal4G this is a point-in-time thing.  We have no intention of releasing `Framework` without `SDK`.  It's just that `Framework` is a rather obvious name change compared to `SDK` which has more design considerations. No need to clutter this PR with `SDK` node design discussions.

davkean:
We're going to approach this a different way; we are adding a framework node.

PR #3464, opened 2018-04-12
Also support GetCommonProperties from InterceptedProjectPropertiesProvider

Pilchie:
CPS has a new optimization that goes through CommonProperties, and without this we
don't get a chance to intercept these values.

<details>
**Customer scenario**

common a scenario is)

**Bugs this fixes:** 

(either VSO or GitHub links)

**Workarounds, if any**

**Risk**

**Performance impact**

**Root cause analysis:**

**How was the bug found?**

</details>


Pilchie:
Tagging @lifengl too

jmarolf:
@Pilchie code coverage will decrease, must not merge! :)

Pilchie:
But I added tests!

Pilchie:

Pilchie:
@davkean  - can you take a look at this?

Pilchie:
Approved to merge via [593209](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593209)

PR #3468, opened 2018-04-16
When nominating NuGet restore, pass MSBuildProjectExtensionsPath inst…

### LINKED ISSUE #3466: Pass MSBuildProjectExtensionsPath instead of BaseIntermediateOutputPath as part of NuGet restore nomination
Context:

- https://github.com/NuGet/NuGet.Client/pull/2131
- https://github.com/Microsoft/msbuild/issues/1603

Basically, NuGet was writing restore output to `BaseIntermediateOutputPath`, but the .props and .targets were being imported from `MSBuildProjectExtensionsPath`, so if you overrode one but not the other (commonly because you set `BaseIntermediateOutputPath` after the Microsoft common .props were evaluated), the .props written by restore wouldn't be imported.

We've now normalized this in MSBuild and NuGet to use `MSBuildProjectExtensionsPath`.  However, in some cases the path is passed to NuGet from the project system.  So those cases need to be updated to pass `MSBuildProjectExtensionsPath`.

Note that `IVsProjectRestoreInfo` has a `BaseIntermediatePath` property.  In order to avoid breaking the interface, we didn't change the name, but the project system should now set that value to the `MSBuildProjectExtensionsPath` from MSBuild.

[NuGetRestore.xaml](https://github.com/dotnet/project-system/blob/3c760034e97c2551a8ff98fe01e46830d34b4213/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml#L45-L47) may also need to be updated, I'm not sure exactly how that works.

@Pilchie I've optimistically set the milestone for this to 15.7.

Pilchie:
…ead of BaseIntermediateOutputPath

Fixes #3466.

<details>
**Customer scenario**

common a scenario is)

**Bugs this fixes:** 

(either VSO or GitHub links)

**Workarounds, if any**

**Risk**

**Performance impact**

**Root cause analysis:**

**How was the bug found?**

</details>

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/ProjectRestoreInfoBuilder.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/ProjectRestoreInfoBuilder.cs
@@ -87,9 +87,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
             return targetFrameworks.Any()
                 ? new ProjectRestoreInfo
                 {
-                    // NOTE: We pass MSBuildProjectExtensionsPath as BaseIntermediatePath, because people often set
-                    // BaseIntermediatePath too late.  See https://github.com/dotnet/project-system/issues/3466 for
-                    // details.
+                    // NOTE: We pass MSBuildProjectExtensionsPath as BaseIntermediatePath instead of using
+                    // BaseIntermediateOutputPath. This is because NuGet switched from using BaseIntermediateOutputPath
+                    // to MSBuildProjectExtensionsPath, since the value of BaseIntermediateOutputPath is often set too
+                    // late (after *.g.props files would need to have been imported from it). Instead of modifying the
+                    // IVsProjectRestoreInfo interface or introducing something like IVsProjectRestoreInfo with an
+                    // MSBuildProjectExtensionsPath property, we opted to leave the interface the same but change the
+                    // meaning of its BaseIntermediatePath proprtey. See
+                    // https://github.com/dotnet/project-system/issues/3466for for details.
                     BaseIntermediatePath = msbuildProjectExtensionsPath,
                     OriginalTargetFrameworks = originalTargetFrameworks,
                     TargetFrameworks = targetFrameworks,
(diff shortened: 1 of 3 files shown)


Pilchie:
Tagging @dsplaisted to take a look too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/ProjectRestoreInfoBuilder.cs
```diff
@@ -87,7 +87,10 @@ internal static class ProjectRestoreInfoBuilder
             return targetFrameworks.Any()
                 ? new ProjectRestoreInfo
                 {
-                    BaseIntermediatePath = baseIntermediatePath,
+                    // NOTE: We pass MSBuildProjectExtensionsPath as BaseIntermediatePath, because people often set
+                    // BaseIntermediatePath too late.  See https://github.com/dotnet/project-system/issues/3466 for
+                    // details.
```
dsplaisted: Note that the MSBuild property is `BaseIntermediateOutputPath`, which is not the same as the property name: `BaseIntermediatePath`.  I would reword the comment similar to this:

NOTE: We pass MSBuildProjectExtensionsPath as BaseIntermediatePath instead of using BaseIntermediateOutputPath.  This is because NuGet switched from using BaseIntermediateOutputPath to MSBuildProjectExtensionsPath, since the value of BaseIntermediateOutputPath is often set too late (after *.g.props files would need to have been imported from it).  Instead of modifying the IVsProjectRestoreInfo interface or introducing something like IVsProjectRestoreInfo with an MSBuildProjectExtensionsPath property, we opted to leave the interface the same but change the meaning of its BaseIntermediatePath proprtey.  See https://github.com/dotnet/project-system/issues/3466 for for details.

Pilchie: Done thanks!

Pilchie:
@dsplaisted - wanted to double check that this will work regardless of the SDK version, right?  We don't need to do any sort of conditioning?

davkean:
Can you expand on that? It's unclear to me if earlier versions of SDKs are broken.

dsplaisted:
Versions 2.1.1xx and prior of the SDK are broken if `MSBuildProjectExtensionsPath` doesn't match `BaseIntermediateOutputPath` (NuGet restore will write the nuget.g.props and .targets files to the latter path, while the MSBuild common targets will try to import them from the prior path).  So there's not any point in trying to make this change to the project system preserve existing behavior for prior SDKs, since the existing behavior is broken anyway.

Pilchie:
Approved to merge via [601526](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/601526)

PR #3469, opened 2018-04-17
Trigger design-time buld when project.assets.json file changes

davkean:
**Customer scenario**

Installing a NuGet package no longer causes the compiler to pick up the new dlls.

**Bugs this fixes:** 

**Workarounds, if any**

No viable workaround.

**Risk**

Low

**Performance impact**

None

Yes, the new design-time build caching only triggers a design-time build if items in the project have changed. It's unaware of the logic that we have that causes a reevaluation if project.assets.json changes - this change makes it aware of that file.

**Root cause analysis:**

We didn't miss it - it was a fallout of the design.

**How was the bug found?**

Ad hoc testing and customers have started running into it.


davkean:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/601563

Pilchie:
Approved to merge via [601563](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/601563)

PR #3470, opened 2018-04-17
LOC CHECKIN | dotnet/project-system dev15.7.x | 20180417 | termchange

v-zbsail:


Pilchie:
@tmeschter - We're not taking this for 15.7, right? Should we just close, and it will get picked up in master?

tmeschter:
@Pilchie No, we're not taking this for 15.7. A future PR will bring these changes into master. Closing.

Issue #3374, opened 2018-03-13
Visual Studio crashes and hangs opening a solution


davkean:
1. Add feed to: https://dotnet.myget.org/feed/dotnet-core/
2. 

``` XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.IO.Pipelines" Version="4.5.0-preview2-26312-02" />
  </ItemGroup>

</Project>
``` 

3. Change netstandard2.0-> netstandard1.3

```

clr!COMNlsHashProvider::HashiStringKnownLower80
clr!COMNlsInfo::InternalGetCaseInsHash
mscorlib_ni!System.Globalization.TextInfo.GetCaseInsensitiveHashCode(System.String, Boolean, Int64)
mscorlib_ni!System.OrdinalComparer.GetHashCode(System.String)
mscorlib_ni!System.Collections.Generic.Dictionary`2[[System.Guid, mscorlib],[System.__Canon, mscorlib]].FindEntry(System.Guid)
mscorlib_ni!System.Collections.Generic.Dictionary`2[[System.Guid, mscorlib],[System.__Canon, mscorlib]].TryGetValue(System.Guid, System.__Canon ByRef)
Microsoft_VisualStudio_ProjectSystem_Managed_VS!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.GetDependencyChildren(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.IDependency)
Microsoft_VisualStudio_ProjectSystem_Managed_VS!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.FindUnresolvedDependenciesRecursive(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.IDependency)
Microsoft_VisualStudio_ProjectSystem_Managed_VS!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.FindUnresolvedDependenciesRecursive(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.IDependency)
[...]

PR #3474, opened 2018-04-18
added check for circular dependencies

etbyrd:
**Customer scenario**

In certain circumstances, there can be a circular dependency that creates a stack overflow when recursively checking for unresolved child dependencies, causing a complete VS crash. 

**Bugs this fixes:** 

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/487858
https://github.com/dotnet/project-system/issues/3374

**Workarounds, if any**

None

**Risk**

Low, there are some threading concerns that I am looking into to make sure that the Hashset is not shared - but I do not think that will be an issue because I am editing a private method and the class is instantiated when creating a snapshot.

**Performance impact**

Low. Just adding a HashSet. 

No

**Root cause analysis:**

There was no check that a dependency was not already resolved. 

**How was the bug found?**

Customer feedback.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
@@ -55,6 +55,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
         private Dictionary<string, bool> _unresolvedDescendantsMap
             = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
 
+        private HashSet<string> _seenDependencies = new HashSet<string>();
+
         private bool? _hasUresolvedDependency;
         public bool HasUnresolvedDependency
         {
@@ -113,7 +115,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
 
         private bool FindUnresolvedDependenciesRecursive(IDependency dependency)
         {
+
             var result = false;
+
+            if (!_seenDependencies.Contains(dependency.Id))
+            {
+                _seenDependencies.Add(dependency.Id);
+            }
+            else
+            {
+                return result;
+            }
+
             if (dependency.DependencyIDs.Count > 0)
             {
                 foreach (var child in GetDependencyChildren(dependency))


jmarolf:
@etbyrd is there a way to write a test for this case?

etbyrd:
retest windows_release please

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -55,6 +55,8 @@ internal class TargetedDependenciesSnapshot : ITargetedDependenciesSnapshot
         private Dictionary<string, bool> _unresolvedDescendantsMap
             = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
 
+        private HashSet<string> _seenDependencies = new HashSet<string>();
```
abpiskunov: > private HashSet<string> _seenDependencies = new HashSet<string>(); [](start = 7, length = 67)

Why keep it global? It seems like implementation detail of FindUnresolvedDependenciesRecursive and don't have to stay in memory? We already keeping too much stuff in memory ...

etbyrd: Good point. I will fix that

etbyrd:
@jmarolf @abpiskunov I will be adding tests for this - I am still worried about some edge cases

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -55,6 +55,8 @@ internal class TargetedDependenciesSnapshot : ITargetedDependenciesSnapshot
         private Dictionary<string, bool> _unresolvedDescendantsMap
             = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
 
+        
+
```
Pilchie: Nit: remove this added blank space.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshotTests.cs
```diff
@@ -573,6 +573,47 @@ public void TFromChanges_RemovedAndAddedChanges()
             Assert.True(snapshot.DependenciesWorld.ContainsKey(@"tfm1\xxx\addeddependency3"));
         }
 
+        /// <summary>
+        /// Added because circular dependencies can cause stack overflows
+        /// https://github.com/dotnet/project-system/pull/3474
```
jmarolf: Link to [https://github.com/dotnet/project-system/issues/3374](https://github.com/dotnet/project-system/issues/3374) instead please

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -32,6 +32,20 @@ internal class TargetedDependenciesSnapshot : ITargetedDependenciesSnapshot
             }
         }
 
+        /// <summary>
+        /// Internal for testing.
+        /// </summary>
+        internal TargetedDependenciesSnapshot(IDictionary<string, IDependency> dependenciesWorld, IEnumerable<IDependency> topLevelDependencies)
```
Pilchie: I would rather make this constructor private, with in `internal static` factory method with a clear name like `CreateForUnitTestingOnly` or something.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -113,14 +127,23 @@ public IEnumerable<IDependency> GetDependencyChildren(IDependency dependency)
...
-                        result = true;
+                        unresolved = true;
+                        break;
+                    }
+
+                    //If the dependency is already in the child map, it is resolved
+                    //Checking here will prevent a stack overflow due to rechecking the same dependencies
+                    if (_dependenciesChildrenMap.ContainsKey(child.Id))
+                    {
+                        unresolved = false;
```
Pilchie: I had to think about why this is sufficient to break cycles.  My conclusion is that we don't have to worry about the unresolved case, because unresolved dependencies wouldn't have any children, is that right?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -113,14 +127,23 @@ public IEnumerable<IDependency> GetDependencyChildren(IDependency dependency)
...
                 foreach (var child in GetDependencyChildren(dependency))
                 {
                     if (!child.Resolved)
                     {
-                        result = true;
+                        unresolved = true;
+                        break;
+                    }
+
+                    //If the dependency is already in the child map, it is resolved
```
Pilchie: Nit: We put a space after `//` in comments.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshotTests.cs
```diff
@@ -573,6 +573,47 @@ public void TFromChanges_RemovedAndAddedChanges()
...
+            var previousSnapshot = new TargetedDependenciesSnapshot(
+                dependenciesWorld: new Dictionary<string, IDependency>()
+                {
+                    { dependencyModelTop1.Id, dependencyModelTop1 },
+                    { dependencyModelTop2.Id, dependencyModelTop2 },
+                },
+                topLevelDependencies: new List<IDependency>() { dependencyModelTop1 });
+
+            // verify it doesn't stack overflow
+            previousSnapshot.CheckForUnresolvedDependencies(dependencyModelTop1);
```
Pilchie: Have you verified that this test fails without your fix?

etbyrd: Yes, it fails without the fix.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshotTests.cs
etbyrd: Done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshotTests.cs
```diff
@@ -573,6 +573,47 @@ public void TFromChanges_RemovedAndAddedChanges()
...
```
jmarolf: I would prefer if we didn't have the failure case here be: crash the test harness.  That is going to be hard to diagnose if we ever regress this.

etbyrd: I agree. I am working a way around this but I wonder if it is a requirement for this fix to be added.

PR #3476, opened 2018-04-18
Fix build error reporting when code coverage is enabled

sharwell:


Pilchie:
Paging @jaredpar for `.ps1` review.

build/build.ps1
```diff
@@ -115,6 +115,9 @@ function Build {
   $useOpenCover = $useCodecov
 
   & $MsbuildExe $ToolsetBuildProj /m /nologo /clp:Summary /nodeReuse:$nodeReuse /warnaserror /v:$verbosity $logCmd /p:Configuration=$configuration /p:SolutionPath=$solution /p:Restore=$restore /p:QuietRestore=true /p:DeployDeps=$deployDeps /p:Build=$build /p:Rebuild=$rebuild /p:Deploy=$deploy /p:Test=$test /p:IntegrationTest=$integrationTest /p:Sign=$sign /p:Pack=$pack /p:UseCodecov=$useCodecov /p:UseOpenCover=$useOpenCover /p:CIBuild=$ci /p:NuGetPackageRoot=$NuGetPackageRoot $properties
```
jaredpar: There is another `$MsbuildExe` invocation below that seems like it should be similarly protected.

build/build.ps1
```diff
@@ -115,6 +115,9 @@ function Build {
...
+  if ((-not $?) -or ($lastExitCode -ne 0)) {
+    throw "Aborting after build failure."
+  }
```
jaredpar: Consider stealing `Exec-Command*` from build-utils.ps1 in the roslyn repo. It has helpers like `Exec-Console` which make all of the error handling automatic.

build/build.ps1
sharwell: It's protected via the call site, which is how the original call was protected before I added code coverage support.

jaredpar: The protection for the `$MsbuildExe` invocation on line 124 seems very fragile. True it's the last line in the `Build` function and we immediately exit after `Build` using `$lastExitCode`. But that seems very fragile to change.

sharwell:
@Pilchie @davkean What should I do here? My original goal was to submit the minimum change that I'm confident corrects the problem at hand and generally preserves the coding style already in use by the rest of the build script. Other refactoring and/or style updates are of course possible, but they introduce other risks at the same time.

jaredpar:
My comments here are just comments. I think your change actually made it better as it validates `$?` and `$lastExitCode`. There are corner cases where Powershell will set one but not the other (this is why I moved to explicit Process and validation of error code). Hence the first, and more crucial, MSBuild check is better. 

Overall though IMHO the latter check is flaky as demonstrated by the need for this PR. I would fix this at some point.

Pilchie:
I'm fine with merging this as is for now.

PR #3479, opened 2018-04-19
Report the correct branch name to codecov.io

sharwell:
Pull request builds use the variable `ghprbTargetBranch`, while primary branch builds use `BranchName`. This pull request updates the build script to support both cases as opposed to just the former.


davkean:
Your build broke - I can't see the issue, code looks good.

davkean:
Strike that I see the issue,you put properties in an item group.

PR #3481, opened 2018-04-19
Mark fields as read-only

davkean:
Used the new code fix.


src/Microsoft.VisualStudio.Editors/XmlIntellisense/XmlIntellisense.vb
```diff
@@ -105,7 +105,7 @@ Namespace Microsoft.VisualStudio.Editors.XmlIntellisense
     '   Methods on this class should only be called on the VS foreground
     '   thread.
     '--------------------------------------------------------------------------
-    <ClassInterface(ClassInterfaceType.None)> _
+    <ClassInterface(ClassInterfaceType.None)>
```
Pilchie: It's interesting that this seems to be pretty-listing the whole file.  @jinujoseph potential bug for you here.

Pilchie:
Oops - this got merged into d15.7.x by mistake :(

@jasonmalinowski @dotnet/roslyn-infrastructure what should we do?

davkean:
Why is that mistake?

Pilchie:
Because 15.7 is in division escrow now, and this isn't the sort of thing that would meet that bar.

Issue #2711, opened 2017-08-15
Need implement IProjectDefaultConfigurationProvider component to prevent loading an empty configuration


lifengl:
.Net Core project currently loads an empty configuration to evaluate the project to understand the right set of project configurations.  It is a huge waste, not only adds one more project evaluation, also that CPS loads the dataflow with it, and it will reevaluate that fake configuration after every change in the project.

We added a new contract to the CPS side, which the .Net Core project can try to infer the first configuration it should load by possibly reading the unevaluated model (or just based on SDK and guess).  If it is implemented correctly, and we can guess it correctly in 90% projects, we can load the first configuration correctly.  Otherwise, it will load an useless configuration just like what we do today.  Basically, CPS will load this configuration instead of the empty one, and we will still use the old logic to know the exact configuration to load.  When it matches the originally loaded one, we will reuse it, and will be faster.

lifengl:
The CPS side change has been made in 15.5.

davkean:
@lifengl Didn't you implement one based on solution config? Is the idea that we do a quick sniff and guess which config it is and use that?

lifengl:
The configuration information from the solution has been passed as a parameter to the component, but we need a component from your team to provide the 3rd dimension of your configuration name (if it exists).  We have a default component to just use the one from solution.

davkean:
If we guess correctly, this will avoid loading an unneeded configuration for multi-targeted projects. We should basically sniff the TargetFrameworks property and return the first TFM combined with the active from the solution config.

lifengl:
Yes

Sent from my phone

On Aug 31, 2017, at 11:30 PM, David Kean <notifications@github.com<mailto:notifications@github.com>> wrote:


If we guess correctly, this will avoid loading an unneeded configuration for multi-targeted projects. We should basically sniff the TargetFrameworks property and return the first TFM combined with the active from the solution config.

lifengl:
Also, if the SDK can be updated to exclude globs or some imports when the TFM doesn't match anything set in the project, it can make the evaluation time of a fake configuration shorter, so in the scenario that the provider makes a wrong guess, it can still be faster.

This issue still impacts Roslyn.sln loading, because it has some multiple targeted projects like TestUtilities.csproj, and ETW trace shows the evaluation time is doubled.

Pilchie:
@lifengl can you take a look at #3104?  @davkean suggests that it should be in CPS instead of here.  Do you agree

Assigning to myself to track for now.

[...]

Issue #1829, opened 2017-03-22
Investigate configuration behavior when the values (Configurations/Platforms/TargetFrameworks) are conditioned or duplicated.


RaulPerez1:
Currently the expectation is that the values are only once in the project and unconditioned (since they are usually what you use as a condition). Now consider the following cases:

Values defined in a conditioned property group

```
<PropertyGroup Condition="'$(OS)' == '1''">
  <Configurations>Debug_OS1</Configurations>
</PropertyGroup>

<PropertyGroup Condition="'$(OS)' == '2''">
  <Configurations>Debug_OS1;Debug_OS2</Configurations>
</PropertyGroup>
```

Values with a specific condition

```
<PropertyGroup>
  <TargetFrameworks>TFM1</TargetFrameworks>
  <TargetFrameworks Condition="'$(OS)' == '1''">$(TargetFrameworks);TFM1</TargetFrameworks>
 </PropertyGroup>
```

Values defined more than once

```
<PropertyGroup>
  <Platforms>x86</Platforms>
  <Platforms>$(Platforms);x64</Platforms>
 </PropertyGroup>
```

When reading the values to populate the dimensions they are populated based on the evaluated values so all these cases are covered. When writing them however it's using the construction model (since values coming from targets can't be written on top of) and picking the first value from the construction model which ignores conditions and duplicated values.

We could change the saving part use the evaluation model if the property is coming from the project and not an import. There are some design issues to take into account with this approach:

1) Things like the configuration manager will only work on the current conditioned value (e.g. with my example above if I delete Debug_OS1 it will only be removed from the current evaluated value). 

2) Similar to the example above renaming something will rename all the conditions in the project that are consuming that configuration. Since the value on the other value will not be renamed when falling into that condition things will not be conditioned properly.

3) Removing something like a platform would remove all of the platform items for the projects regardless of the condition. When going into a separate conditioned platform where the removed platform is still defined there will not be anything conditioned to it.

All of this behavior may or may not be desirable as this is uncharted territory.

PR #3482, opened 2018-04-20
Avoid loading additional configurations for multi-targeting projects

davkean:
To determine the configurations when loading a project, we need to read the project for the dimensions (the `<Configurations>`, `<Platforms>` and `<TargetFrameworks>` properties). This leads to a chicken and egg problem; to figure out all configurations for a given project you need to load at least one configuration to evaluate. CPS made a change to read the same config as the one active in the solution, however, this pattern doesn't work for multi-targeting projects as TargetFramework dimension is never "active", this results us loading an unneeded config that matches the solution just to grab the actual real configurations, see: #2711.

To avoid unnecessarily loading this unneeded config, we "guess" what the defaults will be based on reading the the first value in Configurations, Platforms, TargetFrameworks elements via the MSBuild construction model without evaluating. Configuration and Platform are only read if the solution didn't provide a default. TargetFramework is always read. This avoids unneeded load of a configuration if we guess correctly.

- [x] Need to figure if we should default Platform/Configuration if they aren't present (which is default) to what the SDK provides (Debug/AnyCPU)
- [x] Order matters, we should walk backwards to mimic evaluation
- [x] Add tests for what happens when multiple elements are present
- [x] Need to consider this pattern: https://github.com/MetacoSA/NBitcoin/commit/ca3b52c1b5fc5e25dabaa9e04c57679bbabfcbce?w=1#diff-2b336f1d9878772008f8ab40ba3d1fc7R19.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
@@ -24,7 +24,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
     {
         [ImportingConstructor]
         public PlatformProjectConfigurationDimensionProvider(IProjectAccessor projectAccessor)
-            : base(projectAccessor, ConfigurationGeneral.PlatformProperty, "Platforms")
+            : base(projectAccessor, ConfigurationGeneral.PlatformProperty, "Platforms", "AnyCPU")
         {
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
@@ -19,7 +19,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
     [Export(typeof(IProjectConfigurationDimensionsProvider))]
     [AppliesTo(ProjectCapabilities.ProjectConfigurationsDeclaredDimensions)]
     [Order(DimensionProviderOrder.Platform)]
-    [ConfigurationDimensionDescription(ConfigurationGeneral.ConfigurationProperty)]
+    [ConfigurationDimensionDescription(ConfigurationGeneral.PlatformProperty)]
     internal class PlatformProjectConfigurationDimensionProvider : BaseProjectConfigurationDimensionProvider
     {
         [ImportingConstructor]
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
@@ -19,6 +19,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
     [Export(typeof(IProjectConfigurationDimensionsProvider))]
     [AppliesTo(ProjectCapabilities.ProjectConfigurationsDeclaredDimensions)]
     [Order(DimensionProviderOrder.Platform)]
+    [ConfigurationDimensionDescription(ConfigurationGeneral.ConfigurationProperty)]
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
     [Export(typeof(IProjectConfigurationDimensionsProvider))]
     [AppliesTo(ProjectCapabilities.ProjectConfigurationsDeclaredDimensions)]
     [Order(DimensionProviderOrder.Platform)]
+    [ConfigurationDimensionDescription(ConfigurationGeneral.PlatformProperty)]
```
Pilchie: We don't need a `ConfigurationDimensionDescription` for `TargetFramework`?

Pilchie:
@lifengl - can you take a look?

NicolasDorier:
I changed the pattern to this one https://github.com/MetacoSA/NBitcoin/commit/67736330fd7459894b2246088bd9ec8d1ccc2aa1?w=1 

it turns out that before this commit `NBitcoin.Altcoins` were getting converted by Visual Studio for a reason I still ignore.

davkean:
@NicolasDorier Thanks. You are running into: https://github.com/dotnet/project-system/issues/1829.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/PlatformProjectConfigurationDimensionProvider.cs
davkean: Already have one, because we were opting into "isVariantDimension", but only for that reason. We added it for Platform/Configuration because CPS uses it to avoid calling us when the solution already has a default for those dimensions - solution doesn't have a default for TargetFramework.

davkean:
Hmm for some reason the "read-only" changes got into this PR.

NicolasDorier:
@davkean here is the strange thing: I was using the same pattern for all projects but only NBitcoin.Altcoins were having issues.

Not a big deal as I prefer the pattern now. But I was surprised.
Also, after editing the csproj, for VS to stop asking migration, I needed to restart it.

davkean:
You're seeing a leaky abstraction, details are called out here: https://github.com/dotnet/project-system/blob/master/docs/opening-with-new-project-system.md. The migration is the old project system trying to open the project instead of new project system. If you manually change the GUID in the solution as called out in that doc, we'll correctly open the project.

davkean:
@Pilchie Had to force push this as someone reset 15.7.x underneath me. :)

NicolasDorier:
wow would never had thought looking the SLN for an answer... I was getting crazy.

NicolasDorier:
Ok so something is fishy

```
mkdir test
cd test
dotnet new sln
dotnet new library
dotnet sln add test.csproj
```
the sln then contains the wrong GUID `FAE04EC0-301F-11D3-BF4B-00C04F79EFBC`.

```
dotnet --version
2.1.300-preview1-008174
```

davkean:
It's by design that GUID is burned in - that leads to the "look for TargetFramework/TargetFrameworks elements" behavior. 

If you put a condition on TargetFrameworks/TargetFramework that fools this logic, and hence the bug I pointed to you above. The other GUID as called out in https://github.com/dotnet/project-system/blob/master/docs/opening-with-new-project-system.md forces the project to be loaded by the new project system.

NicolasDorier:
Yes, I would have expected that `dotnet sln add` use the GUID which force the new project system.

davkean:
The "new" GUID is a temporary thing, intentioned as a stop gap until we can open all projects in the new project system.

NicolasDorier:
Similar to a house of card with you guys removing the base while keeping what is above standing. ;)

Switching to new GUID for `dotnet sln add` would allow smoother migration though, as you make sure that new projects don't rely on old system.

davkean:
New changes:

- We now default to Debug|AnyCPU if we cannot figure out the default, typically this occurs when a project is created for the first time, or first time it is added to the solution. Previously, we loaded "NoVariablesDefault" config with no Configuration/Platform set.
- If we cannot parse the default because it made up of another value, for example: `<Platforms>$(Platforms);x64</Platforms>, then we pick the next parsable one. Previously we would have fallen back to "NoVariablesDefault".

davkean:
@Pilchie This is ready for approval, I've tested the hell out of it.

Pilchie:
Approved to merge via 605138.

PR #3485, opened 2018-04-20
Revert "Merge pull request #3464 from Pilchie/Fix593209-InterceptedCo…

Pilchie:
…mmonProperties"

This reverts commit 291f4c0d88bcc05e48f13f03094b62170e0fc953, reversing
changes made to edc5718cee45b8dd31d9f6a207bd5ed844c29bca.

<details>
**Customer scenario**

common a scenario is)

**Bugs this fixes:** 

(either VSO or GitHub links)

**Workarounds, if any**

**Risk**

**Performance impact**

**Root cause analysis:**

**How was the bug found?**

</details>


Pilchie:
Using admin privileges to merge since this is a straight revert.

davkean:
Why the revert?

Pilchie:
This caused a project fault creating new projects .  

The issue here is in [this code](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/TargetFrameworkMonikersValueProvider.cs#L22):

```C#
        public override async Task<string> OnGetEvaluatedPropertyValueAsync(string evaluatedPropertyValue, IProjectProperties defaultProperties)
        {
            var configuredProjects = await _projectProvider.GetActiveConfiguredProjectsAsync().ConfigureAwait(true);
            var builder = ImmutableArray.CreateBuilder<string>();
            foreach (var configuredProject in configuredProjects.Objects)
            {
                var projectProperties = configuredProject.Services.ExportProvider.GetExportedValue<ProjectProperties>();
                var configuration = await projectProperties.GetConfigurationGeneralPropertiesAsync().ConfigureAwait(true);
                var currentTargetFrameworkMoniker = (string)await configuration.TargetFrameworkMoniker.GetValueAsync().ConfigureAwait(true);
                builder.Add(currentTargetFrameworkMoniker);
            }

            return string.Join(";", builder.ToArray());
        }
```

`configuredProjects` is coming back null.  @Lifengl says that we can't rely on other configurations being ready yet.

PR #3497, opened 2018-04-30
Add "Full Path" to project properties

davkean:
- Made project's properties in Properties tool window consistent with project items & legacy project system, favoring the former where they disagreed (such as "Project File").
- Add new "Full Path" property that shows the entire file name of the project

Before:
image

After:

image


davkean:
This has no compatibility concerns as only changed the localized names not the underlying properties.

davkean:
Note, also updated "ConsoleApp297 General" -> "ConsoleApp297 Project Properties" to match legacy.

PR #3504, opened 2018-05-01
A list of integration test scenarios to be automated

Pilchie:
What do people think?  What is missing?  What about the rough priorities?  What is covered by our current (disabled) integration tests?


richaverma1:
@Pilchie, these are applicable to VB also I suppose? Will you also be looking through the scenario list that Matt is creating and updating this list?

Pilchie:
@richaverma1 Yes and Yes.

davkean:
We put repo docs under docs\repo.

davkean:
I'm assuming these are in priority order? The top 4 are the ones I'd focus on and get them passing 100% before moving onto others.

Pilchie:
Moved to `repo`.  Rough priority order, though feedback is welcome.  Agree that the first 4 should be solid before expanding scope.

davkean:
👍

PR #3507, opened 2018-05-02
Fix load of ManagedProjectSystemPackage in F#

Pilchie:
Convert ManagedProjectSystemPackage to use '.NET' for it's load context
capability.


Pilchie:
Tested loading on create of C#/VB/F#, and re-open of a solution with .NET Framework C# and .NET Core VB, and that it *wasn't* loaded on create of a .NET Framework C# project.

PR #3510, opened 2018-05-02
compare DesignTimeBuild and BuildingProject to strong typed bools

SimonCropp:


SimonCropp:
yeah seemed to work for me as well. would be nice to get the approach confirmed though

rainersigwald:
The _literal_ string ` == true OR false != true` is invalid (`error MSB4092: An unexpected token "==" was found at character position 2 in condition " == true OR false != true".`) But putting a property expansion there is legal, regardless of whether it expands to empty at runtime or not.

This change should be ok, but almost all conditions in MSBuild use `'` quoting--why do you think this should be different? It does not affect MSBuild's evaluation of the conditions--MSBuild always attempts to convert strings to bool and compare bools before falling back to string comparisons.

That is, either way MSBuild parses as

image

But the [evaluation](https://github.com/Microsoft/msbuild/blob/8e015cc049b5f3f9cbd9e10ee30ae86998d934d5/src/Build/Evaluation/Conditionals/MultipleComparisonExpressionNode.cs#L50-L85) first checks for empties, then numbers, then bools--which succeeds here.

davkean:
It definitely reads better, I've been conditioned to quote, but if we'd doing boolean comparisons and it works without, why do we bother?

nguerrera:
> in fact i would like to simplify it further:

> `<PropertyGroup Condition="$(DesignTimeBuild) OR !$(BuildingProject)">`

I believe this variant will fail if the variables are not set to anything. @rainersigwald Is that right?

SimonCropp:
@nguerrera sorry yes, you are correct that wont work

davkean:
Thanks for the contribution @SimonCropp!

SimonCropp:
@davkean thank you :)

PR #3515, opened 2018-05-03
Visual Basic projects naming enforcement

davkean:
More editorconfig enforcement. 

Note: I avoided changing any *externally visible* member (public/protected in public class)


davkean:

PR #3521, opened 2018-05-08
Opt Shared Projects into ".NET" capability

davkean:
Binary projects were opt'ing into .NET capability previously via the project type. This opts C#/VB Shared Projects into it.

Unblocks: https://github.com/dotnet/project-system/pull/3299


davkean:
Shared Project's are applicable to C++, but those ones don't import these targets.

PR #3522, opened 2018-05-08
More Visual Basic project cleanup

davkean:
This is more cleanup for the VB projects to match editorconfig settings.

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/IVbpackage.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/IVbpackage.vb
@@ -24,7 +24,9 @@ Namespace Microsoft.VisualStudio.Editors
     Public Class VBPackageUtils
 
         Private Shared s_editorsPackage As IVBPackage
-        Public Delegate Function GetServiceDelegate(ServiceType As Type) As Object
+#Disable Warning IDE1006 ' Naming Styles (Compat)
+        Public Delegate Function getServiceDelegate(ServiceType As Type) As Object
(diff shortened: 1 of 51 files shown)


src/Microsoft.VisualStudio.AppDesigner/IVbpackage.vb
```diff
@@ -24,15 +24,15 @@ Namespace Microsoft.VisualStudio.Editors
     Public Class VBPackageUtils
 
         Private Shared s_editorsPackage As IVBPackage
-        Public Delegate Function getServiceDelegate(ServiceType As Type) As Object
+        Public Delegate Function GetServiceDelegate(ServiceType As Type) As Object
```
Pilchie: Isn't this renaming a public type?

src/Microsoft.VisualStudio.AppDesigner/interop/NativeMethods.vb
```diff
@@ -5,6 +5,8 @@ Imports Microsoft.VisualStudio.OLE.Interop
 Imports ComTypes = System.Runtime.InteropServices.ComTypes
 Imports System.Windows.Forms
 
+#Disable Warning IDE1006 ' Naming Styles
```
Pilchie: Worth dropping a .editorconfig file in this directory instead?

davkean: Meh, probably not going to ever change this code again.

PR #3525, opened 2018-05-08
Fixed inability to add F# files to ASP .NET Core projects

TIHan:
This fixes the issue with F# files not able to be added in the ASP .NET Core projects. The cause was due to the ASP .NET Core project including a file that was part of props file, and our ordering logic did not account for files that are not associated with the main project. We now, simply skip them when doing ordering, but their display order will still be correct.

We are hoping to get this in the next servicing update.

Here is the reported bug:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612703

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.Build.Evaluation;
 using Microsoft.VisualStudio.ProjectSystem.Input;
 using Microsoft.VisualStudio.Shell;
 using Task = System.Threading.Tasks.Task;
@@ -14,25 +15,33 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
         private readonly IUnconfiguredProjectVsServices _projectVsServices;
         private readonly SVsServiceProvider _serviceProvider;
         private readonly OrderAddItemHintReceiver _orderAddItemHintReceiver;
+        private readonly ConfiguredProject _configuredProject;
+        private readonly IProjectAccessor _accessor;
 
         public AbstractAddItemCommand(
             IPhysicalProjectTree projectTree, 
             IUnconfiguredProjectVsServices projectVsServices, 
             SVsServiceProvider serviceProvider,
-            OrderAddItemHintReceiver orderAddItemHintReceiver)
+            OrderAddItemHintReceiver orderAddItemHintReceiver,
+            ConfiguredProject configuredProject, 
+            IProjectAccessor accessor)
         {
             Requires.NotNull(projectTree, nameof(projectTree));
             Requires.NotNull(projectVsServices, nameof(projectVsServices));
             Requires.NotNull(serviceProvider, nameof(serviceProvider));
             Requires.NotNull(orderAddItemHintReceiver, nameof(orderAddItemHintReceiver));
+            Requires.NotNull(configuredProject, nameof(configuredProject));
+            Requires.NotNull(accessor, nameof(_accessor));
 
             _projectTree = projectTree;
             _projectVsServices = projectVsServices;
             _serviceProvider = serviceProvider;
             _orderAddItemHintReceiver = orderAddItemHintReceiver;
+            _configuredProject = configuredProject;
+            _accessor = accessor;
         }
 
-        protected abstract bool CanAdd(IProjectTree target);
+        protected abstract bool CanAdd(Project project, IProjectTree target);
 
         protected abstract Task OnAddingNodesAsync(IProjectTree nodeToAddTo);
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -246,12 +211,22 @@ private static IEnumerable<string> GetEvaluatedIncludes(IProjectTree projectTree
...
                     // Sanity check
-                    if (hashSet.Add(tree2.Item.ItemName))
+                    if (hashSet.Add(include))
                     {
-                        includes.Add(tree2.DisplayOrder, tree2.Item.ItemName);
+                        // GetItemsByEvaluatedInclude is efficient and uses a MultiDictionary underneath.
+                        //     It uses this: new MultiDictionary<string, ProjectItem>(StringComparer.OrdinalIgnoreCase);
+                        var item = project.GetItemsByEvaluatedInclude(include).FirstOrDefault();
+
+                        // We only care about adding one item associated with the evaluated include.
```
Pilchie: Is this true?  What if the include is a glob?

TIHan:
This is ready, just awaiting feedback.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -279,9 +259,20 @@ private static bool IsValidDisplayOrder(int displayOrder)
...
+            return projectTree.Children.Where(x =>
+            {
+                var hasValidDisplayOrder = HasValidDisplayOrder(x);
+                if (!hasValidDisplayOrder)
+                {
+                    return false;
+                }
+
+                // Make sure we have a valid backing xml element that is not part of a .props or .targets files.
+                return GetItemElements(project, x, ImmutableArray<string>.Empty).Any();
```
Pilchie: This pattern (seen a couple of places) coupled with the `ToImmutableArray` in `GetItemElements`, forces you to walk the full tree just to see if it's non empty.  Consider having two methods `GetItemElementsDeferred` and `GetItemElements`, that does the `ToImmutableArray` - then places that use `Any` can use the deferred one and stop at the first `true` value if that's all it's used for.

davkean:
I'm wondering if there's a simpler approach to this - instead of figuring out if a node comes from a targets after the fact, add a project item flags that indicates this during our properties provider that sets the display index. Just check for this flag during query status. 

This prevents us from searching the project (on a background thread, while blocked on the UI) during a query status, which can and will cause UI delays.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: That's what I was thinking before. I can revisit this if we think that is wroth it for perf (I think so)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
TIHan: I'm operating under the assumption in this case that we don't handle globs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AbstractAddItemCommand.cs
```diff
@@ -46,15 +55,19 @@ protected Task ShowAddExistingFilesDialogAsync(IProjectTree target)
             return HACK_AddItemHelper.ShowAddExistingFilesDialogAsync(_projectTree, _projectVsServices, _serviceProvider, target);
         }
 
-        protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree node, bool focused, string commandText, CommandStatus progressiveStatus)
+        protected override async Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree node, bool focused, string commandText, CommandStatus progressiveStatus)
         {
-            if (_projectTree.NodeCanHaveAdditions(GetNodeToAddTo(node)) && CanAdd(node))
+            var canAdd = false;
+
+            await _accessor.OpenProjectForWriteAsync(_configuredProject, project => canAdd = CanAdd(project, node)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
TIHan: Oops, I'll be changing that to read.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
davkean: F# projects will have globs, such as the `<Content/>` one from web projects.

TIHan:
Closing this as we found a much simpler fix without such huge swaths of changes.

PR #3528, opened 2018-05-09
Fixed inability for F# files to be added in ASP .NET Core projects

TIHan:
This fixes the issue with F# files not able to be added in the ASP .NET Core projects. The cause was due to the ASP .NET Core project including a file that was part of props file, and our ordering logic did not account for files that are not associated with the main project. We now, simply skip them when doing ordering, but their display order will still be correct.

Changes in the PR:

- Skipping item elements if they are imported when getting elements 
- Defensive null checks in other places where we call TryGetReferenceElement
- Greatly simplified TryMoveElementsToTop that accounts for skipping imported item elements

Here is the reported bug:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612703

Here is the F# issue:
https://github.com/Microsoft/visualfsharp/issues/4851

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelperTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelperTests.cs
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Immutable;
+using System.IO;
+using Microsoft.Build.Construction;
 using Microsoft.Build.Evaluation;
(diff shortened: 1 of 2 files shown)


TIHan:
@Pilchie We don't have a design for the desired behavior with imported items; so I'm not sure what the tests will cover. 

Currently, it will just skip imported items on manipulation which may result in some odd behavior when moving/adding files in relation with imported items; but it won't throw. We should discuss this with CPS and Phillip about how we can design this.

The PR is only for addressing the inability to add F# files by ignoring the imported items while they retain their proper display order. This gets us moving forward while we come up with a design in a later release.

However, I do think a test to cover the case where it was throwing when adding F# files would be good to have. I'll try to add that.

Pilchie:
Tests do not in general verify correctness.  They protect the current behavior against unintended changes. It's okay if we don't know exactly how we want things to behave, just comment that in the test.

TIHan:
Ok, then I'll try to add some in and comment on it.

TIHan:
Ok, I've added tests for the current behavior.

TIHan:
This was approved for 15.7.2, can we merge it?

Pilchie:
Approved to merge via 612703.

PR #3529, opened 2018-05-09
Use object initialization

davkean:
This leaves ~10 instances of editor config warnings/errors/messages left.


davkean:
My aim is just to stop all the errors/messages showing up in Error List.

PR #3532, opened 2018-05-10
Load package asynchronously when UI context is made active

davkean:
While we were an async package, we were still being loaded synchronously when UI context was made active.

Fixes: 592856.


davkean:
Before this change we were being loaded on UI thread. After we were being loaded on background thread.

Issue #3530, opened 2018-05-09
Break COM aggregation links to lessen severity of leaked IVsHierarchy objects


Pilchie:
See :link: https://developercommunity.visualstudio.com/content/problem/246600/outofmemoryexception-crash-in-visual-studio-1580-p.html

At least one issue discovered is that the E&C manager leaks an IVsHierarchy object, which can hold a large graph.  We should be more defensive and break the link from our aggregated objects on solution close.

davkean:
This has been fixed.

PR #3534, opened 2018-05-10
Reduce leaks when an extension holds onto IVsHierarchy (EditAndContinueProvider)

davkean:
Background: Extensions often hold on to IVsHierarchy waaay after the project has closed. To avoid leaking a bunch of services & data, we null out fields to break the chain to a bunch of services and project itself. See https://github.com/dotnet/project-system/issues/3530.

- Use HResult
- Null out fields in EditAndContinueProvider to reduce leaks https://github.com/dotnet/project-system/issues/3530.
- Avoid aggregating Roslyn's AbstractProject with the hierarchy. We have no control over this object's fields
  - We were aggregating AbstractProject from Roslyn with the IVsHierarchy outer. We don't control this object, and therefore cannot null out fields that it may be holding onto when we are disposed. Instead, do the same pattern as legacy and implement IVsENCRebuildableProjectCfg2/IVsENCRebuildableProjectCfg4 and delegate onto it.


davkean:
Note, I'm going to go and fix all ProjectNodeComService objects in a future change.

davkean:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
```diff
@@ -7,78 +7,217 @@
...
+using ENC_BREAKSTATE_REASON = Microsoft.VisualStudio.LanguageServices.Implementation.EditAndContinue.Interop.ENC_BREAKSTATE_REASON;
 
 namespace Microsoft.VisualStudio.ProjectSystem.Managed.VS.EditAndContinue
 {
-
-    [ExportProjectNodeComService(typeof(IVsENCRebuildableProjectCfg))]
+    [ExportProjectNodeComService(typeof(IVsENCRebuildableProjectCfg), typeof(IVsENCRebuildableProjectCfg2), typeof(IVsENCRebuildableProjectCfg4))]
     [AppliesTo(ProjectCapability.EditAndContinue)]
-    internal class EditAndContinueProvider : IVsENCRebuildableProjectCfg
+    internal class EditAndContinueProvider : IVsENCRebuildableProjectCfg, IVsENCRebuildableProjectCfg2, IVsENCRebuildableProjectCfg4, IDisposable
```
sharwell: :bulb: Mark this `sealed` rather than suppress CA1063.

davkean:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
davkean: CA1063 is designed for public types and should not be firing on internal types.

davkean: Hence the bug.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
```diff
@@ -7,78 +7,217 @@
...
+            return HResult.NotImplemented;
         }
 
         public int SetENCProjectBuildOption([In] ref Guid in_guidOption, [In, MarshalAs(UnmanagedType.LPWStr)] string in_szOptionValue)
         {
-            return VSConstants.E_NOTIMPL;
+            return HResult.NotImplemented;
+        }
+
+        public int StartDebuggingPE()
```
Pilchie: I have a slight preference for using explicit implementation for cases like this, but it's not a big deal either way.

davkean: 👍

davkean:

PR #3540, opened 2018-05-11
Remove dead code

davkean:
These modules have been copied into 4 different projects in VS tree, most of it is dead code.


Pilchie:
Seems odd that this didn't improve code coverage at all.

PR #3541, opened 2018-05-11
Only check compat for projects that support package references

### LINKED ISSUE #3537: NullReferenceException in DotNetCoreProjectCompatibilityDetector.GetProjectCompatibilityAsync
Dump is at `"\\mlangfs1\public\frsilb\dumps\smalis\devenv.dmp"`

@smmalis37 has an internal shproj (in the same directory as the dump) that is causing a NullReferenceException on https://github.com/dotnet/project-system/blob/b3b6ea07cd088926ac7973e1f916ec5be9df55e6/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs#L272. His shproj has no project references, which is causing a non-fatal exception and a popup. The project then "loads", but intellisense is very broken.

Pilchie:
Fixes #3537.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
@@ -255,7 +255,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         private static async Task<CompatibilityLevel> GetProjectCompatibilityAsync(UnconfiguredProject project, VersionCompatibilityData compatData)
         {
-            if (project.Capabilities.AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharp + " & " + ProjectCapability.PackageReferences))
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -255,7 +255,7 @@ public int OnAfterBackgroundSolutionLoadComplete()
 
         private static async Task<CompatibilityLevel> GetProjectCompatibilityAsync(UnconfiguredProject project, VersionCompatibilityData compatData)
         {
-            if (project.Capabilities.AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharp))
+            if (project.Capabilities.AppliesTo(ProjectCapability.CSharpOrVisualBasicOrFSharp + " & " + ProjectCapability.PackageReferences))
```
davkean: this will result in:
```
CSharp | VisualBasic | FSharp & PackageReference
```

Not sure what order of operations is, but can we parenthesis these to make this clear?

```
(CSharp | VisualBasic | FSharp) & PackageReference
```

Pilchie: Updated - thanks for catching.

PR #3542, opened 2018-05-14
Delete our implementation of IVsSingleFileGeneratorFactory

davkean:
In lieu of the CPS version.

This can be merged when https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/122403?_a=files is merged.


davkean:
This has been merged into stg and hence we can check-in our side.

Issue #3535, opened 2018-05-10
We're looking for single file generators under new project system's project type instead of legacy


davkean:
Single File Generators [are being looked up](https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/registering-single-file-generators) under the new project types for C#, VB and F#. We should change this to look under the legacy project types so that providers don't need to register these in multiple places.

There's two places where we look up under the new GUIds:
- CPS\src\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\VsProjectTypeValuesCacheService.cs
- https://github.com/dotnet/project-system/blob/b3b6ea07cd088926ac7973e1f916ec5be9df55e6/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Generators/SingleFileGeneratorFactoryAggregator.cs#L63

This will be the reason no custom 3rd party single file generators work, because we don't find them.

Work to do:

- [x] Move VsSingleFileGeneratorFactory/SingleFileGeneratorFactoryAggregator -> CPS and change it call through CPS's VsProjectTypeValuesCacheService. This a project system feature and tied to the `SingleFileGenerators` capability and not to C#. [PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/122403?_a=files).
- [x] Implement the entire interface, not just the one method we have.
- [x] Delete our version [PR](https://github.com/dotnet/project-system/pull/3542)
- [x] Change VsProjectTypeValuesCacheService to pull ProjectTypeGuid from IItemTypeGuidProvider [PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/122410?_a=overview)
- [x] Remove our registration of single file generators from each package, these are the [RemoteCodeGeneratorRegistration] attributes. We don't need them after we start pulling from legacy. [PR](https://github.com/dotnet/project-system/pull/3543).
- [x] Null out fields in COM service

davkean:
We should be pulling on the "active" IItemTypeGuidProvider.

davkean:
We can delete our registration of them also from the C#/VB package.

hallambaker:
Does not address my issue. I worked out that the GUID had changed and registered under the new ones. The problem is that the generators still don't run.

davkean:
@hallambaker Can you paste your resulting pkgdef containing the registration?

davkean:
[...]

PR #3543, opened 2018-05-14
Remove single file generator registration for our project types

davkean:
We're moving CPS to use our IItemTypeGuidProvider to lookup single file generator registration, which will cause us to use the legacy project's registration instead of our own. After that we no longer specific registration under our project type.

We can merge in when the CPS opting into to use IItemTypeGuidProvider https://github.com/dotnet/project-system/issues/3535 is merged in.


davkean:
I'll merge this in when I have confirmation of a CPS insertion.

Issue #3561, opened 2018-05-21
Implement IExtenderCATIDProvider for F#


davkean:
We need to implement IExtenderCATIDProvider for F# for any properties extenders, which is a little harder than it was for C#/VB.

1) F# uses the runtime generated GUIDs of nodes which need to manually figured out: https://github.com/Microsoft/visualfsharp/blob/20b6f6b62f6911006bc37449ef52c3cbe8090fc0/vsintegration/src/FSharp.ProjectSystem.FSharp/Project.fs#L414. We should then change that project system to hardcode them so that we can guarantee they never change.

2) F# returns different values for VSHPROPID_ExtObjectCATID and VSHPROPID_BrowseObjectCATID based on the item ID passed in via IVsHierarchy.GetProperty. IExtenderCATIDProvider does not currently represent that.

Pilchie:
@davkean can you comment on customer scenarios impacted by this, and the timeline you think it needs to be done?

davkean:
Without this, extenders could add a property to the Properties window for items, references and projects for C# and VB projects but not F# projects. Its really hard to know if there is any existing extenders that do this - figuring the GUIDs would be a good start so we could search for them.

Issue #3560, opened 2018-05-21
AbstractProjectConfigurationProperties needs to implement get_Extender/ExtenderCATID/ExtenderNames


davkean:
These methods/properties are not implemented:

AbstractProjectConfigurationProperties.get_Extender
AbstractProjectConfigurationProperties.ExtenderCATID
AbstractProjectConfigurationProperties.ExtenderNames

They need to call through IExtenderCATIDProvider asking for ProjectConfigurationBrowseObject. Without this, extenders from flavors will not work.

PR #3562, opened 2018-05-21
Return CATID's for project system extenders

davkean:
Implement IExtenderCATIDProvider for C#/VB. F# is being tracked by https://github.com/dotnet/project-system/issues/3561.

Also filed https://github.com/dotnet/project-system/issues/3560 to implement support for ProjectConfigurationBrowseObject and implemented ConfigurationBrowseObject here https://devdiv.visualstudio.com/DevDiv/Default/_git/CPS/pullrequest/123753?_a=overview.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
@@ -1,0 +1,112 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.ProjectSystem.VS.Properties;
+
+using BCLDebug = System.Diagnostics.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation
+{
+    /// <summary>
+    ///     An abstract class that maps <see cref="IExtenderCATIDProvider"/> to a more correct model.
+    /// </summary>
+    internal abstract class AbstractExtenderCATIDProvider : IExtenderCATIDProvider
+    {
+        protected AbstractExtenderCATIDProvider()
+        {
+        }
+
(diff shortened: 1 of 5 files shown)


davkean:
Note, behavior is identical to legacy with exception to; Project and ProjectItem objects. Legacy does not return a CATID for these but flavors do. Implemented them, so that flavors can just hang their extender properties directly on these.

sharwell:
Is there any guidance for extension authors who have since updated their NuGet packages to provide XAML-based customizations for the Properties window? I'd like to avoid a situation where duplicate properties start showing - one through the extender APIs and the other through the XAML rules.

davkean:
Extenders can opt in and out of objects that they extend via `CanExtend(object)`. Depending on the object you are extending, you should be able to walk the parent until you get to the owning project, at which point you can exclude CPS projects via https://github.com/Microsoft/VSProjectSystem/blob/master/doc/automation/finding_CPS_in_a_VS_project.md.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+                    return GetExtenderCATID(ExtendeeObject.FolderBrowseObject);
+
+                default:
+                case ExtenderCATIDType.Unknown:                                                 // EnvDTE.ProjectItem.Properties (when not file, folder, reference)
+                    BCLDebug.Assert(extenderCATIDType == ExtenderCATIDType.Unknown, $"Unrecognized CATID type {extenderCATIDType}");
+                    return null;
+            }
+        }
+
+        protected abstract string GetExtenderCATID(ExtendeeObject extendee);
```
Pilchie: Should this take an `IProjectTree` for the node for when the second issue above is fixed?

davkean: No, 'IProjectTree' is not a good representation of all the objects we're referring to, some, like a configuration, don't have a representation in the tree `at all`. The DTE objects themselves are actually a better representation at this point.

PR #3566, opened 2018-05-22
Use Properties Browser instead of Properties Window 

### LINKED ISSUE #3498: Can open strange Project Properties window
1. Right-click on a file and choose Properties which opens "Properties window" (as opposed to Properties tool window)
2. Click on Project

image

We should just prevent this entire window from showing in general and behave like legacy where only properties tool window is shown.

davkean:
This causes Properties on files and project to behave exactly like legacy by open the Properties Browser instead of Properties window on **Properties**.

Shell ends up querying browse object, which based on this setting decides whether the Properties Browser is used instead of Properties Window.

~This includes https://github.com/dotnet/project-system/pull/3484 and is blocked on it, so only review https://github.com/dotnet/project-system/commit/a0a636bf8a5087904162914892921e7f4e0d48b2 and https://github.com/dotnet/project-system/pull/3566/commits/faf75b8cf40f3e10805cce62c5a2b656a6d1e32e.~

~We need https://devdiv.visualstudio.com/DevDiv/Default/_git/CPS/pullrequest/124002?_a=overview to make this fix complete~ and fix #3498. This is merged in.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AdditionalFiles.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AdditionalFiles.xaml
@@ -5,6 +5,7 @@
   DisplayName="Additional File"
   PageTemplate="generic"
   Description="File Properties"
+  PropertyPagesHidden="true"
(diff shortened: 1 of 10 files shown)


davkean:
Split this apart from your changes @Pilchie.

Issue #3564, opened 2018-05-22
Putting ampersand in debugging profile cause it be treated as a mnemonic 


davkean:
We should escape these before putting in the menu:

![image](https://user-images.githubusercontent.com/1103906/40356171-a7fadaa8-5dfb-11e8-8e05-bad5efb90de3.png)

PR #3570, opened 2018-05-23
Escape mnemonics  for debug launch profiles

davkean:
Fixes: #3564.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/DebugProfileEnumValuesGenerator.cs
```diff
@@ -76,12 +76,16 @@ internal static ICollection<IEnumValue> GetEnumeratorEnumValues(ILaunchSettings
...
+                let value = new EnumValue { Name = profile.Name, DisplayName = EscapeMnemonics(profile.Name) }
                 select ((IEnumValue)new PageEnumValue(value))).ToList()
             );
 
             return result;
+        }
 
+        private static string EscapeMnemonics(string text)
+        {
+            return text?.Replace("&", "&&");
```
Pilchie: Is this a problem in the admittedly edge case where someone already has `&&` (maybe if it represents some condition in code)?

davkean: No I actually tested this, and works as you'd expect.

PR #3571, opened 2018-05-24
Do not use GlobalProvider off the UI thread

davkean:
GlobaProvider returns a provider per-thread, if you are called off the UI thread - you are going to have empty services. Graph Providers are now instantiated off the UI thread, based on the changes that were made to load them async by platform. Just import the new async service provider that was exported in 15.8.


davkean:
Pretty much - especially in a MEF component. in 15.8 both service providers are now provided by MEF. There are still some places where you don't have any context that you need to use GlobaProvider, but need to do that from UI thread.

davkean:
@Pilchie If it wasn't clear from PR - I went and fixed all locations that were broken.

PR #3581, opened 2018-05-26
Skip initial restore for a project if evaluation fails

### LINKED ISSUE #2867: Failed design-time builds can cause infinite loop of restores
When I changed the project to multiple platform, but forgot to change to TargetFrameworks.  It will be in a forever loop to send design time build requests.

```xml
<TargetFramework>netcoreapp1.1;net47</TargetFramework>
```

It looks like this code is executed to remove targets and add them back repeatedly, which causes repeated design time builds. 

```
Microsoft.VisualStudio.ProjectSystem.dll!Microsoft.VisualStudio.ProjectSystem.DataflowExtensions.LinkTo(System.Threading.Tasks.Dataflow.ISourceBlock<Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<Microsoft.VisualStudio.ProjectSystem.IProjectSubscriptionUpdate>> differencingSnapshotBlock, System.Threading.Tasks.Dataflow.ITargetBlock<Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<Microsoft.VisualStudio.ProjectSystem.IProjectSubscriptionUpdate>> target, System.Threading.Tasks.Dataflow.DataflowLinkOptions linkOptions, bool initialDataAsNew, bool suppressVersionOnlyUpdates, System.Collections.Generic.IEnumerable<string> ruleNames) Line 114	C#	Symbols loaded.
Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget.DependencySharedProjectsSubscriber.SubscribeToConfiguredProject(Microsoft.VisualStudio.ProjectSystem.IProjectSubscriptionService subscriptionService)	Unknown	No symbols loaded.
Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget.DependencySharedProjectsSubscriber.AddSubscriptionsAsync(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget.AggregateCrossTargetProjectContext newProjectContext)	Unknown	No symbols loaded.
Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget.CrossTargetSubscriptionHostBase.AddSubscriptionsAsync.AnonymousMethod__0()	Unknown	No symbols loaded.
Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.ExecuteJob<Microsoft.VisualStudio.Threading.EmptyStruct>(System.Func<System.Threading.Tasks.Task> asyncMethod, Microsoft.VisualStudio.Threading.JoinableTask job)	Unknown	No symbols loaded.
Microsoft.VisualStudio.Threading.dll!Microsoft.VisualStudio.Threading.JoinableTaskFactory.RunAsync(System.Func<System.Threading.Tasks.Task> asyncMethod, bool synchronouslyBlocking, Microsoft.VisualStudio.Threading.JoinableTaskCreationOptions creationOptions, System.Delegate entrypointOverride)	Unknown	No symbols loaded.
Microsoft.VisualStudio.ProjectSystem.dll!Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProjectAsync(Microsoft.VisualStudio.ProjectSystem.IProjectAsynchronousTasksService asyncTaskService, System.Func<System.Threading.Tasks.Task> action, bool registerFaultHandler) Line 221	C#	Symbols loaded.
Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll!Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget.CrossTargetSubscriptionHostBase.AddSubscriptionsAsync(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget.AggregateCrossTargetProjectContext newProjectContext)	Unknown	No symbols loaded.
```

jmarolf:
fixes #2867

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs
@@ -29,7 +29,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
             private readonly IActiveConfigurationGroupService _activeConfigurationGroupService;
             private readonly IActiveConfiguredProjectSubscriptionService _activeConfiguredProjectSubscriptionService;
             private readonly IProjectLogger _logger;
-            private bool _isFirstNomination = true;
 #pragma warning disable CA2213 // OnceInitializedOnceDisposedAsync are not tracked correctly by the IDisposeable analyzer
             private IDisposable _configurationsSubscription;
             private DisposableBag _designTimeBuildSubscriptionLink;
@@ -144,8 +143,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
 
             private void NominateProject(ImmutableList<IProjectValueVersions> sources)
             {
-                IVsProjectRestoreInfo projectRestoreInfo = ProjectRestoreInfoBuilder.Build(sources, _projectVsServices.Project, _isFirstNomination);
-                _isFirstNomination = false;
+                IVsProjectRestoreInfo projectRestoreInfo = ProjectRestoreInfoBuilder.Build(sources, _projectVsServices.Project);
 
                 if (projectRestoreInfo != null)
                 {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs
@@ -29,6 +29,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
             private readonly IActiveConfigurationGroupService _activeConfigurationGroupService;
             private readonly IActiveConfiguredProjectSubscriptionService _activeConfiguredProjectSubscriptionService;
             private readonly IProjectLogger _logger;
+            private bool _isFirstNomination = true;
 #pragma warning disable CA2213 // OnceInitializedOnceDisposedAsync are not tracked correctly by the IDisposeable analyzer
             private IDisposable _configurationsSubscription;
             private DisposableBag _designTimeBuildSubscriptionLink;
@@ -143,7 +144,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
 
             private void NominateProject(ImmutableList<IProjectValueVersions> sources)
             {
-                IVsProjectRestoreInfo projectRestoreInfo = ProjectRestoreInfoBuilder.Build(sources, _projectVsServices.Project);
+                IVsProjectRestoreInfo projectRestoreInfo = ProjectRestoreInfoBuilder.Build(sources, _projectVsServices.Project, _isFirstNomination);
+                _isFirstNomination = false;
 
                 if (projectRestoreInfo != null)
                 {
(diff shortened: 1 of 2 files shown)


davkean:
Bit confusing that CPS used the term "evaluation" here because this is processing design time build in the case of _PackageReferences_, _ProjectReferneces_ appears be coming from evaluation which is probably indicative of another bug (means we're not telling NuGet about references added during design-time build) that we should file.

I'm not sure I understand why Nat put a "is first nomination" check here - project state can change through the life of project. We may successfully run the first design-time build but run into more issues latter down the track, you can see this by adding something like to the project after it has loaded:

``` XML
  <Target Name="FailDesignTimeBuild" AfterTargets="CollectPackageReferences">
      <Error Text="Failed design-time build" />
  </Target>
```

We'll never catch this state.

davkean:
Whoops didn't mean to close this.

jmarolf:
> I'm not sure I understand why Nat put a "is first nomination" check here

I agree.  I think we always want to skip restore if a design-time build fails.

Pilchie:
IIRC there are cases where you need to restore in order to get targets files that will allow the design time build to succeed.  A bit of a chicken and egg problem :-/

jmarolf:
> there are cases where you need to restore in order to get targets files that will allow the design time build to succeed.

Can you give me an example.  We'll need to check for these cases or alternatively skip restore if we detect that we are in an endless loop.  I don't know which is going to be harder.

Pilchie:
I forget the details, but there was an issue where @natidea and @lifengl were discussing it. Will take a look when I'm back in the office.

Pilchie:
And also [this one](https://github.com/dotnet/project-system/issues/2867#issuecomment-361425562) and it's links.

davkean:
I think we need to come up with a design that factors that in without having a state that indicates "first open". Given we automatically load changes from disk when switching branches, your project could have drastically changed throughout its lifetime. "First open" as a concept doesn't really mean anything and will led users to restarting Visual Studio just to get back to that "first open" state which is not what we want.

jmarolf:
@davkean @Pilchie @lifengl 

## Current Behavior
here is my understanding of how restore works today.  In the diagram below starts once the project is opened/changed and ends once no more changes are made to the project file(s). Please correct me if I am wrong, but essentially the only way you escape the restore cycle is if no files change on disk.

project_nomination

The advantage of this behavior is that if there is a way for restore to succeed it will.  There are in fact some cases where design-time builds fail but we do a restore and everything starts working.  Unfortunately it also means that its rather easy for incorrectly written or broken project files to get us into an infinite restore state.

## Behavior in this PR

The change as it currently exists in this PR would essentially do the following, allowing us to stop the restore cycle in the event of a failed design-time build.
project_nomination

This means we can guarantee that the user will not get stuck in a infinite restore state.  Unfortunately I think this should be considered a breaking change as projects that used to work without issue will now fail.

## Proposed Behavior

I think the only way to handle this case at a high level is have the `restore info builder` track the design time builds that it queues and ignore changes from evaluation if that design-time build fails.  

project_nomination

@davkean @lifengl I am not aware of a mechanism to track design-time builds this way.  I assume that this would require changes to CPS to accomplish.  Does this approach make sense or is there an easier way?

lifengl:
Nat and I discussed this about 4 month ago, and he took a simple solution at that time to control risks in the ask mode, and plan to revisit it again.  Unfortunately, I forgot the detail of the discussion.

I do remember a few things to make the issue more complex:

1, a simple logic to skip restoring when DT build failed doesn't work.  When a developer made a bad change and downloaded package corrupted the DT build, you need a way to restore to the workable state.

2, using project change to prevent looping won't work.  The bug actually happens when a bad package carries a critical capability change, so every restoring will trigger the entire project system to be reloaded, and the loop happens crossing multiple sessions of the project system without any knowledge about the previous restoring.  It is always doing the first restoring operation after loading a project.

To prevent endless reloading a more persisted state/history of the package restoring will be needed.  The lifespan of that state need beyond the project system itself.

jmarolf:
> To prevent endless reloading a more persisted state/history of the package restoring will be needed. 

I agree that the only sane design I can come up with involves a sort of *restore state manager*.  Fundamentally we can't just keep trying the restore and hoping things work out. We need some kind of loop exit criteria and the only way we can build that is to have something that can track state of all the component across several different types of changes.

PR #3586, opened 2018-05-28
Look up single file generators and set CmdUIContext via legacy locations

davkean:
### Look up single file generators using legacy location
Fixes: #3535.

For C# and Visual Basic, generators in the legacy project system are looked up based on their package (not project type) - use that GUID.

For F#, generators in the legacy project system are looked up based on project type, use that.

### Activate legacy UI context for project is active
C# and VB activate a UI context (VSHPROPID_CmdUIGuid) that matches their package GUID (not their project type) when the project is active.

F# activates a UI context that matches its project type GUID when the project is active.

This is blocked on: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/122410?_a=overview.


davkean:
This can be merged in this, we've got sign-off on the names and have tested this together.

PR #3589, opened 2018-05-29
Change default startup project

davkean:
VS picks the first project in the solution as startup project on first open - make it so that first users get the "right" project as startup.


cartermp:
Interesting. In the Visual F# solution file, [this legacy project](https://github.com/Microsoft/visualfsharp/blob/master/VisualFSharp.sln#L17) is always picked even though it's not the first one listed.

PR #3592, opened 2018-05-29
[WIP] First approach at simplifying package reference usage across the tree

davkean:
This is the first approach at simplifying our package references across tree by following these strategies: https://github.com/davkean/maket & https://github.com/jeffkl/MSBuildSdks/blob/4cf8305df105b1751127c2b01df749c8e299005b/src/CentralPackageVersions/README.md. This is very similar to the what we're looking at productizing with MSBuild/NuGet.

The intention to is get rid of the majority of versions in Versions.props - and leave only those that are used in other contexts other than `<PackageReference />`.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,7 +5,7 @@
   <Import Project="build\RepoToolset\RestoreSources.props"/>
 
   <ItemGroup>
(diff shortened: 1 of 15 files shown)


davkean:
Going to close this out until it's completed.

PR #3597, opened 2018-05-30
Use 15.5 images for CI

davkean:


davkean:
Note, that VSTS is running 15.6, we do not have an image on Jenkins that matches 15.6, only 15.7. so we'll leave at 15.5 so that we don't accidentally take a dependency on 15.6 things

PR #3602, opened 2018-05-30
TFS611255 add UsePreviewSdk option

abpiskunov:
- Adding UserPreviewSdk option to .Net Core options page. 
- This setting has dynamic default value depending if VS was installed from Preview or Release channel.
- When user changing setting value, we create a sdk.txt file in the local AppData VS instance folder containing UsePreviews=True|False with current setting value. Absence of the file means setting has default value. Out-proc components will be able to determine default value by checking VS channel in the same way we do here.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
+++ b/src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
@@ -1,7 +1,13 @@
 ﻿' Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+Option Strict Off
+
+Imports System.IO
 Imports System.Runtime.InteropServices
+Imports System.Text.RegularExpressions
 Imports Microsoft.VisualStudio.Settings
+Imports Microsoft.VisualStudio.Setup.Configuration
+Imports Microsoft.VisualStudio.Shell.Interop
 
 Namespace Microsoft.VisualStudio.Editors.OptionPages
     Public NotInheritable Class GeneralOptions
@@ -9,10 +15,16 @@ Namespace Microsoft.VisualStudio.Editors.OptionPages
         Private Class SVsSettingsPersistenceManager
         End Class
 
+        Private Const PreviewChannelPattern As String = "VisualStudio\.[0-9]+\.Preview/"
+        Private Const UsePreviewSdkMarkerFileName As String = "sdk.txt"
         Private Const FastUpToDateEnabledSettingKey As String = "ManagedProjectSystem\FastUpToDateCheckEnabled"
         Private Const FastUpToDateLogLevelSettingKey As String = "ManagedProjectSystem\FastUpToDateLogLevel"
+        Private Const UsePreviewSdkSettingKey As String = "ManagedProjectSystem\UsePreviewSdk"
 
         Private ReadOnly _settingsManager As ISettingsManager
+        Private ReadOnly _shell As IVsShell
+        Private ReadOnly _localAppData As String
+        Private ReadOnly _isPreview As Boolean
 
         Public Property FastUpToDateCheckEnabled As Boolean
             Get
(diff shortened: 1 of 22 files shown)


src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
```diff
@@ -32,8 +44,110 @@ Namespace Microsoft.VisualStudio.Editors.OptionPages
...
+        End Sub
+
+        ''' <summary>
+        ''' Uses VS setup API, to determine if current VS instance came from Preview channel or not.
+        ''' Note: we depend here on channel manifest format. If it changes, we could be broken. There is
+        ''' an integration test in WTE repo to be executed in nightly vendors' runs to protect form manifest 
+        ''' format changes.
+        ''' </summary>
+        ''' <returns></returns>
+        Private Function IsPreview() As Boolean
```
davkean: Have we reached to setup for an API on this?

davkean:
I think we should add a telemetry point on this so that we can see who opts in and out.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
nguerrera: +1. I'm not looking forward to duplicating this.

abpiskunov: Preview is just the name of the channel. Which could change , but if it is in the API name it would be fixed. 

What exactly are we duplicating here? We are analyzing a string property of the VS instance.

In any case, i can ping them , however do you really believe they would add this right now? I am not sure it would happen, let' see. 


---
In reply to: [191930452](https://github.com/dotnet/project-system/pull/3602#discussion_r191930452) [](ancestors = 191930452)

abpiskunov:
Yes, sounds good.

---
In reply to: [393325610](https://github.com/dotnet/project-system/pull/3602#issuecomment-393325610) [](ancestors = 393325610)

abpiskunov:
I added email thread to the bug, there used to be slide deck from Angelos, but it seems the work was reduced only to this option and there no slide deck anymore. Motivation is to allow preview VS use preview sdk by default for msbuild and templates (when creating new projects) and stable VS should use only stable sdk. Thats why default value for this option is dynamic. If you have more questions check with Angelos, he might share the deck and give you more motivation details.

---
In reply to: [393324612](https://github.com/dotnet/project-system/pull/3602#issuecomment-393324612) [](ancestors = 393324612)

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
nguerrera: > What exactly are we duplicating

This function. Resolver will need to do the same, right? I don't love copying all this. Nobody reading our source code will have any idea what WTE is or why that test is good enough reason to do this via string matching.

davkean: You've written logic that is tied to the a technology we don't own or get notified when it changes, and is being tested by a set of tests that my team don't own or control while taking a dependency on an implementation detail. The first approach should be to talk to the team that owns this to ask for an API.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
```diff
@@ -32,8 +42,83 @@ Namespace Microsoft.VisualStudio.Editors.OptionPages
...
+        ''' Note: This setting is not roamed. After chatting with shell team, we decided not to roam 
+        ''' it due to stability issues.
+        ''' </summary>
+        ''' <returns></returns>
+        Public Property UsePreviewSdk As Boolean
+            Get
+                Return If(_settingsManager?.GetValueOrDefault(UsePreviewSdkSettingKey, _isPreview), _isPreview)
+            End Get
+            Set
+                _settingsManager.SetValueAsync(UsePreviewSdkSettingKey, Value, isMachineLocal:=False)
```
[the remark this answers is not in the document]
abpiskunov: I checked with Huy from shell team and we looked up the code that is using this option. It seems that isMahcineLocal is not related to roaming, it is determines if the setting would be shared on the machine across different SKUs and potentially instances even though we did not find how exactly it is reused since it is being stored to registry storage which is stored under instance folders.  We want False in this case, just to be safe.

---
In reply to: [192156768](https://github.com/dotnet/project-system/pull/3602#discussion_r192156768) [](ancestors = 192156768)

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
```diff
@@ -1,18 +1,28 @@
 ﻿' Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+Option Strict Off
```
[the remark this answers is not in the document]
abpiskunov: I had to add Option Strict Off , because it threw compile exception for this line Dim vsSetupConfig = New SetupConfiguration() , where we try to instantiate embedded object. It could not cast SetupConfiguration to it's embedded type.

---
In reply to: [192156136](https://github.com/dotnet/project-system/pull/3602#discussion_r192156136) [](ancestors = 192156136)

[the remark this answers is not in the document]
abpiskunov: Any other ways to use embedded objects in VB?

---
In reply to: [192230241](https://github.com/dotnet/project-system/pull/3602#discussion_r192230241) [](ancestors = 192230241,192156136)

abpiskunov:
@Pilchie / @davkean I followed with all comments, i assume i can complete this PR?

nguerrera:
One thing: we said honoring this checkbox in resolver is pri-1 vs the pri-0 to do the compat filtering. Are we prepared to hide/back out this checkbox if pri-1 doesn't land in time? Or should we hold this and coordinate when this goes in with when the resolver goes in?

davkean:
Oh the resolver doesn't work yet? We should wait for the resolver to respect this.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb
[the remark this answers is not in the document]
Pilchie: You shouldn't need Option Strict off for NoPIA - I suspect something else is going on, and we should work through that.

abpiskunov:
There are also web templates work that depends on it. I guess we can work on that with private build and hold checkin. However i dont want to checkin everything at the very last moment , it would leave not enough time for e2e testing.

Nick we have 15.8 and 15.9 which is whole summer,  do you feel that it Pri 1 would not fit there?

Thanks,
Anton

On May 31, 2018, at 5:20 PM, Kevin Pilch <notifications@github.com<mailto:notifications@github.com>> wrote:


@Pilchie commented on this pull request.

In src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptions.vb<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdotnet%2Fproject-system%2Fpull%2F3602%23discussion_r192272012&data=02%7C01%7Canton.piskunov%40microsoft.com%7C7367348c443d41a003e908d5c7557979%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636634092284024839&sdata=aph0MMKGj003j1QoYLjkOe8Aujy5NBEC2CEdJotoN%2Bo%3D&reserved=0>:

> @@ -1,18 +1,28 @@
 ﻿' Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.

+Option Strict Off


You shouldn't need Option Strict off for NoPIA - I suspect something else is going on, and we should work through that.

nguerrera:
I believe it will fit. I just want to know what the contingency plan is if we hit an unexpected snag. It could be that we'd have to hide the checkbox later, or maybe indicate temporarily that it only applies to templates.

> We have 15.8 and 15.9

Isn't this change targeting 15.8 already or do I just not understand the branches? If I've got it right and part of the work slips to 15.9, we shouldn't have a checkbox in 15.8 that doesn't do what it says.

davkean:
I completely forgot this applied to templates - can you change the text to indicate that?

Pilchie:
@nguerrera any update on the CLI pieces here?

nguerrera:
@Pilchie In PR here:

https://github.com/dotnet/cli/pull/9547
https://github.com/dotnet/core-setup/pull/4301

The biggest issue remaining is getting the code flow from core-setup to cli that we need. This may get in the way of getting it in for 15.8. I will also add you to a thread discussing this.

Pilchie:
Added the approved to merge label for posterity.  We're going to do a combined insertion of this with @nguerrera's CLI and resolver pieces for a 15.9 preview after some more testing.

Issue #669, opened 2016-10-28
Can't get PlatformTarget from EnvDTE


bricelam:
Similar to #665, I use to be able to do this...

``` C#
var platformTarget = project.ConfigurationManager.ActiveConfiguration.Properties
    .Item("PlatformTarget")?.Value as string;
```

...but in a .NET Core project, the `PlatformTarget` property doesn't exist.

bricelam:
### Workaround
```C#
var context = (IVsBrowseObjectContext)project;
var unconfiguredProject = context.UnconfiguredProject;
var configuredProject = await unconfiguredProject.GetSuggestedConfiguredProjectAsync();
var properties = configuredProject.Services.ProjectPropertiesProvider.GetCommonProperties();
var platformTarget = await properties.GetEvaluatedPropertyValueAsync("Platform");
```

davkean:
This has been fixed in 15.8 Preview 3.

Issue #667, opened 2016-10-28
Can't get OutputFileName from EnvDTE


bricelam:
Similar to #665, I used to be able to do the following.

``` C#
var outputFileName = project.Properties.Item("OutputFileName")?.Value as string;
```

However, the `OutputFileName` property doesn't exist on .NET Core projects.

bricelam:
### Workaround
```C#
var context = (IVsBrowseObjectContext)project;
var unconfiguredProject = context.UnconfiguredProject;
var configuredProject = await unconfiguredProject.GetSuggestedConfiguredProjectAsync();
var properties = configuredProject.Services.ProjectPropertiesProvider.GetCommonProperties();
var outputFileName = await properties.GetEvaluatedPropertyValueAsync("TargetFileName");
```

davkean:
This has been fixed in 15.8 Preview 3.

Issue #665, opened 2016-10-28
Can't get IntermediatePath from EnvDTE


bricelam:
In the old project system, I was able to get the intermediate output path (e.g. `obj\Debug`) using the following.

``` C#
var intermediatePath = project.ConfigurationManager.ActiveConfiguration.Properties
    .Item("IntermediatePath")?.Value as string;
```

However, the `IntermediatePath` property no longer exists.

I'm using a .NET Core Console project.

bricelam:
### Workaround
```C#
var context = (IVsBrowseObjectContext)project;
var unconfiguredProject = context.UnconfiguredProject;
var configuredProject = await unconfiguredProject.GetSuggestedConfiguredProjectAsync();
var properties = configuredProject.Services.ProjectPropertiesProvider.GetCommonProperties();
var intermediatePath = await properties.GetEvaluatedPropertyValueAsync("IntermediateOutputPath");
```

davkean:
This has been fixed in 15.8 Preview 3.

Issue #3622, opened 2018-06-06
We return different browse objects depending 


davkean:
Our implementation of how we implemented project and configuration properties is a mismatch of confusing types. As we bring up legacy compat we need to make sure behavior remains identical, regardless of whether you retrieve objects via IVsHierarchy or via DTE.

Here's our current state:

Object                                                       |   Description                      | CPS Implementation                                                         | Owner | Legacy Implementation
------------------------------------------------------------ |------------------------------------|----------------------------------------------------------------------------|-------|---------------------------------------------
EnvDTE.Project                                               | Project Automation Object          | `OAProject : EnvDTE.Project, EnvDTE.ISupportVSProperties`                  | CPS   | `VBProject : EnvDTE.IDispatch, ISupportVSProperties`
EnvDTE.Project.Object                                        | Project VS Automation Object       | `VSProject : VSLangProj.VSProject, VSLangProj.ProjectProperties`           | ML    | `CVsExtVSProject : VSLangProj.VSProject`
EnvDTE.Project.Properties                                    | Project Browse Object(1)           | `DynamicTypeBrowseObject : : IVsBrowseObject, IDispatch`                   | CPS   | `CVSProjectProperties : IVsBrowseObject, IDispatch, VSLangProj.ProjectProperties`
Sniff'd by AppDesigner to specially handle CPS               | Project Browse Object(2)           | `VSProject : VSLangProj.VSProject, VSLangProj.ProjectProperties`           | ML    | N/A
EnvDTE.Configuration                                         | Configuration Automation Object    | `CConfigurationAssignmentExt : EnvDTE.Configuration`                       | Plat  | `CConfigurationAssignmentExt : EnvDTE.Configuration`
EnvDTE.Configuration.Object/EnvDTE.Configuration.Properties  | Configuration Browse Object (1)    | `AbstractProjectConfigurationProperties : ProjectConfigurationProperties3` | ML    | `CVSProjectConfigProperties : IVsCfgBrowseObject, IDispatch, ProjectConfigurationProperties3`
N/A							                                 | Configuration Browse Object(2)     | `ProjectConfig : IVsCfgBrowseObject`                                       | CPS   | N/A

For both project and configs, we have two representations of our browse objects depending where you get the object from. We shouldn't have that concept, we should do the following:

- [ ] Split VSLangProj.VSProject and VSLangProj.ProjectProperties - these are two different objects.
[...]

PR #3623, opened 2018-06-06
Handle properties that "Update-Database" script from EntityFramework needs

davkean:
Earlier version's of [EF's Update-Database script](https://github.com/aspnet/EntityFrameworkCore/blob/release/2.0/src/EFCore.Tools/tools/EntityFrameworkCore.psm1) did a Project.Kind check to determine if .NET Core and if so read properties directly from via GetCommonProperties. In recent builds this kind changed to return legacy project system, but we missed this usage because it's since been changed to a "CPS" check. Downlevel packages, however, are still broken.

Implement the properties that they need:

#669
#667
#665 

Previous:

```
PM> Install-Package Microsoft.EntityFrameworkCore.Tools -Version 2.0.3
PM> Update-Database
Could not load assembly ''. Ensure it is referenced by the startup project ''.
```

New (which @bricelam tells me indicates that it worked):
```
PM> Install-Package Microsoft.EntityFrameworkCore.Tools -Version 2.0.3
PM> Update-Database
No DbContext was found in assembly 'WebApplication1'. Ensure that you're using the correct assembly and that the type is neither abstract nor generic.
```

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProject_VsLangProjectProperties.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProject_VsLangProjectProperties.cs
@@ -94,6 +94,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation
             }
         }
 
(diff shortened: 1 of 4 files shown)


davkean:
Side note, our browse object's are a little broken and I filed #3622 to track merging them.

PR #3632, opened 2018-06-07
Free threaded create and dispose

Pilchie:
If the Workspace supports it, we don't switch to the UI thread before creating or disposing the `IWorkspaceProjectContext`.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
@@ -83,8 +83,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
                 usedProjectContexts = _configuredProjectContextsMap.Values.ToImmutableHashSet();
             }
 
-            // TODO: https://github.com/dotnet/roslyn-project-system/issues/353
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
```diff
@@ -214,15 +216,18 @@ private void AddConfiguredProjectState(ConfiguredProject configuredProject, IWor
             if (string.IsNullOrEmpty(targetPath))
                 return null;
 
-
-            // TODO: https://github.com/dotnet/roslyn-project-system/issues/353
-            await _commonServices.ThreadingService.SwitchToUIThread();
+            bool continueOnCaturedContext = false;
```
whoisj: nit: misspelling. I think you meant `continueOnCapturedContext`

Pilchie: Thanks, fixed (rebased into original commit)

Pilchie:
Verified that I can open Roslyn.sln with these changes and @jasonmalinowski's free-threaded branch, and the language service does get hooked up.  Did not do any extensive performance testing or anything though.

@davkean - can you take a look and see if you have any concerns?

davkean:
It looks okay.

davkean:
Closing this out as it's no longer relevant.

Issue #3647, opened 2018-06-13
QueryDebugTargets returns an exception for SDK styled .NetFramework based projects


AbhitejJohn:
Repro steps:
1. Pull in the solution from here: https://github.com/dlemstra/NativeDebugTest 
2. Add a breakpoint in the native code.
3. Context click within the test in the editor and "Debug tests"

Expected:
That the breakpoint in the native code is hit because `nativedebugging:true`.

Observed:
The breakpoint isn't hit because querying for the below within Test Explorer returns a -2146233088 (looks like a general exception).
```
hr = queryDebuggableCfg.QueryDebugTargets((uint)flags, 0, null, number);
```

Code link in TE [here](https://devdiv.visualstudio.com/DevDiv/VS%20IDE%20Test%20Experiences/_git/VSUnitTesting?path=%2Fsrc%2FTestWindow%2FCore%2FShellExtensions%2FVsProjectExtensions.cs&version=GBmaster&_a=contents&line=891&lineStyle=plain&lineEnd=908&lineStartColumn=25&lineEndColumn=26)

Linked to #3495 .
/cc: @davkean

PR #3648, opened 2018-06-13
Do not fail IVsQueryDebuggableProjectCfg.QueryDebugTargets when called on library

davkean:
When called on a library, IVsQueryDebuggableProjectCfg.QueryDebugTargets was failing preventing Test Explorer from pulling the current debug settings.

QueryDebugTargetsAsync is called in two situations:

- When launching
- When someone queries the available debug targets

Similar to legacy project system, only fail when called during launch so that we can continue to show an error on F5'ing a library, but also be able to hand back launch data to consumers, such as Test Explorer.

Fixes: #3647.

Note, we now fall back the target path similar if RunCommand is not specified. This brings the behavior in line with legacy when called on a library.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
@@ -359,6 +359,64 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         }
 
         [Fact]
+        public async Task QueryDebugTargetsAsync_WhenLibraryWithRunCommand_ReturnsRunCommand()
+        {
+            var properties = new Dictionary<string, string>() {
+                {"RunCommand", @"C:\dotnet.exe"},
+                {"TargetFrameworkIdentifier", @".NETFramework" }
+                };
+
+            _mockFS.WriteAllText(@"C:\dotnet.exe", string.Empty);
+
+            var debugger = GetDebugTargetsProvider("Library", properties);
+
+            var activeProfile = new LaunchProfile() { Name = "Name", CommandName = "Project" };
+
+            var targets = await debugger.QueryDebugTargetsAsync(0, activeProfile);
+
+            Assert.Single(targets);
+            Assert.Equal(@"C:\dotnet.exe", targets[0].Executable);
+        }
+
+        [Fact]
+        public async Task QueryDebugTargetsAsync_WhenLibraryWithoutRunCommand_ReturnsTargetPath()
+        { 
+            var properties = new Dictionary<string, string>() {
+                {"TargetPath", @"C:\library.dll"},
+                {"TargetFrameworkIdentifier", @".NETFramework" }
+                };
+
+            _mockFS.WriteAllText(@"C:\library.dll", string.Empty);
+
+            var debugger = GetDebugTargetsProvider("Library", properties);
+
+            var activeProfile = new LaunchProfile() { Name = "Name", CommandName = "Project" };
+
+            var targets = await debugger.QueryDebugTargetsAsync(0, activeProfile);
+
+            Assert.Single(targets);
+            Assert.Equal(@"C:\library.dll", targets[0].Executable);
+        }
+
+        [Fact]
+        public async Task QueryDebugTargetsForDebugLaunchAsync_WhenLibrary_Throws()
+        {
+            var properties = new Dictionary<string, string>() {
+                {"TargetPath", @"C:\library.dll"},
+                {"TargetFrameworkIdentifier", @".NETFramework" }
+                };
+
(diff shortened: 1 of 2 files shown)


davkean:
tag @AbhitejJohn.

AbhitejJohn:
@davkean : Thanks for a quick fix. Is there a way I can patch this to buddy test?

davkean:
Yes:
```
> git clone https://github.com/dotnet/project-system/
> cd project-system
> git fetch upstream pull/3648/head:pr-3648
> git checkout 3648
> build.cmd
> Launch.cmd

```

davkean:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
```diff
@@ -360,6 +358,64 @@ public void GetExeAndArgumentsWithEmptyArgs()
...
+            var activeProfile = new LaunchProfile() { Name = "Name", CommandName = "Project" };
+
+            var targets = await debugger.QueryDebugTargetsAsync(0, activeProfile);
+
+            Assert.Single(targets);
+            Assert.Equal(@"C:\library.dll", targets[0].Executable);
+        }
+
+        [Fact]
+        public async Task QueryDebugTargetsForDebugLaunchAsync_WhenLibrary_Throws()
```
Pilchie: So, the difference between this and the test before is just whether the library exists on disk?  Is that right, or should this return the targetpath anyway, assuming that something will cause a build in between?

davkean: Ah, no it's not. This test is showing the previous behavior (that know only occurs during QueryDebugTargetsForDebugLaunch), but given I forgot to create the file on disk it could be either, I'll change the test.

davkean:
> Test Explorer sets VSDebugTargetInfo4.guidLaunchDebugEngine to ManagedAndNative even for NetCore projects and debugging still works.

Can you expand on this? Under what process do you run .NET Core tests?

> We actually set it to mixed explicitly because our process is always managed.

Not sure I understand. If you always set it to mixed, why would this bug exist in the first place where you didn't respect launchsettings?

davkean:
> Test Explorer sets VSDebugTargetInfo4.guidLaunchDebugEngine to ManagedAndNative even for NetCore projects and debugging still works.

I just tested this by returning the "managed" debugging engine for .NET Core projects and it fails badly. Can you point to the code? Are you sure you are launch a dotnet core process?

AbhitejJohn:
>just tested this by returning the "managed" debugging engine for .NET Core projects and it fails badly. Can you point to the code? Are you sure you are launch a dotnet core process?

The test platform does launch a net core host. [Here](https://devdiv.visualstudio.com/DevDiv/_git/VSUnitTesting?path=%2Fsrc%2FTestWindow%2FCore%2FVsHost%2FDebugLauncher.cs&version=GBmaster&_a=contents&line=333&lineStyle=plain&lineEnd=340&lineStartColumn=13&lineEndColumn=14) is the code I'm referring to.

>Not sure I understand. If you always set it to mixed, why would this bug exist in the first place where you didn't respect launchsettings?

The code above should answer this too. We only set it to Mixed if there is a native guid returned by the project system which it wasn't before this change.

davkean:
> We only set it to Mixed if there is a native guid returned by the project system which it wasn't before this change.

Can you point me to this exact code?

davkean:
Ah missed that you pointed to it.

davkean:
That says if we're "native-only" do mixed. I don't think it will be hit in this case.

AbhitejJohn:
@davkean : Well, we actually perform a contains search on all the guids returned which in this case is both the Core CLR and the native. That is just code from pre-core CLR time that needs to be changed.

davkean:
I think this is handling the native project system, and will never be hit for managed debugging. We'll never launch "native-only".

davkean:
Ah, I'm reading the code incorrectly and see what you are saying. I need to walk through this code and see what's going on and why that works for CoreCLR.

davkean:
Okay after debugging this, I agree it looks like unit testing is different from our launch:

### F5:

Type  | guidLaunchDebugEngine | pDebugEngines
-- | -- | --
Managed | ManagedCore | ManagedCore
Mixed | ManagedCore | ManagedCore, NativeOnly

### Unit Testing:

Type  | guidLaunchDebugEngine | pDebugEngines
-- | -- | --
Managed | ManagedAndNative | ManagedAndNative, ManagedCore
Mixed | ManagedAndNative | ManagedAndNative, ManagedCore, NativeOnly

Started a thread to see which one is correct.

davkean:
Either way we can merge this is as we're returning the same engines as F5 and it seemingly works both ways.

davkean:
I made one slight change from previous change - we stop validating other settings during QueryDebugTargetsAsync, which leads to a better experience when the working directory doesn't exist.

PR #3655, opened 2018-06-16
Run formatting across tree

davkean:
Ran the new formatting feature across the tree using _Format All Files_ extension.


src/Microsoft.VisualStudio.Editors/DesignerFramework/AccessModifierComboBox.vb
```diff
@@ -527,7 +527,7 @@ Namespace Microsoft.VisualStudio.Editors.DesignerFramework
             Catch ex As Exception When ReportWithoutCrash(ex, "Failed to determine if the access modifier combobox should be enabled", NameOf(AccessModifierCombobox))
                 Throw
             End Try
-            Return shouldBeEnabled
+            Return ShouldBeEnabled()
         End Function
```
davkean: Note this "appears" like a behavior change, but its not. VB was already binding this to ShouldBeEnabled(). It's actually dead code.

davkean:

sharwell:
I would have loved to see this in two commits:

1. Run formatting across the tree using *only* the settings prior to the new Code Cleanup feature
2. A second commit after running formatting a second time, this time with Code Cleanup

Obviously this would not be necessary if step 1 is a NOP for the current code base.

davkean:
You'd like to see that for interest sake, or for what reason? I'm using the features as they are designed, there's no two passes here that we provide.

sharwell:
The diff here is relatively big. I'm interested in seeing how much of that diff was caused by not running formatting earlier, and compare that to how much of the diff was directly caused by the introduction of a Code Cleanup feature in Format Document.

davkean:
If you are interested from a product perspective, you're welcome to run it yourself. :)

It's a massive pain to do this, so I'd prefer not to do it again; The fixing up of member refs, spacing and removal of line continuation is old formatter. Everything else is new formatting options.

sharwell:
> It's a massive pain to do this, so I'd prefer not to do it again;

Thanks, I understand 😄

PR #3659, opened 2018-06-18
Turn properties into fields

davkean:
Simplify this class into simple read-only fields.


davkean:
Ah, missed this was public. @abpiskunov why was this made public?

abpiskunov:
Hm it probably should not be public, it is not used in web repo and i don't know about other consumers out there.

davkean:
I've changed these back to properties to avoid an accidental break.

PR #3662, opened 2018-06-18
Switch to UI thread before touching UI-bound objects

davkean:
We were getting lucky that we were only ever initialized on UI thread, but we're async package so we can't touch GetService during InitializeAsync.


src/Microsoft.VisualStudio.ProjectSystem.FSharp.VS/Packaging/FSharpProjectSystemPackage.cs
```diff
@@ -30,13 +32,16 @@ public FSharpProjectSystemPackage()
         {
         }
 
-        protected override System.Threading.Tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
sharwell: ❕ Pass `cancellationToken`

src/Microsoft.VisualStudio.ProjectSystem.FSharp.VS/Packaging/FSharpProjectSystemPackage.cs
```diff
@@ -30,13 +32,16 @@ public FSharpProjectSystemPackage()
...
+
             _projectSelectorService = this.GetService<IVsRegisterProjectSelector, SVsRegisterProjectTypes>();
```
sharwell: :bulb: `GetServiceAsync` would be good to use here.

src/Microsoft.VisualStudio.ProjectSystem.FSharp.VS/Packaging/FSharpProjectSystemPackage.cs
```diff
@@ -30,13 +32,16 @@ public FSharpProjectSystemPackage()
...
             Guid selectorGuid = typeof(FSharpProjectSelector).GUID;
             _projectSelectorService.RegisterProjectSelector(ref selectorGuid, new FSharpProjectSelector(), out _projectSelectorCookie);
 
-            return base.InitializeAsync(cancellationToken, progress);
+            await base.InitializeAsync(cancellationToken, progress)
+                      .ConfigureAwait(true); // Avoid forced switch to thread-pool
```
sharwell: :bulb: In the latest vs-threading, I believe this would be an appropriate place to use `.ConfigureAwaitRunInline()` (https://github.com/Microsoft/vs-threading/pull/292). Not sure when that extension will be available for use in this repo though.

sharwell: Sure, was just pointing it out so you knew it was something that will exist in the future. 😄

PR #3679, opened 2018-06-28
Apply code fixes

davkean:
- Collection initialization can be simplified
- Apply accessible modifiers on partial classes
  - Usually we don't apply modifiers on partial declarations other than the main one, but until dotnet/roslyn#27837 is implemented we cannot enforce modifiers everywhere but these cases, so we just enforce them everywhere.


davkean:

davkean:

PR #3686, opened 2018-07-03
report SDK version information

jmarolf:
### Customer scenario
Collect telemetry about what versions of SDKs are in use for .NET Core with the hope to use this to notify users if there is an SDK update for their project.

### Bugs this fixes
[internal tracking bug](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/602105)

### Workarounds, if any
None

### Risk
Low

## FIX INFO

1. As it stands today this change will log telemetry for each configured project due to the way the `ProjectProperties` types lifetime works
2. In testing it was possible to deadlock project creation if we didn't move logging telemetry to a fire-and-forget task
3. Added an interface for testing as there is no way to mock `ProjectProperties`


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
+            {
+                await _projectVsServices.ThreadingService.SwitchToUIThread();
+                ConfigurationGeneral projectProperties = await _projectVsServices.ActiveConfiguredProjectProperties.GetConfigurationGeneralPropertiesAsync().ConfigureAwait(false);
+
+                var name = (string)await projectProperties.SDKIdentifier.GetValueAsync().ConfigureAwait(false);
+                var version = (string)await projectProperties.SDKVersion.GetValueAsync().ConfigureAwait(false);
+
+                if (_projectId == null)
```
Pilchie: When would this ever *not* be null here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,78 @@
...
+                ITelemetryService telemetryService)
+                : base(projectVsServices.ThreadingService.JoinableTaskContext)
+            {
+                _projectVsServices = projectVsServices;
+                _telemetryService = telemetryService;
+            }
+
+            protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
+            {
+                await _projectVsServices.ThreadingService.SwitchToUIThread();
```
davkean: You are switching to the UI thread, only to immediately schedule on ThreadPool very next line (https://github.com/Microsoft/vs-threading/blob/master/doc/cookbook_vs.md#should-i-await-a-task-with-configureawaitfalse).

Does the telemetry service need UI thread?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
davkean: What are we aiming to report here? What is SDKIdentifier/SDKVersion and who populates them?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            protected override Task DisposeCoreAsync(bool initialized)
+            {
+                _projectId = null;
+                return Task.CompletedTask;
+            }
+
+            private void InitializeProjectId()
+            {
+                IProjectGuidService projectGuidService = _projectVsServices.Project.Services.ExportProvider.GetExportedValueOrDefault<IProjectGuidService>();
```
davkean: Please use ISafeProjectGuidSevice and just import it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,78 @@
...
+                if (projectGuidService != null)
+                {
+                    _projectId = projectGuidService.ProjectGuid.ToString();
+                }
+                else
+                {
```
davkean: This path will never be executed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Telemetry;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
+{
+    /// <summary>
+    /// For maintaining light state about the SDK version used in a project
+    /// </summary>
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
```
davkean: You are reporting this for every single configuration that ever gets loaded, but only ever report the "Active" config. That does not seem correct.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,78 @@
...
+        private class SDKVersionTelemetryServiceInstance : AbstractProjectDynamicLoadInstance
+        {
+            private const string TelemetryEventName = "SDKVersion";
+            private const string ProjectProperty = "Project";
+            private const string NameProperty = "Name";
+            private const string VersionProperty = "Version";
+
+            private readonly IUnconfiguredProjectVsServices _projectVsServices;
+            private readonly ITelemetryService _telemetryService;
+            private string _projectId;
```
davkean: Instead of adding state here that you only use once, just have a method that returns the project id.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Telemetry;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
+{
+    /// <summary>
+    /// For maintaining light state about the SDK version used in a project
+    /// </summary>
+    [Export(ExportContractNames.Scopes.UnconfiguredProject, typeof(IProjectDynamicLoadComponent))]
```
davkean: You've now done reverse; you are saying that you live in "UnconfiguredProject" scope, but in your constructor you are importing a ConfiguredProject scope object; ProjectProperties. This won't compose.

jmarolf:

davkean:
Tell me when this ready for re-review.

jmarolf:
@davkean ready

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.cs
jmarolf: We now report for each configuration that is loaded.  Let me know if you think there is a better approach here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Telemetry/SDKVersionTelemetryTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            return new SDKVersionTelemetryServiceComponent(
+                projectProperties,
+                projectGuidSevice,
+                telemetryService,
+                projectThreadingService);
+        }
+
+        private static ITelemetryService CreateITelemetryService(Action<TelemetryParameters> onTelemetryLogged)
+        {
+            var telemetryService = new Mock<ITelemetryService>();
```
davkean: Pull these out into a factory so that we can reuse them.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/NETCoreSdkVersionProperty.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
+{
+    [Export(typeof(INETCoreSdkVersionProperty))]
```
davkean: We mock properties in lots of places (see ProjectPropertiesFactory.Create usage), I don't think you need this indirection.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                // Do not block initialization on reporting the sdk version. It is possible to deadlock.
+                Task.Run(async () =>
+                {
+                    string version = await _sdkVersionProperty.GetValueAsync().ConfigureAwait(false);
+                    string projectId = await GetProjectIdAsync().ConfigureAwait(false);
+
+                    if (string.IsNullOrEmpty(version) || string.IsNullOrEmpty(projectId))
+                    {
+                        _onNoSDKDetected(new NoSDKDetectedEventArgs(projectId, version));
```
davkean: Why isn't this just testing that we don't pass anything to the telemetry service? Is it due the lack of observability? There are probably simpler ways to fix that (such as storing the resulting task and waiting on that in the test).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/NETCoreSdkVersionProperty.cs
jmarolf: we can't mock general properties

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -0,0 +1,76 @@
...
+                _sdkVersionProperty = sdkVersionProperty;
+                _projectGuidSevice = projectGuidSevice;
+                _telemetryService = telemetryService;
+                _onNoSDKDetected = onNoSDKDetected;
+            }
+
+            protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
+            {
+                // Do not block initialization on reporting the sdk version. It is possible to deadlock.
+                Task.Run(async () =>
```
davkean: You have the danger of this task failing when the project is unloaded during load, such as via a veto. I think you'll wait to wrap this behind a LoadedProject.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Telemetry/SDKVersionTelemetryTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            var semaphore = new SemaphoreSlim(0);
+            bool success = false;
+            TelemetryParameters result = default;
+            void onTelemetryLogged(TelemetryParameters callParameters)
+            {
+                success = true;
+                result = callParameters;
+                semaphore.Release();
+            }
+            var component = CreateCoponent(guid, onTelemetryLogged, version);
```
davkean: "Component".

davkean:
I'm still not understanding what the expected behavior here is. Currently, you'll fire an event for _every_ configuration that is loaded. Is that expected? Would we ever see an SDK difference between one configuration to another?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/NETCoreSdkVersionProperty.cs
jmarolf: specifially `ProjectProperties` can be passed a project from which it pull project specific properties.  Unfortunately for this I need to test properties that are created by the dotnet SDK.

davkean:
For future reference, the deadlock probably occurs because of ISafeProjectGuidService - it waits until PrioritizedProjectLoadedInHost is fired, which won't until this component is loaded.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/NETCoreSdkVersionProperty.cs
davkean: See src\Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests\ProjectSystem\VS\Automation\VsLangProjectPropertiesTests.cs - OutputType.

jmarolf:
@davkean 
> Currently, you'll fire an event for every configuration that is loaded.

I pick up the `NETCoreSdkVersion` property from by defining a new read-only-string property in `ConfigurationGeneral.xaml`.  Any properties I define there I can only observe from a configured project. Is there a means to do this without loading for each configuration?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.cs
davkean: Okay, I understand. My comment before was that you needed to be consistent with property reading:

- Use the active configuration's project properties & export into UnconfiguredProject scope

-or-

- Use each configuration's project properties & export into ConfiguredProject scope

I suspect we really just want the first, you were mixing both in previous attempts.

davkean:
@jmarolf See this comment: https://github.com/dotnet/project-system/pull/3686#discussion_r200899705 to see if that clears it up, if not we can have a quick chat about it.

jmarolf:
@davkean updated

davkean:
Heads up you will be affected by https://github.com/dotnet/project-system/pull/3686. You need to change your base class to AbstractMultiLifetimeComponent and implement IProjectDynamicLoadComponent directly on your component.

jmarolf:
@davkean can you review the PR so that I can merge it and not need to deal with #3686?

davkean:
Ah, thought you were out for a week. Yep can do.

jmarolf:
I mean, I have my laptop.  I can probably fix you concerns in the evenings...

jmarolf:
@dotnet/project-system @davkean can I get another review pass please?

jmarolf:
Alright folks speak now or hold your peace

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Telemetry/SDKVersionTelemetryTests.cs
```diff
@@ -0,0 +1,130 @@
...
+using Xunit;
+using static Microsoft.VisualStudio.Telemetry.ITelemetryServiceFactory;
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    [Trait("UnitTest", "ProjectSystem")]
+    public class SDKVersionTelemetryTests
+    {
+        [Fact]
+        public static async Task TestCreateCoponentSDKVersionDefined()
```
Pilchie: Nit: typo: `Coponent`

Pilchie: And other copypasta in this file.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/xlf/ConfigurationGeneral.xaml.cs.xlf
```diff
@@ -152,6 +152,11 @@
         <target state="translated">Zakázat rychlou kontrolu aktuálnosti</target>
         <note />
       </trans-unit>
+      <trans-unit id="StringProperty|NETCoreSdkVersion|Description">
+        <source>Represents the version of the dotnet sdk being used (if any).</source>
+        <target state="new">Represents the version of the dotnet sdk being used (if any).</target>
```
Pilchie: @tmeschter: We're after string freeze for 15.8, but given that this property is marked as `Visible="False"`, I think it's okay, right?

jmarolf: spoke to @tmeschter in person. non-visible properties do not need to be localized

PR #3688, opened 2018-07-03
Improve reliability of TaskDelayScheduler tests

davkean:
These are failing often on Jenkins builds.


Pilchie:
Why does this make them more reliable?  Any way to achieve testing that doesn't depend on physical time delays? (using ManualResetEvent or something?)

davkean:
@Pilchie It makes them more reliable by increasing the time delays. I need to make product changes to make them 100% reliable - which was a distraction I didn't need, just wanted to quickly get in and have them stop failing for now.

PR #3689, opened 2018-07-04
Replace ImplicitlyActivated/Deactivated events with IImplicitlyActiveService

davkean:
See: https://github.com/dotnet/project-system/pull/3423/files#r178951672.

It became clear that consuming the activated/deactivated events in a manner that was thread-safe was difficult. Instead replace it with an service interface that automatically gets Activated/Deactivated when the configured project becomes implicitly activate or deactivated.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ConfiguredProjectImplicitActivationTrackingTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ConfiguredProjectImplicitActivationTrackingTests.cs
@@ -14,48 +14,6 @@ namespace Microsoft.VisualStudio.ProjectSystem
     public class ConfiguredProjectImplicitActivationTrackingTests
     {
         [Fact]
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ConfiguredProjectImplicitActivationTrackingTests.cs
```diff
@@ -324,11 +289,12 @@ public void IsImplicitlyActiveTask_WhenProjectUnloadCancellationTokenSourceHasBe
...
             {
                 callCount++;
-                return Task.CompletedTask;
-            };
+            });
+
+            service.ImplicitlyActiveServices.Add(implicitActiveService);
 
             configurationGroups = IConfigurationGroupFactory.CreateFromConfigurationNames("Debug|x86");
             await source.SendAndCompleteAsync(configurationGroups, service.TargetBlock);
```
davkean: Ugh - really hate how hard it is to test dataflow, need a better pattern.

PR #3706, opened 2018-07-11
Version padding for 15.8

tmat:
Keep this change in 15.8 branch only.

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -4,6 +4,13 @@
     <VersionBase Condition="'$(UseVisualStudioVersion)' != 'true'">2.8.0</VersionBase>
     <PreReleaseVersionLabel>beta2</PreReleaseVersionLabel>
 
+    <!-- 
+      15.8 specific workaround for a change to BUILD_BUILDNUMBER format that added two more digits. 
+      Any subsequent versions must have the same amount of digits to be considered higher.
+      Updates the version 20180102.03 to 20180102.0003.
+    -->
+    <BUILD_BUILDNUMBER Condition="$(BUILD_BUILDNUMBER.Length) == 11">$(BUILD_BUILDNUMBER.Insert(9, '00'))</BUILD_BUILDNUMBER>
+
     <!-- Opt-in repo features -->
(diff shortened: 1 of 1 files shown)


build/Versions.props
```diff
@@ -4,6 +4,13 @@
     <VersionBase Condition="'$(UseVisualStudioVersion)' != 'true'">2.8.0</VersionBase>
     <PreReleaseVersionLabel>beta2</PreReleaseVersionLabel>
 
+    <!-- 
+      15.8 specific workaround for a change to BUILD_BUILDNUMBER format that added two more digits.
```
jasonmalinowski: Tagging @dpoeschl specifically since this will require work to not take it when merging back to master.

jasonmalinowski: (although I guess it's written in a way that will be fine still...)

tmat: Let's not merge it to master.

davkean:
Can we front load the merge so its not accidentally merged? What about 15.9?

dpoeschl:
@davkean It's included in https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=1852025 which I'm making an insertion for currently.

dpoeschl:
Don't know about 15.9.

dpoeschl:
Actually now that I've said all that, I don't know what you meant. "front load the merge so its not accidentally merged"?

davkean:
What I mean is that we're going to get an auto-merge from 15.8 -> 15.9. Are we supposed to throw the commit away? I'd prefer we merge than revert.

tmat:
The change should only be in 15.8. Not in 15.9 nor master.

tmat:
merging and reverting the commit sounds good to me

dpoeschl:
So let it flow to 15.9 and then revert it there before it gets inserted? (noting that 15.9 insertions haven't even been on my radar so we have time)?

davkean:
I've gone ahead and done it here: https://github.com/dotnet/project-system/pull/3708 so that we don't forget about it.

PR #3707, opened 2018-07-11
Hide the "Auto-generated Binding Redirects" check box

tmeschter:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/636036

Hide the "Auto-generated Binding Redirects" check box when targeting .NET Core or .NET Standard. Binding redirects have no meaning for Core or Standard, so rather than showing a check box with no effect one way or the other we simply hide it.

When the target framework changes we re-evaluate and show or hide the check box as necessary. Right now there's no way to switch between Core/Standard and the "desktop" framework, but if that ever becomes possible we'll handle it appropriately.


davkean:
They don't have meaning for UWP, Xamarin, or Tizon either when they are targeted from CPS projects. This check should be the reverse; only .NETFramework.

tmeschter:
@davkean I'll make that change.

PR #3714, opened 2018-07-12
Add timeouts to renaming tests

davkean:
I suspect the renaming tests are blocking runs, this will at least help chase down the cause - and prevent runs being blocked for 60 minutes.


davkean:
@Pilchie These are test only changes, I'm merging them in so that I get headway into investigating failures - they have increased in past week.

PR #3721, opened 2018-07-12
LOC CHECKIN | dotnet/project-system dev15.8.x | 20180712

v-zbsail:


KevinRansom:
@pilchie, I would imagine that untranslated text, is inserted, when they are ready to insert it, rather than waiting on each string to be fully translated in all languages.

So my expectation is that nothing can be inferred from the fact that only 6 languages is in this particular block.

PR #3725, opened 2018-07-13
Add IProjectAccessor.EnterWriteLockAsync

davkean:
Adds the last usage of WriteLockAsync under a unit testable mockable method.

There will be follow-up PRs that consume this. Centralizing all write locks usage let's us react to this change in an easier way than CPS: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/132369.


davkean:
This method isn't testable itself - because IProjectLockService is not unit testable.

PR #3726, opened 2018-07-13
 Avoid blocking thread on releasing write lock

davkean:
Note: This includes https://github.com/dotnet/project-system/pull/3725, so only review https://github.com/dotnet/project-system/commit/66dfc46402d700a94723a3a32b7bdcce39bd176d.

Dispose synchronously blocks the thread until the write lock is released. Under the covers the project service yields on this path, which under thread pool exhaustion situations can cause Dispose to be blocked for a long time, exacerbating the problem. Await on ReleaseAsync so that the current thread isn't blocked and can be released back to the pool.

See https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/132369 for the CPS-version of this change.


davkean:

Issue #1058, opened 2016-12-30
RESX code generation does not respect <LogicalName>


davkean:
The RESX code generation does not respect the `<LogicalName>` metadata item on a `<EmbeddedResource/>` item. 

For example, the following:

``` XML
    <EmbeddedResource Include="Resources\Designer.resx">
      <LogicalName>Microsoft.VisualStudio.Editors.Designer.resources</LogicalName>
      <SubType>Designer</SubType>
      <CustomToolNamespace>My.Resources</CustomToolNamespace>
      <Generator>ResXFileCodeGenerator</Generator>
      <LastGenOutput>Designer.Designer.vb</LastGenOutput>
    </EmbeddedResource>
```

Should result in code gen with:

``` VB
        Friend Shared ReadOnly Property ResourceManager() As Global.System.Resources.ResourceManager
            Get
                If Object.ReferenceEquals(resourceMan, Nothing) Then
                    Dim temp As Global.System.Resources.ResourceManager = New Global.System.Resources.ResourceManager("Microsoft.VisualStudio.Editors.Designer", GetType(Designer).Assembly)
                    resourceMan = temp
                End If
                Return resourceMan
            End Get
        End Property
```

but actually results in:

``` VB
        Friend Shared ReadOnly Property ResourceManager() As Global.System.Resources.ResourceManager
            Get
                If Object.ReferenceEquals(resourceMan, Nothing) Then
                    Dim temp As Global.System.Resources.ResourceManager = New Global.System.Resources.ResourceManager("Designer", GetType(Designer).Assembly)
                    resourceMan = temp
                End If
                Return resourceMan
            End Get
        End Property
```

We hit this fixing: https://github.com/dotnet/roslyn-project-system/issues/1056

drewnoakes:
I believe this occurs because we don't return the correct value for `__VSHPROPID.VSHPROPID_DefaultNamespace` for the `.resx` file item.

http://index/?rightProject=Microsoft.VisualStudio.Design&file=Serialization%5CResXFileCodeGenerator.cs&line=201

drewnoakes:
Looked at this with Dave and concluded that `ResXFileCodeGenerator` needs extra logic to consume the logical name for the item, and decided to do this via the browse object.

drewnoakes:
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/257392?_a=files

davkean:
[...]

PR #3741, opened 2018-07-17
Fix access to resources in Microsoft.VisualStudio.Editors

tmeschter:
Commit 95751b18 added a new string to src/Microsoft.VisualStudio.Editors/Resources/Designer.resx. This caused the associated Designer.Designer.vb to be regenerated. Due to issue #1058, the newly generated code looked for resources under the "Designer" name instead of the proper "Microsoft.VisualStudio.Editors.Designer" name.

The end result was that none of these resources could be accessed at runtime, leading to failures such as https://devdiv.visualstudio.com/DevDiv/_workitems/edit/647167 where several of the project property pages wouldn't load at all.

Until #1058 is dealt with we have two options for fixing this:

1. Manually edit Designer.Designer.vb to switch back to the proper name. This is a small and easy fix, but this is a generated file. The next time someone added a resource to Designer.resx this would happen again, and we would waste another couple of days diagnosing the issue.
2. Rename Designer.resx to Microsoft.VisualStudio.Editors.Designer.resx so that the generated code would always have the proper name. The upside is that we don't have to worry about this issue cropping up again the next time we edit the .resx. On the other hand, changing the file name also changes the name of the generated class in Microsoft.VisualStudio.Editors.Designer.Designer.vb (previously just Designer.Designer.vb), requiring us to fix up the 80 or so files that access the resources through this class.

Here I've gone with option two. While this commit touches a lot of files, most of the individual changes are trivial. Note that I manually renamed and fixed up the associated .xlf files which should generally _not_ be done, but left to its own devices the XliffTasks tooling would generate all-new .xlf files with all the resources listed as "new" and in need of translation. This way we preserve the translations we have rather than having to wait a week for the Loc team to do a pass over resources that haven't really changed.

Also note that simply changing the name of the resource group to "Designer" wasn't an option, as other assemblies have a dependency on the existing "Microsoft.VisualStudio.Editors.Designer" name.


Pilchie:
This is approved to merge for 15.8 P5 by QB.

Issue #3752, opened 2018-07-19
F# - Pasting items that are outside of the project into a F# project results in a watson


TIHan:
After some dogfooding, I noticed that pasting an item that doesn't exist in a project into a F# project will result in a watson.

Pilchie:
@TIHan the fix for this was inserted in 15.8 and this can be closed, right?

TIHan:
I assume so since it was merged into dev15.8.x

davkean:
Fixed by https://github.com/dotnet/project-system/pull/3753.

PR #3753, opened 2018-07-19
Fixed F# project error when pasting items outside of a project into a project

TIHan:
This fixes this issue: https://github.com/dotnet/project-system/issues/3752

Basically, I forgot to capture the drop target in this case.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/PasteOrdering.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/PasteOrdering.cs
@@ -67,6 +67,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
 
         public bool CanHandleDataObject(object dataObject, IProjectTree dropTarget, IProjectTreeProvider currentProvider)
         {
+            _dropTarget = dropTarget;
             return PasteProcessor.CanHandleDataObject(dataObject, dropTarget, currentProvider);
         }
 


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/PasteOrdering.cs
```diff
@@ -67,6 +67,7 @@ private IPasteDataObjectProcessor PasteProcessor
 
         public bool CanHandleDataObject(object dataObject, IProjectTree dropTarget, IProjectTreeProvider currentProvider)
         {
+            _dropTarget = dropTarget;
```
davkean: The change looks fine, we can do this in a followup but why do we store it even when we don't handle it? Also, when do we clear the field? We could leak a bunch of data here from another app, right?

TIHan: I did some testing and each F# project gets an instance of "PasteOrdering". So they are isolated and no chance of leaks, AFAIK.

Hmm. After looking at this a bit more, I think I know how to make this make a bit more sense. We should clear the field after we are finished with it and I think after calling the `base.CanHandleDataObject(`, if that is true, I will store the `dropTarget`.

davkean: Ah I misread this, I thought you were storing "dataObject".

TIHan: Seems I can't do what I wish. So I'll answer your question regarding why we store the drop target even if we don't handle it.

The reason why we have to store the target even if we don't actually paste the item is because there is no other way. We relied on `ProcessDataObjectAsync` to give us the `dropTarget` which seems to have worked in most cases except for what the original issue suggests. If an item that is outside of any project, if I paste that item into a F# project, `ProcessDataObjectAsync` will not be called; I don't know why. But, `CanHandleDataObject` does get called.

TIHan: We don't clear the field, but I think we should after we use the `_dropTarget`.

davkean: Yeah we should clear it - it will hold onto data long after the item might be deleted/removed from the project.

davkean: As a separate bug though, it has nothing to do this with this change.

TIHan: I just tried clearing the field after we use the drop target, but `CanHandleDataObject` gets called again after "PasteItemsAsync" and the drop target gets set again, so clearing is useless.

Pilchie:
Do we have any indication from Watson how often this happens in practice?

TIHan:
I don't have any more info about how often the watson happens, I just know that the watson report gets called.

TIHan:
Ok, I checked and this has been hit ~120 times.

Pilchie:
Okay - can you get a VSO bug filed (I'm surprised there isn't one already if there are that many hits), and we'll see about taking this to VS shiproom for 15.8?

TIHan:
@Pilchie Sent an email with the bug filed and template filled out.

PR #3766, opened 2018-07-21
URL-escape percent signs in pre/post build events

### LINKED ISSUE #3506: Consider escaping % in Build Events
See feedback: https://developercommunity.visualstudio.com/content/problem/237752/prebuild-event-command-line-date-issue.html

MSBuild uses URL escaping to treat MSBuild special characters like `@`, `$`, and `%` as literals.

Unfortunately, if a user enters a Build Event that uses DOS-style environment variable expansion `%VARIABLE%` and the first two characters of the variable are valid hex digits, MSBuild will `urldecode` them into a character. This can be _very_ confusing, like the feedback case where the user was trying to say `%DATE%` but it was interpreted as `Úte%`.

It'd be nice if the project system handled escaping/unescaping such strings for users who use the VS UI to edit pre- and post-build events. I _think_ this would only be `%` characters followed by two hex digits--`urlencode`ing the whole user string would be way too much, since spaces are 100% legal and required for legibility. But scoping the round-trip could be difficult.

wjk:
This PR fixes the AbstractBuildEventValueProvider to properly escape MSBuild metachraacters signs added to the pre-build or post-build event strings in the Project Properties window of a .NET Core project.

Fixes #3506.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.AbstractBuildEventHelper.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.AbstractBuildEventHelper.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Threading.Tasks;
 
 using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties.InterceptedProjectProperties
@@ -78,7 +79,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties.InterceptedProjectP
 
                 if (execTask.Parameters.TryGetValue(Command, out string commandText))
                 {
-                    return commandText;
+                    return ProjectCollection.Unescape(commandText);
                 }
 
                 return null; // exec task as written in the project file is invalid, we should be resilient to this case.
@@ -149,7 +150,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties.InterceptedProjectP
             }
 
             private static void SetExecParameter(ProjectTaskElement execTask, string unevaluatedPropertyValue)
-                => execTask.SetParameter(Command, unevaluatedPropertyValue);
+                => execTask.SetParameter(Command, ProjectCollection.Escape(unevaluatedPropertyValue));
         }
     }
 }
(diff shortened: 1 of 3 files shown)


wjk:
And yet, bizarrely, when I test this functionality in VS _without_ my patched editor, it works just fine! It seems that VS has already fixed this internally. Of course, if I had known, I wouldn't have filed with PR to begin with! 😄 Do with this what you will. Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.cs
```diff
@@ -30,7 +32,22 @@ internal abstract partial class AbstractBuildEventValueProvider : InterceptingPr
             using (ProjectLockReleaser access = await _projectLockService.ReadLockAsync())
             {
                 Microsoft.Build.Construction.ProjectRootElement projectXml = await access.GetProjectXmlAsync(_unconfiguredProject.FullPath).ConfigureAwait(true);
-                return await _helper.GetPropertyAsync(projectXml, defaultProperties).ConfigureAwait(true);
+                string rawValue = await _helper.GetPropertyAsync(projectXml, defaultProperties).ConfigureAwait(true);
+                if (rawValue == null) return rawValue;
+
+                Regex regex = new Regex("%([0-9A-Fa-f][0-9A-Fa-f])");
```
jmarolf: Seems like any escape logic we add should go in `AbstractBuildEventHelper` instead of this class.

jmarolf:
see [here](https://github.com/dotnet/project-system/tree/bcab0444c672fc1a96477b627d8cd2e40b890d4a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties) for where our test currently live.

wjk:
@davkean I've added unit tests for the new functionality. I'm afraid I didn't fully understand what tests you wanted me to write. I'm not that familiar with this codebase. If there are any other tests you want me to write, please let me know (specifically, please) and I'll do my best to write them.

All backed-out commits have been removed.

wjk:
The Jenkins test failures indicate that the `ProjectCollection.Escape()` is escaping _all_ MSBuild metacharacters, not just the problematic `%` signs. Should I keep it this way and update the other tests, or change the escaping process to handle only `%` signs as the legacy project system does?

davkean:
@wjk Yep you are correct. This is my bad, I read the code wrong - digging in.

davkean:
Okay, legacy project system literally replaces `%25` with `%` and vice versa. That's it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.AbstractBuildEventHelper.cs
```diff
@@ -78,7 +79,7 @@ private string GetFromTargets(ProjectRootElement projectXml)
 
                 if (execTask.Parameters.TryGetValue(Command, out string commandText))
                 {
-                    return commandText;
+                    return ProjectCollection.Unescape(commandText);
```
davkean: I don't think this is correct; if I enter:

```echo %24(Bar)```

This will output:

```$(Bar)```

If we read that value, we'll un-encode the `%24` into `$` which will change the meaning of the line.

I think you literally just want the reverse of the encoding.

PR #3767, opened 2018-07-21
Implemented App.xaml special file provider

### LINKED ISSUE #116: Project System: App.xaml special file handler
_From @davkean on March 23, 2016 21:25_

PSFFILEID_AppXaml

_Copied from original issue: dotnet/roslyn#10034_

jp2masa:
Fixes #116

## Changes
- Implemented App.xaml special file provider.

Not sure if this needs tests, nor if the target capability is correct.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.CSharp/ProjectSystem/SpecialFileProviders/CSharpAppXamlFileSpecialFileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.CSharp/ProjectSystem/SpecialFileProviders/CSharpAppXamlFileSpecialFileProvider.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.Composition;
+
+using Microsoft.VisualStudio.IO;
+
+namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
+{
+    [ExportSpecialFileProvider(SpecialFiles.AppXaml)]
+    [AppliesTo(ProjectCapability.CSharp)]
+    internal class CSharpAppXamlFileSpecialFileProvider : AbstractFindByItemTypeSpecialFileProvider
+    {
+        [ImportingConstructor]
+        public CSharpAppXamlFileSpecialFileProvider(
+            IPhysicalProjectTree projectTree,
+            [Import(ExportContractNames.ProjectItemProviders.SourceFiles)] IProjectItemProvider sourceItemsProvider,
+            [Import(AllowDefault = true)] Lazy<ICreateFileFromTemplateService> templateFileCreationService,
+            IFileSystem fileSystem)
+            : base(projectTree, sourceItemsProvider, templateFileCreationService, fileSystem)
+        {
+        }
+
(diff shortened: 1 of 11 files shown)


davkean:
Appreciate the contribution, but I don't believe this is the correct change. The legacy project system logic is the following:

``` C++
    // First look in the current folder
    CLangFileNode* pFile = pFolder->GetFirstFile(FALSE);
    while (pFile != NULL)
    {
        if (pFile->IsInProject())
        {
            if (CompareNoCase(pFile->GetBuildAction(), APPLICATION_DEFINITION_BUILD_ACTION) == 0)
            {
                // We found it - bail out
                *ppAppXamlFileNode = pFile;
                return S_OK;
            }
        }
        pFile = pFolder->GetNextFile(pFile, FALSE);
    }

    // Then recursively look in subfolders
    CLangFolderNode* pSubFolder = pFolder->GetFirstFolder(FALSE);
    while (pSubFolder != NULL)
    {
        if (pSubFolder->IsInProject())
        {
            HRESULT hr = FindAppXamlFileInFolder(pSubFolder, ppAppXamlFileNode);
            if (hr != HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND))
            {
                // We either found it or we have a fatal error - bail out
                return hr;
            }
        }
        pSubFolder = pFolder->GetNextFolder(pSubFolder, FALSE);
    }

    // We couldn't find it
    return HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND);
```

You can see above, that it returns the first file it finds that is in the `ApplicationDefinition` item, not the first item named "App.xaml".

Secondly, looking internally, C# doesn't appear to have a template called "InternalWPFApplicationDefinition.zip" - and hence it's going to fail when it's missing. I think we should avoid heading down the creation path if we're in C#.

This is going to be significantly more work than above, so give me a heads up if you want to continue to do this and I can guide you through it.

jp2masa:
I'd like to continue this, and I'll probably need some help. I have some questions:
1. Should the class implement `ISpecialFileProvider` directly or extend `AbstractSpecialFileProvider`? `AbstractSpecialFileProvider` seems to assume that `TemplateName` is always provided, but in this case there's no template for C#, so it would have to be changed a bit.
2. If I understand correctly, the code from the legacy project system iterates through the project tree to find the first item for which the build action is `ApplicationDefinition`. Should I use `IPhysicalProjectTree` to get the root node and then search from there? If so, how to get the build action?
3. As there's no template for C#, what should be done when `SpecialFileFlags.CreateIfNotExist` is specified?

davkean:
1. IProjectItemProvider.GetItemsAsync(string) provides a way to look up a project item based on build action/item type. You can see how AbstractSpecialFileProvider "imports" it.
2. There's really two ways to look up special files; one by "name" and the other by "item type". I would rename AbstractSpecialFileProvider to indicate the former (maybe "AbstractFindByNameSpecialFileProvider") and introduce a AbstractFindByItemTypeSpecialFileProvider that implements the above logic. There's probably shared logic that they both share that could either come in another base class or shared helpers. Do the rename in a separate commit.
3.  For this, throw NotImplementException.

davkean:
Is this ready for review?

jp2masa:
I think so, but I have no idea how to check if it works and also if this needs tests.

davkean:
@jp2masa Sorry for the late followup on this. This code has significately changed since this PR was last updated. https://github.com/dotnet/project-system/pull/5507 contains the reworked version of this bug on top of the new base classes and closing this in lieu of that.

PR #3768, opened 2018-07-22
Added support for launching a class library project using the run project command

### LINKED ISSUE #1484: Support launching VSIX projects
VSIX projects are class libraries, but the VSSDK sets some variables to say it should launch VS in the dev hive. We should respect those flags so a launchSettings.json isn't needed to start VS.

jp2masa:
Fixes #1484

## Changes

- Added support for launching a class library project using the run project command. This allows running a class library project using the `RunCommand` project property without needing to create a launch profile.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
@@ -413,7 +413,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         }
 
         [Fact]
-        public async Task QueryDebugTargetsForDebugLaunchAsync_WhenLibrary_Throws()
+        public async Task QueryDebugTargetsForDebugLaunchAsync_WhenLibraryAndNoRunCommandSpecified_Throws()
         {
             var properties = new Dictionary<string, string>() {
                 {"TargetPath", @"C:\library.dll"},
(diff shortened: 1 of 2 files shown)


davkean:
Can you explain the behavior you are introducing? I've read it a few times and I'm not understanding it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
```diff
@@ -377,7 +377,7 @@ public void GetExeAndArgumentsWithEmptyArgs()
         }
 
         [Fact]
-        public async Task QueryDebugTargetsAsync_WhenLibraryWithoutRunCommand_ReturnsTargetPath()
+        public async Task QueryDebugTargetsAsync_WhenLibraryWithoutRunCommand_ReturnsEmptyExecutablePath()
```
davkean: This will break unit testing debugging, this was a very deliberate change in https://github.com/dotnet/project-system/pull/3648.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
```diff
@@ -391,25 +391,7 @@ public void GetExeAndArgumentsWithEmptyArgs()
             var targets = await debugger.QueryDebugTargetsAsync(0, activeProfile);
 
             Assert.Single(targets);
-            Assert.Equal(@"C:\library.dll", targets[0].Executable);
```
davkean: This will break Test Explorer debugging, this was a very deliberate change in https://github.com/dotnet/project-system/pull/3648. You should make this test pass.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
```diff
@@ -391,25 +391,7 @@ public void GetExeAndArgumentsWithEmptyArgs()
...
-        }
-
-        [Fact]
-        public async Task QueryDebugTargetsAsync_WhenLibraryWithoutRunCommand_DoesNotManipulateTargetPath()
```
davkean: You should make this test pass, as above this is testing Test Explorer scenarios.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Debug/ConsoleDebugLaunchProviderTests.cs
```diff
@@ -453,7 +435,6 @@ public void GetExeAndArgumentsWithEmptyArgs()
         [Theory]
         [InlineData("winexe")]
         [InlineData("appcontainerexe")]
-        [InlineData("library")]
```
davkean: This test should continue to pass.

davkean:
This update looks great! Thanks for the contribution!

PR #3774, opened 2018-07-24
Handle renames events without new name and old name

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/653144.

The other events do not suffer from the same problem.

You can see how this behavior occurs here: https://referencesource.microsoft.com/#System/services/io/system/io/FileSystemWatcher.cs,541.


davkean:
This has about 2000 lifetime hits, and about 50 in the 15.7. It's our most hit Watson in this code base.

PR #3775, opened 2018-07-24
Fix Resources Designer crashes under Italian

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/653172.

Italian had a bad placeholder causing this path to crash when you entered a resource with a bad identifier.


davkean:
Filed: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/653174 to prevent the next hand-off from overwriting this change.

jmarolf:
While this is the correct fix, won't this get blown away the next time our xlf generation task runs? @nguerrera and @tmeschter what is the correct way to prevent this?

davkean:
I can see it getting blown away next hand off which is what the above linked bug is about (now fixed), but why would it get blown away based on the xlf generation?

jmarolf:
@davkean I may be misunderstanding the root cause.  Was it just a bad translation or did the xlf tasks not escape correctly?

davkean:
Ah, yeah, it's a bad translation.

tmeschter:
@davkean Do we need to fix this for 15.8? That is, what will happen to VS when we try to format the Italian translation of this string?

davkean:
@tmeschter VS crashes. Given the number of hits this has on watson (5), this would not meet the bar.

PR #3776, opened 2018-07-24
[15.8] Fix telemetry loading configuration too early

davkean:
The new SDK telemetry started faulting configuration in at a time that it previously wasn't loaded, which uncovered some race conditions. Wait until the project is loaded before we run.

This is considered the least risky change vs fixing the race conditions which we will do at a later time.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
@@ -42,6 +42,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                 // Do not block initialization on reporting the sdk version. It is possible to deadlock.
                 Task.Run(async () =>
                 {
+                    // Wait for the project to be loaded so that we don't prematurally load the active configuration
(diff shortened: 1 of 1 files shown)


davkean:
This is an alternative fix to https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/653090 and https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/652229.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -42,6 +42,10 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
                 // Do not block initialization on reporting the sdk version. It is possible to deadlock.
                 Task.Run(async () =>
                 {
+                    // Wait for the project to be loaded so that we don't prematurally load the active configuration
+                    await _unconfiguredProjectTasksService.ProjectLoadedInHost
```
davkean: Note, this is our version of this API and will be cancelled if the project is unloaded unlike CPS's IProjectAsyncLoadDashboard.

lifengl:
Yes, I can see the component can be initialized during unconfigured project loading, and it accesses configured project, which can trigger it to be loaded before the project system initialize the COM layer and the intended code to compute and load first configuration, which can lead many things to fail.

jmarolf:
merging so we can get an insertion

PR #3782, opened 2018-07-24
fixing tab ordering in debug options page

jmarolf:
Fixes this VS bug: 
- [622505](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/622505)

Fix description:

- Today we do not set the tab ordering for the debug page.  This change sets the first item on the debug page as a place for tab ordering to stop at as well as indicating that it should be first in the tab order

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageControl.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageControl.xaml
@@ -70,6 +70,8 @@
                 VerticalContentAlignment="Center"
                 MinHeight="23"
                 Margin="5,6,2,7"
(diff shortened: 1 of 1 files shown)


Pilchie:
I thought that in the absence of explicitly setting it, WPF did it in the order controls appear on the page.  Do we understand why that wasn't the case here?

davkean:
Yeah, that Combo is definitely a TabStop - so I have the same question.

jmarolf:
@Pilchie we want this for 15.8 or master?

Pilchie:
Given the low risk, I'd take to the QB for 15.8 still.

jmarolf:
alright, re targeted.  I'll fill out the ask mode template

jmarolf:
@Pilchie I've updated the [bug](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/622505) for QB approval

Pilchie:
Approved by QB on Friday - forgot to update the label here.

Issue #2881, opened 2017-10-11
Shift+Space in the managed resource editor doesn't make a space


natidea:
From [VSFeedback](https://developercommunity.visualstudio.com/content/problem/128679/shiftspace-in-the-managed-resource-editor-doesnt-m.html) | [506825](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/506825)

Repro Steps:

1. If necessary, create a new managed resource file (*.resx)
2. Open it in the managed resource editor.
3. Edit a string value.
4. Try to type something IN ALL CAPS by keeping the shift key held down as you type.

Observe that as you type while holding the shift key, spaces are ignored. This is very annoying if you're trying to type just a few words in caps.

davkean:
Fixed by https://github.com/dotnet/project-system/pull/3786.

PR #3786, opened 2018-07-24
Resx string editor quality-of-life improvements

wjk:
This PR does two things:

1. Fixes the bug described in #2881.
2. Changes the `ResourceStringTable` to automatically change the size of its rows to show the complete cell value immediately upon committing edits to a cell; previously you had to manually click out of and back into an affected cell to have the row resize. This makes editing long strings in resx files much less annoying.

If there is a way to ensure that the text box used to edit the value itself similarly resizes to fit its text, I'm all ears, but I don't know if that is possible. Thanks!

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
@@ -1870,14 +1870,6 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
             End Sub
         End Class
 
-        Protected Overrides Function ProcessDataGridViewKey(e As KeyEventArgs) As Boolean
-            If e.KeyCode = Keys.Space AndAlso e.Shift AndAlso IsCurrentCellInEditMode Then
-                Return False
-            Else
-                Return MyBase.ProcessDataGridViewKey(e)
-            End If
-        End Function
-
     End Class
 
 End Namespace
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
@@ -87,6 +87,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
             AllowUserToDeleteRows = False
             EditMode = DataGridViewEditMode.EditOnKeystrokeOrF2
             AutoSizeColumnsMode = DataGridViewAutoSizeColumnsMode.Fill
+            AutoSizeRowsMode = DataGridViewAutoSizeRowsMode.AllCells
             DefaultCellStyle.WrapMode = DataGridViewTriState.True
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
```diff
@@ -1869,6 +1870,14 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
             End Sub
         End Class
 
+        Protected Overrides Function ProcessDataGridViewKey(e As KeyEventArgs) As Boolean
+            If e.KeyCode = Keys.Space AndAlso e.Shift Then
+                Return False
+            Else
+                Return MyBase.ProcessDataGridViewKey(e)
```
Pilchie: What does the DataGridView normally treat Shift+Space as?  Will this change break expectations for something people expect from DGV in other applications?

wjk: As far as I can tell, Shift+Space selects either the entire row, or the entire column, depending on the value of `DataGridView.SelectionMode`. In practice, this highlights the entire row, without losing the edit focus. This has no practical benefit (to me, at least), and it also eats the space input, which is extremely annoying when typing text with the spacebar held down (to type all-caps text, or to type an ampersand between words, for example). I have never heard of anyone actually using this "feature" anywhere, and in fact I'd reckon that other DataGridView users have become just as confused by the same quirk that the resx editor suffers from.

davkean: I think I'd prefer this to be limited to the only when editing the content of the cells, there's no reason to eat CTRL+Space when not in editing mode. This lets you easily select and delete a cell when you moving the selection around for a single cell.

src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
```diff
@@ -87,6 +87,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
             AllowUserToDeleteRows = False
             EditMode = DataGridViewEditMode.EditOnKeystrokeOrF2
             AutoSizeColumnsMode = DataGridViewAutoSizeColumnsMode.Fill
+            AutoSizeRowsMode = DataGridViewAutoSizeRowsMode.AllCells
```
davkean: To call out the behavior change for reviewers; this means that rows will automatically be sized to fit content - and you can no longer use mouse to resize heights. I'm okay with this.

src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceStringTable.vb
davkean: It looks like DataGridView.IsCurrentCellInEditMode indicates that.

davkean:
This looks great. I just made one change - I pushed it down to the base class so that the Settings editor also gets the new behavior.

davkean:
Merged this in, thanks again for the contribution!

PR #3787, opened 2018-07-24
Display active debug framework menu with multiple startup projects

### LINKED ISSUE #2855: "multiple startup projects" dont mix with "multiple target frameworks"
Add two console apps to a solution
```
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>netcoreapp2.0;net452</TargetFrameworks>
  </PropertyGroup>
</Project>
```

Set one as the startup project and u can choose the framework

image

set both as startup projects

image

you can now no longer select which framework to run 

image

bording:
This fixes #2855 by displaying the active debug framework menu when there are multiple startup projects.

Currently, the menu is only shown when all of the startup projects have an identical `TargetFrameworks` property. Even having the same frameworks but defined in a different order is enough to prevent the menu from displaying. While this could be relaxed, I wasn't sure how to handle the scenario where the menu needs to default to the first entry (when there is no active framework yet) when there are different "first" entries per project.

I've updated and added tests, but let me know if there are any more you'd like for me to add around this.


Pilchie:
Thanks for the contribution @bording!

@davkean @BillHiebert @abpiskunov can you take a look?

davkean:
@Pilchie This is in 15.9, but not sure it fits meets spirit of the 15.9 release. I think we want to pull this in 16.0.

Pilchie:
Yes, I'd like to consider this for 15.9 if it looks solid and is ready soon enough.

bording:
> Yes, I'd like to consider this for 15.9 if it looks solid and is ready soon enough.

@Pilchie Do you have any sort of time frame you can share of the deadline here?

bording:
> One thing we might want to do is to put a disabled placeholder there when we can't show the menu indicating why - it's a little confusing when it doesn't show up when you expect it to, because you had the wrong order of TFMs.

@davkean Can you explain a bit more as to what you'd want to see in this case, what the "indicating why" would look like?

bording:
> For 15.9, we have plenty of time (months) to get it in.
 
Okay, that's good news!

>I thought about this over the weekend, and I think we should live with the current behavior you've implemented for this PR as improves the existing behavior but long term follow up with a change to cover the situations where the menu doesn't show up

While working on this PR, I had been thinking that the current UI was too limiting, but figured it made sense to consider additional changes as a followup to getting this in. 

What you're proposing seems like it would address most of the things I had been thinking was too limiting by the single menu. With your proposed idea, what would happen to the "All" option if `net46` was chosen for ConsoleApp2 since there would no longer be a single common framework for all the projects?

One reason I didn't try to go down the route of doing something like you're proposing here is that I wasn't quite sure how to actually implement something like that in the menu VSCT file. There doesn't appear to be a way to have a dynamically determined menu (a Framework for each project), only for buttons. (the `DynamicItemStart` command flag only works for buttons).

BTW, I'm going on vacation tomorrow and won't be able to respond to additional comments until I get back next week, in case that alters the priority for following up on this.

bording:
@davkean FYI I am back from vacation and can continue working on this now.

davkean:
@BillHiebert Now you are back, can you take a look at this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/IStartupProjectHelper.cs
```diff
@@ -1,14 +1,16 @@
...
 {
     /// <summary>
-    /// Provides a mechanism to get an export from a DotNet project which is the only startup project. The capabilityMatch is  
+    /// Provides a mechanism to get an export from a DotNet project which is the only startup project. The capabilityMatch is
     /// used to refine the projects that are considered
     /// </summary>
     internal interface IStartupProjectHelper
     {
-        T GetExportFromSingleDotNetStartupProject<T>(string capabilityMatch) where T : class;
+        List<T> GetExportFromDotNetStartupProjects<T>(string capabilityMatch) where T : class;
```
drewnoakes: Does this have to return `List<T>`? Can you use `ImmutableArray<T>` or `IReadOnlyList<T>` instead?

bording: Either of those options should work. Which one should I change it to?

drewnoakes: The solution has the following usages:

- 34 `IReadOnlyList<T>`
- 15 `IImmutableList<T>`
- 60 `ImmutableList<T>`
- 168 `ImmutableArray<T>`

Looks like `ImmutableArray<T>` would be the more common choice, especially if the return value is not being manipulated (as the array variant copies internal storage on modification, whereas the list variant shares internal structure).

bording: I've switched to `ImmutableArray<T>`.

bording:
I really want to make sure this gets into 15.9, if at all possible. Is there anything else I can do to help make that happen?

jmarolf:
@BillHiebert and @davkean what else needs to happen for this PR?

bording:
@jmarolf Based on on the fact that you just changed the base branch, I assume this means it won't be able to make 15.9? If so, that's rather disappointing since one of the reasons I opened the PR in the first place was to get it into 15.9, and I had been told there was plenty of time to get it in there when I opened it.

jmarolf:
@bording I am very sorry that this did not make it in time.  I just did a bar check and this did not make it unfortunately.

jmarolf:
@bording There was a mock rename in the dev16 branch https://github.com/bording/project-system/pull/1

PR #3797, opened 2018-07-27
Add support for metafiles to the Resx editor

### LINKED ISSUE #3795: Resources editor does not accept WMF files
In a Windows Forms project, it's not possible to add a WMF file resource using the GUI. 

If I click Add Resource -&gt; Existing Resource, metafiles are not listed in the file browser's list of acceptable file types, however, it does seem that the resx file can actually handle wmf file resources properly, for the most part.

The reason why I think the resx file can handle WMF, is if I hand edit the resx file with the following line:

```
<data name="TestWMF" type="System.Resources.ResXFileRef, System.Windows.Forms">
<value>..\Resources\Sample.wmf;System.Drawing.Imaging.Metafile, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
</data>
```
The editor shows it in an &quot;other&quot; category, but it seems that it has accepted it as a resource, and it can be assigned to controls like ToolStripButton, etc.

Interpreting the term &quot;Image&quot; in a way that implies that it refers to System.Drawing.Image, I would expect that this category would accept both Bitmap and Metafile objects.

I considered posting this as a suggestion but I leaned toward &quot;bug&quot; on this one, sorry for the mixup if this is considered &quot;not a bug&quot;.

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/199180/resources-editor-does-not-accept-wmf-files.html
VSTS ticketId: 568197_
(no comments)
(no solutions)

wjk:
Does just what the title says. I tested it with a metafile of my own creation, and it is added to the resx file, and from there to a PictureBox on a Form, with no issues.

Note: If you need a metafile to test with, just run `sc Sample.wmf $null` at a PowerShell prompt, then double-click the resulting empty file to open it in Paint. Add some content, then save, and poof: you now have a valid WMF file.

Fixes #3795.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceTypeEditorBitmap.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceTypeEditorBitmap.vb
@@ -23,7 +23,8 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
             EXT_GIF,
             EXT_JPEG, EXT_JPG,
             EXT_PNG,
-            EXT_TIF, EXT_TIFF}
+            EXT_TIF, EXT_TIFF,
+            EXT_EMF, EXT_WMF}
 
         ' Extensions supported in a device project
         ' NOTE: WinCE does not support TIF file...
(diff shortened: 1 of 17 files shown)


davkean:
You're on fire! That was quick.

PR #3805, opened 2018-07-30
YAML build for dev15.8.x

jmarolf:
adding yaml build

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -1,0 +1,109 @@
+resources:
+- repo: self
+  clean: true
+queue:
+  name: VSEng-MicroBuildVS2017
+  demands: Cmd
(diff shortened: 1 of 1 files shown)


Pilchie:
How is this different from the version against master (or is it)?

jmarolf:
@Pilchie it is not different.  We do not merge master into dev15.8.x currently.  We would want to merge this if we are OK with the YAML build being turned on for dev15.8 and dev15.9 in addition to master

davkean:
@jmarolf Next time to avoid the additional commits and merge commits (I think there will be in total 9 commits for effectively a single commit) you can create a merge base like so:

```
git merge-base master dev15.8.x
git branch [branchname] [commit-sha]
```

PR #3806, opened 2018-07-30
YAML build for dev15.9.x

jmarolf:
adding yaml build


Pilchie:
How is this different from the version in master/15.8?  SHould we just merge in 15.8 and let it flow with auto-merges?

Issue #3722, opened 2018-07-12
XamlAppDef not available


lokki:
We are in a process of converting a few large-ish solutions to the SDK project format. A few of those contain Workflow Foundation workflow definitions in .xaml files. These should be using the XamlAppDef build action but it's not available in the new projects:
<img width="352" alt="new" src="https://user-images.githubusercontent.com/1587351/42652929-eb5892c8-85e1-11e8-9c86-37c715461331.png">
vs old format:
<img width="350" alt="old" src="https://user-images.githubusercontent.com/1587351/42652956-f548b8e4-85e1-11e8-9976-2494fbe51ccb.png">

Why is XamlAppDef not available? If XAML is not supported then why are Page and ApplicationDefinition available in the new format?

Thanks

davkean:
We don't respect [AvailableItemName](https://github.com/dotnet/project-system/issues/113) yet so they need to manually added to support. What is XamlAppDef used for?

lokki:
It is used for the Workflow Foundation workflows which are defined in .xaml files. Without it they won't compile

lokki:
@davkean is there any chance it's a quick fix I can do myself? Or tooling/project system change needs to happen first?
Thanks

davkean:
If you add the following experimental feature to your project, these items should show up:

``` XML
<ItemGroup>
    <ProjectCapability Include="ShowMissingItemTypes" />
</ItemGroup>
```

Note: This feature hasn't been fully tested yet, so might have bugs.

wjk:
@davkean I wrote an initial implementation of adding this item type as XAML. [You can find it in my branch](https://github.com/wjk/dotnet-project-system/tree/XamlAppDef). It all seems to work just fine, except when I switch the item's type from "None" to "Workflow definition" in the Properties panel the XAML file disappears from the Solution Explorer! I must turn on Show All Files for the file to reappear, and right-clicking > Include in Project do nothing. I have not figured out why this is happening. The only thing I can do to make the XAML file reappear is to change its item type back to None, which of course defeats the whole point. Once we've figured out how to squash this, I can send in a PR. Thanks!

davkean:
@wjk This occurs because you need to include the rule in https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L118

wjk:
[...]

PR #3810, opened 2018-07-31
Add Workflow definition build rule (XamlAppDef)

wjk:
Does mostly what the title says. Some notes:

* I copied the `XamlAppDef.xaml` file from `None.xaml`, because I didn't think that any extra properties would be required. If additional properties needs to be placed in this file, please let me know.
* When I load a project that uses this functionality in the VS Experimental Instance, I still see the issue described in https://github.com/dotnet/project-system/issues/3722#issuecomment-408543998. I think this is because the updated design-time targets are not being referenced properly. Again, please let me know if I've missed anything here.

Thanks!


jp2masa:
> When I load a project that uses this functionality in the VS Experimental Instance, I still see the issue described in https://github.com/dotnet/project-system/issues/3722#issuecomment-408543998.

Did you try setting the `Context` to `File;BrowseObject`?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -171,6 +171,10 @@
       <Context>File</Context>
     </PropertyPageSchema>
     
+    <PropertyPageSchema Include="$(ManagedXamlResourceDirectory)XamlAppDef.xaml">
+      <Context>File</Context>
```
davkean: This needs to be 'File;BrowseObject'.

davkean:
F5 will correctly setup the environment to pick up new rules + targets, so if Experimental Instance isn't working, then the bug hasn't been fixed.

wjk:
Now this is really weird...

I added `BrowseObject` to the XAML file as @jp2masa suggested (in commit 71cb46d), and yet the Experimental Instance still doesn't show the file! I looked inside the Extensions folder for the Exp root-suffix, and none of the design-time targets files show up anywhere! Is this expected, or is something else going wrong?

davkean:
[This](https://github.com/dotnet/project-system/blob/master/src/ProjectSystemSetup/Properties/launchSettings.json#L10) should cause us to pick it up from the recently build artifacts folder. Look in the Output window of the debugger for CPS tracing that might indicate what's going on. I can look latter to see what's going on.

wjk:
I'll do so tomorrow, after I get off work. Thanks!

wjk:
@davkean I've looked at the Output window, and see nothing that looks like CPS tracing. All I see are a lot of rote assembly load/unload log statements. [Here's the full log](https://github.com/dotnet/project-system/files/2246115/vs.log.txt).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/XamlAppDef.xaml
```diff
@@ -0,0 +1,91 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.-->
+<Rule
+  Name="XamlAppDef"
+  DisplayName="General"
+  PageTemplate="generic"
+  Description="File Properties"
```
davkean: This is missing PropertyPagesHidden="true"

davkean:
I'm debugging this to see why this is occuring.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -171,6 +171,10 @@
       <Context>File</Context>
     </PropertyPageSchema>
     
+    <PropertyPageSchema Include="$(ManagedXamlResourceDirectory)XamlAppDef.xaml">
```
davkean: This property is missing an "s"; should be `$(ManagedXamlResourcesDirectory)`.

wjk: Fixed, but it still doesn't seem to work. Please keep looking. Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/XamlAppDef.xaml
```diff
@@ -0,0 +1,91 @@
...
+      Description="The namespace into which the output of the custom tool is placed." />
+
+  <StringProperty
+      Name="Identity"
+      Visible="false"
+      ReadOnly="true"
+      Category="Misc"
+      Description="The item specified in the Include attribute.">
+    <StringProperty.DataSource>
+      <DataSource Persistence="Intrinsic" ItemType="None" PersistedName="Identity" SourceOfDefaultValue="AfterContext" />
```
davkean: Replace every usage of `ItemType="None"` with `ItemType="XamlAppDef"`.

davkean: Even better, delete the `ItemType` from all properties. This is automatically pulled from the DataSource on the rule itself.

davkean: Filed to https://github.com/dotnet/project-system/issues/3815 to basically do this across all the rules so we never run into this again.

davkean: Also filed https://github.com/dotnet/project-system/issues/3816. We have tracing for a bunch of this, but we hook up too late for it to be output into the Output window.

davkean: Strike removing the ItemType, I'm replacing it with XamlAppDef. It's required for the `BrowseObject` but not for the `File` context.

davkean:
With the latest changes I've just pushed, this works. I'll merge it in when we are green.

PR #3818, opened 2018-08-01
Turn on strict mode and opt into VB 15.3

davkean:
- Turn on strict mode (https://github.com/dotnet/roslyn/search?q=FeatureStrictEnabled&unscoped_q=FeatureStrictEnabled)
- Opt into VB 15.3

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -13,12 +13,14 @@
   <PropertyGroup Condition="'$(Language)' == 'C#'">
     <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
     <LangVersion>7.2</LangVersion>
+    <Features>strict</Features>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Language)' == 'VB'">
     <!-- The default is Embed -->
     <VBRuntime>Default</VBRuntime>
     <RemoveIntegerChecks>true</RemoveIntegerChecks>
+    <Features>strict</Features>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


src/Directory.Build.props
```diff
@@ -12,13 +12,18 @@
...
   <PropertyGroup Condition="'$(Language)' == 'C#'">
     <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
+    <!-- Build machines only have Visual Studio 15.5 -->
     <LangVersion>7.2</LangVersion>
+    <Features>strict</Features>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Language)' == 'VB'">
+    <!-- Build machines only have Visual Studio 15.5 (which comes with VB 15.3) -->
+    <LangVersion>15.3</LangVersion>
```
jmarolf: We should be building with a compiler version from nuget not VS. Do 7.3 and 15.5 give errors?

Pilchie: When I looked at it before, we are for the main build, but not for the xaml compiler build in markuppass1 :(

davkean: Sounds like a bug? Is there one on this?

Pilchie: Not sure?  @tmat?

tmat: Could be that `Language` is not set properly in tmp_proj. .NET 4.7.2 changed the name of the temp project, so our detection of the temp proj in the toolset stopped working. I fixed that in recent version of the toolset.

paul1956:
The current version of VS 2017 comes with VB 15.5 but by "default" exposes VB 15.0 unless you opt-in to something else. Why VB 15.3 instead of 15.5? It would be really nice to be able give parameter names to True and False. I have brought this up elsewhere but it seems to apply here as well.

paul1956:
For your repo why not 15_5 because would be nice to use new feature when I add support for comments after _.

PR #3819, opened 2018-08-01
Move xunit/Moq to latest versions

davkean:
This speeds up test invocation by about 35%.

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -94,7 +94,8 @@
     <MicrosoftDotNetProjectTemplatesVersion>1.0.0-beta2-20170629-269</MicrosoftDotNetProjectTemplatesVersion>
 
     <!-- Tests -->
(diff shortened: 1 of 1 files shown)


build/Versions.props
```diff
@@ -94,7 +94,8 @@
     <MicrosoftDotNetProjectTemplatesVersion>1.0.0-beta2-20170629-269</MicrosoftDotNetProjectTemplatesVersion>
 
     <!-- Tests -->
-    <MoqVersion>4.7.99</MoqVersion>
+    <MoqVersion>4.9.0</MoqVersion>
+    <XUnitVersion>2.4.0</XUnitVersion>
```
Pilchie: Before this, `XUnitVersion` was coming from RepoToolset?  Is this early enough that we don't have a mix of versions anywhere?

davkean: They respect it coming from Version.props.

PR #3820, opened 2018-08-01
Moq base

davkean:
This includes https://github.com/dotnet/project-system/pull/3819, so review from https://github.com/dotnet/project-system/commit/4805702534ff178be4a217bc75d820dfeca7acce onwards.

This is an experiment of moving to from a factory-based based mocking model, to an object based model that lets you easily mock multiple methods for a given test. I moved IActiveDebugFrameworkServicesFactory which was already in a similar style it.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/Moq/AbstractMock.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/Moq/AbstractMock.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Moq
+{
+    internal abstract class AbstractMock<T> : Mock<T>
+        where T : class
+    {
+        protected AbstractMock(MockBehavior behavior = MockBehavior.Loose)
+            : base(behavior)
+        {
+            Switches = Switches.CollectDiagnosticFileInfoForSetups;
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IActiveDebugFrameworkServicesMock.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IActiveDebugFrameworkServicesMock.cs
@@ -29,17 +29,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
             return this;
         }
 
-        public IActiveDebugFrameworkServicesMock ImplementGetProjectFrameworksAsync(List<string> frameworsks)
+        public IActiveDebugFrameworkServicesMock ImplementGetProjectFrameworksAsync(List<string> frameworks)
         {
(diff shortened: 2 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/Moq/AbstractMock.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Moq
+{
+    internal abstract class AbstractMock<T> : Mock<T>
+        where T : class
+    {
+        protected AbstractMock(MockBehavior behavior = MockBehavior.Loose)
+            : base(behavior)
+        {
+            Switches = Switches.CollectDiagnosticFileInfoForSetups;
```
davkean: "Friendly" stack traces.

jmarolf:
I like this new pattern.  Can we replace all our `Factory` classes to follow this pattern?

src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/Moq/AbstractMock.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Moq
+{
+    internal abstract class AbstractMock<T> : Mock<T>
+        where T : class
+    {
+        protected AbstractMock(MockBehavior behavior = MockBehavior.Loose)
```
sharwell: :bulb: Suggest defaulting to `MockBehavior.Strict`. It differs from the Moq default but provides a big advantage in test confidence. Moq defaults tend to violate post-conditions which is not a happy place.

davkean: We started with strict originally in the project, and while it resulted a little more confidence - it's turned into convenience vs correctness. I found that it actively caused me to write less tests - because it was harder to get started.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IActiveDebugFrameworkServicesMock.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+using Moq;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    internal class IActiveDebugFrameworkServicesMock : AbstractMock<IActiveDebugFrameworkServices>
```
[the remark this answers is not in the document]
davkean: We stub out like this due to readability of tests, this is typically always our approach vs using Moq directly in tests. 

Your experience is clearly different than me, but the convenience of Moq having stub out the fake than doing it yourself in my mind outweighs the debuggability. I've tried both approaches in different projects, and ended up favoring this one - and what we use around the tree. If Moq auto-generated C# source code to get best of both worlds, I'd use it in a heartbeat.

PR #3827, opened 2018-08-03
handle  version number calculation

jmarolf:
As part of this pr (https://github.com/dotnet/project-system/pull/3706) we attempted to pad the version numbers so 15.8 version number would line up correctly.

Unfortunately this logic was wrong as BUILD_BUILDNUMBER can come in one of the two formats:
- 20180102.03
- 20180102.3

This change accounts for the shorter format (10 characters).

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -7,9 +7,10 @@
     <!-- 
       15.8 specific workaround for a change to BUILD_BUILDNUMBER format that added two more digits. 
       Any subsequent versions must have the same amount of digits to be considered higher.
-      Updates the version 20180102.03 to 20180102.0003.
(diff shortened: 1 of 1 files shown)


jmarolf:
We may need to do the same thing as @davkean did here: https://github.com/dotnet/project-system/pull/3708 so that this doesn't accidentally get merged into 15.9 or master

tmat:
@Pilchie By default the format of the build number is `yyyymmdd.r`, which results in non-padded revision number (1,2,3,...,10,11,...). So the build number can be 10 or 11 characters long.

When adding this workaround I have set the build definition to `yyyymmdd.rr`, which is padded to 2 characters. We lost that setting by converting to YAML definition, so it's back to the default.

Issue #3646, opened 2018-06-12
RoslynInsertionTool needs to update binding redirects for Project System assemblies


tmat:
VSO PR [127593](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/127593) adds 

```
const string MicrosoftVisualStudioProjectSystemVersion = "2.8.0.0";
```

to `/src/ProductData/AssemblyVersions.tt` file. This version needs to be automatically updated on Project System component insertion.

It would also be good to auto-update package versions in `default.config` to avoid having old binaries in External APIs:

```xml
   <package id="Microsoft.VisualStudio.ProjectSystem.Managed" version="2.6.0-beta3-62503-02" tags="exapis" link="src\ExternalAPIs\Microsoft.VisualStudio.ProjectSystem.Managed" />
    <package id="Microsoft.VisualStudio.ProjectSystem.Managed.VS" version="2.6.0-beta3-62503-02" tags="exapis" link="src\ExternalAPIs\Microsoft.VisualStudio.ProjectSystem.Managed.VS" />
```

jmarolf:
fixed here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/138417

PR #3831, opened 2018-08-07
generate dependent assembly versions for insertion tool

jmarolf:
Initial work needed to get https://github.com/dotnet/project-system/issues/3646 fixed.


jmarolf:
@tmat can you give an example of doing that?

Issue #2122, opened 2017-05-04
Incorrect font used for the Debug page on project properties


srivatsn:
Reported through VSFeedback by @pharring 

The "Debug" page in project properties for a .Net Core Console application appears to be using a different font from the rest of the Visual Studio UI. It appears to be using "Microsoft Sans Serif, 8.25" whereas the rest of the UI is Segoe UI 8.

ghuntley:
Confirmed in `15.7.0 Preview 2.0` using netstandard project.

* [ ] Obtain font preferences from ????
* [ ] Update https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WatermarkStyle.xaml
* [ ] Update https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageControl.xaml

PR #3837, opened 2018-08-09
Fixed: Incorrect font used for the Debug page on project properties

nam178:
Fixed issue: https://github.com/dotnet/project-system/issues/2122

Adding a WPF control to ElementHost causes its default font to be removed. I think this is an expected behaviour to make WPF control looks the same to Windows form, however, we don't want that in this case.

This fix captures the font before hooking up the property page control with ElementHost, and restores it after.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.cs
@@ -85,8 +85,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                 HorizontalScrollBarVisibility = ScrollBarVisibility.Auto
             };
 
+            // Capture original WPF font
+            var originalFontFamily = viewer.FontFamily;
+            var originalFontSize = viewer.FontSize;
+
             viewer.Content = _control;
             _host.Child = viewer;
+
+            // Parenting viewer to ElementHost removes its default font,
+            // So we restore the font here.
+            viewer.FontFamily = originalFontFamily;
+            viewer.FontSize = originalFontSize;
 
             wpfHostPanel.Dock = DockStyle.Fill;
             wpfHostPanel.Controls.Add(_host);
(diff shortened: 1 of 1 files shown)


Pilchie:
Adding @dpoeschl too review this too.

@nam178 can you include a before/after example screenshot?

nam178:
Sure @Pilchie, @dpoeschl 

Screenshots attached.
after
before

Pilchie:
Thanks - I'm following up with some people internally to verify that this is the right approach, but thanks for the contribution.

(The other option I've heard of is setting `ElementHost.Font`, which should be propagated down when `Child` is set.

davkean:
Not sure about WPF, but this change would be incorrect for WinForms. Explicitly setting the Font would prevent it from being inherited from the parent, which would break when we change the parent due to style changes. @Pilchie Try changing icon font in Windows which is what we respect if I remember correctly. I would expect the ElementHost to have the correct Font.

nam178:
Thanks @davkean, so in the change I set the font for the WPF view instead of ElementHost, what's your thought on that?

I'll test switching font.

Pilchie:
Just a quick note that some of the relevant parties are on vacation, so it will be a couple of weeks before we get back to this.

Pilchie:
@nam178 Yes, I *think* the correct thing to do would be to set the font in the WPF view to the VS resource for the font that should be used, so that it can pick up theme changes.

@davkean does that sound right to you?

davkean:
Agreed, I don't believe the this is the correct change. These pages are supposed to be inheriting the Font from VS, the fact that the font _looks_ correct is solely because Segue UI is the default for WPF. WPF's pages hosted in ElementHost are supposed to be inheriting their parent (ie AppDesigner panel) 's font: https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/windows-forms-and-wpf-property-mapping, the fix would be to figure out why that isn't working.

davkean:
I spent a little bit of time on this, and this is what I believe to be the correct change: https://github.com/dotnet/project-system/pull/4124. This will cause the page to respect any future changes to the Font while it is open.

davkean:
@nam178 Really appreciate the attempt to fix this - this one was tougher than we thought due to the WinForms/WPF interaction, I got lucky because I'd run into this particular issue before. 

I'm going to close this one out, in lieu of the one I pointed to above.

PR #3850, opened 2018-08-13
[Language Service] Add IImplicitlyActiveDimensionProvider

davkean:
Returns a set of implicitly active dimensions from specified set of dimensions, which can be used for disambiguation purposes. For example, in multi-targeting projects, the language service will use it in the context switch to distinguish in the following cases:

ClassLibrary (net45)
ClassLibrary (net46)

Consumption comes in a later PR.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/ImplicitlyActiveDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/ImplicitlyActiveDimensionProvider.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Linq;
+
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/ImplicitlyActiveDimensionProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var implicitlyActiveDimensions = ImmutableArray.CreateBuilder<string>();
+
+            foreach (Lazy<IProjectConfigurationDimensionsProvider, IConfigurationDimensionDescriptionMetadataView> provider in DimensionProviders)
+            {
+                for (int i = 0; i < provider.Metadata.IsVariantDimension.Length; i++)
+                {
+                    if (provider.Metadata.IsVariantDimension[i])
+                    {
+                        if (!implicitlyActiveDimensions.Contains(provider.Metadata.DimensionName[i], StringComparers.ConfigurationDimensionNames))
```
Pilchie: How many of these are there likely to be?  Should we build this with `HashSet` or something instead?

davkean: Tiny amount 3 currently, maybe 4 if we add a dimension for RID. I've used ImmutableArray also to guarantee order.

PR #3856, opened 2018-08-14
set the archive trigger after the test results trigger

jmarolf:
Without this change, a test failure will not cause an archive to happen


jmarolf:
@dotnet-bot test ci please

PR #3857, opened 2018-08-14
[Language Service] Add ProjectFilePathAndDisplayNameEvaluationHandler

davkean:
Handles changing the file name and display name in reaction to a project file being renamed. This unifies all logic of setting touching IWorkspaceProjectContext in handlers, which lets us coordinate changes and prevent overlapping.


davkean:
The old hookup sets this in the following places:

https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/AggregateWorkspaceProjectContext.cs#L66
https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs#L256

It also makes the assumption that the "implicitly active" dimension is always TargetFramework, which is something we're trying to move away from.

PR #3864, opened 2018-08-15
Simplify dataflow usage

davkean:
Add new overloads to LinkToAsyncAction/LinkToAction that simplifies signing up to dataflow blocks.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
@@ -66,10 +66,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
 
             Assumes.Present(_startupProjectsListService);
 
-            _subscription = _projectSubscriptionService.ProjectRuleSource.SourceBlock.LinkTo(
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -66,10 +65,8 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
 
             Assumes.Present(_startupProjectsListService);
 
-            _subscription = _projectSubscriptionService.ProjectRuleSource.SourceBlock.LinkTo(
-                target: new ActionBlock<IProjectVersionedValue<IProjectSubscriptionUpdate>>(OnProjectChangedAsync),
-                suppressVersionOnlyUpdates: true,
```
abpiskunov: Just checking, why these two lines are not needed anymore?

davkean: supressVersionOnlyUpdates and DataflowOption.PropagateCompletion are the defaults here.

abpiskunov: ah ok :)

PR #3870, opened 2018-08-15
Add basic overview of the Dependencies Node code

tmeschter:
Adds a high-level overview of how we represent dependencies internally, how data flows through the system, and how we translate it to CPS constructs.


docs/dependencies-node-roadmap.md
```diff
@@ -0,0 +1,104 @@
...
+
+> Aside: While there is currently only one implementation of [`IDependenciesTreeViewProvider`][3] you could imagine others that organize the nodes in different ways: flattening the top-level items into a single list or organizing nodes first by type then by target framework. Should we create other implementations we would also need some way to switch between them; there is currently no support for doing so.
+
+The [`GroupedByTargetTreeViewProvider`][8] traverses down the existing `IProjectTree` and the new [`IDependenciesSnapshot`][6] in parallel starting from the Dependencies node itself and proceeding on to target framework, groupings, and then the individual top-level dependencies. Along the way it incrementally generates new `IProjectTree`s as it finds dependencies that have been updated, added, or removed.
```
abpiskunov: One important thing to note here is that since we are updating tree nodes form top to bottom it prevents Solution explorer to flick (close already opened nodes during those updates).

abpiskunov:
Looks good so far for sections that do have description.

davkean:
We put docs related to the repo itself under docs/repo. Docs directly under docs are for consumption by 3rd parties. I'm okay with changing that but we should be consistent.

docs/dependencies-node-roadmap.md
```diff
@@ -0,0 +1,104 @@
+# Dependencies Node Roadmap
+
+This document describes, at a high level, the design and implementation of the Dependencies node with pointers to the important types.
+
+Let's start at the top and work our way down.
+
+## The CPS View of the Project Tree
+
```
davkean: I know you call this out below, but it would be nice to call here that dependency node is actually CPS + GraphNodes.

PR #3871, opened 2018-08-15
Add a Gitter chat badge to README.md

gitter-badger:
### dotnet/project-system now has a Chat Room on Gitter

@panopticoncentral has just created a chat room. You can visit it here: [https://gitter.im/dotnet/project-system](https://gitter.im/dotnet/project-system?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&content=body_link).

This pull-request adds this badge to your README.md:

[Gitter](https://gitter.im/dotnet/project-system?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=body_badge)

If my aim is a little off, please [let me know](https://github.com/gitterHQ/readme-badger/issues).

Happy chatting.

PS: [Click here](https://gitter.im/settings/badger/opt-out) if you would prefer not to receive automatic pull-requests from Gitter in future.

PR #3872, opened 2018-08-16
Add Gitter link to README

etbyrd:
This is the correct markdown of https://github.com/dotnet/project-system/pull/3871


etbyrd:
To be clear: I am not arguing the merits of adding it, just wanted a PR that was correct.

etbyrd:
@Pilchie Should I go ahead and remove the 15.7.x branch?

Pilchie:
Sure, though we should also make sure we have a tag for the bits that shipped in 15.7 (and 15.8) if we haven't yet.  I think @davkean asked @dotnet/roslyn-infrastructure about that the other day.

davkean:
15.7 branch is already gone.

PR #3873, opened 2018-08-16
Simplify VsContainedLanguageComponentsFactory logic

davkean:
- Only look up IVsContainedLanguageFactory/LanguageServiceId once
- Remove base class, only need to just wait on InitializeAsync from the language service
- Move to CPS pattern of sync method calling async method

Tests forecoming as it doesn't have any. Confirmed that Razor plays nicely with these changes.


davkean:
Added tests.

PR #3875, opened 2018-08-16
adding copyright

jmarolf:
testing integration test queue now that it is merged


davkean:
OMG Jon, they are passing!

PR #3877, opened 2018-08-17
Introduce async IVsService<T>/IVsService<TService, TInterface>

davkean:
Previously IVsService<T>/IVsService<TService, TInterface> was synchronous and UI thread bound. This usage has been renamed to IVsUIService<T>/IVsUIService<TService, TInterface>, and IVsService has been turned into a free-threaded API.

Simplify the AsyncLazy<T> pattern of async service lookup seen throughout CPS and elsewhere into a simple MEF import:

``` C#
    public Part(IVsService<SService, IService> service)
```
...
``` C#
    var service = await service.GetValueAsync();
    service.FreeThreadedMethod();
```

Services can be looked with clsid/service type:

``` C#
    private readonly IVsSolution<IVsSolution> _solution;  // IVsService<TService, TInterface> derives from IVsService<T>

    public Part(IVsService<SVsSolution, IVsSolution> solution)
    {
	    _solution = solution;
    }
```

or just based on their interface:

``` C# 
    private readonly IVsService<DTE> _dte; 

    public Part(IVsService<DTE> dte)
    {
	    _dte = dte;
    }
```

IVsOptionalService has been removed. To have both a split based on optional'ness and async'ness would be an explosion of types. At this point, we now no longer enforce that a service exists and instead return null. This works okay in the situations that we're currently using it, but may enforce this via overloads/additional methods in the future if needed.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsService`1.cs
```diff
@@ -2,50 +2,58 @@
...
+        {
+            get { return typeof(T); }
+        }
+
+        private async Task<T> GetServiceAsync()
+        {
+            // If the service request requires a package load, GetServiceAsync will 
+            // happily do that on a background thread.
+            object iunknown = await _serviceProvider.GetServiceAsync(ServiceType)
+                                                    .ConfigureAwait(true);
```
jmarolf: If we switch to the main thread no matter what why do we need to always return to the calling thread?  

Also, this means that all callers of `GetValueAsync` must know to `ConfigureAwait(true)` can you add a comment about that? I would prefer it if we returned a custom awaiter here so that the caller is always returned to their original thread regardless.

jmarolf:
After looking at how these threading rules need to be applied I don't like this approach.  Can we add something such that callers of this API do not need to know the threading semantics at all?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsService`1.cs
davkean: Read this, https://github.com/Microsoft/vs-threading/blob/master/doc/cookbook_vs.md#should-i-await-a-task-with-configureawaitfalse. 

`ConfigureAwait(false)` == always switch to thread-pool thread. If we use `ConfigureAwait(false)`, and we ended up on UI thread in GetServiceAsync, we'll likely be forced on a thread pool thread, only to immediately switch back to the UI thread.

Look through history of teams, I want to change the project system's usage of ConfigureAwait to be always `ConfigureAwait(true)`.

davkean:
@jmarolf I'm going to merge this into features/languageservices as I have a dependency on it, we can continue to discuss before we merge into master.

jmarolf:
@davkean no problem.  Can you file a tracking bug to rationalize our usage of `ConfigureAwait` (or lack therof).

davkean:
@jmarolf Yeah I will file a bug on fixing our usage.

PR #3878, opened 2018-08-17
Use IVsService in StartupProjectRegister

davkean:
This includes https://github.com/dotnet/project-system/pull/3877, so just review https://github.com/dotnet/project-system/commit/4335a070ffc9d6009c44a68b9a4f79200fb57471 onwards.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -87,15 +80,20 @@ internal async Task OnProjectChangedAsync(IProjectVersionedValue<IProjectSubscri
             bool isDebuggable = await _launchProviders.Value.IsDebuggableAsync()
                                                             .ConfigureAwait(false);
 
+            IVsStartupProjectsListService startupProjectsListService = await _startupProjectsListService.GetValueAsync()
+                                                                                                        .ConfigureAwait(true);
```
jmarolf: Why does this need to return to the UI thread?

davkean: Only the cast runs on the UI thread, package init, service creation will occur on a background thread if possible.

PR #3883, opened 2018-08-20
Turn off IDE0001 in test code due to regression

davkean:
'Var' usage in test code is marked as "refactoring-only" but due to https://github.com/dotnet/roslyn/issues/27819, all explicit type usage is marked as an error.

### DIFF
--- a/build/Test.ruleset
+++ b/build/Test.ruleset
@@ -19,7 +19,10 @@
     <Rule Id="RS0023" Action="None" />
   </Rules>
    <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
-     <!-- Naming styles is too noisy as it fires on all async tests -->
+    <!-- Too noisy due to https://github.com/dotnet/roslyn/issues/27819 -->
+    <Rule Id="IDE0001" Action="None" />                       <!-- Simplify names                               System.Version version;                             Version version;                      -->
+
+     <!-- Naming styles is too noisy as it fires on all async tests -->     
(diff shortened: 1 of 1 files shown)


build/Test.ruleset
```diff
@@ -19,7 +19,10 @@
     <Rule Id="RS0023" Action="None" />
   </Rules>
    <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
-     <!-- Naming styles is too noisy as it fires on all async tests -->
+    <!-- Too noisy due to https://github.com/dotnet/roslyn/issues/27819 -->
+    <Rule Id="IDE0001" Action="None" />                       <!-- Simplify names                               System.Version version;                             Version version;                      -->
```
sharwell: :bulb: this should be changed to `Hidden` or simply removed from this file to align with the behavior of the corrected diagnostic. Once the NotConfigurable flag takes effect, this change will be applied automatically regardless of the value placed in a rule set file.

davkean: What's the equivalent?

sharwell: > What's the equivalent?

I'm not sure I understand this question

davkean: >  Once the NotConfigurable flag takes effect, this change will be applied automatically regardless of the value placed in a rule set file.

In 15.7 I could enforce IDE0001 usage. What is the equivalent?

sharwell: There is no single diagnostic which is equivalent. Several other diagnostics allow users to enforce specific subsets (e.g. the use var and qualify member diagnostics), and we will likely continue to expand this.

PR #3884, opened 2018-08-20
Get Visual Studio code coverage working

davkean:
This makes Code Coverage work in Visual Studio again, also updated code coverage settings to match CodeCov.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -10,6 +10,13 @@
   <Import Project="$(RepoToolsetDir)Settings.props" Condition="'$(RepoToolsetDir)' != ''" />
   <Import Project="..\build\VisualStudio.XamlRules.props" />
 
+  <PropertyGroup>
+    <_IsVisualStudioDeveloperBuild Condition="'$(OfficialBuild)' != 'true' AND $(CIBuild) != 'true' AND '$(BuildingInsideVisualStudio)' == 'true'">true</_IsVisualStudioDeveloperBuild>
+
+    <!-- Code Coverage doesn't currently work for portable/embedded PDBs - force to full inside Visual Studio builds -->
+    <DebugType Condition="'$(_IsVisualStudioDeveloperBuild)' == 'true'">full</DebugType>
+  </PropertyGroup>
+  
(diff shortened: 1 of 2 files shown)


src/Directory.Build.props
```diff
@@ -10,6 +10,13 @@
   <Import Project="$(RepoToolsetDir)Settings.props" Condition="'$(RepoToolsetDir)' != ''" />
   <Import Project="..\build\VisualStudio.XamlRules.props" />
 
+  <PropertyGroup>
+    <_IsVisualStudioDeveloperBuild Condition="'$(OfficialBuild)' != 'true' AND $(CIBuild) != 'true' AND '$(BuildingInsideVisualStudio)' == 'true'">true</_IsVisualStudioDeveloperBuild>
```
sharwell: :grey_question: was `$(CIBuild)` supposed to be in single quotes?

jmarolf:
@davkean change the branch to master so we'd run tests against this (thanks for using `git merge-base`).

As an aside, do we want to turn on ci for the language features branch [here](https://github.com/dotnet/dotnet-ci/blob/a7709bfdcc3b9bf463fd8bad550627d8b502e4b0/data/repolist.txt#L165-L169)?

jmarolf:
@davkean this won't work because Apex only support MSTest at the moment and repotoolset only supports running xUnit tests.  I'll submit an update to you branch so this works.

jmarolf:
@davkean the other branch you compared against in `git merge-base` does not have support for integration tests.  you need to base this branch off of the latest master to be able to add support.

davkean:
I think you are thinking this is something different than what is it. This lets you run unit tests inside Visual Studio and see Code Coverage results in the UI itself. This isn't enabling coverage for integration tests. I need it in language services as I'm writing a ton of tests.

jmarolf:
Whoops, sorry for messing with your PR @davkean

jmarolf:
My Comment about CI still stands though, do you want me to turn it on for the language service branch?

davkean:
If you'd like - I've been okay so far without it, as I merge into master every couple of days but if you'd like that confidence I'll set one up.

PR #3895, opened 2018-08-22
Centralize telemetry events

davkean:
We have complicated logic inside VsTelemetryService to avoid allocating event and property names, centralize all names into a static class.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -24,7 +24,6 @@ internal sealed class BuildUpToDateCheck : OnceInitializedOnceDisposed, IBuildUp
         private const string CopyToOutputDirectory = "CopyToOutputDirectory";
```
Pilchie: Not used below?

davkean: It's used to look up metadata.

PR #3901, opened 2018-08-23
Rename mock factory IWorkspaceProjectContextFactory

davkean:
It clashes with the actual real "IWorkspaceProjectContextFactory" from Roslyn.


davkean:
Ignore the status checks on this - these aren't supposed to be running on this branch.

PR #1002, opened 2016-12-20
Turn on code style rules by default

davkean:
This turns on what I believe to be non-controversial code style rules on by default. Others which I found to be noisy, such as expression body and object initialization I've left turned off.

This a bit of an experiment over Christmas to see if it's too annoying, at which we can turn them from Error -> Info.

Note, I'll going to do one last pass to make sure we've fixed all current violations of these rules.

PR #3904, opened 2018-08-23
Remove configuration of IDE0001 severity

sharwell:
This diagnostic is not covered by a stable versioning policy (i.e. the behavior of the diagnostic is subject to change between VS releases). It is not intended for use at any severity other than Hidden (the default) or Suggestion. This commit removes the use of a rule set file to set this to Error, since it can (and in some cases does) result in unspecified diagnostic churn between VS updates.

### DIFF
--- a/build/Default.ruleset
+++ b/build/Default.ruleset
@@ -122,7 +122,6 @@
   </Rules>
   <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
                                                               <!-- Name:                                        Before:                                             After:                                -->
-    <Rule Id="IDE0001" Action="Error" />                      <!-- Simplify names                               System.Version version;                             Version version;                      -->
     <Rule Id="IDE0002" Action="Error" />                      <!-- Simplify (member access)                     System.Version.Equals("1", "2");                    Version.Equals("1", "2");             -->
     <Rule Id="IDE0003" Action="Error" />                      <!-- Remove qualification                         this.X = x;                                         X = x;                                -->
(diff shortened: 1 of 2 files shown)


davkean:
Can you point me to documentation that states this "a stable versioning policy (i.e. the behavior of the diagnostic is subject to change between VS releases)"?

davkean:
I cannot find any documentation that calls out IDE0001 as being special. @mavasani signed off on the [original change](https://github.com/dotnet/project-system/pull/1002) with zero concerns, so this policy is a take back from the IDE.

sharwell:
> Can you point me to documentation that states this:

The lack of a clear definition for what IDE0001 is enforcing is the reason for this.

The behavior of IDE0001 is stable in the context of a particular release, but as you've seen, it can change between releases resulting in inconsistent enforcement for users who increase the severity. This was not rigorously evaluated in the past because IDE diagnostics do not currently impact builds, but it would be one reason why Simplify Type Name (the current analyzer producing IDE0001) would likely not be migrated to the NuGet package for build-time enforcement.

davkean:
As a consumer of this, how do I determine the difference between rules that are stable and rules that are not stable?

tmeschter:
So even if it isn't stable and we can only use it within the IDE it still has value while editing code. Turning it off now because there _might_ be problems down the road seems premature--we can always wait until it's actually causing problems and disable it then.

davkean:
We're not going to merge this, so going to close it out.

sharwell:
> As a consumer of this, how do I determine the difference between rules that are stable and rules that are not stable?

The only rules that are truly stable are version-locked rules included via NuGet packages. VSIX-installed analyzers would always be expected to have some non-zero tolerable level of churn and/or discrepancies between different machines at any given point.

PR #3905, opened 2018-08-24
Add WorkspaceContextProvider

davkean:
Responsible for creating/disposing a IWorkspaceProjectContext.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/LazyExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/LazyExtensions.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft
+{
+    public static class LazyExtensions
+    {
+        public static Lazy<T> AsLazy<T>(this T instance)
+        {
+            return new Lazy<T>(() => instance);
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IWorkspaceProjectContextFactoryFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IWorkspaceProjectContextFactoryFactory.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using Moq;
+
+namespace Microsoft.VisualStudio.LanguageServices.ProjectSystem
+{
+    internal static class IWorkspaceProjectContextFactoryFactory
+    {
+        public static IWorkspaceProjectContextFactory Create()
+        {
+            return Mock.Of<IWorkspaceProjectContextFactory>();
+        }
+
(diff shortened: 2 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/LazyExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft
+{
+    public static class LazyExtensions
+    {
+        public static Lazy<T> AsLazy<T>(this T instance)
```
Pilchie: Wow, shows up everywhere in our tests without even requiring a using.  Bold.

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IWorkspaceProjectContextFactoryFactory.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using Moq;
+
+namespace Microsoft.VisualStudio.LanguageServices.ProjectSystem
+{
+    internal static class IWorkspaceProjectContextFactoryFactory
```
Pilchie: Have I mentioned that I hate `FactoryFactory` yet?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.ForegroundWorkspaceProjectContext.cs
```diff
@@ -0,0 +1,136 @@
...
+
+                _threadingService = threadingService;
+                _underlyingContext = underlyingContext;
+            }
+
+            public IWorkspaceProjectContext UnderlyingContext
+            {
+                get
+                {
+                    _threadingService.VerifyOnUIThread();
```
jmarolf: Why not just switch to the UI thread and return a task?

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IWorkspaceProjectContextFactoryFactory.cs
jmarolf: We're officially enterprise-grade now :)

---
In reply to: [212745005](https://github.com/dotnet/project-system/pull/3905#discussion_r212745005) [](ancestors = 212745005)

davkean: What would you call this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.ForegroundWorkspaceProjectContext.cs
davkean: 1) I'm implementing an existing interface that I have no control over.
2) I very deliberately coordinate changes to the context - I don't want random consumers switching to the UI and applying things out of order. Once Jason's work is in, this will be free-threaded and we won't need to enforce this.

PR #3913, opened 2018-08-28
Never report up-to-date if CopyAlways item exists

### LINKED ISSUE #3568: Updated file with "copy always" does not get copied
I have a configuration file in my .NET Core 2.0 console app's root directory. It is set to &quot;Build Action: None&quot; and &quot;Copy Always&quot;. The idea is that it will end up in the bin directory.

On first build, all is well.

Then I modify the config file and press Control+Shift+B to build the solution.

I then observe that the file in the bin directory remains out of date, despite it having changed and Copy Always being set.

Build output says that it skipped all my projects. Possibly the cause of the issue? It does not seem to consider this file changing as something that requires a build.

Behavior is same with &quot;Build Action: Content&quot;

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/252209/updated-file-with-copy-always-does-not-get-copied.html
VSTS ticketId: 615693_
(no comments)
(no solutions)

drewnoakes:
Fixes #3568 

Prior code logged the fact that the project was not up to date, but didn't signal that information back to the calling method. If any items are `CopyAlways` then the up-to-date check can be short circuited.

Questions for reviewers:

1. Code in checks immediately above this change use the `Fail` method to capture a telemetry event. Should this also happen when we detect a `CopyAlways` item?
1. Unit tests... I looked into adding tests for this class. It's quite a meaty one with substantial dependencies, and I'm new to the code. Please advise!

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -349,8 +349,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             (string Path, string Link, CopyToOutputDirectoryType CopyType) copyAlwaysItem = _items.SelectMany(kvp => kvp.Value).FirstOrDefault(item => item.CopyType == CopyToOutputDirectoryType.CopyAlways);
             if (copyAlwaysItem.Path != null)
             {
-                logger.Info("Item '{0}' has CopyToOutputDirectory set to 'Always', not up to date.", copyAlwaysItem.Path);
-                return false;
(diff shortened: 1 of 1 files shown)


tmeschter:
@panopticoncentral I would think we would want to use the `Fail` method here as well. Any reason not to?

tmeschter:
@drewnoakes I think it's worth your time to create a unit test for your particular scenario. That would set us up for adding further unit tests for this type in the future, as we have need.
Now, if you come back and say "there's no reasonably way to unit test this" then I'd probably accept that answer, too. ;-)

drewnoakes:
> I would think we would want to use the Fail method here as well. Any reason not to?

See 557302b8b, which adds telemetry reason `CopyAlwaysItemExists`.

This will also always interpolate the logged string, even if the log level is too low. How sensitive are we to such allocations?

> I think it's worth your time to create a unit test for your particular scenario.

👍 I'll take a look at this tomorrow.

drewnoakes:
Merging this to base another PR off it related to #3918.

PR #3914, opened 2018-08-28
Propagate cancellation via IProjectSystemOptions

drewnoakes:
The implementations don't do much, but in case of cancellation it may avoid an operation on the main thread.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
@@ -49,12 +49,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         public async Task<bool> GetIsFastUpToDateCheckEnabledAsync(CancellationToken cancellationToken = default)
         {
             await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+            cancellationToken.ThrowIfCancellationRequested();
+
             return _settingsManager.Value?.GetValueOrDefault(FastUpToDateEnabledSettingKey, true) ?? true;
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
```diff
@@ -45,15 +46,15 @@ public bool IsProjectOutputPaneEnabled
             }
         }
 
-        public async Task<bool> GetIsFastUpToDateCheckEnabledAsync()
+        public async Task<bool> GetIsFastUpToDateCheckEnabledAsync(CancellationToken cancellationToken = default)
         {
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
```
davkean: Confusingly it appears this doesn't do what we would expect. According to docs and the code (http://index/?leftProject=Microsoft.VisualStudio.Threading&leftSymbol=hyuxmvwxl8sv&file=JoinableTaskFactory.cs&line=708):

> A token whose cancellation will immediately schedule the continuation on a threadpool thread.

This means that instead of throwing `OperationCanceledException`, it just executes the continuation (ie `_settingsManager.GetValueOrDefault`) on the thread-pool. I think you want to follow up after the switch with `cancellationToken.ThrowIfCancellationRequested()`.

drewnoakes: Thank for catching that. I've pushed a fix.

I'm not sure this PR adds much, so am happy to close it out.

davkean: I'm okay with this - it's good practice to propagate cancellation tokens and gets you used to how we handle PRs and such.

PR #3916, opened 2018-08-28
adding NUnit templates

jmarolf:
adding templates for NUnit test framework

fixes https://github.com/Microsoft/vstest/issues/1570


davkean:
👍

PR #3920, opened 2018-08-30
Remove async'ness of ICrossTargetSubscriber

davkean:
I need to consume the new Microsoft.VisualStudio.Threading package in the language services, and the new analyzers are flagging certain usage within the dependency node because it's not awaiting results in certain places.

Looking at this usage, this entire infrastructure has no async code whatsoever and other code is not used at all, just remove all async from these code paths and removed unused methods.

Fixes two violations in https://github.com/dotnet/project-system/issues/3921.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/ICrossTargetRuleHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/ICrossTargetRuleHandler.cs
@@ -43,14 +43,5 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
                          ITargetedProjectContext targetedProjectContext,
                          bool isActiveContext,
                          T ruleChangeContext);
-
-        /// <summary>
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/ICrossTargetRuleHandler.cs
```diff
@@ -43,14 +43,5 @@ internal interface ICrossTargetRuleHandler<T> where T : IRuleChangeContext
...
                          T ruleChangeContext);
-
-        /// <summary>
-        /// Handles clearing any state specific to the given context being released.
-        /// </summary>
-        /// <param name="context">
-        ///     A <see cref="ITargetedProjectContext"/> being released.
-        /// </param>
-        Task OnContextReleasedAsync(ITargetedProjectContext context);
-
```
tmeschter: Were there any implementations that weren't effectively no-ops?

davkean: No, other than delegation on the the "nested" ones.

PR #3923, opened 2018-08-30
Remove workaround in DesignTimeAssemblyResolution

davkean:
The bug the workaround was referrring to was fixed in 15.3 and verified that the part no longer gets rejected.


davkean:

PR #3908, opened 2018-08-27
Fire up/tear down workspace context based on implicitly active

davkean:
Fire up and tear down IWorkspaceProjectContext based on being implicitly active/deactive. This lets us react to active configuration/TFM changes.

~- Tests coming~
- Hidden behind a project capability
- IsActiveContext will be implemented in upcoming PR

This includes https://github.com/dotnet/project-system/pull/3948, so only review https://github.com/dotnet/project-system/pull/3908/commits/174ecb0f0ccb45361ef96eed99ab57f0b3673fdc onwards.

PR #3926, opened 2018-08-30
Turn off "CA2213: Disposable fields should be disposed" due to noise

davkean:
Per https://github.com/dotnet/project-system/pull/3908#discussion_r212955729, this rule is mostly noise for project-system because we derive from OnceInitializedOnceDisposed/OnceInitializedOnceDisposedAsync.

There's other usages of `#pragma warning disable` that I left because they aren't actually disposing the field that has been suppressed, such as https://github.com/dotnet/project-system/blob/dfe534a3e7803efb763ba9dcd35c5ce60c9104a5/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs#L20. 

Note, that even after we dispose these usages, the rule will still fire - hence why we're turning it off. I left these suppressions so that we can find them, and using https://github.com/dotnet/project-system/issues/3927 to track fixing them.

### DIFF
--- a/build/Default.ruleset
+++ b/build/Default.ruleset
@@ -48,6 +48,7 @@
     <Rule Id="CA1815" Action="None" />
     <Rule Id="CA2007" Action="Warning" />
     <Rule Id="CA2211" Action="None" />
+    <Rule Id="CA2213" Action="None" /> <!-- https://github.com/dotnet/roslyn-analyzers/issues/1796 -->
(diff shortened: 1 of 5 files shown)


jmarolf:
@davkean can we give this feedback to the analyzer team about `CA2213`? 

CC: @mavasani

PR #3936, opened 2018-08-31
Add graph nodes documentation

tmeschter:
Add documentation on the role of graph nodes in the Dependencies node implementation.

### DIFF
--- a/docs/repo/dependencies-node-roadmap.md
+++ b/docs/repo/dependencies-node-roadmap.md
@@ -99,7 +99,7 @@ New `GraphNode`s are added to the graph as a result of operations initiated by t
 
 ### Connecting `GraphNode`s to `IProjectTree` nodes
 
-`GraphNode`s are automatically created for `IProjectTree` nodes by CPS and the core graph logic. For example, we will create an `IProjectTree` for a top-level NuGet package and an associated `GraphNode` will be generated automatically. We will later see this `GraphNode` as an input node in an `IGraphContext` and create and link new `GraphNode`s for its transitive dependencies.
+CPS and the core graph logic automatically create `GraphNode`s for `IProjectTree` nodes marked with the `ProjectTreeFlags.Common.ResolvedReference` flag. For example, we will create an `IProjectTree` for a top-level NuGet package and an associated `GraphNode` will be generated automatically. We will later see this `GraphNode` as an input node in an `IGraphContext` and create and link new `GraphNode`s for its transitive dependencies.
 
 While we don't need to create these top-level `GraphNode`s we do sometimes need to adjust their properties.
 
--- a/docs/repo/dependencies-node-roadmap.md
+++ b/docs/repo/dependencies-node-roadmap.md
@@ -12,7 +12,7 @@ On the CPS side the project tree is composed of instances of the `IProjectTree` 
 
 An `IProjectTree` is immutable. When a part of the tree needs to be updated, we need to replace it and form a new tree.
 
-Components wishing to add items to the project tree we must implement and export the `IProjectTreeProvider` interface. In practice we implement the interface by deriving from the abstract `ProjectTreeProviderBase` class provided by CPS. In general, we receive events/messages from CPS detailing changes to the project, generated an updated `IProjectTree`, and then pass it back to CPS via `ProjectTreeProviderBase.SubmitTreeUpdateAsync`.
+Components wishing to add items to the project tree must implement and export the `IProjectTreeProvider` interface. In practice we implement the interface by deriving from the abstract `ProjectTreeProviderBase` class provided by CPS. In general, we receive events/messages from CPS detailing changes to the project, generate an updated `IProjectTree`, and then pass it back to CPS via `ProjectTreeProviderBase.SubmitTreeUpdateAsync`.
 
 ### The Graph Node Provider View of Dependencies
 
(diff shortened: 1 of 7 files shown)


docs/repo/dependencies-node-roadmap.md
```diff
@@ -26,7 +28,13 @@ The [`IDependenciesSnapshotProvider`][2] is responsible for providing access to
 
 Much of the code for the Dependencies node is concerned with creating [`IDependency`][4]s and translating them into new `IProjectTree`s when they change.
 
-## DependenciesProjectTreeProvider
+## CPS/Project System Interaction
+
+In general, items _directly_ referenced by the project file (e.g., through `Reference`, `ProjectReference`, and `Analyzer` items in the project file) are represented with `IProjectTree` nodes as opposed to `GraphNodes`. This makes it possible for them to the represented in the project's `IVsHierarchy` and, crucially, makes it easier to code the sorts of interactions users expect for these items. For example, a user should be able to right-click on an assembly reference and remove it, or modify the properties of an assembly refrence.
```
drewnoakes: Minor grammatical problem around _for them to the represented_. Also a typo in _refrence_.

docs/repo/dependencies-node-roadmap.md
```diff
@@ -4,19 +4,21 @@ This document describes, at a high level, the design and implementation of the D
...
-## The CPS View of Dependencies
+## Overview
+
+### The CPS View of Dependencies
 
 On the CPS side the project tree is composed of instances of the `IProjectTree` and `IProjectItemTree` interfaces. `IProjectTree` captures the structure of the tree (e.g. it has properties to access the parent and child nodes) and the "UI" aspects of the node--name, icons, visibility, etc. An `IProjectItemTree` captures all of that but also represents a concrete item within the project like a file, assembly reference, or NuGet package.
 
 An `IProjectTree` is immutable. When a part of the tree needs to be updated, we need to replace it and form a new tree.
 
 Components wishing to add items to the project tree we must implement and export the `IProjectTreeProvider` interface. In practice we implement the interface by deriving from the abstract `ProjectTreeProviderBase` class provided by CPS. In general, we receive events/messages from CPS detailing changes to the project, generated an updated `IProjectTree`, and then pass it back to CPS via `ProjectTreeProviderBase.SubmitTreeUpdateAsync`.
```
drewnoakes: I know it's not part of your PR but there are two typos here:

- project tree ~~we~~ must implement
- changes to the project, generate~~d~~ an updated

I can open a separate PR for these if you prefer.

docs/repo/dependencies-node-roadmap.md
```diff
@@ -26,7 +28,13 @@ The [`IDependenciesSnapshotProvider`][2] is responsible for providing access to
...
+
+> Aside: There are exceptions to this. In practice analyzers are not directly referenced but rather brought in as part of a NuGet package. We still represent all analyzers as `IProjectTree` items directly under the "Analyzers" node. This makes it much easier for the C#/VB language service to add nodes for each diagnostic underneath the analyzer's node, and it makes it easy for the user to find so they can check the severity of the diagnostic and potentially change it using the context menu.
+
+### DependenciesProjectTreeProvider
 
 The primary connection point between CPS and the Dependencies node is the [`DependenciesProjectTreeProvider`][1], implementing the the `IProjectTreeProvider` interface. It is directly responsible for the following:
```
drewnoakes: Also not in this PR, but there's a double _the the_.

docs/repo/dependencies-node-roadmap.md
```diff
@@ -57,15 +65,47 @@ The [`GroupedByTargetTreeViewProvider`][8] traverses down the existing `IProject
...
+New `GraphNode`s are added to the graph as a result of operations initiated by the user. For example, when the user expands a node representing a NuGet package:
+
+1. The `IGraphProvider.BeginGetGraphData` implementation in [`DependenciesGraphProvider`][22] is called with an `IGraphContext` describing the current graph, the input node (e.g. the node for the NuGet Package) and the operation (e.g. "get the children of the input node).
+2. Each implementation of [`IDependenciesGraphActionHandler`][23] that can handle the request is asked to do so.
+3. We retrieve the current [`IDependenciesSnapshot`][6] for the project as well as the [`IDependency`][4] for the input node (via [`IAggregateDependenciesSnapshotProvider`][10]/[`IDependenciesSnapshotProvider`][2]).
+4. We find the first [`IDependenciesGraphViewProvider`][25] that supports the given [`IDependency`][4] and ask it to build up the corresponding parts of the graph.
+5. The [`IDependenciesGraphViewProvider`][25] decides what nodes to add to the graph, and calls [`IDependenciesGraphBuilder`][24]`.AddGraphNode` (implemented by [`DependenciesGraphProvider`][22]) to handle the actual mechanics.
```
abpiskunov: - Very important detail that is not obvious is how GraphNodes are mapped to CPS IProjectTree nodes. It is done by GraphActionHandlerBase.GetDependency method.  IGraph Provider receives all kind of nodes .cs files Properties etc and we need to correctly identify a Dependency type of nodes. At the very beginning we identify CPS nodes by it's id serialized in the file path, but then when CPS node was found first time we add to it;s GraphNode ID property to simplify identification next time we process this node. This connection puts some restriction on what can be put into FilePath of IProjectTree and IProjectItemTree nodes.
- Also please notice, that now GraphNOdes can display any custom properties in Properties panel. I added a change to GraphNodes code sometime back to support it by providing extensions vis IGraphProvider.GetExtension method. We never ended up adding them to Dependencies code because of priorities at that time, but support for that was checked in in band. If you ever need to use it ping me and i will find related code changes.

tmeschter: @abpiskunov Yes, I was going to ask about your first point. When we create an `IProjectTree` instance for a NuGet package (for example) does CPS automatically create an equivalent `GraphNode` for us? I think that's what you're saying, but just want to be sure.

abpiskunov: CPS calls progression API (graph core logic) to notify and "map" a tree node to some graph node when tree node has ProjectTreeFlags.Common.ResolvedReference flag. Progression manages graph nodes accordingly. CPS does not explicitly kn ow anything about GraphNode objects.

tmeschter: Ah, ok. My main point, though, is that these are graph nodes that represent parts of the dependencies graph, but _aren't_ created by the code in dotnet/project-system, correct?

abpiskunov: Yes, we are not deleting or creating those top level GraphNodes, however we are changing their properties. We are managing everything below them, though if IProjectTree node is deleted, CPS signals to progression and progression deletes the whole tree under corresponding top level GraphNode eventually, that's why we have weak collection for expanded IGraphContexts to not hold them when they are deleted.

tmeschter:
@etbyrd Since you've been working in this area as well, I would appreciate any feedback you have on the document.

docs/repo/dependencies-node-roadmap.md
```diff
@@ -97,6 +97,12 @@ New `GraphNode`s are added to the graph as a result of operations initiated by t
 4. We find the first [`IDependenciesGraphViewProvider`][25] that supports the given [`IDependency`][4] and ask it to build up the corresponding parts of the graph.
 5. The [`IDependenciesGraphViewProvider`][25] decides what nodes to add to the graph, and calls [`IDependenciesGraphBuilder`][24]`.AddGraphNode` (implemented by [`DependenciesGraphProvider`][22]) to handle the actual mechanics.
 
+### Connecting `GraphNode`s to `IProjectTree` nodes
+
+`GraphNode`s are automatically created for `IProjectTree` nodes by CPS and the core graph logic. For example, we will create an `IProjectTree` for a top-level NuGet package and an associated `GraphNode` will be generated automatically. We will later see this `GraphNode` as an input node in an `IGraphContext` and create and link new `GraphNode`s for its transitive dependencies.
```
abpiskunov: Automtically for nodes with ResolvedRefernce flag, not for others.

PR #3944, opened 2018-09-05
Remove ConfigureAwait(true) and disable CA2007 (was RS0003)

davidwengier:
Guidance for VS is to use `.ConfigureAwait(true)` on async calls. Since this is the default when directly awaiting an async call we can disable the analyzer and remove the existing calls.

Guidance is here: https://github.com/Microsoft/vs-threading/blob/master/doc/cookbook_vs.md#should-i-await-a-task-with-configureawaitfalse


davidwengier:
Not a very exciting first PR, but you gotta start somewhere. Hopefully I haven't misunderstood the fix, or offline discussions around it.

PR #3955, opened 2018-09-06
Validate after replacing invalid chars

### LINKED ISSUE #3954: Resx editor prints a warning for for resource name \ identifiers with spaces
_From @pranavkm on June 16, 2016 18:26_

The way localization is designed to work in AspNetCore is that you use non-localized string to look up localized strings. For e.g.

``` C#
IStringLocalizer<HomeController> _localizer;
...

public IActionResult Index()
{
  ViewData["Message"] = _localizer["Hello World"];
}
```

This would print the message `Hello World` when a localized string with the identifier `Hello World` cannot be found or the localized string otherwise. Specifying a resource name `Hello World` works at runtime, however the designer prints a warning anytime the resx designer is opened:

```
Warning     The resource name 'Hello world' is not a valid identifier.      c:\release\WebApplication1\src\WebApplication1\Resources\Controllers\HomeController.resx
```

Is there a way to disable these messages?

_Copied from original issue: aspnet/Tooling#593_

davidwengier:
Rather than turn off the validation altogether, we now validate with the same method that re-writes the resource name for codegen. I couldn't actually get a validation failure this way as the codegen is pretty resilient, but it feels more future-proof.

Fixes #3954

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
@@ -2035,14 +2035,10 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
             'Is it a legal identifier?  We only check this if strongly-typed resource generation is set up for this
             '  resx file.
             If Not FastChecksOnly AndAlso _parentResourceFile IsNot Nothing AndAlso _parentResourceFile.IsGeneratedToCode() Then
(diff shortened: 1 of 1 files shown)


davkean:
Would a repro be invalid resource names, such as "$"?

davidwengier:
No, that gets stopped earlier on. Which makes me wonder if the idea behind the red error bubble was more as a warning to mean "The identifier is not going to exactly match what you've typed in", rather than a full on error which is this dialog:
image

PR #3956, opened 2018-09-06
Open the project file for editing on double click or ENTER

### LINKED ISSUE #1224: Show project file itself in Solution Explorer
One of the big benefits of `project.json` was ease of editing. Unfortunately, that's now hidden behind a right click and a menu scan for the user. Here's what that looks like:
screen shot 2017-01-17 at 16 40 23

In the `project.json` world, this access for human editing was far easier:
screen shot 2017-01-17 at 16 48 34

Can we please add the `.csproj` back to the tree for easy editing? It could take many forms. The project.json was easy because the name was consistent, which would be a challenge here. Either it could be a special node always at the top, or some other access.

davidwengier:
Open project file for editing on double click. Fixes #1224

See that issue for a gif of the feature.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/OpenProjectEditorOnDefaultActionCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/OpenProjectEditorOnDefaultActionCommand.cs
@@ -11,7 +11,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
 {
     // Opens the Project file in the text editor when the user double-clicks or presses ENTER on the project file while its selected
     [ProjectCommand(CommandGroup.UIHierarchyWindow, UIHierarchyWindowCommandId.DoubleClick, UIHierarchyWindowCommandId.EnterKey)]
-    [AppliesTo(ProjectCapability.OpenProjectFile)]
+    [AppliesTo(ProjectCapability.DoubleClickEditProjectFile)]
     [Order(Order.Default)]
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/OpenProjectEditorOnDefaultActionCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Input;
+using Microsoft.VisualStudio.ProjectSystem.Input;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    // Opens the Project file in the test editor when the user double-clicks or presses ENTER on the project file while its selected
+    [ProjectCommand(CommandGroup.UIHierarchyWindow, UIHierarchyWindowCommandId.DoubleClick, UIHierarchyWindowCommandId.EnterKey)]
+    [AppliesTo(ProjectCapability.AppDesigner)]
```
davkean: This capability should be "OpenProjectFile".

davidwengier: Thanks

davkean:
@Pilchie @cartermp As discussed, our thoughts for this is that we'll turn this on for Preview 1 and monitor feedback on it.

davkean:
tag @jviau - I'm guessing we'll want this in CPS long term.

PR #3957, opened 2018-09-06
Updating PreReleaseVersionLabel due to creation of dev15.9-preview2

Shyam-Gupta:
Updating PreReleaseVersionLabel due to creation of dev15.9-preview2 branch


Shyam-Gupta:
Not sure if this change is required.
Will confirm from @jasonmalinowski whether we need to bump up the version even for preview branches.

Closing the PR for now.

PR #3959, opened 2018-09-06
Add an icon set cache

tmeschter:
When we create a `Dependency` from another `Dependency` or a `DependencyModel` we re-use their existing `DependencyIconSet`s. This greatly reduces memory consumption related to `ImageMoniker`s as there are usually only a couple dozen distinct icon sets in use within a solution, even with tens of thousands of dependencies.

However, this doesn't help when we're creating a `Dependency` from an `IDependencyModel` supplied by another component--for example, the NPM/Bower package dependencies added by ASP.NET tooling. In those cases we always create a new `DependencyIconSet`.

This change reduces memory usage for those implementations by adding a cache of `DependencyIconSet`s. Every time a `Dependency` creates a new `DependencyIconSet` it runs it through the cache to see if there is an existing set with the same icons; if so, we use that one instead. So while we don't create any fewer instances of `DependencyIconSet` this way most of them will now be garbage collected quickly and only a few will survive long-term.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSet.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSet.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using Microsoft.VisualStudio.Imaging.Interop;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
@@ -10,7 +11,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
     /// sets; we can save considerable amounts of memory by given the sets their
     /// own type and sharing instances.
     /// </summary>
-    internal sealed class DependencyIconSet
+    internal sealed class DependencyIconSet : IEquatable<DependencyIconSet>
     {
         public DependencyIconSet(ImageMoniker icon, ImageMoniker expandedIcon, ImageMoniker unresolvedIcon, ImageMoniker unresolvedExpandedIcon)
         {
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSetCache.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Caches and reuses <see cref="DependencyIconSet"/> instances.
+    /// In practice dependencies use a relatively small number of distinct icon
+    /// sets; we can save considerable amounts of memory by ensuring that the same
+    /// logical sets are represented with the same instances.
+    /// </summary>
+    internal sealed class DependencyIconSetCache
+    {
+        private readonly ConcurrentDictionary<DependencyIconSet, DependencyIconSet> _iconSets = new ConcurrentDictionary<DependencyIconSet, DependencyIconSet>();
```
davkean: Nit: ThreadingTools.ApplyChangeOptimistically + ImmutableHashSet provides a nice lock-free implementation of this, see https://github.com/dotnet/project-system/pull/3960/files#diff-0e5ddc3ccc529a971c09299d53572076R67 for an example.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
```diff
@@ -196,13 +197,13 @@ protected set
...
-        private DependencyIconSet _iconSet;
 
-        public ImageMoniker Icon => _iconSet.Icon;
-        public ImageMoniker ExpandedIcon => _iconSet.ExpandedIcon;
-        public ImageMoniker UnresolvedIcon => _iconSet.UnresolvedIcon;
-        public ImageMoniker UnresolvedExpandedIcon => _iconSet.UnresolvedExpandedIcon;
+        public ImageMoniker Icon => IconSet.Icon;
+        public ImageMoniker ExpandedIcon => IconSet.ExpandedIcon;
+        public ImageMoniker UnresolvedIcon => IconSet.UnresolvedIcon;
+        public ImageMoniker UnresolvedExpandedIcon => IconSet.UnresolvedExpandedIcon;
```
abpiskunov: Are these needed now when we added IconSet as public property?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSetCache.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public DependencyIconSet GetOrAddIconSet(DependencyIconSet iconSet)
+        {
+            return _iconSets.GetOrAdd(iconSet, set => set);
+        }
+
+        public DependencyIconSet GetOrAddIconSet(ImageMoniker icon, ImageMoniker expandedIcon, ImageMoniker unresolvedIcon, ImageMoniker unresolvedExpandedIcon)
+        {
+            return GetOrAddIconSet(new DependencyIconSet(icon, expandedIcon, unresolvedIcon, unresolvedExpandedIcon));
```
abpiskunov: >new DependencyIconSet [](start = 35, length = 21)

You can avoid creating extra icon sets by having dictionary<hash, DependencyIconSet> ? Have static method GetHashCode(<icons> on the DependencyIconSet that would give you a hash to check for existence in the dictionary?

abpiskunov: Your implementation looks good and clean, i just thought if we could avoid unnecessary instantiations...

---
In reply to: [216084356](https://github.com/dotnet/project-system/pull/3959#discussion_r216084356) [](ancestors = 216084356)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
tmeschter: `Dependency` implements `IDependencyModel` which is a public interface, so unfortunately yes.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSetCache.cs
tmeschter: @davkean I'll implement that. Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSet.cs
```diff
@@ -44,5 +45,36 @@ public DependencyIconSet WithExpandedIcon(ImageMoniker newExpandedIcon)
...
+        {
+            var hashCode = Icon.Id.GetHashCode();
+            hashCode = hashCode * -1521134295 + Icon.Guid.GetHashCode();
+            hashCode = hashCode * -1521134295 + ExpandedIcon.Id.GetHashCode();
+            hashCode = hashCode * -1521134295 + ExpandedIcon.Guid.GetHashCode();
+            hashCode = hashCode * -1521134295 + UnresolvedIcon.Id.GetHashCode();
+            hashCode = hashCode * -1521134295 + UnresolvedIcon.Guid.GetHashCode();
+            hashCode = hashCode * -1521134295 + UnresolvedExpandedIcon.Id.GetHashCode();
+            hashCode = hashCode * -1521134295 + UnresolvedExpandedIcon.Guid.GetHashCode();
+            return hashCode;
```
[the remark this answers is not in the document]
tmeschter: I don't think so.

PR #3969, opened 2018-09-07
Use static classes where possible

drewnoakes:
These classes are not instantiated and only contain static/const members.

They are all test support types except for:

- `DependenciesGraphSchema`
- `ImmutableStringDictionary<TValue>`
- `TaskResult`

These three types are internal.


drewnoakes:
> I assume these changes came from an analyzer? Would be good to turn on/increase severity of same now that all of the issues are fixed.

I used R#. I'll be dogfooding without it from now on.

There's an analyser as part of the Roslynator. It looks like an interesting project, though I haven't tried it:

https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1102.md
https://github.com/JosefPihrt/Roslynator/blob/master/src/Analyzers/README.md

PR #3970, opened 2018-09-07
Use readonly fields where possible

drewnoakes:
All changed fields are private except for `TargetFramework.Empty`, `TargetFramework.Any`, `PackageFrameworkAssembliesViewModel.RegularIcon` and `ImmutableStringHashSet.*` which public fields on on internal types.

---

We could also add:

```
[*.{cs,vb}]
dotnet_style_readonly_field = true:warning
```

Right now VS shows them as suggestions.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.CSharp.UnitTests/ProjectSystem/LanguageServices/CSharpSyntaxFactsServiceTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.CSharp.UnitTests/ProjectSystem/LanguageServices/CSharpSyntaxFactsServiceTests.cs
@@ -7,7 +7,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
     [Trait("UnitTest", "ProjectSystem")]
     public class CSharpSyntaxFactsServiceTests
     {
-        private static ISyntaxFactsService s_service = new CSharpSyntaxFactsService(null);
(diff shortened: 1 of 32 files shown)


davkean:
We're already have them marked as warnings: https://github.com/dotnet/project-system/blob/master/.editorconfig#L55, so if they are showing up as suggestions - looks like Roslyn isn't respecting the setting?

drewnoakes:
> I'm definitely in favour on increasing severity level if everything is fixed.

There are quite a few VB cases that I didn't modify as less confident with VB.NET. If there's interest I can file another PR. It's quick to do.

> if they are showing up as suggestions - looks like Roslyn isn't respecting the setting

I've been using R# till now. It likely hides Roslyn's settings. As mentioned in #3969, I'll be dogfooding without it from now on.

davkean:
@drewnoakes It sounds like these are all the ones that R# found but not Roslyn - please open bugs against [Roslyn](github.com/dotnet/roslyn/issues/new) for every unique case like this.

drewnoakes:
@davkean sure thing. I don't know if Roslyn missed them. I'll investigate and open issues as necessary.

davkean:
I think they did; I've already made a pass to fix them all, these other others are in very old code and were not touched.

drewnoakes:
What about the VB occurrences?

davkean:
I've fixed every one that Roslyn found probably in VB if there were violations there.

drewnoakes:
Isolated the issue and reported as dotnet/roslyn#29746.

Created #3977 for the VB fields.

PR #3971, opened 2018-09-07
Fix Dependencies Node

tmeschter:
Commit b064bdf8 replaced a couple of lists of `IDisposable` items within `CrossTargetRuleSubscriberBase<T>` with a single `DisposableBag`. This broke the Dependencies node--it would still appear in Solution Explorer, but would never have any children.

The lists were holding various dataflow link subscriptions. These links may be added, disposed, and replaced multiple times during the life of the `CrossTargetRuleSubscriberBase<T>`. The issue is that while the original lists could be cleared and reused after disposing of their contents a `DisposableBag` can only be disposed once. After that any attempt to add an item to the bag simply causes it to be disposed immediately. So after the first call to `CrossTargetRuleSubscriberBase<T>.ReleaseSubscriptions` subsequent attempts to create and store new dataflow links would just cause them to be immediately disposed.

The fix here is to dispose and null out the `DisposableBag` when subscriptions are released, and then instantiate a new one as needed.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
@@ -19,7 +19,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
     {
 #pragma warning disable CA2213 // OnceInitializedOnceDisposedAsync are not tracked corretly by the IDisposeable analyzer
         private readonly SemaphoreSlim _gate = new SemaphoreSlim(initialCount: 1);
-        private readonly DisposableBag _subscriptions = new DisposableBag();
+        private DisposableBag _subscriptions;
 #pragma warning restore CA2213
         private readonly IUnconfiguredProjectCommonServices _commonServices;
         private readonly IProjectAsynchronousTasksService _tasksService;
@@ -79,7 +79,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
         public void ReleaseSubscriptions()
         {
             _currentProjectContext = null;
-            _subscriptions.Dispose();
+
+            if (_subscriptions != null)
+            {
+                _subscriptions.Dispose();
+                _subscriptions = null;
+            }
         }
 
         private void SubscribeToConfiguredProject(
@@ -101,6 +106,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
                     {
                         NameFormat = "CrossTarget Intermediate Evaluation Input: {1}"
                     });
+
+            if (_subscriptions == null)
+            {
+                _subscriptions = new DisposableBag();
+            }
 
             _subscriptions.AddDisposable(
                 subscriptionService.JointRuleSource.SourceBlock.LinkTo(


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
```diff
@@ -79,7 +79,12 @@ public void AddSubscriptions(AggregateCrossTargetProjectContext newProjectContex
         public void ReleaseSubscriptions()
         {
             _currentProjectContext = null;
-            _subscriptions.Dispose();
+
+            if (_subscriptions != null)
+            {
+                _subscriptions.Dispose();
+                _subscriptions = null;
+            }
```
Pilchie: Consider: `_subscriptions?.Dispose(); _subscriptions = null;`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
```diff
@@ -102,6 +107,11 @@ public void ReleaseSubscriptions()
                         NameFormat = "CrossTarget Intermediate Evaluation Input: {1}"
                     });
 
+            if (_subscriptions == null)
+            {
+                _subscriptions = new DisposableBag();
+            }
+
```
Pilchie: Consider `_subscriptions = _subscriptions ?? new DisposableBag();`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
```diff
@@ -102,6 +107,11 @@ public void ReleaseSubscriptions()
                         NameFormat = "CrossTarget Intermediate Evaluation Input: {1}"
                     });
 
+            if (_subscriptions == null)
```
abpiskunov: Multithreading?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
tmeschter: Sure.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
tmeschter: Sure

tmeschter:
@Pilchie I can't see a good way to unit test this. We would be better off with an integration test covering basic Dependencies node operations (ignoring the fact that integration tests are just not working right now).

abpiskunov:
Yes, we have integration tests in WTE enlistment for tree nodes (and used to have them for dnx in Dev14)  and they were always pretty reliable. Tom ping me and i will send you some TFS links to existing APEX tests to reuse as examples..

Pilchie:
Okay - in the meantime, how about a comment at the place where we set it to null indicating why?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
```diff
@@ -79,7 +79,9 @@ public void AddSubscriptions(AggregateCrossTargetProjectContext newProjectContex
         public void ReleaseSubscriptions()
         {
             _currentProjectContext = null;
-            _subscriptions.Dispose();
+
+            _subscriptions?.Dispose();
+            _subscriptions = null;
```
Pilchie: As @abpiskunov pointed out, I *think* these are multi-threaded, so both this and the assignment in the creation would need to be locked, or `Interlocked`.

tmeschter: No, the type does not support concurrent calls to `AddSubscriptions`/`ReleaseSubscriptions`. If it is _supposed_ to (and I don't think it is) then we need to change a lot more than this and we should handle that as a separate change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetRuleSubscriberBase.cs
Pilchie: Looks like calls are currently protected by `CrossTargetSubscriptionHostBase._linksLock`.

PR #3972, opened 2018-09-07
Cancel delay tasks, dispose of token source, and observe the task.

davidwengier:
Found this while doing other ConfigureAwait work. Uses the pattern from [here](https://github.com/dotnet/corefx/blob/851a53bee21639cd65e109b72fd5f9ef90cc0f39/src/Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs) except adds in disposal of the cts. No idea why System.Threading only does that in one of the methods.

Only "breaking" change here is the observation of the task so that we don't swallow exceptions. Currently the only place this method is used, outside of tests, is here:
https://github.com/dotnet/project-system/blob/dfe534a3e7803efb763ba9dcd35c5ce60c9104a5/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs#L807

No chance of throwing with that since its not a real task that does any work, just a completion source. See:
https://github.com/dotnet/project-system/blob/dfe534a3e7803efb763ba9dcd35c5ce60c9104a5/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs#L162

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskExtensions.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.VisualStudio.Threading.Tasks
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskExtensions.cs
```diff
@@ -26,12 +27,16 @@ internal static class TaskExtensions
         /// </exception>
         public static async Task<bool> TryWaitForCompleteOrTimeout(this Task task, int millisecondsTimeout)
```
sharwell: 💡 Consider replacing this method with `WithTimeout` from vs-threading:

https://github.com/Microsoft/vs-threading/blob/34c6eae89e0514ca6c6e72d411629cdb4cd88d2e/src/Microsoft.VisualStudio.Threading/TplExtensions.cs#L73

It's not a direct replacement since `WithTimeout` throws a `TimeoutException`, but it may work for what you need in practice.

davkean: I wouldn't use it here, because it's not a direct replacement.

davidwengier: Thanks for the tip @sharwell but I think the exception makes it not a great fit here. Will keep it in mind for the future.

PR #3974, opened 2018-09-09
Don't disponse HttpClient

davidwengier:
This is both good practice ([blog post](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/)) but also made the download of the file actually work in my dev environment. Previously I would regularly get task cancellations.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/DefaultHttpClient.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/DefaultHttpClient.cs
@@ -10,6 +10,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
     [Export(typeof(IHttpClient))]
     internal class DefaultHttpClient : IHttpClient
     {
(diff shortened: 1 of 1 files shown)


sharwell:
📝 The bug here is a failure to `await` inside the `using` statement. https://github.com/dotnet/roslyn-analyzers/issues/1752

davidwengier:
Can't believe I didn't see that!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/DefaultHttpClient.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
     [Export(typeof(IHttpClient))]
     internal class DefaultHttpClient : IHttpClient
     {
+        private static readonly Lazy<HttpClient> s_client = new Lazy<HttpClient>(CreateClient);
```
drewnoakes: `LazyThreadSafetyMode.ExecutionAndPublication`?

sharwell:
> Can't believe I didn't see that!

I can. It's scary-easy to miss.

PR #3978, opened 2018-09-10
Use pattern matching

drewnoakes:
Avoid duplicate type tests.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ErrorProfileDebugTargetsProvider.cs
```diff
@@ -62,10 +62,11 @@ public Task OnAfterLaunchAsync(DebugLaunchOptions launchOptions, ILaunchProfile
         /// </summary>
         public Task<IReadOnlyList<IDebugLaunchSettings>> QueryDebugTargetsAsync(DebugLaunchOptions launchOptions, ILaunchProfile activeProfile)
         {
-            if (activeProfile.OtherSettings.TryGetValue("ErrorString", out object oErrorString) && oErrorString is string)
+            if (activeProfile.OtherSettings.TryGetValue("ErrorString", out object oErrorString) && oErrorString is string s)
             {
-                throw new Exception(string.Format(VSResources.ErrorInProfilesFile2, Path.GetFileNameWithoutExtension(_configuredProject.UnconfiguredProject.FullPath), (string)oErrorString));
+                throw new Exception(string.Format(VSResources.ErrorInProfilesFile2, Path.GetFileNameWithoutExtension(_configuredProject.UnconfiguredProject.FullPath), s));
             }
```
tmeschter: I would prefer "objErrorString" or "errorStringObj" instead of "oErrorString", and "errorString" instead of just "s". In general I dislike single letter variable names as they convey very little information.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -97,7 +97,7 @@ public async Task InitializeAsync()
 
             // Check to see if a solution is already open. If so we set _solutionOpened to true so that subsequent projects added to 
             // this solution are processed.
-            if (ErrorHandler.Succeeded(solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out object isFullyLoaded)) && isFullyLoaded is bool && (bool)isFullyLoaded)
+            if (ErrorHandler.Succeeded(solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out object isFullyLoaded)) && isFullyLoaded is bool b && b)
             {
```
tmeschter: Again, I would prefer "objIsFullyLoaded" (or "isFullyLoadedObj") and "isFullyLoaded" instead of "isFullyLoaded" and "b".

drewnoakes:
@tmeschter thanks for the review. Your comments have been addressed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -97,7 +97,9 @@ public async Task InitializeAsync()
 
             // Check to see if a solution is already open. If so we set _solutionOpened to true so that subsequent projects added to 
             // this solution are processed.
-            if (ErrorHandler.Succeeded(solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out object isFullyLoaded)) && isFullyLoaded is bool && (bool)isFullyLoaded)
+            if (ErrorHandler.Succeeded(solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out object objIsFullyLoaded)) &&
+                objIsFullyLoaded is bool isFullyLoaded &&
+                isFullyLoaded)
```
davidwengier: Pretty sure this can be simplified further to just `objIsFullyLoaded is true` instead of the last two conditions.

[Seems to work without issue](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAIAYAEqCMBuAWAChjUBmbHANmwCZKB2TAb2M3ewtxtQBZMAsgEMAllAAUAShZsOcgG5CwmIZgC8mcQHsARgCsApgGNgkgERCAznABmZwkTkKlmHes27DJycEgGHThyygSI2mqoilpi+EAaSwU6sjoGYAJC4AJziFpiRUADkwNF+ZpJ4cglyAL6VQckhYeJukcWx8fWJtYGZ2ToQRc2WpQEpNR0VHWNVQA)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Converters.cs
```diff
@@ -99,7 +99,7 @@ public object Convert(object[] values, Type targetType, object parameter, Cultur
             foreach (object v in values)
             {
                 // Any false, the result is false
-                if (v is bool && !(bool)v)
+                if (v is bool b && !b)
```
davidwengier: As above, `v is false` should work

davkean:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
drewnoakes: You're right. That is cleaner.

I was curious what the `is` syntax actually does. Turns out it calls `object.Equals(object,object)` which means [it boxes the `true`](https://sharplab.io/#v2:C4LglgNgPgAgzAAhgRgGxIEwIMIIN4CwAUAqUoiugEYD2NECAsgBQ1UBWApgMbAI0BKfMTKikAdn4IwAZwTAATgFdOAbhFkAvsU1A===) (IL_0001), which means a small heap allocation.

Is this code called a lot? Where do we sit on the performance vs. readability spectrum in general?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Converters.cs
drewnoakes: As above.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
davkean: This is called once throughout lifetime of VS, readability wins out. That boxing sucks though for that pattern, but won't show up here at all.

drewnoakes: Fixed, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Converters.cs
drewnoakes: This is a multi-value converter that boxes every input and the output.

Take a look at c494b64944a0300cb0ee6271eb6e95ec34ba47ef which prevents any boxing allocations other than two during the type initialiser. Does that reduce readability too much?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
davidwengier: TIL, thanks. I thought the constant pattern did a type check first. I am less enamored by it.

PR #3980, opened 2018-09-10
Merges/languageservice to master

davkean:
Merge latest changes from features/LanguageService -> master. 

All these changes have already been reviewed in PRs.

PR #3982, opened 2018-09-11
Add doc around debugging async test hangs

davkean:
Ran into a hang in https://github.com/dotnet/project-system/pull/3980 - thought I would write up based on an internal doc that @jaredpar wrote how I figured out test that caused it.

### DIFF
--- a/docs/repo/debugging/async-test-hangs.md
+++ b/docs/repo/debugging/async-test-hangs.md
@@ -1,0 +1,271 @@
+# Debugging Async Test Hangs
+
+(Adapted from an internal document originally written by @jaredpar)
+
+Tests hanging on CI machines are often caused by async tests that are blocked waiting for results that will never complete. Unlike traditional synchronous hangs, attaching to an async hang and looking at executing threads typically will not provide anything useful as there won't be any threads actually executing the "awaiting"  methods.
+
+This document walks through debugging a test hang using WinDBG and SOS.
+
+## Prerequisites 
+
+- [WinDBG](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/debugger-download-tools)
+- Basic understanding of async machinery  
+- WinDBG attached to the hanging process, or opened with a dump from the hanging process
+
+## Loading SOS
+
+32-bit dump/process:
+
+```
+> .load C:\Windows\Microsoft.NET\Framework\v4.0.30319\sos.dll
+```
+
+64-bit dump/process:
+
+```
+> .load C:\Windows\Microsoft.NET\Framework64\v4.0.30319\sos.dll
+```
+
+## Determining the executing test
+
+We need to find the xUnit types in memory that track test exception, in this case; [InvokeTestAsync](https://github.com/xunit/xunit/blob/9d10262a3694bb099ddd783d735aba2a7aac759d/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestRunner.cs#L67). It is an async method, so we're going to find the blocked tests by looking through instances of its async state machine.
+
+```
+> !dumpheap -type InvokeTestAsync
+```
+
+```
+ Address       MT     Size
+036c8cb0 08389c3c       44     
+037a1828 08389c3c       44     
+039d9c40 08389c3c       44     
+03adb7c4 08389c3c       44     
+03b9cf68 08389c3c       44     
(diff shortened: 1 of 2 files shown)


docs/repo/debugging/async-test-hangs.md
```diff
@@ -0,0 +1,270 @@
...
+
+```
+
+The value of the `<>1__state` field is what we're interested in, this represents the "current state" of the state machine:
+
+Value|Meaning
+---:|---
+-2|Finished executing
+-1|Not started or currently excecuting (should be active call stack)
+>=0| Blocked in an await. The number indicates the zero-based ordinal of which await in the method is currently waiting.
```
davidwengier: Markdown formatter isn't liking this line, its not part of the table and rendering as a quote. Maybe try &gt;?

docs/repo/debugging/async-test-hangs.md
```diff
@@ -0,0 +1,270 @@
...
+7243f2dc  4000237       14         System.Int32  1 instance        0 <>1__state
+015c84d8  4000238       18 ...lib]], mscorlib]]  1 instance 036c8cc8 <>t__builder
+08383de4  4000239        4 ...k.XunitTestRunner  0 instance 036b843c <>4__this
+07869498  400023a        8 ...ceptionAggregator  0 instance 036b84e8 aggregator
+7243d488  400023b        c        System.String  0 instance 03501228 <output>5__2
+08389e5c  400023c       10 ....TestOutputHelper  0 instance 00000000 <testOutputHelper>5__3
+015c8dc4  400023d       24 ...cimal, mscorlib]]  1 instance 036c8cd4 <>u__1
+```
+
+In the above state machine instance, the value of ``<>1__state` is `0` indicating this currently blocked on the first await in [InvokeTestAsync](https://github.com/xunit/xunit/blob/9d10262a3694bb099ddd783d735aba2a7aac759d/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestRunner.cs#L67) which executes the test.
```
davidwengier: Double back-tick at the start of `<>1__state` is breaking the rendering

docs/repo/debugging/async-test-hangs.md
```diff
@@ -0,0 +1,270 @@
...
+(Adapted from an internal document originally written by @jaredpar)
+
+Tests hanging on CI machines are often caused by async tests that are blocked waiting for results that will never complete. Unlike traditional synchronous hangs, attaching to an async hang and looking at executing threads typically will not provide anything useful as there won't be any threads actually executing the "awaiting"  methods.
+
+This document walks through debugging a test hang using WinDBG and SOS.
+
+
+## Prerequisites 
+
+- [WinDBG](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/debugger-download-tools)
```
Pilchie: Is this really the primary external link to get WinDBG?

docs/repo/debugging/async-test-hangs.md
```diff
@@ -0,0 +1,270 @@
...
+0c037b98 08389c3c       44     
+0c05c2b0 08389c3c       44     
+
+Statistics:
+      MT    Count    TotalSize Class Name
+08389c3c       18          792 Xunit.Sdk.XunitTestRunner+<InvokeTestAsync>d__4
+Total 18 objects
+```
+
+The output tells me there's 18 instances of the async state machine in memory. What we are looking for are state machines that have not completed.
```
Pilchie: I'm sure there must be a foreach command to dump all of these and find the ones in the right state.

When looking for all uncollected exceptions in a dump, I often use something like this, which could probably be adapted here:

`.foreach (ex {!dumpheap -type Exception -short}){.echo "********************************";!pe ${ex} }`

docs/repo/debugging/async-test-hangs.md
davkean: As far as I can see yes. This seems to be no other page like this on Microsoft.

docs/repo/debugging/async-test-hangs.md
davkean: I timeboxed this and gave up trying to write something here - you can use `.if` works but couldn't work a syntax that worked.

In the end, I think this doc is going to be temporary, the CPS one is much better at finding this:

image
 
Started a conversation to figure out how to make this extension public.

Issue #3990, opened 2018-09-12
CreateProject_CreateAndBuild integration test incorrectly passes


davkean:
CreateProject_CreateAndBuild integration tests, incorrectly assumed that "true" returns from ProjectBuilt indicates that the project successfully built. It doesn't, it means that it was included in the last build. 
``` C#
                var success = VS.ObjectModel.Solution.BuildManager.Verify.ProjectBuilt(consoleProject); // <-- Does not mean "passed"
                string[] errors = new string[] { };
                if (!success)
                {
                    VS.ObjectModel.Shell.ToolWindows.ErrorList.WaitForErrorListItems();
                    errors = VS.ObjectModel.Shell.ToolWindows.ErrorList.Errors.Select(x => $"Description:'{x.Description}' Project:{x.ProjectName} Line:'{x.LineNumber}'").ToArray();
                }
                
                Assert.IsTrue(success, $"project '{consoleProject.FileName}' failed to build.{Environment.NewLine}errors:{Environment.NewLine}{string.Join(Environment.NewLine, errors)}");
```

In fact, the test is really failing with errors in the error list:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	NETSDK1004	Assets file 'C:\school\ApexSolutions\TestCase6\Prj_Cu6Xt\obj\project.assets.json' not found. Run a NuGet package restore to generate this file.	Prj_Cu6Xt	C:\Program Files\dotnet\sdk\2.1.401\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets	198	

```

davkean:
This will be the timing issue where we're scheduling the task that blocks the build on restore too late.

drewnoakes:
This is no longer an issue.

https://github.com/dotnet/project-system/blob/cbbaee5e95dc1af364ef7bddca58dd6125cba09c/tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/OpenProjectTests.cs#L56-L58

PR #3991, opened 2018-09-12
Allow integration tests to run from VS

davkean:
Integration tests were;

- Expecting that VsRootSuffix was alway passed to it from config settings - it's not. We respect the ROOTSUFFIX environment variable, or default to 'Exp'
- Always running against the last build installed, not the build from which you run the test. We now respect the environment settings set via VS to pick the right version when running in UI. Will not affect outside of VS.
- VS was left running. I have no idea how Apex is supposed to work, but right now we only have a single test - I just close it after it by opting into DTE mode (I can't figure out how to do this outside of that), we'll need to think about how to do this for multiple tests. What is the expectation here?
- Introduced TestBase class, MSTest has a really weird init model and won't let me put ClassInitialize on a base or as instance method. We should just move to xUnit if we don't need MSTest

After these changes, the test "passes" for 15.8, it fails for 16.0 because of missing 16.0.0.0 Apex dlls: https://github.com/dotnet/project-system/issues/3992.

The tests however, are really failing under the covers because of https://github.com/dotnet/project-system/issues/3990. I scoped that out of this change.

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -54,6 +54,7 @@
     <MicrosoftVisualStudioTextManagerInterop12Version>12.0.30110</MicrosoftVisualStudioTextManagerInterop12Version>
     <MicrosoftVisualStudioTextManagerInterop121DesignTimeVersion>12.1.30328</MicrosoftVisualStudioTextManagerInterop121DesignTimeVersion>
     <MicrosoftVisualStudioSetupConfigurationVersion>1.16.30</MicrosoftVisualStudioSetupConfigurationVersion>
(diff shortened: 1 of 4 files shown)


jmarolf:
> VS was left running. I have no idea how Apex is supposed to work

APEX typically leaves a VS instance open after the test run so that subsequent tests can re-use it (VS starting for each test adds a lot of extra time).  We should be able to modify APEX code to improve this.

davkean:
Okay thanks for that - based on that pointing me to the code, I can simplify the code a bunch.

davkean:
Merge this in and make follow up changes.

PR #3994, opened 2018-09-12
Simplify integration tests

davkean:
- Use default VisualStudio host
- Avoid second instance of Visual Studio opening
- Simplify test initialization (just use AssemblyInitialize, instead of ClassInitialize)
- Shutdown Visual Studio automatically after each test rather than making each test do it

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/TestBase.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/TestBase.cs
@@ -36,6 +36,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.IntegrationTests
             return visualStudioHostConfiguration;
         }
 
+        protected override void DoHostTestCleanup()
+        {
+            base.DoHostTestCleanup();
+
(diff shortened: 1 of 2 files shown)


davkean:
@jmarolf Going to merge this - but take a look and tell me if you have any feedback.

src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/TestBase.cs
```diff
@@ -5,39 +5,51 @@
...
-
-            SetTestInstallationDirectoryIfUnset();
+            // TestCleanup will fire up another instance of Visual Studio to reset 
+            // the AutoloadExternalChanges if it thinks the default changed even if
+            // that was just caused by settings to be sync'd. Just turn this feature off.
+            SuppressReloadPrompt = false;
         }
 
-        protected VisualStudioHostConfiguration DefaultHostConfiguration
+        protected override VisualStudioHostConfiguration GetHostConfiguration()
```
jmarolf: Nice! thanks for this

PR #4000, opened 2018-09-12
Pass cancellation tokens

drewnoakes:
These are occasions where we have a cancellation token in flight and were calling a method that supports cancellation without passing the token.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
@@ -78,7 +78,7 @@ namespace Microsoft.VisualStudio.Threading.Tasks
                 // First we wait the delay time. If another request has been made in the interval, then this task
                 // is cancelled. To avoid unnecessary OperationCanceled exceptions it tests to see if the token has
                 // been canceled
-                await Task.Delay(TaskDelayTime);
+                await Task.Delay(TaskDelayTime, token);
 
                 bool isCanceled = token.IsCancellationRequested;
                 lock (_syncObject)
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
```diff
@@ -78,7 +78,7 @@ private async Task ThrottleAsync(Func<CancellationToken, Task> asyncFnctionToCal
                 // First we wait the delay time. If another request has been made in the interval, then this task
                 // is cancelled. To avoid unnecessary OperationCanceled exceptions it tests to see if the token has
                 // been canceled
-                await Task.Delay(TaskDelayTime);
+                await Task.Delay(TaskDelayTime, token);
```
davkean: This method doesn't look like it expects OperationCancelledException, see comment about it.

drewnoakes: Quite right. I've pushed an update that addresses this. Please review again.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
```diff
@@ -76,9 +76,18 @@ private async Task ThrottleAsync(Func<CancellationToken, Task> asyncFnctionToCal
...
                 // First we wait the delay time. If another request has been made in the interval, then this task
-                // is cancelled. To avoid unnecessary OperationCanceled exceptions it tests to see if the token has
-                // been canceled
-                await Task.Delay(TaskDelayTime);
+                // is cancelled.
+                try
+                {
+                    if (!token.IsCancellationRequested)
+                    {
+                        await Task.Delay(TaskDelayTime, token);
```
Pilchie: I wonder if this will make debugging unit tests more painful.  Still seems better though.

drewnoakes: I usually disable breaking on `*CanceledException`. It probably still slows down both the debugger and debuggee.

Pilchie: Same, and exactly - it does slow them down some.

PR #4001, opened 2018-09-12
Fix typos in identifiers

drewnoakes:
Another pass at typos, this time in identifiers (riskier than in comments/docs) and this time watching brownbags.

I believe all changed identifiers are either local, private, internal or test names, and therefore that there are no breaking API changes here.

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/PropPages/PropertyPageException.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/PropPages/PropertyPageException.vb
@@ -33,9 +33,9 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Me.HelpLink = helpLink
         End Sub
 
-        Public Sub New(message As String, innerException As Exception, ShowHeaderandFooterInErrorControl As Boolean)
+        Public Sub New(message As String, innerException As Exception, ShowHeaderAndFooterInErrorControl As Boolean)
             MyBase.New(message, innerException)
-            _showHeaderAndFooterInErrorControl = ShowHeaderandFooterInErrorControl
+            _showHeaderAndFooterInErrorControl = ShowHeaderAndFooterInErrorControl
         End Sub
 
         ''' <summary>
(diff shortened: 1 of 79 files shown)


src/Microsoft.VisualStudio.AppDesigner/PropPages/PropertyPageException.vb
```diff
@@ -33,9 +33,9 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Me.HelpLink = helpLink
         End Sub
 
-        Public Sub New(message As String, innerException As Exception, ShowHeaderandFooterInErrorControl As Boolean)
+        Public Sub New(message As String, innerException As Exception, ShowHeaderAndFooterInErrorControl As Boolean)
```
Pilchie: This is *technically* a source breaking change, but I'm willing to live with it if anyone is consuming this and using named parameters ;)

davkean: Also can be a runtime one too if you are using dynamic binding, but I also came to the same conclusion. Make note that editors internals are public to the flavors as well, so technically also breaking when we change internals. I've ignored it for the most part as they really only use the property page infrastructure.

drewnoakes: What's the mechanism that allows internal members to be public to the flavors?

davkean: InternalsVisible listed under `My Project\AssemblyInfo.vb`.

drewnoakes: I see. We own them and they're a closed set, right? We're not talking about breaking customers here are we?

davkean: No, we own them. Don't worry about them - just making you aware that they exist and they tend to use Property Page functionality.

PR #4008, opened 2018-09-13
Plumb through IsActiveContext

davkean:
Register/Unregister the context when it is created so that the host can now query if the context is hosting is the active one.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextProviderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextProviderTests.cs
@@ -189,6 +189,23 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         }
 
         [Fact]
+        public async Task CreateProjectContextAsync_RegistersContextWithTracker()
+        {
+            IWorkspaceProjectContext result = null;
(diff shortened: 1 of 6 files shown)


davkean:
Windows_integration_debug was added because it originally targeted against master. It's never going to succeed on this branch.

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextProviderTests.cs
```diff
@@ -211,7 +243,7 @@ public async Task ReleaseProjectContextAsync_WhenContextThrows_SwallowsException
             await provider.ReleaseProjectContextAsync(projectContext);
         }
 
-        private static WorkspaceProjectContextProvider CreateInstance(UnconfiguredProject project = null, IProjectThreadingService threadingService = null, IWorkspaceProjectContextFactory workspaceProjectContextFactory = null, ISafeProjectGuidService projectGuidService = null, IProjectRuleSnapshot projectRuleSnapshot = null)
+        private static WorkspaceProjectContextProvider CreateInstance(UnconfiguredProject project = null, IProjectThreadingService threadingService = null, IWorkspaceProjectContextFactory workspaceProjectContextFactory = null, IActiveWorkspaceProjectContextTracker activeWorkspaceProjectContextTracker = null, ISafeProjectGuidService projectGuidService = null, IProjectRuleSnapshot projectRuleSnapshot = null)
```
Pilchie: Consider wrapping some lines here.

davkean: Yeah, I'll do it in later checkin.

PR #4009, opened 2018-09-13
Remove laziness of services

davkean:
Includes https://github.com/dotnet/project-system/pull/4008 so just review: https://github.com/dotnet/project-system/commit/a210f6bbfbdec86342be3f86ce7ae1ac6f54a521.

These paths are only activated when a configuration is made implicitly active - at which point, the language service is always going to be activated and these services always created. CPS has found that Lazy<T> usage over large number of projects adds up, so we should restrict its usage for uncommon paths.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceContextHost.WorkspaceContextHostInstance.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceContextHost.WorkspaceContextHostInstance.cs
@@ -1,6 +1,5 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-using System;
 using System.ComponentModel.Composition;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceContextHost.WorkspaceContextHostInstance.cs
```diff
@@ -52,8 +54,8 @@ public Task InitializeAsync()
 
             protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
             {
-                _context = await _workspaceProjectContextProvider.Value.CreateProjectContextAsync(_project)
-                                                                       .ConfigureAwait(true);
+                _context = await _workspaceProjectContextProvider.CreateProjectContextAsync(_project)
+                                                                  .ConfigureAwait(true);
```
drewnoakes: Unrelated to this PR, but do we still need `ConfigureAwait` in some cases?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceContextHost.WorkspaceContextHostInstance.cs
```diff
@@ -80,8 +82,8 @@ protected override async Task DisposeCoreUnderLockAsync(bool initialized)
 
                     if (_context != null)
                     {
-                        await _workspaceProjectContextProvider.Value.ReleaseProjectContextAsync(_context)
-                                                                    .ConfigureAwait(true);
+                        await _workspaceProjectContextProvider.ReleaseProjectContextAsync(_context)
+                                                              .ConfigureAwait(true);
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceContextHost.WorkspaceContextHostInstance.cs
davkean: In languageservices - I haven't back ported the rule turn off, so this will break the build if I remove it.

davkean:
Hmm, looks like blob storage went down. 

@dotnet-bot test this please

Issue #3744, opened 2018-07-18
BuildUpToDateCheck.OnChanged allocates 4% of all data opening Roslyn.sln


davkean:
GC time (both Gen0 and Gen1) is hurting the scenario of opening Roslyn.sln, on this path, below is allocating ~4% of all allocations splitting strings:

![image](https://user-images.githubusercontent.com/1103906/42854842-11513518-8a81-11e8-9934-f628311c630b.png)

davkean:
This is caused by splitting the `MSBuildAllProjects` property.

davkean:
This might be fixed by https://github.com/Microsoft/msbuild/issues/1299. We should verify.

davkean:
@drewnoakes Can you verify that this no longer shows up?  If you've not used PerfView I can show how use this next week. Jeff [said](https://github.com/Microsoft/msbuild/issues/1299#issuecomment-416996582) this went into Dev16, so you'll need a recent build of that.

davkean:
We might want to avoid the split too (to avoid the string[]) if there's no ";" as the MSBuild change means only a single target should appear in it.

drewnoakes:
@davkean Running VS 16.0 28022.4003.master the `MSBuildAllProjects` still splits into an array of 36 elements.

I'm not sure how to verify the version of msbuild running though.

![image](https://user-images.githubusercontent.com/350947/45320977-f604ae80-b53b-11e8-89bb-eef1cb25567f.png)

drewnoakes:
> We might want to avoid the split too (to avoid the string[]) if there's no ";" as the MSBuild change means only a single target should appear in it.

@davkean I wasn't able to observe a situation where `MSBuildAllProjects` contained a single path. Above I quoted 36 strings for a solution I generated. For `Roslyn.sln` on Dev16 I saw 54 strings.

My interpretation of the change to `MSBuildAllProjects` ([this comment](https://github.com/Microsoft/msbuild/issues/1299#issuecomment-412894038) in particular) is that the last-modified path will always be _first_. So if we are dealing with MSBuild 16.0 or greater, we need only take the first value for our checks.

Any thoughts on the best way to perform this version test?

Pilchie:
@jeffkl @agerlicher - do you know if an MSBuild with these changes has been inserted into Dev16 builds yet?

AndyGerlicher:
[...]

PR #4012, opened 2018-09-13
UpToDateCheck string allocation reduction

drewnoakes:
Relates to #3744.

Probably easiest to review commit-by-commit:

- a340a4d9c4cfbd8876770e8d74b3b5d24d544527 defers the splitting of the `MSBuildAllProjects` property value until it is used, reducing retained memory and preventing the majority of these split values being promoted from Gen0
- 882d5a067161a814bd258a1b70cdfdd41a929e53 adds another item to `Delimiter` to reduce params array allocations
- 7b729cc156bf63179254c85e4bf87574d6448b07 introduces `LazyStringSplit` and unit tests, supporting splitting strings lazily, and with considerably fewer allocated bytes than `string.Split`
- 496708c7a5fefad1fb925c79255966b5b6033a74 uses `LazyStringSplit` to split the `MSBuildAllProjects` value into short-lived strings, and also prevents allocating redundant strings that exist after one that identifies the build is out of date

I ran some benchmarks to verify behaviour of `LazyStringSplit` using the value of `MSBuildAllProjects` observed when building `Roslyn.sln`:

```
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-6700HQ CPU 2.60GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
  [Host]       : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3132.0
  Clr          : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3132.0
  LegacyJitX86 : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3132.0
Runtime=Clr
```

|    Method |          Job |       Jit | Platform |      Mean |     Error |    StdDev | Rank |   Gen 0 | Allocated |
|---------- |------------- |---------- |--------- |----------:|----------:|----------:|-----:|--------:|----------:|
|     Split |          Clr |    RyuJit |      X64 | 13.489 us | 0.2454 us | 0.2410 us |    3 | 11.1694 |   35265 B |
| LazySplit |          Clr |    RyuJit |      X64 |  7.359 us | 0.1086 us | 0.1016 us |    1 |  3.9825 |   12553 B |
|     Split | LegacyJitX86 | LegacyJit |      X86 | 16.707 us | 0.3169 us | 0.3523 us |    4 | 10.8643 |   34242 B |
| LazySplit | LegacyJitX86 | LegacyJit |      X86 | 10.134 us | 0.1881 us | 0.1760 us |    2 |  3.7384 |   11789 B |

`Roslyn.sln` has 164 projects, so the reduction in the _Allocated_ column can be multiplied out to around 3.7MB.

The _Gen0_ column is the number of collections per 1,000 invocations of the operation.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -119,11 +119,7 @@ private void OnProjectChanged(IProjectSubscriptionUpdate e)
             _msBuildProjectFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectFullPathProperty, _msBuildProjectFullPath);
             _msBuildProjectDirectory = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectDirectoryProperty, _msBuildProjectDirectory);
             _outputRelativeOrFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.OutputPathProperty, _outputRelativeOrFullPath);
-
-            string[] allProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, string.Empty)
-                .Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
-            _imports.Clear();
-            _imports.AddRange(allProjects);
+            _msBuildAllProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
```
Pilchie: @davkean does this repo have a stated preference for `string.Empty`/`String.Empty`/`""`?

Pilchie:
Nice!

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Text/LazyStringSplitTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [InlineData("", ';', new string[0])]
+        [InlineData(";", ';', new string[0])]
+        [InlineData(";;", ';', new string[0])]
+        [InlineData(";;;", ';', new string[0])]
+        [InlineData(";;;a", ';', new[] { "a" })]
+        [InlineData("a;;;", ';', new[] { "a" })]
+        [InlineData(";a;;", ';', new[] { "a" })]
+        [InlineData(";;a;", ';', new[] { "a" })]
+        [InlineData("a", ';', new[] { "a" })]
+        [InlineData("aa", ';', new[] { "aa" })]
```
davkean: Nit, I find it a little easier to view inputs/expectations if you turn this into table

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
```diff
@@ -0,0 +1,93 @@
...
+    ///     methods on <see cref="string"/>.
+    /// </remarks>
+    internal readonly struct LazyStringSplit : IEnumerable<string>
+    {
+        private readonly string _input;
+        private readonly char _delimiter;
+
+        public LazyStringSplit(string input, char delimiter)
+        {
+            _input = input ?? throw new ArgumentNullException(nameof(input));
```
davkean: We use Requires for contract checking.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
```diff
@@ -0,0 +1,93 @@
...
+                        _index++;
+                        continue;
+                    }
+
+                    Current = _input.Substring(_index, length);
+                    _index = delimiterIndex + 1;
+                    return true;
+                }
+
+                return false;
```
davkean: Just as a small note of interest, String.Split returns a 1 element array on empty string, you return zero elements, this is behavor acts more like the SplitReturningEmptyIfEmpty extension method we already have.

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Text/LazyStringSplitTests.cs
drewnoakes: Fixed in 189d881616fe3dfead1d3861b6ab2645559251d8.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
drewnoakes: Fixed in e678f4d4d2edfbb17cf687b64cb24d7caba0173d.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
drewnoakes: This type mimics the behaviour of `string.Split` when passing `StringSplitOptions.RemoveEmptyEntries`, which returns an empty array for on empty string.

I added an equivalence check to the unit test, and updated the documentation in 889f0d9c4cdd998d88f5ee60311321ce84a642e5.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: Find in files shows:

- 94 instances of `string.Empty`
- 119 instances of `String.Empty`
- 2575 instances of `""` (probably lots of false positives)

I don't feel strongly either way, though tend to use `""`.

davidwengier: String 🤮

Its the wrong colour! How can anyone live like this :P

jmarolf: I vote `string.Empty` please

drewnoakes: Sure. I won't change any more of these.

Pilchie: There are people who think that you should use `string` as a type name for variables, but `String` any time you access a static member off of the type.  I'm not sure *why* people think that, but some do.  🤷‍♂️

PR #4017, opened 2018-09-14
Minor changes to get language service working

davkean:
- Remove unused cancellation token, which simplifies a change I'm making in the future
- Opt components/services into both new and old language service, so when I turn on LanguageService2 they still turn on
- Remove "LanguageService" from dynamic capabilities
  - This capability is always a fixed capability (comes from project factory registration), so this change has zero impact on the product, but prevents me from manually switching from LanguageService -> LanguageService2 while testing.


davkean:

Looks like AV, which I suspect is CodeCov:

```
Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.DependencyTests.Dependency_Constructor_WhenRequiredParamsNotProvided_ShouldThrow()
   --- End of inner exception stack trace ---
```

davkean:

PR #4022, opened 2018-09-14
BuildUpToDateCheck documentation and refactor

drewnoakes:
Adds `docs/up-to-date-check.md` that describes the process used by `BuildUpToDateCheck`.

Also some minor refactoring and code documentation.

### DIFF
--- a/docs/up-to-date-check.md
+++ b/docs/up-to-date-check.md
@@ -1,0 +1,84 @@
+# Up-to-date Check
+
+If a project's outputs are up-to-date with its inputs, there's no need to run a build.
+
+The _Fast Up To Date Check_ attempts to perform a fast validation of whether a build is required.
+
+The `IBuildUpToDateCheckProvider` interface (from CPS) has two members:
+
+- `IsUpToDateCheckEnabledAsync`, which is serviced by `IProjectSystemOptions.GetFastUpToDateLoggingLevelAsync`
+- `IsUpToDateAsync` which performs the checks described below
+
+## What is checked
+
+There are several checks which must all pass. 
+
+All :heavy_check_mark: statements must be true for everything to be up to date.
+
+Checks occur in the order listed.
+
+As soon as one returns `false`, we are _not_ up to date and return immediately without performing further checks.
+
+### Environment
+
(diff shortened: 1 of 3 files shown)


docs/up-to-date-check.md
```diff
@@ -0,0 +1,84 @@
+# Up-to-date Check
```
Pilchie: The normal convention here is that only docs that we expect to point end-users to go in this directory, and docs about how to work in the repo, the implementation, etc go in the `docs/repo` directory.  This seems like more of the latter.

docs/up-to-date-check.md
```diff
@@ -0,0 +1,84 @@
...
+- [`BuildUpToDateCheck.cs`](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs)
+- [`BuildUpToDateCheckTests.cs`](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs)
+
+## Debugging
+
+The `BuildUpToDateCheck` class uses a `BuildUpToDateCheckLogger` to log verbose and info level information that is very useful during debugging.
+
+The log level is obtained via `IProjectSystemOptions.GetFastUpToDateLoggingLevelAsync`.
+
+TODO document how to view this log
```
Pilchie: https://github.com/dotnet/project-system/blob/master/docs/repo/debugging/logging.md documents this already.

docs/up-to-date-check.md
drewnoakes: Makes sense. I'll move it in a subsequent PR as #4023 follows on from this PR and modifies the file.

docs/up-to-date-check.md
drewnoakes: I don't see any logging related to the up to date check in the "Project" output panel. I'll have a look at CPS to see what it does with the `TextWriter` it passes on the interface.

drewnoakes: Am looking into this and will update once resolved.

PR #4023, opened 2018-09-14
Fix up-to-date check race condition

drewnoakes:
(Includes PR #4022, so just review 73438f95b3b52c101f430477bc5c851817298d7e.)

Issue #4014 describes a race condition. This PR fixes that race by introducing an additional check. If an input has been modified since the last time a check was requested, then the project is considered _not_ up to date.

PR #4024, opened 2018-09-14
[WIP] Update use of MSBuildAllProjects property

drewnoakes:
Microsoft/msbuild#3605 began a change where `MSBuildAllProjects` is used differently:

1. `MSBuildAllProjects` should not have items appended to it.
2. The first project listed in `MSBuildAllProjects` will be the newest one.

This PR includes #4023 so only review:

- 7344db713fba01e036f53a56dfef4364d7747d46 Don't append to `MSBuildAllProjects` property
- c8f28b220a31768b79269dbd7244d5b58ecfa4da Treat the first `MSBuildAllProjects` item as newest

Some questions I have (hence WIP):

- Do we need to put this new behaviour behind a gate based upon the version of MSBuild?
- What branch should this PR be targetting?

/cc: @jeffkl @AndyGerlicher

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -65,7 +65,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
         private string _msBuildProjectDirectory;
         private string _markerFile;
         private string _outputRelativeOrFullPath;
-        private string _msBuildAllProjects;
+        private string _newestProject;
 
         private DateTime _lastCheckTimeUtc = DateTime.MinValue;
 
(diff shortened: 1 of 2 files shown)


jeffkl:
As long as this code will _only_ ship with MSBuild 16.0+, you are safe to remove any MSBuildAllProjects that is part of the import graph.  

If you need to maintain back compat because your props/targets will be imported by older version of MSBuild, you would want to add a condition instead of removing the property:

```xml
<MSBuildAllProjects Condition="'$(MSBuildAssemblyVersion)' == '' Or '$(MSBuildAssemblyVersion)' < '16.0'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
```

This would reduce allocations in newer MSBuild and not break older versions.

AndyGerlicher:
I'm a little concerned about changing the consumption logic here. Consider the two cases:
```XML
<MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
```
```XML
<MSBuildAllProjects>$(MSBuildAllProjects);..\foo.txt</MSBuildAllProjects>
```
The first case is what we want to remove. Since MSBuild will magically add the newest project to the list, there's no reason to add `$(MSBuildThisFileFullPath)` for up-to-date project reasons. But that doesn't stop people from doing the 2nd option there to artificially influence the up-to-date checks. So it's good to not set that property anymore, but for legacy reasons we should really still treat it as a potential list of up-to-date candidates (guaranteed to be 1 or more items in MSBuild 16+).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -120,7 +126,9 @@ private void OnProjectChanged(IProjectSubscriptionUpdate e)
             _msBuildProjectFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectFullPathProperty, _msBuildProjectFullPath);
             _msBuildProjectDirectory = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectDirectoryProperty, _msBuildProjectDirectory);
             _outputRelativeOrFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.OutputPathProperty, _outputRelativeOrFullPath);
-            _msBuildAllProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
+            string msBuildAllProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
+
+            _newestProject = new LazyStringSplit(msBuildAllProjects, ';').FirstOrDefault();
```
AndyGerlicher: This is the newest of the MSBuild import graph, but it may not be the newest file in the `MSBuildAllProjects` "list". I don't think it buys much to assume there's only the single newest item at the front of the list.

PR #4033, opened 2018-09-18
Enforce UI-thread for EditAndContinueProvider

davkean:
Enforce UI-thread for EditAndContinueProvider. Underlying edit & continue makes the assumption that this is only ever called on UI thread. Enforce that.

Need to change this implementation to consume new APIs for language-service and it's less churn if I pull this out into its own change.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
@@ -20,153 +20,114 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.EditAndContinue
     internal class EditAndContinueProvider : IVsENCRebuildableProjectCfg, IVsENCRebuildableProjectCfg2, IVsENCRebuildableProjectCfg4, IDisposable
     {
         private ILanguageServiceHost _host;
+        private IProjectThreadingService _threadingService;
 
         [ImportingConstructor]
-        public EditAndContinueProvider(ILanguageServiceHost host)
+        public EditAndContinueProvider(ILanguageServiceHost host, IProjectThreadingService threadingService)
         {
             _host = host;
+            _threadingService = threadingService;
         }
 
         public int StartDebuggingPE()
         {
-            if (_host?.HostSpecificEditAndContinueService is IVsENCRebuildableProjectCfg2 encProvider)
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
```diff
@@ -20,160 +20,122 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.EditAndContinue
...
+            HResult hr = Invoke(encProvider => encProvider.GetExceptionSpanCount(out pcExceptionSpanResult));
+
+            pcExceptionSpan = pcExceptionSpanResult;
 
-            pcExceptionSpan = 0;
-            return HResult.Unexpected;
+            return hr;
         }
 
         public int GetExceptionSpans(uint celt, ENC_EXCEPTION_SPAN[] rgelt, ref uint pceltFetched)
```
drewnoakes: `pceltFetched` could be `out` rather than `ref`. Any input value is ignored.

drewnoakes: Just noticed this is probably on an interface and not changeable.

davkean: Yeah COM interfaces don't have concept of "out".

davkean: I stand corrected as GetExceptionSpanCount defines an `out` parameter. The IDL states this is `[in, out]`: https://devdiv.visualstudio.com/DevDiv/_git/VS.ref_archive?path=%2Fsrc%2FVSL%2Fidl%2FVBPublic%2Fencbuild.idl&version=GBmaster&line=241&lineStyle=plain&lineEnd=242&lineStartColumn=1&lineEndColumn=1, so either definition is wrong or there's a case where it provides an input (I can't actually find any callers). Either way, I should be respecting the definition.

davkean: Fixed: https://github.com/dotnet/project-system/pull/4040

PR #4040, opened 2018-09-19
Correctly assign ref argument

davkean:
As per: https://github.com/dotnet/project-system/pull/4033/files#r218715300.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/EditAndContinue/EditAndContinueProvider.cs
```diff
@@ -81,7 +81,7 @@ public int GetExceptionSpanCount(out uint pcExceptionSpan)
 
         public int GetExceptionSpans(uint celt, ENC_EXCEPTION_SPAN[] rgelt, ref uint pceltFetched)
         {
-            uint pceltFetchedResult = 0;
+            uint pceltFetchedResult = pceltFetched;
```
drewnoakes: You can lose the local var here and forward the ref directly.

drewnoakes: Might behave differently in case of exception.

davkean: I can't lose the local, can't capture a ref parameter in a lambda.

PR #4042, opened 2018-09-19
Use computed flags

### LINKED ISSUE #3996: Possible bug in GroupedByTargetTreeViewProvider
`GroupedByTargetTreeViewProvider.CreateOrUpdateProjectItemTreeNode` computes `flags` (line 404) but then never uses it. This computation is the only usage of its arguments `additionalFlags` and `excludedFlags`.

My _guess_ is `flags` should be used in place of `viewModel.Flags` below on line 421, as happens in related method `CreateOrUpdateProjectTreeNode`. If not then some code can be removed.

https://github.com/dotnet/project-system/blob/cc4af275e0f6b0a02462f6d291b58b82e9166dee/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GroupedByTargetTreeViewProvider.cs#L395-L431

tmeschter:
Fixes #3996.

The `GroupedByTargetTreeViewProvider.CreateOrUpdateProjectItemTreeNode` takes `additionalFlags` and `excludedFlags` parameters and uses them to update `viewModel.Flags`. However, the updated flags are not currently used.

This commit updates the method to make use of the flags. This will not change the current behavior as we do not actually pass in any relevant values for `additionalFlags` and `excludedFlags` but we want to avoid any surprises if/when we start to do so.


abpiskunov:
👍

PR #4043, opened 2018-09-19
Remove redundant check

tmeschter:
There are a number of places where we have a nullable value type and we're performing a null check like so:

``` C#
struct Foo { ... }
Foo? f = ...
if (f != null && h.HasValue) { ... }
```

Unless the `Foo` type has an operator that takes a `Foo?` as a parameter this check is redundant; we don't need both the null check and the call to `HasValue`.

This commit removes the null check in these cases.


tmeschter:
@Pilchie I just prefer the way it looks to have `foo.HasValue`/`foo.Value` rather than `foo != null`/`foo.Value`.

tmeschter:
@drewnoakes Thanks for the tip; not sure how I missed those. I've updated the PR to deal with those, too.

PR #4044, opened 2018-09-20
Port test hangs fixes from 15.9/master

davkean:
Porting test hangs fixes that I'm running into from 15.9/master.


davkean:
Going to merge this in as they have all been reviewed previously.

PR #4048, opened 2018-09-20
Replace Linq Count() with Any/Length

drewnoakes:
~~Replace O(N) with O(1).~~

`Enumerable.Count` on `string[]` is O(1) but still looks weird.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -924,8 +924,8 @@ internal void CreateProfile(string name, string commandName)
 
         internal bool IsNewProfileNameValid(string name)
         {
-            return LaunchProfiles.Where(
-                profile => LaunchProfile.IsSameProfileName(profile.Name, name)).Count() == 0;
+            return LaunchProfiles.Any(
```
davkean: You've reversed the lookup, you are missing a `!`.

PR #4055, opened 2018-09-24
Changed Neutral Language ComboBoxStyle to DropDownList

etbyrd:
Fixes: [646618](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646618)

The text that it announces isn't exactly what the "expected result" was, but the bug asked for the combo box to be announced, which it does now.

This also fixes [646590](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646590/)

Pressing Escape with the Neutral Language ComboBox expanded no longer closes the entire Assembly Information window, instead collapsing the Neutral Language ComboBox.

I feel like there is an added benefit to this change as well:

The Neutral Language ComboBox in the Assembly Information window checks to make sure that you inputted a correct Neutral Language if you try to enter your own. This essentially means that editing it is useless, because the only selections that you can use are already in the drop down list. If you enter something that does not match what is available you get this:
notalanguage

This means that it should be changed to a DropDownList style because it doesn't allow editing. 

Another case in the prop pages that already uses a DropDownList style is under Publish->Description->Publish language - and this is seemingly pulling from the same list of available languages, so the behavior should probably match.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoPropPage.Designer.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoPropPage.Designer.vb
@@ -256,6 +256,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Me.NeutralLanguageComboBox.AutoCompleteMode = System.Windows.Forms.AutoCompleteMode.SuggestAppend
             Me.NeutralLanguageComboBox.AutoCompleteSource = System.Windows.Forms.AutoCompleteSource.ListItems
             resources.ApplyResources(Me.NeutralLanguageComboBox, "NeutralLanguageComboBox")
+            Me.NeutralLanguageComboBox.DropDownStyle = System.Windows.Forms.ComboBoxStyle.DropDownList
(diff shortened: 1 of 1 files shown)


etbyrd:
If this is to be merged, we should probably remove the check for valid NeutralResourcesLanguage. It cannot be reached anymore if the ComboBox is not editable.

etbyrd:
I accidentally linked the same bug twice, and updated the PR request for clarity. 

I think those two behaviors are a regression and I am trying to find a better solution that fixes those. 

@davkean 

> I'm trying to understand, why does this fix the problem? I would expect the same behavior here after this after you've dropped the combo list.

Which behavior are you referring to?

etbyrd:
Closing this.

- [646618](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646618/) was a WinForms bug.

- [646590](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646590/) needs to be fixed a different way - a regression isn't acceptable. I will have to check to make sure that the window does not close when the ComboBox is opened, it is not the ComboBoxStyle's fault. Working on that PR now.

PR #4064, opened 2018-09-28
removing templates from project system (moving them to dotnet/templates)

jmarolf:


davkean:
Jon has the move already been done? I cannot see them in dotnet/templates?

jmarolf:
PR for dotnet/templates is here: https://github.com/dotnet/templates/pull/67

PR #4073, opened 2018-09-29
upgrade the version of VS that we run our tests against

jmarolf:
**NOTE**: this cannot be merged until we have new jenkins machines with dev16


jmarolf:
❔ 	We could make it so that we select which version of the nuget package to build against based off some variable so people could continue to use 15.8 for testing.  Seems like a lot of complexity to add though

PR #4099, opened 2018-10-10
Change default wording of the not supported version fallback message

davidwengier:
Previous message looks like it was written assuming compat information was updated early, but these days bits are available _very_ early.

Reported(?) on Twitter: https://twitter.com/ben_a_adams/status/1049847644745584641
Possibly the compat file should be updated to look further ahead than 3.0? Looks like @AngelosP owns that.


davidwengier:
@drewnoakes put down your phone and go to sleep :P

PR #4102, opened 2018-10-11
C# needs real unit/void support

davkean:
Add Task<T> overloads for downstream consumers (latter PR)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IWorkspaceProjectContextHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IWorkspaceProjectContextHost.cs
@@ -37,5 +37,23 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         ///     The result is awaited and the <see cref="ConfiguredProject"/> is unloaded.
         /// </exception>
         Task OpenContextForWriteAsync(Func<IWorkspaceProjectContextAccessor, Task> action);
+
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
```diff
@@ -71,6 +108,30 @@ protected Task ExecuteUnderLockAsync(Func<CancellationToken, Task> action, Cance
             return ExecuteUnderLockCoreAsync(action, cancellationToken);
         }
 
+        private async Task<T> ExecuteUnderLockCoreAsync<T>(Func<CancellationToken, Task<T>> action, CancellationToken cancellationToken = default)
+        {
+            Requires.NotNull(action, nameof(action));
```
drewnoakes: Callers already check this. Can remove one or the other.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
```diff
@@ -71,6 +108,30 @@ protected Task ExecuteUnderLockAsync(Func<CancellationToken, Task> action, Cance
             return ExecuteUnderLockCoreAsync(action, cancellationToken);
```
drewnoakes: Looks like you could merge the two `ExecuteUnderLockCoreAsync` methods, keeping only the generic `Task<T>` overload and passing a lambda here that produces an ignored return value. Would that introduce significant overhead?

PR #4103, opened 2018-10-11
Merge active context

davkean:
This includes https://github.com/dotnet/project-system/pull/4102 so only review https://github.com/dotnet/project-system/commit/ce49d0f9e6462a9f98685da6c3374a8c848d63fc.


davkean:
@drewnoakes Possibly but at the moment that's going to result in an explosion of overloads, I'll let the dust settle on the design before I plumb through that.

PR #4106, opened 2018-10-11
Refactor blocks

jmarolf:
There is now a `DataflowBlockSlim` class that exists to reduce memory pressure associated with using dataflow types.  This PR switches to use it instead of creating new dataflow objects.


jmarolf:
*NOTE to internal reviewers, DataflowBlockSlim is defined [here](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem%2FCore%2FDataflow%2FDataflowBlockSlim.cs&version=GBmaster)*

PR #4108, opened 2018-10-11
use delegate-based API instead of using-based API

jmarolf:
This was added by CPS to fix known dead lock issues


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -204,8 +204,7 @@ IReceivableSourceBlock<IProjectVersionedValue<IProjectTreeSnapshot>> IProjectTre
             // Get the list of normal reference Item Nodes (this excludes any shared import nodes).
             IEnumerable<IProjectTree> referenceItemNodes = nodes.Except(sharedImportNodes);
 
-            using (ProjectWriteLockReleaser access = await ProjectLockService.WriteLockAsync())
-            {
+            await ProjectLockService.WriteLockAsync(async access => {
```
Pilchie: brace on next line.

PR #4109, opened 2018-10-12
Add Resource rule for file and browseobject contexts

### LINKED ISSUE #3620: Graphics files in .net standard 2.0 library do have have build action of "Resource"
We have a few graphics files in .net standard 2.0 libraries, and .net 4.7.1 libraries. We are using     
MSBuildSdkExtras in these libraries. All the libraries have been converted to the new project system. However, when we try to set the build action to "Resource", on an image file like png or gif, it 
is not an available option in .net standard 2.0 libraries.

davidwengier:
This change allows selecting Resource as a build action in the properties dialog and ensures that is saved to the project file correctly. 

Fixes #3620


src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -263,6 +266,12 @@
       <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
       <SubType>Designer</SubType>
     </XamlPropertyRule>
+    <XamlPropertyRuleNoCodeBehind Include="ProjectSystem\Rules\Resource.BrowseObject.xaml" />
+    <XamlPropertyRule Include="ProjectSystem\Rules\Resource.xaml">
+      <XlfInput>false</XlfInput>
```
davkean: Why? The BrowseObject has strings.

davkean: Ugh, mb - this isn't the browseobject.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Resource.BrowseObject.xaml
```diff
@@ -0,0 +1,100 @@
...
+    <StringProperty.DataSource>
+      <DataSource PersistenceStyle="Attribute" SourceOfDefaultValue="AfterContext" />
+    </StringProperty.DataSource>
+  </StringProperty>
+  <StringProperty Name="Extension" Visible="False" ReadOnly="true">
+    <StringProperty.DataSource>
+      <DataSource Persistence="Intrinsic" ItemType="Resource" PersistedName="Extension" SourceOfDefaultValue="AfterContext" />
+    </StringProperty.DataSource>
+  </StringProperty>
+  <StringProperty Name="LastGenOutput" Visible="false" Description="The filename of the last file generated as a result of the SFG." />
```
davkean: For future reference during cleanup, we don't need descriptions for non-visible properties.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectItemsSchema.xaml
```diff
@@ -66,12 +66,15 @@
 
     <ContentType Name="XamlAppDef" DisplayName="Workflow definition" ItemType="XamlAppDef" />
 
+    <ContentType Name="Resource" DisplayName="Resource" ItemType="Resource" />
```
davkean: I wonder aloud if we want to go back and give these things better names to describe what they are used for; Resource (WPF), Application definition (WPF)?

davidwengier: I was wondering that too but at least with the Resource case I seem to remember using it in Winforms at least once. From memory it made it an embedded resource with no namespace prefix (as distinct from Embedded Resource), but maybe it doesn't matter if it says WPF in the name anyway

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Resource.BrowseObject.xaml
davidwengier: 👍 Makes sense. Probably more I can clean up in the other PR then too

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectItemsSchema.xaml
davkean: I don't believe this used by WinForms, or if you used it in WinForms, it was via the WinFx targets. See FileClassification target/FileClassifier task which handles these in Microsoft.WinFx.targets.

davkean: You might be thinking of csc's `-resource` switch, which is how `<EmbeddedResource/>` indirectly ends up in the assembly.

davidwengier: I must be remembering wrong, I've just checked and the Resource item type does nothing in a standard WinForms app.

PR #4110, opened 2018-10-12
Fix Tab Order on Build Event Command Line Prop Page

etbyrd:
Fixes: [646597](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646597) and [646592](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646592)

Previously, after expanding or collapsing the macro list, the tab order would move down to the OK button instead of staying on the Macro button. This PR fixes that behavior.

After clicking on Macros >> before fix:
taborderbeforefix
It moves to the OK button.

After clicking on Macros >> After fix:
taborderafterfix
It stays on the (now visible)  << Macros Button.

It also fixes it in the reverse case.


davkean:
@Pilchie Do we need approval for these changes for Preview 1?

PR #4112, opened 2018-10-15
 Fixed NullReferenceException when failing to resolve type

davkean:
Fixes: devdiv.visualstudio.com/DevDiv/_workitems/edit/705725.

Finally got some cabs for this watson that I've been tracking for a while. GetRuntimeType returns null on a variety of situations under the covers, for example, when it fails load a type because it references another assembly that isn't available to the project.

I repro'd this simply by adding a reference to Microsoft.AspNetCore.Mvc.Abstractions and attempting to expand it in the TypePickerDialog.


src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
```diff
@@ -197,11 +197,17 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
...
-                                If availableType.FullName.Contains(".") AndAlso SettingTypeValidator.IsValidSettingType(mtsrv.GetRuntimeType(availableType)) Then
-                                    _typeTreeView.AddTypeNode(Node, availableType.FullName)
+                            Dim availableTypes As Type() = mtsrv.GetSupportedTypes(Node.Text, AddressOf GetAssemblyCallback)
+                            For Each availableType As Type In availableTypes
+
+                                If availableType.FullName.Contains(".") Then
+
+                                    ' NOTE that GetRuntimeType returns null when there's a failure resolving the type
+                                    Dim runtimeType = mtsrv.GetRuntimeType(availableType)
+                                    If runtimeType IsNot Nothing And SettingTypeValidator.IsValidSettingType(runtimeType) Then
```
davidwengier: Should this And be an AndAlso? Without that the behaviour is unchanged isn't it (ie, SettingTypeValidator.IsValidSettingType can still be passed a null).

PR #4113, opened 2018-10-15
Fix crash when adding new setting

davkean:
Another Watson I was watching with cabs.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/705752

CreateUniqueName and EqualIdentifiers had different definitions of what a unqiue identifier was, so if you had a setting called 'setting', and clicked the setting dropdown on the next row, we'd watson because we'd attempt to add new setting called by the default "Setting". Treated them as the same, causes the new setting to be called "Setting1" and avoid the watson.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsDesigner/DesignTimeSettings.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsDesigner/DesignTimeSettings.vb
@@ -1,4 +1,4 @@
-' Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿' Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 Imports System.ComponentModel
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editors/SettingsDesigner/DesignTimeSettings.vb
```diff
@@ -177,15 +177,15 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
...
+            Return StringComparers.SettingNames.Equals(Id1, Id2)
         End Function
 
         Friend Function CreateUniqueName(Optional Base As String = Nothing) As String
             If String.IsNullOrEmpty(Base) Then
                 Base = My.Resources.Microsoft_VisualStudio_Editors_Designer.SD_DefaultSettingName
             End If
 
-            Dim ExistingNames As New Hashtable
+            Dim ExistingNames As New Hashtable(StringComparers.SettingNames)
```
Pilchie: It would be nice to ban `New Dictionary(of String, T)()` and force specifying a comparer.

Pilchie: Not sure how it handles specifically instantiated generics.  And of course, this is just a `Hashtable` that you are putting strings in, so even harder to find...

PR #4114, opened 2018-10-15
Avoid crash when StartProgram has invalid characters

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/705763

When StartProgram has invalid characters and remote debugging is enabled, there are certain cases where the property page crashes during validation. When remote debugging is disabled, the validation check was being done in File.Exists which will always return false for invalid characters and would never hit this case.

~Merging into 15.9~, because this is our highest watson, with about ~100 hits in the past month.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/DebugPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/DebugPropPage.vb
@@ -340,7 +340,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                         If Trim(StartProgram.Text).Length = 0 Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_NeedExternalProgram
                             Return ValidationResult.Warning
-                        ElseIf Not Path.GetExtension(StartProgram.Text).Equals(".exe", StringComparison.OrdinalIgnoreCase) Then
+                        ElseIf Not StartProgram.Text.EndsWith(".exe", StringComparison.OrdinalIgnoreCase) Then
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/DebugPropPage.vb
```diff
@@ -340,7 +340,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                         If Trim(StartProgram.Text).Length = 0 Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_NeedExternalProgram
                             Return ValidationResult.Warning
-                        ElseIf Not Path.GetExtension(StartProgram.Text).Equals(".exe", StringComparison.OrdinalIgnoreCase) Then
+                        ElseIf Not StartProgram.Text.EndsWith(".exe", StringComparison.OrdinalIgnoreCase) Then
```
drewnoakes: Is there a constant for the comparer?

Pilchie: I believe this now matches a file named just `.exe` where the previous one would not.  Don't think it necessarily matters...

Pilchie:
15.9 is in QB approval mode now, so we'll need approval before merging.

src/Microsoft.VisualStudio.Editors/PropPages/DebugPropPage.vb
davkean: String comparers don't do EndsWith, so my usual pattern doesn't work.

drewnoakes: I confused `StringComparison` with `StringComparer`.

PR #4119, opened 2018-10-15
Miscellaneous changes to dependencies code

drewnoakes:
A grab bag of stuff I saw while working on dependency node bugs today.

Easiest to review commits separately. Descriptions in commit messages.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
@@ -368,11 +368,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
             }
             finally
             {
-                sb.Clear();
-
                 // Prevent holding on to large builders
-                if (sb.Length < 1000)
+                if (sb?.Capacity < 1000)
                 {
+                    sb.Clear();
                     s_builderPool.Add(sb);
                 }
             }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Globalization;
@@ -102,13 +103,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
             }
             else
             {
-                ImmutableList<string>.Builder normalizedDependencyIDs = ImmutableList.CreateBuilder<string>();
-                foreach (string id in dependencyModel.DependencyIDs)
-                {
-                    normalizedDependencyIDs.Add(GetID(TargetFramework, ProviderType, id));
-                }
-
-                DependencyIDs = normalizedDependencyIDs.ToImmutable();
+                IEnumerable<string> ids = dependencyModel.DependencyIDs.Select(
+                    id => GetID(TargetFramework, ProviderType, id));
+                DependencyIDs = ImmutableArray.CreateRange(ids);
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
@@ -103,9 +103,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
             }
             else
             {
-                IEnumerable<string> ids = dependencyModel.DependencyIDs.Select(
-                    id => GetID(TargetFramework, ProviderType, id));
-                DependencyIDs = ImmutableArray.CreateRange(ids);
+                int count = dependencyModel.DependencyIDs.Count;
+                ImmutableArray<string>.Builder ids = ImmutableArray.CreateBuilder<string>(count);
+                for (int i = 0; i < count; i++)
+                    ids.Add(GetID(TargetFramework, ProviderType, dependencyModel.DependencyIDs[i]));
+                DependencyIDs = ids.MoveToImmutable();
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/AnalyzerRuleHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/AnalyzerRuleHandler.cs
@@ -24,9 +24,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscription
             unresolvedIcon: ManagedImageMonikers.CodeInformationWarning,
             unresolvedExpandedIcon: ManagedImageMonikers.CodeInformationWarning);
 
-        protected override string UnresolvedRuleName { get; } = AnalyzerReference.SchemaName;
-        protected override string ResolvedRuleName { get; } = ResolvedAnalyzerReference.SchemaName;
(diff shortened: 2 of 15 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/AnalyzerRuleHandler.cs
```diff
@@ -24,9 +24,9 @@ internal class AnalyzerRuleHandler : DependenciesRuleHandlerBase
             unresolvedIcon: ManagedImageMonikers.CodeInformationWarning,
             unresolvedExpandedIcon: ManagedImageMonikers.CodeInformationWarning);
 
-        protected override string UnresolvedRuleName { get; } = AnalyzerReference.SchemaName;
-        protected override string ResolvedRuleName { get; } = ResolvedAnalyzerReference.SchemaName;
-        public override string ProviderType { get; } = ProviderTypeString;
+        protected override string UnresolvedRuleName => AnalyzerReference.SchemaName;
+        protected override string ResolvedRuleName => ResolvedAnalyzerReference.SchemaName;
+        public override string ProviderType => ProviderTypeString;
```
Pilchie: I assume the right-hand-sides here are all immutable?  Otherwise this is a change in behavior.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleHandlerBase.cs
```diff
@@ -71,16 +59,16 @@ public virtual bool ReceiveUpdatesWithEmptyProjectChange
         {
             if (projectChanges.TryGetValue(UnresolvedRuleName, out IProjectChangeDescription unresolvedChanges))
             {
-                HandleChangesForRule(false /*unresolved*/,
-                    unresolvedChanges, context, ruleChangeContext,
-                        (itemSpec) => { return true; });
+                HandleChangesForRule(
+                    resolved: false, unresolvedChanges, context, ruleChangeContext,
+                    itemSpec => true);
```
Pilchie: Not introduce, but a name for what this delegate is would be nice too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
```diff
@@ -368,11 +368,10 @@ public static string GetID(ITargetFramework targetFramework, string providerType
             }
             finally
             {
-                sb.Clear();
-
                 // Prevent holding on to large builders
-                if (sb.Length < 1000)
+                if (sb?.Capacity < 1000)
```
Pilchie: Whoops - my fault.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
```diff
@@ -102,13 +103,9 @@ public Dependency(IDependencyModel dependencyModel, ITargetFramework targetFrame
...
-                ImmutableList<string>.Builder normalizedDependencyIDs = ImmutableList.CreateBuilder<string>();
-                foreach (string id in dependencyModel.DependencyIDs)
-                {
-                    normalizedDependencyIDs.Add(GetID(TargetFramework, ProviderType, id));
-                }
-
-                DependencyIDs = normalizedDependencyIDs.ToImmutable();
+                IEnumerable<string> ids = dependencyModel.DependencyIDs.Select(
+                    id => GetID(TargetFramework, ProviderType, id));
+                DependencyIDs = ImmutableArray.CreateRange(ids);
```
Pilchie: Could we use an `ImmutableArrayBuilder` here, and pre-size it with `dependencyModel.DependencyIDs.Count`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/AnalyzerRuleHandler.cs
drewnoakes: Yes, they're all const strings.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
drewnoakes: It would have been hard to hit this in practice. Perhaps if `targetFramework.ShortName` was `null`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleHandlerBase.cs
drewnoakes: Done.

davkean:
This looks like real test failures.

drewnoakes:
> This looks like real test failures.

Yep, I wanted to pack it in for the day but get this stuff visible for review overnight. I didn't run the tests before pushing, sorry. Picking it up now, along with other review feedback.

davkean:
@drewnoakes Feel free to push away - much better to use Azure resources to find failures for you than be blocked on tests running.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
drewnoakes: I couldn't find an approach here that didn't involve redundant copying. The builder always does a defensive copy in the end. The closest was an `ImmutableArray.CreateRange` overload that takes a func for mapping (so can create the backing array internally without defensive copies), but it only accepts `ImmutableArray` as an argument, and `DependencyIds` is `IImmutableList<string>`. It would be possible to [add such a method](https://github.com/dotnet/corefx/issues/32839) to `ImmutableArray`.

Even with the redundant copies, the builder is still faster and allocates less than `AddRange` 👍 

|                             Method | Count |       Mean |     Error |     StdDev |  Gen 0 | Allocated |
|----------------------------------- |------ |-----------:|----------:|-----------:|-------:|----------:|
|         ImmutableArray_CreateRange |     3 |   470.1 ns | 10.096 ns |  16.013 ns | 0.0196 |     104 B |
|     ImmutableArray_Builder |     3 |   107.2 ns |  1.483 ns |   1.315 ns | 0.0122 |      64 B |
|         ImmutableArray_CreateRange |    30 | 1,634.1 ns | 32.726 ns |  83.297 ns | 0.0896 |     472 B |
|     ImmutableArray_Builder |    30 |   585.9 ns | 11.733 ns |  18.609 ns | 0.0525 |     280 B |

davkean: MoveToImmutable avoids a copy.

drewnoakes: Very nice. Updated.

|                                 Method | Count |        Mean |      Error |      StdDev |  Gen 0 | Allocated |
|--------------------------------------- |------ |------------:|-----------:|------------:|-------:|----------:|
|             ImmutableArray_CreateRange |     3 |   473.69 ns |  9.3715 ns |  10.0274 ns | 0.0191 |     104 B |
|     ImmutableArray_Builder_ToImmutable |     3 |   102.87 ns |  2.1851 ns |   3.8271 ns | 0.0122 |      64 B |
| ImmutableArray_Builder_MoveToImmutable |     3 |    73.28 ns |  0.7152 ns |   0.5584 ns | 0.0075 |      40 B |
|             ImmutableArray_CreateRange |    30 | 1,860.60 ns | 72.7219 ns | 209.8194 ns | 0.0896 |     472 B |
|     ImmutableArray_Builder_ToImmutable |    30 |   534.11 ns | 11.9213 ns |  24.6195 ns | 0.0525 |     280 B |
| ImmutableArray_Builder_MoveToImmutable |    30 |   492.87 ns | 10.0324 ns |  14.7054 ns | 0.0277 |     148 B |

drewnoakes:
> My personal struggle with expression bodied members continues on 😛

As much as I like them for short expressions, I'll try and avoid introducing them in places where the difference is only one of style. Three snuck through here. For the others (the majority) we can lose the backing fields.

PR #4123, opened 2018-10-16
 Avoid crashing when editing settings while debugging

davkean:
(This PR is easier to consume if you look at a commit at a time.)

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/706537.

If you edit the settings designer while debugging the project, Visual Studio crashes because checkout throws an ApplicationException when debugging.

There's two case that turn the "setting designer" into read-only mode; building the project (IsReadOnly) and debugging the project (!IsDesignMode). This change centralizes all usage of these cases into a single "IsDesignerEditable" property to avoid cases where we check one of those cases and not the other. Also removes the need to check these when we attempt to check-out the project. 

This crash has about 32 hits over the past month, so putting it in for 15.9 candidate.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsDesignerLoader.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsDesignerLoader.vb
@@ -772,6 +772,14 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
             End If
         End Sub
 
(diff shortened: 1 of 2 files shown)


davkean:
@jmarolf I'm not sure what's going on with this spanish run.

jmarolf:
@davkean ignore this failure. Looks like the dnceng team took all the agents out of that queue.  I will follow up on them regarding why they did that.

Issue #646, opened 2016-10-25
New debug dialog isn't scaling correctly with High DPI


dsplaisted:
I see the following when I open up the Debug page:

![image](https://cloud.githubusercontent.com/assets/145043/19669319/510ba144-9a13-11e6-8bb6-f559ab08c156.png)

This is on a laptop that is normally at 200% DPI but I'm logged into via Remote Desktop with 100% DPI.

davkean:
Hmm, both cases looks like its scaling via DPI, and not Font. It should be scaling via Font, similar to what we do for WinForm's pages.

davkean:
But the Font in both cases looks wrong.

davkean:
Feedback item here: https://github.com/dotnet/project-system/issues/3372.

davkean:
@dsplaisted Is this still repro'ing for you? I've tried lots of different situations, and none of them look as bad as you have above. This is me remote desktop'ing at 100% into a 200% Surface book:

![image](https://user-images.githubusercontent.com/1103906/46998100-4b506300-d16d-11e8-8278-bd22f2d9ce5a.png).

Regardless, after https://github.com/dotnet/project-system/pull/4124 the Fonts will be inherited from VS itself so will match the same size/family as the everything else.

Pilchie:
Windows has been making changes to scaling for HDPI, so it might depend on exact Windows version and settings too...

dsplaisted:
@davkean  I haven't seen this recently

davkean:
@Pilchie Yes I expect this as a result of a Windows change.

PR #4124, opened 2018-10-16
Inherit Font/Colors in WPF property pages

davkean:
Fixes: #2122 
Fixes: #646 
WPF will automatically inherit Windows Forms fonts https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/windows-forms-and-wpf-property-mapping. Make sure we set the Control.Parent if it's available, which sets this up.


davkean:
@dotnet/project-system Can someone with a high DPI machine pull this change down and share me a screenshot of Application + Debug page?

davkean:
@davidwengier Pulled down this, and it looks correct. 

I've tried this at:

100% on a Surface Book (267)
150% on a Surface Book (200)
200% on a Surface Book (133.5)
100% on a HP ZR30w (102)
100% on a HP ZR30w remoting into a Surface Book at 200%.

I cannot repro the situations called out here: https://github.com/dotnet/project-system/issues/646 - but if they are still occuring this should be fixed by this change as the Font will match the rest of VS.

PR #4126, opened 2018-10-16
SDKVersion was pushing telemetry under wrong name

davkean:
This broke when d897cd2d merged into master, this code path changed to no longer automatically add the event prefix. Change this code to use the same pattern everywhere else.

This was only broken in Dev16.


davkean:
@jmarolf I was making the assumption I broke this - but tests seem to indicate that the name of the event/properties have never had a prefix. Is that right?

jmarolf:
@davkean I thought that the telemetry service auto-appended `vs/projectsystem/managed` to every event that it logged.  Mock of telemetry service didn't do this.

davkean:
@jmarolf Ah, yes that explains it - and does confirm that I broke it. Thanks!

PR #4036, opened 2018-09-18
Various XML doc fixes

drewnoakes:
Fixes various issues with XML documentation.

- Add using directives that are required only to resolve symbols in XML docs
- Fixed some references/signatures
- Minor formatting

PR #4129, opened 2018-10-17
Update version of RoslynTools.RepoToolset

drewnoakes:
Relates to #3700 (excludes the YAML bit) and hopefully unblocks #4036.

Version number sourced from https://dotnet.myget.org/feed/roslyn-tools/package/nuget/RoslynTools.RepoToolset

/cc @jmarolf @tmat

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -11,7 +11,7 @@
     <UsingToolIbcOptimization>true</UsingToolIbcOptimization>
 
     <!-- Toolset -->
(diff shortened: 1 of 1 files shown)


drewnoakes:
Turns out this wasn't as simple as I'd hoped and is now leading into areas I'm not familiar with.

@jmarolf what will it take to get the toolset updated?

drewnoakes:
Closing this as it doesn't appear to be anywhere close to what we need.

PR #4142, opened 2018-10-18
Use tuple deconstruction with foreach

drewnoakes:
This is a style thing mostly. I find it clearer, but it's newish syntax.

Some cases of this conversion highlighted that only the value was needed. Such cases were changed to enumerate `Values` instead. This potentially changes the order of enumeration, though dictionaries are not generally considered ordered. Note in review that there's a mix of mutable and immutable dictionaries here.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
@@ -201,12 +201,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
             ImmutableDictionary<ITargetFramework, ITargetedProjectContext>.Builder innerProjectContextsBuilder = ImmutableDictionary.CreateBuilder<ITargetFramework, ITargetedProjectContext>();
             ITargetFramework activeTargetFramework = TargetFramework.Empty;
 
-            foreach (KeyValuePair<string, ConfiguredProject> kvp in configuredProjectsMap)
+            foreach ((string tfm, ConfiguredProject configuredProject) in configuredProjectsMap)
             {
-                ConfiguredProject configuredProject = kvp.Value;
                 ProjectProperties projectProperties = configuredProject.Services.ExportProvider.GetExportedValue<ProjectProperties>();
                 ConfigurationGeneral configurationGeneralProperties = await projectProperties.GetConfigurationGeneralPropertiesAsync();
-                ITargetFramework targetFramework = await GetTargetFrameworkAsync(kvp.Key, configurationGeneralProperties);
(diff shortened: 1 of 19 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -201,12 +201,11 @@ private async Task<AggregateCrossTargetProjectContext> CreateProjectContextAsync
             ImmutableDictionary<ITargetFramework, ITargetedProjectContext>.Builder innerProjectContextsBuilder = ImmutableDictionary.CreateBuilder<ITargetFramework, ITargetedProjectContext>();
             ITargetFramework activeTargetFramework = TargetFramework.Empty;
 
-            foreach (KeyValuePair<string, ConfiguredProject> kvp in configuredProjectsMap)
+            foreach ((string tfm, ConfiguredProject configuredProject) in configuredProjectsMap)
```
davkean: How does this work? Where is the deconstruction method coming from?

jmarolf: KeyValuePair [already has a](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.keyvaluepair-2.deconstruct?view=netcore-2.1&viewFallbackFrom=netframework-4.7.2) deconstruction method in .NET Framework

davkean: That's only .NET Core 2.0/2.1. I can't see that in .NET Framework.

davidwengier: Looks like Drew added it in his second week: https://github.com/dotnet/project-system/blob/29b4f9c4f2378aa9b20a1b6e269f4ca177f0fc45/src/Microsoft.VisualStudio.ProjectSystem.Managed/Composition/DeconstructionExtensions.cs#L7

The long con...

drewnoakes: Ha, I forgot adding that.

davkean: Ah that makes sense.

drewnoakes:
I broke GitHub. This repo contains a merge commit, but this PR's status is unmerged. It's stuck on this:

image

Will close it just to clean up.

drewnoakes:
I broke GitHub. This repo contains a merge commit, but this PR's status is unmerged. It's stuck on this:

image

PR #4143, opened 2018-10-18
Dependency node grooming

### LINKED ISSUE #4140: Coordination of updates in DependenciesProjectTreeProvider
Whenever a dependency's snapshot changes, `DependenciesProjectTreeProvider` queues up an async `BuildTreeForSnapshotAsync` call, serialising operations with a lock and task:

https://github.com/dotnet/project-system/blob/730c8bb5672b28618017f4e68d31466a1d08f372/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs#L367-L383

However `BuildTreeForSnapshotAsync` is not actually async. It always returns `Task.CompletedTask`:

https://github.com/dotnet/project-system/blob/730c8bb5672b28618017f4e68d31466a1d08f372/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs#L387-L412

It does, however, kick off some async work, but ignores the task.

The other call to `SubmitTreeUpdateAsync` (during initialisation) doesn't wait on the returned task either.

I believe the current handling (`_treeUpdateLock`, `_treeUpdateQueueTask` and JTF) can be replaced with a lock-free cancellation token sequence, relying on `SubmitTreeUpdateAsync` to do the serialisation.

I'll put a PR together with these changes. If I'm barking up the wrong tree, please let me know.

drewnoakes:
Fixes #4140.

More changes made while reading through dependency node code. Easier to review commit-by-commit.

Starts out with removing unused code, tweaking documentation and some formatting.

I focussed on `DependenciesProjectTreeProvider`. It converts dependency snapshots into tree updates for CPS, among other things.

Some private methods were converted to local functions (as discussed over breakfast in Redmond with @davkean and @davidwengier). I generally find that if a private method is only used in one other member, it's a great candidate for conversion to a local function, and doing so often helps identify further opportunities. Unfortunately these transformations wreak havoc on the overall diff, hence being pulled out to separate commits: 7a46b4d6e5556d427656ee149e2dab92c9f8d3ce, 28fa3538b7b6fe36e5a8854f18a27d49c99d92bd, 4983dd77509b560ed169ed89c9774d9a3996b9d4, 4229d747b8f45ab7953d3d3a349e05042def5b4c. 

This lead to the observations discussed in #4140 and the following meatier changes:

- ea1d2e42080aba2fce3406aba35b1d8803178dfe, in which `BuildTreeForSnapshotAsync` is stripped of its false async-ness
- 13c190108056a07c9239d3b0fcd88ab52b93383b, in which the lock, sequencing of tasks and calls through JTF are removed as the code becomes synchronous (it was anyway, in reality, with some tail yielding via schedulers)
- 45ad3fd5c3662d2e37694a4a4fe2c53425b018c3 in which calls to CPS's `SubmitTreeUpdateAsync` are provided a `CancellationToken` that is cancelled as soon as the next update is created

This ends up using CPS's facility for cancellation of async updates, rather than something in the `DependenciesProjectTreeProvider` class itself.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -47,9 +47,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
             IProjectThreadingService threadingService,
             UnconfiguredProject project,
             IDependenciesSnapshotProvider dependenciesSnapshotProvider,
-            [Import(DependencySubscriptionsHost.DependencySubscriptionsHostContract)]
-            ICrossTargetSubscriptionsHost dependenciesHost,
-            [Import(ExportContractNames.Scopes.UnconfiguredProject)]IProjectAsynchronousTasksService tasksService,
+            [Import(DependencySubscriptionsHost.DependencySubscriptionsHostContract)] ICrossTargetSubscriptionsHost dependenciesHost,
+            [Import(ExportContractNames.Scopes.UnconfiguredProject)] IProjectAsynchronousTasksService tasksService,
             IDependencyTreeTelemetryService treeTelemetryService)
             : base(threadingService, project)
         {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -395,28 +395,30 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                     {
                         if (TasksService.UnloadCancellationToken.IsCancellationRequested)
                         {
-                            return;
+                            return Task.CompletedTask;
                         }
 
-                        await BuildTreeForSnapshotAsync();
+                        BuildTreeForSnapshot();
+
+                        return Task.CompletedTask;
                     }).Task;
                 }
                 else
                 {
                     _treeUpdateQueueTask = _treeUpdateQueueTask.ContinueWith(
-                        t => BuildTreeForSnapshotAsync(), TaskScheduler.Default);
+                        t => BuildTreeForSnapshot(), TaskScheduler.Default);
                 }
             }
 
             return;
 
-            Task BuildTreeForSnapshotAsync()
+            void BuildTreeForSnapshot()
             {
                 Lazy<IDependenciesTreeViewProvider, IOrderPrecedenceMetadataView> viewProvider = ViewProviders.FirstOrDefault();
 
                 if (viewProvider == null)
                 {
-                    return Task.CompletedTask;
+                    return;
                 }
 
                 Task<IProjectVersionedValue<IProjectTreeSnapshot>> nowait = SubmitTreeUpdateAsync(
@@ -434,8 +436,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                         // way of merging, mute them for now and get to it in U1
                         return new TreeUpdateResult(dependenciesNode, false, null);
                     });
-
-                return Task.CompletedTask;
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -326,9 +326,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
 
                         lock (SyncObject)
                         {
+                            Verify.NotDisposed(this);
+
                             DependenciesSnapshotProvider.SnapshotChanged += OnDependenciesSnapshotChanged;
 
-                            Verify.NotDisposed(this);
                             Task<IProjectVersionedValue<IProjectTreeSnapshot>> nowait = SubmitTreeUpdateAsync(
                                 (treeSnapshot, configuredProjectExports, cancellationToken) =>
                                 {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -339,7 +339,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                                     //                                                      cancellationToken);
 
                                     return Task.FromResult(new TreeUpdateResult(dependenciesNode, true));
-                                });
+                                },
+                                GetNextTreeUpdateCancellationToken());
                         }
 
                     },
@@ -380,6 +381,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
         private void OnDependenciesSnapshotChanged(object sender, SnapshotChangedEventArgs e)
         {
             IDependenciesSnapshot snapshot = e.Snapshot;
+
             if (snapshot == null)
             {
                 return;
@@ -417,7 +419,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                         // TODO We still are getting mismatched data sources and need to figure out better 
                         // way of merging, mute them for now and get to it in U1
                         return new TreeUpdateResult(dependenciesNode, false, null);
-                    });
+                    },
+                    GetNextTreeUpdateCancellationToken());
             }
         }
 
@@ -541,6 +544,45 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
 
                 return NamedCatalogs;
             }
+        }
+
+        #endregion
+
+        #region Tree update cancellation
+
+        private CancellationTokenSource _treeUpdateCancellationTokenSource = new CancellationTokenSource();
+
+        private CancellationToken GetNextTreeUpdateCancellationToken()
+        {
+#pragma warning disable CA2000 // Dispose objects before losing scope
+            var nextSource = new CancellationTokenSource();
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+            CancellationToken nextToken = nextSource.Token;
+            CancellationTokenSource priorSource = Volatile.Read(ref _treeUpdateCancellationTokenSource);
+
+            // Loop until we succeed in swapping a known prior for our next.
+            // This ensures a proper sequence is observed, which is important
+            // as otherwise we may cancel a prior twice, which would coincide
+            // with a prior not being cancelled at all. We need a strict order
+            // to maintain one-in-one-out.
+            while (true)
+            {
+                CancellationTokenSource snapshot = Interlocked.CompareExchange(
+                    ref _treeUpdateCancellationTokenSource, nextSource, priorSource);
+
+                if (ReferenceEquals(snapshot, priorSource))
+                {
+                    break;
+                }
+
+                priorSource = snapshot;
+            }
+
+            priorSource.Cancel();
+            priorSource.Dispose();
+
+            return nextToken;
         }
 
         #endregion
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -328,6 +328,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                         {
                             Verify.NotDisposed(this);
 
+                            // Issue this token before hooking the SnapshotChanged event to prevent a race
+                            // where a snapshot tree is replaced by the initial, empty tree created below.
+                            // The handler will cancel this token before submitting its update.
+                            CancellationToken initialTreeCancellationToken = GetNextTreeUpdateCancellationToken();
+
                             DependenciesSnapshotProvider.SnapshotChanged += OnDependenciesSnapshotChanged;
 
                             Task<IProjectVersionedValue<IProjectTreeSnapshot>> nowait = SubmitTreeUpdateAsync(
@@ -341,7 +346,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
 
                                     return Task.FromResult(new TreeUpdateResult(dependenciesNode, lazyFill: true));
                                 },
-                                GetNextTreeUpdateCancellationToken());
+                                initialTreeCancellationToken);
                         }
 
                     },
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -392,36 +392,29 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                 return;
             }
 
-            BuildTreeForSnapshot();
+            Lazy<IDependenciesTreeViewProvider, IOrderPrecedenceMetadataView> viewProvider = ViewProviders.FirstOrDefault();
 
-            return;
-
-            void BuildTreeForSnapshot()
+            if (viewProvider == null)
             {
-                Lazy<IDependenciesTreeViewProvider, IOrderPrecedenceMetadataView> viewProvider = ViewProviders.FirstOrDefault();
-
-                if (viewProvider == null)
-                {
-                    return;
-                }
-
-                Task<IProjectVersionedValue<IProjectTreeSnapshot>> nowait = SubmitTreeUpdateAsync(
-                    async (treeSnapshot, configuredProjectExports, cancellationToken) =>
-                    {
-                        IProjectTree dependenciesNode = treeSnapshot.Value.Tree;
-                        if (!cancellationToken.IsCancellationRequested)
-                        {
-                            dependenciesNode = await viewProvider.Value.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
-
-                            _treeTelemetryService.ObserveTreeUpdateCompleted(snapshot.HasUnresolvedDependency);
-                        }
-
-                        // TODO We still are getting mismatched data sources and need to figure out better 
-                        // way of merging, mute them for now and get to it in U1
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -65,7 +65,16 @@ internal class DependenciesProjectTreeProvider :
             TasksService = tasksService;
             _treeTelemetryService = treeTelemetryService;
 
+            // Hook this so we can unregister the snapshot change event when the project unloads
             unconfiguredProject.ProjectUnloading += OnUnconfiguredProjectUnloading;
+
+            Task OnUnconfiguredProjectUnloading(object sender, EventArgs args)
```
davidwengier: Is there a benefit (or cost?) to not changing this to be a normal void returning event handler?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -602,35 +501,88 @@ public async Task<IRule> GetRuleAsync(IDependency dependency, IProjectCatalogSna
...
+
+        #endregion
+
+        #region Tree update cancellation
+
+        private CancellationTokenSource _treeUpdateCancellationTokenSource = new CancellationTokenSource();
+
+        private CancellationToken GetNextTreeUpdateCancellationToken()
+        {
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
jmarolf: Can you add a more descriptive comment about how we don't need to dispose this because `nextSource` becomes  `priorSource`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -427,95 +357,70 @@ private IProjectTree CreateDependenciesFolder(IProjectTree oldNode)
...
+                    browseObjectProperties: null,
+                    icon: values.Icon,
+                    expandedIcon: values.ExpandedIcon,
+                    visible: true,
+                    flags: values.Flags);
             }
         }
 
-        private async Task<IImmutableDictionary<string, IPropertyPagesCatalog>> GetNamedCatalogsAsync(IProjectCatalogSnapshot catalogs)
+        private void OnDependenciesSnapshotChanged(object sender, SnapshotChangedEventArgs e)
```
jmarolf: Do we have any tests that will fail if the order changes are processed is not correct?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
```diff
@@ -144,12 +144,7 @@ public IDependency FindDependency(string id, bool topLevel = false)
 
         public bool Equals(IDependenciesSnapshot other)
         {
-            if (other != null && other.ProjectPath.Equals(ProjectPath, StringComparison.OrdinalIgnoreCase))
-            {
-                return true;
-            }
-
-            return false;
+            return other != null && other.ProjectPath.Equals(ProjectPath, StringComparison.OrdinalIgnoreCase);
```
davidwengier: I'm curious, whats your personal rule for when to use expression bodies and not?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: This is required per CPS's public API. It consumes tasks from all handlers and awaits them in parallel.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
drewnoakes: There are few hard and fast rules in programming. I use them for readability, which is very subjective. In a tie break, I favour brevity. Any opinions here are loosely held.

I suspect I would use them more liberally than others on the team, so I'm exercising restraint. This PR changes `Dependency.Equals` to expression form, yet this one remains in block form. I would convert this on my own project, absent any restraint. I'd convert several other members on `DependenciesSnapshot`.

What about you/others?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -325,71 +326,78 @@ async delegate
...
                     },
                     registerFaultHandler: true);
             }
-        }
 
-        private Task OnUnconfiguredProjectUnloading(object sender, EventArgs args)
-        {
-            UnconfiguredProject.ProjectUnloading -= OnUnconfiguredProjectUnloading;
-            DependenciesSnapshotProvider.SnapshotChanged -= OnDependenciesSnapshotChanged;
+            IProjectTree CreateDependenciesFolder()
```
abpiskunov: >IProjectTree CreateDependenciesFolder() [](start = 12, length = 39)

I don't know what discussion you guys had over breakfast, but these local functioned code is harder to read .... :)

abpiskunov: What exactly is the new convention and what are benefits?

---
In reply to: [226698249](https://github.com/dotnet/project-system/pull/4143#discussion_r226698249) [](ancestors = 226698249)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
abpiskunov: In general it is hard to test such things via unit tests and i was going to suggest you to hold on your refactorings and create first a set of E2E integration tests(Apex) + stress tests, which would shield you from regressions , especially for things like here which can have timing issues.

---
In reply to: [226568463](https://github.com/dotnet/project-system/pull/4143#discussion_r226568463) [](ancestors = 226568463)

abpiskunov: All you changes look good, however they are basically untested, since manual sanity verification does not count for many scenarios.

---
In reply to: [226698874](https://github.com/dotnet/project-system/pull/4143#discussion_r226698874) [](ancestors = 226698874,226568463)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -405,118 +413,9 @@ private Task BuildTreeForSnapshotAsync(IDependenciesSnapshot snapshot)
...
-                        ReferencesProjectTreeCustomizablePropertyValues values)
-        {
-            foreach (IProjectTreePropertiesProvider provider in ProjectTreePropertiesProviders.ExtensionValues())
-            {
-                provider.CalculatePropertyValues(context, values);
-            }
+                    return new TreeUpdateResult(dependenciesNode, lazyFill: false);
+                },
+                GetNextTreeUpdateCancellationToken());
         }
```
abpiskunov: As i mentioned in the github issue, this change is wrong. Before we would have a queue that would schedule next SubmitTreeUpdateAsync only after all previous ones were called. When now OnDependenciesSnapshotChanged can be called after 2 different DT builds/evaluations carrying conflicting changes and if timing will be wrong OnDependenciesSnapshotChanged will be called for them in wrong order which would result in sporadic tree behavior in solution explorer with tree showing wrong dependencies and their states..

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -602,35 +495,91 @@ public async Task<IRule> GetRuleAsync(IDependency dependency, IProjectCatalogSna
...
             {
-                if (dependency.Resolved)
+                return configuredProjectExports.RuleFactory.CreateResolvedReferencePageRule(
+                    schema,
+                    context,
+                    dependency.Name,
+                    dependency.Properties);
+            }
+
+            return browseObjectsCatalog.BindToContext(schema.Name, context);
```
abpiskunov: conditional operator?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: Readability is pretty subjective, so let me explain how I see the benefit.

The more top level members a type has, the more potential dependencies there are between them. I find this makes it harder to reason about invariants and usage of members.

By nesting members within other members, it communicates (enforces) separation. These aggregate members are easier to understand in isolation. You can also see the domain of values exchanged and other invariants, which often highlights opportunities to simplify things, IME.

Complexity with local functions can come about when you capture and mutate many variables. There's talk of adding `static` local functions that can't capture, for this reason. For now it's a good idea to avoid capturing in complex ways, especially if mutating.

Introducing local functions where none previously existed can increase readability. Function names are documentation.

This style of nesting is familiar to me from work in other languages where I grew to really like it. I encourage folks to give it a go. I'm doing a bunch of work on the dependency node at the moment and will limit such changes to that area of the code for now.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: Apex sounds interesting, I'll look into it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
abpiskunov: Snapshot1 (assemblies) and Snapshot2 (assemblies+projects) come very quickly one after another, so quick that OnDependenciesSnapshotChanged executes simultaneously. Which one would first reach line 403?

---
In reply to: [226705432](https://github.com/dotnet/project-system/pull/4143#discussion_r226705432) [](ancestors = 226705432)

abpiskunov: before i had a lock at the begining of the method that would guaranty first come first serve

---
In reply to: [226714600](https://github.com/dotnet/project-system/pull/4143#discussion_r226714600) [](ancestors = 226714600,226705432)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
abpiskunov: still confusing.. separating code to a private function is done to reuse some functionality and it is fine if other class members call it. Yes if it is not done right then it could lead to some harder to follow logic that might change the state of the object. But that can b e said about anything.. 

On the other side to me if the method has 20+ lines it is already too complicated and local functions making it a little harder to read. Anyway, it is subjective as you said .. 

---
In reply to: [226707589](https://github.com/dotnet/project-system/pull/4143#discussion_r226707589) [](ancestors = 226707589)

abpiskunov: i will try it out and will see if i get used to it :)

---
In reply to: [226717284](https://github.com/dotnet/project-system/pull/4143#discussion_r226717284) [](ancestors = 226717284,226707589)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
abpiskunov: Ping me offline, i will send you a TFS link to my integration tests for nesting , they do check the tree and i have a stress test there too. IT is actually pretty handy to have.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: But it's still first come, first served, whether you add the lock or not. Even with a lock, one thread could start, switch out before the lock, and another could race past and get the lock.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: Oh, one more thing about local functions&mdash;it sometimes means you can push state "down" in scope, from field to local.

I guess the overall benefit I see is restricting the scope of things.

If a local function becomes useful in another scenario, it can be lifted out.

A pattern I like (and I think it's used in Roslyn) is to make sure all local functions are at the end of the method, and if the method has `void` return type, to include a redundant `return;` before the first local function, just to make it clear that the top level method is finished.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: Thanks. Message sent on Teams.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
abpiskunov: My point was when two threads execute OnDependenciesSnapshotChanged it could be pretty random who first would hit the line 403 and call CPS SubmitTreeUpdateAsync, because of logic in previous lines in that method event though they are quick. If you feel there will be no timing, its ok try it out. However i would add stress tests to verify snapshots are applied right, otherwise  it would be hard to investigate ven if customers open issues.

---
In reply to: [226736814](https://github.com/dotnet/project-system/pull/4143#discussion_r226736814) [](ancestors = 226736814)

abpiskunov:
>                         _treeTelemetryService.ObserveTreeUpdateCompleted(snapshot.HasUnresolvedDependency);

Btw, unrelated to this PR, i remember that telemetry calculation logic was kind of controversial at the time. I think it would be nice if you would look at it with your fresh eye and see if it is does what we would expect from it and fires telemetry events reliably.

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs:411 in e9642f8. [](commit_id = e9642f838a0a6db9f8f7b38a4ce3094ed4cb015e, deletion_comment = False)

drewnoakes:
> telemetry calculation logic 

Sure, I'll take a look at it. I haven't done so yet.

BTW what tool are you using to review? I've seen those `[](...)` comments a fair bit lately and don't know where they're coming from.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: I'll follow the data backwards and see whether the code that's firing the dependencies `SnapshotChanged` event is serialising invocations or not.

abpiskunov:
I use CodeFlow . You can install CodeFLow plugin in chrome and it would appear as a link in Guthub PRs. It is much easier to review difs in CodeFlow than on github website , but again it might be subjective :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
davidwengier: > This PR changes `Dependency.Equals` to expression form, yet this one remains in block form

That's why I asked, was curious about the reasoning for this one versus that one.

I'm just trying to get an idea of where everyone's preference sits wrt this sort of thing.  My personal habits steer away from brevity when I'm writing, but I have no problem with reading it, so I don't really know where I stand ¯\\\_(ツ)_/¯

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -37,35 +37,35 @@ internal class DependenciesProjectTreeProvider :
...
+        private Task _treeUpdateQueueTask;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DependenciesProjectTreeProvider"/> class.
         /// </summary>
         [ImportingConstructor]
         public DependenciesProjectTreeProvider(
             IProjectThreadingService threadingService,
-            UnconfiguredProject project,
+            UnconfiguredProject unconfiguredProject,
```
Pilchie: @davkean didn't you have a preferred naming convention here to follow along with CPS?

drewnoakes: In this instance I made the parameter name match the base ctor's.

PR #4146, opened 2018-10-19
Add try/catch around setting up the file watcher

BillHiebert:
This handles the case where the project folder path goes invalid during a branch switch while the project is loading. This addresses the following feedback issue: https://developercommunity.visualstudio.com/content/problem/361196/project-system-encounters-a-fault-after-a-branch-s.html

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
@@ -724,7 +724,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
                 }
                 catch (Exception ex)  when (ex is IOException || ex is ArgumentException)
                 {
-                    // IF the project folder is no longer available this will throw, which can happen during branch switching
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -712,13 +712,20 @@ private void WatchLaunchSettingsFile()
...
+                    FileWatcher = new SimpleFileWatcher(Path.GetDirectoryName(CommonProjectServices.Project.FullPath),
+                                                        true,
+                                                        NotifyFilters.FileName | NotifyFilters.Size | NotifyFilters.LastWrite,
+                                                        LaunchSettingsFilename,
+                                                        LaunchSettingsFile_Changed,
+                                                        LaunchSettingsFile_Changed);
+                }
+                catch (Exception ex)  when (ex is IOException || ex is ArgumentException)
+                {
+                    // IF the project folder is no longer available this will throw, which can happen during branch switching
```
Pilchie: Nit: Capitalization of `IF`

BillHiebert: fixed

Issue #4105, opened 2018-10-11
Clean up rules


davidwengier:
- [ ] We don't need BrowseObject rules to have properties that legacy doesn't expose. This will save memory usage as we won't be moving around those values. First task is to determine exactly which properties are exposed by legacy, so we can match. See comment for more info: https://github.com/dotnet/project-system/issues/2966#issuecomment-367222311
- [x] We should also check as part of this and make sure all of the rules are consistent across None/Content/Compile etc. These files should be identical apart from their Page.Name attribute, and the value of DataSource.ItemType. (~#4635~)
- [ ] Consistently apply DoNotCopyAcrossProjects
- [x] Clean up space, and layout of the elements attributes, so that it is identical across all rules - makes it easier to diff/read. (~#4620~)
- [ ] Add PersistenceStyle="Attribute" at the data source level for the rule
- [ ] Figure out default properties (such as datasource for a property, and see if you need to restate them - if not remove them).
- [ ] Consistently name all rules; RuleName.xaml, or RuleName.BrowseObject.xaml
- [x] Move rules into sub directories for different types to help find them, one for "Items", "Project", etc (~#4618~)
- [x] Finally we don't need localisable strings on non-Visible properties, since they're not shown to the user anyway (~#4270~)
- [ ] Ensure all visible browseobject properties have display names (and descriptions?)
- [ ] Write tests to ensure we don't regress any of the above (started with ~#4641~) 
- [x] Test for xaml files included in CommonFiles.swr (~#4967~)
- [ ] Test for xaml files in design time targets (and neutral directory as appropriate?) (from #4775)
- [ ] We currently have CustomTool and Generator properties in some cases - do we need both?

davidwengier:
See this comment for info about which properties are necessary: https://github.com/dotnet/project-system/issues/2966#issuecomment-367222311

davidwengier:
As for the first checklist item in this issue, I looked into the BrowseObject rules and all of them seem fine as far as only exposing things that legacy has. Seems to me that the things that do have properties that legacy doesn't expose is things like `SdkReference.xaml` so we'd have to split all of those into their separate parts (eg that one is `Project`, `ProjectSubscriptionService` and `BrowseObject`). Can you confirm my understanding there is correct @davkean ?

davkean:
The easiest way to find underused properties, it to open VS Telemetry monitor and look for the `vs/projectsystem/cps/rulepropertiesusage` event, this tells you what properties are unused:

For example:

```
[...]

PR #4148, opened 2018-10-22
Rule cleanup

davidwengier:
Part of #4105. This splits up 3 more rule types, and removes localizable strings from hidden properties.

When this is merged it means all of the file rules are separated and minimal.


src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -344,4 +350,18 @@
   <ItemGroup>
     <DesignTimeTargetsFile Include="ProjectSystem\DesignTimeTargets\*.targets" />
   </ItemGroup>
+  <ItemGroup>
+    <None Update="ProjectSystem\Rules\ApplicationDefinition - Copy.xaml">
```
Pilchie: This doesn't look right :)

davidwengier: Ugh.. I specifically did the file duplication out of VS to stop it doing this, must have switched back too quickly.

Issue #4080, opened 2018-10-02
Razor language service initialization


davkean:
- [ ] IProjectEvaluationHandler that takes Content files and pushes them to IWorkspaceProjectContext 

- [ ] API for blocking on us pushing through a configured project version to the language service. 
     - Razor is going to be blocking on design-time build via OutputGroups anyway, so really only needed for correctness

davkean:
Currently blocked on IWorkspaceProjectContext new overloads

PR #4152, opened 2018-10-23
Apply/suppress new analyzers warnings

davkean:
As a part of https://github.com/dotnet/project-system/issues/4080, we need to update our Roslyn bits which also updates analyzers. Address, suppress or baseline new warnings.

If the fix was simple without any thought I just fixed it. If it was a result of a bug in the rule, I suppressed it. If it required thought to fix, I baselined along with the rest to be addressed in https://github.com/dotnet/project-system/issues/121.


src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsValueSerializer.vb
```diff
@@ -11,9 +11,12 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
     ''' a culture info.
     ''' </summary>
     ''' <remarks></remarks>
-    Friend Class SettingsValueSerializer
+    Friend NotInheritable Class SettingsValueSerializer
```
davidwengier: I assume this can't be shared because of one of the InternalsVisibleTo attributes? If so, how does the existing changes not break that scenario already?

davkean: VB doesn't have an equivalent of "static" classes. They have modules, but that results in VB importing those methods for usage without class prefix everywhere.

davidwengier: Thank you, I suspect this won't be the last time I forget that :)

PR #4164, opened 2018-10-24
IDependenciesSnapshotFilter.BeforeRemove returns bool

drewnoakes:
Unlike `IDependenciesSnapshotFilter.BeforeAdd`, it doesn't make sense for `IDependenciesSnapshotFilter.BeforeRemove` to modify the dependency. It either removes the dependency or does not.

Updated documentation to hopefully make the meaning of return and `out` vars clearer.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
@@ -39,7 +39,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.Fil
             IDependency dependency,
             ImmutableDictionary<string, IDependency>.Builder worldBuilder,
             ImmutableHashSet<IDependency>.Builder topLevelBuilder,
-            Dictionary<string, IProjectDependenciesSubTreeProvider> subTreeProviders,
+            IReadOnlyDictionary<string, IProjectDependenciesSubTreeProvider> subTreeProviders,
             HashSet<string> projectItemSpecs,
             out bool filterAnyChanges);
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
@@ -40,7 +40,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.Fil
             ImmutableDictionary<string, IDependency>.Builder worldBuilder,
             ImmutableHashSet<IDependency>.Builder topLevelBuilder,
             IReadOnlyDictionary<string, IProjectDependenciesSubTreeProvider> subTreeProviders,
-            HashSet<string> projectItemSpecs,
+            IImmutableSet<string> projectItemSpecs,
             out bool filterAnyChanges);
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/SdkAndPackagesDependenciesSnapshotFilter.cs
```diff
@@ -87,12 +87,10 @@ internal class SdkAndPackagesDependenciesSnapshotFilter : DependenciesSnapshotFi
...
             filterAnyChanges = false;
-            if (!dependency.TopLevel || !dependency.Resolved)
-            {
-                return dependency;
-            }
 
-            if (dependency.Flags.Contains(DependencyTreeFlags.PackageNodeFlags))
+            if (dependency.TopLevel && 
+                dependency.Resolved && 
+                dependency.Flags.Contains(DependencyTreeFlags.PackageNodeFlags))
```
Pilchie: The side-effects below are a bit icky :(

drewnoakes: Via the builders? In this case it's the design. The add/remove filters can manipulate the builders, as well as reject the dependency or (in the case of add) modify it. Maybe the name _filter_ is misleading if you think of it as a pure predicate. Not sure what a better name would be.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
```diff
@@ -42,16 +44,16 @@ internal interface IDependenciesSnapshotFilter
...
         /// <param name="targetFramework">Target framework for which dependency was resolved.</param>
         /// <param name="dependency">The dependency to which filter should be applied.</param>
         /// <param name="worldBuilder">Builder for immutable world dictionary of updating snapshot.</param>
         /// <param name="topLevelBuilder">Top level dependencies list builder of updating snapshot.</param>
-        /// <param name="filterAnyChanges">True if filter did any changes in the snapshot</param>
-        /// <returns>Dependency itself if removal is approved, null if dependency should not be removed</returns>
-        IDependency BeforeRemove(
+        /// <param name="filterAnyChanges"><see langword="true"/> if <paramref name="worldBuilder"/> or <paramref name="topLevelBuilder"/> changed, otherwise <see langword="false"/>.</param>
+        /// <returns><see langword="true"/> if removal is approved, or <see langword="false"/> if <paramref name="dependency"/> should not be removed.</returns>
+        bool BeforeRemove(
```
jmarolf: Would `CanRemove` make more sense as a name? I get that `BeforeRemove` is supposed to imply that the filter could modify things before removal but seems like a thing we don't want to happen anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/SdkAndPackagesDependenciesSnapshotFilter.cs
Pilchie: Yes, I meant because filter implies a pure predicate.  Not sure of a better name either.

davkean:
@jmarolf What's going on with the builds?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
davkean: Agreed, CanRemove, ShouldExclude, Filter, IsIgnorable, IsExcludable

jmarolf:
@davkean odd, looks like there are plenty of agents.  going to close and re-open

drewnoakes:
I pushed two more commits to make parameters of `IDependenciesSnapshotFilter.BeforeAdd` immutable, making it clearer which parameters implementations are expected to modify.

As for the name, I think `BeforeAdd` and `BeforeRemove` are as good as any other name proposed so far. It's the `Filter` name that's causing confusion. Implementations are used in a pipeline, or chain of responsibility, similar to how an HTTP request might be passed along a series of middleware.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IDependenciesSnapshotFilter.cs
drewnoakes: These "filters" are (currently) intended to observe dependencies being added and removed in order to make other state consistent. If we keep that behaviour, then `Filter` is the misleading bit.

drewnoakes:

PR #4166, opened 2018-10-24
More dependency node grooming

drewnoakes:
Follows on from #4143, containing several small thematic commits with changes made while walking through the rest of the dependencies code.
 
Easiest to review commit by commit. I tried splitting this up after the fact but the changes interleave and conflict.

Particularly interesting commits:

- b34f5001ef06a99d9bf329a5c1f372eda2963706 which is similar to ea1d2e42080aba2fce3406aba35b1d8803178dfe from #4143 which addressed issue #4140 
- e4728901c257900b76c29e93424b35f3faeeeea3, af0d22ca981eef58a96a0cda9fa400ebeed894f0 and bcf32fbf06a943b28819f39aa37853ebad3cee07 which uses pooled `StringBuilder`s more efficiently
- 42de3fa5f181b6034a57eb6461b578a3bfe426e0 which significantly reduces string allocations in `DependenciesSnapshot.FindDependency`

I have a question about style for the team off the back of this PR. Several types have private properties as well as private fields. Historically I have preferred making private data fields, and using properties for non-private. I find it consuming code clearer. Is there some other pattern at play here, or do we have a mixture of different styles? See 3f2d7d7e7054c5502227c7187a4bb6088c7ae8b0 for an example.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -196,7 +196,7 @@ private void TrackChanges(SnapshotChangedEventArgs updatedProjectContext)
 
             var actionHandlers = GraphActionHandlers.Select(x => x.Value).Where(x => x.CanHandleChanges()).ToList();
 
-            if (!actionHandlers.Any())
+            if (actionHandlers.Count == 0)
```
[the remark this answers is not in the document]
drewnoakes: Probably depends upon the version of the framework. I F12'd into decompiled source and saw it invoke the enumerator. Some Linq operators do special case collection types, others don't.

drewnoakes:
@Pilchie thanks for the review. I'll get to your other comments tomorrow.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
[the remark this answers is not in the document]
Pilchie: See https://github.com/dotnet/corefx/issues/32569

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -151,10 +160,10 @@ internal class TelemetryState
         {
             private readonly ConcurrentDictionary<string, bool> _observedRules = new ConcurrentDictionary<string, bool>(StringComparers.RuleNames);
 
-            internal bool InitializeRule(string rule) =>
+            internal void InitializeRule(string rule) =>
```
jmarolf: Why this change?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -340,9 +339,6 @@ private static GraphNodeId GetTopLevelGraphNodeId(string projectPath, string mod
                 GraphNodeId.GetPartial(CodeGraphNodeIdName.File, filePath));
         }
 
-        private static string GetIconStringName(ImageMoniker icon)
-        {
-            return $"{icon.Guid.ToString()};{icon.Id}";
-        }
+        private static string GetIconStringName(ImageMoniker icon) => $"{icon.Guid:D};{icon.Id}";
```
davidwengier: Since D is the default for Guid.ToString() it might be better to remove the formatter specifier to be sure the compiler optimizes this to string concatenation

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/GraphViewProviderBase.cs
```diff
@@ -145,7 +145,8 @@ public virtual bool ShouldTrackChanges(string projectPath, string updatedProject
         {
             nodesToRemove = existingChildren.Except(updatedChildren).ToList();
             nodesToAdd = updatedChildren.Except(existingChildren).ToList();
-            return nodesToAdd.Any() || nodesToRemove.Any();
+
+            return nodesToAdd.Count != 0 || nodesToRemove.Count != 0;
```
davidwengier: I wonder if in this case keeping the `.Any()` and removing the `.ToList()` might actually be better overall? Depends on how big the incoming lists are likely to be and how often there is expected to be changes, which is info I don't have.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/GraphViewProviderBase.cs
```diff
@@ -126,29 +126,29 @@ public virtual bool ShouldTrackChanges(string projectPath, string updatedProject
...
-            out IEnumerable<DependencyNodeInfo> nodesToRemove)
+            out IReadOnlyList<DependencyNodeInfo> nodesToAdd,
+            out IReadOnlyList<DependencyNodeInfo> nodesToRemove)
         {
-            IEnumerable<DependencyNodeInfo> existingChildrenInfo = GetExistingChildren(dependencyGraphNode);
+            IReadOnlyList<DependencyNodeInfo> existingChildrenInfo = GetExistingChildren(dependencyGraphNode);
             IEnumerable<IDependency> updatedChildren = targetedSnapshot.GetDependencyChildren(updatedDependency)
                 ?? Enumerable.Empty<IDependency>();
-            IEnumerable<DependencyNodeInfo> updatedChildrenInfo = updatedChildren.Select(x => DependencyNodeInfo.FromDependency(x));
+            IReadOnlyList<DependencyNodeInfo> updatedChildrenInfo = updatedChildren.Select(x => DependencyNodeInfo.FromDependency(x)).ToList();
```
davidwengier: There could also be an argument for removing the `.ToList()` in the `AnyChanges` method and here - could prevent any construction of DependencyNodeInfo if the circumstances are right. I'm just thinking out loud though mainly.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -104,10 +104,7 @@ protected override Task DisposeCoreAsync(bool initialized)
         /// </summary>
         public void BeginGetGraphData(IGraphContext context)
         {
-            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
-            {
-                await BeginGetGraphDataAsync(context);
-            });
+            ThreadHelper.JoinableTaskFactory.RunAsync(() => BeginGetGraphDataAsync(context));
```
davidwengier: This reminds me, I saw [this guidance](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task) the other day and thought it was interesting after our discussions about this in the past.

For the record I support this type of change :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/ProjectRuleHandler.cs
```diff
@@ -37,19 +37,35 @@ internal class ProjectRuleHandler : DependenciesRuleHandlerBase
...
-            AggregateSnapshotProvider.SnapshotChanged += OnSnapshotChanged;
-            AggregateSnapshotProvider.SnapshotProviderUnloading += OnSnapshotProviderUnloading;
+            aggregateSnapshotProvider.SnapshotChanged += OnAggregateSnapshotChanged;
+            aggregateSnapshotProvider.SnapshotProviderUnloading += OnAggregateSnapshotProviderUnloading;
 
-            CommonServices.Project.ProjectUnloading += OnUnconfiguredProjectUnloading;
+            // Unregister event handlers when the project unloads
+            commonServices.Project.ProjectUnloading += OnUnconfiguredProjectUnloading;
+
+            Task OnUnconfiguredProjectUnloading(object sender, EventArgs e)
```
davidwengier: Unrelated, I wrote a local function the other day and even in VS with proper syntax highlighting I still kinda want to add `// ------------ end of method -------------` just above them :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/AssemblyDependencyModel.cs
```diff
@@ -32,14 +32,10 @@ internal class AssemblyDependencyModel : DependencyModel
...
         {
-            string fusionName = null;
-            if (Properties != null)
-            {
-                Properties.TryGetValue(ResolvedAssemblyReference.FusionNameProperty, out fusionName);
-            }
-
             if (Resolved)
             {
+                string fusionName = null;
```
davidwengier: Nit: Unnecessary initialisation

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GroupedByTargetTreeViewProvider.cs
```diff
@@ -142,33 +142,12 @@ public override IProjectTree FindByPath(IProjectTree root, string path)
...
-                if (string.Equals(node.FilePath, path, StringComparison.OrdinalIgnoreCase))
-                    return node;
-            }
+            IProjectTree dependenciesNode = root.Flags.Contains(DependencyTreeFlags.DependenciesRootNodeFlags) 
+                ? root 
+                : root.GetSubTreeNode(DependencyTreeFlags.DependenciesRootNodeFlags);
 
-            return null;
+            return dependenciesNode?.GetSelfAndDescendentsBreadthFirst()
+                .FirstOrDefault(node => string.Equals(node.FilePath, path, StringComparison.OrdinalIgnoreCase));
```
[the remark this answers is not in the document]
drewnoakes: `GetSelfAndDescendentsBreadthFirst` returns `IEnumerable<IProjectItem>` and internally is implemented using a non-recursive generator function, so it'll walk the tree lazily.

However the general pattern of O(N) scanning (worst case) for items in `FindByPath` might be suboptimal. It's not immediately obvious to me how we'd maintain a lookup here. I'll keep an eye out for it while profiling.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -51,6 +51,9 @@ internal class DependencyTreeTelemetryService : IDependencyTreeTelemetryService
         /// </summary>
         public void InitializeTargetFrameworkRules(ITargetFramework targetFramework, IEnumerable<string> rules)
         {
+            if (_stopTelemetry)
```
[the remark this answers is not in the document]
drewnoakes: These are opportunistic. If the flag's set (it only transitions from `false` to `true` once, within the lock) then we _definitely_ don't need to enter the lock. If it's `false`, we enter the lock and test the flag again just to be sure.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/GraphViewProviderBase.cs
```diff
@@ -147,7 +147,6 @@ public virtual bool ShouldTrackChanges(string projectPath, string updatedProject
             out IEnumerable<DependencyNodeInfo> nodesToAdd,
             out IEnumerable<DependencyNodeInfo> nodesToRemove)
         {
-            var comparer = new DependencyResolvedStateComparer();
             nodesToRemove = existingChildren.Except(updatedChildren).ToList();
             nodesToAdd = updatedChildren.Except(existingChildren).ToList();
```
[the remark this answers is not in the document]
drewnoakes: `DependencyNodeInfo` implements its own `Equals`/`GetHashCode` so it should be fine.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyTreeTelemetryService.cs
drewnoakes: The return values were unused. Advertising them gives the false sense that they do something.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
drewnoakes: Looks like the [compiler emits](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQFEAeAHJzkCWA9gHYBqUCBUwANigDQwgG0A+AAgAwAEHAjAG4AsACgxHAMx8ATDwDCYgN5ieavtI4AWHgFkAFAHE4BACY8A5idMBKVepWj1zngDdKPZPx4BeHgBIAImMzHyUrMwBfQJEnFzV3BE85PyCQ0zCI0xAAEWjY50ixSKA) a call to `System.String::Format` with the `D` specifier in the format string.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/GraphViewProviderBase.cs
drewnoakes: The `ToList` calls support returning `IReadOnlyList<T>` via the `out` params, not just the `Any` calls.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/GraphViewProviderBase.cs
drewnoakes: The `Any` calls would kick off an enumeration to the first item. Doing so creates an `Enumerable.ExceptIterator` which fully consumes one enumerable to create a set to compare against. `TrackChanges` fully enumerates both results too. Without `ToList` we end up allocating redundant sets internally to support `Except`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
davidwengier: Looks like the optimization I was thinking of only applies to strings, so you have to call .ToString() on the guid to get it, and use the C# 8 compiler. Without those things the `:D` makes no difference.

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQGMoBssQaiADwAcllkBLAewDsAaGECrAHwAEAGAAjYEYBuALAAoEWwDMPAExcAwiIDeIrip6S2AFi4BZABQBxOBQAmXAOZHjASmWqlw1Y64A3KAi7JeXALxcAJABEhibeChYmAL4BQg5OKq7uyDK+gcHGoeHGIAAiUTFx8W4ekilBlhmWAHQAKlQAyjAIFDRmulZ5tioRIhFAA==

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
drewnoakes: Thanks for the link, I hadn't seen that.

We have a lot of `return Task.CompletedTask` throughout the code and CPS which goes against this guidance.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/ProjectRuleHandler.cs
drewnoakes: I generally like to add a redundant `return` in these cases. Missed this one, thanks. Added.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/AssemblyDependencyModel.cs
drewnoakes: It's needed here because the call to `TryGetValue` is conditional.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependenciesViewModelFactory.cs
```diff
@@ -44,7 +44,7 @@ public ImageMoniker GetDependenciesRootIcon(bool hasUnresolvedDependencies)
 
         private IProjectDependenciesSubTreeProvider GetProvider(string providerType)
         {
-            return SubTreeProviders.FirstOrDefault(x => x.Value.ProviderType.Equals(providerType)).Value;
+            return SubTreeProviders.First(x => x.Value.ProviderType.Equals(providerType)).Value;
```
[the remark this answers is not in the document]
drewnoakes: This method's so simple that either exception would be meaningful in reality.

I generally favour the non-NRE when possible. My preference comes from trying to guess what threw NRE while looking at a stack trace from a user/customer.

My understanding is the info about what was null is only available in the debugger, not in a stack trace. 

Doesn't `First` say something like "sequence contained no elements"? It will also have `First` at the top of the stack trace, so it's clear it came from a call to `First`. By contrast a NRE can happen almost anywhere, making debugging harder.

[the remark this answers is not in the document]
davidwengier: > My understanding is the info about what was null is only available in the debugger, not in a stack trace.

I'm sure you're right, I have no idea.

My preference is to use *OrDefault and check for null and log/throw a meaningful message, but my experience in this area is mostly from ORMs that tend to have lots of "sequences" in them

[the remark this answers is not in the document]
drewnoakes: > I'm sure you're right

Don't be :) I assume the exception type wasn't extended and that this is a debugger feature.

PR #4170, opened 2018-10-25
Handle more cases when calculating folder names

davkean:
Pull out folder name calculation into separate method and handle more cases. This is needed for Razor support.

Note: This doesn't produce anymore allocations than previously, but now shows at least one place where we can reduce allocations (getting the "project directory" to pass to GetLogicalFolderNames).


Pilchie:
Preview 1 changes require QB approval - can we just take all of them at once?

Issue #4163, opened 2018-10-24
Provide an implementation of BuildManager to produce TempPE assembly during type resolution


davidwengier:
When opening designers the type resolution service makes calls to the projects BuildManager in order to get it to compile design time files into a TempPE assembly. CPS has everything we need to be able to provide our own implementation of this.

- [x] Create DesignTimeInputsDataSource, that produces a set of design time inputs and shared inputs
- [x] Create DesignTimeInputsFileWatcher, that consumes DesignTimeInputsDataSource and watches inputs for changes
- [x] Create TempPECompilerManager that consumes DesignTimeInputsDataSource and DesignTimeInputsFileWatcher and compiles DLLs
- [x] ~Create TempPEBuildManager that consumes DesignTimeInputsDataSource and DesignTimeInputsFileWatcher and fires events~ (Removed: Too much overlap with TempPECompilerManager
- [x] Split TempPECompilerManager into two components, one to produce changes to design time inputs, and one to compile said changes into DLLs (#5231)
- [x] Create DesignTimeInputsBuildManager that consumes the changes datasource from above, and fires events (#5232)
- [x] Wire up BuildManager to call into DesignTimeInputsBuildManager

Design (subject to change):
![image](https://user-images.githubusercontent.com/754264/62267561-d3d79000-b46f-11e9-9707-033a9cf59cfe.png)

Relates to [AB#825236](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/825236)

davidwengier:
In a project with just over 1000 `.cs` files, one of which is a design time input and one of which is a shared design time input, mutating the solution state to compile just those two files is only 5.8% of the CPU time spent, compared to 94.2% doing the actual compilation.

davidwengier:
This is related to AB#825236

dmonroym:
@davidwengier interestingly enough when you mention the work item in a comment it doesn't add the link in the AzDo side. This means that we have to add it either in the description (when it's a PR or issue) or on the commit message.

davidwengier:
It did end up adding a link about 2 hours ago. Azure was just busy yesterday :)

dmonroym:
I actually added the comment on the description, that's why it got linked 😆

davidwengier:
Oh, thank you! GitHub must not notify about edits I guess.

PR #4173, opened 2018-10-25
Implement build manager

davidwengier:
This is the first two checklist items in #4163. Thought I'd start PRing early to make sure everything is as expected. The [current CPS BuildManager](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.VS.Implementation%2FPackage%2FAutomation%2FVSProject%2FOAVSBuildManager.cs&version=GBmaster&line=78&lineStyle=plain&lineEnd=79&lineStartColumn=1&lineEndColumn=1) implementation simply throws `NotImplementedException`'s so I don't think there is any harm in merging this even though its incomplete. If that's not a good idea please let me know.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
@@ -23,15 +23,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation
     {
         private readonly VSLangProj.VSProject _vsProject;
 
+        private readonly IUnconfiguredProjectCommonServices _unconfiguredProjectServices;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="VsBuildManager"/> class.
         /// </summary>
         [ImportingConstructor]
         internal VsBuildManager(
-            [Import(ExportContractNames.VsTypes.CpsVSProject)] VSLangProj.VSProject vsProject)
+            [Import(ExportContractNames.VsTypes.CpsVSProject)] VSLangProj.VSProject vsProject,
+            IUnconfiguredProjectCommonServices unconfiguredProjectServices)
         {
             AddEventSource(this as IEventSource<_dispBuildManagerEvents>);
             _vsProject = vsProject;
+            _unconfiguredProjectServices = unconfiguredProjectServices;
         }
 
         #region _dispBuildManagerEvents_Event Members
@@ -71,7 +75,32 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation
         {
             get
             {
-                throw new NotImplementedException();
+                var project = _unconfiguredProjectServices.ActiveConfiguredProject;
+                var ruleSource = project.Services.ProjectSubscription.ProjectRuleSource;
+                var update = _unconfiguredProjectServices.ThreadingService.ExecuteSynchronously(() => ruleSource.GetLatestVersionAsync(project, new string[] { Compile.SchemaName }));
+                var snapshot = update[Compile.SchemaName];
+
+                var monikers = new List<string>();
+                foreach (var item in snapshot.Items.Values)
+                {
+                    bool isLink = GetBooleanPropertyValue(item, Compile.LinkProperty);
+                    bool designTime = GetBooleanPropertyValue(item, Compile.DesignTimeProperty);
+
+                    if (!isLink && designTime)
+                    {
+                        if (item.TryGetValue(Compile.FullPathProperty, out string path))
+                        {
+                            monikers.Add(path);
+                        }
+                    }
+                }
+                
+                return monikers.ToArray();
+
+                bool GetBooleanPropertyValue(System.Collections.Immutable.IImmutableDictionary<string, string> item, string propertyName)
+                {
+                    return item.TryGetValue(propertyName, out string value) && StringComparers.PropertyValues.Equals(value, "true");
+                }
             }
         }
 
@@ -91,6 +120,20 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation
         /// </summary>
         public virtual string BuildDesignTimeOutput(string bstrOutputMoniker)
         {
+            var languageName = _unconfiguredProjectServices.ThreadingService.ExecuteSynchronously(async () =>
+            {
+                var property = await _unconfiguredProjectServices.ActiveConfiguredProjectProperties.GetConfigurationGeneralPropertiesAsync();
+                return await property.LanguageServiceName.GetValueAsync();
+            });
+
+            // TODO: Once Roslyn work is complete:
+            //
+            //   var compiler = _serviceProvider.GetService(typeof(Microsoft.VisualStudio.LanguageServices.Implementation.ProjectSystem.CPS.ITempPECompilerHost));
+            //  (after importing a service provider in the constructor as: [Import(typeof(SVsServiceProvider))]IServiceProvider serviceProvider)
+            //  (or maybe can just import a Lazy<ITempPECompilerHost> in the constructor?)
+            //
+            //   compiler.Compile(languageName ... );
+
             throw new NotImplementedException();
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
@@ -1,0 +1,137 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.ProjectSystem.VS.ConnectionPoint;
+using Microsoft.VisualStudio.ProjectSystem.VS.Interop;
+using Microsoft.VisualStudio.Shell;
+using VSLangProj;
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProject.cs
```diff
@@ -30,18 +30,20 @@ public partial class VSProject : VSLangProj.VSProject, IConnectionPointContainer
...
+        private readonly BuildManager _buildManager;
 
         [ImportingConstructor]
         internal VSProject(
             [Import(ExportContractNames.VsTypes.CpsVSProject)] VSLangProj.VSProject vsProject,
             IProjectThreadingService threadingService,
             ActiveConfiguredProject<ProjectProperties> projectProperties,
-            UnconfiguredProject project)
+            UnconfiguredProject project,
+            [Import(typeof(BuildManager))]BuildManager buildManager)
```
davkean: Don't need the typeof(BuildManager) - that's implied.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                return monikers.ToArray();
+
+                bool GetBooleanPropertyValue(System.Collections.Immutable.IImmutableDictionary<string, string> item, string propertyName)
+                {
+                    return item.TryGetValue(propertyName, out string value) && StringComparers.PropertyValues.Equals(value, "true");
+                }
+            }
+        }
+
+        /// <summary>
```
davkean: Please remove the "undocumented" blocks - I know CPS does this but it's really unhelpful.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// Undocumented.
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1065:DoNotRaiseExceptionsInUnexpectedLocations")]
+        public virtual object Parent
+        {
+            get { throw new NotImplementedException(); }
```
davkean: Why virtual, and what does this return for legacy?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        #region BuildManager Members
+
+        /// <summary>
+        /// Undocumented.
+        /// </summary>
+        public virtual string BuildDesignTimeOutput(string bstrOutputMoniker)
+        {
+            var languageName = _unconfiguredProjectServices.ThreadingService.ExecuteSynchronously(async () =>
```
davkean: Validate the bstroutputMoniker the same way legacy does.

davkean: Why is this virtual?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davkean: This returns VSProject in legacy.

davidwengier: The virtual is just because I copied this file direct from CPS. Will clean it up.

The only thing worth looking at in this file at the moment is the DesignTimeOutputMonikers property, but in legacy this returns the project, so presumably _vsProject.

davkean: This needs to be "our" VSProject not CPS's version.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                var property = await _unconfiguredProjectServices.ActiveConfiguredProjectProperties.GetConfigurationGeneralPropertiesAsync();
+                return await property.LanguageServiceName.GetValueAsync();
+            });
+
+            // TODO: Once Roslyn work is complete:
+            //
+            //   var compiler = _serviceProvider.GetService(typeof(Microsoft.VisualStudio.LanguageServices.Implementation.ProjectSystem.CPS.ITempPECompilerHost));
```
davkean: Use IVsService<T>.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davkean: LanguageName can be empty here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: As in `IVsService<ITempPECompilerHost>`? I'll try it out when I get up to that bit, but in the experimentation to get to here, roslyn exports weren't available when this was being constructed

davkean: This isn't an export - this is a proffered service. I think if this setup correctly, Roslyn package should be initialized by the fact that you are pull on the service. IVsService is identical to this except it forces UI threadiness.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davkean: Any method that uses Roslyn service should be switching to the UI thread, basically should wrap this method in ExecuteSync.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,180 @@
...
+    /// <summary>
+    /// Undocumented.
+    /// </summary>
+    [Export(typeof(BuildManager))]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
+    [Order(Order.Default)]
+    internal class VsBuildManager : ConnectionPointContainer,
+                                    IEventSource<_dispBuildManagerEvents>,
+                                    BuildManager,
+                                    BuildManagerEvents
```
davkean: This needs to be returns by VSProject.Events.BuildManagerEvents somehow.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                        {
+                            monikers.Add(path);
+                        }
+                    }
+                }
+                
+                return monikers.ToArray();
+
+                bool GetBooleanPropertyValue(System.Collections.Immutable.IImmutableDictionary<string, string> item, string propertyName)
+                {
```
davkean: I don't think we want to use snapshot data here; or we only want to use it there's an update-to-date one. I owe you a pattern for this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: We have `VisualBasicVSProjectEvents` which is easy to hook this into and fills the need. Is there a good reason not to just remove the VisualBasic capability check from that? It doesn't have anything interesting in it that I can see, I assume its there mainly to allow our `VisualBasicVSImports` to be imported, but that has the capability check so shouldn't be a problem.

davkean: It was added to support "ImportsEvents"; but assuming VisualBasicVSImports continues to only apply to VB than that should be okay.

davidwengier: I can't see why this pattern from VsProject.cs wouldn't work. I'll give it a go.

https://github.com/dotnet/project-system/blob/c2e466aabd03b83de4a3b0f44a488cc6d13ce790/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProject.cs#L69-L83

davidwengier:
Updated with all of the PR feedback addressed, except some of the things in `BuildDesignTimeOutput(string)` as that whole method is really a TODO, and its implementation will come later. Merging this should allow for that to be PRed easily though.

PR #4177, opened 2018-10-25
Refactor IProjectTreeExtensions

drewnoakes:
- Remove unused methods `AddElementsToTree`, `RemoveElementsFromTree`, and `FindImmediateChildByPath`
- Rename `FindNodeByCaption` -> `GetChildWithCaption` (it only checks children)
- Inline method to simplify and improve perf

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectTreeExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectTreeExtensions.cs
@@ -13,7 +13,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         /// Gets the direct child of <paramref name="tree"/> with <paramref name="caption"/>
         /// if found, otherwise <see langword="null"/>.
         /// </summary>
-        public static IProjectTree GetChildWithCaption(this IProjectTree tree, string caption)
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectTreeExtensions.cs
```diff
@@ -12,120 +10,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
...
-        ///     <para>
-        ///         -or-
-        ///     </para>
-        ///     <paramref name="path"/> is <see langword="null"/>.
-        /// </exception>
-        /// <exception cref="ArgumentException">
-        ///     <paramref name="path"/> is an empty string ("")
-        /// </exception>
-        public static IProjectTree FindImmediateChildByPath(this IProjectTree tree, string path)
+        public static IProjectTree GetChildWithCaption(this IProjectTree tree, string caption)
```
davkean: Interesting that you changed this to "Get"; across the framework we tend to use "Find" to indicate an API that can return null, such as this one.

davkean: And CPS uses FindXXX/TryFindXXX, I would favor it.

drewnoakes:
@dotnet-bot retest windows_integration_release please

PR #4180, opened 2018-10-25
Simplify package reference version

davkean:
This is an updated attempt at simplifying the package reference versioning. This will make the move to Dev16 a lot simpler as we won't need to copy around the XXXVersion pattern.


Pilchie:
Some restore errors in the CI builds...

PR #4184, opened 2018-10-28
adding code coverage

jmarolf:
Have code coverage work in azure pipelines


jmarolf:
> Maybe we live without CC for the rest of 15.9, and just merge this into dev16.0.x?

I don't have a problem with this.  I'll retarget.

jmarolf:
I'll need to take a different approach here

PR #4194, opened 2018-10-30
Update Roslyn dependencies

davkean:
Update Roslyn to the version currently in Preview 1 so that we can consume the new APIs for Razor.

### DIFF
--- a/src/HostAgnostic.props
+++ b/src/HostAgnostic.props
@@ -17,6 +17,8 @@
     <!-- Roslyn -->
     <PackageReference Include="Microsoft.Net.RoslynDiagnostics" />
     <PackageReference Include="Microsoft.VisualStudio.LanguageServices" />
+    <PackageReference Include="Microsoft.CodeAnalysis" />
+    
 
     <!-- MSBuild -->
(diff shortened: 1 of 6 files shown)


davkean:
Only took me 3 days to figure out the right numbers.

src/HostAgnostic.props
```diff
@@ -17,6 +17,8 @@
     <!-- Roslyn -->
     <PackageReference Include="Microsoft.Net.RoslynDiagnostics" />
     <PackageReference Include="Microsoft.VisualStudio.LanguageServices" />
+    <PackageReference Include="Microsoft.CodeAnalysis" />
```
Pilchie: Really - we didn't use to have a reference to all of Roslyn - why do we need it now?  For `ProjectId`?

davkean: LanguageSevices changed its dependencies; before we had access to Microsoft.CodeAnalysis.CSharp/Microsoft.CodeAnalysis.VisualBasic (parsers) and the associated language-specific workspace APIs (rename) through it - now we have to explicitly state it.

PR #4195, opened 2018-10-30
Improve image moniker debugging

drewnoakes:
Adds `DebuggerDisplay` attributes for `ImageMoniker` and `ProjectImageMoniker` that produce strings such as `SdkWarning` for instances of these types in the debugger.

Previously the debugger would just show the type name, and when expanded the GUID and Id.

These changes have no effect unless a debugging.

Before:

image

After:

image

You can still see the GUID and Id if you expand the node.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Resources/ManagedImagesMonikers.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Resources/ManagedImagesMonikers.cs
@@ -1,40 +1,127 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 
+using Microsoft.VisualStudio.Imaging;
 using Microsoft.VisualStudio.Imaging.Interop;
+using Microsoft.VisualStudio.ProjectSystem;
+
+[assembly: DebuggerDisplay("{Microsoft.VisualStudio.ProjectSystem.VS.ManagedImageMonikers.ImageMonikerDebugDisplay(this)}", Target = typeof(ImageMoniker))]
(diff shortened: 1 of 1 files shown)


drewnoakes:
@dotnet-bot retest windows_integration_release please

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Resources/ManagedImagesMonikers.cs
```diff
@@ -1,40 +1,127 @@
...
+
+        internal static string ProjectImageMonikerDebugDisplay(ProjectImageMoniker moniker) => DebugDisplay(moniker.Guid, moniker.Id);
+
+        private static string DebugDisplay(Guid guid, int id)
+        {
+            if (guid == s_manifestGuid)
+            {
+                switch (id)
+                {
+                    case 0:  return nameof(ApplicationPrivate);
```
davkean: An alternative for this would to be checking against the static fields themselves to avoid getting out of sync with the id.

drewnoakes: I agree it'd be good to prevent them getting out of sync.

`case` needs a compile time constant. We'd have to pull out consts for the IDs. Do you think it'd be worth it?

Another alternative would be use reflection and a loop. Not sure if that overhead would adversely impact the debugging experience. It feels a bit dirty.

davkean: I thought pattern matching might help here?

drewnoakes: I'll try it out. Pattern matching falls back to chained if/else rather than a jump table. Might not be a big deal. Still orders of magnitude better than reflection.

drewnoakes: Pattern matching didn't work. It might in a future version of C#. Our options are to leave it or pull out constants. Or add some documentation saying "if you change these, update this over here too".

PR #4198, opened 2018-10-30
Run sign on official build.

brettfo:


davkean:
What's the back story on this?

Pilchie:
I merged the equivalent fix in https://github.com/dotnet/project-system/pull/4197.

This was accidentally broken in https://github.com/dotnet/project-system/pull/4107.

davkean:
Are we missing verification for this?

Pilchie:
DevDiv insertions have validations, but we don't in our signed build leg.  I *think* there is now a signcheck task we could enable more locally, right @jmarolf ?

PR #4199, opened 2018-10-30
Reduce allocation and contention within dependencies node 

drewnoakes:
I'm continuing to learn my way around the dependencies node code. Here is a series of small changes that reduce allocations, reduce contention, reduce unnecessary work and tidy up little bits and pieces.

Most commits are small and focussed, so most commits will be quick to review. Apologies for the bulk&mdash;I did spend some time trying to pull smaller bits out but they're all focussed in the same areas and conflict heavily. This would require more time than I felt it was worth.

Some quick things to note:

- Added `StringComparisons` class to mirror `StringComparers` (89be343270f80d1856bf69b792ef7686d15657a8).
- Usage of `readonly struct` with `in` modifier to reduce defensive copying (9652302bccd52ac1c4f96845f8d541d9d38d878f, 9652302bccd52ac1c4f96845f8d541d9d38d878f). This is something we may be able to use more. The language feature is much newer than most of this code.
- Many properties on `DependencyModel` were made virtual/abstract to exchange dedicated storage on the base class with method-based (getter) access for constant or simple conditional values such as icon sets, priorities, versions, etc. (several commits towards the end).

Areas I want to draw particular review attention to:

- Make `TaskDelayScheduler` non-locking (61a5b6dda51877ac8239634868f7677da2595395, 1e0868519fcb884a259db8c358aeca7e12ea452a). I saw very high CPU usage with `Roslyn.sln` across all cores, with most time spent in the kernel. Looking at performance counters, the number of contentions was increasing very quickly (I saw the total exceed 100,000 in total after a few minutes). Breaking into the debugger showed several threads trying to enter the lock in `TaskDelayScheduler`. I changed this type to not require any locking, using user-space interlocked instructions instead. CPU usage is still high, but I don't see nearly as many contentions any more.

- As project changes run from Dataflow through the dependency subsystem, they are debounced via `TaskDelayScheduler`. This schedules a `Func<CancellationToken, Task>`, allowing an in-flight operation to cancel itself when a new update arrives. This cancellation token was not being passed forward, but is now (4357fd96a215c4dc3c565bb9afd3c09294a40f85). I want to get this PR out for review before EOD here, but I have in my notes to perform another review to verify there's no issue due to cancelling before applying deltas. It should be possible to compute deltas each time from full snapshots, so if one is skipped at any level, it'll catch up. There's also some re-entrancy here where one update can schedule another, hence the ability to pass a cancellation token per-request to `TaskDelayScheduler` (90518b7b4ea128a712007b638b24d3429fd8bc8a).

- Removal of some async state machines (17dcd2b3b48da748a41c8ce6428dac49cc259bcb, 349367b8d40177a09934f878e819d53f87442920). This can change behaviour in case of exceptions, and can remove these non-`async` methods from call stacks. I believe these scenarios are fine.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesChanges.cs
```diff
@@ -5,32 +5,49 @@
...
 
-        public void IncludeAddedChange(IDependencyModel model)
+        public ImmutableArray<IDependencyModel> AddedNodes
         {
-            Added.Add(model);
+            get
+            {
+                lock (_added)
+                {
+                    return ImmutableArray.CreateRange(_added);
```
davkean: Is there benefit to move these to immutable to avoid the copying, and then use ApplyChangeOptimistically to avoid the locking?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleChangeContext.cs
```diff
@@ -18,55 +18,36 @@ public DependenciesRuleChangeContext(ITargetFramework target, IProjectCatalogSna
...
-                change.IncludeRemovedChange(ruleMetadata);
-            }
+            GetChanges(targetFramework).IncludeRemovedChange(ruleMetadata);
         }
 
         private DependenciesChanges GetChanges(ITargetFramework targetFramework)
         {
-            if (!_changes.TryGetValue(targetFramework, out IDependenciesChanges change))
+            lock (_changesLock)
             {
```
davkean: Ditto, I would consider using ThreadingTools.ApplyChangeOptimistically to avoid the lock; it's perfect for this sort of pattern.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
```diff
@@ -110,30 +111,28 @@ private void SubscribeToConfiguredProject(IProjectSubscriptionService subscripti
...
             EnsureInitialized();
 
-
-            await _tasksService.LoadedProjectAsync(async () =>
+            await _tasksService.LoadedProjectAsync(() =>
             {
                 if (_tasksService.UnloadCancellationToken.IsCancellationRequested)
                 {
-                    return;
+                    return Task.CompletedTask;
```
davkean: The pattern here is usually to ThrowIfCancellationRequested and to fault the block - that's exactly what LoadProjectAsync throws if the project is already unloaded at this point.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
```diff
@@ -185,41 +186,33 @@ private void SubscribeToConfiguredProject(IProjectSubscriptionService subscripti
             IEnumerable<string> addedSharedImportPaths = sharedFolderProjectPaths.Except(currentSharedImportNodePaths);
             foreach (string addedSharedImportPath in addedSharedImportPaths)
             {
-                IDependencyModel added = CreateDependencyModel(addedSharedImportPath, targetContext.TargetFramework, resolved: true);
+                IDependencyModel added = CreateDependencyModel(addedSharedImportPath, resolved: true);
```
davkean: Why isn't TargetFramework required anymore? Shared imports aren't expected to vary per framework?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -646,7 +646,14 @@ protected async Task CheckoutSettingsFileAsync()
         /// file with the name LaunchSettings.json. We don't need to special case because, if a file with this name
         /// changes we will only check if the one we cared about was modified.
         /// </summary>
-        protected void LaunchSettingsFile_Changed(object sender, FileSystemEventArgs e)
+        private void LaunchSettingsFile_Changed(object sender, FileSystemEventArgs e)
+        {
+#pragma warning disable VSTHRD110 // Observe result of async calls
+            HandleLaunchSettingsFileChangedAsync();
```
davkean: .Forgot is the pattern.

davkean:

davkean:
While the commits were concise I'd prefer in future that they we split unrelated ones into different PRs and keep the PR focused on one thing; remove locks from TaskSchedular, make use of 'in', reduce memory overhead of DependencyModel, etc.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <para>Consuming code is responsible for managing overlapping asynchronous operations.</para>
+    /// <para>This class has a lock-free implementation to minimise latency and contention.</para>
+    /// </remarks>
+    public sealed class CancellationSeries : IDisposable
+    {
+        private CancellationTokenSource _cts = new CancellationTokenSource();
+
+        private readonly CancellationToken[] _superToken;
```
Pilchie: Why is this an array? It never has more than one element.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            priorSource.Cancel();
+            priorSource.Dispose();
+
+            return nextToken;
+        }
+
+        /// <summary>Gets whether this object has been disposed.</summary>
+        public bool IsDisposed => Volatile.Read(ref _cts) == null;
```
Pilchie: Doesn't seem to be used?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    /// <summary>
+    /// Produces a series of <see cref="CancellationToken"/> objects such that requesting a new token
+    /// causes the previously issued token to be cancelled.
+    /// </summary>
+    /// <remarks>
+    /// <para>Consuming code is responsible for managing overlapping asynchronous operations.</para>
+    /// <para>This class has a lock-free implementation to minimise latency and contention.</para>
+    /// </remarks>
+    public sealed class CancellationSeries : IDisposable
```
Pilchie: I wonder if @stephentoub would like to review this class.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -98,7 +98,8 @@ public async Task InitializeAsync()
             // Check to see if a solution is already open. If so we set _solutionOpened to true so that subsequent projects added to 
             // this solution are processed.
             if (ErrorHandler.Succeeded(solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out object isFullyLoaded)) &&
-                isFullyLoaded is true)
```
Pilchie: I didn't know that boxed.  That's too bad.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySubscriptionsHost.cs
```diff
@@ -270,13 +265,14 @@ private void OnSubtreeProviderDependenciesChanged(object sender, DependenciesCha
...
-            UpdateDependenciesSnapshotAsync(changes, catalogs: null, activeTargetFramework: null);
+            UpdateDependenciesSnapshotAsync(changes, catalogs: null, activeTargetFramework: null, e.Token);
         }
 
         private void UpdateDependenciesSnapshotAsync(
             ImmutableDictionary<ITargetFramework, IDependenciesChanges> changes,
             IProjectCatalogSnapshot catalogs,
-            ITargetFramework activeTargetFramework)
+            ITargetFramework activeTargetFramework,
+            CancellationToken token = default)
```
Pilchie: I don't like making these optional for things that aren't part of a public API.  It's too easy to avoid passing them along.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IProjectDependenciesSubTreeProvider.cs
```diff
@@ -27,19 +28,23 @@ public interface IProjectDependenciesSubTreeProvider
...
             Provider = provider;
             TargetShortOrFullName = targetShortOrFullName;
             Changes = changes;
+            Token = token;
         }
 
         public IProjectDependenciesSubTreeProvider Provider { get; }
         public string TargetShortOrFullName { get; }
         public IDependenciesChanges Changes { get; }
+        public CancellationToken Token { get; }
```
Pilchie: Not thrilled with having these on eventargs, but I understand it...

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -58,7 +58,7 @@ internal TargetedDependenciesSnapshot(IDictionary<string, IDependency> dependenc
 
         private readonly Dictionary<string, IDependency> _topLevelDependenciesByPathMap = new Dictionary<string, IDependency>(StringComparer.OrdinalIgnoreCase);
         private readonly ConcurrentDictionary<string, IList<IDependency>> _dependenciesChildrenMap = new ConcurrentDictionary<string, IList<IDependency>>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, bool> _unresolvedDescendantsMap = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
+        private readonly ConcurrentDictionary<string, bool> _unresolvedDescendantsMap = new ConcurrentDictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
Pilchie: Also, adding this comment again in case you don't see the Outdated one:

Unless this was a known contentious path, this may not be worth it. In any case, you might want to limit the concurrencyLevel of the new dictionary, since ConcurrentDictionary tends to be a very large allocation by default.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
drewnoakes: Updated.

Looking at the implementation of `LoadProjectAsync`, it seems to do this check synchronously before calling our func anyway. No harm in doing it again here though.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
drewnoakes: `SharedProjectDependencyModel` (nor any other `DependencyModel` subclass) doesn't store a target framework.

These models are currently added to a `DependenciesRuleChangeContext` which provides the framework.

In general I'm trying to remove stuff that's provably unused (even if it might be useful one day) as I find it makes the code easier to understand. It's easier to add code than remove it.

I'm still getting my head around the layering here. These are good questions for me, cheers.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
drewnoakes: Updated.

We have a few other unobserved tasks throughout the code (search for `nowait` for example). Are these candidates for `.Forget()` as well? They don't raise `VSTHRD110`.

It seems that `.Forget()` doesn't actually do anything with the tasks beyond suppress warnings. Is that really a great idea? Can we do something more sensible with them that's specific to our code, such as provide a continuation in case of fault that logs something sensible, or launches a debugger in `DEBUG` builds, or...

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleHandlerBase.cs
```diff
@@ -14,26 +14,38 @@ internal abstract class DependenciesRuleHandlerBase
...
+
+        protected DependenciesRuleHandlerBase(
+            string unresolvedRuleName,
+            string resolvedRuleName)
+        {
+            UnresolvedRuleName = unresolvedRuleName;
+            ResolvedRuleName = resolvedRuleName;
+
+            _evaluationRuleNames = ImmutableStringHashSet.EmptyOrdinal.Add(unresolvedRuleName);
+            _designTimeBuildRuleNames = ImmutableStringHashSet.EmptyOrdinal.Add(unresolvedRuleName).Add(resolvedRuleName);
```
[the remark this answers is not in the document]
drewnoakes: Good idea. Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
```diff
@@ -117,10 +117,10 @@ public IDependency FindDependency(string id, bool topLevel = false)
                     subTreeProviders,
                     projectItemSpecs,
                     out bool anyTfmChanges);
-                builder[targetFramework] = newTargetedSnapshot;
```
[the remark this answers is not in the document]
drewnoakes: In later commits I got to the point where `anyTfmChanges` being `true` means `ReferenceEquals(newTargetedSnapshot, previousSnapshot)`. The `out` parameter is redundant in that sense and could be removed. So in this case we'd be overwriting the dictionary entry with an identical reference.

I've removed the `out` param and added some API docs to `FromChanges`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: It's passed to a `params` method. This avoids allocating on each call.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: You're right. I've removed it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: It'd be great to hear his thoughts.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
drewnoakes: Yeah it could possibly generate better code, but it's not a biggie. Might be nice to have a `System.Boolean.TrueObject`/`FalseObject` for these kinds of comparisons.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySubscriptionsHost.cs
drewnoakes: Updated.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IProjectDependenciesSubTreeProvider.cs
drewnoakes: Agree but I don't know a better approach for this scenario. I don't think it's a good fit for MEF's `ImportMany`. We have Dataflow coming in to us, but then transition to an imperative/event based means to propagate data forward.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesChanges.cs
drewnoakes: Possibly. I did consider it. I felt this approach might come out ahead, but I should do some measurements to find out. With the optimisic pattern we'd need to use `ImmutableList<T>` rather than `ImmutableArray<T>`, which has worse space complexity.

I experimented with separating these two phases out (builder and snapshot), but `DependenciesRuleChangeContext` _can_ be modified after would first need the snapshot. There's possibly something more we can do here.

drewnoakes: I ran a benchmark to learn more about this.

Here's the code:

```c#
[ClrJob]
[MemoryDiagnoser]
public class BuildImmutableFromSetBenchmarks
{
    [Params(0, 1, 10, 100, 1000)]
    public int Count;

    private string[] _strings;
    private HashSet<string> _hashSet;
    private ImmutableHashSet<string> _immutableHashSet;

    [GlobalSetup]
    public void Setup()
    {
        _strings = Enumerable
            .Range(0, Count)
            .Select(i => i.ToString())
            .ToArray();
    }

    [Benchmark]
    public ImmutableHashSet<string> OptimisticImmutableHashSet()
    {
        _immutableHashSet = ImmutableHashSet.Create<string>(StringComparer.OrdinalIgnoreCase);

        foreach (var s in _strings)
        {
            var before = Volatile.Read(ref _immutableHashSet);
            while (Interlocked.CompareExchange(ref _immutableHashSet, before.Add(s), before) != before)
            {

            }
        }

        return _immutableHashSet;
    }

    [Benchmark]
    public ImmutableArray<string> HashSetToImmutableArray()
    {
        _hashSet = new HashSet<string>(StringComparer.OrdinalIgnoreCase);

        foreach (var s in _strings)
        {
            lock (_hashSet)
            {
                _hashSet.Add(s);
            }
        }

        lock (_hashSet)
        {
            return ImmutableArray.CreateRange(_hashSet);
        }
    }
}
```

I tried to replicate the scenario as best as possible. Obviously there's no concurrency here. It also doesn't simulate producing the `ImmutableArray` more than once, which can happen in theory. Actually we can cache the result, with invalidation on write to help with that.

The results:

|                           Method | Count |           Mean |         Error |        StdDev |         Median |   Gen 0 |  Gen 1 | Allocated |
|--------------------------------- |------ |---------------:|--------------:|--------------:|---------------:|--------:|-------:|----------:|
|       OptimisticImmutableHashSet |     0 |       621.5 ns |     12.103 ns |     15.306 ns |       616.6 ns |  0.0029 |      - |      20 B |
|          HashSetToImmutableArray |     0 |       145.2 ns |      3.935 ns |     11.603 ns |       138.6 ns |  0.0074 |      - |      40 B |
|       OptimisticImmutableHashSet |     1 |       848.4 ns |     16.019 ns |     20.829 ns |       849.9 ns |  0.0134 |      - |      72 B |
|          HashSetToImmutableArray |     1 |       275.4 ns |      5.205 ns |      5.112 ns |       276.9 ns |  0.0286 |      - |     152 B |
|       OptimisticImmutableHashSet |    10 |     4,547.7 ns |     89.233 ns |    158.612 ns |     4,555.7 ns |  0.2289 |      - |    1212 B |
|          HashSetToImmutableArray |    10 |     1,579.6 ns |     19.275 ns |     17.086 ns |     1,583.4 ns |  0.1144 |      - |     620 B |
|       OptimisticImmutableHashSet |   100 |    71,160.5 ns |  1,463.607 ns |  4,055.647 ns |    70,170.6 ns |  4.3945 |      - |   23300 B |
|          HashSetToImmutableArray |   100 |    13,689.1 ns |    264.648 ns |    325.012 ns |    13,730.5 ns |  1.1749 |      - |    6220 B |
|       OptimisticImmutableHashSet |  1000 | 1,035,607.0 ns | 17,529.009 ns | 14,637.515 ns | 1,032,324.7 ns | 68.3594 | 3.9063 |  360931 B |
|          HashSetToImmutableArray |  1000 |   131,551.6 ns |  2,597.237 ns |  2,027.752 ns |   130,749.8 ns | 11.7188 | 1.2207 |   62429 B |

```text
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.345 (1803/April2018Update/Redstone4)
Intel Xeon Silver 4110 CPU 2.10GHz, 2 CPU, 32 logical and 16 physical cores
Frequency=2045972 Hz, Resolution=488.7652 ns, Timer=TSC
  [Host] : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3190.0
  Clr    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3190.0

Job=Clr  Runtime=Clr
```

Summary from those results is that the `HashSet` approach, with `Monitor`-based locking, is strictly faster on all counts, however for lower counts it allocates more memory. As the number of items grows however, the optimistic immutable collection approach allocates considerably more. Somewhere between 1 and 10 elements, the `HashSet` approach takes the lead.

image

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
stephentoub: There's an overload of CreateLinkedTokenSource that accepts two tokens... instead of using an array here, you could just pass this "super token" as one of the inputs and pass `default`/`CancellationToken.None` as the other.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,112 @@
...
+        public CancellationToken GetToken(CancellationToken token = default)
+        {
+            // Create the next CTS. There are four scenarios, depending upon whether
+            // a super token was given to the constructor, and whether a token was
+            // passed to this method.
+
+            CancellationTokenSource nextSource = _superToken == null
+                ? token == default
+                    ? new CancellationTokenSource()
+                    : CancellationTokenSource.CreateLinkedTokenSource(token)
```
stephentoub: Same here: if you pass `default` as the second argument, you'll avoid the params array.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,112 @@
...
+                : token == default
+                    ? CancellationTokenSource.CreateLinkedTokenSource(_superToken)
+                    : CancellationTokenSource.CreateLinkedTokenSource(token, _superToken[0]);
+
+            CancellationToken nextToken = nextSource.Token;
```
stephentoub: Why store this into a local, rather than just `return nextSource.Token` at the end?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// A cancellation token that will be cancelled when either:
+        /// <list type="bullet">
+        /// <item><see cref="GetToken"/> is called again</item>
+        /// <item>The token passed to this method (if any) is cancelled</item>
+        /// <item>The token passed to the constructor (if any) is cancelled</item>
+        /// <item><see cref="Dispose"/> is called</item>
+        /// </list>
+        /// </returns>
+        /// <exception cref="ObjectDisposedException">This object has been disposed.</exception>
+        public CancellationToken GetToken(CancellationToken token = default)
```
stephentoub: FWIW, just looking at this method name doesn't suggest to me personally that it's canceling/disposing the previous token and creating and storing a new one.  But, maybe in context it'd be more clear?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// <summary>
+    /// Produces a series of <see cref="CancellationToken"/> objects such that requesting a new token
+    /// causes the previously issued token to be cancelled.
+    /// </summary>
+    /// <remarks>
+    /// <para>Consuming code is responsible for managing overlapping asynchronous operations.</para>
+    /// <para>This class has a lock-free implementation to minimise latency and contention.</para>
+    /// </remarks>
+    public sealed class CancellationSeries : IDisposable
+    {
```
stephentoub: What are the chances that an instance of this class gets dropped without being Dispose'd?  Maybe add a debug-only finalizer to catch any such misuse, as that would potentially mean a registration from a linked token source was leaked.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Great, didn't realise `CreateLinkedTokenSource` had that property. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,112 @@
...
+                }
+
+                priorSource = snapshot;
+            }
+
+            priorSource.Cancel();
+            priorSource.Dispose();
+
+            return nextToken;
+        }
```
stephentoub: This Interlocked.CompareExchange loop looks fine.  I'm just wondering if it's overkill, and if just using:
```C#
CancellationTokenSource priorSource = Interlocked.Exchange(ref _cts, nextSource);
if (priorSource == null) throw ...;
priorSource.Cancel();
priorSource.Dispose();
return nextSource.Token;
```
would be sufficient?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: No reason. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                if (ReferenceEquals(snapshot, priorSource))
+                {
+                    break;
+                }
+
+                priorSource = snapshot;
+            }
+
+            priorSource.Cancel();
```
stephentoub: Maybe this isn't possible based on construction, but do you need to handle the case where a delegate registered with a cancellation token throws an unhandled exception?  That would emerge from here and prevent priorSource from being disposed and prevent the token from being returned.  Does that matter?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: I believe you're right. That's much simpler. Thank you!

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Current usages are long lived but it seems like a good idea to be able to catch this.

Is there any reason not to add a finalizer that does this, even in non-DEBUG builds?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: This sounds entirely reasonable. I'll look into the mechanics.

For now the only protection that comes to mind is catching exceptions thrown from `Cancel`, disposing both sources and rethrowing.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Agree, though I can't think of a better name off hand. Would `GetNext` be better? Everything else that comes to mind is quite wordy.

stephentoub: Adding "Next" to the name sounds good.  And maybe "Create" instead of "Get".  But up to you, of course.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
stephentoub: > Is there any reason not to add a finalizer that does this, even in non-DEBUG builds?

There's a small run-time cost to adding a finalizer, but it might be insignificant compared to everything else. The real issue would be what the finalizer does.... if it crashes, you probably wouldn't want that in release :)  If it just logs, and pops an assert in a debug build, should be fine.

drewnoakes: Fixed in eb02e72be2f150de51dfc12eea88ce4e539e8892. I went for an assert.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Changed in 78a446498287aefba469ed90b5393d10178afde6.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Fixed in 6d9b69a7f82774095a81960b2ad4dfe80c89bb24, along with a test.

I couldn't find a way to verify that `priorSource` was disposed via public APIs on `CancellationToken`.

drewnoakes:
Thanks very much @stephentoub for the fantastic review comments!

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -29,23 +30,30 @@ public CancellationSeries(CancellationToken token = default)
             _superToken = token;
         }
 
+#if DEBUG
+        ~CancellationSeries()
+        {
+            Debug.Assert(_cts == null, "Instance of CancellationSeries not disposed before being finalized");
```
Pilchie: In Roslyn we had a lot of trouble with things like this in unit tests due to shutdown.   We ended up wrapping most of them in `if (!Environment.ShutdownStarted)`.

Pilchie: Also, probably doesn't matter since this is debug only, but consider adding a `SupppressFinalize` to the `Dispose` method.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleChangeContext.cs
drewnoakes: Looking at `ApplyChangeOptimistically`, it could do with an overload that takes an operand to pass to the func, so you don't have to allocate a lambda when calling. I'll file a PR, if I can find the repo :)

A while back I put together set/list types that hid this internally behind normal collection interfaces/APIs. I'll dig the code out and push it to GitHub somewhere. Might be useful.

drewnoakes: Microsoft/vs-threading#408

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Thanks, fixed both. Also noticed I didn't protect against exceptions from `Cancel` in `Dispose` so fixed that too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleChangeContext.cs
drewnoakes: I found the code on an old laptop. It's from 2017, is completely undocumented and untested, but shows the idea. Pushed here: https://github.com/drewnoakes/optimistic-immutable-collections

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesChanges.cs
davkean: What kind of contention do we expect to see here? It's not clear me to how many threads are actually looking that are adding nodes and inspecting them? I guess not a lot for the latter given you're not worried about the ImmutableArray allocations?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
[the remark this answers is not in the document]
drewnoakes: You're right. It prevents the corruption. The race isn't such an issue as it would only result in duplicated work. But actually it's not hard to address that too. I've pushed a change that uses `ConcurrentDictionary<,>.GetOrAdd`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -55,10 +56,8 @@ internal TargetedDependenciesSnapshot(IDictionary<string, IDependency> dependenc
 
         public ImmutableDictionary<string, IDependency> DependenciesWorld { get; private set; } = ImmutableStringDictionary<IDependency>.EmptyOrdinalIgnoreCase;
 
-        private readonly object _snapshotLock = new object();
-
         private readonly Dictionary<string, IDependency> _topLevelDependenciesByPathMap = new Dictionary<string, IDependency>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, IList<IDependency>> _dependenciesChildrenMap = new Dictionary<string, IList<IDependency>>(StringComparer.OrdinalIgnoreCase);
+        private readonly ConcurrentDictionary<string, IList<IDependency>> _dependenciesChildrenMap = new ConcurrentDictionary<string, IList<IDependency>>(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
drewnoakes: The alternative here is to introduce locks around some fairly long-running operations. `FindUnresolvedDependenciesRecursive` can walk the whole dependency graph, looking for issues.

I've never measured memory usage of `ConcurrentDictionary` before, basing decisions only on a general understanding of the implementation. Your comments prompted me to run a benchmark with different concurrency levels and different numbers of entries. Here are the results:

|                             Method | Count |          Mean | Allocated |
|----------------------------------- |------ |--------------:|----------:|
| ConcurrentDictionary_Concurrency32 |     0 |     370.23 ns |     868 B |
| ConcurrentDictionary_Concurrency16 |     0 |     223.92 ns |     484 B |
|  ConcurrentDictionary_Concurrency8 |     0 |     127.51 ns |     292 B |
|  ConcurrentDictionary_Concurrency4 |     0 |      90.44 ns |     196 B |
|  ConcurrentDictionary_Concurrency2 |     0 |      71.72 ns |     148 B |
|  ConcurrentDictionary_Concurrency1 |     0 |      61.67 ns |     124 B |
|                         Dictionary |     0 |     16.65 ns  |      48 B |
| ConcurrentDictionary_Concurrency32 |     1 |     451.41 ns |     892 B |
| ConcurrentDictionary_Concurrency16 |     1 |     291.28 ns |     508 B |
|  ConcurrentDictionary_Concurrency8 |     1 |     209.89 ns |     316 B |
|  ConcurrentDictionary_Concurrency4 |     1 |     170.77 ns |     220 B |
|  ConcurrentDictionary_Concurrency2 |     1 |     162.44 ns |     172 B |
|  ConcurrentDictionary_Concurrency1 |     1 |     143.28 ns |     148 B |
|                         Dictionary |     1 |      70.65 ns |     132 B |
| ConcurrentDictionary_Concurrency32 |    10 |   1,164.41 ns |    1108 B |
| ConcurrentDictionary_Concurrency16 |    10 |   1,016.14 ns |     724 B |
|  ConcurrentDictionary_Concurrency8 |    10 |   1,502.08 ns |     928 B |
|  ConcurrentDictionary_Concurrency4 |    10 |   1,268.21 ns |     832 B |
|  ConcurrentDictionary_Concurrency2 |    10 |     818.93 ns |     420 B |
|  ConcurrentDictionary_Concurrency1 |    10 |     804.48 ns |     400 B |
|                         Dictionary |    10 |    268.57 ns  |     292 B |
| ConcurrentDictionary_Concurrency32 |   100 |  11,349.97 ns |    6880 B |
| ConcurrentDictionary_Concurrency16 |   100 |  10,859.56 ns |    6496 B |
|  ConcurrentDictionary_Concurrency8 |   100 |  11,377.01 ns |    6304 B |
|  ConcurrentDictionary_Concurrency4 |   100 |   7,928.09 ns |    2980 B |
|  ConcurrentDictionary_Concurrency2 |   100 |   7,941.07 ns |    2940 B |
|  ConcurrentDictionary_Concurrency1 |   100 |   7,788.00 ns |    2920 B |
|                         Dictionary |   100 |  2,056.53 ns  |    2212 B |
| ConcurrentDictionary_Concurrency32 |  1000 | 104,636.70 ns |   60761 B |
| ConcurrentDictionary_Concurrency16 |  1000 | 102,770.06 ns |   60377 B |
|  ConcurrentDictionary_Concurrency8 |  1000 |  72,621.69 ns |   28261 B |
|  ConcurrentDictionary_Concurrency4 |  1000 |  75,603.64 ns |   28181 B |
|  ConcurrentDictionary_Concurrency2 |  1000 |  74,025.54 ns |   28141 B |
|  ConcurrentDictionary_Concurrency1 |  1000 |  77,335.37 ns |   28121 B |
|                         Dictionary |  1000 | 20,503.37 ns  |   22148 B |

<details>
<summary>Code for benchmark</summary>

```c#
[ClrJob]
[MemoryDiagnoser]
public class ConcurrentDictionaryBenchmarks
{
    [Params(0, 1, 10, 100, 1000)]
    public int Count;

    [Benchmark]
    public ConcurrentDictionary<int, string> ConcurrentDictionary_Concurrency32()
    {
        var dic = new ConcurrentDictionary<int, string>(concurrencyLevel: 32, capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }

    [Benchmark]
    public ConcurrentDictionary<int, string> ConcurrentDictionary_Concurrency16()
    {
        var dic = new ConcurrentDictionary<int, string>(concurrencyLevel: 16, capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }

    [Benchmark]
    public ConcurrentDictionary<int, string> ConcurrentDictionary_Concurrency8()
    {
        var dic = new ConcurrentDictionary<int, string>(concurrencyLevel: 8, capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }

    [Benchmark]
    public ConcurrentDictionary<int, string> ConcurrentDictionary_Concurrency4()
    {
        var dic = new ConcurrentDictionary<int, string>(concurrencyLevel: 4, capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }

    [Benchmark]
    public ConcurrentDictionary<int, string> ConcurrentDictionary_Concurrency2()
    {
        var dic = new ConcurrentDictionary<int, string>(concurrencyLevel: 2, capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }

    [Benchmark]
    public ConcurrentDictionary<int, string> ConcurrentDictionary_Concurrency1()
    {
        var dic = new ConcurrentDictionary<int, string>(concurrencyLevel: 1, capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }

    [Benchmark]
    public Dictionary<int, string> Dictionary()
    {
        var dic = new Dictionary<int, string>(capacity: Count);

        for (int i = 0; i < Count; i++)
        {
            dic[i] = "Hello";
        }

        return dic;
    }
}
```
</details>

Changing the concurrency level does reduce the size of the collection considerably when there are few entries, but there's still a lot of overhead there.

Here's a plot of bytes-per-element, based off the above data:

image

Given there's unlikely to be much contention here I agree and have reverted to using `Dictionary<,>` with a lock.

Thanks again for the insight.

drewnoakes:
@dotnet-bot retest unit-tests.debug please

drewnoakes:
@dotnet-bot retest unit-tests.debug please

drewnoakes:
@dotnet-bot retest unit-tests.debug please

drewnoakes:
@dotnet-bot retest unit-tests.debug.spanish please

drewnoakes:
@dotnet-bot retest please

PR #4204, opened 2018-11-01
 Add language-service support for Razor CSHTML

davkean:
This includes https://github.com/dotnet/project-system/pull/4170.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/HandlerServices.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
+{
+    internal static class HandlerServices
+    {
+        /// <summary>
+        ///     Normalizes <see cref="IProjectChangeDiff.RenamedItems"/> to <see cref="IProjectChangeDiff.AddedItems"/>
+        ///     and <see cref="IProjectChangeDiff.RemovedItems"/>.
+        /// </summary>
+        public static IProjectChangeDiff NormalizeRenames(IProjectChangeDiff difference)
```
drewnoakes: Seems like a good candidate for an extension method.

drewnoakes:
What is the significance of a _dynamic_ item? What makes it dynamic, what's it dynamic relative to, what's the alternative...?

davkean:
@drewnoakes You make good point on the dynamic; it's not clear enough. I'll make a follow up that addresses this - going to leave as it as we need this to get into Preview 1 and don't want to reset testing.

drewnoakes:
I don't necessarily have a problem with the name. I just don't know what it means, and the XML doc was a bit light.

PR #4176, opened 2018-10-25
Update for new MSBuild versions

Pilchie:

PR #4217, opened 2018-11-02
Correctly locate MSBuild on Dev16

davkean:
This is an alternative to https://github.com/dotnet/project-system/pull/4176, this handles both Dev15/Dev16. It does not, however, make uninstalling of the integration tests work - we'll need a follow up for that.

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -138,6 +138,19 @@ function LocateVisualStudio {
   return $vsInstallDir
 }
 
+function LocateMSBuild {
+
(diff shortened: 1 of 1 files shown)


Pilchie:
There is a reference to the `15.0` extensions folder in `build.ps1` that will probably also need to be updated?

davkean:
Yeah I called that out for the integration tests - I don't know how we're going to do that yet in a light up fashion, any suggestions?

Issue #3992, opened 2018-09-12
Integration tests are failing on 16.0 with missing "Microsoft.Test.Apex.Framework, Version=16.0.0.0"


davkean:
Integration tests are failing with the following under 16.0:

```
   Could not load file or assembly 'Microsoft.Test.Apex.Framework, Version=16.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
   at System.Reflection.RuntimeAssembly.nLoadFile(String path, Evidence evidence)
   at System.Reflection.Assembly.LoadFile(String path)
   at MessageHookProc(Int32 nCode, UInt32 wparam, Int32 lparam)
```

jmarolf:
Looks like we'll need to ask @AArnott to upload nuget packages for the 16.0 versions of APEX when they are ready.

AArnott:
To have seen this error at all, it seems you'd have to have something that has already seen Apex 16. so I guess it's already around. I'm no longer in charge of that though. @javierdlg may be the one to do it.

jmarolf:
Currently blocked on not having Dev16 images available

javierdlg:
This is odd, I only got this notification now.
Either way, dev16 bits are already released on the myget repo here:
https://vside.myget.org/feed/vs-impl/package/nuget/Microsoft.Test.Apex.VisualStudio

You should be able to update the nuget dependency to the latest one.

davkean:
Thanks, just need Dev16 images in VSTS.

javierdlg:
Can you go into a bit more detail? Do you need the Apex dlls published on the vs-impl feed?

davkean:
Dependencies are fine, we have all we need from Apex. Just referring to the fact that we don't yet have Virtual Machine images running Dev16 in our build lab that we can use to run against Dev16.

davkean:
When fixing this we also need to fix the reference to 15.0 in build.cmd for uninstall:

`AppData\Local\Microsoft\VisualStudio\15.0_$($vsid)$($hive)\Extensions`

jmarolf:
@javierdlg the 16.0 Apex package is unusable without 16.0 versions of the following packages:

- Microsoft.VisualStudio.Composition
- Microsoft.VisualStudio.Shell.15.0
- Microsoft.VisualStudio.WCFReference.Interop
- Microsoft.VSDesigner
[...]

PR #4218, opened 2018-11-02
remove groovy script for integration tests

jmarolf:
remove jenkins as none of the runs there will succeed. This bug https://github.com/dotnet/project-system/issues/3992 is what I am using to track re-enabling these tests once we have the correct images


jmarolf:
@dotnet-bot test ci please

Issue #4213, opened 2018-11-02
package properties tab should expose the new license property


karann-msft:
License URL is being deprecated in favor of License.

Licenses are now going to be a part of the NuGet package - [Packaging License within the nupkg](https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg)

There are now two ways to specify the package license:
1. using an SPDX identifier OR
2. including a license file within the nupkg

The package property page needs to be updated to surface these options:

![image](https://user-images.githubusercontent.com/16904420/47885964-cadc6400-ddf4-11e8-820c-e43217d4bf30.png)

If the expression is populated, the project file should be updated with:
`<PackageLicenseExpression>MIT</PackageLicenseExpression>`

If a license file is selected, it gets copied to the project root, and the project file should be updated with:
```
<PropertyGroup>
    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
</PropertyGroup>

<ItemGroup>
    <None Include="LICENSE.txt" Pack="true" PackagePath="$(PackageLicenseFile)" Visible="false" />
</ItemGroup>

```

etbyrd:
Merged https://github.com/dotnet/project-system/pull/4223 - should be in Preview 5

PR #4223, opened 2018-11-06
Adding License Expression and License File to Package Property Page

etbyrd:
This will be a fix for https://github.com/dotnet/project-system/issues/4213

Core Requirements:
- [x] Update UI
- [x] Verify Accessibility 
- [x] Setting up the controls (Browse for a .txt/.md/no extension file and update the TextBox)
- [x] Update .csproj with both PackageLicenseFile and PackageLicenseExpression properties
- [x] Update .csproj with License Item 

Potential Requirements:
- Match exact ItemGroup format
  - The current format works and is perfectly valid, but is not the exact form used in the Nuget docs. It should work by changing the None.BrowseObject.xaml, but that seems to be broken
- Adding item using NuGet.Build.Tasks.Pack.targets
  - It was decided to not include this behavior

For reviewing, there are some behaviors that are intended:
- If there is already a license file item, changing the license file will try to change a matching item
- The item metadata (PackagePath and Pack) is not editable in the UI
- The license file text box is not directly linked to the PackageLicenseFile property. It is supposed to display the relative path to the file. It is not editable.
- Switching between License File and License Expression will remove the text from the other TextBox, but not save the change
- Adding a property and then removing it will keep the property in the project file but it will be empty. This is not ideal, but it is consistent with the behavior of other properties

### Current version 
Initial State:
noneselected
Relative License File:
fileselected
License Expression:
licenseexpressionselected
Package License URL detected:
licenseurldetected


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
```diff
@@ -145,7 +145,7 @@
         </StringProperty.DataSource>
     </StringProperty>
     <BoolProperty Name="PackageRequireLicenseAcceptance" DisplayName="Package Require License Acceptance" Default="False" Visible="False"/>
-    <StringProperty Name="PackageLicenseUrl" DisplayName="Package License Url" Visible="False"/>
+    <StringProperty Name="PackageLicenseExpression" DisplayName="Package License Expression" Visible="False"/>
```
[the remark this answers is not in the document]
davidwengier: Yep, don't need DisplayName or Description for any non-visible properties.

karann-msft:
if it is not too much effort, may I recommend adding just one validation for the UI - block on save if the user has specified both a license expression and a license file.

etbyrd:
Would it be a good idea to have a radio button in that case? It would force the user to specify without having to do validation - and might be more intuitive

karann-msft:
actually, scratch that....we don't do any UI validations including when "Require license acceptance" is checked and the License URL is null. That validation is handled by the pack task. We can go the same route with this validation.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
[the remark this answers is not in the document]
etbyrd: Across all of the rules? Should I make a separate PR for that or put it in here?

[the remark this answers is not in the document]
etbyrd: Fixed in https://github.com/dotnet/project-system/pull/4270

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -231,7 +233,39 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         ''' </summary>
         Private Sub NeutralLanguageComboBox_DropDown(sender As Object, e As EventArgs) Handles NeutralLanguageComboBox.DropDown
             PopulateNeutralLanguageComboBox(NeutralLanguageComboBox)
-            Common.SetComboBoxDropdownWidth(NeutralLanguageComboBox)
+            SetComboBoxDropdownWidth(NeutralLanguageComboBox)
+        End Sub
+
+        Private Sub LicenseBrowseButton_Click(sender As Object, e As EventArgs) Handles LicenseBrowseButton.Click
+            Dim sInitialDirectory As String = Nothing
```
Pilchie: We don't typically use prefixes for simple data types like `String`.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -231,7 +233,39 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            Dim sFileName As String
+
+            If sInitialDirectory = "" Then
```
Pilchie: Since it's a local, `sInitialDirectory` will always be `Nothing` here.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -231,7 +233,39 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+                sFileName = ""
+                sInitialDirectory = ""
+            Else
+                sFileName = Path.GetFileName(sInitialDirectory)
+                sInitialDirectory = Path.GetDirectoryName(sInitialDirectory)
+            End If
+
+            Dim fileNames As ArrayList = GetFilesViaBrowse(ServiceProvider, Handle, sInitialDirectory, My.Resources.Microsoft_VisualStudio_Editors_Designer.PPG_AddExistingFilesTitle,
+                    CombineDialogFilters(
+                        CreateDialogFilter(My.Resources.Microsoft_VisualStudio_Editors_Designer.RSE_Filter_Text, "txt"),
```
Pilchie: What is the value of `RSE_Filter_Text` and `PPG_AddExistingFilesTitle` here?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -231,7 +233,39 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+                        GetAllFilesDialogFilter()
+                        ),
+                        0, False, sFileName)
+            If fileNames IsNot Nothing AndAlso fileNames.Count = 1 Then
+                sFileName = CStr(fileNames(0))
```
Pilchie: I would tend to use `DirectCast` here, not `CStr`

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -231,7 +233,39 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+                If File.Exists(sFileName) Then
+                    LicenseFileNameTextBox.Text = sFileName
+                    SetDirty(LicenseFileNameTextBox, True)
+                Else
+                    DelayValidate(LicenseFileNameTextBox)
```
Pilchie: What does `DelayValidate` do?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: These are for looking up the localized strings in the browse for file menu.
RSE_Filter_Text is "Text Files": 
textfile
PPG_AddExistingFilesTitle is "Add existing file to project":
filetitle

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Good point, fixed.

This code was borrowed for testing purposes from ApplicationPropPage.vb, which seems to not have changed in a meaningful way since 2003. I am going to clean up the whole function and I should probably make a separate PR for cleaning up the other PropPages as well.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Correct - and the whole if else block can be removed since the else block will never be reached...

karann-msft:
hey @etbyrd, just wanted to check-in on this item....can we push this into 16.0 Preview 2?

etbyrd:
@karann-msft I will be working on finishing this by the end of the week.

etbyrd:
Note: This is not quite ready for review.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -186,35 +241,35 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                     datalist.Add(data)
                     data = New PropertyControlData(105, "Copyright", Copyright, ControlDataFlags.None, New Control() {CopyrightLabel})
                     datalist.Add(data)
-                    data = New PropertyControlData(107, "PackageLicenseUrl", PackageLicenseUrl, ControlDataFlags.None, New Control() {PackageLicenseUrlLabel})
+                    data = New PropertyControlData(106, "PackageLicenseExpression", PackageLicenseExpression, ControlDataFlags.None, New Control() {PackageLicenseLabel})
```
drewnoakes: Should the control be renamed to `PackageLicenseExpressionLabel`?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -50,6 +81,32 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+        ''' <summary>
+        ''' This checks the properties for licenses to determine the initial state of the licensing section.
+        ''' Currently, If both are set, it will default to just enabling the PackageLicenseExpression.
+        ''' In the future, it might be a good idea to have a warning and only output one value, but the nuget error after packing is helpful.
+        ''' TryGetNonCommonPropertyValue will get the property if it is set even if it is empty, but because the empty properties 
+        ''' are ignored elsewhere, they are ignored here.
+        ''' </summary>
+        Private Sub InitializeLicensing()
+            Dim PackageLicenseFileSet = TryCast(TryGetNonCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile")), String)
+            If (PackageLicenseFileSet IsNot Nothing And PackageLicenseFileSet IsNot "") Then
```
drewnoakes: Consider using `AndAlso`. Not a big deal though.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -50,6 +81,32 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                 AssemblyVersionMajorTextBox, AssemblyVersionMinorTextBox, AssemblyVersionBuildTextBox, AssemblyVersionRevisionTextBox}
         End Sub
 
+        ''' <summary>
+        ''' This checks the properties for licenses to determine the initial state of the licensing section.
+        ''' Currently, If both are set, it will default to just enabling the PackageLicenseExpression.
```
drewnoakes: When referencing symbols in doc comments, link to types (eg. `<see cref="PackageLicenseExpression" />`) so that refactorings keep things up to date.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -50,6 +81,32 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+        ''' In the future, it might be a good idea to have a warning and only output one value, but the nuget error after packing is helpful.
+        ''' TryGetNonCommonPropertyValue will get the property if it is set even if it is empty, but because the empty properties
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,117 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+        End Sub
+
+        'These GotFocus methods are for when the property page is first entered and neither have a value
+        'It would make sense to selected the corresponding radio button when the text box recieves focus
+        Private Sub PackageLicenseExpression_GotFocus(sender As Object, e As EventArgs) Handles PackageLicenseExpression.GotFocus
+            SetLicenseRadioButtons(True)
+        End Sub
+
+        Private Sub LicenseFileNameTextBox_GotFocus(sender As Object, e As EventArgs) Handles LicenseFileNameTextBox.GotFocus
+            SetLicenseRadioButtons(False)
```
drewnoakes: Are you intentionally not calling `LicenseTypeFirstSelected` here?

If not, you could move the four identical lines in the radio button change handlers into `SetLicenseRadioButtons` directly.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,117 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            Dim projectSourceItemProvider = configuredProject.Services.ExportProvider.GetExportedValue(Of ProjectSystem.IProjectSourceItemProvider)()
+            Dim correctDirectory = Directory.GetParent(unconfiguredProject.FullPath).ToString
+            Dim relativePath As String = GetRelativePath(Path.GetFullPath(fileName), correctDirectory + "\")
+            LicenseFileNameTextBox.Text = relativePath
+            'The TextBox needs to have the relative path, so the property isn't linked to the TextBox. It must be set manually.
+            SetCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile"), Path.GetFileName(fileName))
+            ThreadHelper.JoinableTaskFactory.Run(Function()
+                                                     Return projectSourceItemProvider.AddAsync("None", relativePath, {(New KeyValuePair(Of String, String)("Pack", "True")), New KeyValuePair(Of String, String)("PackagePath", "")})
+                                                 End Function)
+        End Sub
```
drewnoakes: This sub has two calls to `JTF.Run`. Making a single call (and doing the intermediate work within the lamda) would reduce the overhead.

Alternatively make `AddLicenseItemToProject` async and move the JTF call to its call in the click handler below.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,117 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+
+        Private Sub LicenseBrowseButton_Click(sender As Object, e As EventArgs) Handles LicenseBrowseButton.Click
+            Dim fileName = ""
+            Dim fileNames As ArrayList = GetFilesViaBrowse(ServiceProvider, Handle, GetProjectPath(), My.Resources.Microsoft_VisualStudio_Editors_Designer.PPG_AddExistingFilesTitle,
+                    CombineDialogFilters(
+                        "License File (*.txt, *.md)|*.txt;*.md;",
```
drewnoakes: Trailing semicolon is not needed here, though probably harmless.

drewnoakes: Does `License File` need to be localised?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,117 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            LicenseFileNameTextBox.Text = relativePath
+            'The TextBox needs to have the relative path, so the property isn't linked to the TextBox. It must be set manually.
+            SetCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile"), Path.GetFileName(fileName))
+            ThreadHelper.JoinableTaskFactory.Run(Function()
+                                                     Return projectSourceItemProvider.AddAsync("None", relativePath, {(New KeyValuePair(Of String, String)("Pack", "True")), New KeyValuePair(Of String, String)("PackagePath", "")})
+                                                 End Function)
+        End Sub
+
+        Private Sub LicenseBrowseButton_Click(sender As Object, e As EventArgs) Handles LicenseBrowseButton.Click
+            Dim fileName = ""
```
drewnoakes: This can be declared in a tighter scope.

src/Microsoft.VisualStudio.Editors/PropPages/xlf/PackagePropPage.cs.xlf
```diff
@@ -137,6 +132,121 @@
...
+        <target state="new">&amp;File:</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="LicenseBrowseButton.Text">
+        <source>Browse...</source>
+        <target state="new">Browse...</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="Copyright.AccessibleName">
+        <source>Package Copyright</source>
```
drewnoakes: The items from here down do not have `&`-style accelerators. Are we supposed to add them?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
```diff
@@ -145,6 +145,9 @@
         </StringProperty.DataSource>
     </StringProperty>
     <BoolProperty Name="PackageRequireLicenseAcceptance" Default="False" Visible="False"/>
+    <StringProperty Name="PackageLicenseFileLocation" Visible="False"/>
```
drewnoakes: Is this needed? I don't see any usages.

etbyrd: No, that was left over from the pack target implementation.

src/Microsoft.VisualStudio.Editors/PropPages/xlf/PackagePropPage.cs.xlf
etbyrd: I don't believe so - These are auto generated.  Just to make sure though: @tmeschter

tmeschter: Don't worry about the accelerators in the .xlf files. However, have proper accelerators been defined for the newly added controls?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: I added it as a property, so it will be localized

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Good point, fixed.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Fixed.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: That is a good idea, I will change it

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: No, it was just referencing the wrong control. It should have been the expression label.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Great catch! Fixed.

src/Microsoft.VisualStudio.Editors/PropPages/xlf/PackagePropPage.cs.xlf
etbyrd: Working on that - I am having trouble even getting the old accelerators responding correctly

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,113 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            End If
+            LicenseFileRadioButton.Checked = Not setLicenseExpression
+            LicenseExpressionRadioButton.Checked = setLicenseExpression
+            PackageLicenseExpression.Enabled = setLicenseExpression
+            LicenseFileNameTextBox.Enabled = Not setLicenseExpression
+            LicenseBrowseButton.Enabled = Not setLicenseExpression
+            If setLicenseExpression AndAlso Not LicenseFileNameTextBox.Text = "" Then
+                SetCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile"), "")
+                LicenseFileNameTextBox.Text = ""
+            ElseIf (Not PackageLicenseExpression.Text = "") Then
```
davidwengier: I could be reading this wrong but if setLicenseExpression is true and the LicenseFileNameTextBox is empty, won't this clear an existing value in the license expression text box? Seems like thats unwanted if setLicenseExpression is true.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,113 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+
+        Private Sub LicenseFileNameTextBox_GotFocus(sender As Object, e As EventArgs) Handles LicenseFileNameTextBox.GotFocus
+            SetLicenseRadioButtons(False)
+        End Sub
+
+        Private Sub LicenseBrowseButton_GotFocus(sender As Object, e As EventArgs) Handles LicenseBrowseButton.GotFocus
+            SetLicenseRadioButtons(False)
+        End Sub
+
+        Private Shared Function GetRelativePath(fullPath As String, basePath As String) As String
```
davidwengier: Microsoft.VisualStudio.Editor, Utils.vb has a function for this.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,113 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            Return relativeUri.ToString().Replace("/", "\")
+        End Function
+
+        Private Sub AddLicenseItemToProject(fileName As String)
+            Dim unconfiguredProject = GetUnconfiguredProject(ProjectHierarchy)
+            Dim configuredProject As ProjectSystem.ConfiguredProject = ThreadHelper.JoinableTaskFactory.Run(Function()
+                                                                                                                Return unconfiguredProject.GetSuggestedConfiguredProjectAsync
+                                                                                                            End Function)
+            Dim projectSourceItemProvider = configuredProject.Services.ExportProvider.GetExportedValue(Of ProjectSystem.IProjectSourceItemProvider)()
+            Dim correctDirectory = Directory.GetParent(unconfiguredProject.FullPath).ToString
```
davidwengier: This is probably just a personal preference thing but if you just need the string, not the full DirectoryInfo, I prefer using Path.GetDirectoryName

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -50,6 +81,32 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+                SetLicenseRadioButtons(False)
+            End If
+            Dim PackageLicenseExpressionSet = TryCast(TryGetNonCommonPropertyValue(GetPropertyDescriptor("PackageLicenseExpression")), String)
+            If (PackageLicenseExpressionSet IsNot Nothing AndAlso PackageLicenseExpressionSet IsNot "") Then
+                _newLicensePropertyDetectedAtInit = True
+                SetLicenseRadioButtons(True)
+            End If
+            Dim PackageLicenseUrlSet = TryCast(TryGetNonCommonPropertyValue(GetPropertyDescriptor("PackageLicenseUrl")), String)
+            If (PackageLicenseUrlSet IsNot Nothing AndAlso PackageLicenseUrlSet IsNot "") Then
+                LicenseLineLabel.BackColor = Drawing.SystemColors.Control
```
davkean: What is this trying to do? Why are we changing the color?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -230,7 +285,113 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+                LicenseFileNameTextBox.Text = ""
+            ElseIf (Not PackageLicenseExpression.Text = "") Then
+                PackageLicenseExpression.Text = ""
+            End If
+        End Sub
+
+        Private Sub LicenseTypeFirstSelected()
+            'When the project has neither of the new license properties AND it has the license URL property and a new license type is selected
+            If (_licenseUrlDetected) Then
+                LicenseLineLabel.BackColor = Drawing.SystemColors.ControlDark
```
davkean: Why are you seeing the color?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: In some of the other property pages, the separators between sections is just an empty label with a ControlDark BackColor. I am using the same method here but I am also using it to display the deprecated warning for PackageLicenseUrl.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
davkean: The dependency on CPS here is not needed, and will actually cause troubles - there is no guarantee that the tree has updated with the new item by the time this call returns. Please just use Project.ProjectItem which makes it project system agnostic and guarantees that the tree has been updated.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Same reason as above, I am using that label as the separator and in the case that there is a LicenseUrl, it will display a warning.

davkean: This will not play nicely with screen readers - you are using color to hide the warning?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Good point. I'll change that.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Ah, didn't see that. Good catch.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: You are correct. I wasn't seeing that in my tests because there would never be any text in PackageLicenseExpression if setLicenseExpression was true because it was already handled. But, I am going to change the check so it can be used in the future without confusion.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: I am not sure that this will work - Nuget Pack needs an item in the project file. It doesn't seem there is a way for Project.ProjectItem.AddItem does not generate an item in the project file. If it can generate the item, is there a way to change the metadata on the item? Pack also needs specific metadata. From a cursory glace, it seems like a ProjectItem's ProjectPropertiesContext does not have metadata. 

Also, the desired behavior is that the item is not copied to project.

src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/Microsoft.VisualStudio.ProjectSystem.CSharp.VS.csproj
```diff
@@ -1,20 +1,23 @@
...
+  <Import Project="..\VisualStudio.props" />
   <PropertyGroup>
     <RootNamespace>Microsoft.VisualStudio</RootNamespace>
     <!-- VSIX -->
     <GeneratePkgDefFile>true</GeneratePkgDefFile>
     <CreateVsixContainer>false</CreateVsixContainer>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <PlatformTarget>x86</PlatformTarget>
+  </PropertyGroup>
```
jjmew: Was this intentional?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -388,26 +387,19 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
-                                Next
-                            Next
-                            If (Not foundAndUpdated) Then
+                            Dim foundItem = projectXML.ItemGroups.SelectMany(Function(x) x.Items).FirstOrDefault(Function(x) x.Include = oldInclude)
+                            If foundItem IsNot Nothing Then
+                                foundItem.Include = newInclude
+                            Else
                                 'We couldn't find one to change so we should add it 
-                                projectXML.AddItem("None", newInclude, {(New KeyValuePair(Of String, String)("Pack", "True")), New KeyValuePair(Of String, String)("PackagePath", "")})
+                                projectXML.AddItem("None", newInclude, {New KeyValuePair(Of String, String)("Pack", "True"), New KeyValuePair(Of String, String)("PackagePath", "")})
```
davkean: This logic of modifying the project XML does not belong in the UI layer. If we make changes to how things are written to the project file (for example, when we changed from elements -> attributes) this will be affected. 

The relationship here is that the UI sets/reads properties from the project system via property APIs, and the project system is responsible for mapping the XML => how its represented in-memory. I get this late now, but post Preview 4, please move this logic behind a project property (in a rule file with a datasource) and have that property read/write the item.

davkean:
Can you please try the following scenarios which I think will be broken:

- Use a $(Property) inside the item include
- Put the item inside a targets file and try and edit the property.
- Open this page, edit the project file manually (or switch branches) to add a license file and come back to this page. Does it pick it?
- Open this page, edit the project file manually (or switch branches) to remove a license file and come back to this page, does it pick up the removal?

etbyrd:
* Use a $(Property) inside the item include
  - This works - the full file path still needs to include the PackageLicenseFile because the text box is populated with the first item that has the same file name as PackageLicenseFile in the include. For example these would work:
`<FileName>LICENSE.txt</> <PackageLicenseFile>$(FileName)</> <FullPackagePath>..\FolderName\$(PackageLicenseFile)</> <ItemGroup> <None Include="$(FullPackagePath)"> ... </>`
* Put the item inside a targets file and try and edit the property.
  - It will read the item just fine, but it will not update the item in the targets file - it just adds a new one to the .csproj. This matches the behavior for the all of the other properties in the package property page.
* Put a '&' in the folder name of the license file.
  - You can't just put a & in the .csproj - it isn't valid. However:
  - When picking a file, assuming the folder exists and the file exists, this works fine. It will add a `&amp;` to the path
  - Putting a `&amp;` in the `<PackageLicenseFile>` and the `<PackagePath>` will pack as expected
* Open this page, edit the project file manually (or switch branches) to add a license file and come back to this page. Does it pick it?
  - No, the project needs to be unloaded and reloaded. This matches the behavior of all of the other properties in the package property page
* Open this page, edit the project file manually (or switch branches) to remove a license file and come back to this page, does it pick up the removal?
  - Same as above

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Understood. I will get a PR out to update this behavior.

src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/Microsoft.VisualStudio.ProjectSystem.CSharp.VS.csproj
etbyrd: oops - that was not intentional.

davkean:
> No, the project needs to be unloaded and reloaded. This matches the behavior of all of the other properties in the package property page

This is a regression - please assume this is supposed to work while we figure out the regression.

davkean:
I filed https://github.com/dotnet/project-system/issues/4549 for the regression, this was regressed in 16.0.

etbyrd:
> This is a regression - please assume this is supposed to work while we figure out the regression.

I see. Should this block merging?

etbyrd:
Modifying the project file is now reflected correctly in the UI

jjmew:
I am seeing the error below each time I open the properties pages. Ethan is investigating. 

image

etbyrd:
This assert failure is expected - PackageLicenseFile is not linked to a control, and it is not hidden, so it cannot be handled automatically. It must be handled manually, which is how it is designed. The user does not set the PackageLicenseFile directly - they pick a relative path to a file and the PackageLicenseFile is determined from that. Then, the textbox displays the relative path to the file. Every other property is directly linked to a control and updated automatically.

PR #4224, opened 2018-11-06
Also publish the raw test result log files directory

Pilchie:
Against dev16.0.x so that if any other PRs to that run into trouble we'll already have it there.

### DIFF
--- a/.vsts-dnceng.yml
+++ b/.vsts-dnceng.yml
@@ -25,6 +25,13 @@ phases:
         publishLocation: Container
       continueOnError: true
       condition: failed()
+    - task: PublishBuildArtifacts@1
+      inputs:
+        PathtoPublish: '$(Build.SourcesDirectory)\artifacts\$(_configuration)\TestResults'
+        ArtifactName: '$(_configuration) Test Result Logs'
+        publishLocation: Container
+      continueOnError: true
+      condition: failed()
     - task: PublishTestResults@2
       inputs:
         testRunner: 'xUnit'


drewnoakes:
Ironically we won't know if this PR works unless it fails :)

Once merged I'll merge it into #4199 and take it for a spin.

Pilchie:
Merging infra only change.

Pilchie:
@drewnoakes This isn't against master yet, so you might want to `cherry-pick` the commit into your PR to get results before the automerge happens.

drewnoakes:
Worked perfectly. I can now see what the error is, thank you 🎉

Pilchie:
@jmarolf Do you know the right contacts in ADO to give some feedback about this?  Ideally it would work with just the test result publishing as well...

jmarolf:
I don't know anyone over there but I'll start a mail with @jaredpar.  This seems like a bug on their side.

jmarolf:
Looking at the problem it does not appear to be ADO's fault.   Here is the log:

```ini
xUnit.net Console Runner v2.4.0 (32-bit Desktop .NET 4.5.2, runtime: 4.0.30319.42000)
  Discovering: Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests
  Discovered:  Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests
  Starting:    Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests
Microsoft.VisualStudio.Diagnostics.ThrowingTraceListener+DebugAssertFailureException: Instance of CancellationSeries not disposed before being finalized

   en Microsoft.VisualStudio.Diagnostics.ThrowingTraceListener.Fail(String message, String detailMessage) en F:\vsagent\2\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.TestServices\Diagnostics\ThrowingTraceListener.cs:l¡nea 25
   en System.Diagnostics.TraceListener.Fail(String message)
   en System.Diagnostics.TraceInternal.Fail(String message)
   en System.Diagnostics.Debug.Assert(Boolean condition, String message)
   en Microsoft.VisualStudio.Threading.Tasks.CancellationSeries.Finalize() en F:\vsagent\2\s\src\Microsoft.VisualStudio.ProjectSystem.Managed\Threading\Tasks\CancellationSeries.cs:l¡nea 36
=== COMMAND LINE ===
"C:\Users\vsagent\.nuget\packages\xunit.runner.console\2.4.0\tools\net452\xunit.console.x86.exe" "F:\vsagent\2\s\artifacts\Debug\bin\UnitTests\Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests.dll" -noshadow -xml "F:\vsagent\2\s\artifacts\Debug\TestResults\Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests_net461_x86.xml" -html "F:\vsagent\2\s\artifacts\Debug\TestResults\Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests_net461_x86.html"
```

`ThrowingTraceListener` is throwing asynchronously and bringing down the test process so no xml or html logs are written out?  At least I do not see `Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests_net461_x86.xml` written to the artifacts directory.

drewnoakes:
That's an assertion thrown during the execution of tests. That class validates it is disposed in debug builds.

It gets thrown at some later time, during finalisation, which might cause the test runner to consider this an internal error. I'm not sure how isolated this is.

Am considering removing the feature however... have pushed another commit that'll include the stack at the time the object was allocated to help identify which usage is causing the failure, then I'll decide.

jaredpar:
> ThrowingTraceListener is throwing asynchronously and bringing down the test process so no xml or html logs are written out? 

Agree that seems to be the problem. If the process crashes then xunit won't emit any logs. 

> Am considering removing the feature however...

Which feature: ThrowingTraceListenter or the feature that triggered this failure?

drewnoakes:
> Which feature: ThrowingTraceListenter or the feature that triggered this failure?

The feature that triggered the failure.

It is a `DEBUG` only assertion in the finalizer of a type that should always be disposed. It's a nice idea in theory, but debugging such failures can be painful as they're non-deterministic and don't drop useful logs.

https://github.com/dotnet/project-system/blob/4f563eff786e99dad55d7e4b780a798d353d0e29/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs#L37-L49

As it stands I weeded out all the violations and fixed them. Future violations, when we've long forgotten about this, might evoke some cursing.

jaredpar:
Yeah you should definitely remove that assert ... because it fails to use interpolated strings 😄

drewnoakes:
Consider me an optimist 😃

davkean:
ThrowingTraceListener exists to avoid getting dialog prompts, which will result in the result except the process will be torn down after 60 minutes.

PR #4226, opened 2018-11-06
Passing GeneratePathProperty metadata as part of nomination

jainaashish:
AS part of https://github.com/NuGet/NuGet.Client/pull/2271 We added a new feature to generate msbuild property to nuget package path if someone set `GeneratePathProperty` on a `PackageReference`. But Project system wasn't updated to pass this new metadata to NuGet as part of nomination.

This PR is to pass that and pass this new metadata to NuGet.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
@@ -84,10 +84,10 @@
                     DisplayName="Suppress warnings"
                     Description="Comma-delimited list of warnings that should be suppressed for this package" />
 
-    <StringProperty Name="GeneratePathProperty" 
+    <BoolProperty Name="GeneratePathProperty" 
                     Visible="True"
-                    DisplayName="Generate Path Property"
-                    Description="Generates msbuild property to the path of this reference" />
+                    DisplayName="Generate path property"
+                    Description="Indicates whether to generate an MSBuild property with the location of the package's root directory. The generated property name is in the form of 'Pkg[PackageID]', where '[PackageID]' is the ID of the package with any periods '.' replaced with underscores '_'." />
 
     <BoolProperty Name="Visible"
                   Visible="False"
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
@@ -84,6 +84,11 @@
                     DisplayName="Suppress warnings"
                     Description="Comma-delimited list of warnings that should be suppressed for this package" />
 
+    <StringProperty Name="GeneratePathProperty" 
+                    Visible="True"
+                    DisplayName="Generate Path Property"
+                    Description="Generates msbuild property to the path of this reference" />
+
     <BoolProperty Name="Visible"
                   Visible="False"
                   ReadOnly="True" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
@@ -112,11 +112,6 @@
                     Description="Version of reference.">
     </StringProperty>
 
-    <StringProperty Name="GeneratePathProperty" 
-                    Visible="True"
-                    DisplayName="GeneratePathProperty"
-                    Description="Generates msbuild property to the path of this reference" />
-
     <!-- Hidden properties -->
     <BoolProperty Name="LinkLibraryDependencies"
                   Visible="False" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
@@ -112,6 +112,11 @@
                     Description="Version of reference.">
     </StringProperty>
 
+    <StringProperty Name="GeneratePathProperty" 
+                    Visible="True"
+                    DisplayName="GeneratePathProperty"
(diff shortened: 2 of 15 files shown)


jeffkl:
@jainaashish it looks like you need to update the XLF

```
'ProjectSystem\Rules\xlf\ProjectReference.xaml.cs.xlf' is out-of-date with 'ProjectSystem\Rules\ProjectReference.xaml'. Run `msbuild /t:UpdateXlf` to update .xlf files or set UpdateXlfOnBuild=true to update them on every build
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
```diff
@@ -112,6 +112,11 @@
                     Description="Version of reference.">
     </StringProperty>
 
+    <StringProperty Name="GeneratePathProperty"
```
davkean: What does this represent? This name is not clear to me.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
```diff
@@ -112,6 +112,11 @@
                     Description="Version of reference.">
     </StringProperty>
 
+    <StringProperty Name="GeneratePathProperty" 
+                    Visible="True"
+                    DisplayName="GeneratePathProperty"
```
davkean: Don't specific DisplayName if its the same as name.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
jainaashish: this generates msbuild property `Pkg<Package_Name>` which points to the package root path. This is what I've tried to explain in description. Let me know if you've better text.

`Generates msbuild property to the path of this reference`

jainaashish:
@davkean `ProjectReference` was updated by mistake so I've corrected that. I wasn't able to update XLF since `msbuild /t:UpdateXlf` or `build.cmd` both are failing to build stating missing settings. So is there other way to generate these?

davkean:
@jainaashish Type "build.cmd".

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ProjectReference.xaml
davkean: "Indicates whether to generate an MSBuild property with the location of the package's root directory. The generated property name is in the form of 'Pkg[PackageID]', where '[PackageID]' is the ID of the package with any periods '.' replaced with underscores '_'."

davkean:
This should also be `<BoolProperty/>` not `<StringProperty`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
```diff
@@ -85,6 +85,11 @@
                     DisplayName="NoWarn"
                     Description="Comma-delimited list of warnings that should be suppressed for this package" />
 
+    <StringProperty Name="GeneratePathProperty" 
+                    Visible="True"
+                    DisplayName="Generate Path Property"
```
davkean: "Generate path property" I know that PackageReference isn't consistent here, I've filed https://github.com/dotnet/project-system/issues/4231 to fix this.

jainaashish:
`This should also be <BoolProperty/> not <StringProperty.`

What exactly is the diff? I saw some existing Boolean property like `IsImplicitlyDefined` are also defined as `StringProperty`

davkean:
It controls how it is persisted and displayed in the Properties window. IsImplicitlyDefined isn't displayed in the UI.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/xlf/PackageReference.xaml.cs.xlf
```diff
@@ -82,6 +82,16 @@
         <target state="translated">Umístění odkazovaného balíčku</target>
         <note />
       </trans-unit>
+      <trans-unit id="BoolProperty|GeneratePathProperty|DisplayName">
```
Pilchie: Tagging @tmeschter.  Note that we don't currently have any more loc handbacks planned for 15.9, so this string would remain untranslated until Dev16.

tmeschter: If the display name or description are going to be visible to users then we can't take this change. Loc cut-off was over a week ago.

davkean:
@tmeschter Too late for what release?

tmeschter:
@davkean 15.9 RTW.

davkean:
Oh, completely missed the branch here. @jainaashish Is this deliberate for 15.9? We'll need approval for the change itself and loc changes. To avoid approval for the loc changes, we can turn off visible, and then turn on visible in master. 

//cc @Pilchie

jainaashish:
@Pilchie already mentioned about these loc changes and the fact that they will only be localized in dev16. We'll mention that once we ask for the approval for 15.9.1

dasMulli:
But even having it in there with Visible="false" would fix the VS restore?

davkean:
Yes.

davkean:
@jainaashish What's going on with this PR?

jainaashish:
Last I knew, we had to bring it for 15.9.x approval but afterwards I don't know what happened. @rrelyea what happened to this?

bill-msft:
Happy new year! Has there been any progress on this?

jainaashish:
Rebased this PR to master branch for dev16. @tmeschter can you unblock the PR to merge?

jainaashish:
@davkean can you complete this PR?

KirillOsenkov:
There is tremendous value in this PR. It would be nice to ship this soon.

bill-msft:
Any update on this? It looks like the build agent shut down in the middle of testing.

Pilchie:
Close and re-open to kick the CI.

PR #4228, opened 2018-11-07
Wrap context usage in a batch

davkean:
Roslyn will avoid updating VisualStudioWorkspace until after a batch has finished.


davkean:
Note, I had to redo this commit (#834b4e8) as I accidentally blew it away in an update.

PR #4230, opened 2018-11-07
Remove usage of AbstractProject

davkean:
We can use the new IWorkspaceProjectContext.Id property.


davkean:
The test that is failing in Spanish, will be fixed by this change: https://github.com/dotnet/project-system/pull/4229, so merging in.

PR #4235, opened 2018-11-07
adding public api analyzer

jmarolf:
This adds public api checking for the assemblies that we ship

NOTE: since we've already shipped most of these I am marking them as `Shipped`. In the future new APIs should be added to the `Unshipped` file until released to the public


Pilchie:
cc @abpiskunov - here's where we added public API analyzers for `project-system`. Roslyn has had them for years.

abpiskunov:
Awesome, thanks!

drewnoakes:
This looks really great @jmarolf 👍

jmarolf:
@Pilchie yeah, can you file a bug to review this at some point? we should rationalize what we are currently shipping.  Dev16.0 would be the time if there is something bad we wanted to remove.

jmarolf:
@davidwengier yep, adding those files "turns-on" the analyzer

PR #4242, opened 2018-11-08
Reduce work and allocations in dependency node

drewnoakes:
Pushing these small changes ahead of a larger PR based off this, to make the next PR easier to review.

---

Out of curiousity I loaded `Roslyn.sln`, performed a search in the solution, and counted the number of each kind of `DependencyModel` subclass created. Results:

| Type | Count |
|----|----:|
| `AssemblyDependencyModel` | 17,315 |
| `PackageDependencyModel` | 14,819 |
| `PackageDependencyModel` | 7,764 |
| `ProjectDependencyModel` | 3,756 |
| `AnalyzerDependencyModel` | 2,238 |
| `SdkDependencyModel` | 1,218 |
| `PackageAnalyzerDependencyModel` | 1,119 |
| `SharedProjectDependencyModel` | 10 |
| `SubTreeRootDependencyModel` | 5 |

Note this includes the change in 31813b78708fb9f498c58a2e5f8c31e6099a1b50 where root nodes are reused (hence the 5). That number would probably be just under 1,000 otherwise.

The gains here are modest, but they keep adding up.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependencyModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependencyModel.cs
@@ -37,6 +37,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Models
 
             if (isImplicit)
             {
+                // Cannot remove implicit dependencies
                 Flags = Flags.Except(DependencyTreeFlags.SupportsRemove);
             }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependencyModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependencyModel.cs
@@ -9,44 +9,63 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Models
 {
     internal abstract class DependencyModel : IDependencyModel
     {
+        [Flags]
+        private enum DependencyFlags
+        {
+            Resolved = 1 << 0,
+            TopLevel = 1 << 1,
+            Implicit = 1 << 2,
+            Visible  = 1 << 3
+        }
+
         protected DependencyModel(
             string path,
             string originalItemSpec,
             ProjectTreeFlags flags,
-            bool resolved,
+            bool isResolved,
             bool isImplicit,
-            IImmutableDictionary<string, string> properties)
+            IImmutableDictionary<string, string> properties,
+            bool isTopLevel = true,
+            bool isVisible = true)
         {
             Requires.NotNullOrEmpty(path, nameof(path));
 
             Path = path;
             OriginalItemSpec = originalItemSpec ?? path;
-            Resolved = resolved;
-            Implicit = isImplicit;
             Properties = properties ?? ImmutableStringDictionary<string>.EmptyOrdinal;
             Caption = path;
 
-            if (resolved)
-            {
-                Flags = flags.Union(DependencyTreeFlags.GenericResolvedDependencyFlags);
-            }
-            else
-            {
-                Flags = flags.Union(DependencyTreeFlags.GenericUnresolvedDependencyFlags);
-            }
+            flags += isResolved 
+                ? DependencyTreeFlags.GenericResolvedDependencyFlags 
+                : DependencyTreeFlags.GenericUnresolvedDependencyFlags;
 
             if (isImplicit)
             {
                 // Cannot remove implicit dependencies
-                Flags = Flags.Except(DependencyTreeFlags.SupportsRemove);
+                flags -= DependencyTreeFlags.SupportsRemove;
             }
 
-            if (Properties.TryGetValue("Visible", out string visibleMetadata)
-                && bool.TryParse(visibleMetadata, out bool visible))
+            Flags = flags;
+
+            if (Properties.TryGetValue("Visible", out string visibleString)
+                && bool.TryParse(visibleString, out bool visibleBool))
             {
-                Visible = visible;
+                isVisible = visibleBool;
             }
+
+            DependencyFlags depFlags = 0;
+            if (isResolved)
+                depFlags |= DependencyFlags.Resolved;
+            if (isVisible)
+                depFlags |= DependencyFlags.Visible;
+            if (isImplicit)
+                depFlags |= DependencyFlags.Implicit;
+            if (isTopLevel)
+                depFlags |= DependencyFlags.TopLevel;
+            _flags = depFlags;
         }
+
+        private readonly DependencyFlags _flags;
 
         public abstract string ProviderType { get; }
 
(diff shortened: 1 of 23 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+            projectRuleSnapshot.Items.TryGetValue(itemSpec, out IImmutableDictionary<string, string> properties);
+
+            return properties;
+        }
+
+        public static bool TryGetStringProperty(this IImmutableDictionary<string, string> properties, string key, out string stringValue)
+        {
+            if (properties != null &&
+                properties.TryGetValue(key, out stringValue) &&
+                !string.IsNullOrEmpty(stringValue))
```
jmarolf: Why would we have a null or empty string in there to begin with?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <param name="itemSpec">The unresolved reference item name.</param>
+        /// <returns>The key is item name and the value is the metadata dictionary.</returns>
+        public static IImmutableDictionary<string, string> GetProjectItemProperties(this IProjectRuleSnapshot projectRuleSnapshot, string itemSpec)
+        {
+            Requires.NotNull(projectRuleSnapshot, nameof(projectRuleSnapshot));
+            Requires.NotNullOrEmpty(itemSpec, nameof(itemSpec));
+
+            projectRuleSnapshot.Items.TryGetValue(itemSpec, out IImmutableDictionary<string, string> properties);
+
+            return properties;
```
jmarolf: Do we want to explicitly return null if `TryGetValue` fails?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
drewnoakes: I'm not sure why, but this check is pretty common amongst existing code that reads properties from dictionaries. Would `<EmptyElement></EmptyElement>` not produce an empty property?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
drewnoakes: If `TryGetValue` fails, `properties` is definitely `null`. Are you saying the code could be clearer?

jmarolf: Yes, `properties` being bull is an implementation detail.  I assume we want to be explicit that we intend to return null

---
In reply to: [231938165](https://github.com/dotnet/project-system/pull/4242#discussion_r231938165) [](ancestors = 231938165)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
jmarolf: That still doesn't explain the null case for me

---
In reply to: [231937916](https://github.com/dotnet/project-system/pull/4242#discussion_r231937916) [](ancestors = 231937916)

jmarolf: And if the property is an empty string why is it incorrect to return it?

---
In reply to: [231938983](https://github.com/dotnet/project-system/pull/4242#discussion_r231938983) [](ancestors = 231938983,231937916)

drewnoakes: What would be better here?

jmarolf: I don't know, I am moistly asking about the intended semantics of this method.  Its clear that if we have a value for a key that is null or an empty string we consider that the wrong result and return false, but it is unclear to me why 

---
In reply to: [231939349](https://github.com/dotnet/project-system/pull/4242#discussion_r231939349) [](ancestors = 231939349)

drewnoakes: The intent of the `Get*` variants is that they return `null` if no sensible value could be obtained, and the caller can then `??` a default in. It's the more convenient form. You can `?? string.Empty` if you want.

I'll document these methods.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
drewnoakes: I'll make it explicit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/MetadataExtensions.cs
jmarolf: Yes but if someone added an empty string or even null intentially to this dictionary why are those values not sensible? Essentially I am asking why this is wrong:

```C#
if (properties != null && properties.TryGetValue(key, out stringValue) {
   return true;
}
```

drewnoakes: The decision was based on existing usages. Many test for empty strings and do something else if they're observed. This just makes that simpler. My understanding is that in MSBuild and empty string is equivalent to a null, as far as property values are concerned. Correct me if I'm mistaken.

drewnoakes:
@jmarolf I've addressed your feedback. Thanks.

@all is there a reason we can't bump the langversion to 7.3? We could use the `Enum` type constraint.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependencyModel.cs
```diff
@@ -9,45 +9,64 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Models
 {
     internal abstract class DependencyModel : IDependencyModel
     {
+        [Flags]
+        private enum DependencyFlags
```
Pilchie: Not sure if it would help due to field packing, but consider making this have an underlying type of `byte`?

jmarolf:
@Pilchie they have not. We either need newer CI machines or move our build to be completely independent of the VS compiler and targets

Pilchie:
> move our build to be completely independent of the VS compiler and targets

Looking forward to .NET Core 3 for WPF support for this part..

PR #4246, opened 2018-11-08
More reduction in work/allocation within dependency node

drewnoakes:
Follows #4242.

See individual commits for details.

Main points:

- c938fd29dfbcbe59d68c44790122ef3fa5a47128 Fixed a potential NRE 
- f5b3c92bee9930e942d43fa6b8dc2f307d3717b1 Cache `ProjectTreeFlags` to avoid immutable set manipulations that always produce the same results
- ce29706ec010690feaeedaaec5f6986aaaae3153 Inline `PackageDependencyMetadata` in `PackageRuleHandler` which makes the logic clearer and helped avoid allocation and redundant work
- 651dc3c389f9e26c91bbf5fedd66e8ee29ac068d Reuse `LazyStringSplit` to avoid array/string allocations

I stumbled upon what I think is a nice alternative to `Lazy<T>` that doesn't allocate, for cases where you're on the stack (in f714231e1711c83f81cacf8e38049343672061e6). Finally found a use for a custom `ref struct`! If it turns out to be useful elsewhere I'll pull it out and document it. **EDIT** `StackLazy<T>` ends up breaking if the user copies it, and there's [no language support](https://github.com/dotnet/csharplang/issues/859#issuecomment-437331337) to prevent copying. Using external storage via ref fields only works for reference types. I've removed this for now but may revisit it in future.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DependencyModel.cs
```diff
@@ -19,6 +19,29 @@ private enum DependencyFlags : byte
...
+                _lookup = new ProjectTreeFlags[4];
+                _lookup[Index(true,  false)] = resolved;
+                _lookup[Index(true,  true )] = resolved.Except(DependencyTreeFlags.SupportsRemove);
+                _lookup[Index(false, false)] = unresolved;
+                _lookup[Index(false, true )] = unresolved.Except(DependencyTreeFlags.SupportsRemove);
+            }
+
+            public ProjectTreeFlags Get(bool isResolved, bool isImplicit) => _lookup[Index(isResolved, isImplicit)];
+
+            private static int Index(bool isResolved, bool isImplicit) => (isResolved ? 2 : 0) | (isImplicit ? 1 : 0);
```
davidwengier: It might just be me, but this is very difficult to understand. I don't think some comments or maybe even an if statement or two, would hurt here. A binary literal perhaps?

drewnoakes: Thanks. Added some documentation.

PR #4254, opened 2018-11-09
Don't allocate dependency model just to remove it

drewnoakes:
Includes #4246.

Exclusive diff: https://github.com/dotnet/project-system/compare/9625a97^..0acec3f

Only review 9625a97d75852ef0b36d7c65bc8614bcaf626957, 8ba54f9b3a3410f8ea235d4d2ec9712681d0cf15, f658bd978115fd2ba25003dccec157fb30dd18da and 0acec3fdbe6b776699e88ad144ebdca55e35f836.

The main change here is in 0acec3fdbe6b776699e88ad144ebdca55e35f836 where property `IDependenciesChanges.RemovedNodes` is changed from `IImmutableList<IDependencyModel>` to `IImmutableList<(string providerType, string dependencyId)>`.

Previously whenever a dependency was removed, we'd create a whole `IDependencyModel` instance to add to `IDependenciesChanges`. Removing a dependency only requires its key (provider type and ID).

Seeing as this public API changed earlier this week and WebTools is going to update, it makes sense to try and roll this change in as well.

//cc: @abpiskunov


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IDependenciesChanges.cs
```diff
@@ -10,6 +10,6 @@ public interface IDependenciesChanges
 
         ImmutableArray<IDependencyModel> AddedNodes { get; }
 
-        ImmutableArray<IDependencyModel> RemovedNodes { get; }
+        ImmutableArray<(string providerType, string dependencyId)> RemovedNodes { get; }
```
jmarolf: Well I guess we get to have the "should we use tuples in a public api" discussion :)

I would prefer if we defined our own stuct to contain this infromation

jmarolf:
> Could someone please summarise the rationale here for me?

Yeah, my 2 cents.  

Tuples are great for storing data that you are passing around as part of a computation but they are not great if you a public API because:
- You cannot attach doc comments to anything so a user of your type has no way to understand what, for example, `dependencyId` means.
- Adding a tuple element is _always_ a breaking change because it completely changes the type identity (get hash code is generated differently etc.) where are a custom type could be evolved without being a breaking change
- (**Not Applicable in this case**) `ValueTuple` is not something that we did a great job of shipping. In some cases the consumer needs to add a nuget package, in others they do not.  If we were shipping to a broader set of consumers this would be adding unnecessary friction for them as they may need to change their nuget dependency graph to consume this API, or at least need to reason about what this dependency means for their project.

drewnoakes:
More discussion about using `ValueTuple` in public APIs here https://github.com/dotnet/corefx/issues/33553

PR #4256, opened 2018-11-09
Change case of display names in PackageReference XAML

### LINKED ISSUE #4231: PackageReference isn't consistent with ProjectReference in it's display name
Properties that are displayed in the properties window for the new project system are "Sentence cased". PackageReference uses "PasalCased". We should fix that.

wjk:
The user-visible display names now use sentence case. Fixes #4231.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
@@ -29,17 +29,17 @@
 
     <StringProperty Name="IncludeAssets" 
                     Visible="True"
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PackageReference.xaml
```diff
@@ -82,7 +82,7 @@
 
     <StringProperty Name="NoWarn" 
                     Visible="True"
-                    DisplayName="NoWarn"
+                    DisplayName="Warnings to suppress"
```
davkean: "Suppress warnings" to be consistent with "Build" properties.

davkean:
Thanks for this, just a single comment.

PR #4262, opened 2018-11-12
Cache ImageMoniker strings, reducing GC heap by 0.5%

drewnoakes:
VS `GraphNode`s specify their icons as `string`. This string is created from an `ImageMoniker`'s Guid and integer ID.

There are many more graph nodes than distinct icons, hence creating a new string for each node results in a large number of duplicate string objects on the heap. This change results in strings being reused.

Loading `Roslyn.sln` and performing a search in the solution explorer produced the following counts:

|Icon String|Count|Wasted Bytes|
|----|----:|-----------:|
|`ae27a6b0-e345-4288-96df-5eaf394ee369;3644`|12,210|1,172,064|
|`259567c1-aa6b-46bf-811c-c145dd9f3b48;3`|11,491|1,057,080|
|`259567c1-aa6b-46bf-811c-c145dd9f3b48;11`|5,617|516,672|
|`259567c1-aa6b-46bf-811c-c145dd9f3b48;9`|1,423|130,824|
|`259567c1-aa6b-46bf-811c-c145dd9f3b48;1`|1,368|125,764|
|`259567c1-aa6b-46bf-811c-c145dd9f3b48;0`|91|8,280|
|**Total**|**32,200**|**3,010,684**|

Caching these strings reduced the managed heap by 0.5% in the above run.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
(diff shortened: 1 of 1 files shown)


davkean:
Does it make sense to cache these at the DependencyIconSet level?

drewnoakes:
> Does it make sense to cache these at the DependencyIconSet level?

Are you talking about caching the strings as properties on `DependencyIconSet`, or moving the cache logic?

The former would require a wider change across a range of types (`IDependencyViewModelExtensions`, `IDependencyViewModel`, ...) because of [this call](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs#L265). I think the API would end up in a worse place.

The latter is easy to do.

I'm starting to look at the graph provider in depth and have observed some other perf opportunities here.

PR #4266, opened 2018-11-13
Implement RunCodeAnalysis property to prevent watsons

davidwengier:
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/680205

This at least solves the 99% case of this property always being hit for SDK projects and correctly handles returning false for the empty string default value. In testing even when RunCodeAnalysis is specifically set to true, none of the other CodeAnalysis* properties are hit.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs
@@ -124,6 +124,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties
                 });
             }
         }
+        public bool RunCodeAnalysis
+        {
+            get
+            {
+                return _threadingService.ExecuteSynchronously(async () =>
+                {
+                    ConfiguredBrowseObject browseObjectProperties = await _projectProperties.GetConfiguredBrowseObjectPropertiesAsync();
+                    string value = await browseObjectProperties.RunCodeAnalysis.GetEvaluatedValueAtEndAsync();
+                    return string.Equals(value, bool.TrueString, StringComparisons.PropertyValues);
+                });
+            }
+            set
+            {
+                _threadingService.ExecuteSynchronously(async () =>
+                {
+                    ConfiguredBrowseObject browseObjectProperties = await _projectProperties.GetConfiguredBrowseObjectPropertiesAsync();
+                    await browseObjectProperties.RunCodeAnalysis.SetValueAsync(value);
+                });
+            }
+        }
+
         public object ExtenderNames => null;
         public string __id => throw new System.NotImplementedException();
         public bool DebugSymbols { get => throw new System.NotImplementedException(); set => throw new System.NotImplementedException(); }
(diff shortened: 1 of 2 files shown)


davidwengier:
@Pilchie The work item said this is "Valid for 16 preview 1". I'll leave it to you to make that ruling, I based the change off d16 so its easy enough to retarget the PR.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralConfiguredBrowseObject.xaml
```diff
@@ -33,6 +33,7 @@
     </StringProperty.DataSource>
   </StringProperty>
 
+  <BoolProperty Name="RunCodeAnalysis" DisplayName="Run Code Analysis" Visible="False" />
```
davkean: I know you're being consistent, but non-visible properties don't need to have DisplayName.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs
```diff
@@ -124,6 +124,27 @@ public string IntermediatePath
...
         }
+        public bool RunCodeAnalysis
+        {
+            get
+            {
+                return _threadingService.ExecuteSynchronously(async () =>
+                {
+                    ConfiguredBrowseObject browseObjectProperties = await _projectProperties.GetConfiguredBrowseObjectPropertiesAsync();
+                    string value = await browseObjectProperties.RunCodeAnalysis.GetEvaluatedValueAtEndAsync();
+                    return string.Equals(value, bool.TrueString, StringComparisons.PropertyValues);
```
davkean: GetValueAsync will return a (bool) here.

davidwengier: Yeah, I wasn't 100% sure about this so I went with the assumption that the since all of the other properties in this file use `EvaluatedValueAtEnd` that it was more important that worrying too much about a conversion. Easy to change though.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralConfiguredBrowseObject.xaml
davidwengier: I'll come through and clean all of these up eventually anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs
davkean: GetValue is probably the correct way to read these properties anyway - as the properties could be marked as "BeforeContext" (it's value as defined in the project file) and "AfterContext" (EvaluatedValueAtEnd) and we should be respecting this.

davidwengier: Updated to use GetValueAsync. It indeed returns a `bool?`.

Pilchie:
> The work item said this is "Valid for 16 preview 1". I'll leave it to you to make that ruling, I based the change off d16 so its easy enough to retarget the PR.

Forgot to reply to this part - I'm okay deferring this to Preview 2.

Issue #4269, opened 2018-11-13
Dependency upon order of items in TargetedDependenciesSnapshot


drewnoakes:
#4264 (57bec6fcb4143ae6c30fc56e5cd82ba2ca7a1ea9) should have fixed the regression I observed. I was very surprised to discover it did not.

In 77ca8fac8d423fbbfb1a43f5b1d71e0af3cb80f1 I'd incorrectly concluded that `ImmutableHashSet<>.Builder.Add` would replace an existing item. It is actually necessary to `Remove` and `Add`.

The change to usage of `ImmutableDictionary<TKey,TValue>.Builder` from `Remove`/`Add` to `this[TKey] = value` correctly replaces items in-place (the source shows it looks up the hash bucket and replaced the item directly). By contrast `Remove`/`Add` churns the data structure, changing the order of items.

Changing the item in-place causes most dependency nodes to appear with unresolved icons.

I _believe_ this means downstream code depends upon the order of items in the collection, and am looking to confirm this.

I'll first file a PR that reverts to `Remove`/`Add` so that we can prepare an insertion, and continue investigating this issue separately.

drewnoakes:
Ugh. Checking out 77ca8fac8d423fbbfb1a43f5b1d71e0af3cb80f1 and making the change at that commit fixes things, but the same change on `master` doesn't work. I'll have to do some more bisecting. There's clearly some fragility here.

Really looking forward to getting some integration tests on this.

drewnoakes:
I cannot reproduce this problem any more.

PR #4271, opened 2018-11-14
Fix dependencies node showing unresolved icons

drewnoakes:
Some fixes that restore the correct presentation of the dependencies node.

As mentioned in #4269, the change in e8d5becd53e75ac5382772124341912ae1eaac14 should not be necessary. As far as I can tell it only changes the order of items in the dictionary, which shouldn't have a bearing on anything of consequence. I'll follow up on that tomorrow.

Also added `Equals`/`GetHashCode` to the custom struct recently introduced in place of the `ValueTuple`. It's used in a set, and the default `ValueType` implementations are slow and likely to cause collisions (it's my understanding they'd only compute the hash code on the first field, `ProviderType` which is commonly duplicated). This

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IDependenciesChanges.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IDependenciesChanges.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Immutable;
 using System.Diagnostics;
 
@@ -15,13 +16,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
     }
 
     [DebuggerDisplay("({" + nameof(ProviderType) + ("}, {" + nameof(DependencyId) + "})"))]
-    public readonly struct RemovedDependencyIdentity
+    public readonly struct RemovedDependencyIdentity : IEquatable<RemovedDependencyIdentity>
     {
         public string ProviderType { get; }
         public string DependencyId { get; }
 
         public RemovedDependencyIdentity(string providerType, string dependencyId) : this()
         {
+            Requires.NotNull(providerType, nameof(providerType));
+            Requires.NotNull(dependencyId, nameof(dependencyId));
+
             ProviderType = providerType;
             DependencyId = dependencyId;
         }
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IDependenciesChanges.cs
```diff
@@ -31,5 +35,25 @@ public void Deconstruct(out string providerType, out string dependencyId)
...
+                   string.Equals(DependencyId, other.DependencyId, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public override bool Equals(object obj) => obj is RemovedDependencyIdentity other && Equals(other);
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (StringComparer.OrdinalIgnoreCase.GetHashCode(ProviderType) * 397) ^
```
jmarolf: Was this hand generated or did the codefix pick this prime?

drewnoakes: That was the largest prime I could compute on this piece of scrap paper with a hand-drawn sieve of Eratosthenes. <sup>R# generated it</sup>

Would rather use [`System.HashCode`](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode?view=netcore-2.1).

jmarolf: Huh, I guess it makes sense.  Prime is a little small for Fowler–Noll–Vo hash function.  For reference, this is what vanilla VS generates:

```C#
public override int GetHashCode()
{
    var hashCode = -427567219;
    hashCode = hashCode * -1521134295 + StringComparer.OrdinalIgnoreCase.GetHashCode(ProviderType);
    hashCode = hashCode * -1521134295 + StringComparer.OrdinalIgnoreCase.GetHashCode(DependencyId);
    return hashCode;
}
```

---
In reply to: [233276869](https://github.com/dotnet/project-system/pull/4271#discussion_r233276869) [](ancestors = 233276869)

davkean:
ValueTuple uses the hash codes of all items.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IDependenciesChanges.cs
davkean: I trick is make use of ValueTuple; (ProviderType, DependencyId).GetHashCode().

jmarolf: VS generates this if tuples are available in the project that you are generating into.

drewnoakes:
@davkean, agreed. ValueType does not. When we changed from ValueTuple, we have to add this boilerplate.

Maybe the hash code calculation could use `(a, b). GetHashCode()` in the absence of the new HashCode type.

davkean:
Ah, thought you said "ValueTuple".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/IDependenciesChanges.cs
```diff
@@ -31,5 +35,25 @@ public void Deconstruct(out string providerType, out string dependencyId)
...
+                        StringComparer.OrdinalIgnoreCase.GetHashCode(DependencyId);
```
Pilchie: Late, and doesn't apply as well when you need `OrdinalIgnoreCase`, but a handy trick to make the compiler generate this (and .Equals), is to use tuples under the hood:

```C#
    return (ProviderType, DependencyId).GetHashCode();
```

and 

```C#
    return (ProviderType, DependencyId) == (other.ProviderType, other.DependencyId);
```

drewnoakes: We can use this now we are on C# 7.3.

https://docs.microsoft.com/en-us/dotnet/csharp/tuples#equality-and-tuples

Issue #3425, opened 2018-03-28
Add an API that lets consumers wait on language service to be up-to-date with a given project version


davkean:
We need an API that lets consumers wait on the language service to be up-to-date with a given project version, similar to IProjectTreeService.PublishTree:

```c#
Task PublishLatestAsync(
  CancellationToken cancellationToken = default(CancellationToken));

Task PublishAsync(
  IImmutableDictionary<NamedIdentity, IProjectVersionRequirement> minimumRequiredDataSourceVersions,
  CancellationToken cancellationToken = default(CancellationToken));
```

This lets services such as CodeModel provider to block this before returning a CodeModel item, or the renamer to block on this attempting to rename. Also lets us block VS integration tests on this work before attempting to test state.

Without this, it's impossible to know when ProjectItem.CodeModel is going to return a result.

AbhitejJohn:
@davkean : Did we get a chance to schedule this work in yet...

davkean:
We're doing this as part of the language service hookup rewrite in 15.8. First part is in review, but we've only started on 15.8 this week. We're at least a sprint away from it.

japj:
Any update on this? It seems this is a blocking issue for 
https://developercommunity.visualstudio.com/content/problem/145930/create-unit-test-context-menu-missing-net-core-pro.html

Pilchie:
Note also that as part of this work we should use this API to push information to the VS Platform's new Operation Progress API so that we can track the async initialization.

kristianjaeger:
@davkean Any updates on this? Thank you.

tolgabalci:
Looking forward to getting the option for "Create Unit Tests" for our .Net Core apps.  This is a great time saver and helps Jr developers be more comfortable with creating unit tests.

kristianjaeger:
@Pilchie or @davkean Any updates on this?  From May 2018 to January 2019 is a long sprint ;) . Thank you.

davkean:
Appreciate your patience here, but due to the nature of software development it is hard to give a timeframe especially given the feature it was blocked on, only got to the point 5 minutes of ago of [being ready to merge](https://github.com/dotnet/project-system/pull/4419). We have tentatively put this in 16.1 timeframe but its priority is behind our focus for that release; .NET Core 3.0.

iWangJiaxiang:
[...]

PR #4273, opened 2018-11-14
 IWorkspaceProjectContextHost.Loaded -> PublishAsync

davkean:
This fixes a deadlock when the UI thread is blocked on Loaded, such as during PrioritizedProjectLoad. This occurred because MultiLifetimeComponent.Loaded was backed by a TaskCompletionSource that did not have a JoinableTask relationship to the code that completed it (MultiLifetimeComponent.LoadAsync).

I sat down with Andrew Arnott came up with a design that creates the relationship. I've used the name Publish (mimicking ITreeService, etc) as this method is basically the precursor to #3425.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
@@ -38,7 +38,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
         ///     This method does not initiate loading of the <see cref="AbstractMultiLifetimeComponent{T}"/>, however,
         ///     it will join the load when it starts.
         /// </remarks>
-        protected async Task<T> PublishInstanceAsync(CancellationToken cancellationToken = default)
+        protected async Task<T> WaitForLoadedAsync(CancellationToken cancellationToken = default)
         {
             // Wait until LoadAsync has been called, force switching to thread-pool in case
             // there's already someone waiting for us on the UI thread.
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
@@ -16,7 +16,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
         where T : class, IMultiLifetimeInstance
     {
         private readonly object _lock = new object();
-        private TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)> _instanceTaskSource = new TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)>();
+        private TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)> _instanceTaskSource = new TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)>(TaskCreationOptions.RunContinuationsAsynchronously);
 
         protected AbstractMultiLifetimeComponent(JoinableTaskContextNode joinableTaskContextNode)
              : base(joinableTaskContextNode)
@@ -89,7 +89,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
                 {
                     // Should throw TaskCancellatedException if already cancelled in Dispose
                     (instance, _) = _instanceTaskSource.Task.GetAwaiter().GetResult();
-                    _instanceTaskSource = new TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)>();
+                    _instanceTaskSource = new TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)>(TaskCreationOptions.RunContinuationsAsynchronously);
                 }
             }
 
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
```diff
@@ -11,33 +12,44 @@ namespace Microsoft.VisualStudio.ProjectSystem
...
+        ///     The result is awaited and <paramref name="cancellationToken"/> is cancelled.
+        /// </exception>
         /// <remarks>
-        ///     The returned <see cref="Task"/> is canceled when the <see cref="AbstractMultiLifetimeComponent"/> 
-        ///     is disposed.
+        ///     This method does not initiate loading of the <see cref="AbstractMultiLifetimeComponent{T}"/>, however,
+        ///     it will join the load when it starts.
         /// </remarks>
-        public Task Loaded
+        protected async Task<T> PublishInstanceAsync(CancellationToken cancellationToken = default)
```
drewnoakes: `PublishInstanceAsync` sounds like it initiates publish (a directive). Would `GetInstanceAsync` or `WhenPublished` be clearer?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
```diff
@@ -47,38 +59,37 @@ public async Task LoadAsync()
...
         }
 
-        private async Task LoadCoreAsync()
+        public async Task LoadCoreAsync()
         {
-            TaskCompletionSource<object> loadedSource = null;
-            IMultiLifetimeInstance instance;
+            JoinableTask initializeAsyncTask;
+
             lock (_lock)
```
jmarolf: Could we use something like AsyncReaderWriterLock? I agree that this logic looks correct (as much as I can understand it).   But do we really want to be blocking threadpool threads on initialization?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
```diff
@@ -47,38 +59,37 @@ public async Task LoadAsync()
...
             {
-                if (_instance != null)
+                if (_instanceTaskSource.Task.IsCompleted)
                 {
-                    instance = _instance;
-                    _instance = null;
-                    _loadedSource = new TaskCompletionSource<object>();
+                    // Should throw TaskCancellatedException if already cancelled in Dispose
+                    (instance, _) = _instanceTaskSource.Task.GetAwaiter().GetResult();
+                    _instanceTaskSource = new TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)>();
```
drewnoakes: Probably want to pass `TaskCreationOptions.RunContinuationsAsynchronously` here to prevent continuations running synchronously when the lock is held and `SetResult` or `TrySetCancelled` is called.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
```diff
@@ -11,33 +12,44 @@ namespace Microsoft.VisualStudio.ProjectSystem
...
     ///     a component that is loaded and unloaded multiple times.
     /// </summary>
-    internal abstract class AbstractMultiLifetimeComponent : OnceInitializedOnceDisposedAsync
+    internal abstract class AbstractMultiLifetimeComponent<T> : OnceInitializedOnceDisposedAsync
+        where T : class, IMultiLifetimeInstance
     {
         private readonly object _lock = new object();
-        private TaskCompletionSource<object> _loadedSource = new TaskCompletionSource<object>();
-        private IMultiLifetimeInstance _instance;
+        private TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)> _instanceTaskSource = new TaskCompletionSource<(T instance, JoinableTask initializeAsyncTask)>();
```
drewnoakes: Ditto (`TaskCreationOptions.RunContinuationsAsynchronously`).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
```diff
@@ -47,38 +59,37 @@ public async Task LoadAsync()
...
-                    _instance = CreateInstance();
-                    loadedSource = _loadedSource;
+                    (T instance, JoinableTask initializeAsyncTask) result = CreateInitializedInstanceAsync();
+                    _instanceTaskSource.SetResult(result);
                 }
 
-                instance = _instance;
+                Assumes.True(_instanceTaskSource.Task.IsCompleted);
+
+                // Should throw TaskCancellatedException if already cancelled in Dispose
```
drewnoakes: Nit: typo (TaskCancellatedException) appears twice

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
davkean: I changed this name, but left WorkspaceContextHost to be still be called `PublishAsync`. This follows the same pattern for other components that "publish", such as the tree, but really just waiting for the specified version to published.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
davkean: As mentioned on the PR that does this - this is lock has basically zero contention.

Issue #4058, opened 2018-09-27
Add support for Nullable build setting


jcouv:
This is a tracking work item related to the "nullable reference types" feature in C# 8.0 (for dev16).
We'll add design notes here as it crystallizes (is it a property or some other form of API?)
Tagging @davkean

jcouv:
The design for turning on the nullable feature changed since our last discussion.
Previously, we would have `[NonNullTypes]` (for class- or method-level control) or `[module:NonNullTypes]` (for module/project-level).
Now, we'll have `#nullable ...` for scoped control in source, and a compiler flag for project-level control. That compiler flag is exposed via [MSBuild property](https://github.com/dotnet/roslyn/pull/30574) `NullableReferenceTypes`, as part of dev16 preview1.

Based on this, there are two scenarios of interest:
1. exposing a checkbox in Build options ("Enable Nullable Reference Types")
2. allowing a codefixer to change this project property (like UpgradeProject does for LangVersion)

@davkean I assume that those scenarios translate into three work items:
1. some common work in project-system (object model or API change)
2. IDE work to add Build option UI (or is this also owned by project-system)
3. codefixer work (IDE/myself)

@davkean @Pilchie Do you think the API work (1) could be reasonably scheduled as part of preview 2? If yes, I would do the codefixer (3) in preview 2 as well.

etbyrd:
I could get a PR out for the checkbox in the Build property page - do we have an idea of where on the page we would want it?

jcouv:
For adding the checkbox to the Build property page, I think putting it under "Check for arithmetic overflow/underflow" would make sense.
@davkean Does that sounds ok?

tmeschter:
I wouldn't want to add the checkbox without adding the underlying property support to csproj/msvbprj, and that's the expensive part.

chucker:
In my testing, it looks like the `<NullableReferenceTypes>` property only works with new-style projects. Are there plans to make it work with the old project system (e.g., for an ASP.NET project)?

jcouv:
@chucker Yes, this property will be made to work in old project system too, before C# 8 ships as RTM, but likely after dev16.0 (where C# 8.0 language features are in beta).
The present issue tracks this remaining work.

jcouv:
@davkean We had to modify the options to add a third one (`true`/`false` were not enough) and changed the name of the option in the process.
I assume that the change of name affects project system, but the change of values doesn't. Let me know if that's not the case, as one of the options might still change.
[...]

PR #4277, opened 2018-11-15
Add NullableReferenceTypes property

davkean:
Fixes adding configuration-specific property for #4058

Note: This is a configuraiton-specific property but writes itself without a configuration condition, similar to LangVersion. We need this because it's going to live on the Build property page which pulls from configuration.

You can access this using DTE.Configuration.Properties["NullableReferenceTypes"]. We do not yet have a strongly typed API for it and will do that when legacy adds support for it.


davkean:
tag @jcouv

davkean:
@jcouv here's an example of checking for CPS: https://github.com/dotnet/roslyn/blob/027d0f2535403ae8140d7669035956c2b1713c88/src/VisualStudio/Core/Def/Implementation/ProjectSystem/VisualStudioWorkspaceImpl.cs#L298. I have no idea how to this in a code fix.

davkean:
Yes, though I think this will be one of the things that get "evaluation" handled.

etbyrd:
I see in https://github.com/dotnet/roslyn/issues/22152 it is mentioned that we will need a UI change on the Build prop page. Do we have an idea of what that will look like?

KenFoskeyInfo:
Just received this in VS 2019.

The "NullableReferenceTypes" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property.

Is this related to the above?

chucker:
@KenFoskeyInfo 

* is this in an SDK-style project?
* did you set `LangVersion` to `8.0`?

PR #4283, opened 2018-11-16
very small bit of nits

jmarolf:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
```diff
@@ -25,7 +25,7 @@ internal partial class LanguageServiceErrorListProvider : IVsErrorListProvider
         private IVsLanguageServiceBuildErrorReporter2 _languageServiceBuildErrorReporter;
 
         [ImportingConstructor]
-        public LanguageServiceErrorListProvider(UnconfiguredProject project, ILanguageServiceHost host)
+        public LanguageServiceErrorListProvider(ILanguageServiceHost host)
```
drewnoakes: Does removing `UnconfiguredProject` imports here and elsewhere change the scope of the export? Even though they're unused references, perhaps they have an impact on the lifecycle of the components.

https://github.com/Microsoft/VSProjectSystem/blob/master/doc/overview/scopes.md

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectDebuggerProvider.cs
```diff
@@ -89,14 +89,10 @@ public static Guid GetManagedDebugEngineForFramework(string targetFramework)
...
-            const string NetStandardAppPrefix = ".NetStandardApp";
             const string NetStandardPrefix = ".NetStandard";
             const string NetCorePrefix = ".NetCore";
-            const string NetCoreAppPrefix = ".NetCoreApp";
-            return targetFramework.StartsWith(NetCoreAppPrefix, StringComparison.OrdinalIgnoreCase) ||
-                   targetFramework.StartsWith(NetCorePrefix, StringComparison.OrdinalIgnoreCase) ||
-                   targetFramework.StartsWith(NetStandardPrefix, StringComparison.OrdinalIgnoreCase) ||
-                   targetFramework.StartsWith(NetStandardAppPrefix, StringComparison.OrdinalIgnoreCase);
+            return targetFramework.StartsWith(NetCorePrefix, StringComparison.OrdinalIgnoreCase) ||
+                   targetFramework.StartsWith(NetStandardPrefix, StringComparison.OrdinalIgnoreCase);
```
drewnoakes: Why is it safe to remove these checks?

jmarolf: In this case:
```C#
targetFramework.StartsWith(".NetStandard", StringComparison.OrdinalIgnoreCase)
targetFramework.StartsWith(".NetStandardApp", StringComparison.OrdinalIgnoreCase)
```
If something starts with `.NetStandardApp` it will also start with `.NetStandard`.  Case 1 is a super-set of case 2. The same is true of `.NetCore` and `.NetCoreApp` right?

drewnoakes: Ah yes. It's clear once pointed out.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
jmarolf: Good question.  You've spooked me enough that I am going to back these out as I have no idea how to rationalize implicit scopes here.  @davkean should we add comments for apparently unnecessary imports like this?

PR #4287, opened 2018-11-16
Restore public API of IDependenciesChanges

drewnoakes:
This gets the `IDependenciesChanges` interface back to where it was, restoring binary compatability with WTE and other consumers of the public API.

It's not a complete revert of the behaviour. Originally a fully populated `IDependencyModel` object would be created. With this change, a lightweight private implementation is created with just the necessary fields implemented.

I browsed through the NPM/Bower implementations WTE and am content that the integration should work. I'm running some tests locally now and if all's well, will remove WIP. Getting this up first for review.


drewnoakes:
After testing I'm content that WTE integrates correctly with our stuff again.

PR #4289, opened 2018-11-16
Test tidying

drewnoakes:
- Reinstate commented test
- Use assertions that provide more details when they fail

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyTests.cs
@@ -247,41 +247,38 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
             Assert.Equal(expectedDependencyId, newDependency.DependencyIDs[0]);
         }
 
-        //[Fact]
-        //public void Dependency_DependenciesProperty()
-        //{
-        //    var mockModel1 = IDependencyModelFactory.Implement(
-        //        providerType: "providerType",
-        //        id: "someId1",
-        //        dependencyIDs: new[] { "someId2", "someId_other" });
+        [Fact]
+        public void Dependency_DependencyIDsProperty()
+        {
+            var mockModel1 = IDependencyModelFactory.Implement(
+                providerType: "providerType",
+                id: "someId1",
+                dependencyIDs: new[] { "someId2", "someId_other" });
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyTests.cs
```diff
@@ -247,41 +247,38 @@ public void Dependency_SetProperties_PreservesDependencyIDs()
...
-        //    };
-
-        //    mockSnapshot.Setup(x => x.DependenciesWorld).Returns(
-        //        ImmutableDictionary<string, IDependency>.Empty.AddRange(children));
-
-        //    Assert.True(dependency1.Dependencies.Count() == 2);
-        //    Assert.True(dependency1.Dependencies.Any(x => x.Equals(dependency2)));
-        //    Assert.True(dependency1.Dependencies.Any(x => x.Equals(dependency3)));
-        //}
+        [Fact]
```
etbyrd: I wonder why this was commented out here: https://github.com/dotnet/project-system/commit/c9f4b6c943a04c0b9df1928824327a4951fff263#diff-da666934981933a9cc43c6fdaf1b6d4aL227

drewnoakes: I can't tell from context. Perhaps @abpiskunov can comment when he's back in the office.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyTests.cs
```diff
@@ -234,7 +234,7 @@ public void Dependency_SetProperties_PreservesDependencyIDs()
             var expectedDependencyId = "tfm1\\providerType\\glass";
 
             Assert.Equal(expectedId, dependency.Id);
-            Assert.True(dependency.DependencyIDs.Count == 1);
+            AssertEx.CollectionLength(dependency.DependencyIDs, 1);
```
davidwengier: Nit: You used `.Single` for this case elsewhere. I have no idea which gives a better message though ¯\\\_(ツ)_/¯

drewnoakes: Thanks. Fixed.

PR #4293, opened 2018-11-19
Fix compile error after parallel PR merges

drewnoakes:
One PR added a test using an API that another PR changed.


drewnoakes:
I'm merging without review to keep `master` building.

Issue #2714, opened 2017-08-15
Large amount of lock contention in dependency node


davkean:
In this bug: https://github.com/dotnet/project-system/issues/2713, I saw lots of lock contention, GetTargetFrameworkAsync has a huge amount - but that may be caused by what is the root cause of that bug of whatever cause us to continually call it. However, we should look at GetDependencyModel:

![image](https://user-images.githubusercontent.com/1103906/29308310-c3ebe72e-81e8-11e7-8fa6-47cef4d558b6.png)

Pilchie:
Same basic issue as #2707

PR #4294, opened 2018-11-20
No lock

jmarolf:
There are several situations where we lock inside of an async method when we do no need to.  Instead we can yield the thread until the lock in available


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -241,6 +245,18 @@ private static string GetDisplayName(ConfiguredProject configuredProject, Projec
                 projectData.DisplayName;
         }
 
+        private Task ExecuteWithinLockAsync(Action action) => _gate.ExecuteWithinLockAsync(JoinableCollection, JoinableFactory, action);
+
+        private int _isInitialized;
+
+        private async Task EnsureInitialized()
```
drewnoakes: Would this be a good addition to `OnceInitializedOnceDisposedAsync`? Maybe push it down and make it `protected`, then remove the dupe from `UnconfiguredProjectContextProvider`

Also, should have `Async` suffix.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -241,6 +245,18 @@ private static string GetDisplayName(ConfiguredProject configuredProject, Projec
...
+        {
+            if (Interlocked.CompareExchange(ref _isInitialized, 1, 0) == 0)
```
drewnoakes: Would `Interlocked.Exchange(ref _isInitialized, 1) == 0` be simpler?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -241,6 +245,18 @@ private static string GetDisplayName(ConfiguredProject configuredProject, Projec
...
+            {
+                await InitializeAsync();
```
drewnoakes: Could just return the `Task` here, and `Task.CompletedTask` in the usual case (zero allocation).

drewnoakes:
Potentially fixes #2714.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
drewnoakes: I ran a benchmark. `CompareExchange` is quite a bit faster.

| Method |        Mean |     Error |     StdDev | Median |
------------------------ |------------:|----------:|-----------:|-------:|
| Exchange | 158.0000 ns | 81.148 ns | 239.267 ns | 0.0 ns |
| CompareExchange_Match |  16.8293 ns | 16.653 ns |  44.162 ns | 0.0 ns |
| CompareExchange_NoMatch |   0.5814 ns |  1.393 ns |   3.790 ns | 0.0 ns |

```c#
[ClrJob]
public class ExchangeVsCompareExchangeBenchmarks
{
    private int _i;

    [IterationSetup]
    public void Setup() => _i = 0;

    [Benchmark]
    public void Exchange() => Interlocked.Exchange(ref _i, 1);

    [Benchmark]
    public void CompareExchange_Match() => Interlocked.CompareExchange(ref _i, 0, 1);

    [Benchmark]
    public void CompareExchange_NoMatch() => Interlocked.CompareExchange(ref _i, 1, 1);
}
```

```
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17763
Intel Xeon Silver 4110 CPU 2.10GHz, 2 CPU, 32 logical and 16 physical cores
  [Host] : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3190.0
  Clr    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3190.0
```

drewnoakes: I am quite surprised by the results there actually. Let me dig a little more...

drewnoakes: Yeah those results above are garbage.

| Method |     Mean |     Error |    StdDev |   Median |
| ------------------------ |---------:|----------:|----------:|---------:|
| Exchange | 7.222 ns | 0.1440 ns | 0.4246 ns | 6.985 ns |
| CompareExchange_Match | 9.680 ns | 0.1886 ns | 0.2880 ns | 9.555 ns |
| CompareExchange_NoMatch | 7.586 ns | 0.0116 ns | 0.0109 ns | 7.587 ns |

```c#
[ClrJob]
public class ExchangeVsCompareExchangeBenchmarks
{
    private const int OperationCount = 100_000_000;

    private int _i;

    [Benchmark(OperationsPerInvoke = OperationCount * 2)]
    public void Exchange()
    {
        _i = 0;

        for (var i = 0; i < OperationCount; i++)
        {
            Interlocked.Exchange(ref _i, 0);
            Interlocked.Exchange(ref _i, 1);
        }
    }

    [Benchmark(OperationsPerInvoke = OperationCount * 2)]
    public void CompareExchange_Match()
    {
        _i = 0;

        for (var i = 0; i < OperationCount; i++)
        {
            Interlocked.CompareExchange(ref _i, 0, 1);
            Interlocked.CompareExchange(ref _i, 1, 0);
        }
    }

    [Benchmark(OperationsPerInvoke = OperationCount * 2)]
    public void CompareExchange_NoMatch()
    {
        _i = 0;

        for (var i = 0; i < OperationCount; i++)
        {
            Interlocked.CompareExchange(ref _i, 1, 1);
            Interlocked.CompareExchange(ref _i, 1, 1);
        }
    }
}
```

There's not much in it. I'd go for `Exchange` based on it being the simpler form, but it's not a biggie. My curiousity's been satisfied.

drewnoakes:
Looking at the source of `OnceInitializedOnceDisposedAsync.InitializeAsync`, I'm thinking it already handles the 'ensure' logic. Maybe the extra layer is redundant. The docs suggest as much:

> Initializes this instance if it has not yet been.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/EnsureOnceInitializedOnceDisposedAsync.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    internal abstract class EnsureOnceInitializedOnceDisposedAsync : OnceInitializedOnceDisposedAsync
```
davkean: This isn't needed, OnceInitializedOnceDisposedAsync already gives this guarantee.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -64,7 +67,7 @@ public void ReleaseProjectContext(AggregateCrossTargetProjectContext context)
         {
             Requires.NotNull(context, nameof(context));
 
-            lock (_gate)
+            using (_gate.DisposableWait())
```
drewnoakes: Looks like you can use the async form here and flow the task backwards.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -146,15 +150,15 @@ private ProjectData GetProjectData()
...
             {
                 return _configuredProjectContextsMap.TryGetValue(configuredProject, out targetedProjectContext);
             }
         }
 
         private void AddConfiguredProjectState(ConfiguredProject configuredProject, ITargetedProjectContext projectContext)
         {
-            lock (_gate)
+            using (_gate.DisposableWait())
```
drewnoakes: Ditto (async form).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -146,15 +150,15 @@ private ProjectData GetProjectData()
 
         private bool TryGetConfiguredProjectState(ConfiguredProject configuredProject, out ITargetedProjectContext targetedProjectContext)
         {
-            lock (_gate)
+            using (_gate.DisposableWait())
```
drewnoakes: Ditto (async) via `Task<(bool found, ITargetedProjectContext targetedProjectContext)>` or similar.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetSubscriptionHostBase.cs
```diff
@@ -284,7 +267,7 @@ protected override async Task DisposeCoreAsync(bool initialized)
 
         private void DisposeAndClearSubscriptions()
         {
-            lock (_linksLock)
+            using (_linksLock.DisposableWait())
```
drewnoakes: Ditto (async).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
```diff
@@ -5,10 +5,12 @@
...
 using System.ComponentModel.Composition;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 
 using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
+using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
```
drewnoakes: Not sure it's worth changing stuff in the language service. @davkean may say otherwise.

jmarolf:
@drewnoakes in a separate PR we can look at making all methods async and not blocking any of the calling threads.  If I were to do this it would change about ~1000 methods, some of which are part of the public API.  This PR focuses on the less controversial fix of not blocking a thread when a method is already async. 

**NOTE:** you cannot use `DisposableWaitAsync` safely.  You need to instead use `ExecuteWithinLockAsync` in order to prevent deadlocks.

https://github.com/dotnet/project-system/blob/b72d88016d51f59cebca83b84428d494c66aaa11/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/SemaphoreSlimExtensions.cs#L41-L57

drewnoakes:
No worries if you want to do it in a separate PR&mdash;this is already a big improvement in reduction of blocking. I'd be keen to see you merge this sooner anyway, as I have some overlapping changes inbound.

> it would change about ~1000 methods, some of which are part of the public API

In my quick look at this I felt the ones I highlighted would require asyncifying fewer than 10 (only one or two levels back), none of which are public.

And yes, the goal was to replace the blocking `DisposableWait` with `ExecuteWithinLockAsync`, even if the func is synchronous. No one ends up in a blocking wait.

jmarolf:
@davkean 	any problems here I should know about?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
jmarolf: I've added async versions for these. As you said, it didn't end up spreading the virus very far :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetSubscriptionHostBase.cs
```diff
@@ -216,9 +201,9 @@ private Task<AggregateCrossTargetProjectContext> UpdateProjectContextAsync()
             });
         }
 
-        private void DisposeAggregateProjectContext(AggregateCrossTargetProjectContext projectContext)
+        private async Task DisposeAggregateProjectContextAsync(AggregateCrossTargetProjectContext projectContext)
         {
-            _contextProvider.Value.ReleaseProjectContext(projectContext);
+            await _contextProvider.Value.ReleaseProjectContextAsync(projectContext);
```
davkean: Unneeded state machine, just return.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/CrossTargetSubscriptionHostBase.cs
```diff
@@ -227,9 +212,9 @@ private async Task AddSubscriptionsAsync(AggregateCrossTargetProjectContext newP
 
             await _commonServices.ThreadingService.SwitchToUIThread();
 
-            await _tasksService.LoadedProjectAsync(() =>
+            await _tasksService.LoadedProjectAsync(async () =>
             {
-                lock (_linksLock)
+                await _linksLock.ExecuteWithinLockAsync(JoinableCollection, JoinableFactory, () =>
```
davkean: Unneeded state machine, just return.

davkean:
I'm unconvinced by some of these, they basically have zero/very little contentention, don't continue on another thread and we've introduced JoinableTask overhead. We've probably made some of these situations worse as some of them just update a field, or add something to a collection - it's probably going to take longer for the async plumbing to schedule a continuation than its going to unblock the thread. I unconvinced that thread pool usage won't go up running all the continuations.

Convince me why it a good thing to replace all locks with async machinery?

jmarolf:
## TL;DR

Collected some traces for this and it looks like this change is not worth it.  Thanks @davkean for challenging me on this. Below is the record of what I did, hopefully this is helpful to someone in the future.

## Process
### Setup
1. Enlist in `dotnet/project-system`
2. Enlist in `dotnet/roslyn`
3. install latest lab/d16.0stg build of VS

### Collection
For each branch (master and no-lock)
1. checkout project-system branch
2. build release version (`build.cmd /release /rootSuffix Exp`)
3. sync roslyn repo to [this commit](https://github.com/dotnet/roslyn/commit/547fbc0e803540c3b494e927ae680894cd30f2ef)
4. run `git clean -xdf` in the roslyn repo
5. run `Restore.cmd` in the roslyn repo
6. Start perf trace 
`>PerfView64.exe collect C:\perfTrace\Trace.etl -CircularMB:4096 -BufferSizeMB:256 -Merge:true -Zip:true -Providers:641d7f6c-481c-42e8-ab7e-d18dc5e5cb9e,*Microsoft-VisualStudio-Common,.NETTasks:0:0 -ThreadTime -NoV2Rundown -NoNGenRundown`
7. start VS `devenv /rootSuffix Exp C:\source\roslyn\Roslyn.sln`
8. Wait for all the nodes in the project system tree to not have yellow triangles
9. Stop perf collection

### Analysis
1. Find the *start* timestamp by looking at the process start event time
2. Find the *stop* event time by finding the last tree update event that was fired

**NOTE:** Traces can be found here: \\mlangfs1\public\jmarolf\traces

## Results
|                            |    [master](https://github.com/dotnet/project-system/commit/b713612310d72c45476c40b9b0ff9b5603c47e55)     |  [jmarolf:no-lock](https://github.com/jmarolf/project-system/commit/812dd253c5a7d9f7b5a1a28ccb42ff4a7b467cb9) |  Delta   |
|----------------------------|:-------------:|:----------------:|:--------:|
| Start (ms)                 | 1,751.522     |  723.096         |          |
| Stop (ms)                  | 41,743.366    |  40,316.409      |          |
| Elapsed Time (ms)          | 39,991.844    |  39,593.313      | 1.00%    |
| CPU Stack Time Sum (ms)    | 188,536.20    |  222,953.30      | -18.25%  |
| Thread Time Stacks Sum (ms)| 188,536.20    |  222,953.30      | -57.96%  |
| # Threads                  | 118           |  117             | 0.85%    |
| BLOCKED_TIME Sum (ms)      | 2,823,503.50  |  4,514,176.50    | -59.88%  |
| BLOCKED_TIME Count         | 134,592       |  382,517         | -184.20% |
| CPU_TIME Sum (ms)          | 173,050.13    |  222,757.33      | -28.72%  |
| CPU_TIME Count             | 330,200       |  605,195         | -83.28%  |
| READIED_TIME Sum (ms)      | 4,194.45      |  5,129.56        | -22.29%  |
| READIED_TIME Count         | 2,493         |  3,255           | -30.57%  |

PR #4296, opened 2018-11-20
Dependency snapshot filtering API

drewnoakes:
`IDependenciesSnapshotFilter`'s add/remove methods previously had implicit usage contracts, and it was easy for implementations to accidentally violate them.

This change introduces new classes (`AddDependencyContext` and `RemoveDependencyContext`) that wrap the mutable state exposed to filter implementations, and codifies the kinds of interactions filters can perform on that state. They then correctly track whether changes have occurred so we know whether to produce a new snapshot.

Dependency filter unit tests have been modified to be state-based rather than mock-based. The `TestDependency` class has been extended to assist with this.

Unfortunately the nature of the changes to the test code means the diff is pretty horrible. The best place to start reading from is `IDependenciesSnapshotFilter`, as this shows the fundamental API change that all other changes then stem from.

No public APIs were harmed in the making of this PR.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -140,6 +93,79 @@ public static ITargetedDependenciesSnapshot CreateEmpty(string projectPath, ITar
...
+                        projectPath,
+                        targetFramework,
+                        dependency,
+                        context);
+
+                    anyChanges |= context.Changed;
+
+                    if (!context.GetResult(filter))
+                    {
+                        // TODO breaking here denies later filters the opportunity to modify builders
```
jmarolf: change to `return`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/AddDependencyContext.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                throw new InvalidOperationException(
+                    $"Filter type {filter.GetType()} must either accept or reject the dependency.");
+            }
+
+            _acceptedOrRejected = false;
+            return _acceptedDependency;
+        }
+
+        #region IAddDependencyContext
```
davkean: We're a region free zone.

https://github.com/dotnet/project-system/blob/master/docs/repo/coding-conventions.md#code

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IAddDependencyContext.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Immutable;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.Filters
+{
+    /// <summary>
+    /// Context used by <see cref="IDependenciesSnapshotFilter"/> implementations when filtering
+    /// a dependency that is being added to a <see cref="DependenciesSnapshot"/> by
+    /// <see cref="DependenciesSnapshot.FromChanges"/>.
+    /// </summary>
+    internal interface IAddDependencyContext
```
davkean: Is this following a particular pattern? I've not seen it before.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IAddDependencyContext.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <remarks>
+        /// In the course of filtering one dependency, the filter may wish to modify or add other
+        /// dependencies in the project's tree. This method allows that to happen.
+        /// </remarks>
+        void AddOrUpdate(IDependency dependency);
+
+        /// <summary>
+        /// Attempts to find the dependency in the project's tree with specified <paramref name="dependencyId"/>.
+        /// </summary>
+        bool TryGetDependency(string dependencyId, out IDependency dependency);
```
davkean: Is there a reason you decided to expose these three methods, instead of just exposing the "world" as a property? You're already tied to an implementation via ImmutableDictionary? It is to avoid accidental boxing the of the enumerator?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IRemoveDependencyContext.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.Filters
+{
+    /// <summary>
+    /// Context used by <see cref="IDependenciesSnapshotFilter"/> implementations when filtering
+    /// a dependency that is being removed from a <see cref="DependenciesSnapshot"/> by
+    /// <see cref="DependenciesSnapshot.FromChanges"/>.
+    /// </summary>
+    internal interface IRemoveDependencyContext
+    {
+        void Accept();
```
davkean: Does this mean "Accept" removal?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IAddDependencyContext.cs
davkean: Ah, to avoid exposing the "builder"?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/IAddDependencyContext.cs
drewnoakes: See https://github.com/dotnet/project-system/pull/4296#discussion_r235331280

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/AddDependencyContext.cs
```diff
@@ -0,0 +1,93 @@
...
+        private IDependency _acceptedDependency;
+
+        public bool Changed { get; private set; }
+
+        public AddDependencyContext(ImmutableDictionary<string, IDependency>.Builder worldBuilder)
+        {
+            _worldBuilder = worldBuilder;
+        }
+
+        public void Initialize()
```
davkean: This is called Reset elsewhere. Initialize tends to single use.

davkean:
👍

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
drewnoakes: Isn't it already returning?

drewnoakes:
Thanks for the reviews. Updates pushed.

PR #4302, opened 2018-11-21
Revert https://github.com/dotnet/project-system/pull/4003

davidwengier:
Seems we do need to define these, but getting .vsct changes to appear in the experimental instance is very temperamental so my testing of the above PR was testing old bits.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Menus.vsct
@@ -67,10 +67,16 @@
         <Strings>
           <ButtonText>&amp;Add Below</ButtonText>
         </Strings>
-      </Menu>  
+      </Menu>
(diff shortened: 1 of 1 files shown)


davidwengier:
I've tried to "fix" CPS by implementing Menus.vsct there to no effect. The package is loaded and the menu resource is created but it doesn't get added to the private registry. I've asked around about tooling to help diagnose issues, but am timeboxing this work for now.

The alternative is the remove the Menus.vsct from CPS and reinstate this PR. That has the advantage of not affecting any other consumers of CPS who currently don't have an Edit .csproj menu item and don't want one.

davkean:
> That has the advantage of not affecting any other consumers of CPS who currently don't have an Edit .csproj menu item and don't want one.

There isn't an advantage. Both should result in exactly the same thing, they affect the product as a whole. The commands are hidden/disabled when you don't have OpenProjectFile capability.

PR #4305, opened 2018-11-21
Fix endless loop if optimistic set update fails

drewnoakes:
This would occur after an optimistic update failed. The loop needs to update the value of `prior` before continuing, otherwise it's an endless loop.

I observed this as a hang when unloading projects in d16.0stg. Let's get this in for the next insertion.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSetCache.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSetCache.cs
@@ -52,6 +52,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                 }
 
                 // Optimistic update failed, so loop around and try again.
+                prior = original;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyIconSetCache.cs
```diff
@@ -52,6 +52,7 @@ private static T GetOrAdd<T>(ref ImmutableHashSet<T> location, T value)
                 }
 
                 // Optimistic update failed, so loop around and try again.
+                prior = original;
```
davkean: Why can't this entire method be replaced with ApplyChangeOptimistically?

drewnoakes: The improvement here is lookup within the loop.

drewnoakes: `ImmutableInterlocked` has `GetOrAdd` methods for `ImmutableDictionary` but not `ImmutableHashSet`. At some point I'll file a PR with this addition for corefx.

davkean: Hmm, but I like tried and tested methods also, avoids bugs. 

Did this actually show up on the radar? If so we should be pushing these changes into VS threading.

drewnoakes: I tried pushing a similar change into vs-threading. The response suggested this kind of thing belongs in `ImmutableInterlocked`, and I tend to agree.

drewnoakes:
You can verify it by attaching a debugger. You should see a stack with the icon cache at the top, adding to an immutable collection, and probably a bunch of other stacks blocked on the locks it's holding.

I'm looking at ordering/race conditions and at the same time wondering if we can get rid of that lock. Possibly not.

PR #4306, opened 2018-11-21
Memory reduction

drewnoakes:
Various reductions to memory, mainly avoiding temporary strings and arrays. I searched the project for `Trim` and `Split`.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
@@ -56,8 +56,7 @@ namespace Microsoft.VisualStudio.Build
             Requires.NotNull(project, "project");
 
             return project.Properties
-                .Where(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase))
-                .FirstOrDefault();
+                .FirstOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase));
         }
 
         /// <summary>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
@@ -71,11 +71,9 @@ namespace Microsoft.VisualStudio.Build
             // We need to ensure that we return values in the specified order.
             foreach (string value in new LazyStringSplit(propertyValue, delimiter))
             {
-                string s = value.Trim();
-
-                if (!string.IsNullOrEmpty(s))
+                if (!string.IsNullOrWhiteSpace(value))
                 {
-                    yield return s;
+                    yield return value.Trim();
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
```diff
@@ -37,6 +37,25 @@ public LazyStringSplit(string input, char delimiter)
 
         IEnumerator<string> IEnumerable<string>.GetEnumerator() => GetEnumerator();
 
+        public IEnumerable<T> Select<T>(Func<string, T> func)
```
abpiskunov: Nice util, I wish this type of helpful utils were shared somewhere across VS for others to reuse ..

src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
```diff
@@ -66,21 +66,16 @@ public static ProjectPropertyElement GetProperty(ProjectRootElement project, str
...
             {
-                if (!string.IsNullOrEmpty(value))
+                if (!string.IsNullOrWhiteSpace(value))
                 {
-                    valuesBuilder.Add(value);
+                    yield return value.Trim();
                 }
             }
-
-            return valuesBuilder.Distinct(StringComparer.OrdinalIgnoreCase).ToImmutableArray();
```
drewnoakes: This change might not be correct actually. The previous code eliminated duplicates. Will review.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -71,7 +72,7 @@ protected virtual async Task<ImmutableArray<string>> GetOrderedPropertyValuesAsy
             }
             else
             {
-                return BuildUtilities.GetPropertyValues(propertyValue);
+                return BuildUtilities.GetPropertyValues(propertyValue).ToImmutableArray();
```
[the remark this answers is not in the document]
drewnoakes: The change was sympathetic to other callers that could consume values lazily. Though that may not make sense if it turns out I reinstate de-duplicating the output of GetPropertyValues, in which case I may restore the immutable return type.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
drewnoakes: Where would it live?

drewnoakes: Where would it live?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
drewnoakes: I restored deduplication and added a test with duplicates.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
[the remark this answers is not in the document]
drewnoakes: A ImmutableArray builder versus the `ToImmutableArray` method are on par for perf. However for lazy consumers (there are a few), we avoid allocating more than just the enumerator, and if they don't consume all elements then we don't allocate substrings that would be unused.

PR #4318, opened 2018-11-26
[WIP] Implement TempPEBuildManager using UnconfiguredProjectHostBridge

davidwengier:
There is still lots to be done in this, and most of it not documented in comments, but just wanted to get a quick look over the work so far to make sure its on the right track. The main thing is the TempPEManager and its implementation of UnconfiguredProjectHostBridge.

The idea goes like this:
* TempPEBuildManager receives updates to source items view PreprocessAsync
* That method pulls out the changes relevant to us (TODO: ensure ShouldApplyChanges returns false when there are no design time changes)
* Those changes are then pushed to the UI thread in ApplyAsync, including creating a CancellationSeries for each file
* In GetTempPEBlob we utilise the CancellationSeries (thanks Drew) in order to cancel any in-progress compiles for the input file, so the new one is the only one.
  * The theory here is that we can avoid the need for the event queue that legacy has since our builds are cancellable

Big TODO items are:
* Events
* Taking a dependency on new Roslyn bits (commented out code is correct though)
* Checking if we actually need to compile based on file datetime
  * This is arguably unnecessary with project changes being pushed to us, but loading a project could benefit from it
* Listening to events (if applicable?)

Have at it. I expect lots of comments :P


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            var applied = AppliedValue.Value;
+            var designTimeInputs = applied.Inputs.ToBuilder();
+            var designTimeSharedInputs = applied.SharedInputs.ToBuilder();
+
+            foreach (var item in value.AddedItems)
+            {
+                // TODO: Raise events
+
+                designTimeInputs.Add(item, new CancellationSeries());
```
davidwengier: Another TODO here would be to move this `new CancellationSeries` to `GetTempPEBlob` so that its lazy, and only created if we need to ever compile a TempPE DLL, but I want to get the tests written for that first so I can make sure that calling ApplyAsync from that method works as I expect it to.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal interface ITempPEBuildManager
```
jmarolf: Could you add comments about this interface such as why `bstrOutputMoniker` is the best name for a parameter

davidwengier: Yeah, I should probably rename that :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,315 @@
...
+            }
+
+            return base.DisposeCoreAsync(initialized);
+        }
+
+        /// <summary>
+        /// Use the project subscription service to read connected services data from the tree service.
+        /// </summary>
+        [Import]
+        private IActiveConfiguredProjectSubscriptionService ProjectSubscriptionService { get; set; }
```
jmarolf: Can we just import this [in the constructor?](https://github.com/dotnet/project-system/blob/master/docs/repo/coding-conventions.md#mef)

davidwengier: Good point, thank you. This is a copy-paste from CPS who use property imports more.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        [ImportMany(typeof(BuildManager))]
+        internal OrderPrecedenceImportCollection<BuildManager> BuildManager { get; set; }
+
+        protected override Task DisposeCoreAsync(bool initialized)
+        {
+            if (AppliedValue != null)
+            {
+                // dispose any of our cancellation series we still have
+                foreach (var item in AppliedValue.Value.Inputs.Values)
```
jmarolf: While I love `var` personally.  Its use in this repository is restricted to only cases where they type is apprent (such as `var f = new Foo();`)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                new DataflowLinkOptions { PropagateCompletion = true },
+                cancellationToken: ProjectAsynchronousTasksService.UnloadCancellationToken);
+        }
+
+        protected override Task<DesignTimeInputsDelta> PreprocessAsync(IProjectVersionedValue<InputTuple> input, DesignTimeInputsDelta previousOutput)
+        {
+            var project = input.Value.Item1.ProjectInstance;
+            var changes = input.Value.Item2.ProjectChanges[Compile.SchemaName];
+            
+            var addedDesignTimeInputs = ImmutableList.CreateBuilder<string>();
```
jmarolf: Why are we using `ImmutableList` here and not `ImmutableArray`?

jmarolf:
@davidwengier 	will this every be called from background threads / multiple threads at once?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProjectEvents.cs
```diff
@@ -6,29 +6,32 @@
...
         [ImportMany]
         protected OrderPrecedenceImportCollection<ImportsEvents> ImportsEventsImpl { get; set; }
 
         public ReferencesEvents ReferencesEvents => _vsProject.Events.ReferencesEvents;
 
-        public BuildManagerEvents BuildManagerEvents => _vsProject.Events.BuildManagerEvents;
+        public BuildManagerEvents BuildManagerEvents => _buildManager as BuildManagerEvents ?? _vsProject.Events.BuildManagerEvents;
 
-        public ImportsEvents ImportsEvents => ImportsEventsImpl.First().Value;
+        public ImportsEvents ImportsEvents => ImportsEventsImpl.FirstOrDefault()?.Value ?? _vsProject.Events.ImportsEvents;
```
davkean: What does C# legacy return for this property? ImportsEvent is for VB only.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,132 @@
...
+        private readonly ITempPEBuildManager _tempPEManager;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="VSBuildManager"/> class.
+        /// </summary>
+        [ImportingConstructor]
+        internal VSBuildManager(IUnconfiguredProjectCommonServices unconfiguredProjectServices,
+            ITempPEBuildManager tempPEManager)
+        {
+            AddEventSource(this as IEventSource<_dispBuildManagerEvents>);
```
davkean: Why "as"? We expect this to fail?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,132 @@
...
+            _unconfiguredProjectServices = unconfiguredProjectServices;
+            _tempPEManager = tempPEManager;
+
+            Project = new OrderPrecedenceImportCollection<VSLangProj.VSProject>(projectCapabilityCheckProvider: _unconfiguredProjectServices.Project);
+        }
+
+        [ImportMany(ExportContractNames.VsTypes.VSProject)]
+        internal OrderPrecedenceImportCollection<VSLangProj.VSProject> Project { get; set; }
```
davkean: You can remove the set.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        #region _dispBuildManagerEvents_Event Members
```
davkean: #noregions

jmarolf:
> Why do you ask?

Just trying to reason about the logic.  My understanding is that
1. We only want one compilation to be in-progress at any given point
2. We want calls to `GetTempPEBlobAsync` to cancel active compilations and queue a new one

What you have seems correct. Are callers to `GetTempPEBlobAsync` expected to handle operation cancelled exceptions and re-queue?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,315 @@
...
+            }
+
+            AppliedValue = new ProjectVersionedValue<DesignTimeInputsItem>(new DesignTimeInputsItem
+            {
+                Inputs = designTimeInputs.ToImmutable(),
+                SharedInputs = designTimeSharedInputs.ToImmutable(),
+            }, value.DataSourceVersions);
+
+            // Fire off the events
+            var buildManager = BuildManager.First().Value as VSBuildManager;
```
davkean: We have a hard dependency on the BuildManager being; BuildManagerEvents and VSBuildManager. Please just import that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,315 @@
...
+            {
+                // otherwise just the ones that we've added are dirty
+                foreach (var item in addedDesignTimeInputs)
+                {
+                    buildManager.OnDesignTimeOutputDirty(item);
+                }
+            }
+            foreach (var item in removedDesignTimeInputs)
+            {
+                buildManager.OnDesignTimeOutputDeleted(item);
```
davkean: Are we on the UI thread here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
jmarolf: `ImmutableArray` is superior to `IReadOnlyList` on a number of fronts so if I had choice I would always choose it for returning snapshots of data in an API.

`ImmutableList` etc. assumes that you will mutate this in some way, however `ImmutableList.CreateBuilder<string>()` is already stating "I am going to mutate this then create an immutable version" so its unclear to me why `ImmutableList` is desired here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        #endregion
+
+        /// <summary>
+        /// Occurs when a design time output moniker is deleted.
+        /// </summary>
+        internal virtual void OnDesignTimeOutputDeleted(string outputMoniker)
+        {
+            DesignTimeOutputDeleted?.Invoke(outputMoniker);
```
davkean: Is this what CPS does? In legacy, any consumers of BuildManager will be RPC marshalled to the UI thread, and the these events will only ever being fired on UI thread. Please assert/check that we're on UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: ¯\\\_(ツ)_/¯

Most of this file is copied straight from the current CPS implementation of BuildManager

davkean:
Cancellation is a little unclear to me, how does this work? Looks like we're compiling on the UI thread in response to someone asking on the UI thread blocking us from presumably updating the AppliedValue, how exactly does cancellation work?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        /// <summary>
+        /// Use the project subscription service to read connected services data from the tree service.
+        /// </summary>
+        [Import]
+        private IActiveConfiguredProjectSubscriptionService ProjectSubscriptionService { get; set; }
+
+        [ProjectAutoLoad]
+        [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
```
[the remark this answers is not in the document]
davidwengier: That doesn't exist. Do you mean `DotNetLanguageService`?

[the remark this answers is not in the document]
davkean: Depends, what happens when you call into Roslyn and the language is F#?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: Yes, this is straight from CPS too. The ReferencesHostBridge uses `HResult.Invoke` to fire its events.

Once I change ApplyAsync to be on the UI thread this will be too though.

davkean: Please assert that.

davidwengier:
The cancellation stuff is still just a theory, but you raise a good point. The cancellation would need to happen in PreprocessAsync or ApplyAsync but off the UI thread. The theory is simply to avoid needing the event loop that legacy has, we can simply cancel any current compilation and let the next one happen as change flow through to the UI thread. Now that I'm writing that out though, perhaps thats not necessary at all, and the fact that compilation is happening on the UI thread will be enough - we can just let the current compilations finish and then the next set of changes will push through anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: I thought ProjectHostBridge guaranteed that (isn't that the whole point? Only changing AppliedValue on the  UI thread) but I see now that ReferencesHostBridge has a VerifyOnUIThread at the start so I will add that here.

davkean: Anywhere we think we're on the UI thread; assert.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        #region BuildManager Members
+
+        /// <summary>
+        /// Builds a temporary portable executable (PE) and returns its description in an XML string.
+        /// </summary>
+        public string BuildDesignTimeOutput(string bstrOutputMoniker)
+        {
+            if (bstrOutputMoniker == null) return null;
```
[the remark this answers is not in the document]
davidwengier: Legacy returns E_INVALIDARG but VSTypeResolutionService doesn't do anything in the catch block except trace so I figured I'd save the exception overhead. Should I change it to throw an ArgumentException?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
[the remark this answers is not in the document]
davidwengier: I don't know. Do we want TempPE for F#? Whats the difference between `CSharpOrVisualBasic` and `CSharpOrVisualBasicLanguageService`?

davidwengier:
> Are callers to GetTempPEBlobAsync expected to handle operation cancelled exceptions and re-queue?

No, the theory is that the thing doing the cancellation is doing so because its in the process of re-queuing.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: > While I love `var` personally.  Its use in this repository is restricted to only cases where they type is apprent

I don't know how I got it in my head but I thought we were var everywhere. Which is weird because personally I'm not a fan of var :P

Will definitely be changing this one!

davidwengier: It is. Sometimes I open solutions that are not so pure as ours, and I had Message off as a result

davkean:
The compilation is in response to a call from the UI thread. The UI thread is blocked on the call, nothing else will execute on the UI thread (unless it unblocks the call) until the call is finished. There's probably no benefit to it running on a background thread.

However, perhaps we could reconsider the design where there is a benefit? I can't remember what legacy does here, do you remember?

davidwengier:
I don't know why it is using a background thread, there aren't any comments indicating any reasons [in the source](http://index/?query=%60_CSharpTempPEMgrThreadProc&rightProject=VisualStudio(Repo)&file=vsproject%5Ccool%5Ccoolbld%5CCCSharpTempPECompiler.cpp&line=326). I thought we wanted to compile on the UI thread because that way we're guaranteed that the project state is consistent because we're not passing all of the info to Roslyn any more, but relying on the fact that `AppliedValue` and roslyns `CurrentSolution` are both only changed on the UI thread.

davkean:
UI thread is blocked, so CurrentSolution/AppliedValue cannot change regardless of where you compile. Let me try and understand why/what it is doing.

davidwengier:
I don't claim to be able to read the C++ code well, but in theory the UI thread doesn't have to be blocked on the compile, just the returning of the XML string. That can be done using only file names.

Why you would want to tell someone a DLL is somewhere, then schedule a task to make it appear there at some point in time in the future, I certainly don't know.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Can you expand on that? Do you mean directly importing VSBuildManager (need to add another Export attribute to it?) or do you mean importing VSProjectEvents and access through BuildManagerEvents so that we're ensuring our dependency is hooked up correctly?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProjectEvents.cs
davkean: Ah, I thought this started returning null, so it's not a change in behavior, and hence not changed in this PR. Legacy is relevant because we need to be compatible with it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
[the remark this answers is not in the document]
davkean: Yes, please do.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsProjectEvents.cs
davidwengier: I can actually revert all of the changes to this file as it turns out. CPS's OAVSProjectEvents gets its BuildManagerEvents through a VSProject import, which ends up being our implementation anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
drewnoakes: > I thought Array was for when adding is rare.

It is.

Here's my mental model for `ImmutableArray<>` in case it's helpful:

- `ImmutableArray<>` is expensive to mutate (linear space/time complexity as internal array is copied)
- `ImmutableArray<>.Builder` is no more expensive to mutate than `List<>` (amortised constant space/time complexity)
- Calling `builder.ToImmutable` involves a defensive copy of the builder's internal array
- Calling `builder.MoveToImmutable` moves the builder's inner array to the returned `ImmutableArray<>` without copying, but only works if the builder is created with the correct initial size (otherwise it throws)
- `ImmutableArray<>` cannot be `null`, however `default(ImmutableArray<>)` throws when used, so you must check `IsDefault` or `IsDefaultOrEmpty` to avoid exceptions
- `ImmutableArray<>` implements `IReadOnlyList<>` but the conversion boxes
- `ImmutableArray<>`  allows allocation-free enumeration

> `ImmutableList.CreateBuilder<string>()` is already stating "I am going to mutate this then create an immutable version"

Agreed. It also says to me that the consumer of that immutable version is likely to make further changes over time. If that's not the case, `ImmutableArray<T>` is probably a better choice.

davidwengier: Thank you that's awesome.

> ImmutableArray<>.Builder is no more expensive to mutate than List<> (amortised constant space/time complexity)
> Calling builder.ToImmutable involves a defensive copy of the builder's internal array

In that case it would seem like in this case (always starting with an empty builder, collection of unknown size) there would be no downside to just using a `List<>` and using `CreateRange` to convert it to an ImmutableArray at the end right?

drewnoakes: `CreateRange` takes `IEnumerable<>` but sniffs the fact that it can grab the length and use `CopyTo`. So in theory they should be roughly the same.

I ran a quick benchmark while waiting for a dump to copy. They allocate roughly the same amount of memory, but vary on speed.

| Method | Count | Mean | Allocated |
|------------------------------------- |------ |----------:|----------:|
| ImmutableArray_CreateRange_FromList |     3 | 182.44 ns |      76 B |
| ImmutableArray_Builder_UnknownLength |     3 | 125.56 ns |      84 B |
| ImmutableArray_Builder_KnownLength |     3 |  65.98 ns |      40 B |
| ImmutableArray_CreateRange_FromList |    30 | 714.48 ns |     444 B |
| ImmutableArray_Builder_UnknownLength |    30 | 622.25 ns |     408 B |
| ImmutableArray_Builder_KnownLength |    30 | 373.65 ns |     148 B |

Note I didn't use `MoveToImmutable` for `KnownLength`. It's still the best approach in all cases. WIth move it would probably be faster, but my file's almost finished copying...

I guess the variation in allocation is due to different initial array capacities and growth strategies, or fixed overheads. I'm not really sure. There's not much in it.

I'm surprised it's 15-30% faster to use the builder over the list. `CreateRange` does have to do a few type checks, but I wouldn't expect them to account for that much. Haven't looked further.

In short, if it's easy to use the builder, why not.

PR #4321, opened 2018-11-27
Do not warn about using a pre-release sdk to build

jmarolf:
removes noise on the commandline build.


drewnoakes:
Thank you for this. Will make eyeball parsing output easier.

PR #4323, opened 2018-11-27
Fix merge error

### LINKED ISSUE #4313: Dependency node no longer show content of packages
Version: Version 16.0.0 Preview 2.0 [28321.115.d16.0stg]

**Expected:** 
image

**Actual:**
image

drewnoakes:
This error was introduced in 3744d8f6cbb1d03da506e7b30bae4d8acffcac7d

Fixes #4313.


drewnoakes:
> Is there a test that could be added that would fail without this?

I've tried to write sensible tests for the rule handlers, but they end up being 99% setup. There's so much prerequisite state. I hope to do so one day for all of them, but that's more than I have the stomach for today.

I'll merge this now to fix the regression. Integration tests would be a good fit here too.

davidwengier:
> Integration tests would be a good fit here too

I would love to be able to write more integration tests. Most of the stuff I've worked on until now has been impossible to test without mocking out 99% of CPS and VS, which is clearly not something I want to do, and results in a very implementation specific test anyway.

PR #2666, opened 2017-07-26
Updating the property pages to follow the recommended keyboard navigation behavior for 'Navigating Between Window Panes'.

tannergooding:
**Customer scenario**

Customer attempts to navigate using the keyboard

**Bugs this fixes:** 

https://devdiv.visualstudio.com/DevDiv/_queries?id=470148

~~https://github.com/dotnet/project-system/issues/2650 and https://github.com/dotnet/project-system/issues/2651~~

**Workarounds, if any**

None

**Risk**

Low. This just modifies a code path to reactivate the page rather than transfer focus to the next button

**Performance impact**

Low.

**Is this a regression from a previous update?**

Not a regression. This was new code fixing accessibility issues.

**Root cause analysis:**

The focus order was not done in consideration to the guidelines set here: https://msdn.microsoft.com/en-us/library/ms971323.aspx

**How was the bug found?**

Ad-hoc testing

PR #2268, opened 2017-05-23
Updating the DesignerTabControl to allow keyboard navigation on the DesignerTabButton(s)

tannergooding:
**Customer scenario**

Customers who rely on keyboard navigation will not be able to navigate through the available property pages.

**Bugs this fixes:** 

https://devdiv.visualstudio.com/DevDiv/_workitems?id=409661

**Workarounds, if any**

None

**Risk**

Low, this adds the DesignerTabControl to the list of tab stops, but it requires some manual focus invocations to get it functioning with the existing codebase.

**Performance impact**

Low, this only adds a couple of additional checks on the existing code paths to determine if the keyboard based navigation is being used.

**Is this a regression from a previous update?**

No

**Root cause analysis:**

Known accessibility issue

**How was the bug found?**

Accessibility Tenant

PR #4331, opened 2018-11-29
Fixing Tab and Shift-Tab behavior on Property Pages

etbyrd:
This is a fix for: [646608](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646608/) and [646587](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646587/)

In the WinForm property pages, tabbing at the last item in a page will take you to the page's title tab item. For example, tabbing at the "Advanced..." button on the Build property page (the last control on the page) would take you to the "Build" tab. But, If you tabbed again, it would not take you back into the property page as expected, but go down to the next title tab item. There was no way to tab back into the property page, and tabbing kept you in the property page named tabs.

The base for this PR is https://github.com/dotnet/project-system/pull/2666. It is contained in the first commit.

To make the previous comments easier to read, I put them in a table:
<details><summary>Organized Comments from last PR</summary>
<p>

| Dave Kean's Issue     | Tanner's Response  | My Notes
| -                     | -                  | -     
| If I am in the Assembly name textbox on the Application page and I navigate backwards onto the Application tab, my computer beeps | I'm not sure what's causing 1 for you. I am not able to get it to repro currently. | Not a regression. I can repro this. Interestingly, the opposite happens on the Package Property page - shift-tab onto the Package tab does not make a beep, but tabbing all the way around to the tab will beep|
| The currently Focused tab isn't remembered when I tab from the tabs -> page and back, instead it always takes me the current selected tab | Not a regression. The previous implementation did not remember the currently focused tab and also did not provide a way to navigate onto the last control of the page. | Not a regression. I can repro this. |
| Navigating backwards once on the titles - doesn't navigate backwards and take me onto the last control, but rather navigates me forward onto the first control | Not a regression. The previous implementation did not remember the currently focused tab and also did not provide a way to navigate onto the last control of the page. | Not a regression. I can repro this. |
| Starting in the Build events textbox, tabing backwards and then back into the text box enters a TAB character into the box | This is occurring on Shift+Tab, and only on Shift+Tab (when navigating from the first dialog to the build events button). It is likely because the message is continuing to be processed by the text box. | Not a regression. I can repro this. This seems like a WinForms bug - the AcceptTab property is false (and why you can tab and shift-tab in the first place) so it shouldn't even be possible for you to input a tab into the textbox at all. |
| Dirty state of the Tab titles is changed as I tab around the pages: 1. Open Application page, make a change to AssemblyName, SHIFT-TAB to Application title, notice it has * 2. TAB forward onto the page, the * disappears | Something about reactivating the page seems to be causing us to lose the save state. | **This is a new issue.** The dirty marker on the Application tab does go away when tabbed into the page, but the change is still there, pending a save. |
| I can't set focus to the overflow button nor drop it via the keyboard | Doesn't look like we ever handled this, because it is not part of the list of buttons and is instead a standalone control. | Not a regression. This behavior is not handled. |
| I can't navigate from the WPF Debug or Package page onto the page titles | I'm seeing this for WPF Debug but not for the Package page. Probably something was missed since it is hosted WPF rather than WinForms. | This will have to be handled separately in those pages. |
| If you navigate from the Resources (and probably Settings) page onto Page titles, the focus rectangle disappears | Seems the pages in general don't display focus rectangles as well. | Not sure about this one, I can see the focus rectangles, but the shift-tabbing does not work on the the elements in the Resources property page. Either way, it is not a regression. |
| If **underline keyboard shortcuts and access keys** is on, setting focus via the mouse on mouse down on the button and dragging the mouse off the titles does not draw focus rectangles. Also if that setting is off, we give zero indication that button currently has focus at that time. | From David Poeschl: repro on the existing property pages, so not regressions | If you have that setting on, it does not show the focus boxes, but this is not a regression. |
</p>
</details>

<p> </p>

There were some concerns about regressions when it was created, but only one was a new issue - the dirty state of the tab titles was being reset if you shift-tabbed to the title and tabbed back into the property page.

The second commit fixes the issue where shift-tabbing out of property page and tabbing back into would reset the dirty status of the title tab.

The second commit also fixed an issue with the original PR where shift-tab would not behave as expected on the title tab.

This PR is also related to [646614](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646614/), but that is concerning the WPF property pages.

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
@@ -1090,7 +1090,8 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                         Dim PropPageView As PropPageDesigner.PropPageDesignerView
                         PropPageView = TryCast(NewCurrentPanel.DocView, PropPageDesigner.PropPageDesignerView)
                         If PropPageView IsNot Nothing Then
-                            PropPageView.ActivatePage(PropPageView.PropPage)
+                            'We are looping in the same page, do not set the undo status to clean
+                            PropPageView.SetControls(PropPageView.PropPage, True)
                         Else
                             'Must have had error loading
                         End If
@@ -1232,11 +1233,21 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
         Public Overrides Sub OnItemClick(item As ProjectDesignerTabButton, reactivatePage As Boolean)
             Common.Switches.TracePDFocus(TraceLevel.Warning, "ApplicationDesignerView.OnItemClick")
             MyBase.OnItemClick(item, reactivatePage)
-            ShowTab(SelectedIndex, ForceShow:=True, ForceActivate:=reactivatePage)
+            ShowTab(SelectedIndex, ForceShow:=True)
 
             ' we need set back the tab, if we failed to switch...
             If SelectedIndex <> _activePanelIndex Then
                 SelectedIndex = _activePanelIndex
+            End If
+        End Sub
+
+        Friend Overrides Sub SetControl(firstControl As Boolean)
+            Dim NewCurrentPanel As ApplicationDesignerPanel = _designerPanels(SelectedIndex)
+            Dim PropPageView As PropPageDesigner.PropPageDesignerView
+            PropPageView = TryCast(NewCurrentPanel.DocView, PropPageDesigner.PropPageDesignerView)
+            If PropPageView IsNot Nothing Then
+                'We are looping in the same page, do not set the undo status to clean
+                PropPageView.SetControls(PropPageView.PropPage, firstControl)
             End If
         End Sub
 
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/DesignerTabButton.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/DesignerTabButton.vb
@@ -151,28 +151,43 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                     If parent IsNot Nothing Then
                         parent.OnItemClick(Me, reactivatePage:=True)
                     End If
-                Case Keys.Tab, Keys.Up, Keys.Down
-
-                    If keyCode <> Keys.Tab OrElse (keyData And Keys.Control) <> Keys.Control Then                                   ' Don't process if Ctrl+Tab, but do Process for Ctrl+Up and Ctrl+Down
+                Case Keys.Up
+                    Dim parent As ProjectDesignerTabControl = ParentTabControl
+                    If parent IsNot Nothing Then
+                        Dim nextIndex As Int32 = ButtonIndex - 1
+                        If nextIndex < 0 Then
+                            nextIndex = parent.TabButtonCount - 1
+                        End If
+                        Dim nextButton = parent.GetTabButton(nextIndex)
+                        nextButton.Focus()
+                        nextButton.FocusedFromKeyboardNav = True
+                        Return True
+                    End If
+                Case Keys.Down
+                    Dim parent As ProjectDesignerTabControl = ParentTabControl
+                    If parent IsNot Nothing Then
+                        Dim nextIndex As Int32 = ButtonIndex + 1
+                        If nextIndex >= parent.TabButtonCount Then
+                            nextIndex = 0
+                        End If
+                        Dim nextButton = parent.GetTabButton(nextIndex)
+                        nextButton.Focus()
+                        nextButton.FocusedFromKeyboardNav = True
+                        Return True
+                    End If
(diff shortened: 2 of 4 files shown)


davkean:
How did you verify that the first three cases were not regressions?

davkean:
I went back to VS 2010 (it's only prerelease I have installed) and those regressions were introduced by the changes to the tabbing support for the AppDesigner.

jmarolf:
@davkean 	If we regressed this behavior in Dev11 (after VS 2010) shouldn't we file bugs on that? I don't see how *this* PR introduces a regression when VS behaves this way today.

davkean:
The bug was introduced by the change in Dev15 to start handing TABing behavior ourselves, instead of relying on Win32. The follow up PR pointed out above #https://github.com/dotnet/project-system/pull/2666, was "fixing" some of the issues caused by that original change. My feedback back then was, please use Win32 to handling tabbing and we don't have any of these probl3ems. This change is identical to the original one and the one I pushed back on.

davkean:
ie This is a patch on top of a shaky ground. We should strengthen the ground first, which means we don't keep needing to special case these situations.

jmarolf:
@davkean 	its unclear to me what the exact fix is you are suggesting, could you elaborate? What would we do to undo the behavior introduced in Dev15 (and why was this allowed to happen in the first place?)

davkean:
My concrete proposal, is revert https://github.com/dotnet/project-system/pull/2268 and figure out why we can't use normal TAB'ing behavior inherited from WinForms/Win32. Been a long time since I've looked at this, but this addresses the majority of the issues I raised above.

jmarolf:
My understanding is that these are working around known bugs in Winforms that are not going to be fixed for .NET 4.7.2 but maybe I misunderstood something.  @davkean 	can you file a bug about this so we can track it?  If we can just revert most of our custom logic and things will just work then I am all ears.

davkean:
> My understanding is that these are working around known bugs in Winforms that are not going to be fixed for .NET 4.7.2

Point me to how you came to that conclusion?

jmarolf:
@etbyrd 	was stepping through the winforms source on this one so I'll let him reply but that we my understanding of what happened

davkean:
Let's step back:

We are effectively a TabControl (ProjectDesignerTabControl) with TabPages (PropPageDesignerView). A little differently, though, we have DesignerTabButton that represent the "tab buttons".

Here are some examples of TabControls or TabControl-like controls:

image
image

Go and use them, and then use our AppDesigner, can you see the difference?

The difference is that unlike all these other controls where the "strip" of buttons have a single tab stop, each button in our "strip" has a tab stop. **Problem 1.**

Secondly, go and read the code of TabControl and TabPage, do you see any code that specifically handles "TAB" (not TAB+CTRL which is specifically a TabControl behavior)? Instead of relying on Win32 & WinForms to handle TAB, we specifically handling it ourselves. **Problem 2**.

Thirdly, go and read the code of TabControl and TabPage, do you see any code that specially remembers, or handles that the navigation occurred via the keyboard? **Problem 3.**

- **Problem 1:** I think is because we're incorrectly reading the [original bug](https://devdiv.visualstudio.com/DevDiv/_workitems?id=409661), and misinterpreted how we should fix it. I believe the behavior should be identical to the Windows Settings tab control.
- **Problem 2:** Revert https://github.com/dotnet/project-system/pull/2268, override ProcessTabKey in ProjectDesignerTabControl and have it simply navigate the selected PropPageDesignerView. Figure out why navigating backwards from the PropertyPage is not automatically giving focus to its ProjectDesignerTabControl and hence its current DesignerTabButton. We must be blocking this, or turned off IsTabStop or something.
- **Problem 3:** Figure out exact reason why we can't trust ShowFocusCues.

This addresses almost all my issues raised above.

davkean:
I can sit with you @etbyrd and figure out this thing together if you would like.

jmarolf:
@davkean 	alright thanks for this. Let's investigate.  I'll ping you on chat

jjmew:
> why we can't use normal TAB'ing behavior inherited from WinForms/Win32. Been a long time since I've looked at this, but this addresses the majority of the issues I raised above.

My suggestion is that we create a new work item to revisit the design. Clearly we are in shaky ground and we should revisit the design. I don't think that right now is the right time to do this. We have the potential of going down the rabbit hole. I wouldn't block this on redesigning the whole tab interaction. We can prioritize the redesign against our current priorities

davkean:
This change introduces a regression (see above table) and doesn't fully address the original bug (we still have broken pages).

etbyrd:
I agree that we should eventually revert and redesign this - but this change does not introduce any regressions that are not also fixed in this PR. 

The only regression that was introduced was the dirty state of the tabs, which was also fixed in this PR. 

Also, this PR fixes [646587](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646587/) 

I believe this should be merged, because although the existing ground is shaky, it fixes these issues without introducing any regressions.

jjmew:
> This change introduces a regression (see above table) and doesn't fully address the original bug (we still have broken pages).

My understanding is that we are not aware of any regression introduced by this change. I think we should go ahead merge.

PR #4333, opened 2018-11-29
UpToDateCheck uses first (newest) MSBuildAllProjects only

### LINKED ISSUE #3744: BuildUpToDateCheck.OnChanged allocates 4% of all data opening Roslyn.sln

drewnoakes:
Fixes #3744 

//cc: @AndyGerlicher @rainersigwald


drewnoakes:
Next week I'll do another pass of validation and add some explanation in code comments.

drewnoakes:
> You've confirmed that this always includes the latest input?

I ran some tests and this seems to work as expected. My test approach was to manually modify items from the `MSBuildAllProjects` variable (eg. the project's .g.props and .g.targets files) and verify that a project change is observed with the newly edited file as the first item.

PR #4335, opened 2018-11-29
Dependencies change data

drewnoakes:
Some cleanup around how changes to dependency data are gathered and emitted.

- Split dependency change building from modelling
- Reduce the length of time builder objects are kept alive, reducing chance they end up in gen1
- Also recognise that change building only occurs "on the stack", so no locking is needed
- Make allocation of collections lazy
- Rename `DependenciesRuleChangeContext` to `CrossTargetDependenciesChangesBuilder`
- Don't raise `DependenciesChanged` events when there are no changes
- Remove 'AnyChanges' from event args as there are always changes
- Fix some comments that should have been updated in a recent PR


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesChangesBuilder.cs
```diff
@@ -3,58 +3,56 @@
...
             {
-                lock (_added)
-                {
-                    return ImmutableArray.CreateRange(_added);
-                }
+                _added = new HashSet<IDependencyModel>();
             }
+
+            _added.Remove(model);
+            _added.Add(model);
```
davidwengier: I trust that you're really smart and stuff, so can you explain what this is for? Different IDependencyModel objects that have the same hashcode?

drewnoakes: This pattern's been there for a while, I've just moved it. It looks weird, but is necessary because `Add` is a no-op if the item's already present.

There's currently an implicit requirement that an `IDependencyModel` be equatable with others of its own kind. Formalising this is on my radar, but it's part of the public API so it might not be possible or practical to do so.

davidwengier: > is necessary because Add is a no-op if the item's already present.

You say that like its a bad thing....

drewnoakes: You prompted me to look at this again.

`DependencyRulesSubscriber` has a list of `IDependenciesRuleHandler` implementations, which it goes through in a precedence order in response to `IProjectChangeDescription`s.

In theory two different handlers could produce different versions of the same dependency. I don't know if this happens, or if there's a reason to block it. However in this case it's not part of the public API so it's probably safe to change, maybe throwing if `Add` returns `false`.

If it meant we could get rid of the equatability requirement on `IDependency` it might be worth the work.

PR #4337, opened 2018-11-30
Always ensure that the 4.7.2 targeting pack is installed

jmarolf:


davkean:
Installing machine-wide software during the build?

jmarolf:
@davkean 	yep

davkean:
Can we not do that and just make sure the image has the correct prerequisites.

jmarolf:
@davkean 	like have the right version of the .NET Core SDK installed?

https://github.com/dotnet/project-system/blob/2e418bb5ee703a452e582e78b664aafdd7326fdf/build/CIBuild.cmd#L6-L10

In all seriousness just letting it come from the image is the correct approach but it takes weeks/months to get new images deployed. If the turnaround time was reasonable I would just do that.

davkean:
To deploy our bits and run them successfully, we need updated VS versions on these machines that contain at minimum the CPS breaking changes, as well as updated Roslyn bits with the new APIs we consumed from them. When we update the image, can we just make sure the right stuff is installed on it?

jmarolf:
For roslyn I was planning on installing their VSIXes to get around the fact that we can't seem to get new VSIXes but you bring up CPS which is a real problem.  We will need a process that allows us to rapidly create new images.  Ideally the image would be updated nightly with whatever the latest build of VS is.

PR #4340, opened 2018-12-03
Dependencies graph tweaks

drewnoakes:
I'm starting to focus on the "graph" parts of the dependencies node code.

Please review commit-by-commit.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -58,37 +46,44 @@ internal class DependenciesGraphProvider : OnceInitializedOnceDisposedAsync, IGr
...
 
-        private IAsyncServiceProvider ServiceProvider { get; }
+        private readonly IAggregateDependenciesSnapshotProvider _aggregateSnapshotProvider;
+
+        private readonly IAsyncServiceProvider _serviceProvider;
+
+        private GraphIconCache _iconCache;
+
+        [ImportingConstructor]
+        public DependenciesGraphProvider(
```
jmarolf: Why move the constructor around?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -342,11 +332,39 @@ private static GraphNodeId GetTopLevelGraphNodeId(string projectPath, string mod
...
-
-        private static string GetIconStringName(ImageMoniker icon)
+        private sealed class GraphIconCache
         {
-            return s_iconNameCache.GetOrAdd((icon.Id, icon.Guid), i => $"{i.guid:D};{i.id}");
+            /// <summary>The set of icons that have already been registered.</summary>
+            private ImmutableHashSet<ImageMoniker> _registeredIcons = ImmutableHashSet<ImageMoniker>.Empty;
+
+            /// <summary>A cache of icon names, the reuse of which helps reduce allocations.</summary>
+            private readonly ConcurrentDictionary<(int id, Guid guid), string> _iconNameCache = new ConcurrentDictionary<(int id, Guid guid), string>();
```
jmarolf: How many threads are going to access this cache at once?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
drewnoakes: Short answer is: it makes more sense to me this way.

Longer answer is actually a question: Do we have a standard order for members? I tend to use:

- static fields
- static members
- instance fields
- constructors
- computed properties
- methods

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
drewnoakes: I'm not certain. This is the technique in use beforehand. Once populated, this is likely to be fairly read-only, so maybe an immutable interlocked pattern would be more suitable.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
jmarolf: I don't have a preference for ordering.  My only line in the sand is I don't want to be unable to debug because something is not in the correct order.  So I am not super-motivated to enforce anything with analyzers at the moment.

My only objection is that moving things around makes the diff harder to review.  I would prefer if style changes were just separate PRs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
jmarolf: I guess the correct answer is "we should measure" :), but I don't know if we want to pay for the cost of `ConcurrentDictionary` locking and memory complexity if there are not going to be a lot of parallel read/writes

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
drewnoakes: Also not convinced we need analyzers for this.

Understood regarding style changes. Sometimes it's only after moving things around that I see patterns such as pulling out related fields into a new class, as done here. Each commit here is straightforward.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
drewnoakes: To measure meaningfully we'd need to know the level of concurrency, and whether we're seeing regular writes. I suspect writes to be up front then infrequent, after which `ImmutableInterlocked` is a great fit both in terms of memory and locking.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -342,11 +331,39 @@ private static GraphNodeId GetTopLevelGraphNodeId(string projectPath, string mod
                 GraphNodeId.GetPartial(CodeGraphNodeIdName.File, filePath));
         }
 
-        private static readonly ConcurrentDictionary<(int id, Guid guid), string> s_iconNameCache = new ConcurrentDictionary<(int id, Guid guid), string>();
-
-        private static string GetIconStringName(ImageMoniker icon)
+        private sealed class GraphIconCache
         {
-            return s_iconNameCache.GetOrAdd((icon.Id, icon.Guid), i => $"{i.guid:D};{i.id}");
+            /// <summary>The set of icons that have already been registered.</summary>
```
davkean: Do you find these comments useful? I don't - they are pretty much just restating the name of the field.

drewnoakes: No they're not useful. IIRC they moved with the members.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -132,9 +132,14 @@ internal async Task BeginGetGraphDataAsync(IGraphContext context)
...
 
-                IEnumerable<Lazy<IDependenciesGraphActionHandler, IOrderPrecedenceMetadataView>> actionHandlers = GraphActionHandlers.Where(x => x.Value.CanHandleRequest(context));
-                bool shouldTrackChanges = actionHandlers.Aggregate(
-                    false, (previousTrackFlag, handler) => previousTrackFlag || handler.Value.HandleRequest(context));
+                bool shouldTrackChanges = false;
+                foreach (Lazy<IDependenciesGraphActionHandler, IOrderPrecedenceMetadataView> handler in GraphActionHandlers)
+                {
+                    if (handler.Value.CanHandleRequest(context))
+                    {
+                        shouldTrackChanges = shouldTrackChanges || handler.Value.HandleRequest(context);
```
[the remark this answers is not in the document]
drewnoakes: Can't break early, but the `if` statement's easier to read. The `b = b || c` form came from unravelling the `Aggregate` code. I'll update to use `if`.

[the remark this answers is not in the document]
drewnoakes: Hang on, it's early. I take it back -- we can break early.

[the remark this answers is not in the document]
davidwengier: You should take it as a compliment that I double checked a dozen times in my head that an early exit was possible, because you seem to rarely do anything without a good reason.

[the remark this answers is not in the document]
drewnoakes: Flattery will get you beer when you're in London.

PR #4341, opened 2018-12-03
Use object pool

jmarolf:
Roslyn has an internal implementation of object pools that they ship as a [source based nuget package](https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.CodeAnalysis.PooledObjects/2.11.0-beta3-63519-01).  Unfortunately we cannot consume this package directly without changing our build infrastructure.  

I've added them here as a generalized machanism for re-using common builder objects.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
@@ -6,6 +6,7 @@ using System.Collections.Immutable;
 using System.ComponentModel.Composition;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Managed.PooledObjects;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
 {
@@ -39,7 +40,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
             ImmutableDictionary<string, ConfiguredProject> configuredProjectsMap = await _activeConfiguredProjectsProvider.GetActiveConfiguredProjectsMapAsync();
 #pragma warning restore CS0618 // Type or member is obsolete
             ProjectConfiguration activeProjectConfiguration = _commonServices.ActiveConfiguredProject.ProjectConfiguration;
-            ImmutableArray<ITargetFramework>.Builder targetFrameworks = ImmutableArray.CreateBuilder<ITargetFramework>(initialCapacity: configuredProjectsMap.Count);
+            var targetFrameworks = PooledArray<ITargetFramework>.GetInstance();
             ITargetFramework activeTargetFramework = TargetFramework.Empty;
 
             foreach ((string tfm, ConfiguredProject configuredProject) in configuredProjectsMap)
@@ -61,7 +62,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
 
             return new AggregateCrossTargetProjectContext(
                 isCrossTargeting,
-                targetFrameworks.MoveToImmutable(),
+                targetFrameworks.ToImmutableAndFree(),
                 configuredProjectsMap,
                 activeTargetFramework,
                 _targetFrameworkProvider);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
@@ -6,7 +6,6 @@ using System.Collections.Immutable;
 using System.ComponentModel.Composition;
 using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.VisualStudio.ProjectSystem.Managed.PooledObjects;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
 {
@@ -40,7 +39,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
             ImmutableDictionary<string, ConfiguredProject> configuredProjectsMap = await _activeConfiguredProjectsProvider.GetActiveConfiguredProjectsMapAsync();
 #pragma warning restore CS0618 // Type or member is obsolete
             ProjectConfiguration activeProjectConfiguration = _commonServices.ActiveConfiguredProject.ProjectConfiguration;
-            var targetFrameworks = PooledArray<ITargetFramework>.GetInstance();
+            ImmutableArray<ITargetFramework>.Builder targetFrameworks = ImmutableArray.CreateBuilder<ITargetFramework>(initialCapacity: configuredProjectsMap.Count);
             ITargetFramework activeTargetFramework = TargetFramework.Empty;
 
             foreach ((string tfm, ConfiguredProject configuredProject) in configuredProjectsMap)
@@ -62,7 +61,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
 
             return new AggregateCrossTargetProjectContext(
                 isCrossTargeting,
-                targetFrameworks.ToImmutableAndFree(),
+                targetFrameworks.MoveToImmutable(),
                 configuredProjectsMap,
                 activeTargetFramework,
                 _targetFrameworkProvider);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
@@ -2,7 +2,7 @@
 
 using System;
 using System.Collections.Immutable;
-using Microsoft.VisualStudio.ProjectSystem.Managed.PooledObjects;
+using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.Imaging.Interop;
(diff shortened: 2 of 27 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -39,7 +40,7 @@ public async Task<AggregateCrossTargetProjectContext> CreateProjectContextAsync(
             ImmutableDictionary<string, ConfiguredProject> configuredProjectsMap = await _activeConfiguredProjectsProvider.GetActiveConfiguredProjectsMapAsync();
 #pragma warning restore CS0618 // Type or member is obsolete
             ProjectConfiguration activeProjectConfiguration = _commonServices.ActiveConfiguredProject.ProjectConfiguration;
-            ImmutableArray<ITargetFramework>.Builder targetFrameworks = ImmutableArray.CreateBuilder<ITargetFramework>(initialCapacity: configuredProjectsMap.Count);
+            var targetFrameworks = PooledArray<ITargetFramework>.GetInstance();
```
drewnoakes: What's the impact of losing the capacity specifier here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
```diff
@@ -95,10 +92,10 @@ public Dependency(IDependencyModel dependencyModel, ITargetFramework targetFrame
             else
             {
                 int count = dependencyModel.DependencyIDs.Count;
-                ImmutableArray<string>.Builder ids = ImmutableArray.CreateBuilder<string>(count);
+                var ids = PooledArray<string>.GetInstance();
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
```diff
@@ -334,39 +331,18 @@ public static string GetID(ITargetFramework targetFramework, string providerType
...
-            finally
-            {
-                // Prevent holding on to large builders
-                if (sb?.Capacity < 1000)
-                {
-                    sb.Clear();
-                    s_builderPool.Add(sb);
-                }
-            }
+            var sb = PooledStringBuilder.GetInstance();
```
drewnoakes: Ditto.

Also, with this change `s_builderPool` should probably be removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/ObjectPool.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// Not returning objects to the pool in not detrimental to the pool's work, but is a bad practice. 
+    /// Rationale: 
+    ///    If there is no intent for reusing the object, do not use pool - just use "new". 
+    /// </summary>
+    internal class ObjectPool<T> where T : class
+    {
+        [DebuggerDisplay("{Value,nq}")]
+        private struct Element
+        {
+            internal T Value;
```
drewnoakes: This is a nice trick for safe, bounds checked array element references.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsProvider.cs
```diff
@@ -121,7 +121,7 @@ public async Task<ActiveConfiguredObjects<ProjectConfiguration>> GetActiveProjec
 
             IImmutableSet<ProjectConfiguration> configurations = await _services.ProjectConfigurationsService.GetKnownProjectConfigurationsAsync();
 
-            ImmutableArray<ProjectConfiguration>.Builder builder = ImmutableArray.CreateBuilder<ProjectConfiguration>(configurations.Count);
+            var builder = PooledArray<ProjectConfiguration>.GetInstance();
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsProvider.cs
```diff
@@ -101,7 +101,7 @@ public async Task<ActiveConfiguredObjects<ConfiguredProject>> GetActiveConfigure
             if (configurations == null)
                 return null;
 
-            ImmutableArray<ConfiguredProject>.Builder builder = ImmutableArray.CreateBuilder<ConfiguredProject>(configurations.Objects.Count);
+            var builder = PooledArray<ConfiguredProject>.GetInstance();
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ContextHandlerProvider.cs
```diff
@@ -56,8 +56,8 @@ public void ReleaseHandlers(IWorkspaceProjectContext context)
 
         private Handlers CreateHandlers(IWorkspaceProjectContext context)
         {
-            ImmutableArray<(IProjectEvaluationHandler handler, string evaluationRuleName)>.Builder evaluationHandlers = ImmutableArray.CreateBuilder<(IProjectEvaluationHandler handler, string evaluationRuleName)>(s_handlerFactories.Length);
-            ImmutableArray<ICommandLineHandler>.Builder commandLineHandlers = ImmutableArray.CreateBuilder<ICommandLineHandler>(s_handlerFactories.Length);
+            var evaluationHandlers = PooledArray<(IProjectEvaluationHandler handler, string evaluationRuleName)>.GetInstance();
+            var commandLineHandlers = PooledArray<ICommandLineHandler>.GetInstance();
```
drewnoakes: Ditto x 2.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/PooledArray.cs
```diff
@@ -0,0 +1,437 @@
...
+            {
+                builder.Add(fillWithValue);
+            }
+
+            return builder;
+        }
+
+        public static ObjectPool<PooledArray<T>> CreatePool()
+        {
+            return CreatePool(128); // we rarely need more than 10
```
drewnoakes: This line confuses me.

jmarolf: `128` is the size of the array of objects, however in practice (for roslyn where this was stolen) it rarely allocates 128 objects.  most of the objects in the array are null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
drewnoakes: In cases where we know _exactly_ what the array size will be upon construction, we do not want to use pooled builders. It's better to allocate the array size exactly then move it to the returned `ImmutableArray<>`. A pooled builder performs a defensive copy, in addition to potentially resizing during construction.

On that basis, the three removals of `MoveToImmutable` are all regressions.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/PooledArray.cs
drewnoakes: I was mostly referring to the comment.

drewnoakes: ...which I see you've just fixed. 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
jmarolf: @drewnoakes 	did some prefview trace'n and confirmed that memory and cpu time is sub-optimal for all cases if we do not use `MoveToImmutable`.  I'll revert this change for those cases

davkean:
Why can't we reuse their package without a change to build infrastructure?

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/ObjectPool.cs
```diff
@@ -0,0 +1,145 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Managed.PooledObjects
```
davkean: We don't use "Managed" in a namespace. For this, I'd copy the CoreFx naming; "VisualStudio.Buffers".

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsProvider.cs
davkean: This was should be reverted, right? We know the size upfront.

davkean: Strike that, this is only the "maximum", the resulting size might be different, carry one.

jmarolf:
> but I'd really love if we could just reuse the existing source.

No problem the source in [this nupkg](https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.CodeAnalysis.PooledObjects/2.11.0-beta3-63519-01) would require the following:

1. Modify global suppression to ignore these files and update the suppressions as IDE fixes change (the source in roslyn is very different style wise from what we enforce)
2. Be OK with the files being placed at the root of the project like so:
image
3. Be OK with contributors being able to modify these files on disk and having it work locally but fail on PR
4. Being unable to change the default pool sizes without a PR to rolsyn.

I would prefer to just use the nuget package as well so long as none of that raises objections.

davkean:
I could go either way, I'll let you decide.

PR #4346, opened 2018-12-04
Move rule handlers into sub-namespace

drewnoakes:
These types are all internal.

Splitting the namespace makes it a bit easier to see the mechanics here.


drewnoakes:
@jmarolf what does 🕐 mean?

jmarolf:
@drewnoakes 	it typically indicated to github that the reviewer is waiting on the original poster of a response.  I mis-clicked.  You can ignore

Issue #3156, opened 2018-01-18
ResXFileCodeGenerator does not run on linked files


nguerrera:
## Repro steps

* [repro.zip](https://github.com/dotnet/project-system/files/1641163/repro.zip)
* Open CPS\ConsoleApp29.sln
* Edit linked file Resources\Resource1.resx in designer, save

## Expected result
* Resource1.Designer.cs is updated with changes made to resx
* Note that this worked in classic project system (see Classic\ConsoleApp31.sln)

## Actual result
* No changes are applied to Resource1.Desgner.cs

Pilchie:
@brettfo - is this what you were seeing yesterday in the F# build?

davkean:
I suspect not related to #2687 - I suspect a CPS issue.

ocallesp:
PR https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/CPS/pullrequest/253809

PR #4349, opened 2018-12-05
Create a TempPE manager and compile TempPE DLLs as required

### LINKED ISSUE #4060: Compile TempPE output
This is a tracking item for the work to wire up the TempPE compiler in order to bring up WinForms

- [ ] Compile TempPE as part of file generator for resx files
- [ ] Compile TempPE via BuildManager for design time builds (#4163)

### LINKED ISSUE #4163: Provide an implementation of BuildManager to produce TempPE assembly during type resolution

davidwengier:
Fixes #4163 and fixes #4060 

This PR contains:
* An implementation of VSLangProj.BuildManager, largely copied from CPS, that hooks up our TempPEBuildManager
* A TempPEBuildManager that lazily initialiaes itself when the first TempPE information is requested
  * TempPEBuildManager inherits from UnconfiguredProjectHostBridge so that data flow changes to project files flow to it in order to update TempPE information and file TempPE events
* A hint receiver for triggering recompiles and firing events when single file generators run, that only starts triggering after the first TempPE info is requested
* A file update action handler for triggering recompiles and firing events when manual file edits are made to TempPE inputs, that only starts triggering after the first TempPE info is requested
* Some tests of processing data flow updates

This can't be merged until Roslyn and CPS dependencies are new enough but it is feature complete as far as parity with legacy goes.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+        }
+
+        private static string GetOutputFileName(string outputPath, string fileName)
+        {
+            return Path.Combine(outputPath, fileName.Replace('\\', '.') + ".dll");
+        }
+
+        protected virtual Task CompileTempPEAsync(HashSet<string> filesToCompile, string outputFileName)
+        {
+            return _sequentialTaskQueue.ExecuteTask(async () =>
```
davidwengier: The sequential task queue is here to ensure that TempPE files get compiled in the order they were asked for. I cannot fathom a way that there could ever be multiple tasks in the queue because we always block the UI thread when compiling, but I left this here for parity with legacy, and to trigger some discussion around it.

Legacy has an event queue that it adds events to in order to compile TempPEs, however it too waits on the UI thread for a compilation, and it never adds an item to the queue without also immediately draining the queue and compiling all items, so I similarly cannot fathom a way that legacy could ever has multiple items in its queue. If anyone has any context around that, or wants to go and look at [the code](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2Flangbuild%2FCLangTempPEManager.cpp&version=GBmaster&line=795&lineStyle=plain&lineEnd=796&lineStartColumn=1&lineEndColumn=1) and work out why it might have a queue, I'd welcome it.

If nobody can work out why I don't see a reason for this to exist, though it is a fairly low overhead insurance policy.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,135 @@
...
+                return _tempPEManager.Value.GetTempPESourceFileNames();
+            }
+        }
+
+        /// <summary>
+        /// Builds a temporary portable executable (PE) and returns its description in an XML string.
+        /// </summary>
+        public string BuildDesignTimeOutput(string bstrOutputMoniker)
+        {
+            if (bstrOutputMoniker == null)
```
davkean: Requires.NotNull.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            DesignTimeOutputDeleted?.Invoke(outputMoniker);
+        }
+
+        /// <summary>
+        /// Occurs when a design time output moniker is dirty.
+        /// </summary>
+        internal virtual void OnDesignTimeOutputDirty(string outputMoniker)
+        {
+            UIThreadHelper.VerifyOnUIThread();
```
davkean: _threadingSevice.VerifyOnUIThread(), we don't need to use the static helper.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="sourceFile">A source file that needs a TempPE library created. Must be one of the values returned from <see cref="GetTempPESourceFileNames"/>.</param>
+        /// <returns>An XML string formatted the same as the return from <see cref="VSLangProj.BuildManager.BuildDesignTimeOutput(string)"/></returns>
+        /// <remarks>The XML format is documented at https://docs.microsoft.com/en-us/dotnet/api/vslangproj.buildmanager.builddesigntimeoutput </remarks>
+        Task<string> GetTempPEDescriptionXmlAsync(string sourceFile);
+        
+        /// <summary>
+        /// Fires the TempPEDirty event if the source file is a file that the ITempPEBuildManager cares about
+        /// </summary>
+        /// <param name="sourceFile">The evaluated include of any source file</param>
+        Task TryFireTempPEDirtyAsync(string sourceFile);
```
davkean: Why "Try"? It doesn't return a result.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// Provides TempPE build information and services to the build manager
+    /// </summary>
+    internal interface ITempPEBuildManager
+    {
+        /// <summary>
+        /// Get the list of filenames that need to have TempPE libraries created
+        /// </summary>
+        /// <returns></returns>
+        string[] GetTempPESourceFileNames();
```
davkean: You have "TempPE" name already in the interface, do you need it in the members?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+            Lazy<ITempPEBuildManager> tempPEBuildManager)
+        {
+            _buildManager = buildManager;
+            _tempPEBuildManager = tempPEBuildManager;
+        }
+
+        public async Task<bool> TryHandleFileSavedAsync(IProjectTree tree, string newFilePath, bool saveAs)
+        {
+            // We don't want to be the one to realise the Lazy<ITempPEManager> so ignore changes in projects that don't need TempPE
+            if (!_buildManager.ProjectNeedsTempPE) return false;
```
davkean: This feels like the wrong dependency; this class only a dependency on BuildManager just to ask if ITempPEBuildManager is actually doing work.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+    [Export(typeof(IFileActionHandler))]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
+    internal class TempPEFileSaveWatcher : IFileActionHandler
+    {
+        private readonly VSBuildManager _buildManager;
+        private readonly Lazy<ITempPEBuildManager> _tempPEBuildManager;
+
+        [ImportingConstructor]
+        public TempPEFileSaveWatcher(
+            [Import(typeof(BuildManager))]VSBuildManager buildManager,
```
davkean: Again this class says "I only have a dependency on BuildManager" but it really has a dependency on VSBuildManager.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileEditHintReceiver.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            // We don't want to be the one to realise the Lazy<ITempPEManager> so ignore changes in projects that don't need TempPE
+            if (!_buildManager.ProjectNeedsTempPE) return;
+
+            IImmutableSet<IProjectChangeHint> files = hints[ProjectChangeFileSystemEntityHint.EditedFile];
+            foreach (IProjectChangeFileHint hint in files)
+            {
+                foreach (string fileName in hint.Files)
+                {
+                    // The magic of MEF means that the instance we get from .Value is the same as the build managers instance, so it has all of the info it needs
```
davkean: This isn't magic, this is as a result of shared instances. The couple here is I think needs to be removed - I don't it's BuildManager concept to know if TempPE is running.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.VisualStudio.ProjectSystem.VS.Automation;
+using VSLangProj;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    /// <summary>
+    /// This class handles being notified whenever a source file is saved by the hierarchy
+    /// </summary>
+    [Export(typeof(IFileActionHandler))]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
```
davkean: We don't need this in shared projects; this should be CSharpOrVisualBasicLanguageServices.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            string path = tree.BrowseObjectProperties.ItemName;
+            // Save As will come through as a file rename so we don't need to double handle
+            if (!saveAs && path != null)
+            {
+                // The magic of MEF means that the instance we get from .Value is the same as the build managers instance, so it has all of the info it needs
+                await _tempPEBuildManager.Value.TryFireTempPEDirtyAsync(path);
+            }
+
+            // we're just listening in here, not taking action, so return false always
```
davkean: Anyone running before that handles save can prevent us from seeing this notification.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+                if (!result)
+                {
+                    try
+                    {
+                        if (_fileSystem.FileExists(outputFileName))
+                        {
+                            _fileSystem.RemoveFile(outputFileName);
+                        }
+                    }
+                    catch (IOException)
```
davkean: "UnauthorizedAccessException" as well.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+            return _sequentialTaskQueue.ExecuteTask(async () =>
+            {
+                bool result = await _compiler.CompileAsync(_languageServiceHost.ActiveProjectContext, outputFileName, filesToCompile, default);
+
+                // if the compilation failed or was cancelled we should clean up any old TempPE outputs lest a designer gets the wrong types, plus its what legacy did
+                if (!result)
+                {
+                    try
+                    {
+                        if (_fileSystem.FileExists(outputFileName))
```
davkean: RemoveFile already does this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+        internal bool CompilationNeeded(HashSet<string> files, string outputFileName)
+        {
+            if (!_fileSystem.FileExists(outputFileName))
+                return true;
+
+            DateTime outputDateTime = _fileSystem.LastFileWriteTime(outputFileName);
+
+            foreach (string file in files)
+            {
+                DateTime fileDateTime = _fileSystem.LastFileWriteTime(file);
```
davkean: LastFileWriteTimeUtc

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+  />
+</root>";
+        }
+
+        internal bool CompilationNeeded(HashSet<string> files, string outputFileName)
+        {
+            if (!_fileSystem.FileExists(outputFileName))
+                return true;
+
+            DateTime outputDateTime = _fileSystem.LastFileWriteTime(outputFileName);
```
davkean: LastFileWriteTimeUtc

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: What if this throw UnauthorizedAccessException/IOException?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: What if this throws?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
davidwengier: It might not fire the event, because it gets called whether the sourceFile parameter is a TempPE input or not.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+                if (fileDateTime > outputDateTime)
+                    return true;
+            }
+
+            return false;
+        }
+
+        private static string GetOutputFileName(string outputPath, string fileName)
+        {
+            return Path.Combine(outputPath, fileName.Replace('\\', '.') + ".dll");
```
davkean: What is "fileName", is this relative to the project? Can we name it as such?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+
+        /// <summary>
+        /// Called privately to actually fire the TempPE events and optionally recompile the TempPE library for the specified inpu
+        /// </summary>
+        private async Task FireTempPEDirtyAsync(string fileName, bool shouldCompile)
+        {
+            // Not using use the ThreadingService property because unit tests
+            await _unconfiguredProjectServices.ThreadingService.SwitchToUIThread();
+
+            if (shouldCompile)
```
davkean: Why do we fire OnDesignTimeOutputDirty when we didn't compile anything?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+            bool hasAddedDesignTimeSharedInputs = false;
+
+            if (value.HasFileChanges)
+            {
+                var designTimeInputs = previousValue.Inputs.ToBuilder();
+                var designTimeSharedInputs = previousValue.SharedInputs.ToBuilder();
+
+                foreach (string item in value.AddedItems)
+                {
+                    // If a property changes we might be asked to add items that already exist, so we just ignore them as we're happy using our existing CancellationSeries
```
davkean: This comment is out of date.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+            ImmutableHashSet<string> newDesignTimeInputs;
+            ImmutableHashSet<string> newSharedDesignTimeInputs;
+            var addedDesignTimeInputs = new List<string>();
+            var removedDesignTimeInputs = new List<string>();
+            bool hasRemovedDesignTimeSharedInputs = false;
+            bool hasAddedDesignTimeSharedInputs = false;
+
+            if (value.HasFileChanges)
+            {
+                var designTimeInputs = previousValue.Inputs.ToBuilder();
```
davkean: Can we switch to non-var usage? I have no idea what this type is.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+                    }
+                }
+
+                newDesignTimeInputs = designTimeInputs.ToImmutable();
+                newSharedDesignTimeInputs = designTimeSharedInputs.ToImmutable();
+            }
+            else
+            {
+                // If there haven't been file changes we can just flow our previous collections to the new version to avoid roundtriping our collections to builders and back
+                // Normally ShouldValueBeApplied can be used for this but our project properties require comparison to previous values so we have to apply every update
```
davkean: Can you expand or reword this? I'm not sure what "project properties require comparison to previous values" means.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davkean: We had this conversation from last PR that wasn't resolved: https://github.com/dotnet/project-system/pull/4318#discussion_r236899719.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davidwengier: Yes, that is true, and it would also prevent single file generators from running. I couldn't see any other way to find out about files changing. There is a data flow update when it happens, but there is no information about what file changed, so I'm using this to avoid having to keep a list of inputs in memory with their last write times, and then checking last write times on every update.

Even if there is an action handler that blocks this from running, when anything actually asks for the TempPE it will still be recompiled as necessary due to the file date checking on the input file and DLL.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Provides TempPE build information and services to the build manager
+    /// </summary>
+    internal interface ITempPEBuildManager
+    {
+        /// <summary>
+        /// Get the list of filenames that need to have TempPE libraries created
+        /// </summary>
+        /// <returns></returns>
```
davkean: Remove, it's empty.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        [ImportMany(ExportContractNames.VsTypes.VSProject)]
+        internal OrderPrecedenceImportCollection<VSLangProj.VSProject> Project { get; }
+
+        /// <summary>
+        /// Occurs when a design time output moniker is deleted.
+        /// </summary>
+        public event _dispBuildManagerEvents_DesignTimeOutputDeletedEventHandler DesignTimeOutputDeleted;
+
+        /// <summary>
```
davkean: We don't program against this type, so these comments are already represented by the interface it implements, and hence we don't need them.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+
+        /// <summary>
+        /// ApplyAsync is called on the UI thread and its job is to update AppliedValue to be correct based on the changes that have come through data flow after being processed
+        /// </summary>
+        protected override async Task ApplyAsync(DesignTimeInputsDelta value)
+        {
+            // Not using use the ThreadingService property because unit tests
+            _unconfiguredProjectServices.ThreadingService.VerifyOnUIThread();
+
+            DesignTimeInputsItem previousValue = AppliedValue.Value;
```
davkean: I can't find this type?

davidwengier: This, and `DesignTimeInputsDelta` are at the bottom of the file. Seemed to be the convention of other ProjectHostBridge classes.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+                Inputs = newDesignTimeInputs,
+                SharedInputs = newSharedDesignTimeInputs,
+                RootNamespace = value.RootNamespace,
+                OutputPath = value.OutputPath
+            }, value.DataSourceVersions);
+
+
+            // Fire off any events necessary
+
+            // We ignore blank values for previous value so that the initial load of the project doesn't trigger a compile of all PEs. It also makes testing easier.
```
davkean: This doesn't sound correct, at least for RootNamespace, it's definitely value to go from empty -> value -> empty.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+                // False                  | True              | PreprocessAddItem will return true, ApplyAsync will process as an add, so all good
+                // True                   | False             | PreprocessAddItem will return false, ApplyAsync will process remove, so all good
+                // True                   | True              | PreprocessAddItem will return true, ApplyAsync will not process add because key already exists
+                if (!PreprocessAddItem(item))
+                {
+                    removedDesignTimeInputs.Add(item);
+                }
+            }
+
+            string rootNamespace = project.GetPropertyValue(ConfigurationGeneral.RootNamespaceProperty);
```
davkean: This should be gathered by a "ConfigurationGeneral" rule (similar to your Compile rule) that is an input for HostBridge, here reading from the evaluated instance bypassing all the rule & caching layer.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+        public ImmutableHashSet<string> Inputs { get; set; } = ImmutableHashSet.Create(StringComparers.Paths);
+        public ImmutableHashSet<string> SharedInputs { get; set; } = ImmutableHashSet.Create(StringComparers.Paths);
+        public string OutputPath { get; internal set; }
+        public string RootNamespace { get; internal set; }
+    }
+
+    internal class DesignTimeInputsDelta
+    {
+        public bool ShouldCompile { get; set; } = true;
+        public ImmutableArray<string> AddedSharedItems { get; set; } = ImmutableArray.Create<string>();
```
davkean: While the same thing, we usually use "Empty" for things like this so that someone doesn't think "Oh does Create return a new value?".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// This method is where we tell data flow which blocks we're interested in receiving updates for
+        /// </summary>
+        protected override IDisposable LinkExternalInput(ITargetBlock<IProjectVersionedValue<InputTuple>> targetBlock)
+        {
+            return ProjectDataSources.SyncLinkTo(
+                _projectSubscriptionService.ProjectSource.SourceBlock.SyncLinkOptions(),
```
davkean: Below when I said don't read from "ProjectInstance", You want ProjectRuleSource passing it ConfigurationGeneral similar to generators: http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem&leftSymbol=ixccm82ohk0j&rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=CodeGenerators%5CSingleFileGeneratorsService.cs&line=158.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Mainly to match legacy behaviour. VSTypeResolutionService gets the event, and if it asks for the TempPE info again we'll recompile then. Legacy doesn't immediately compile for any shared input changes so I mirrored that behaviour.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+            ImmutableArray<string>.Builder addedDesignTimeSharedInputs = ImmutableArray.CreateBuilder<string>();
+
+            foreach (string item in changes.Difference.AddedItems)
+            {
+                PreprocessAddItem(item);
+            }
+
+            foreach (string item in changes.Difference.RemovedItems)
+            {
+                // the RemovedItems doesn't have any project info any more so we can't tell if an item was a design time input or not
```
davkean: It does; IProjectChangeDescription.Before.Items contains metadata.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Thank you, I never did see ProjectRuleSource ever have any information in it, and the extra things passed to SyncLinkOptions probably explains that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileEditHintReceiver.cs
```diff
@@ -0,0 +1,51 @@
...
+    [ProjectChangeHintKind(ProjectChangeFileSystemEntityHint.EditedFileAsString)]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
+    internal class TempPEFileEditHintReceiver : IProjectChangeHintReceiver
+    {
+        private readonly VSBuildManager _buildManager;
+        private readonly Lazy<ITempPEBuildManager> _tempPEBuildManager;
+
+        [ImportingConstructor]
+        public TempPEFileEditHintReceiver(
+            [Import(typeof(BuildManager))]VSBuildManager buildManager,
```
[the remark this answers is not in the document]
davidwengier: To be clear, this means adding a second `Export` to `VSBuildManager`, correct?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: Why would we want VSTypeResolutionService to do anything in response to this change if the dll we return hasn't changed?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+    internal class TempPEBuildManager : UnconfiguredProjectHostBridge<IProjectVersionedValue<InputTuple>, DesignTimeInputsDelta, IProjectVersionedValue<DesignTimeInputsItem>>, ITempPEBuildManager
+    {
+        protected readonly IUnconfiguredProjectCommonServices _unconfiguredProjectServices;
+        private readonly ILanguageServiceHost _languageServiceHost;
+        private readonly IActiveConfiguredProjectSubscriptionService _projectSubscriptionService;
+        private readonly IFileSystem _fileSystem;
+        private readonly SequentialTaskExecutor _sequentialTaskQueue = new SequentialTaskExecutor();
+
+        // protected for unit test purposes
+        protected ITempPECompiler _compiler;
```
davkean: Why? You pass the thing in the constructor why can't you use that?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davidwengier: There were a few questions I asked on that PR that didn't get answered, but the commit stream was getting too noisy so I closed it. Now that I've finished all of the functionality I intend this to be the "real" PR so will keep things cleaner, and hopefully will get some more answers. That comment thread you linked to still doesn't clarify for me what you mean by suggesting I change the `AppliedTo` as that capability doesn't exist.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: Can we split them out into a separate file (but nested), to match tree convention.

davkean:
One thing that isn't clear to me is the how you are handling "file names" throughout the system; it's perfectly valid to have both relative and rooted file names:

```
<ItemGroup>
   <Compile Include="C:\Project\Source.cs" />
   <Compile Include="Source.cs" />
   <Compile Include="..\Source.cs" />
   <Compile Include="C:Source.cs" />
   <Compile Include="E:\AnotherProject\Source.cs" />
</ItemGroup>
```

The code above appears to be making the assumption that the item is always "relative" to the project.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileEditHintReceiver.cs
```diff
@@ -0,0 +1,51 @@
...
+            Lazy<ITempPEBuildManager> tempPEBuildManager)
+        {
+            _buildManager = buildManager;
+            _tempPEBuildManager = tempPEBuildManager;
+        }
+
+        public async Task HintedAsync(IImmutableDictionary<Guid, IImmutableSet<IProjectChangeHint>> hints)
+        {
+            // We don't want to be the one to realise the Lazy<ITempPEManager> so ignore changes in projects that don't need TempPE
+            if (!_buildManager.ProjectNeedsTempPE) return;
```
davkean: I think here you need to separate the concept from; _ITempPEManager has been created_ vs. _it is actually doing work_. Once you start having a large set of Lazy<>'s, the overhead adds up. In total, the 3 `Lazy<ITempPEBuildManager>` (and MEF baggage) are actually bigger in size than ITempPEBuildManager itself. I would remove this "ProjectNeedsTempPE" concept.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+            _buildManager = buildManager;
+            _tempPEBuildManager = tempPEBuildManager;
+        }
+
+        public async Task<bool> TryHandleFileSavedAsync(IProjectTree tree, string newFilePath, bool saveAs)
+        {
+            // We don't want to be the one to realise the Lazy<ITempPEManager> so ignore changes in projects that don't need TempPE
+            if (!_buildManager.ProjectNeedsTempPE) return false;
+
+            string path = tree.BrowseObjectProperties.ItemName;
```
davkean: This isn't a path.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davkean: Can we chat about this? In the world of branch switches I don't think listening to Saves is the right approach.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davkean: CSharpOrVisualBasic is not correct, given this is not applicable to Shared Projects.
DotNetLanguageService is not correct given this is not applicable to F#.

This needs to be either "CSharpOrVisualBasic" + "LanguageService", a new capability.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileEditHintReceiver.cs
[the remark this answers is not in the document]
davkean: Are we the only ones importing BuildManager? Or does CPS import this?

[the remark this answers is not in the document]
davidwengier: CPS imports it in its OAVSProject

davkean:
I've done a quick pass; note I still want to dig deep into the logic of the compilation, some random thoughts:

- BuildManager can be entered by any thread
- Branch switches might result in a bunch of changes to a tonne of files across the solution. I don't think we want to compile all those on the UI thread.
- We have a compile on demand concept but also a precompile concept, but we don't verify or check that we're not already compiling the thing you are asking for
- Switching between between a branches might result in a tonne of throwaway work or useless work, but we'll keep doing that work.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: From reading the legacy code the only thing I can deduce is that it would have been done this way for performance. When a shared input changes it fires the dirty event for all inputs so recompiling all inputs could take time. If there isn't a designer open, then nobody is listening to the events, so a recompile will be delayed until the next time someone asks for a dll.

> the dll we return hasn't changed

When we get asked to return it, we check the file date and recompile it, so by the time it is returned it will have changed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: I'll change it in this case, since I'm just catching all exceptions anyway, but this feels like coding to an implementation detail, which I don't think is normally a good idea.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: I tested with non-existent files and LastFileWriteTime doesn't throw, but its a good idea to add some belt and braces as I'm certainly not doing exhaustive testing around this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: This just makes my test harness easier, so the test compiler can be given a reference to the test build manager.

I can rework the test infrastructure if this is important enough to worry about.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileEditHintReceiver.cs
davidwengier: I think I just went a bit far here with responding to your comment about not being loaded unless its needed. I can remove this concept easily, it was a late addition anyway, it will just mean there will be a TempPEBuildManager created for every unconfigured project, whether its needed or not. The data flow structures still won't be initialised and collecting data until TempPE is needed though.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileEditHintReceiver.cs
davidwengier: This is related to https://github.com/dotnet/project-system/pull/4349#discussion_r239280321 and will be removed with that change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davidwengier: This is related to https://github.com/dotnet/project-system/pull/4349#discussion_r239280321 and will be removed with that change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davidwengier: This is related to https://github.com/dotnet/project-system/pull/4349#discussion_r239280321 and will be removed with that change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davidwengier: This feels a lot like bikeshedding, as I suspect if I'd called this `itemName` then it simply would have elicited a comment of "Why are you passing an item name to a method that excepts a path?".

Additionally I think it looks a lot like a path.
image

For reference the above screenshot is from this definition in the project file:
```
    <Compile Include="..\..\OtherResources\Relative\Relative.Designer.cs">
      <AutoGen>True</AutoGen>
      <DesignTime>True</DesignTime>
    </Compile>
```

I'm open to any suggestions anyone might have around this however, I don't feel strongly about it at all.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: I edit the project file and change `<RootNamespace>Foo</RootNamespace>` -> `<RootNamespace></RootNamespace>`. Or I use Property Pages against a CPS project and change Default namespace to empty.

davidwengier:
> it's perfectly valid to have both relative and rooted file names

I've done some testing around this and it seems like there are quite a few kinks when doing this. In legacy when a relative .resx file is changed its generated file gets a `Link` property added which means its excluded from TempPE. Not all code paths check that however so at times a TempPE is still produced. For CPS projects with a relative resource file the single file generator doesn't regenerate the code behind file at all.

Exploring this did show up a bug in my code for handling the Link property (I thought it was a boolean because the legacy code takes a boolean parameter) so I will fix that. Enabling support for advanced scenarios where user files with relative paths are marked as DesignTime in the project file should be straight forward to add.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: > I edit the project file and change <RootNamespace>Foo</RootNamespace> -> <RootNamespace></RootNamespace>

This is the scenario I mentioned above that is broken. After doing this File -> New Class generates a file that doesn't compile (produces `namespace ` as the namespace), as will running the single file generator for a resource file in a subfolder of the project (produces `namespace .FolderName` as the namespace). The single file generator for a `.settings` file in the `Properties` folder didn't run at all.

> Or I use Property Pages against a CPS project and change Default namespace to empty.

This writes `<RootNamespace />` to the project file so results in the same broken experience as above.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: This has been resolved by changes due to the suggestion ProjectRuleSource hint in https://github.com/dotnet/project-system/pull/4349#discussion_r238940967

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,462 @@
...
+                ShouldCompile = (previousOutput != null),
+
+                DataSourceVersions = input.DataSourceVersions
+            };
+
+            return Task.FromResult(result);
+
+            bool PreprocessAddItem(string item)
+            {
+                ProjectItemInstance projItem = project.GetItemsByItemTypeAndEvaluatedInclude(Compile.SchemaName, item).FirstOrDefault();
```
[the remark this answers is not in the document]
davidwengier: This has been resolved by changes due to the suggestion ProjectRuleSource hint in https://github.com/dotnet/project-system/pull/4349#discussion_r238940967

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: This has been resolved by changes due to the suggestion ProjectRuleSource hint in https://github.com/dotnet/project-system/pull/4349#discussion_r238940967

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: This has been resolved by changes due to the suggestion ProjectRuleSource hint in https://github.com/dotnet/project-system/pull/4349#discussion_r238940967

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: This has been resolved by changes due to the suggestion ProjectRuleSource hint in https://github.com/dotnet/project-system/pull/4349#discussion_r238940967

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEFileSaveWatcher.cs
davkean: Just because it looks like a path, doesn't make it path. Regardless, I won't dig into this until we've thought about this approach - gone are the days where all manipulation of a file occurs via VS when the code in legacy was originally written. A simply branch switch will cause you to miss a change. I think we want to either watch these files, or have Roslyn figure that out. It already watches all buffers and files on disk.

davidwengier:
> Enabling support for advanced scenarios where user files with relative paths are marked as DesignTime in the project file should be straight forward to add.

Turns out this in not necessary, there is no combination of project file craziness I can create that results in a relative or rooted path being represented as anything without a Link property, which is [not a valid TempPE input](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2Flangproj%2FCLangFileNode.cpp&version=GBmaster&line=2975&lineStyle=plain&lineEnd=2978&lineStartColumn=1&lineEndColumn=1) so the paths in question for this code will always be relative to the project.

davkean:
Related: https://github.com/dotnet/project-system/issues/3156.

davidwengier:
I've pushed commits to resolve all of the comments except for the ones that require further discussion. Each new commit resolves a single issue and is probably easier to review individually.

davidwengier:
Moved to use the VS file watcher which handily covers the single file generator case, as well as directly modifying an input.

davidwengier:
@davkean Can you have another pass at this some time.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IProjectSnapshotFactory.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.Build.Construction;
```
drewnoakes: Missing copyright banner.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.DesignTimeInputsDelta.cs
```diff
@@ -0,0 +1,24 @@
...
+        internal class DesignTimeInputsDelta
+        {
+            public bool ShouldCompile { get; set; } = true;
+            public ImmutableArray<string> AddedItems { get; set; } = ImmutableArray<string>.Empty;
+            public ImmutableArray<string> RemovedItems { get; set; } = ImmutableArray<string>.Empty;
+            public ImmutableArray<string> AddedSharedItems { get; set; } = ImmutableArray<string>.Empty;
+            public ImmutableArray<string> RemovedSharedItems { get; set; } = ImmutableArray<string>.Empty;
+            public IImmutableDictionary<NamedIdentity, IComparable> DataSourceVersions { get; set; } = ImmutableSortedDictionary<NamedIdentity, IComparable>.Empty;
+            public bool NamespaceChanged { get; internal set; }
+            public string OutputPath { get; internal set; }
```
drewnoakes: These setters are already internal. Why mark them in particular?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.DesignTimeInputsItem.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal partial class TempPEBuildManager
+    {
+        internal class DesignTimeInputsItem
+        {
+            public ImmutableHashSet<string> Inputs { get; set; } = ImmutableHashSet.Create(StringComparers.Paths);
+            public ImmutableHashSet<string> SharedInputs { get; set; } = ImmutableHashSet.Create(StringComparers.Paths);
+            public string OutputPath { get; internal set; }
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.DesignTimeInputsDelta.cs
davidwengier: That's the default behavior for the "Add Property" code fixer, must have forgotten to clean them up

davidwengier: Oops! I should have closed this review, #4740 is the current one. I'm breaking this up into pieces to make reviewing easier. The "ditto" comment is still valid in that branch though, so thanks :D

davidwengier:
Closing in favour of #4740

Issue #3548, opened 2018-05-16
Memory allocation in DependenciesSnapshotProvider.UpdateDependenciesSnapshot.GetProjectItemSpecs


lifengl:
This method calls into IsImported(), which will get the property where the item is imported.  This property is quite expensive, and msbuild will recalculate it with some regular expression work every time.  Because the rule snapshot has already captured that data in a built-in property, there is no reason to pay that price again in the dependencies tree.  Can this code switches to use the data captured in project rule results.

This contributes 1.4% of total allocation in Roslyn.sln loading time, but also contributes to the extra CPU overhead.

Name | Inc % | Inc
-- | -- | --
Microsoft.Build.Shared.FileUtilities+ItemSpecModifiers.GetItemSpecModifier (System.String, System.String, System.String, System.String, System.String ByRef) | 1.4 | 43,575,560

lifengl:
Yes.

Sent from my phone

On May 15, 2018, at 5:23 PM, David Kean <notifications@github.com<mailto:notifications@github.com>> wrote:


Oh DefiningProjectFullPath?

tmeschter:
As best I can tell the only place this information is used is in `ImplicitTopLevelDependenciesSnapshotFilter.BeforeAdd`.

tmeschter:
@lifengl @davkean What specific type and member already captures this information?

drewnoakes:
@lifeng I'm seeing this and would like to fix it.

On a basic .NET Core project we call `IsImported` 132 times on _every_ dependency update, of which there are many.

> Because the rule snapshot has already captured that data in a built-in property, there is no reason to pay that price again in the dependencies tree. Can this code switches to use the data captured in project rule results.

Can you please point me to the built-in property you're referring to?

The closest thing I've found after hunting around in the debugger is an `Microsoft.Build.Execution.ProjectInstance` within a `Microsoft.VisualStudio.ProjectSystem.Designers.ProjectSnapshot`. Does that sound about right?

drewnoakes:
I'm seeing calls to `IsImported` allocating 3.8% on a large solution:

![image](https://user-images.githubusercontent.com/350947/49252621-68d74480-f41c-11e8-853f-9b165ecf09ad.png)

@lifengl made it sound like this might be straightforward to fix, and I've had a look throught the project but I'm not sure which property to use instead.

davkean:
Can you show the allocations under that?

drewnoakes:
[...]

PR #4352, opened 2018-12-06
Locking, allocations, documentation and cleanup

drewnoakes:
While working through #3548 I found a few places we could tighten things up, which I've pulled out into this PR.

- 91560a37e15fa7d797002e124ef1dc72465aa3f8 remove locks made redundant by previous changes
- 21abe962d98829591d296a1213a7960dd3b5970e 144ac2a9d697378454bd6fcc5a8856fde1035bb0 7310b4052000f3a308bcf9632708559dc221cf68 d9c83c6aeacf6cb539326447904e78bff1d5d994 f92789ddb7175bf51d74ecdfc3a811e2815a99f9 reduce allocations
- 2cd92e1e95e03b7d088ebdb1a2a5aa2aa060c064 documentation

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ImmutableCollectionLinqExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ImmutableCollectionLinqExtensions.cs
@@ -39,6 +39,14 @@ namespace Microsoft.VisualStudio
             return false;
         }
 
+        public static IEnumerable<TResult> Select<TSource, TResult>(this ImmutableArray<TSource> immutableArray, Func<TSource, TResult> selector)
+        {
+            foreach (TSource obj in immutableArray)
+            {
+                yield return selector(obj);
+            }
+        }
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ImmutableCollectionLinqExtensions.cs
```diff
@@ -39,6 +39,14 @@ internal static class ImmutableCollectionLinqExtensions
             return false;
         }
 
+        public static IEnumerable<TResult> Select<TSource, TResult>(this ImmutableArray<TSource> immutableArray, Func<TSource, TResult> selector)
```
Pilchie: Doesn't this already exist?https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/src/System/Linq/ImmutableArrayExtensions.cs#L30

Did it not get added to the version that we use?

drewnoakes: You're right. I missed that.

I checked the other extensions we have and they're not duplicates.

PR #4361, opened 2018-12-10
Replace LanguageService2 -> LanguageService capability

davkean:
LanguageService2 was a temporary capability.


davkean:
Let me switch on the new one while at the same time disabling the old one.

Issue #2795, opened 2017-09-07
Remove concept of "short name" in TargetFramework


davkean:
TargetFramework has a short name concept - this has been fraught with bugs https://github.com/dotnet/project-system/issues/2713 and https://github.com/dotnet/project-system/issues/2388 because it's really tempting to compare it. Short name is not-canonical and should never be compared, it is solely a convenience used for developers writing project files.

We should remove concept of TargetFramework.ShortName.

drewnoakes:
Currently we use the short form in dependency ID strings. Using the long form will increase memory consumption.

Blocking this on #5051.

drewnoakes:
#5051 has been addressed, unblocking this.

`TargetFramework.FrameworkName` was removed in dfc437c6462b115a917075eafcd5aeaf985b541a, leaving only properties `FullName` and `ShortName`. When we remove `ShortName`, we may be able to remove `TargetFramework` altogether.

PR #4364, opened 2018-12-10
Remove lock from TargetFrameworkProvider

### LINKED ISSUE #2714: Large amount of lock contention in dependency node

drewnoakes:
Fixes #2714, which identified `TargetFrameworkProvider.GetTargetFramework` as a source of high contention, albeit under the influence of a since-fixed evaluation loop bug.

`GetTargetFramework` looks up `ITargetFramework` instances by name. It did this via O(N) scanning of a mutable list within a lock (where N is usually a very small number, but still required the protection of a lock).

This change introduces lock-free caching of these instances using O(1) lookup in an immutable dictionary keyed by the input string. Cache misses should be very rare. Hits are now the fast path and have zero contention.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/TargetFrameworkProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/TargetFrameworkProvider.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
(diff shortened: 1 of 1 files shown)


davkean:
It's hard to see if there is a behavior change here; unit tests would be helpful. :)

I've been meaning to remove comparisons to short names completely because they are not canonical which is tracked by https://github.com/dotnet/project-system/issues/2795.

drewnoakes:
> unit tests would be helpful. :)
>
> I've been meaning to remove comparisons to short names completely because they are not canonical which is tracked by #2795.

Hopefully this class along with `ITargetFramework` goes away before long, along with the cross targeting stuff. It's on my mind as I work through this stuff.

davkean:
What I foresee us having is something that maps dimension values -> some display name, and we have a target framework that we use in Debugging TFM selector, dependency node and the editor context/error list disambiguation.

PR #4365, opened 2018-12-10
Merge consecutive iteration loops

drewnoakes:
CPU profiling shows about 50% of `DependenciesSnapshot.FromChanges` time was spent in `DuplicatedDependenciesSnapshotFilter.BeforeAddOrUpdate`, mostly enumerating immutable dictionaries.

On loading `Roslyn.sln` <sup>1</sup> we spent 4% CPU time producing dependencies snapshots, half of which was in this method:

image

The normal case here is that no duplicate caption is found. This change halves the number of iterations required.

I'm looking at separating handling of top-level and non-top-level nodes across the dependencies node, at which point the number of items being iterated will drop considerably, along with that CPU percentage. In the meantime, this is an easy win.

<sup>1</sup> Based on a 30 second PerfView capture started when the "Preparing projects" dialog disappears.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/DuplicatedDependenciesSnapshotFilter.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/DuplicatedDependenciesSnapshotFilter.cs
@@ -30,35 +30,39 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.Fil
             IAddDependencyContext context)
         {
             IDependency matchingDependency = null;
-            foreach ((string _, IDependency x) in context)
-            {
-                if (x.TopLevel 
-                     && !x.Id.Equals(dependency.Id, StringComparison.OrdinalIgnoreCase)
-                     && StringComparers.DependencyProviderTypes.Equals(x.ProviderType, dependency.ProviderType)
-                     && x.Caption.Equals(dependency.Caption, StringComparison.OrdinalIgnoreCase))
-                {
-                    matchingDependency = x;
-                    break;
-                }
-            }
+            bool shouldApplyAlias = false;
 
-            // If found node with same caption, or if there were nodes with same caption but with Alias already applied
-            // NOTE: Performance sensitive, so avoid formatting the Caption with parenthesis if it's possible to avoid it.
-            bool shouldApplyAlias = matchingDependency != null;
-            if (!shouldApplyAlias)
+            foreach ((string _, IDependency other) in context)
             {
-                int adjustedLength = dependency.Caption.Length + " (".Length;
-                foreach ((string _, IDependency x) in context)
+                if (!other.TopLevel ||
+                    other.Id.Equals(dependency.Id, StringComparison.OrdinalIgnoreCase) ||
+                    !StringComparers.DependencyProviderTypes.Equals(other.ProviderType, dependency.ProviderType))
                 {
-                    if (x.TopLevel
-                         && !x.Id.Equals(dependency.Id, StringComparison.OrdinalIgnoreCase)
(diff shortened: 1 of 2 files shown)


davkean:
For the future, you should be able to scope to particular events to get more concrete numbers, for example, I'd consider; 

Start: Microsoft-VisualStudio-Common/vs_solution_perf_load/Stop
Stop: https://github.com/dotnet/project-system/blob/d9f0d1d40b81fe3f50317a758b3474426c69cb63/src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryEventName.cs#L35

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/DuplicatedDependenciesSnapshotFilter.cs
```diff
@@ -30,52 +30,59 @@ internal sealed class DuplicatedDependenciesSnapshotFilter : DependenciesSnapsho
...
-                    if (x.TopLevel
-                         && !x.Id.Equals(dependency.Id, StringComparison.OrdinalIgnoreCase)
-                         && StringComparers.DependencyProviderTypes.Equals(x.ProviderType, dependency.ProviderType)
-                         && x.Caption.StartsWith(dependency.Caption, StringComparison.OrdinalIgnoreCase)
-                         && x.Caption.Length >= adjustedLength
-                         && string.Compare(x.Caption, adjustedLength, x.OriginalItemSpec, 0, x.OriginalItemSpec.Length, StringComparison.OrdinalIgnoreCase) == 0)
+                    if (other.Caption.Length == dependency.Caption.Length)
                     {
+                        // Exact match.
+                        System.Diagnostics.Debug.Assert(matchingDependency == null, "matchingDependency should be null");
```
davkean: Assumes.Null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/DuplicatedDependenciesSnapshotFilter.cs
```diff
@@ -30,52 +30,59 @@ internal sealed class DuplicatedDependenciesSnapshotFilter : DependenciesSnapsho
...
+                        matchingDependency = other;
                         shouldApplyAlias = true;
                         break;
                     }
+
+                    // Prefix matches.
+                    // Check whether we have a match of form "Caption (ItemSpec)".
+
+                    string itemSpec = other.OriginalItemSpec;
+                    int expectedItemSpecIndex = dependency.Caption.Length + 2;
```
davkean: I found the previous way this was adding this more obvious; can you put this `// Account for " ("` or somethign after it?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/DuplicatedDependenciesSnapshotFilter.cs
drewnoakes: `Assert` is conditional so wouldn't be in release code.

It's impossible for this value to be anything other than `null`, given the `break`. I'll just remove the assert.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Filters/DuplicatedDependenciesSnapshotFilter.cs
drewnoakes: Comment added.

It'd be nice if `" (".Length` was reduced to a constant by the compiler.

PR #4366, opened 2018-12-11
Fix non english rules

davidwengier:
Fixes [#746371](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/746371)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -23,6 +23,14 @@
     <SuppressOutOfDateMessageOnBuild Condition="'$(SuppressOutOfDateMessageOnBuild)' == ''">true</SuppressOutOfDateMessageOnBuild>
   </PropertyGroup>
 
+  
+  <!--
+    Rule files that don't need localization go in the neutral directory to save duplicating files into each language
+  -->
+  <PropertyGroup Condition="'$(ManagedXamlNeutralResourcesDirectory)' == ''">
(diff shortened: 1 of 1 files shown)


davkean:
We don't need to duplicate these; add a concept to Microsoft.Managed.DesignTime.targets of non-localizable files.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -52,6 +52,11 @@
     <DebuggerFlavor Condition = "'$(DebuggerFlavor)' == ''">ProjectDebugger</DebuggerFlavor>
   </PropertyGroup>
 
+  <!-- For rule files that don't need localization we use the neutral directory to save duplicating files into each language -->
+  <PropertyGroup Condition="'$(ManagedXamlNeutralResourcesDirectory)' == ''">
```
davkean: I'm thinking that English above should fall back to `ManagedXamlNeutralResourcesDirectory`, so that to control this directory you only need to set one property.

davidwengier: Yeah, that makes sense. It might even make sense, in light of this change, to create an `\en\` subdirectory for the English files, just so that neutral can really mean "has no translations", rather than "has no translations or is English".

davkean: In VS, neutral almost always means English. In Windows, they do put English in its own directory.

davidwengier: In that case lets leave it for another day, since this is a P2 bug

PR #4367, opened 2018-12-11
Match legacy behaviour and don't restore designer on solution open

### LINKED ISSUE #1486: Find All Files causes project properties to be opened
I can't figure out exact repro steps for this, but it has happened to me many times with the SDK solution in VS:

- Place the cursor on a word you want to search for in the editor
- Hit CTRL+SHIFT+F to bring up the Find All Files dialog
- Press Enter to start the Find All operation

**EXPECTED:** Find all operation starts, and the focus moves to the Find Results window
**ACTUAL:** Find all operation starts, but a project properties window is opened and the focus moves to the newly opened project properties window

Once this happens for me, it does not appear to happen again if I repeat the same steps immediately.

davidwengier:
Fixes #1486

The current app designer code is littered with spots where it forcibly "needs" to show the designer in order to do its work, but that breaks when it is restored to a background tab during solution load, in particular when Find is used (see issue). Since this is a regression from legacy which never restored app designer tabs, this PR reinstates that behaviour. In future if the app designer code changes significantly this could potentially be removed.

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerWindowPane.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerWindowPane.vb
@@ -71,6 +71,9 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                 'Must marshal ourselves as IUnknown wrapper
                 VSErrorHandler.ThrowOnFailure(WindowFrame.SetProperty(__VSFPROPID.VSFPROPID_ViewHelper, New UnknownWrapper(_viewHelper)))
 
+                'Don't reopen the designer during solution restore
+                VSErrorHandler.ThrowOnFailure(WindowFrame.SetProperty(__VSFPROPID5.VSFPROPID_DontAutoOpen, True))
+
                 ' make sure scrollbars in the view are not themed
                 VSErrorHandler.ThrowOnFailure(WindowFrame.SetProperty(__VSFPROPID5.VSFPROPID_NativeScrollbarThemeMode, __VSNativeScrollbarThemeMode.NSTM_None))
             End If


davkean:
Can you explain why this would be ideal behavior?

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerWindowPane.vb
```diff
@@ -71,6 +71,9 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                 'Must marshal ourselves as IUnknown wrapper
                 VSErrorHandler.ThrowOnFailure(WindowFrame.SetProperty(__VSFPROPID.VSFPROPID_ViewHelper, New UnknownWrapper(_viewHelper)))
 
+                'Don't reopen the designer during solution restore
+                VSErrorHandler.ThrowOnFailure(WindowFrame.SetProperty(__VSFPROPID5.VSFPROPID_DontAutoOpen, True))
```
davidwengier: This works, but one question about it, presumably for @davkean: I was told that we share designers and editor factories with legacy for the app designer, so I don't understand how this regression came about. There is no reference to this property, or its equivalent RDT_DontAutoOpen property, in any of the legacy code, or pkgdef files, in the entire vsproject folder in the VS repo.

davkean: Clearly legacy is doing this a different way; I'd be worried about branch switches that reloaded the solution - does this mean that the designer will now close in this situation?

This fix (IMO) is worse than the "bug".

davkean: I just tested legacy; the AppDesigner is restored if it active, not if it is a background tab.

davidwengier:
> We've already shipped the Find bug.

Fair enough, I didn't think Preview 2 counted as shipped. I was trying to get this in before the dev cut off for that.

> What's preventing this?

I haven't dug in to the specifics, but there are at least 3 spots I found in the code where it does it (see https://github.com/dotnet/project-system/issues/1486#issuecomment-440460864) with comments around them implying its necessary, which makes me think it would be a longer job than possible to hit the Preview 2 cut off. Based on the above maybe that isn't relevant?

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerWindowPane.vb
davidwengier: > I'd be worried about branch switches that reloaded the solution

Indeed the design does close when doing branch switches, which is not ideal. Legacy has the same behaviour if there are changes to the project file across branches (and possibly other scenarios)

PR #4372, opened 2018-12-12
Fix both NeutralLanguageComboBox and AssemblyInfoPropPage closing on Escape

etbyrd:
Fixes: [646590](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646590/)

On the Assembly Information property page, if the Neutral Language ComboBox is dropped down the escape key will close both the ComboBox and the property page. This will preview the key pressed and not close the property page if escape was used to close the ComboBox.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoPropPage.vb
@@ -172,6 +172,27 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Common.SetComboBoxDropdownWidth(NeutralLanguageComboBox)
         End Sub
 
+        Private _neutralLanguageWasDroppedDown As Boolean = False
+
+        ''' <summary>
+        ''' For checking if the NeutralLanguageComboBox was DroppedDown when the escape key was pressed.
+        ''' With a combination of AutoCompleteMode.SuggestAppend and AutoCompleteSource.ListItems the 
+        ''' NeutralLanguageComboBox and the Assembly Info window will both handle a Escape key press and both close.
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoPropPage.vb
```diff
@@ -172,6 +172,27 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+        Private _neutralLanguageWasDroppedDown As Boolean = False
+
+        ''' <summary>
+        ''' For checking if the NeutralLanguageComboBox was DroppedDown when the escape key was pressed.
+        ''' With a combination of AutoCompleteMode.SuggestAppend and AutoCompleteSource.ListItems the 
+        ''' NeutralLanguageComboBox and the Assembly Info window will both handle a Escape key press and both close.
+        ''' We only want the NeutralLanguageComboBox to close in that case.
+        ''' </summary>
+        Private Sub NeutralLanguageComboBox_PreviewKeyDown(sender As Object, e As PreviewKeyDownEventArgs) Handles NeutralLanguageComboBox.PreviewKeyDown
+            If (e.KeyData And Keys.KeyCode) = Keys.Escape And NeutralLanguageComboBox.DroppedDown Then
```
Pilchie: I'm a bit surprised this is necessary, but I can repro it in a standalone app.  Have we filed a bug on Winforms to see if this is by design?

I would have thought that if the combo box had focus it would handle the escape before it got to the dialog.

etbyrd: I was very surprised as well. I will be filing a bug for this.

Issue #159, opened 2016-05-17
Related files support


davkean:
C# and VB have a registration API for automatically registering dependency relationship between files so that they are automatically parented underneath and hidden in some cases. 

For example, in VB, Foo.Designer.cs and Foo.resx is automatically grouped under Foo.cs, and both are hidden by default.

We should implement similar support - but I'd like to see an MSBuild equivalent of this - not via the registry as it is today:

![image](https://cloud.githubusercontent.com/assets/1103906/15332731/ab08370a-1c1b-11e6-97a8-936f423cc27b.png)

See: `CVbPackage::InitializeRelatedFiles` and `CCSharpPackage::InitializeRelatedFiles` internally.

We should make sure that we have sensible defaults that do not require DependantUpon metadata item to be written to the project file.

paul1956:
Today in VB if you add an existing form, the related don't always get copied and if you use the OS to copy all the files you have to restart Visual Studio for the nesting to happen. It would be nice if these were cleaned up as part of the change.

davkean:
:+1:

davkean:
Wonder if a hybrid approach of using project.json's heuristics + flags to hide nested types for the VB cases might be enough.

asbjornu:
I've experienced cases where related files like these aren't automatically grouped unless you hand edit the `.csproj` file and add some MSBuild voodoo. What and where, I don't remember. I just mention it as another argument for doing something about this grouping, which I agree is a nice feature. :smiley:

davkean:
Also see: https://visualstudio.uservoice.com/forums/121579-visual-studio-2015/suggestions/4280193-identify-and-nest-dependent-files-in-solution-expl

asbjornu:
[People seem to want this feature turned off](https://visualstudio.uservoice.com/forums/121579-visual-studio-2015/suggestions/10924407-option-to-disable-file-nesting-in-asp-net-5-mvc) too, so a flexible way to configure which files are nested by which pattern and the ability to turn this on and off on a per-file basis (without writing megabytes of XML) would be neat.

NickCraver:
I'm really interested in this - we make extensive use of it at Stack Overflow. For example, for a `User` object, we have `User.cs`, `User.Render.cs`, etc. Here's a view:

<img width="297" alt="screen shot 2016-12-15 at 20 28 34" src="https://cloud.githubusercontent.com/assets/454813/21248436/170424ba-c305-11e6-876f-4a06dab9c124.png">

[...]

PR #4376, opened 2018-12-13
Add project capability DynamicDependentFile

drewnoakes:
Now that WTE has inserted their opt-out logic, we can enable this.

//cc: @abpiskunov @jmarolf

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -33,6 +33,7 @@ namespace Microsoft.VisualStudio.Packaging
 
 
         public const string DefaultCapabilities = ProjectCapability.AppDesigner + "; " +
+                                                  ProjectCapability.DynamicDependentFile + "; " +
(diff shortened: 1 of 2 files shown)


drewnoakes:
Fixes #159.

abpiskunov:
👍

davkean:
Is it required that this is a fixed (vs dynamic) capability? ie Is it deliberate that this in our project registration vs targets file?

drewnoakes:
If this can be a dynamic dependency, do we need the CPS precedence order suppression logic for this? Can't we just add the capability dynamically, and WTE can turn it off? Can we make this all a bunch simpler?

lifengl:
We can handle it dynamically.  You can try to add/remove it in the project file.

Sent from my phone

On Dec 12, 2018, at 4:54 PM, Drew Noakes <notifications@github.com<mailto:notifications@github.com>> wrote:


If this can be a dynamic dependency, do we need the CPS precedence order suppression logic for this? Can't we just add the capability dynamically, and WTE can turn it off? Can we make this all a bunch simpler?

drewnoakes:
@lifeng thanks. Does this mean that if we add the capability then WTE can remove it again, and we end up in the same place without needing `SuppressLowerPriority` logic?

@abpiskunov would that work for you?

drewnoakes:
Closing in favour of #4378.

Issue #4375, opened 2018-12-13
Remove usages of banned APIs


drewnoakes:
#4374 introduces a banned API analyzer to help prevent future usages of certain APIs, as discussed in #3777.

It suppressed existing usages of banned APIs. Most of those should be replaced.

The suppressions were in 047b1f3f3b1bbf6b4f3862e4a443745fe408a9f2, and can be found by searching for `#pragma warning disable RS0030`.

PR #4379, opened 2018-12-13
remove useage of IProjectGuidService

jmarolf:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -132,20 +133,18 @@ public void ObserveTreeUpdateCompleted(bool hasUnresolvedDependency)
 
             bool ObservedAllRules() => _telemetryStates.All(state => state.Value.ObservedAllRules());
 
-            void InitializeProjectId()
+            async Task InitializeProjectIdAsync()
             {
-#pragma warning disable RS0030 // symbol IProjectGuidService is banned
-                IProjectGuidService projectGuidService =
-                    _project.Services.ExportProvider.GetExportedValueOrDefault<IProjectGuidService>();
+                ISafeProjectGuidService projectGuidService =
```
davkean: Do you mind moving this to just an import in the constructor?

jmarolf: no problem

davkean: It's not optional so you can delete the null path.

jmarolf: I'll add a check for whether the Guid is the default guid

davkean: This will never return a default GUID, that would be a bug.

davkean: In fact, that's why this new service exists - to make sure we're never in that situation.

jmarolf: https://github.com/dotnet/project-system/blob/bed629b18ba823e49daabfb32f21f777d29fd2db/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSafeProjectGuidService.cs#L39

jmarolf: Looks like I should check to see if `Guid.Empty` is returned then

drewnoakes:
Relates to #4375.

src/Microsoft.VisualStudio.ProjectSystem.Managed/GlobalSuppressions.cs
```diff
@@ -1,3 +1,5 @@
...
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "VSTHRD004:Await SwitchToMainThreadAsync", Justification = "https://github.com/Microsoft/vs-threading/issues/377", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.ThreadingExtensions.SwitchToUIThread(Microsoft.VisualStudio.ProjectSystem.IProjectThreadingService,System.Threading.CancellationToken)~Microsoft.VisualStudio.Threading.JoinableTaskFactory.MainThreadAwaitable")]
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Scope = "member", Target = "~M:Microsoft.VisualStudio.Threading.Tasks.CancellationSeries.CreateNext(System.Threading.CancellationToken)~System.Threading.CancellationToken")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0030:Do not used banned APIs", Justification = "<Pending>", Scope = "type", Target = "~T:Microsoft.VisualStudio.ProjectSystem.ISafeProjectGuidService")]
```
drewnoakes: Do we have guidance on when we use a suppression attribute vs. when we use `#pragma warning'? We have a mix of both through the code.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/GlobalSuppressions.cs
```diff
@@ -4,3 +4,5 @@
...
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "VSTHRD110:Observe result of async calls", Justification = "https://github.com/dotnet/project-system/issues/3921", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPage.SetObjects(System.UInt32,System.Object[])")]
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "VSTHRD110:Observe result of async calls", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.InitializeCoreAsync(System.Threading.CancellationToken)~System.Threading.Tasks.Task")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0030:Do not used banned APIs", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.VS.VsSafeProjectGuidService.GetProjectGuidAsync~System.Threading.Tasks.Task{System.Guid}")]
```
davkean: I don't understand why this was added?

src/Microsoft.VisualStudio.ProjectSystem.Managed/GlobalSuppressions.cs
davkean: I don't get why this was added, why is the rule firing?

drewnoakes: Both of these suppressions are replacements for `#pragma warning` items that've been removed from other files.

In both cases it's due to legitimate usages of the banned symbol. The 'safe' implementation of this interface needs to use the actual interface, so gets a pass.

I'm undecided on whether it's clearer to have the suppression here or alongside the code in question.

drewnoakes: > why is the rule firing

The rule's firing in `ISafeProjectGuidService.cs` due to the reference of `IProjectGuidService` in the XML docs.

davkean: I use suppression files when; 1 it's a bug in the rule, 2 I'm baselining. I'll usually inline the rest.

drewnoakes: I think that's a sensible approach. Under those guidelines these'd be inline.

jmarolf: My thought was basically:

- If the rule is likely to be re-enabled at some point on that code then use `#pragma` suppression so it's easy to search for these

- If a rule is being suppressed because we don't want to change / will never change the code and it's not a bug then I think moving it to a suppression file and not having `#pragma`'s clutter things up is nice

Don't have any strong feelings about this though.  If someone has a preference here I will revert it.

PR #4380, opened 2018-12-13
 Add an implemention of IActiveWorkspaceProjectContext for new language service

davkean:
- Add an implemention IActiveWorkspaceProjectContext for the new language service integration
  - This lets non-multi-targeting aware consumers to use the "active" configuration's version of the workspace host

- Ensure that we've initialized the language service before project load has completed

This is the last of the changes needed to make the language-service rewrite work.


davidwengier:
> last of the changes
🎉 Congratulations 🎉

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceContextHost.cs
```diff
@@ -75,7 +76,7 @@ public async Task<T> OpenContextForWriteAsync<T>(Func<IWorkspaceProjectContextAc
 
             WorkspaceContextHostInstance instance = await WaitForLoadedAsync();
 
-            // Throws OperationCanceledException if 'instance' is Disposed
+            // Throws ActiveProjectConfigurationChangedException if 'instance' is Disposed
```
drewnoakes: Would these be better as `<throws` docs on the member (or interface member)?

PR #4385, opened 2018-12-13
remove use of `IProjectLockService` in favor of `IProjectAccessor`

jmarolf:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.cs
```diff
@@ -2,52 +2,48 @@
...
             IProjectProperties defaultProperties,
             IReadOnlyDictionary<string, string> dimensionalConditions = null)
         {
-#pragma warning disable RS0030 // symbol IProjectLockService is banned
-            await _projectLockService.WriteLockAsync(async access =>
+            await _projectAccessor.OpenProjectXmlForWriteAsync(_unconfiguredProject, projectXml =>
             {
-                ProjectRootElement projectXml = await access.GetProjectXmlAsync(_unconfiguredProject.FullPath);
-                await _helper.SetPropertyAsync(unevaluatedPropertyValue, defaultProperties, projectXml);
+                _threadingService.ExecuteSynchronously(() => _helper.SetPropertyAsync(unevaluatedPropertyValue, defaultProperties, projectXml));
```
davkean: We should avoid blocking a thread pool thread for this, we should separate the async path from the reading/updating path, or look into deeper why this is async.

jmarolf: It comes from reading the project properties
https://github.com/dotnet/project-system/blob/81492404fee8a9e9a53f07ed68958836e3d1a0f0/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.AbstractBuildEventHelper.cs#L39

Would love to change this to be better. What should we do here instead?

drewnoakes: Was just thinking the same.

Also, what's the difference between `ExecuteSynchronously` and `JTF.Run`?

The instance above looks different. The return value from `ExecuteSynchronously` is returned into `OpenProjectXmlForReadAsync`.

jmarolf: Lets see. I can pull the async call to `IProjectProperties defaultProperties` up so that only synchronus code runs in inside the lock.  Let me try that

davidwengier: > Also, what's the difference between ExecuteSynchronously and JTF.Run?

[Nothing](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.Implementation%2FProjectServices%2FProjectMultiThreadedService.cs&version=GBmaster&line=144&lineStyle=plain&lineEnd=149&lineStartColumn=1&lineEndColumn=1)

jmarolf: alright, this should be fixed now

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/AbstractBuildEventValueProvider.AbstractBuildEventHelper.cs
```diff
@@ -33,27 +33,43 @@ public abstract class AbstractBuildEventHelper
             private string BuildEvent { get; }
             private string TargetName { get; }
 
-            public async Task<string> GetPropertyAsync(ProjectRootElement projectXml, IProjectProperties defaultProperties)
+            public async Task<(bool succes, string property)> TryGetPropertyAsync(IProjectProperties defaultProperties)
```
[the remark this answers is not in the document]
jmarolf: This class is internal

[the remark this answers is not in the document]
davidwengier: Oops, I was going on GitHub's little blue helper bar.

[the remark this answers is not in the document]
jmarolf: No problem.  This is actually why I use CodeFlow a lot, I need to see _all the code_

PR #4386, opened 2018-12-14
Small bits pulled from larger work

drewnoakes:
To keep the size of a future PR down a bit, here's a grab bag of small tidy ups.

The only functional change here is in e89b1f955024d2860a7c2b24c96554a8054c6aaf which fixes a memory leak in `AggregateDependenciesSnapshotProvider` in cases where the project path changes, as it will not be removed from its collection.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySubscriptionsHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySubscriptionsHost.cs
@@ -27,6 +27,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscription
     {
         public const string DependencySubscriptionsHostContract = "DependencySubscriptionsHostContract";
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySubscriptionsHost.cs
```diff
@@ -101,21 +105,11 @@ internal sealed class DependencySubscriptionsHost : OnceInitializedOnceDisposedA
                 _dependenciesUpdateThrottleInterval,
                 commonServices.ThreadingService,
                 tasksService.UnloadCancellationToken);
-
-            ProjectFilePath = commonServices.Project.FullPath;
```
davkean: This would be out of sync with rename, so good change.

drewnoakes: That's the motivation for the change. It resulted in a leak in `AggregateDependenciesSnapshotProvider` which keyed a dictionary on project path. There may have been other consequences I didn't see too.

On a local branch I remove the `ProjectFilePath` property from `DependencySubscriptionHost` completely, and source it directly from the snapshot, so we have a single source of truth.

PR #4390, opened 2018-12-14
Ban ConfigureAwait method

### LINKED ISSUE #4389: Ban use of ConfigureAwait
#4374 introduced the banned symbol analyzer.

We should add `ConfigureAwait` to the list of banned symbols.

The diagnostic message should capture that:

- `ConfigureAwait(true)` should be removed
- `ConfigureAwait(false)` should be replaced with `await TaskScheduler.Default`

drewnoakes:
Fixes #4389

There were no usages, so nothing else to tidy up.

Searching for the string "ConfigureAwait" turned up two comments, both of which are pretty much redundant, so removed.

### DIFF
--- a/src/BannedSymbols.txt
+++ b/src/BannedSymbols.txt
@@ -9,3 +9,5 @@ M:Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProject(Mi
 
 T:Microsoft.VisualStudio.ProjectSystem.ProjectAutoLoadAttribute; Using IProjectDynamicLoadComponent avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.
 T:Microsoft.VisualStudio.ProjectSystem.ConfiguredProjectAutoLoadAttribute; Using IProjectDynamicLoadComponent avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.
+
+M:System.Threading.Tasks.Task.ConfigureAwait(System.Boolean); "ConfigureAwait(true)" should be removed, and "ConfigureAwait(false)" should be replaced with "await TashScheduler.Default"
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/PasteOrdering.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/PasteOrdering.cs
@@ -101,7 +101,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
         {
             Assumes.NotNull(_dropTarget);
 
-            // ConfigureAwait is true because we need to come back for PasteItemsAsync to work. If not, PasteItemsAsync will throw.
             ImmutableHashSet<string> previousIncludes = await OrderingHelper.GetAllEvaluatedIncludes(_configuredProject, _accessor);
             PasteItemsResult result = await PasteHandler.PasteItemsAsync(items, effect);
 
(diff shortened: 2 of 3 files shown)


src/BannedSymbols.txt
```diff
@@ -9,3 +9,5 @@ M:Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProject(Mi
 
 T:Microsoft.VisualStudio.ProjectSystem.ProjectAutoLoadAttribute; Using IProjectDynamicLoadComponent avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.
 T:Microsoft.VisualStudio.ProjectSystem.ConfiguredProjectAutoLoadAttribute; Using IProjectDynamicLoadComponent avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.
+
+M:System.Threading.Tasks.Task.ConfigureAwait(System.Boolean); "ConfigureAwait(true)" should be removed, and "ConfigureAwait(false)" should be replaced with "await TashScheduler.Default"
```
sharwell: 💡 

```suggestion
M:System.Threading.Tasks.Task.ConfigureAwait(System.Boolean); "ConfigureAwait(true)" should be removed, and "ConfigureAwait(false)" should be replaced with "await TaskScheduler.Default"
```

sharwell: 💡 If you reference ValueTask, it might make sense to add that one as well

sharwell: ❔  Should this also include `Task<TResult>.ConfigureAwait`?

drewnoakes: I believe so. Even though it's a subclass, it returns `Task<T>` so must be hiding the base.

davkean:
I'd consider adding `JoinableTask`/`JoinableTask<T>`

src/BannedSymbols.txt
drewnoakes: This highlighted a bug in the analyzer around generic types/methods. Fixed in dotnet/roslyn-analyzers#1927.

drewnoakes:
> I'd consider adding `JoinableTask`/`JoinableTask<T>`

They don't have `ConfigureAwait` and I can't find an extension. Is that what you were referring to?

davkean:
@drewnoakes My bad, I thought they did, but I'm mixing it with `JoinableTask.Task.ConfigureAwait` which will be captured by above.

drewnoakes:
@sharwell other task types are covered now. Thanks.

The analyzer version we're using has a bug around open/closed generic symbol comparisons (fixed in dotnet/roslyn-analyzers#1927). Once that's merged and built, I'll make another PR here with the updated version. I want to get dotnet/roslyn-analyzers#1925 in too, but need to find a repro so I can add a test.

Will merge if CI gives a green light.

src/BannedSymbols.txt
```diff
@@ -9,3 +9,8 @@ M:Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProject(Mi
 
 T:Microsoft.VisualStudio.ProjectSystem.ProjectAutoLoadAttribute; Using IProjectDynamicLoadComponent avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.
 T:Microsoft.VisualStudio.ProjectSystem.ConfiguredProjectAutoLoadAttribute; Using IProjectDynamicLoadComponent avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.
+
+M:System.Threading.Tasks.Task.ConfigureAwait(System.Boolean); "ConfigureAwait(true)" should be removed, and "ConfigureAwait(false)" should be replaced with "await TaskScheduler.Default"
```
davidwengier: This is most likely a nit, but ConfigureAwait(false) should be _preceeded_ by a call to `await TaskScheduler.Default`. 

ie 

```
await SomeThing().ConfigureAwait(false);
```
should become
```
await TaskScheduler.Default;
await SomeThing();
```

drewnoakes: Hmm, yeah, I don't think we can fit the explanation of how to do this properly. I'll back out the last commit. If someone knows how to fit that into the one-liner, we can improve it.

PR #4393, opened 2018-12-14
Import props/targets from NuGet in WPF tmpproj builds

Pilchie:
Thanks @tmat for the inspiration.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,5 +1,29 @@
 <!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
 <Project>
+  <!--
+    Determine whether the project is WPF temp project.
+    Since .NET Framework 4.7.2 WPF temp project name ends with _wpftmp suffix and keeps the language specific extension (e.g. csproj). 
+  -->
+  <PropertyGroup>
+    <IsWpfTempProject>false</IsWpfTempProject>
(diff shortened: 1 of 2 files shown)


Pilchie:
Confirmed with a binlog that this seems to work

davkean:
@Pilchie Does affect PackageReference in legacy as well? We should get a bug to fix this also in MSBuild/SDK.

Pilchie:
Yes, and there is already bug on MSBuild to fix it.

Pilchie:
This seems to not be working on the build machines after all.  I'll have to investigate further :(

davkean:
Didn't the project name change between .NET versions just recently?

nguerrera:
> Didn't the project name change between .NET versions just recently?

Yes

Pilchie:
Ah, maybe the build machines don't have a new enough .NET Framework?  This was a change in the runtime, and not the targeting pack, right?

nguerrera:
> This was a change in the runtime, and not the targeting pack, right?

Correct.

Pilchie:
Maybe the issue is that the build machines in CI aren't running `net472`?  If so, that sounds like a problem waiting to happen, since we've updated this project to target `net472`, so tests could fail at any time in the future.  @jmarolf do you know if we do/can require `net472` for CI builds?

Pilchie:
Nevermind - seems problem is our old version of repo toolset.  `_IsWpfTempProject` is `true`, but `_WpfTempProjectNuGetFileNameNoExt` isn't getting the right path to the artifacts directory.

PR #4394, opened 2018-12-14
Start using nullable

Pilchie:
@jcouv @cston

### DIFF
--- a/build/Test.ruleset
+++ b/build/Test.ruleset
@@ -37,5 +37,10 @@
   <Rules AnalyzerId="xunit.analyzers" RuleNamespace="xunit.analyzers">
     <Rule Id="xUnit1026" Action="None" /> <!-- Theory methods must use all parameters -->
     <Rule Id="xUnit1004" Action="None" /> <!-- Test methods should not be skipped -->
+
+    <!-- Disabled these two rules until we pick up a compiler with the fix for workaround https://github.com/dotnet/roslyn/issues/31740. -->
+    <Rule Id="xUnit1009" Action="None" /> <!-- InlineData values must match the number of method parameters -->
+    <Rule Id="xUnit1010" Action="None" /> <!-- The value is not convertible to the method parameter '{0}' of type '{1}'. -->
+    <Rule Id="xUnit1019" Action="None" /> <!-- MemberData must reference a data type assignable to '{0}'. The reference type '{1}' is not valid. -->
   </Rules>
 </RuleSet>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -23,8 +23,9 @@
 
   <PropertyGroup Condition="'$(Language)' == 'C#'">
     <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <Features>strict</Features>
+    <NullableReferenceTypes>true</NullableReferenceTypes>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Language)' == 'VB'">
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -26,6 +26,7 @@
     <LangVersion>8.0</LangVersion>
     <Features>strict</Features>
     <NullableReferenceTypes>true</NullableReferenceTypes>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);8600;8601;8602;8603;8604;8618;8619;8620;8622;8625;8626;8629</WarningsNotAsErrors>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Language)' == 'VB'">
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
@@ -363,7 +363,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                 }
             }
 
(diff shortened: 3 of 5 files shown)


build/Test.ruleset
```diff
@@ -37,5 +37,10 @@
   <Rules AnalyzerId="xunit.analyzers" RuleNamespace="xunit.analyzers">
     <Rule Id="xUnit1026" Action="None" /> <!-- Theory methods must use all parameters -->
     <Rule Id="xUnit1004" Action="None" /> <!-- Test methods should not be skipped -->
+
+    <!-- Disabled these two rules until we pick up a compiler with the fix for orkaround https://github.com/dotnet/roslyn/issues/31740. -->
```
jcouv: >orkaround [](start = 79, length = 9)

typo: workaround

Did you understand why those XUnit diagnostics fire?

Pilchie: I believe it was the same issue as attribute array nullability comparisons.

src/Directory.Build.props
```diff
@@ -47,8 +47,10 @@
 
   <PropertyGroup Condition="'$(Language)' == 'C#'">
     <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <Features>strict</Features>
+    <NullableReferenceTypes>true</NullableReferenceTypes>
+    <WarningsNotAsErrors>CS8625</WarningsNotAsErrors>
```
jcouv: >CS8625 [](start = 25, length = 6)

Why tune down this diagnostic?
The diagnostic that we're expecting people to tune down or disable is CS8600.

Pilchie:
Wait till you see the warning count in the CI Builds though 😛

build/Test.ruleset
jcouv: Since I was not able to repro the XUnit issue, I'm not sure the compiler fix fixes it.

src/Directory.Build.props
Pilchie: I think this might be another issue.  Our CI normally builds with `/warnaserror` but this seems to be the only warning that gets promoted?

build/Test.ruleset
Pilchie: This is a repro if you want to enlist ;-)

src/Directory.Build.props
Pilchie: N/M - the others are too.

Pilchie:
@agocke - do you know how to keep these warnings as warnings?

davkean:
Before we merge can we just all take a stab at fixing them all? I can take a look at this tomorrow and add this this PR.

agocke:
I believe the format for compiler warnings is just "<number>" not "CS<number>"

src/Directory.Build.props
```diff
@@ -47,8 +47,10 @@
 
   <PropertyGroup Condition="'$(Language)' == 'C#'">
     <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <Features>strict</Features>
+    <NullableReferenceTypes>true</NullableReferenceTypes>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);CS8600;CS8601;CS8602;CS8603;CS8618;CS8619;CS8625</WarningsNotAsErrors>
```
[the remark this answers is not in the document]
Pilchie: That doesn't work either :(

jmarolf:
Looks Like there are a lot of violations

```
2525 Error(s)
```

instead of turning this on for the entire solution all at once can't we use `#nullable enable` on a per-file basis Or enable
```xml
<NullableReferenceTypes>true</NullableReferenceTypes>
```

Per project?

Pilchie:
@jmarolf absolutely - this is more just a proof of concept to show that it can be enabled.  I don't think we should throw the switch like this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
```diff
@@ -363,7 +363,7 @@ public void SetObjects(uint cObjects, object[] ppunk)
                 }
             }
 
-            OnSetObjects(false);
+            WaitForAsync(() => OnSetObjects(false));
```
Pilchie: Interestingly, this triggers a VSTHRD010 violation that doesn't fire without the switch to `/langver:8.0` and `<NullableReferenceTypes>true</NullableReferenceTypes>`.  Possible change in API behavior?

Pilchie:
Also, 27 unit tests seem to fail on this branch that don't on `master`, which seems like a potential compat issue?

Pilchie:
> Also, 27 unit tests seem to fail on this branch that don't on master, which seems like a potential compat issue?

Seems related to the attribute argument thing where I disabled the xunit analyzers, as it looks like the attributes aren't persisted properly.

build/Test.ruleset
Pilchie: @jcouv: FYI - Updated to a compiler with the fix for 31740, and this still repros. Also seems to cause a code-gen change leading to test failures when these errors are suppressed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
sharwell: ❔ Was this intended to be fire-and-forget? Because this now changed the behavior to block on completion.

Pilchie:
Filed https://github.com/dotnet/roslyn/issues/31872 to explicitly track the xunit/theory issue.

davkean:
Going to close this as we've decided to take a different approach.

Issue #3777, opened 2018-07-24
Add analyzers to catch incorrect usage within tree


davkean:
I see the same bugs over and over again. We should write tests or analyzers to catch them so that new devs don't run into them:

- Use IProjectAccessor over IProjectLockService. This avoids skipping:
  - Checkout from source control on write 
  - Call to ReleaseAsync to avoid blocking a thread pool thread on release of the lock

- Use/centralized usage of MSBuild construction model. This makes sure that we:
  - Escape/Unescape values
  - Implement the same behavior of walking, looking for properties, items, etc.

- Use ISafeProjectGuidService vs IProjectGuidService, IProjectGuidService2. This avoids:
  - Reading the GUID before we've been set it in the cases where the solution gives us a new one

- Use IUnconfiguredProjectTasksService.ProjectLoadInHost vs IProjectAsyncLoadDashboard.ProjectLoadInHost. This handles:
  - The situation where the project is closed before it has finished loading, instead of waiting forever.

- Use IUnconfiguredProjectTasksService.LoadedProjectAsync vs CommonProjectSystemTools.LoadedProject<T>.
  - This interface is unit testable.

- Use IProjectDynamicLoadComponent vs [ProjectAutoLoad] and [ConfiguredProjectAutoLoad]. This avoids accidently reloading the project while loading, switching branches, installing NuGet package, etc.

- Reading TargetFramework/TargetFrameworks properties. Only the configuration service should read these properties. Everyone else should be looking at "isVariantDimension" dimensions, or TargetFrameworkMoniker - and not know anything about these properties.

Pilchie:
I can try to work on this a little bit.

davkean:
What always prevented me from doing this, is there a way to consume this from the same solution that you build them in?

davkean:
What might be interesting, is a way to mark an API as obsolete that you don't have access to - that way, you right the rule once and we just obsolete all the replement APIs with the new ones.

Pilchie:
Not as P2P refs, no. You’d have to package them and refer to an LKG version.

Pilchie:
Have some of these set up in [this branch](https://github.com/Pilchie/project-system/tree/UseBannedApiAnalyzer).  Results below.
 
Need more specifics about the MSBuild construction model, and the TargetFramework/TargetFrameworks properties.  Also need to flesh out the analyzer to allow finding members, (and the places it doesn't catch yet).

```
Severity	Code	Description	Project	File	Line	Suppression State
[...]

PR #4395, opened 2018-12-15
WIP Use IProjectDynamicLoadComponent over attributed load methods

drewnoakes:
Use `IProjectDynamicLoadComponent` over `ProjectAutoLoad` and `ConfiguredProjectAutoLoad` which were banned in #4374.

Relates to #3777, #4375.

## THIS IS CURRENTLY COMPLETELY BROKEN

These changes were made mechanically given my interpretation of advice in #3777. 

The removed attributes make specifications that `IProjectDynamicLoadComponent` does not allow for! Distinct examples from the base code:

```c#
[ProjectAutoLoad(startAfter: ProjectLoadCheckpoint.ProjectInitialCapabilitiesEstablished)]
[ProjectAutoLoad(startAfter: ProjectLoadCheckpoint.ProjectFactoryCompleted)]
[ProjectAutoLoad(completeBy: ProjectLoadCheckpoint.ProjectFactoryCompleted)]
[ConfiguredProjectAutoLoad(RequiresUIThread = false)]
```

Loading `ProjectSystem.sln` in d16.0stg under the debugger and the dependencies node doesn't populate. Probably more stuff is broken.

Leaving this here to start a discussion about what's needed if we do indeed wish to ban `ProjectAutoLoad` and `ConfiguredProjectAutoLoad`.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ActiveConfiguredProjectsLoaderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ActiveConfiguredProjectsLoaderTests.cs
@@ -25,7 +25,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
             });
 
             var loader = CreateInstance(project, out ProjectValueDataSource<IConfigurationGroup<ProjectConfiguration>> source);
-            await loader.InitializeAsync();
+            await loader.LoadAsync();
 
             // Change the active configurations
             await source.SendAndCompleteAsync(configurationGroups, loader.TargetBlock);
@@ -47,7 +47,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
             });
 
             var loader = CreateInstance(project, tasksService, out ProjectValueDataSource<IConfigurationGroup<ProjectConfiguration>> source);
-            await loader.InitializeAsync();
+            await loader.LoadAsync();
 
             var configurationGroups = IConfigurationGroupFactory.CreateFromConfigurationNames("Debug|AnyCPU");
 
@@ -71,8 +71,8 @@ namespace Microsoft.VisualStudio.ProjectSystem
 
             var loader = CreateInstance(project, out var source);
 
-            await loader.InitializeAsync();
-            await loader.InitializeAsync();
+            await loader.LoadAsync();
+            await loader.LoadAsync();
 
             var configurationGroups = IConfigurationGroupFactory.CreateFromConfigurationNames("Debug|AnyCPU");
 
@@ -106,7 +106,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
             });
 
             var loader = CreateInstance(project, out ProjectValueDataSource<IConfigurationGroup<ProjectConfiguration>> source);
-            await loader.InitializeAsync();
+            await loader.LoadAsync();
             loader.Dispose();
 
             var configurationGroups = IConfigurationGroupFactory.CreateFromConfigurationNames("Debug|AnyCPU");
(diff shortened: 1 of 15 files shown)


davidwengier:
There is no magic that picks up classes that implement IProjectDynamicLoadComponent, but there is magic that picks up classes that `[Export(typeof(IProjectDynamicLoadComponent)]`. Or to be more correct, one of:

```
[Export(ExportContractNames.Scopes.UnconfiguredProject, typeof(IProjectDynamicLoadComponent))]
```
or 
```
[Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
```

I don't know about the `startAfter` or `completedBy` stuff, the above is just from reading code not actual use of the component.

davkean:
Need some background, this should be "remove uses of [ProjectAutoLoad] and [ConfiguredProjectAutoLoad] with a *".

The * is, if it uses a "dynamic" capability. A dynamic capability is one that changes over the lifetime of a project;

- A targets file carries the capability via `<ProjectCapability Include="Foo"/>`
- A package is installed with the capability
- It comes and goes dynamically from other sources

A fixed capability is fixed for the lifetime of a project. It cannot be changed without reloading the project. Fixed capabilities tend to come from project type registration (in our package).

Why is a "targets file" considered dynamic? Because the outer UnconfiguredProject "inherits" capabilities from any active configuration, targets are only ever evaluated inside a configuration. There's a time early in a project's lifetime before `ProjectLoadCheckpoint.ProjectInitialCapabilitiesEstablished` where there is no active configuration, or the active configuration's capabilities haven't been applied to the UnconfiguredProject. Also, the active configuration can change causing a new set of capabilities to be applied to the UnconfiguredProject.

This presents a problem; [ProjectAutoLoad] says "load me automatically at the stage I've indicated" - but that can before the set of capabilities you are "applied" to have even been determined. You've said "I need to be loaded before we've determined the active configuration", but your capability doesn't appear until after that. Another problem is, what if we've already loaded you, but capability disappeared when someone changed the active configuration?

To handle these situations, we've decided that changes to capabilities that are applied to [ProjectAutoLoad] components if we've already loaded them, or the stage has passed where we should have loaded them, will cause us to automatically reload the project. We did this for compatibility mostly. We did also consider disposing autoload components if their capability disappeared, but it was very likely that _their_ dependencies probably weren't prepared for that.

To handle these dynamic capabilities that come and go without requiring us to reload the project, we introduced a new concept `IProjectDynamicLoadComponent`; LoadAsync gets called when the capability first appears, UnloadAsync gets called when the capability disappears. Most of our newish components opt into this. This isn't a direct replacement for [ProjectAutoLoad] or [ConfiguredProjectAutoLoad] because components will get loaded _later_ (typically around `ProjectInitialCapabilitiesEstablished`), but is a requirement if you are a dynamic capabilit

There are still components that use [ProjectAutoLoad]/[ConfiguredProjectAutoLoad] but these are components that contained "fixed" capabilities, such as ".NET" or "LanguageService". There might one or two old components that have a dynamic capability, but are still using these attributes - they are a timing bug waiting to happen, and will likely result in us reloading the project as it changes.

davkean:
Each one of these requires careful review and play around with. None of them actually suffer from the "dynamic problem" except for the VisualBasic one for shared projects. VisualBasic is dynamic for shared projects cos history.

I think it's all goodness that we move them all to IProjectDynamicLoadComponent, but I'll need to very that we don't regress anything.

UnloadAsync isn't a no-op - these features need to turn off when that is called, hence why we use AbstractMultiLifetimeComponent. You are missing an `[Export(typeof(IProjectDynamicLoadComponent))]` for all of them.

drewnoakes:
@davkean thanks very much for the detailed explanation here. It filled in a lot of gaps in my understanding and showed me a few interesting places in the CPS code base to dig through.

The project load checkpoint aspect implies there's no simple mechanical way to apply this change. Each needs consideration individually, and the unload operations need implementing.

Questions:

- Is a `IProjectDynamicLoadComponent` implementation constructed and disposed with the project? I assume that's the case.
- Is there an easy way to determine if a capability is fixed or dynamic? If it's on `[ProjectTypeRegistration]` is't obviously fixed, but beyond that I'm not completely sure.

I'll close this PR for now and will make some more specific comment's on #4401.

Issue #4377, opened 2018-12-13
Enable some disabled analyzers


drewnoakes:
#4374 upgraded the Roslyn Analyzers we import.

There are some new diagnostics in `Microsoft.NetFramework.Analyzers` that are worth looking at. In that PR I disabled them:

- [x] CA2153: Do not catch CorruptedStateExceptions 
- [x] CA2235: Mark all non-serializable fields (PR #4402)

Another in `Microsoft.CodeQuality.Analyzers`:

- [x] CA1820: Test for empty strings using string length

PR #4402, opened 2018-12-17
Enable CA2235 (non serializable fields on serializable types)

drewnoakes:
Relates to #4377

~These changes fix the errors, but looking at these I wonder whether the types wouldn't failed serialisation anyway, and if it wouldn't be better to just make them non-serialisable.~

### DIFF
--- a/build/Default.ruleset
+++ b/build/Default.ruleset
@@ -65,7 +65,7 @@
   </Rules>
   <Rules AnalyzerId="Microsoft.NetFramework.Analyzers" RuleNamespace="Microsoft.NetFramework.Analyzers">
     <Rule Id="CA2153" Action="None" /> <!-- TODO do not catch CorruptedStateExceptions -->
-    <Rule Id="CA2235" Action="None" /> <!-- TODO mark all non-serializable fields -->
+    <Rule Id="CA2235" Action="Error" />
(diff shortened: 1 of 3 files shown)


davkean:
This is just noise as it just does it own serialization; implements GetObjectData.

Pilchie:
So, potentially use a local suppressmessage attribute on each of those types, instead of doing all the fields?

drewnoakes:
I believe the diagnostic to be incorrect, so opened:

- https://github.com/dotnet/roslyn-analyzers/issues/1970
- https://github.com/dotnet/roslyn/issues/32133 (related, also to #4415)

Will revisit this PR soon to get it cleaned up.

drewnoakes:
If https://github.com/dotnet/roslyn-analyzers/pull/1971 is merged then we can ignore suppression altogether.

drewnoakes:
It's merged, but the build's borked. 

https://devdiv.visualstudio.com/DevDiv/_build?definitionId=9244&_a=summary

drewnoakes:
The build recovered. I've updated the analyzer package to include dotnet/roslyn-analyzers#1971 which fixes the false positives.

The one remaining diagnostic was valid. Please review the type for which I removed `SerializableAttribute`.

jmarolf:
Looks good 👍

PR #4404, opened 2018-12-17
Start annotating project-system for nullability.

Pilchie:
e7bb224 annotates everything except the generated code for Microsoft.VisualStudio.ProjectSystem.Managed.dll

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
@@ -81,8 +81,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         private readonly IVsService<SVsAppId, IVsAppId> _vsAppIdService;
         private readonly IVsService<SVsShell, IVsShell> _vsShellService;
 
-        private RemoteCacheFile _versionDataCacheFile;
-        private Version _ourVSVersion;
+        private RemoteCacheFile? _versionDataCacheFile;
+        private Version? _ourVSVersion;
 
         private uint _solutionCookie = VSConstants.VSCOOKIE_NIL;
         private bool _solutionOpened;
@@ -90,15 +90,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         // Tracks how often we meed to look for new data
         private DateTime _timeCurVersionDataLastUpdatedUtc = DateTime.MinValue;
-        private VersionCompatibilityData _curVersionCompatibilityData;
-        private IVsSolution _vsSolution;
+        private VersionCompatibilityData? _curVersionCompatibilityData;
+        private IVsSolution? _vsSolution;
 
         public async Task InitializeAsync()
         {
             await _threadHandling.Value.SwitchToUIThread();
 
             // Initialize our cache file
-            string appDataFolder = await _shellUtilitiesHelper.Value.GetLocalAppDataFolderAsync(_vsShellService);
+            string? appDataFolder = await _shellUtilitiesHelper.Value.GetLocalAppDataFolderAsync(_vsShellService);
             if (appDataFolder != null)
             {
                 _versionDataCacheFile = new RemoteCacheFile(Path.Combine(appDataFolder, VersionDataFilename), VersionCompatibilityDownloadFwlink,
@@ -376,7 +376,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             try
             {
                 // Try the cache file
-                Dictionary<Version, VersionCompatibilityData> versionCompatData = GetCompabilityDataFromCacheFile();
+                Dictionary<Version, VersionCompatibilityData>? versionCompatData = GetCompabilityDataFromCacheFile();
 
                 // See if the cache file needs refreshing and if so, kick off a task to do so
                 if (_versionDataCacheFile != null && _versionDataCacheFile.CacheFileIsStale())
@@ -390,8 +390,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
                 if (versionCompatData != null)
                 {
+                    // This is never null after InitializeAsync (which is called during package load)
+                    Version ourVSVersion = _ourVSVersion!;
+
                     // First try to match exactly on our VS version and if that fails, match on just major, minor
-                    if (versionCompatData.TryGetValue(_ourVSVersion, out VersionCompatibilityData compatData) || versionCompatData.TryGetValue(new Version(_ourVSVersion.Major, _ourVSVersion.Minor), out compatData))
+                    if (versionCompatData.TryGetValue(ourVSVersion, out VersionCompatibilityData compatData) || versionCompatData.TryGetValue(new Version(ourVSVersion.Major, ourVSVersion.Minor), out compatData))
                     {
                         // Now fix up missing data
                         if (string.IsNullOrEmpty(compatData.OpenSupportedMessage))
@@ -423,7 +426,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
                 });
             }
 
-            return _curVersionCompatibilityData;
+            return _curVersionCompatibilityData!;
         }
 
         private void UpdateInMemoryCachedData(VersionCompatibilityData newData)
@@ -435,11 +438,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         /// <summary>
         /// If the cached file exists reads the data and returns it
         /// </summary>
-        private Dictionary<Version, VersionCompatibilityData> GetCompabilityDataFromCacheFile()
+        private Dictionary<Version, VersionCompatibilityData>? GetCompabilityDataFromCacheFile()
         {
             try
             {
-                string data = _versionDataCacheFile?.ReadCacheFile();
+                string? data = _versionDataCacheFile?.ReadCacheFile();
                 if (data != null)
                 {
                     return VersionCompatibilityData.DeserializeVersionData(data);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
@@ -16,7 +16,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
         private readonly IProjectThreadingService _threadingService;
         private readonly IVsService<SVsSolutionBuildManager, IVsSolutionBuildManager2> _vsSolutionBuildManagerService;
         private readonly GeneratePackageOnBuildPropertyProvider _generatePackageOnBuildPropertyProvider;
-        private IVsSolutionBuildManager2 _buildManager;
+        private IVsSolutionBuildManager2? _buildManager;
         private uint _solutionEventsCookie;
 
         protected AbstractGenerateNuGetPackageCommand(
@@ -58,7 +58,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
             // Ensure build manager is initialized.
             await EnsureBuildManagerInitializedAsync();
 
-            ErrorHandler.ThrowOnFailure(_buildManager.QueryBuildManagerBusy(out int busy));
+            ErrorHandler.ThrowOnFailure(_buildManager!.QueryBuildManagerBusy(out int busy));
             return busy == 0;
         }
 
@@ -90,14 +90,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
 
                 // Save documents before build.
                 var projectVsHierarchy = (IVsHierarchy)Project.Services.HostObject;
-                ErrorHandler.ThrowOnFailure(_buildManager.SaveDocumentsBeforeBuild(projectVsHierarchy, (uint)VSConstants.VSITEMID.Root, 0 /*docCookie*/));
+                ErrorHandler.ThrowOnFailure(_buildManager!.SaveDocumentsBeforeBuild(projectVsHierarchy, (uint)VSConstants.VSITEMID.Root, 0 /*docCookie*/));
 
                 // Enable generating package on build ("GeneratePackageOnBuild") for all projects being built.
                 _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(true);
 
                 // Kick off the build.
                 uint dwFlags = (uint)(VSSOLNBUILDUPDATEFLAGS.SBF_SUPPRESS_SAVEBEFOREBUILD_QUERY | VSSOLNBUILDUPDATEFLAGS.SBF_OPERATION_BUILD);
-                ErrorHandler.ThrowOnFailure(_buildManager.StartSimpleUpdateProjectConfiguration(projectVsHierarchy, null, null, dwFlags, 0, 0));
+                ErrorHandler.ThrowOnFailure(_buildManager!.StartSimpleUpdateProjectConfiguration(projectVsHierarchy, null, null, dwFlags, 0, 0));
             }
 
             return true;
@@ -148,9 +148,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
 
                         if (_buildManager != null)
                         {
+#pragma warning disable CS8602 // Workaround https://github.com/dotnet/roslyn/issues/31620
                             // Unregister solution build events.
                             _buildManager.UnadviseUpdateSolutionEvents(_solutionEventsCookie);
                             _buildManager = null;
+#pragma warning restore CS8602
                         }
                     });
                 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
@@ -10,6 +10,7 @@ using System.Windows.Forms;
 using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
 {
@@ -362,7 +363,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                 }
             }
 
-            WaitForAsync(() => OnSetObjects(false));
+            OnSetObjects(false).Forget();
         }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
@@ -362,7 +362,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                 }
             }
 
-            OnSetObjects(false);
+            WaitForAsync(() => OnSetObjects(false));
         }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
@@ -18,10 +18,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
        IPropertyPage,
        IVsDebuggerEvents
     {
-        private IPropertyPageSite _site = null;
+        private IPropertyPageSite? _site = null;
         private bool _isDirty = false;
         private readonly bool _ignoreEvents = false;
-        private IVsDebugger _debugger;
+        private IVsDebugger? _debugger;
         private uint _debuggerCookie;
         private bool _isActivated = false;
         internal IProjectThreadingService _threadHandling;
@@ -42,7 +42,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
             AutoScroll = false;
         }
 
-        internal UnconfiguredProject UnconfiguredProject { get; set; }
+        internal UnconfiguredProject? UnconfiguredProject { get; set; }
 
 
         ///--------------------------------------------------------------------------------------------
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.Designer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.Designer.cs
@@ -5,7 +5,7 @@
         /// <summary> 
         /// Required designer variable.
         /// </summary>
-        private System.ComponentModel.IContainer _components = null;
+        private System.ComponentModel.IContainer? _components = null;
 
         /// <summary> 
         /// Clean up any resources being used.
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
@@ -17,11 +17,12 @@ namespace Microsoft.VisualStudio
                     return obj;
             }
 
-            return default;
+            return default!;
         }
 
         [Pure]
-        public static T SingleOrDefault<T, TArg>(this IEnumerable<T> source, Func<T, TArg, bool> predicate, TArg arg)
+        public static T? SingleOrDefault<T, TArg>(this IEnumerable<T> source, Func<T, TArg, bool> predicate, TArg arg)
+            where T : class
         {
             using (IEnumerator<T> enumerator = source.GetEnumerator())
             {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
@@ -65,7 +65,9 @@ namespace Microsoft.VisualStudio.ProjectSystem
             // While all callers should wait on InitializeAsync, 
             // only one should complete the completion source
             await instance.InitializeAsync();
-            loadedSource?.SetResult(null!);
+#pragma warning disable CS8625 // Workaround for https://github.com/dotnet/roslyn/issues/31865
+            loadedSource?.SetResult(null);
+#pragma warning restore CS8625
         }
 
         public Task UnloadAsync()
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
@@ -14,7 +14,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
     internal abstract class AbstractMultiLifetimeComponent : OnceInitializedOnceDisposedAsync
     {
         private readonly object _lock = new object();
-        private TaskCompletionSource<object> _loadedSource = new TaskCompletionSource<object>();
+        private TaskCompletionSource<object?> _loadedSource = new TaskCompletionSource<object?>();
         private IMultiLifetimeInstance? _instance;
 
         protected AbstractMultiLifetimeComponent(JoinableTaskContextNode joinableTaskContextNode)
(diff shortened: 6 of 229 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractMultiLifetimeComponent.cs
```diff
@@ -65,12 +65,12 @@ private async Task LoadCoreAsync()
             // While all callers should wait on InitializeAsync, 
             // only one should complete the completion source
             await instance.InitializeAsync();
-            loadedSource?.SetResult(null);
+            loadedSource?.SetResult(null!);
```
Pilchie: Fixed to `#pragma` instead

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsData.cs
```diff
@@ -5,12 +5,12 @@
...
-    /// This class is used to hold the data serialized from the json file.  
+    /// This class is used to hold the data serialized from the json file.
     /// </summary>
     internal class LaunchSettingsData
     {
-        public Dictionary<string, object> OtherSettings { get; set; }
+        public Dictionary<string, object>? OtherSettings { get; set; }
 
-        public List<LaunchProfileData> Profiles { get; set; }
+        public List<LaunchProfileData>? Profiles { get; set; }
```
Pilchie: This required a bunch of changes.  It might be better to ensure that it's default initialized to empty instead.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettings.cs
```diff
@@ -62,7 +65,7 @@ public LaunchSettings(IWritableLaunchSettings settings)
                 else
                 {
                     string jsonString = JsonConvert.SerializeObject(value, Formatting.Indented, jsonSerializerSettings);
-                    object clonedObject = JsonConvert.DeserializeObject(jsonString, value.GetType());
+                    object clonedObject = JsonConvert.DeserializeObject(jsonString, value?.GetType());
```
Pilchie: Switched to `#pragma` with link to issue.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -138,16 +138,16 @@ public IReceivableSourceBlock<ILaunchSettings> SourceBlock
             get
             {
                 EnsureInitialized();
-                return _changedSourceBlock;
+                return _changedSourceBlock!;
```
Pilchie: Legitimate because `EnsureInitialized` causes it to be non-null

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -208,9 +208,14 @@ protected async Task ProjectRuleBlock_ChangedAsync(IProjectVersionedValue<Tuple<
...
                 ruleSnapshot.Properties.TryGetValue(ProjectDebugger.ActiveDebugProfileProperty, out string activeProfile);
-                ILaunchSettings snapshot = CurrentSnapshot;
+                ILaunchSettings? snapshot = CurrentSnapshot;
                 if (snapshot == null || !LaunchProfile.IsSameProfileName(activeProfile, snapshot.ActiveProfile?.Name))
                 {
+                    if (_sequentialTaskQueue == null)
+                    {
+                        throw new ObjectDisposedException(nameof(LaunchSettingsProvider));
+                    }
+
```
Pilchie: Should probably extract a helper here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -137,7 +137,7 @@ protected override void Initialize()
 
             foreach (ExportLifetimeContext<IWorkspaceContextHandler> handler in _handlers)
             {
-                handler.Value.Initialize(_context);
+                handler.Value.Initialize(_context!);
```
Pilchie: This might not be right?  Maybe need to annotate `Initialize` as taking possible nulls?  @davkean ?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractWorkspaceContextHandler.cs
```diff
@@ -20,7 +20,7 @@ protected IWorkspaceProjectContext Context
             {
                 Assumes.NotNull(_context);
 
-                return _context;
+                return _context!;
```
Pilchie: Legit due to the `Assumes.NotNull` just above.

src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
```diff
@@ -17,11 +17,12 @@ internal static class LinqExtensions
                     return obj;
             }
 
-            return default;
+            return default!;
```
jmarolf: see other comment about why we return `T` here instead of `T?`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -146,47 +146,48 @@ private Task ExecuteWithinLockAsync(Func<Task> task)
...
             {
                 await _commonServices.ThreadingService.SwitchToUIThread();
 
-                string newTargetFramework = null;
+                string? newTargetFramework = null;
                 ConfigurationGeneral projectProperties = await _commonServices.ActiveConfiguredProjectProperties.GetConfigurationGeneralPropertiesAsync();
 
                 // Check if we have already computed the project context.
-                if (_currentAggregateProjectContext != null)
+                AggregateWorkspaceProjectContext? currentAggregateProjectContext = _currentAggregateProjectContext;
```
Pilchie: This is a workaround for https://github.com/dotnet/roslyn/issues/31620 which I'm still hitting with Preview 1 at home

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
```diff
@@ -181,8 +181,8 @@ private bool TryGetConfiguredProjectState(ConfiguredProject configuredProject, o
                 }
                 else
                 {
-                    workspaceProjectContext = null;
-                    configuredProjectHostObject = null;
+                    workspaceProjectContext = null!;
```
Pilchie: I decided to do this rather than annotate them, as we don't have the attributes to annotate `Try` methods properly yet.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Logging/ProjectLoggerExtensions.ProjectLoggerBatch.cs
```diff
@@ -43,41 +43,43 @@ public void WriteLine(in StringFormat format)
             {
                 if (IsEnabled)
                 {
-                    Init();
-                    WriteNewLineIfNeeded();
-                    WriteIndent();
-                    Write(format);
+                    StringBuilder builder = GetOrCreateBuilder();
```
Pilchie: This code was fine, but this pattern allowed the compiler to track things easier.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/AssemblyAttributeProperties/AssemblyInfoProperties.cs
```diff
@@ -99,12 +99,15 @@ private async Task<string> GetPropertyValueFromSourceAttributeAsync(string prope
...
         /// Set the value of a property.
         /// </summary>
-        public override async Task SetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IReadOnlyDictionary<string, string> dimensionalConditions = null)
+        public override async Task SetPropertyValueAsync(string propertyName, string? unevaluatedPropertyValue, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
         {
             if (_attributeValueProviderMap.TryGetValue(propertyName, out SourceAssemblyAttributePropertyValueProvider provider) &&
                 !await IsAssemblyInfoPropertyGeneratedByBuild(propertyName))
             {
+#pragma warning disable CS8604
                 await provider.SetPropertyValueAsync(unevaluatedPropertyValue);
```
Pilchie: Please review - not sure if we should be allowing nulls here, and if not, what we should do?  Throw/skip the call/etc?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AppDesignerFolderProjectTreePropertiesProvider.cs
```diff
@@ -58,11 +58,11 @@ public void UpdateProjectTreeSettings(IImmutableDictionary<string, IProjectRuleS
             //
             // TODO: Read these default values from the rules themselves
             // See: https://github.com/dotnet/roslyn-project-system/issues/209
-            string folderName = ruleSnapshots.GetPropertyOrDefault(AppDesigner.SchemaName, AppDesigner.FolderNameProperty, "Properties");
-            string contextsVisibleOnlyInShowAllFiles = ruleSnapshots.GetPropertyOrDefault(AppDesigner.SchemaName, AppDesigner.ContentsVisibleOnlyInShowAllFilesProperty, "false");
+            string? folderName = ruleSnapshots.GetPropertyOrDefault(AppDesigner.SchemaName, AppDesigner.FolderNameProperty, "Properties");
```
jmarolf: won't this never be null because it will be set to `"Properties"` if there is no value?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -384,7 +387,7 @@ private IEnumerable<string> CollectInputs(BuildUpToDateCheckLogger logger)
             if (_customInputs.Count != 0)
             {
                 logger.Verbose("Adding " + UpToDateCheckInput.SchemaName + " inputs:");
-                foreach (string input in _customInputs.Select(_configuredProject.UnconfiguredProject.MakeRooted))
+                foreach (string input in _customInputs.Select(s => _configuredProject.UnconfiguredProject.MakeRooted(s)))
```
Pilchie: This looks like another compiler bug, but I wasn't sure how to file it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
```diff
@@ -67,7 +67,7 @@ internal Enumerator(in LazyStringSplit split)
                 _index = 0;
                 _input = split._input;
                 _delimiter = split._delimiter;
-                Current = null;
+                Current = null!;
```
Pilchie: I decided to allow this, as it's set only when the enumerator is in an invalid state...

Pilchie:
@jcouv @cston @jaredpar would love if you could skim through this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/WritableLaunchSettings.cs
```diff
@@ -42,19 +42,20 @@ public WritableLaunchSettings(ILaunchSettings settings)
                     else
                     {
                         string jsonString = JsonConvert.SerializeObject(value, Formatting.Indented, jsonSerializerSettings);
-                        object clonedObject = JsonConvert.DeserializeObject(jsonString, value.GetType());
+                        // TODO: Bug in nullable flow analysis her?
```
jmarolf: I think this is also [https://github.com/dotnet/roslyn/issues/31803](https://github.com/dotnet/roslyn/issues/31803)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -147,15 +147,19 @@ private void ProcessProjectBuildFailure(IProjectRuleSnapshot snapshot)
             //
             // We still forward those 'removes' of references, sources, etc onto Roslyn to avoid duplicate/incorrect results when the next
             // successful build occurs, because it will be diff between it and this failed build.
-            _context.LastDesignTimeBuildSucceeded = snapshot.IsEvaluationSucceeded();
+            if (_context != null)
```
jmarolf: I assume all this is rewritten in @davkean 's language service branch.  But it feels like we would want to throw something in this case

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
jmarolf: We could add it easily right?

```C#
namespace System.Runtime.CompilerServices
{
    [AttributeUsage(AttributeTargets.Parameter,
                   AllowMultiple = false)]
    public class NotNullWhenTrueAttribute : Attribute
    {
        public NotNullWhenTrueAttribute() { }
    }
}
```

---
In reply to: [242257953](https://github.com/dotnet/project-system/pull/4404#discussion_r242257953) [](ancestors = 242257953)

jmarolf: [https://github.com/dotnet/roslyn/blob/master//src/Compilers/Test/Utilities/CSharp/CSharpTestBase.cs#L55-L64](https://github.com/dotnet/roslyn/blob/master//src/Compilers/Test/Utilities/CSharp/CSharpTestBase.cs#L55-L64)

---
In reply to: [242265057](https://github.com/dotnet/project-system/pull/4404#discussion_r242265057) [](ancestors = 242265057,242257953)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -128,12 +128,12 @@ private void OnProjectChanged(IProjectSubscriptionUpdate e)
             _msBuildProjectFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectFullPathProperty, _msBuildProjectFullPath);
             _msBuildProjectDirectory = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectDirectoryProperty, _msBuildProjectDirectory);
             _outputRelativeOrFullPath = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.OutputPathProperty, _outputRelativeOrFullPath);
-            string msBuildAllProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
+            string? msBuildAllProjects = e.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
```
jmarolf: How can this ever be null?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
```diff
@@ -363,7 +363,7 @@ public void SetObjects(uint cObjects, object[] ppunk)
                 }
             }
 
-            OnSetObjects(false);
+            WaitForAsync(() => OnSetObjects(false));
```
jmarolf: This should either be a call to `SetObjects` if we intend to wait on the result of use a discard if we do not

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
Pilchie: Even having it still seems to require `!`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
Pilchie: Anyone know what we intended?  Feels like wait, but I'm not familiar with this at all.

Pilchie: I guess Discard is the thing that doesn't change behavior...

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/UnconfiguredProjectContextProvider.cs
jmarolf: It should work if you also update the parameters to be nullable.

```C#
private bool TryGetConfiguredProjectState(ConfiguredProject configuredProject, 
									      [NotNullWhenTrue] out IWorkspaceProjectContext? workspaceProjectContext,
                                          [NotNullWhenTrue] out IConfiguredProjectHostObject? configuredProjectHostObject)
{
	...
	workspaceProjectContext = null;
	configuredProjectHostObject = null;
	return false;
}
```

That way the compiler _should_ not warn at the call-site if you check for `true` but _will_ warn if you do not

Pilchie: Ok, done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AppDesignerFolderProjectTreePropertiesProvider.cs
drewnoakes: Is there a way to flow through the nullness of the default value?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPage.cs
drewnoakes: `.Forget()`?

jmarolf:
Remaining warnings are all from auto-generated files in the obj directory.  Is there no way to turn this analysis off for generated code?  Will we need to put `#nullable enable` at the top of all of our source files and remove the project item?

Pilchie:
See https://github.com/dotnet/roslyn/issues/31875

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -390,8 +390,11 @@ private VersionCompatibilityData GetVersionCompatibilityData()
 
                 if (versionCompatData != null)
                 {
+                    // This is never null after InitializeAsync (which is called during package load)
+                    Version ourVSVersion = _ourVSVersion!;
```
jmarolf: Perhaps we want to throw an object disposed exception in this case?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
```diff
@@ -58,7 +58,7 @@ private async Task<bool> IsReadyToBuildAsync()
             // Ensure build manager is initialized.
             await EnsureBuildManagerInitializedAsync();
 
-            ErrorHandler.ThrowOnFailure(_buildManager.QueryBuildManagerBusy(out int busy));
+            ErrorHandler.ThrowOnFailure(_buildManager!.QueryBuildManagerBusy(out int busy));
```
jmarolf: I wish there was a way to restructure this so the compiler could track this better

Pilchie: Agreed.  It's getting to be a pain.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.Designer.cs
```diff
@@ -57,4 +57,4 @@ private void InitializeComponent()
 
         private System.Windows.Forms.Panel wpfHostPanel;
     }
-}
\ No newline at end of file
+}
```
jmarolf: remove added whitespace

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -23,7 +23,7 @@ internal class ConfiguredProjectImplicitActivationTracking : OnceInitializedOnce
         private readonly IActiveConfigurationGroupService _activeConfigurationGroupService;
         private readonly ITargetBlock<IProjectVersionedValue<IConfigurationGroup<ProjectConfiguration>>> _targetBlock;
         private TaskCompletionSource<object> _isImplicitlyActiveSource = new TaskCompletionSource<object>();
```
cston: >TaskCompletionSource<object> _isImplicitlyActiveSource [](start = 16, length = 54)

Should this be `TaskCompletionSource<object?>`, to allow `TrySetResult(null)`?

cston:
>         private readonly TaskCompletionSource<object> _projectLoadedInHost = new TaskCompletionSource<object>();

`TaskCompletionSource<object?>` here and below?

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UnconfiguredProjectTasksService.cs:21 in 1c5b339. [](commit_id = 1c5b3395b8f2d12208c1a945c6b8923fe1e8bc78, deletion_comment = False)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -313,6 +323,11 @@ protected static void MergeExistingInMemoryProfiles(LaunchSettingsData newSnapsh
                     {
                         // Create a new one from the existing in-memory profile and insert it in the same location, or the end if it
                         // is beyond the end of the list
+                        if (newSnapshot.Profiles == null)
```
cston: >newSnapshot.Profiles == null [](start = 28, length = 28)

Is this ever true or would `FirstOrDefault` have thrown `ArgumentNullException` in that case? Consider replacing with `Debug.Assert(newSnapshot.Profiles != null);` which should prevent subsequent warnings.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -663,15 +681,21 @@ protected Task HandleLaunchSettingsFileChangedAsync()
...
 
+                        SequentialTaskExecutor? sequentialTaskQueue = _sequentialTaskQueue;
+                        if (sequentialTaskQueue == null)
+                        {
+                            throw new ObjectDisposedException(nameof(LaunchSettingsProvider));
+                        }
+
                         // Updates need to be sequenced
-                        return _sequentialTaskQueue.ExecuteTask(() => UpdateProfilesAsync(null));
+                        return sequentialTaskQueue.ExecuteTask((() => UpdateProfilesAsync(null)));
```
cston: >((() [](start = 62, length = 4)

Are the extra parentheses necessary?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
drewnoakes: Such restructuring will likely help formalise object lifecycles that were previously implicit.

On this class it's possible to convert `Task EnsureBuildManagerInitializedAsync()` into `Task<IBuildManager> GetBuildManager()` which always returns a non-null instance to a local variable (and itself is the only method to interact with `_buildManager`), and `IsReadyToBuildAsync()` could be converted to a synchronous extention method.

In more complex cases where initialization populates a lot of state, moving that state into a new class where a `EnsureInitialized`-like method returns a non-null instance might help.

Such changes require a fair bit of thought per instance. It'll be interesting to see what patterns emerge and how they effect code quality.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.Designer.cs
Pilchie: Our `.editorconfig` says to have this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
Pilchie: Yes, done.  Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
Pilchie: Nope - artifact of the Expand/Reduce API going wrong. Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
Pilchie: Definitely agree, and have been thinking about the opportunities to extract things into classes together to help enforce things.  See the (simple) change I made in the up to date logging as an example.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
Pilchie: Fixed to force non-null, since it gets the default.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
Pilchie: Good catch - thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: This actually gives better performance anyway as the compiler caches the delegate for lambdas but not for method groups.

Pilchie: I thought that changed very recently, but I could be wrong.

drewnoakes: Nice! I can't reproduce the old behaviour any more, even on 7.3:

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQFEAeAHJzkCWA9gHYBqUCBUwANigD4ACADAARMCMA3ALABQTAMwcATGwDCbAN4C28jiKYAWNgFkAFAEo2chbP4KjbAIIa1WvoePyz2tgF4AfOu2W9Nj8YBiGgHLu1ja+9s5sfm5WNvJeRgBulGwAJo5sJBAA7mwAItpR0b5JAHQW+cEaoS7FmlqBRgC+Xl64BAkwEGwEJDDh9tIcAOxsnKJC3GyNQbpTwhyqZlziUDrSk0azKmzeADwAKk4aTACsey5gK2sKXkzi2TKxCrNdPTUyTEMjY5fyk/VAA===

davkean:
@Pilchie How do you want to move this forward?

Pilchie:
I think this is the wrong approach.  Instead I think we should opt-in to C# 8, but leave nullable off, and let people do PRs a file/directory at a time using `#nullable` until it reaches critical mass.

davkean:
👍 Yeah I was thinking as much.

PR #4405, opened 2018-12-17
remove use of IProjectLockService

jmarolf:
I _think_ this is the correct way to remove the use of `IProjectLockService` but would love some reviewer to correct me.


jmarolf:
@drewnoakes 	This is the last usage of `IProjectLockService` in the codebase.  I'll wait for @davkean 	to wake up and sign off on this before merging

PR #4407, opened 2018-12-18
Add minimal project for documentation

drewnoakes:
This brings the documentation files into the solution so that they are easier to edit and appear in search results.

This probably shouldn't be a `csproj` file. If there's a better project format that allows globbing and is in-box, let me know.

I looked at shared projects but concluded they didn't support globbing (IIRC&mdash;I authored this a while back).


jmarolf:
I would probably just add these as solution files and make sure any new markdown files were added manually.

jmarolf:
If the maintenance burden is felt I would be amenable to adding a project with globbing (or we could just add globbing support to solution files 😃)

drewnoakes:
Is there a reason against doing it as done here? The project isn't built in any configuration, so it seems we get globbing with no downside.

jmarolf:
I don't have any real objections.

PR #4411, opened 2018-12-24
Fix violations of IDE0059

davidwengier:
Just removing some noise from the Error List.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ProjectRootImageProjectTreeModifierTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ProjectRootImageProjectTreeModifierTests.cs
@@ -157,9 +157,14 @@ Root (flags: {ProjectRoot})
 ")]
         public void CalculatePropertyValues_NonProjectRootAsTree_DoesNotSetIcon(string input)
         {
-            var tree = ProjectTreeParser.Parse(input);
+            var imageProvider = IProjectImageProviderFactory.ImplementGetProjectImage(ProjectImageKey.ProjectRoot, new ProjectImageMoniker(new Guid("{A140CD9F-FF94-483C-87B1-9EF5BE9F469A}"), 1));
 
-            Assert.Null(tree.Icon);
+            var propertiesProvider = CreateInstance(imageProvider);
+
+            var tree = ProjectTreeParser.Parse(input);
+            var result = propertiesProvider.ChangePropertyValuesForEntireTree(tree.Children[0]);
+
+            Assert.Null(result.Icon);
         }
 
         [Fact]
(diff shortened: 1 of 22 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/ProjectRootImageProjectTreeModifierTests.cs
```diff
@@ -157,12 +157,7 @@ public void CalculatePropertyValues_WhenSharedItemsImportFileAsTree_SetsIconToSh
 ")]
         public void CalculatePropertyValues_NonProjectRootAsTree_DoesNotSetIcon(string input)
         {
-            var imageProvider = IProjectImageProviderFactory.ImplementGetProjectImage(ProjectImageKey.ProjectRoot, new ProjectImageMoniker(new Guid("{A140CD9F-FF94-483C-87B1-9EF5BE9F469A}"), 1));
-
-            var propertiesProvider = CreateInstance(imageProvider);
-
             var tree = ProjectTreeParser.Parse(input);
-            var result = propertiesProvider.ChangePropertyValuesForEntireTree(tree.Children[0]);
```
davkean: The bug here is that the Assert should be checking "result" not "tree".

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UnconfiguredProjectTasksServiceTests.cs
```diff
@@ -101,7 +101,6 @@ public void UnloadCancellationToken_WhenUnderlyingUnloadCancellationTokenCancell
         {
             var source = new CancellationTokenSource();
             var tasksService = IProjectAsynchronousTasksServiceFactory.ImplementUnloadCancellationToken(source.Token);
-            var service = CreateInstance(tasksService);
 
             Assert.False(tasksService.UnloadCancellationToken.IsCancellationRequested);
```
davkean: Bug here is that Assert should be checking "service".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/OutputTypeExValueProviderTests.cs
```diff
@@ -52,8 +52,7 @@ public async Task SetValue(string incomingValue, string expectedOutputTypeValue)
                     SetValues = setValues
                 });
             var provider = new OutputTypeExValueProvider(properties);
-
-            var actualPropertyValue = await provider.OnSetPropertyValueAsync(incomingValue, null);
+            await provider.OnSetPropertyValueAsync(incomingValue, null);
             Assert.Equal(setValues.Single(), expectedOutputTypeValue);
```
davkean: Looks like it's checking the wrong result, I suspect "actualPropertyValue" is supposed to be in the assert.

davidwengier: Looks like `OnSetPropertyValueAsync` has a hardcoded null return, and this test has been broken in this way for a while, so the removal makes sense.

davidwengier:
With respect to the commit stream that makes up this PR: This place is not a place of honor.. no highly esteemed deed is commemorated here... nothing valued is here.

😛

PR #4415, opened 2018-12-27
Fix ide0060 warnings

davidwengier:
Some more compiler noise since updating roslyn packages.

I've gone through the test changes and they seem to be fine to me, but some of them do seem a little light on substance now, so I could be mistaken.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/CreateFileFromTemplateServiceTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/CreateFileFromTemplateServiceTests.cs
@@ -59,11 +59,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
 
         [Theory]
-        [InlineData(@"Root (flags: {ProjectRoot}), FilePath: ""C:\Foo\""", @"C:\Path\To\TemplateFile", true)]
-        [InlineData(@"Properties, FilePath: ""C:\Foo\""", @"C:\Path\To\TemplateFile", true)]
-        [InlineData(@"Root (flags: {ProjectRoot}), FilePath: ""C:\Foo\""", @"C:\Path\To\TemplateFile", false)]
-        [InlineData(@"Root (flags: {ProjectRoot}), FilePath: ""C:\Foo\""", null, false)]
-        public async Task CreateFile(string input, string templateFilePath, bool expectedResult)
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/CreateFileFromTemplateServiceTests.cs
```diff
@@ -59,11 +59,10 @@ public async Task CreateFile_NullFileName_ThrowsArgumentNull()
...
         [Theory]
-        [InlineData(@"Root (flags: {ProjectRoot}), FilePath: ""C:\Foo\""", @"C:\Path\To\TemplateFile", true)]
-        [InlineData(@"Properties, FilePath: ""C:\Foo\""", @"C:\Path\To\TemplateFile", true)]
-        [InlineData(@"Root (flags: {ProjectRoot}), FilePath: ""C:\Foo\""", @"C:\Path\To\TemplateFile", false)]
-        [InlineData(@"Root (flags: {ProjectRoot}), FilePath: ""C:\Foo\""", null, false)]
-        public async Task CreateFile(string input, string templateFilePath, bool expectedResult)
+        [InlineData(@"C:\Path\To\TemplateFile", true)]
+        [InlineData(@"C:\Path\To\TemplateFile", false)]
+        [InlineData(null, false)]
+        public async Task CreateFile(string templateFilePath, bool expectedResult)
```
davidwengier: `input` was unused after [this change](https://github.com/dotnet/project-system/commit/425106035bb404a5ef720e664ad073c2dedf5324) that removed the project tree parsing, but it does make this test a little smelly IMO.. It seems to me that its basically asserting some of its own setup code. Also one of the sets of parameters was redundant.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -51,9 +51,6 @@ internal class LaunchSettingsProvider : OnceInitializedOnceDisposed, ILaunchSett
             _launchSettingsFilePath = new AsyncLazy<string>(GetLaunchSettingsFilePathNoCacheAsync, commonProjectServices.ThreadingService.JoinableTaskFactory);
         }
 
-        // TODO: Add error list support. Tracked by https://github.com/dotnet/roslyn-project-system/issues/424
```
davidwengier: I've commented on this issue with a link to this commit, so it can be easily reverted when/if that job is ever done.

PR #4422, opened 2018-12-31
Update name for nullable property. Change to string

jcouv:
Relates to https://github.com/dotnet/project-system/issues/4058 (adding project system support for nullable)
Relates to https://github.com/dotnet/project-system/pull/4277 (PR adding boolean `NullableReferenceTypes` setting)

@davkean for review. Please let me know what kind of validation I should do, if any.

FYI @AlekseyTs

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralConfiguredBrowseObject.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralConfiguredBrowseObject.xaml
@@ -32,11 +32,11 @@
       <DataSource Persistence="ProjectFile" PersistedName="IntermediateOutputPath" HasConfigurationCondition="True" SourceOfDefaultValue="AfterContext" />
     </StringProperty.DataSource>
   </StringProperty>
-  <BoolProperty Name="NullableReferenceTypes" Visible="False">
-    <BoolProperty.DataSource>
+  <StringProperty Name="NullableContextOptions" Visible="False">
+    <StringProperty.DataSource>
       <DataSource Persistence="ProjectFile" SourceOfDefaultValue="AfterContext" HasConfigurationCondition="False" />
-    </BoolProperty.DataSource>
-  </BoolProperty>
+    </StringProperty.DataSource>
+  </StringProperty>
   <BoolProperty Name="RunCodeAnalysis" Visible="False" />
   <BoolProperty Name="Prefer32Bit" Visible="False"/>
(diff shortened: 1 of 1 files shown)


Pilchie:
Don't we want this to be an enum property instead?

jcouv:
Just to clarify, I assume this PR can go in for preview2 sometime next week. If not, let me know, as it means I need to sort this PR out today.

Pilchie:
I don't have a great handle on the differences for things that don't show up in the Property Grid (toolwindow).  Maybe @davkean or @davidwengier knows better?

As for timing, if this isn't merged and inserted into the VS build by 6pm today (already highly unlikely), it will need QB approval for Preview 2.  Based on previous discussions, I'm not sure how likely that is.

jcouv:
@Pilchie Jared thinks this can go in (along with other roslyn PRs) per last discussion in infrastructure meeting.

jcouv:
@davkean for second review. 
Please advise on practical difference between using a string or an enum. In particular, is a string ok for preview2 (assuming we can get this PR approved/merged/inserted) since no UI?

jcouv:
So, I tried this scenario on "Int Preview". Created a C# 8.0 Core project with `<NullableContextOptions>enable</NullableContextOptions>`.
To my surprise, proper squiggles showed up, and the warnings showed up in the Error List for "Intellisense" and for "Build" (once I built).

The only other impact I can foresee is that the property would not be visible to a fixer (that enables the feature). We don't have such a fixer in preview2.

Let me know if I missed something. Since I didn't observe any noticeable degradation of experience, I am fine to pushing this out.
If you want to use an enum instead of a string, feel free to close this PR.

jcouv:
@davkean Could you advise on string versus enum?

Also, I expected things would be more broken, but in my testing I could not find obvious symptoms of the mismatch (compiler and project-system recognizing different names). Did I miss something? Answering this would be useful to determine the priority of fixing this.

jcouv:
@davkean I didn't hear back from you.
Two questions:
- should the new property be a string or an enum?
- why aren't things observably broken in the current state (am I missing something)?

jcouv:
I'm going to close this PR. 
@jjmew is scheduled to work on integrating new build settings (in particular the legacy project system part), and I'll let him cover this too.

davkean:
@jcouv What is the current state of this property? This is public API, so if it has the wrong shape I'd like to remove it.

jcouv:
@davkean The `<NullableReferenceTypes>true/false<...>` property was replaced with `<NullableContextOptions>` with values such as `enable`, `disable`, `safeonly`, ... and others listed [here](https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-reference-types.md#setting-project-level-nullable-context).

PR #4426, opened 2019-01-04
add packages.lock files for deterministic restore

jmarolf:
As per [this blog post](https://blog.nuget.org/20181217/Enable-repeatable-package-restores-using-a-lock-file.html) by checking in packages.lock files we can ensure deterministic restore for this repository

### DIFF
--- a/src/DeployIntegrationDependencies/packages.lock.json
+++ b/src/DeployIntegrationDependencies/packages.lock.json
@@ -1,0 +1,1215 @@
+{
+  "version": 1,
+  "dependencies": {
+    ".NETFramework,Version=v4.7.2": {
+      "EnvDTE": {
+        "type": "Direct",
+        "requested": "[8.0.1, )",
+        "resolved": "8.0.1",
+        "contentHash": "kcxy9mqppXJpvBWyY2KMAosyyVWJWMYOb4zhVpC2s5ggFbpNcUHOILsCUGTT9gZbj+O7ILH8R1yryuEwGxqnmg==",
+        "dependencies": {
+          "stdole": "7.0.3301"
+        }
+      },
+      "EnvDTE80": {
+        "type": "Direct",
+        "requested": "[8.0.1, )",
+        "resolved": "8.0.1",
+        "contentHash": "07uC6dkry4+6iXaubgShBvjvPLautWExE9UHnfuOcXWqgZ6YSXcAYRBpBNe3Vrxj8XB6FdgCjy088BWly9OAKw==",
+        "dependencies": {
+          "EnvDTE": "8.0.1",
+          "stdole": "7.0.3301"
+        }
+      },
+      "EnvDTE90": {
+        "type": "Direct",
+        "requested": "[9.0.1, )",
+        "resolved": "9.0.1",
+        "contentHash": "kbhX8MkY2GeB+eWKnT6yqoF0XiANNOw6SHzZ1LMfl2Fxlv+02L8EBmPyyMOTxidtsXCiDSEjjM9mD0V3/Kyr8Q==",
+        "dependencies": {
+          "EnvDTE": "8.0.1",
+          "EnvDTE80": "8.0.1",
+          "stdole": "7.0.3301"
+        }
+      },
+      "Microsoft.Build": {
+        "type": "Direct",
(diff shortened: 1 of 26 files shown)


jmarolf:
@drewnoakes 	Today we restore a _lot_ of our packages from myget. Roslyn for example deploys all builds from all branches to the same myget feed.  It is possible that a future deployment (or incorrect build) by roslyn would cause us to pick up the wrong package without this.  Also, I'm paranoid.

jmarolf:
Turns out this does not play well with the way our implicit toolset restore works.  I can see no means of fixing it without changing out repo toolset which I am not inclined to do.

davkean:
Given our current setup, I believe this protects us from the same package from being replaced with another package. Due to warnings as errors, any roll forward should already break the build.

jmarolf:
@davkean 	we do not require that our version number be exact today.  We specify that it must be >= the version number that we specify.  As per the [lowest applicable version](https://docs.microsoft.com/en-us/nuget/consume-packages/dependency-resolution#lowest-applicable-version) rule, if we depend on Roslyn `3.1.0-beta` and that package is deleted in the future, nuget will silently select the next lowest version, say `3.2.0-beta`.  This _could_ mean that when we try and build a servicing branch in the future (perhaps several months or years from now) instead of failing the build, telling us that it cannot find our dependency, nuget will silently upgrade our package graph.

I will readily admit that this is a rather unlikely scenario, but again, I'm paranoid about deterministic builds.

jmarolf:
Specifically, if either of these three situations occurs today in myget our build will not be deterministic

1. **Intermediate versions:** A missing version of the package, matching PackageReference version requirements, is published:
Day 1: If you specified <PackageReference Include="My.Sample.Lib" Version="4.0.0"/> but the versions available on the NuGet repositories were 4.1.0, 4.2.0 and 4.3.0, NuGet resolves to 4.1.0 as it the nearest minimum version.
Day 2: Version 4.0.0 gets published. NuGet now restores version 4.0.0 as it is an exact match.
2. **Package deletion:** Though nuget.org does not allow package deletions, not all package repositories have this constraint. Deletion of a package version results in NuGet finding the best match when it cannot resolve to the deleted version.
3. **Package content mismatch:** If the same package (id and version) is present with different content across repositories, then NuGet cannot ensure the same package (with the same content hash) gets resolved every time. It also does not warn/error out in such cases

**NOTE:** 3 would likely mean that the nuget feed was compromised and was serving up bogus packages

davkean:
It will not silently upgrade the version. Here's an example, change a version number in Packages.targets to lower than the version you asked for and build and you'll get:

```
\project-system2\src\Microsoft.VisualStudio.ProjectSystem.FSharp.VS.UnitTests\Microsoft.VisualStudio.ProjectSystem.FSharp.VS.UnitTests.csproj : error NU1603: Microsoft.VisualStudio.ProjectSystem.FSharp.VS.UnitTests depends on Moq (>= 1.0.0) but Moq 1.0.0 was not found. An approximate best match of Moq 3.1.416.3 was resolved.
```

jmarolf:
For example, we don't explicitly depend on `System.Collections.Immutable` but every assembly in the repo implicitly depends on version `1.5.0`.

davkean:
An inherited dependency can change based on a new or removed project reference, but if one of our direct dependencies says it depends on `1.5.0` of immutable and that is upgraded to `1.6.0` that would also result in an error.

jmarolf:
>  if one of our direct dependencies says it depends on 1.5.0 of immutable and that is upgraded to 1.6.0 that would also result in an error.

Today we have several dependencies that specify multiple versions of lots of assemblies. `System.Collections.Immutable` for example has two versions in our graph `1.3.1` from our CPS dependency, and `1.5.0` from our roslyn dependency.  My understanding of what you've said is: if we add a new dependency that requires version `1.5.1` it will unify to that, but if version `1.5.0` was deleted it would never select `1.5.1` without issuing `NU1603`. Sounds good.

My final bit of paranoia is this: imagine that someone uploaded package version `3.1.0-beta` to two myget feeds, both of which we reference.  I do not think the restore operation is deterministic in that case because nuget does not check the hash of the contents so impersonation is possible. Essentially nuget does not take the packages location or content into account when deciding about where to pull from.

PR #4430, opened 2019-01-06
Dependencies tree and graph changes

drewnoakes:
Following #4429, more changes to the dependencies node's tree and graph logic.

In particular:

- 97f66bb6c18c2c1a9c6cf25867e1470b64c1e381 Rename `GroupedByTargetTreeViewProvider` to `DependenciesTreeViewProvider`
- 6cdde4b36082aa654bfac3e6b1f957586d840222 Extract tracking of changes to graph contexts from `DependenciesGraphProvider` to new `DependenciesGraphChangeTracker` class
- 81b81c0a1268f305f7ff71c2e6a1725e3bdfcc63 More API documentation
- da983990ed8294a6644753d95df2b2ab0089eb90 Remove duplication of project path from `IDependenciesSnapshotProvider` and its snapshot&mdash;allow the snapshot to be the source of truth over time, as mentioned in https://github.com/dotnet/project-system/pull/4386#discussion_r241693587
- 141f7453c1f4984d69aca64164c411d28c6c8b13 Remove `JTF.RunAsync` from `DependenciesGraphProvider.BeginGetGraphData` by implementing/exporting `IProjectDynamicLoadComponent` for async initialisation
- 16071092a58f71a3ad50fac22ed47e0dade002ab `DependenciesGraphChangeTracker` avoids notifying of graph context changes when nothing actually changed

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/Actions/TrackChangesGraphActionHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/Actions/TrackChangesGraphActionHandler.cs
@@ -39,6 +39,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.GraphNodes.A
                 return false;
             }
 
+            bool anyChanges = false;
+
             foreach (GraphNode inputGraphNode in graphContext.InputNodes.ToList())
             {
                 string existingDependencyId = inputGraphNode.GetValue<string>(DependenciesGraphSchema.DependencyIdProperty);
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphChangeTracker.cs
```diff
@@ -0,0 +1,130 @@
...
+    internal sealed class DependenciesGraphChangeTracker : IDependenciesGraphChangeTracker
+    {
+        private readonly object _snapshotChangeHandlerLock = new object();
+
+        /// <summary>
+        /// Remembers expanded graph nodes to track changes in their children.
+        /// </summary>
+        private readonly WeakCollection<IGraphContext> _expandedGraphContexts = new WeakCollection<IGraphContext>();
+
+        [ImportMany] private readonly OrderPrecedenceImportCollection<IDependenciesGraphActionHandler> _graphActionHandlers;
```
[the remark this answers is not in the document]
drewnoakes: I'm not sure if that works. Every example I've seen does it this way. The consumer can parameterise the import collection when constructing it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/Actions/TrackChangesGraphActionHandler.cs
```diff
@@ -73,18 +75,21 @@ public override bool HandleChanges(IGraphContext graphContext, SnapshotChangedEv
...
+                    {
+                        anyChanges = true;
+                    }
 
                     scope.Complete();
                 }
             }
 
-            return false;
+            return anyChanges;
```
Pilchie: Was this a bug that it always returned `false`?

drewnoakes: The return value wasn't being used before.

PR #4432, opened 2019-01-07
Pass through RootNamespace to language-service

davkean:
Introduce a new rule; LanguageService.xaml that lets us pass through properties to the language-service, such as RootNamespace, without needing to write any additional code - similar to NuGutRestore.xaml.


genlu:
I'm trying to test this locally, but it seems the new code in `ProjectPropertiesItemHandler` isn't called though.

davkean:
RootNamespace comes from both. This is untested yet, I'll look this morning.

genlu:
@davkean Any update on this?

davkean:
I had a bug where the format string I was passing the logger was wrong, after I fixed this - this worked for me.

genlu:
I tried to build and deploy from latest source again, still couldn't get this to work. Same behavior as I mentioned before, when opening a solution, "roonamespace" property is part of the project change of `ConfigurationGeneral` evaluation rule, not `LanguageService`. Did I miss anything?

davkean:
RootNamespace will be a part of both rules.

You need launch using Launch.cmd or F5 from our solution for the new rules to take effect. If you don't do this, the new rule will not be picked up.

PR #4434, opened 2019-01-07
Graph action handlers

drewnoakes:
Following #4430, changes to the dependencies node graph action handlers.

In particular:

- a049042851da43fd1eb08eab6340bfa21095d9b4 More efficient implementation of `WeakReference` tracking of known `IGraphContext` objects, avoiding secondary locking, `ToList` calls and allocation during enumeration, on a code path that's called heavily on solutions with many projects, especially during solution load.
- 75f3feeb16d969835c25c7f6b3e45646f92f383b Divides `IDependenciesGraphActionHandler` so that it handles one kind of operation, not two, which simplified the surrounding implementation.
- db6445eed7c5f4aabe9ecfa8dddcc610228e8801 Merges the two members (`CanHandleRequest` and `HandleRequest`) on `IDependenciesGraphActionHandler` into `TryHandleRequest`, as the single caller (`DependenciesGraphProvider`) always called them in succession.

Misc other refactoring and documentation.

Please review commit by commit.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphChangeTracker.cs
```diff
@@ -126,5 +114,69 @@ public void Dispose()
         {
             _aggregateSnapshotProvider.SnapshotChanged -= OnSnapshotChanged;
         }
+
+        private sealed class WeakCollection<T> where T : class
```
Pilchie: I suggest a comment about what this is replacing and why it isn't good enough.

Alternatively, how would a `ConditionalWeakTable` work for this case?

drewnoakes: 👍 I'll look at this in my morning tomorrow.

drewnoakes: Documentation added in #4439.

`ConditionalWeakTable` is useful when you want to add arbitrary fields to objects you don't control. In this case we don't control the lifetime of the `IGraphContext` instances, so use `WeakReference` to infer when they're no longer in the graph. I am not familiar enough yet with the VS graph code to know if there's a better way to track this, but I do know that the change in this PR preserves functionality while avoiding a bunch of allocation, especially when expanding lots of nodes (i.e. searching).

PR #4435, opened 2019-01-07
Address code duplication

drewnoakes:
Follows #4434.

- 27a9164521d955c3626faa640080d9772a72d195 extracts near-identical code from `GetChildrenGraphActionHandler` and `CheckChildrenGraphActionHandler` into new `InputNodeGraphActionHandlerBase`.
- d2179ec808afbec5736db45a286517a6339de808 moves identical `GetSnapshot` methods from `UnsupportedProjectsSnapshotFilter` and `ProjectGraphViewProvider` to `IAggregateDependenciesSnapshotProvider`. This makes tests nicer too due to less mocking.
- f67bcb34affd7770e4051f7bba7d24e0baa004e8 simplifies `IAggregateDependenciesSnapshotProvider` as all consumers wanted snapshots, not snapshot providers. This makes tests nicer too due to less mocking.
- 4b6feb3a48595e276aeea5fedcc8e678f269a883 replaces an O(N) scan through a snapshot's `TopLevelDependencies` array with an O(1) lookup in `WorldDependencies`, as the latter is a superset of the former.

Misc refactoring and API documentation.


drewnoakes:
Will review test failure in my morning tomorrow.

PR #4436, opened 2019-01-08
Support getting compatible property values via GetCommonProperties

davidwengier:
CPS supports tagging properties providers that can get property values from a ProjectInstance, which is more efficient, so this PR creates a new provider that opts into it (via the `HasEquivalentProjectInstancePropertiesProvider` metadata), and then moves any of our intercepted properties that are compatible with it, to that provider.

Compatibility was determined by simply testing within Visual Studio - TargetFrameworkMonikers fails here, as it needs access to all configured projects, which is not available sometimes with this model - and by querying each property via DTE - PreBuildEvent and PostBuildEvent fail here because of a NotSupportedException in the PageProperty class.

The two tests in `InterceptedProjectPropertiesProviderTests.cs` have been cribbed from the now-reverted https://github.com/dotnet/project-system/pull/3464 (thanks @Pilchie 😄)

Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/593209

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
@@ -42,7 +42,7 @@
     </StringProperty>    
     <StringProperty Name="TargetFrameworkMoniker" DisplayName="Target Framework Moniker">
         <StringProperty.DataSource>
-            <DataSource Persistence="ProjectFileWithInterception" PersistedName="TargetFrameworkMoniker" HasConfigurationCondition="False" SourceOfDefaultValue="AfterContext" />
+            <DataSource Persistence="ProjectFileWithInterceptionViaSnapshot" PersistedName="TargetFrameworkMoniker" HasConfigurationCondition="False" SourceOfDefaultValue="AfterContext" />
         </StringProperty.DataSource>
     </StringProperty>
     <StringProperty Name="TargetFrameworkMonikers" DisplayName="Target Framework Monikers" ReadOnly="True">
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -42,7 +42,7 @@
     </StringProperty>    
     <StringProperty Name="TargetFrameworkMoniker" DisplayName="Target Framework Moniker">
         <StringProperty.DataSource>
-            <DataSource Persistence="ProjectFileWithInterception" PersistedName="TargetFrameworkMoniker" HasConfigurationCondition="False" SourceOfDefaultValue="AfterContext" />
```
davkean: What did you take this approach instead of just replacing existing version? When we would ever want to use ProjectFileWithInterception?

davidwengier: The existing version is still needed for TargetFrameworkMonikers, PreBuildEvent and PostBuildEvent as they use things that are not available when opting in to CPSes new code path.

Originally https://github.com/dotnet/project-system/pull/3464 tried to change the existing implementation, but it got reverted because of the errors from those properties (well, specifically from TargetFrameworkMonikers). In the DevDiv work item linked above there is a comment from Lifeng (sorry, can't link directly to comments it seems) suggesting that it would be beneficial to add a new provider and move over whichever properties can be moved, to remove UI delays, so thats what I did.

davkean: It's not clear to me when or when we wouldn't use this implementation. Those properties that ring a bell as to why they would be special. How do we make a decision to use the data source or not?

davidwengier: I don't claim to know the full context of the CPS changes, but anecdotally the new provider can be used if the value provider can do its work based on the unevaluated value (eg OutputType converting strings to ints), but the existing provider if more there are more complex requirements (eg TargetFrameworkMonikers which iterates through all configurations).

Realistically for new properties being added, it probably wouldn't matter which provider is used unless the property value is requested enough to warrant the new provider for performance reasons (values are read from the snapshot, instead of under a lock).

davkean:
You can merge this in, those other properties need a bunch of work to make happen.

davidwengier:
I'll log jobs :)

PR #4439, opened 2019-01-08
Document weak collection

drewnoakes:
Add documentation to the `WeakCollection<T>` type as discussed in https://github.com/dotnet/project-system/pull/4434#discussion_r245808709.

Also use the generic form of `WeakReference<T>`.


drewnoakes:
@Pilchie I can't see a way to enumerate the contents of a `ConditionalWeakTable<TKey, TValue>`, which is a requirement for our purposes. Basically every time we get an update to a project's dependencies, we go through all known graph context to see if they need updating.

Pilchie:
Huh, never noticed that!  Definitely a problem for this use case :)

PR #4440, opened 2019-01-08
Dependencies node graph work

drewnoakes:
Follows #4435.

- c363e9428418b753ed41182bcd217f16bc95a616 merges identical code in `GraphViewProviderBase` hierarchy
- b53dfdf4aea1b1dec457a13cc83c08e778f7fd67 reduces allocations when comparing sets of data during dependencies updates
- 5529d6661c1dbed43ee12bc7082764fac25d1784 renames `DependencySubscriptionsHost` to `DependenciesSnapshotProvider` (what it does rather than how it does it)

Misc refactoring and documentation.


drewnoakes:
Build failure is due to xunit/xunit#1877. That diagnostic's not failing my local build though.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
```diff
@@ -172,9 +173,10 @@ private async Task HandleAsync(Tuple<IProjectSubscriptionUpdate, IProjectSharedF
                 .ToList();
             IEnumerable<string> currentSharedImportNodePaths = currentSharedImportNodes.Select(x => x.Path);
 
+            var diff = new SetDiff<string>(currentSharedImportNodePaths, sharedFolderProjectPaths);
+
             // process added nodes
-            IEnumerable<string> addedSharedImportPaths = sharedFolderProjectPaths.Except(currentSharedImportNodePaths);
-            foreach (string addedSharedImportPath in addedSharedImportPaths)
+            foreach (string addedSharedImportPath in diff.Added)
```
jmarolf: While this is simpler could we, instead of allocating a lot of enererables can't we just keep things an Immutable Array since that's what `targetedSnapshot.TopLevelDependencies` starts as?

drewnoakes: There's no allocation here for the enumerable. The `SetDiff<>` type uses a single dictionary to track which items have been added/removed, and the `Added` and `Removed` properties use struct enumerators to walk through that dictionary without allocating.

If you want an immutable array, you can call `ToImmutableArray()` and pay the cost.

As it stands, the approach here cuts allocations in half and is about 75% faster in my benchmarks.

Or am I missing something?

jmarolf: I guess I'll need to run some benchmarks and see if this changes things

```C#
ImmutableArray<IDependency> topLevelDependencies = targetedSnapshot.TopLevelDependencies;
var builder = PooledArray<string>.GetInstance(topLevelDependencies.Length);
for (int i = 0; i < topLevelDependencies.Length; i++)
{
    IDependency topLevelDependency = topLevelDependencies[i];
    if (topLevelDependency.Flags.Contains(DependencyTreeFlags.SharedProjectFlags))
    {
        builder.Add(topLevelDependency.Path);
    }
}
ImmutableArray<string> currentSharedImportNodePaths = builder.ToImmutableAndFree();
```

Can you share the benchmarks you are using?

---
In reply to: [245998020](https://github.com/dotnet/project-system/pull/4440#discussion_r245998020) [](ancestors = 245998020)

jmarolf:
>             var currentSharedImportNodes = targetedSnapshot.TopLevelDependencies

Sorry, I am talking about here where we are given an array and then turn it into an IEnumerable

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs:171 in 4dc2309. [](commit_id = 4dc23096077e383abe3e026d0acfc6c803e2ce72, deletion_comment = False)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
drewnoakes: Oh I see, we're talking about different bits of code. I was referring to the change you highlighted, which replaces two symmetric Linq `Except` calls with `SetDiff`. You're referring to the code beforehand which isn't changed in this commit. Is that right? I agree we can do more work to improve the code before this. I haven't spent much time on the shared project code, nor the graph search stuff. Your code looks sensible.

drewnoakes: In this case I would probably not go for an immutable collection as the collection object never escapes the method. It is preferable to reduce iteration here by moving the `Select` before the `ToList` however, which I've done on the tip of my branch and which should surface for review before long.

PR #4448, opened 2019-01-09
Don't allocate to compare strings during graph search

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/Actions/SearchGraphActionHandler.cs
```diff
@@ -164,7 +164,7 @@ private static HashSet<IDependency> SearchFlat(string searchTerm, IDependenciesS
             {
                 foreach ((string _, IDependency dependency) in targetedSnapshot.DependenciesWorld)
                 {
-                    if (dependency.Visible && dependency.Caption.ToLowerInvariant().Contains(searchTerm))
+                    if (dependency.Visible && dependency.Caption.IndexOf(searchTerm, StringComparison.InvariantCultureIgnoreCase) != -1)
```
davkean: Is this supposed to be en-US? I would imagine this is supposed to be culturally aware.

drewnoakes: Seems reasonable. Updated.

PR #4460, opened 2019-01-14
Make ConcreteApplicationSettings Serializable Again

drewnoakes:
Provides an alternative fix to that in #4402 where we enabled an analyzer that detects serializable types with non-serializable fields.

The VS shell needs this type to be serializable even though it must not be doing any serialization (or was swallowing exceptions).

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/765098

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsGlobalObject/SettingsGlobalObjectProvider.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsGlobalObject/SettingsGlobalObjectProvider.vb
@@ -2312,11 +2312,17 @@ Namespace Microsoft.VisualStudio.Editors.SettingsGlobalObjects
         ''' <summary>
         ''' 
         ''' </summary>
-        ''' <remarks></remarks>
+        ''' <remarks>
+        ''' This type must be serializable to satisfy the VS shell.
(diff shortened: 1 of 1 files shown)


jmarolf:
@drewnoakes 	just to confirm: we want this fix in preview 3 not preview 2 correct?

Pilchie:
Tagging @jjmew.  We should probably follow-up with the filing team about what timeline they want to see a fix for this in.

jjmew:
> Tagging @jjmew. We should probably follow-up with the filing team about what timeline they want to see a fix for this in.

Which team filed the issue? was it Office? I tried to find it but the GitHub issue was filed by Drew

jjmew:
@drewnoakes  if there is already an email thread, could you add me to it?

drewnoakes:
@jjmew so far all discussion's taken place on the ADO thread linked in the original issue. I tagged @jmarolf there too so he can let everyone know once it's inserted (this is still a manual process as I understand it).

drewnoakes:
> @drewnoakes just to confirm: we want this fix in preview 3 not preview 2 correct?

I think you're right, by reading the linked issue. I'll ask there to verify.

PR #4461, opened 2019-01-14
LOC CHECKIN | dotnet/project-system master | 20190114

v-zbsail:


drewnoakes:
The diff would be much smaller if there weren't any BOM changes.

sharwell:
@v-zbsail I'm going to apply a patch to this PR to revert BOM changes. We will review the content after that change is applied.

tmeschter:
@v-zbsail A `<target>` element has been deleted from src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/xlf/PropPageDesignerView.pt-BR.xlf, and this will break our build.

PR #4463, opened 2019-01-15
Optimise linq a bit

drewnoakes:
Reduces the amount of enumeration.

Following suggestion by @jmarolf in https://github.com/dotnet/project-system/pull/4440#pullrequestreview-190241568.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
```diff
@@ -168,10 +168,10 @@ private async Task HandleAsync(Tuple<IProjectSubscriptionUpdate, IProjectSharedF
             }
 
             IEnumerable<string> sharedFolderProjectPaths = sharedFolders.Value.Select(sf => sf.ProjectPath);
-            var currentSharedImportNodes = targetedSnapshot.TopLevelDependencies
+            var currentSharedImportNodePaths = targetedSnapshot.TopLevelDependencies
                 .Where(x => x.Flags.Contains(DependencyTreeFlags.SharedProjectFlags))
+                .Select(x => x.Path)
                 .ToList();
```
jmarolf: Why do we need to call `ToList` here?

drewnoakes: The result's used in the foreach below.

PR #4464, opened 2019-01-15
Only update graph context when something changes

drewnoakes:


jmarolf:
>                 }

What can cause this to throw?

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs:108 in fa552d7. [](commit_id = fa552d7e7cabcdcc2e413ef0c86666486316025c, deletion_comment = False)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -105,7 +109,10 @@ public void BeginGetGraphData(IGraphContext context)
                 finally
                 {
                     // OnCompleted must be called to display changes 
-                    context.OnCompleted();
+                    if (anyChanges)
+                    {
+                        context.OnCompleted();
```
tmeschter: What's the downside to calling `OnCompleted` if there are no changes?

Pilchie: This seems good, but agree that it would be nice to know that the assumption that `OnCompleted` is expensive has been validated.

PR #4489, opened 2019-01-23
Stop localizing AppDesigner/CompilerCommandLineArgs rules

davkean:
These do not contain any user-visible strings.


davkean:
As a heads up this went a little more extensive, I went and removed all the new rules that we've added just for the ProjectSubscriptionService and stopped localizing them.

PR #4500, opened 2019-01-25
Remove IServiceProvider usage from AbstractAddClassProjectCommand

davkean:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddClassProjectCommand.cs
```diff
@@ -13,19 +13,19 @@ internal abstract class AbstractAddClassProjectCommand : AbstractSingleNodeProje
...
 
         protected abstract string DirName { get; }
 
-        protected AbstractAddClassProjectCommand(IPhysicalProjectTree projectTree, IUnconfiguredProjectVsServices projectVsServices, SVsServiceProvider serviceProvider)
+        protected AbstractAddClassProjectCommand(IPhysicalProjectTree projectTree, IUnconfiguredProjectVsServices projectVsServices, IVsUIService<IVsAddProjectItemDlg> addItemDialog)
         {
             Requires.NotNull(projectTree, nameof(IPhysicalProjectTree));
             Requires.NotNull(projectVsServices, nameof(IUnconfiguredProjectVsServices));
-            Requires.NotNull(serviceProvider, nameof(SVsServiceProvider));
+            Requires.NotNull(addItemDialog, nameof(SVsServiceProvider));
```
jmarolf: Should the nameof expressions be changed here?

davkean: Yes, weirdly they are referring to the type rather than the parameter so I didn't get a compile error. Fixed.

PR #4504, opened 2019-01-29
Add UI for a setting to turn off double click to edit project file

### LINKED ISSUE #4493: We need a setting to define the what double clicking project does
Version: 16.0 Preview 2

### Steps to reproduce:

1. Open a solution with a CPS-based C# project
2. Double click to expand the project in Solution Explorer.

### Expected results:

The project expands in Solution Explorer.

### Actual results:

The project does not expand in Solution Explorer, and instead the project opens.

📝 I was aware of the behavior change but avoided filing this issue until I actually hit it by accident. It happened within 5 minutes of installing Preview 2.

davidwengier:
The code that checks this setting is in CPS.

Fixes #4493.

Note: This may not go ahead as the UI may be moved to the VS repo.


src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptionPageResources.resx
```diff
@@ -138,4 +138,7 @@
   <data name="General_UsePreviewSdk" xml:space="preserve">
     <value>Use previews of the .NET Core SDK</value>
   </data>
+  <data name="General_DoubleClickOpenProjectFile" xml:space="preserve">
+    <value>Double click on project node to edit project file.</value>
```
davkean: This should not have a trailing ".".

davkean: Also missing an access key.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptionPageControl.xaml
```diff
@@ -17,6 +17,9 @@
             <StackPanel Orientation="Horizontal">
                 <CheckBox x:Name="UsePreviewSdk" x:Uid="General_UsePreviewSdk" Content="{x:Static r:GeneralOptionPageResources.General_UsePreviewSdk}" Margin="13, 10" />
             </StackPanel>
+            <StackPanel Orientation="Horizontal">
+                <CheckBox x:Name="DoubleClickOpenProjectFile" x:Uid="General_DoubleClickOpenProjectFile" Content="{x:Static r:GeneralOptionPageResources.General_DoubleClickOpenProjectFile}" Margin="13, 10" />
+            </StackPanel>
```
drewnoakes: Stack panels with a single child aren't necessary.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptionPageResources.resx
davidwengier: My WPF is weak... pointers?
image
image

drewnoakes: Use a single underscore instead of an ampersand.

davidwengier: WPF is weird. Bring back Windows Forms I say!

Thanks guys :P

drewnoakes: `&` has a special meaning in XML. You'd have to use `&amp;`.

davkean:
Given https://dev.azure.com/devdiv/DevDiv/_workitems/edit/779596?src=WorkItemMention&src-action=artifact_link, and the more I think about this, I'm thinking we might want to push this setting along with the Preview settings instead of shipping the org chart here.

davidwengier:
Hadn't seen that. Makes the UI tricky, as a setting for previewing .NET SDK project files would apply to all three of the current preview window sub-settings, and I would have thought we'd still want a separate setting for the double click behaviour.

image

davkean:
I'd say that it's only relevant for Solution Explorer, though I'd need to ask the customer what exactly bothers them about it.

If you wanted to switch both, it's weird that one would be under .NET Core and one under that.

davidwengier:
For sure the Tabs and Windows section is a better spot for both settings, just not sure what layout would make sense.

davidwengier:
Renamed the setting based on feedback from the UX meeting today, and changed a couple of other strings even though they're not used, because they might be once #4561 is done.

davidwengier:
UX board has recommended against adding an option to control this behaviour.

PR #4509, opened 2019-01-30
don't use Task.Run in the project system

jmarolf:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/SDKVersionTelemetryServiceComponent.SDKVersionTelemetryServiceInstance.cs
```diff
@@ -35,7 +35,7 @@ internal class SDKVersionTelemetryServiceInstance : OnceInitializedOnceDisposedA
             protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
             {
                 // Do not block initialization on reporting the sdk version. It is possible to deadlock.
-                Task.Run(async () =>
+                _projectVsServices.ThreadingService.Fork(async () =>
                 {
                     // Wait for the project to be loaded so that we don't prematurely load the active configuration
                     await _unconfiguredProjectTasksService.ProjectLoadedInHost;
```
davkean: You can remove usage of LoadedProjectAsync as they project is guaranteed to stay open while this task is running.

davkean: In fact, going to ban Task.Run for exactly this reason.

jmarolf: I couldn't find any other uses in the code base but we should definitely ban it so new uses don't accidentally creep in.

davkean: Working on it...

PR #4510, opened 2019-01-30
Add LoadProjectAsync to the banned symbol list

davkean:


src/BannedSymbols.txt
```diff
@@ -6,6 +6,8 @@ T:Microsoft.VisualStudio.ProjectSystem.VS.IProjectGuidService2; Using ISafeProje
 P:Microsoft.VisualStudio.ProjectSystem.VS.IProjectAsyncLoadDashboard.ProjectLoadedInHost; Using IUnconfiguredProjectTasksService.ProjectLoadedInHost prevents waiting indefinitely when the project is closed before it has finished loading
 
 M:Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProject(Microsoft.VisualStudio.ProjectSystem.IProjectAsynchronousTasksService); Using IUnconfiguredProjectTasksService.LoadedProjectAsync is unit testable
+M:Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProjectAsync(Microsoft.VisualStudio.ProjectSystem.IProjectAsynchronousTasksService,System.Func{System.Threading.Tasks.Task},System.Boolean); Using IUnconfiguredProjectTasksService.LoadedProjectAsync is unit testable
+M:Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.LoadedProjectAsync`1(Microsoft.VisualStudio.ProjectSystem.IProjectAsynchronousTasksService,System.Func{System.Threading.Tasks.Task{``0}},System.Boolean); Using IUnconfiguredProjectTasksService.LoadedProjectAsync is unit testable
```
drewnoakes: Is it worth mentioning `UnconfiguredProjectTasksService` as a replacement in the messages? Also on line 8.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
```diff
@@ -186,11 +186,6 @@ private IReadOnlyCollection<string> GetWatchedRules(RuleHandlerType handlerType)
 
             await _tasksService.LoadedProjectAsync(async () =>
             {
-                if (_tasksService.UnloadCancellationToken.IsCancellationRequested)
-                {
-                    return;
```
drewnoakes: Was this doing anything useful? Doesn't `LoadedProjectAsync` prevent unloading during the callback anyway?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
```diff
@@ -108,8 +108,6 @@ private async Task OnProjectChangedAsync(Tuple<IProjectSubscriptionUpdate, IProj
 
             await _tasksService.LoadedProjectAsync(() =>
             {
-                _tasksService.UnloadCancellationToken.ThrowIfCancellationRequested();
```
drewnoakes: Ditto.

src/BannedSymbols.txt
davkean: UnconfiguredProjectTasksService is the implementation, but you import IUnconfiguredProjectTasksService.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
davkean: This line is effectively what LoadProjectAsync does before calling this callback, it won't start the callback if this is true.

PR #4514, opened 2019-01-30
Tidying

drewnoakes:
Some small commits pulled out of #4513 for easier review.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
@@ -164,6 +164,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
         public string Name { get; }
         public string OriginalItemSpec { get; }
         public string Path { get; }
+
         public string FullPath
         {
             get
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs
```diff
@@ -60,9 +60,20 @@ public Dependency(IDependencyModel dependencyModel, ITargetFramework targetFrame
             // This is needed for tree update logic to track if tree node changing state from unresolved 
             // to resolved or vice-versa (it helps to decide if we need to remove it or update in-place
             // in the tree to avoid flicks).
-            Flags += Resolved
-                ? DependencyTreeFlags.ResolvedFlags
-                : DependencyTreeFlags.UnresolvedFlags;
+            if (Resolved)
```
davkean: Why? This is much easier to read before.

drewnoakes: Described in d8245f4d0f21092d1b43be851d11c0558b87e2f9:

> The += operator will always allocate, even if no change occurs. This longer form can avoid such allocations.
>
> This check is only to assist 3rd party dependency node providers that may omit this step. The in-box providers should never need this code and so should not be paying any overhead.

PR #4480, opened 2019-01-18
LOC CHECKIN | dotnet/project-system master | 20190118

v-zbsail:

PR #4516, opened 2019-01-30
LOC CHECKIN | dotnet/project-system dev16.0.x | 20190130

tmeschter:
This is a cleaned-up version of https://github.com/dotnet/project-system/pull/4480. We needed those change in the dev16.0.x branch rather than master, but the commit history has become pretty messy. Further, we really only need the final result, not the intermediate commits, and there were still some things that needed to be cleaned up. Namely, some files had double BOMs at the beginning, which prevented us from reading them as XML.

That PR will be closed in favor of this one.

This isn't perfect, but the translation state is a big improvement.


tmeschter:
Now to get QB approval.

PR #4520, opened 2019-01-31
Fix changed project references not causing rebuild

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/781913?src=WorkItemMention&src-action=artifact_link.

This was accidentally removed in 0ae2a36f.


davkean:
@jjmew This is a regression and needs to be be a part of Preview 3.

jmarolf:
We _really_ like this PR @davkean

Pilchie:
This has been approved - we need to get an insertion going for it now. @jmarolf can you help with that?

jmarolf:
Do we need a dummy pr or if it's already approved can we merge and use the real pr?

davkean:
I've merged just use the real PR.

jmarolf:
Yep confirmed with Kevin.  I'll tag Doug to approve it once its created

PR #4529, opened 2019-01-31
Little bits

drewnoakes:
Small commits pulled from larger work for separate review.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -339,14 +339,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
 
                             _dependenciesSnapshotProvider.SnapshotChanged += OnDependenciesSnapshotChanged;
 
-                            Task<IProjectVersionedValue<IProjectTreeSnapshot>> nowait = SubmitTreeUpdateAsync(
-                                (treeSnapshot, configuredProjectExports, cancellationToken) =>
+                            _ = SubmitTreeUpdateAsync(
+                                delegate
                                 {
                                     IProjectTree dependenciesNode = CreateDependenciesFolder();
-
-                                    // TODO create providers nodes that can be visible when empty
-                                    //dependenciesNode = CreateOrUpdateSubTreeProviderNodes(dependenciesNode, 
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -339,15 +339,11 @@ async delegate
 
                             _dependenciesSnapshotProvider.SnapshotChanged += OnDependenciesSnapshotChanged;
 
-                            Task<IProjectVersionedValue<IProjectTreeSnapshot>> nowait = SubmitTreeUpdateAsync(
-                                (treeSnapshot, configuredProjectExports, cancellationToken) =>
+                            _ = SubmitTreeUpdateAsync(
+                                delegate
```
jmarolf: The code was previously like this so its not a problem with this PR but I have questions...

1. Why are we calling a fire-and-forget async  task inside of a lock and not awaiting it? Does it need to be in this lock?
2. Fire and forget tasks are dangerous because they can be run after the project us unloaded, causing them to crash VS.  Could we use the `Fork` method from JTF so we cancel the work if the project is unloaded?

jmarolf: Scratch my second comment I scrolled up and saw that we are using `UnconfiguredProjectAsynchronousTasksService.LoadedProjectAsync`.  But I am still curious if we want to block project unload on this, wouldn't we just want to cancel the work instead and let the user unload the project?

drewnoakes: In general I agree that fire and forget is a bad idea. There is another instance of this a little lower in the file too around line 420. We also have two calls to `TplExtensions.Forget` (which does nothing with the task at all) and one call to `VsTaskLibraryHelper.FileAndForget` (which handles faults and reports them via telemetry).

I'll create an issue as I don't want to go down this rabbit hole right now.

drewnoakes: #4530

jmarolf: :+1:

PR #4531, opened 2019-02-01
Format all rule files

davkean:
Formats all rule files against settings in .editorconfig.


davidwengier:
Could be GitHub display issues, but it looks like `COMReference.xaml` has been added, but no deletion of `ComReference.xaml`. Additionally do we need to rename all of the .xlf files to pick up the case change?

davkean:
Ugh, no this looks like some sort of bug - I thought I had removed this but VS is picking it up as COMReference, and if I save that - causes above. Need to investigate.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/UpToDateCheckBuilt.xaml
```diff
@@ -1,10 +1,11 @@
...
-    </Rule.DataSource>
-    <BoolProperty Name="Visible" Visible="False" ReadOnly="True" />
-    <StringProperty Name="Original" />    <!-- If set, specifies that this built output comes from this source file. -->
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFile" HasConfigurationCondition="False" SourceOfDefaultValue="AfterContext"
+                SourceType="TargetResults" MSBuildTarget="CollectUpToDateCheckBuiltDesignTime"/>
+  </Rule.DataSource>
+  <BoolProperty Name="Visible" Visible="False" ReadOnly="True" />
+  <StringProperty Name="Original" />
+  <!-- If set, specifies that this built output comes from this source file. -->
```
sharwell: 💡 This comment should be placed before the property

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/COMReference.xaml
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Rule Name="ComReference" DisplayName="COM Reference" PageTemplate="generic" Description="COM Reference Properties" xmlns="http://schemas.microsoft.com/build/2009/properties">
```
sharwell: ❔ Where did this come from?

PR #4535, opened 2019-02-04
Adding F# OtherFlags to GeneralConfiguredBrowseObject

TIHan:
This allows the "Other flags" text field in the F# property page to be editable and not grayed out in SDK style projects.

otherflags

On the F# side, we still have a bug to fix the implicit flags that get added; this is a visualfsharp bug, not project-system.

// cc @KevinRansom


davkean:
Yes, long term we'll split these out into F# specific rules.

davkean:
@TIHan What's blocking this?

TIHan:
Nothing that I'm aware. I can merge it in.

PR #4548, opened 2019-02-07
Serialise dependencies node data updates

drewnoakes:
The dependencies node currently receives both evaluation and design time data updates via DataFlow from CPS, each from its own block.

A single block is guaranteed to not overlap async data callbacks, but because we have two we require synchronisation between them.

**NOTE** This is not urgent and can wait to be merged, or changed to target a different branch.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
@@ -161,12 +161,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscription
                     OnActiveConfiguredProjectEvaluatedAsync);
 
                 // Each of the host's subscribers are initialized.
-                foreach (IDependencyCrossTargetSubscriber subscriber in Subscribers)
-                {
-                    subscriber.InitializeSubscriber(this, _activeConfiguredProjectSubscriptionService);
-                }
-
-                return Task.CompletedTask;
+                return Task.WhenAll(
+                    Subscribers.Select(
+                        subscriber => subscriber.InitializeSubscriberAsync(this, _activeConfiguredProjectSubscriptionService)));
             }
 
             async Task OnActiveConfiguredProjectEvaluatedAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> e)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
@@ -5,20 +5,24 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
 
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.ProjectSystem.Utilities;
 using Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget;
+using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions
 {
     [Export(typeof(IDependencyCrossTargetSubscriber))]
     [AppliesTo(ProjectCapability.DependenciesTree)]
-    internal sealed class DependencyRulesSubscriber : OnceInitializedOnceDisposed, IDependencyCrossTargetSubscriber
+    internal sealed class DependencyRulesSubscriber : OnceInitializedOnceDisposedAsync, IDependencyCrossTargetSubscriber
     {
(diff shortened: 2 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -161,12 +161,9 @@ Task AddInitialSubscriptionsAsync()
...
                 // Each of the host's subscribers are initialized.
-                foreach (IDependencyCrossTargetSubscriber subscriber in Subscribers)
-                {
-                    subscriber.InitializeSubscriber(this, _activeConfiguredProjectSubscriptionService);
-                }
-
-                return Task.CompletedTask;
+                return Task.WhenAll(
+                    Subscribers.Select(
+                        subscriber => subscriber.InitializeSubscriberAsync(this, _activeConfiguredProjectSubscriptionService)));
```
drewnoakes: This change initialises subscribers concurrently. I've verified that's ok to do.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
```diff
@@ -184,12 +188,18 @@ private IReadOnlyCollection<string> GetWatchedRules(RuleHandlerType handlerType)
                 return;
             }
 
-            await _tasksService.LoadedProjectAsync(async () =>
+            // Ensure updates don't overlap
+            await _updateGate.ExecuteWithinLockAsync(JoinableCollection, JoinableFactory, async () =>
```
drewnoakes: All the other changes in this file are to support this line.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
```diff
@@ -184,12 +188,18 @@ private IReadOnlyCollection<string> GetWatchedRules(RuleHandlerType handlerType)
...
             {
-                using (ProjectCapabilitiesContext.CreateIsolatedContext(configuredProject, capabilities))
+                // Ensure the project doesn't unload during the update
+                await _tasksService.LoadedProjectAsync(async () =>
```
jmarolf: Do we actually want to block project unload on this or just cancel this operation if the project is unloaded?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
```diff
@@ -256,16 +266,18 @@ private IReadOnlyCollection<string> GetWatchedRules(RuleHandlerType handlerType)
...
         }
 
-        protected override void Initialize()
+        protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
         {
+            return Task.CompletedTask;
         }
 
-        protected override void Dispose(bool disposing)
+        protected override Task DisposeCoreAsync(bool initialized)
```
davkean: This can overlap with with your ExecuteWithinAsync, i'd consider deriving from OnceInitializedOnceDisposedUnderLockAsync which prevents thats.

davkean: This also reduces the JoinableTaskFactory overhead in ExecuteWithinLockAsync which was added to workaround a bug in VS Threading.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
davkean: Yes, very likely it is not resilient against unload - though probably want to eventually pass through CancellationToken to bail earlier.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
davkean: And then we can delete our usage of SemaphoreSlim and ban it in lieu of the new ReentrantSemaphore.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
drewnoakes: Added a TODO in fa9b08210.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
drewnoakes: @davkean I believe this is now covered in 11a3c28e1. Please take another look.

The work I'm doing on #4550 is easier to debug with this PR's changes, so I'm basing it off this branch.

Issue #4530, opened 2019-01-31
Review fire-and-forget asynchrony


drewnoakes:
As highlighted by @jmarolf in https://github.com/dotnet/project-system/pull/4529#discussion_r252798620, we have some async operations which fire-and-forget off into the ether. We should review whether these are subject to potential problems in case of project closure or task faults.

- There are two discarded tasks in `DependenciesProjectTreeProvider`
~- We have two calls to `TplExtensions.Forget` (which does nothing with the task at all)~
~- We have one call to `VsTaskLibraryHelper.FileAndForget` (which handles faults and reports them via telemetry)~
- Possibly more cases elsewhere

davkean:
Fire and forgets were banned here: https://github.com/dotnet/project-system/pull/4554. Discarded tasks were not reviewed, but should call IProjectFaultHandlerService.RegisterFaultHandler2 (or whatever the names ends up in the previous PR).

PR #4554, opened 2019-02-08
Add replacements for Watson reporting APIs

davkean:
Replacements for Watson reporting APIs across the project-system that end up:

- Ensuring that they are bucked under our component and under the new DevDiv watson event name
- Remove the need to pass a JoinableTaskFactory
- Require that a project is always specified even if null, so that the error message includes the project at fault
- Replaces all usages of Forget and FireAndForget


davkean:
@drewnoakes Did you ask about FireAndForget the other day? I've banned all of those APIs, and all should go through RegisterFaultHandler.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="unconfiguredProject">
+        ///     The unconfigured project which the delegate operates on, if applicable. Can be <see langword="null"/>.
+        /// </param>
+        /// <param name="faultSeverity">
+        ///     Suggests to the user how severe the fault is if the delegate throws.
+        /// </param>
+        /// <param name="options">
+        ///     Influences the environment in which the delegate is executed.
+        /// </param>
+        public static void Fork2(this IProjectThreadingService threadingService, Func<Task> asyncAction, UnconfiguredProject unconfiguredProject, ProjectFaultSeverity faultSeverity = ProjectFaultSeverity.Recoverable, ForkOptions options = ForkOptions.Default)
```
jmarolf: Could we have a more descriptive name than `Fork2`?

jmarolf: Maybe `ForkWithDefaultFaultHander` or at least `SafeFork`?

davkean: Sure, what do you think? (TBH, Fork2 is an obvious replacement for Fork)

src/BannedSymbols.txt
```diff
@@ -17,11 +17,21 @@ M:System.Threading.Tasks.Task`1.ConfigureAwait(System.Boolean); "ConfigureAwait(
...
+M:System.Threading.Tasks.Task.Run`1(System.Func`1{``0},System.Threading.CancellationToken); Task.Run should be replaced with IProjectThreadingService.Fork2, which prevents the project being unloaded while the Task is running and ensures that failed tasks are reported as non-fatal watsons
+M:System.Threading.Tasks.Task.Run(System.Func{System.Threading.Tasks.Task},System.Threading.CancellationToken); Task.Run should be replaced with IProjectThreadingService.Fork2, which prevents the project being unloaded while the Task is running and ensures that failed tasks are reported as non-fatal watsons
+M:System.Threading.Tasks.Task.Run`1(System.Func{System.Threading.Tasks.Task{``0}},System.Threading.CancellationToken); Task.Run should be replaced with IProjectThreadingService.Fork2, which prevents the project being unloaded while the Task is running and ensures that failed tasks are reported as non-fatal watsons
+
+M:Microsoft.Threading.Tasks.TplExtensions.Forgot(System.Threading.Tasks.Task); Use IProjectFaultHandlerService.RegisterFaultHandler2 to ensure that failed tasks are reported as non-fatal watsons and assigned to the correct component
```
davkean: This is Microsoft.VisualStudio.Threading.TplExtensions.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
jmarolf: `ForkWithDefaultReportSettings` would be most descriptive which I am partial to, but if you really like `Fork2` for its terseness I can live with it.

src/BannedSymbols.txt
drewnoakes:
```suggestion
M:Microsoft.Threading.Tasks.TplExtensions.Forget(System.Threading.Tasks.Task); Use IProjectFaultHandlerService.RegisterFaultHandler2 to ensure that failed tasks are reported as non-fatal watsons and assigned to the correct component
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
drewnoakes: `Fork_final`?

src/BannedSymbols.txt
drewnoakes: i.e. it's Forget, not Forgot

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
davkean: What does the _final_ mean?

drewnoakes:
The fire and forget issue was in https://github.com/dotnet/project-system/issues/4530.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
drewnoakes: I refuse to explain my joke other than to say `Fork_final2` has a nice ring to it.

davkean: Yeah I can play around with that name @jmarolf.

drewnoakes: The numerical suffix thing on interfaces seemed weird to me when I joined MS but I understand it now as meaning the interface has been extended through versioning and derives from N-1. That's not the case here. `Fork` is already a rebranding of `Run` if I understand correctly. How about `RunWithErrorReporting` or `ForkWithErrorReporting`?

davkean: I did consider Run, but I thought that was confusing with `JoinableTaskFactory.Run` (which is equivalent to `IProjectThreadingService.ExecuteSynchronously`). This is really; 'RunAndForget'

davidwengier: I like `RunAndForget`. Neither `Fork` nor `ForkWithDefaultReportSettings` make it obvious to me that its forgetting, though I guess the void return helps.

davkean: How about this;

Fork2 -> RunAndForget
RegisterFaultHandler2 -> Forget.

davkean: Mind you Fork is more than JoinableTaskFactory.Run because it will switch threads and hide locks, which is why it was probably named as such.

davkean: I've fixed the names.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="task">
+        ///     The task to attach error handling to.
+        /// </param>
+        /// <param name="project">
+        ///     The project related to the failure, if applicable. Can be <see langword="null"/>.
+        /// </param>
+        /// <param name="severity">
+        ///     The severity of the failure.
+        /// </param>
+        public static void Forget(this IProjectFaultHandlerService faultHandlerService, Task task, UnconfiguredProject project, ProjectFaultSeverity severity = ProjectFaultSeverity.Recoverable)
```
drewnoakes: Consider having `Forget` take `Func<Task>` so that exceptions thrown before yielding receive the same treatment as those thrown in continuations.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="task">
+        ///     The task to attach error handling to.
+        /// </param>
+        /// <param name="project">
+        ///     The project related to the failure, if applicable. Can be <see langword="null"/>.
+        /// </param>
+        /// <param name="severity">
+        ///     The severity of the failure.
+        /// </param>
+        public static void Forget<TResult>(this IProjectFaultHandlerService faultHandlerService, Task<TResult> task, UnconfiguredProject project, ProjectFaultSeverity severity = ProjectFaultSeverity.Recoverable)
```
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectFaultServiceExtensions.cs
davkean: I'm wrapping FaultHandlerService here, I don't control over this without reimplementing.

PR #4555, opened 2019-02-08
Opt out of banned symbols in AppDesigner/Editors

davkean:
Adding additional APIs to the BannedSymbols in the future that are very specific to the project system such as IServiceProvider, just opt out to avoid us having to suppress all usage.


davkean:
Basically all usage of banned symbols in the list are already very specific to the project system.

PR #4556, opened 2019-02-08
Ban IServiceProvider/IAsyncServiceProvider usage

davkean:
This includes https://github.com/dotnet/project-system/pull/4555 so only review https://github.com/dotnet/project-system/pull/4556/commits/9e5672dafca8f8518be1e044571ce8f3da43ea01.

Ban IServiceProvider/IAsyncServiceProvider usage

IVsUIService/IVsService should be used instead. Existing usages are baselined and being tracked by #4375.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/Xproj/GlobalJsonRemover.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/Xproj/GlobalJsonRemover.cs
@@ -74,8 +74,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Xproj
         public int OnAfterOpenSolution(object pUnkReserved, int fNewSolution)
         {
             UIThreadHelper.VerifyOnUIThread();
+#pragma warning disable RS0030 // Do not used banned APIs
             DTE2 dte = _serviceProvider.GetService<DTE2, DTE>();
+#pragma warning restore RS0030 // Do not used banned APIs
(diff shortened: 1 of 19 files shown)


src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/Xproj/GlobalJsonRemover.cs
```diff
@@ -74,8 +74,10 @@ internal GlobalJsonRemover(IServiceProvider serviceProvider, IFileSystem fileSys
         public int OnAfterOpenSolution(object pUnkReserved, int fNewSolution)
         {
             UIThreadHelper.VerifyOnUIThread();
+#pragma warning disable RS0030 // Do not used banned APIs
             DTE2 dte = _serviceProvider.GetService<DTE2, DTE>();
```
jmarolf: Do we want to eventually refactor this someday or do we feel this usage is acceptable?

jmarolf: I guess I am actually asking: do we want to file bugs for some of these cases to refactor them later?

davkean: Tracked by https://github.com/dotnet/project-system/issues/4375, usually I'd baseline them in GlobalSupressions but just following existing usage in that bug.

PR #4563, opened 2019-02-12
Fix hang in GetWorkspaceContextIdFromActiveContext

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/782754.

Avoid blocking on initialization of the active workspace context just to get the default workspace context id for a multi-targeting projects it hasn't been set. Instead, just return the first one that is registered. This returns the behavior to what it was previously before the new language-service integration was checked in. 

This means that there is a chance that it may return no longer the "active configuration" context but instead the first one to register itself - but practically won't occur because we block on the active config registration in WorkspaceProjectContextHostInitiator.

Due to a flaw in the new integration, there's timing bug where IActiveWorkspaceProjectContext.OpenContextForWrite can hang (or never complete) if it is called on a configuration that previously was, but behind the scenes is no longer implicitly active. This will be fixed by https://github.com/dotnet/project-system/issues/3425, but in the meantime just avoid calling in this case.

To simplify this, I've:

- Remove the need to store the actual context in the tracker, we only use the "id"
- Replaced the hash table with a list so that we have guaranteed order (and therefore can return the first one registered). We only have a small number of elements anyway (1 per TFM in a project), so lookup should not be a concern.

@jjmew Like to take this to shiproom for Preview 4, will fill in ask mode tomorrow after I've finished testing it.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ActiveEditorContextTracker.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ActiveEditorContextTracker.cs
@@ -3,9 +3,9 @@
 using System;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
+using System.Linq;
 using System.Threading.Tasks;
 
-using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
 using Microsoft.VisualStudio.TextManager.Interop;
 using Microsoft.VisualStudio.Threading;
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ActiveEditorContextTracker.cs
```diff
@@ -56,57 +52,42 @@ public int GetProjectName(uint itemid, out string pbstrProjectName)
...
         {
-            Requires.NotNull(context, nameof(context));
+            Requires.NotNullOrEmpty(contextId, nameof(contextId));
 
-            if (_contexts.TryGetValue(context, out string projectContextId))
-            {
-                return StringComparers.WorkspaceProjectContextIds.Equals(ActiveIntellisenseProjectContext, projectContextId);
-            }
+            if (!_contexts.Contains(contextId))
+                throw new InvalidOperationException($"'nameof{contextId}' has not been registered or has already been unregistered");
```
davidwengier: Nit: This looks odd... is nameof not supposed to be there, supposed be inside the braces, or both?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ActiveEditorContextTracker.cs
```diff
@@ -56,57 +52,42 @@ public int GetProjectName(uint itemid, out string pbstrProjectName)
...
+                {
+                    projectContexts = projectContexts.Add(contextId);
+                }
+
+                return projectContexts;
+            });
 
             if (!changed)
-                throw new InvalidOperationException("'context' has already been registered.");
+                throw new InvalidOperationException($"'{nameof(contextId)}' has already been registered.");
```
davidwengier: Based on this, looks like for my above comment nameof should be inside the braces, but perhaps the value of contextId help diagnosis too?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ActiveEditorContextTracker.cs
```diff
@@ -25,21 +25,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
...
+        public ActiveEditorContextTracker()
         {
-            _threadingService = threadingService;
-            _activeWorkspaceProjectContextHost = activeWorkspaceProjectContextHost;
         }
 
         public string ActiveIntellisenseProjectContext
         {
-            get { return _activeIntellisenseProjectContext ?? GetWorkspaceContextIdFromActiveContext(); }
+            get { return _activeIntellisenseProjectContext ?? _contexts.FirstOrDefault(); }
```
davidwengier: Might not make a difference or be worth changing, but `FirstOrDefault` could have just been used on the `Values` property of the ImmutableDictionary

davkean: ImmutableDictionary doesn't have order semantics, meaning that it would no longer be the first one registered.

davidwengier: Makes sense, thanks

davkean:
@jjmew This is ready for shiproom, filled out ask mode template.

PR #4568, opened 2019-02-14
Add accessible help text to the debug property page

etbyrd:
Fixes for [785764](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/785764/) and [786001](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/786001/)

Adds AutomationProperties.HelpText to 5 items in the debug property page and the new profile dialog. These strings are read by screen readers to give a better description of what will actually happen when the buttons are used - especially when the result is a UI only change. The strings were just first intuitive strings I thought of - I would appreciate any input @cartermp @AbhitejJohn 

New Button (Debug Property Page) - Open the new profile dialog
Delete Button (Debug Property Page) - Delete the current profile and change to the next available profile
Launch Browser Checkbox (Debug Property Page)  - Checking enables a textbox to specify an absolute or relative URL
OK Button (New Profile Dialog) - Create the profile and return to the debug property page with the new profile selected
Cancel Button (New Profile Dialog) - Return to the debug property page with no changes


etbyrd:
@Pilchie  This will be for 16.1, which is master, right? Also, what documentation are you referring to?

PR #4570, opened 2019-02-15
Remove unnecessary rules

davidwengier:
These rules are the same as the None rule so aren't necessary to have separate XAML files for, as they'll be picked up anyway now that we support AvailableItemName.

Note that I don't 100% trust my local testing of these (manually deleting XAML files in my VS install) so if I've missed something let me know.

I separated things into individual commits to make reviewing easier btw.


davidwengier:
@jmarolf is there a not-too-difficult way to get a clean VS build with these changes?

davkean:
Can you make sure they are actually picked up due to AvailableItemName? ie Diff the Build Action drop down, both CLI SDKs versions and target frameworks might affect this.

davkean:
(I saw that because I only 4 available item names when I was debugging this the other day)

davidwengier:
You can see when something is pulled in by AvailableItemName as CPS indicates its a generated rule:
image

I tested locally and was able to select each one of these removed item types in a .NET Core WPF app and a .NET Core console app, but I can certainly try more, I'm just not confident that the F5 experience/experimental hive is doing a representing the real world accurately here.

davkean:
Is that text ("Rule generated for") deliberate?

.NET SDK version can affect this, so you'll need to opt in via global.js, or uninstall the SDK. I would like to try 2.2 and 3.0 at least. Also please try targeting .NET Framework.

davidwengier:
> Is that text ("Rule generated for") deliberate?

That's what [CPS inserts](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.Implementation%2FPropertyPages%2FPropertyPagesCatalog.cs&version=GBteam%2Fv16&line=517&lineStyle=plain&lineEnd=518&lineStartColumn=1&lineEndColumn=1) when it creates a rule that doesn't physically exist.

> .NET SDK version can affect this, so you'll need to opt in via global.js, or uninstall the SDK. I would like to try 2.2 and 3.0 at least. Also please try targeting .NET Framework.

👍

davkean:
Going to PR a better description for that.

davidwengier:
Closing this for now, needs more work. Because we define <ItemType> nodes for these, this makes the experience worse as CPS now adds item types for things it has no rules for (if the AvailableItemName's aren't in scope).

PR #4574, opened 2019-02-15
Synchronise up-to-date check query and project updates

drewnoakes:
Previously it was possible for an update to occur during the up to date computation. Its collections are not thread safe so could report incorrect values or even raise exceptions.

This commit makes the following changes:

- Derive from `OnceInitializedOnceDisposedUnderLockAsync`
- Export `IProjectDynamicLoadComponent` to respond to project capabilites
- Perform update and query within `ExecuteUnderLockAsync` for isolation

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -108,7 +108,32 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
 
         public Task UnloadAsync()
         {
-            return Task.CompletedTask;
+            return ExecuteUnderLockAsync(_ =>
+            {
+                _link?.Dispose();
+                _link = null;
+
+                _lastCheckTimeUtc = DateTime.MinValue;
+                _isDisabled = true;
+                _itemsChangedSinceLastCheck = true;
+                _msBuildProjectFullPath = null;
+                _msBuildProjectDirectory = null;
+                _markerFile = null;
+                _outputRelativeOrFullPath = null;
+                _newestImportInput = null;
+
+                _itemTypes.Clear();
+                _items.Clear();
+                _customInputs.Clear();
+                _customOutputs.Clear();
+                _builtOutputs.Clear();
+                _copiedOutputFiles.Clear();
+                _analyzerReferences.Clear();
+                _compilationReferences.Clear();
+                _copyReferenceInputs.Clear();
+
+                return Task.CompletedTask;
+            });
         }
 
         protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -20,8 +20,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
 {
     [AppliesTo(ProjectCapability.DotNet + "+ !" + ProjectCapabilities.SharedAssetsProject)]
     [Export(typeof(IBuildUpToDateCheckProvider))]
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
     [ExportMetadata("BeforeDrainCriticalTasks", true)]
-    internal sealed class BuildUpToDateCheck : OnceInitializedOnceDisposed, IBuildUpToDateCheckProvider
+    internal sealed class BuildUpToDateCheck : OnceInitializedOnceDisposedUnderLockAsync, IBuildUpToDateCheckProvider, IProjectDynamicLoadComponent
     {
         private const string CopyToOutputDirectory = "CopyToOutputDirectory";
         private const string PreserveNewest = "PreserveNewest";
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -98,26 +101,33 @@ internal sealed class BuildUpToDateCheck : OnceInitializedOnceDisposed, IBuildUp
...
-        [ConfiguredProjectAutoLoad]
-#pragma warning restore RS0030 // symbol ConfiguredProjectAutoLoad is banned
-        [AppliesTo(ProjectCapability.DotNet + "+ !" + ProjectCapabilities.SharedAssetsProject)]
-        internal void Load()
+        public Task LoadAsync()
+        {
+            return InitializeAsync();
+        }
+
+        public Task UnloadAsync()
```
jmarolf: What does it mean that this component doesn't unload itself? Why make it a `IProjectDynamicLoadComponent` if unloads do nothing?

davkean: Agreed. This is also unrelated to the thing you are fixing. AutoLoad can be async, just have it return Task.

drewnoakes: Ah, I didn't know `[ConfiguredProjectAutoLoad]` would work with an async method. I'll back it out, although it probably does make sense to subscribe and unsubscribe from Dataflow in response to capabilities in future.

drewnoakes: @davkean are you sure `[ConfiguredProjectAutoLoad]` can return a task? It doesn't seem to work for me.

For now I'll make `UnloadAsync` clean up properly.

drewnoakes:
I wonder if deferring the Dataflow subscription until the first build would make sense (rather than when the project loads). We always report the first build as not up to date.

`BuildUpToDateCheck` is created per configuration. Most folks will only use a single configuration throughout a given VS session. So if you have two configurations, that's half the number of subscriptions, and half the work at solution load time.

//cc: @panopticoncentral

davkean:
> We always report the first build as not up to date. 

Why is the first build out of date?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
davkean: MB, looks like only '[ProjectAutoLoad]' supports this not '[ConfiguredProjectAutoLoad]'

drewnoakes:
> Why is the first build always out of date?

I assume because we can't know if the set of input items changed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: Np. Take a look at the updated implementation and let me know if you see any issues.

Pilchie:
That was presumably true before we cached designtime build results between solution loads, but now that we do, we should be able to tell, right?

davkean:
Even without caching between loads - after the first design-time build has finished, and the project has built at least once - why would it be considered out of date? We've gathered all inputs at that point, right?

drewnoakes:
Let's continue the discussion about the first build being not up-to-date in #4585, as it's unrelated to this PR.

I think this is good to go in. Can it get a review?

davkean:
It looks good, I would have considered a different approach, where you didn't lock but instead produced a bunch of immutable data which you just published to a single field when you were done. 

> Most folks will only use a single configuration throughout a given VS session. So if you have two configurations, that's half the number of subscriptions, and half the work at solution load time.

If you don't switch configurations the config will never get loaded.

drewnoakes:
> I would have considered a different approach, where you didn't lock but instead produced a bunch of immutable data which you just published to a single field when you were done.

Essentially the innars of the lock do take an immutable prior and produce the N+1 version of it. The problem is that there are two triggers for this (evaluation and DTB) and they need to be processed sequentially without overlapping.

davkean:
Are we looking at different code? There's no overlap between evaluation and DTB in this class.

drewnoakes:
> Are we looking at different code?

Sorry. Late here and my brain's not working. Was thinking of a different PR... I'll look at this again tomorrow.

drewnoakes:
> I would have considered a different approach, where you didn't lock but instead produced a bunch of immutable data which you just published to a single field when you were done.

@davkean I've taken a look at this again and see what you mean. It sounds better in theory than it looks in practice. It would still likely require locking for correctness as the act of testing up-to-date-ness actually mutates state (it's not an idempotent operation, even if it should be).

Specifically the `_lastCheckTimeUtc` and `_itemsChangedSinceLastCheck` fields.

I will push the code around a bit and produce an immutable state object to see how it looks in another branch. It'll allocate more due to all the immutable sets/dictionaries. Let's see if it ends up being more readable.

davkean:
I'm okay with this change as is - don't feel like you need to go down that approach.

drewnoakes:
@davkean too late :) The first stab at making state immutable is at https://github.com/drewnoakes/project-system/commit/00000003b70ac119ebfae9fa2b01893fbf71b02a. Is that what you were referring to?

I will probably go ahead and merge this PR. I'll take another look at the above commit tomorrow and see how it looks with fresh eyes.

davidwengier:
> at [drewnoakes@0000000](https://github.com/drewnoakes/project-system/commit/00000003b70ac119ebfae9fa2b01893fbf71b02a)

Nice commit hash... you should buy a lottery ticket

drewnoakes:
Wow, twice in one day. What are the odds? https://github.com/drewnoakes/gitlucky/commit/0000000febe641f4658c0ec10d20fafe6dd528ea

PR #4584, opened 2019-02-19
Final loc changes for 16.0

tmeschter:
Bring in what should be the final set of loc changes for 16.0


Pilchie:
This was approved, so I've merged it.

davidwengier:
Late, but the stuff I was waiting for is here 👍

PR #4588, opened 2019-02-20
 Scope ActiveEditorContextTracker per project

davkean:
#4563 caused ActiveEditorContextTracker to be scoped across the solution instead of per project, resulting in the "default" context for a project to be the first one registered across the solution, instead for the project. This broke Razer for solutions with more than one project.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/795781.

tag @jjmew for approval.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/LanguageServices/ActiveEditorContextTrackerTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/LanguageServices/ActiveEditorContextTrackerTests.cs
@@ -236,7 +236,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
 
         private static ActiveEditorContextTracker CreateInstance()
         {
-            return new ActiveEditorContextTracker();
+            return new ActiveEditorContextTracker((UnconfiguredProject)null);
         }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IActiveEditorContextTracker.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IActiveEditorContextTracker.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
-
+using Microsoft.VisualStudio.Composition;
 using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IActiveEditorContextTracker.cs
```diff
@@ -20,6 +20,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
     ///         tracked via <see cref="IActiveWorkspaceProjectContextHost"/>.
     ///     </para>
     /// </remarks>
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
```
davkean: This is not enforced "yet" so it's just documenting what scope this is supposed to live. Coincidentally I was reviving a branch early this week I had with a set of tests that would have caught this and will respect this attribute.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/LanguageServices/ActiveEditorContextTrackerTests.cs
```diff
@@ -236,7 +236,7 @@ public void GetProjectName_WhenActiveIntellisenseProjectContextIdSetToNull_Retur
 
         private static ActiveEditorContextTracker CreateInstance()
         {
-            return new ActiveEditorContextTracker();
+            return new ActiveEditorContextTracker((UnconfiguredProject)null);
```
Pilchie: Could we add a test that uses two projects and verifies that they get different trackers?

jjmew:
Thanks for fixing this. I have 2 questions:

- What percentage of users get affected by this? Is it high?

-  How much testing has been done here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/LanguageServices/ActiveEditorContextTrackerTests.cs
davkean: Unlike CPS, we don't have integration tests at that level, we need their implementation assemblies to be able to correctly setup "scopes" that would mimic this bug.

davkean: Given our lack of VS tests right now, and for the test verification that I'm reviving above - I'll figure out a way to get access to their implementation binaries so that we can do this sort of thing.

Pilchie:
Proactively merged to get a val build going for validation.

jjmew:
@davkean  can you fill the Ask Mode template?

PR #4589, opened 2019-02-20
 Move IVsService to global scope

davkean:
This moves IVsService from ProjectService scope -> VS default container, which lets components in the tree outside of the project system (such as other global services) import it. For more information, see: https://github.com/Microsoft/VSProjectSystem/blob/master/doc/overview/scopes.md.


davkean:
IVsUIService will be forthcoming.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsService`1.cs
```diff
@@ -22,10 +23,10 @@ internal class VsService<T> : IVsService<T>
         private readonly AsyncLazy<T> _value;
 
         [ImportingConstructor]
-        public VsService([Import(typeof(SAsyncServiceProvider))]IAsyncServiceProvider serviceProvider, IProjectThreadingService threadingService)
+        public VsService([Import(typeof(SAsyncServiceProvider))]IAsyncServiceProvider serviceProvider, JoinableTaskContext joinableTaskContext)
```
drewnoakes: Why not import the factory here instead of the context? All usages want a factory.

davkean: Only the context is added to the MEF container.

PR #4593, opened 2019-02-21
Simplify IVsService<T>/IVsUIService<T>

davkean:
Fields don't need to restate the service type - it's only needed for the import.


davkean:
@drewnoakes FYI on above ^, just for future heads up.

PR #4596, opened 2019-02-21
Moved StartupProjectHelper to IVsUIService usage

davkean:
Moved StartupProjectHelper to IVsUIService usage

Removed null checks for services because if these two services do not exist, we shouldn't be running.

This is blocked on https://github.com/dotnet/project-system/pull/4592 merging.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/StartupProjectHelper.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/StartupProjectHelper.cs
@@ -4,6 +4,8 @@ using System;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
 
+using EnvDTE;
+
 using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.ProjectSystem.VS.Extensibility;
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/StartupProjectHelper.cs
```diff
@@ -33,29 +39,24 @@ public StartupProjectHelper([Import(typeof(SVsServiceProvider))] IServiceProvide
         /// </summary>
         public ImmutableArray<T> GetExportFromDotNetStartupProjects<T>(string capabilityMatch) where T : class
         {
-#pragma warning disable RS0030 // Do not used banned APIs
```
davidwengier: I know it didn't before, but should this switch to the main thread, just in case? Thats kinda the rule when using IVsUIService, no?

davkean: There's two rules;

- Either you are async and switch to the UI thread
- Or you are synchronous and throw if you are not the UI thread

This now does the latter by its usage of IVsUIService<T>. If I was writing this - I would have an explicit verification that I'm the UI thread, but I was just removing IServiceProvider usage.

Issue #4544, opened 2019-02-06
Investigate whether there is a benefit to merging dlls


davkean:
Visual Studio load/start up time is directly related to the number of dlls that are loaded on that path. The more dlls that are loaded, the slower the load.

We should investigate whether our managed scenarios are better off by merging .CSharp/.VisualBasic/.FSharp and .VS into a single dll or whether increasing the size of the dlls regresses performance.

drewnoakes:
From the [architecture](https://github.com/dotnet/project-system/blob/master/docs/repo/architecture.md) document:

> C# assemblies cannot depend on any F# or Visual Basic assemblies and vice versa.

If merged into a single assembly we lose an easy way to enforce this separation. An analyzer could achieve the same thing.

For example, given:

```c#
[assembly: IsolateNamespace("Microsoft.VisualStudio.ProjectSystem.LanguageServices.CSharp")]
[assembly: IsolateNamespace("Microsoft.VisualStudio.ProjectSystem.LanguageServices.FSharp")]
[assembly: IsolateNamespace("Microsoft.VisualStudio.ProjectSystem.LanguageServices.VisualBasic")]
```

An analyzer would enforce that code defined outside each of these namespaces could not reach into that namespace.

If this sounds useful I'll take a look at putting something together.

davkean:
It sounds very useful.

davkean:
This is done, left open to track modifying the architecture diagram.

drewnoakes:
A first draft of the analyzer is up at https://github.com/dotnet/roslyn-analyzers/pull/2161.

davkean:
This diagram has been updated.

PR #4597, opened 2019-02-21
Remove non-VS layer language-specific dlls

davkean:
This is part 1 of 2 of removing the language-specific dlls in lieu of two dlls; Microsoft.VisualStudio.ProjectSystem.Managed.dll and Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll

Fixes part of https://github.com/dotnet/project-system/issues/4544.

This removes the following dlls and their tests:

    Microsoft.VisualStudio.ProjectSystem.CSharp
    Microsoft.VisualStudio.ProjectSystem.VisualBasic
    Microsoft.VisualStudio.ProjectSystem.FSharp

Their code/tests have been moved into language-specific namespaces inside Microsoft.VisualStudio.ProjectSystem.Managed.

The VS layer is a little more involved and will happen in a follow up change.


drewnoakes:
https://github.com/dotnet/project-system/blob/master/docs/repo/architecture.md should be updated too.

davkean:
Yep, good point, I'll do it at end after all are deleted.

Issue #3817, opened 2018-08-01
Merge CSharp, VB, FSharp & Managed packages


davkean:
We have 4 packages; CSharpProjectSystemPackage, VisualBasicProjectSystemPackage, FSharpProjectSystemPackage and ManagedProjectSystemPackage for no real good reason. We should just merge them all in one, you don't need a package to register a pkgdef.

- [x] Delete CSharp
- [x] Delete Visual Basic
- [x] Delete F#

davkean:
There's a couple of things that are blocking this;

- MigrateXprojProjectFactory from C# package
- FSharpProjectSelector from F# package

These are small enough that we should just move them into Managed.VS and get rid of CSharpProjectSystemPackage, VisualBasicProjectSystemPackage and FSharpProjectSystemPackage.

davkean:
The last of this was done here: https://github.com/dotnet/project-system/pull/4613. We now have a single package.

PR #4603, opened 2019-02-22
Remove VisualBasicProjectSystemPackage

davkean:
Fixes part of https://github.com/dotnet/project-system/issues/3817.

This package was never used and was solely present to find two strings. Delete package and move registration to Managed.VS.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
@@ -1,34 +1,17 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-using System.Runtime.InteropServices;
-
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Packaging;
 using Microsoft.VisualStudio.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.VS;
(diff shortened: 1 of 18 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
```diff
@@ -1,34 +1,17 @@
...
 using Microsoft.VisualStudio.Packaging;
 using Microsoft.VisualStudio.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.VS;
-using Microsoft.VisualStudio.Shell;
 
-// Register ourselves as a CPS project type
+// Visual Basic
 [assembly: ProjectTypeRegistration(
     projectTypeGuid: ProjectType.VisualBasic,
     displayName: "#1",                      // "Visual Basic"
```
davidwengier: Can we make these nicer names, given that we'll have a bunch of them in the same assembly now? Or do they have to be positional?

davkean: We cannot, they must be integers.

davkean: The design is based on Win32 resources and managed resources were kinda hacked on top of it.

davkean:
Note this slightly changed; fixed the name of the "strings" - they were wrong and should been without a '#' prefix. Left the resources in a VSPackage.resx that I moved over, as VS package resources need to be named this.

PR #4604, opened 2019-02-22
Remove FSharpProjectSystemPackage

davkean:
Fixes part of #3817.

This package was used to register a selector which has been moved into Managed.VS.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -50,6 +53,12 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
         {
```
davkean: This method has become a bit of mess and will cleanup at after I've merged all these packages together.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
```diff
@@ -1,3 +1,6 @@
+Microsoft.VisualStudio.ProjectSystem.VS.FSharp.FSharpProjectSelector
```
davkean: It's unfortunate that these become public, but it's a requirement for COM visible objects. Maybe obsolete it to avoid consumption?

davkean: Actually, I stand corrected maybe only needed for IDispatch, I see MigrateXprojProjectFactory is internal and its almost identical to this.

Pilchie: Yes, I'm pretty sure `public` is only required for `IDispatch` interfaces.

PR #4617, opened 2019-02-27
Introduce component rules

davkean:
This is fork of the MEF component rules that have been implemented in CPS. I've rewritten them so that they valid a single thing at a time, instead of validating the entire tree at once. This results in each violation being a single test failure.

These tests enforces correct usage of CPS and our components, it validates the following:

- Imports must respect the capabilities of the components they import, ie if  component blindly imports something without checking capabilities but exports say they apply to something - then that import needs to filter, either via `OrderPrecedenceImportCollection` or in the case of singular, `Lazy<T, IAppliesToMetadataView>` and checking that against the current scope they are in.

- Exports from the same part must apply to the same capabilities (prevents you from running on startup under ".NET", but only exporting something under ".NET & LanguageService" or vice versa)

- Exports must be marked with AppliesTo if there's an import of that contract that is using IAppliesToMetadataView (such as `OrderPrecedenceImportCollection`, or `Lazy<T, IAppliesToMetadataView>`)

- Imports must respect the cardinality (`[Import]` vs `[ImportMany]`) of a contract

- Imports and exports must agree on the scope, ie can't import `ConfiguredProject` if exporting to the `UnconfiguredProject` scope

- Imports/Exports must import interfaces marked with [ProjectSystemContract] - this is the thing that makes a lot of the above rules work. 

This is a work in progress, especially given our components currently violate the last rules. The validation is also incomplete because we're missing CPS's implementation assemblies from the composition.

One question that I know will come up, "why aren't these analyzers?" The biggest reason is that tests use real "reflection" types via VS MEF's analysis.  

Theoretically we could sit VS MEF's composition on top of Roslyn types (via its MemberRef/TypeRef concepts) but it's a non-trivial amount of work. I want to get these tests in to stop the bleeding.


davkean:
@dotnet/project-system This is ready for review, [this PR](https://github.com/dotnet/project-system/pull/4627/files) fixes the test breaks.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentVerificationTests.cs
```diff
@@ -0,0 +1,292 @@
...
+using Xunit;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    public partial class ComponentVerificationTests
+    {
+        [Theory]
+        [ClassData(typeof(ComposedPartTestData))]
+        public void ImportsMustFilterBasedOnCapabilities(Type type)
+        {   // Imports should respect/filter the capabilities of the exports they receive
```
jmarolf: place this on its own line?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentVerificationTests.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            var part = ComponentComposition.Instance.FindComposedPart(type);
+
+            foreach (KeyValuePair<ImportDefinitionBinding, IReadOnlyList<ExportDefinitionBinding>> binding in part.SatisfyingExports)
```
drewnoakes: Seems like a nice candidate for destructuring.

davkean: 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentVerificationTests.cs
davkean: 👍

PR #4618, opened 2019-02-27
Move item rules into items folder

davidwengier:
Part of #4105 to clean up rules: Moving rules about item types into an Items subdirectory for organizational purposes.

Doing this in pieces to make things easier to review (hopefully)


davkean:
Does deployment and setup just work with this change?

davidwengier:
I'm never 100% confident that the experimental instance reflects this type of change but I checked that:

* Everything gets deployed correctly to `artifacts\Debug\VSSetup\Rules` in the expected flat structure
* DLLs in ILSpy looks the same before and after
* A new property on a rule in the `Items` subfolder is correctly shown in the property grid in the experimental instance

I was thinking of perhaps doing a manual insertion to get a full build to install and verify, just in case.

davkean:
> * Everything gets deployed correctly to `artifacts\Debug\VSSetup\Rules` in the expected flat structure

That's good enough coverage for me.

PR #4619, opened 2019-02-27
[WIP] Initial support for .editorconfig files

tmeschter:
Initial support for .editorconfig files.

1. Update `BuildOptions` to take the set of analyzer config files, and read them from the `CommandLineArguments` type. This requires a newer version of the Microsoft.CodeAnalysis package.
2. Pass the analyzer config files to the language service through the workspace project context. This requires a newer version of the Microsoft.VisualStudio.LanguageServices package.
3. Add a type to watch for changes (specifically add and remove) to .editorconfig files above the project cone and trigger a new evaluation/design-time build. Changes to .editorconfig files within the project already trigger evaluation/build and we don't need to do anything special for that. This new type, `EditorConfigFileWatcher`, is based off of the `ProjectAssetsFileWatcher` but could probably be greatly improved.

### DIFF
--- a/build/Packages.targets
+++ b/build/Packages.targets
@@ -13,6 +13,7 @@
       https://dotnet.myget.org/F/project-system-interop/api/v3/index.json;
       https://vside.myget.org/F/vs-impl/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
+      C:\Users\tomescht\Documents\LocalPackageSource
     </RestoreSources>
   </PropertyGroup>
 
@@ -95,8 +96,8 @@
     <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.0.201-pre-g7d366164d0" />
 
     <!-- Roslyn -->
-    <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="2.11.0-beta2-63529-05" />
-    <PackageReference Update="Microsoft.CodeAnalysis" Version="2.11.0-beta2-63529-05" />
+    <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="3.0.0-dev" />
+    <PackageReference Update="Microsoft.CodeAnalysis" Version="3.0.0-dev" />
     <PackageReference Update="Microsoft.VisualStudio.IntegrationTest.Utilities" Version="2.6.0-beta1-62113-02" />
 
     <!-- Analyzers -->
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
@@ -1,0 +1,271 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.Threading.Tasks;
+
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
+{
+    /// <summary>
+    ///     Watches for adds and deletes of .editorconfig files *above* the project
+    ///     directory and triggers evaluation as needed. Note that adds and deletes of
+    ///     files within the project directory directory already trigger evaluation, so
+    ///     this type doesn't handle that case.
+    /// </summary>
+    internal class EditorConfigFileWatcher : OnceInitializedOnceDisposedAsync, IVsFreeThreadedFileChangeEvents2
+    {
+        private static readonly TimeSpan s_notifyDelay = TimeSpan.FromMilliseconds(100);
+
+        private readonly IActiveConfiguredProjectSubscriptionService _activeConfiguredProjectSubscriptionService;
+        private readonly IVsService<IVsAsyncFileChangeEx> _fileChangeService;
+        private readonly IProjectTreeProvider _fileSystemTreeProvider;
+        private readonly IUnconfiguredProjectCommonServices _projectServices;
+        private readonly IUnconfiguredProjectTasksService _projectTasksService;
+
+        private Dictionary<string, uint> _filesBeingWatched;
+        private IDisposable _treeWatcher;
+        private ITaskDelayScheduler _taskDelayScheduler;
+        private CancellationTokenSource _taskDelaySchedulerCancellationTokenSource;
+
+        [ImportingConstructor]
+        public EditorConfigFileWatcher(
+            IVsService<SVsFileChangeEx, IVsAsyncFileChangeEx> fileChangeService,
+            [Import(ContractNames.ProjectTreeProviders.FileSystemDirectoryTree)] IProjectTreeProvider fileSystemTreeProvider,
+            IUnconfiguredProjectCommonServices projectServices,
+            IUnconfiguredProjectTasksService projectTasksService,
+            IActiveConfiguredProjectSubscriptionService activeConfiguredProjectSubscriptionService)
+            : base(projectServices.ThreadingService.JoinableTaskContext)
+        {
+            Requires.NotNull(fileChangeService, nameof(fileChangeService));
+            Requires.NotNull(fileSystemTreeProvider, nameof(fileSystemTreeProvider));
+            Requires.NotNull(projectServices, nameof(projectServices));
+            Requires.NotNull(projectTasksService, nameof(projectTasksService));
+            Requires.NotNull(activeConfiguredProjectSubscriptionService, nameof(activeConfiguredProjectSubscriptionService));
+
+            _fileChangeService = fileChangeService;
+            _fileSystemTreeProvider = fileSystemTreeProvider;
+            _projectServices = projectServices;
+            _projectTasksService = projectTasksService;
+            _activeConfiguredProjectSubscriptionService = activeConfiguredProjectSubscriptionService;
+        }
+
+#pragma warning disable RS0030 // Do not used banned APIs
+        [ConfiguredProjectAutoLoad(RequiresUIThread = false)]
+#pragma warning restore RS0030 // Do not used banned APIs
+        [AppliesTo(ProjectCapability.DotNet)]
+        internal void Load()
+        {
+            var _ = InitializeAsync();
+        }
+
+        /// <summary>
+        /// Called on changes to the project tree.
+        /// </summary>
+        internal async Task DataFlow_ChangedAsync(IProjectVersionedValue<Tuple<IProjectTreeSnapshot, IProjectSubscriptionUpdate>> dataFlowUpdate)
+        {
+            await InitializeAsync();
+
+            IProjectTreeSnapshot treeSnapshot = dataFlowUpdate.Value.Item1;
+            IProjectTree newTree = treeSnapshot.Tree;
+            if (newTree == null)
+            {
+                return;
+            }
+
+            // If tree changed when we are disposing then ignore the change.
+            if (IsDisposing)
+            {
+                return;
+            }
+
+            IProjectSubscriptionUpdate projectUpdate = dataFlowUpdate.Value.Item2;
+            SortedSet<string> updatedEditorConfigFilePaths = GetEditorConfigFilePaths(newTree, projectUpdate);
+
+            var filesToStopWatching = _filesBeingWatched.Keys.Where(p => !updatedEditorConfigFilePaths.Contains(p)).ToArray();
+            await UnregisterFileWatchersAsync(filesToStopWatching);
+
+            var filesToStartWatching = updatedEditorConfigFilePaths.Where(p => !_filesBeingWatched.TryGetValue(p, out var _)).ToArray();
+            await RegisterFileWatchersAsync(filesToStartWatching);
+        }
+
+        private async Task UnregisterFileWatchersAsync(string[] filesToStopWatching)
+        {
+            IVsAsyncFileChangeEx fileChangeService = await _fileChangeService.GetValueAsync();
+            foreach (string editorConfigFilePath in filesToStopWatching)
+            {
+                uint cookie = _filesBeingWatched[editorConfigFilePath];
+                if (cookie != VSConstants.VSCOOKIE_NIL)
+                {
+                    await fileChangeService.UnadviseFileChangeAsync(cookie);
+                }
+
+                _filesBeingWatched.Remove(editorConfigFilePath);
+            }
+        }
+
+        private async Task RegisterFileWatchersAsync(string[] filesToStartWatching)
+        {
+            IVsAsyncFileChangeEx fileChangeService = await _fileChangeService.GetValueAsync();
+            foreach (string editorConfigFilePath in filesToStartWatching)
+            {
+                _filesBeingWatched[editorConfigFilePath] =
+                    await fileChangeService.AdviseFileChangeAsync(
+                        editorConfigFilePath,
+                        _VSFILECHANGEFLAGS.VSFILECHG_Add | _VSFILECHANGEFLAGS.VSFILECHG_Del,
+                        sink: this);
+            }
+        }
+
+        /// <summary>
+        /// Initialize the watcher.
+        /// </summary>
+        protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
+        {
+            // Explicitly get back to the thread pool for the rest of this method so we don't tie up the UI thread;
+            await TaskScheduler.Default;
+
+            await _projectTasksService.LoadedProjectAsync(() =>
+                {
+                    _filesBeingWatched = new Dictionary<string, uint>(StringComparers.Paths);
+                    _taskDelaySchedulerCancellationTokenSource = CreateLinkedCancellationTokenSource();
+                    _taskDelayScheduler = new TaskDelayScheduler(
+                        s_notifyDelay,
+                        _projectServices.ThreadingService,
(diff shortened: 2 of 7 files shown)


build/Packages.targets
```diff
@@ -13,6 +13,7 @@
       https://dotnet.myget.org/F/project-system-interop/api/v3/index.json;
       https://vside.myget.org/F/vs-impl/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
+      C:\Users\tomescht\Documents\LocalPackageSource
```
tmeschter: This will be removed before the work is merged in, once suitable official packages are available for consumption.

build/Packages.targets
```diff
@@ -95,8 +96,8 @@
     <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.0.201-pre-g7d366164d0" />
 
     <!-- Roslyn -->
-    <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="2.11.0-beta2-63529-05" />
-    <PackageReference Update="Microsoft.CodeAnalysis" Version="2.11.0-beta2-63529-05" />
+    <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="3.0.0-dev" />
+    <PackageReference Update="Microsoft.CodeAnalysis" Version="3.0.0-dev" />
```
tmeschter: As before, these will be updated once suitable official packages are available.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+            _fileChangeService = fileChangeService;
+            _fileSystemTreeProvider = fileSystemTreeProvider;
+            _projectServices = projectServices;
+            _projectTasksService = projectTasksService;
+            _activeConfiguredProjectSubscriptionService = activeConfiguredProjectSubscriptionService;
+        }
+
+#pragma warning disable RS0030 // Do not used banned APIs
+        [ConfiguredProjectAutoLoad(RequiresUIThread = false)]
+#pragma warning restore RS0030 // Do not used banned APIs
```
tmeschter: What should I be doing instead of using this banned API?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+                        editorConfigFilePath,
+                        _VSFILECHANGEFLAGS.VSFILECHG_Add | _VSFILECHANGEFLAGS.VSFILECHG_Del,
+                        sink: this);
+            }
+        }
+
+        /// <summary>
+        /// Initialize the watcher.
+        /// </summary>
+        protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
```
tmeschter: I believe this whole method is overly complicated, but I'm not sure what the simpler version looks like. Notably, I don't actually care about the .editorconfig files passed to the compiler, and within this method I don't even care that files have changed in the general case.

All I really need to be able to do is identify the set of .editorconfig files above the project cone when the project is created/opened and when it is moved. Other changes are not relevant.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+        {
+            try
+            {
+                TraceUtilities.TraceVerbose("A .editorconfig file has been added or removed. Marking project dirty.");
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+#pragma warning disable RS0030
+                await _projectServices.Project.Services.ProjectAsynchronousTasks.LoadedProjectAsync(async () =>
+#pragma warning restore RS0030
```
tmeschter: When one of the .editorconfig files outside the project cone has changed we need to re-evaluate/re-run the design-time build. What should I use instead of this banned API?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/BuildOptions.cs
```diff
@@ -8,12 +8,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
 {
     internal class BuildOptions
     {
-        public BuildOptions(ImmutableArray<CommandLineSourceFile> sourceFiles, ImmutableArray<CommandLineSourceFile> additionalFiles, ImmutableArray<CommandLineReference> metadataReferences, ImmutableArray<CommandLineAnalyzerReference> analyzerReferences)
+        public BuildOptions(ImmutableArray<CommandLineSourceFile> sourceFiles, ImmutableArray<CommandLineSourceFile> additionalFiles, ImmutableArray<CommandLineReference> metadataReferences, ImmutableArray<CommandLineAnalyzerReference> analyzerReferences, ImmutableArray<string> analyzerConfigFiles)
```
jmarolf: This seems long enough.  Can you wrap this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerConfigItemHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+    [Export(typeof(IWorkspaceContextHandler))]
+    internal class AnalyzerConfigItemHandler : AbstractWorkspaceContextHandler, ICommandLineHandler
+    {
+        private readonly UnconfiguredProject _project;
+        private readonly HashSet<string> _paths = new HashSet<string>(StringComparers.Paths);
+
+        [ImportingConstructor]
+        public AnalyzerConfigItemHandler(UnconfiguredProject project)
+        {
+            Requires.NotNull(project, nameof(project));
```
jmarolf: When can this happen?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
jmarolf: use `_projectTasksService`

---
In reply to: [260990526](https://github.com/dotnet/project-system/pull/4619#discussion_r260990526) [](ancestors = 260990526)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerConfigItemHandler.cs
tmeschter: This is heavily based on the `AdditionalFilesItemHandler`, right down to the doc comment. It may also handle the evaluation side of things; I'm not entirely sure. :-)

Also, you could conceivably have a target that generates MSBuild items for .editorconfig files, so we can't just consider the evaluation model.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
jmarolf: Implement `IProjectDynamicLoadComponent`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+                });
+        }
+
+        private CancellationTokenSource CreateLinkedCancellationTokenSource()
+        {
+            // we want to cancel when we switch what file is watched, or when the project is unloaded
+            if (_projectServices.Project?.Services?.ProjectAsynchronousTasks?.UnloadCancellationToken != null)
+            {
+                return CancellationTokenSource.CreateLinkedTokenSource(
+                    _projectServices.Project.Services.ProjectAsynchronousTasks.UnloadCancellationToken);
```
davidwengier: Question: Does this need to link to this classes Dispose token? Or would it do the same job as this effectively?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+
+            // TODO (tmeschter): if $(DiscoverEditorConfigFiles) is "false", disable all file watchers by returning an empty set.
+
+            string projectDirectory = Path.GetDirectoryName(projectFilePath);
+            string directory = Path.GetDirectoryName(projectDirectory);
+            while (directory != null)
+            {
+                string editorConfigFilePath = Path.Combine(directory, ".editorconfig");
+                editorConfigFilePaths.Add(editorConfigFilePath);
+            }
```
davidwengier: Needs `directory = Path.GetDirectoryName(projectDirectory);` in here to prevent infinite loop :P

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
davkean: If the warning didn't provide enough info, can you please PR a better message?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+            _projectServices = projectServices;
+            _projectTasksService = projectTasksService;
+            _activeConfiguredProjectSubscriptionService = activeConfiguredProjectSubscriptionService;
+        }
+
+#pragma warning disable RS0030 // Do not used banned APIs
+        [ConfiguredProjectAutoLoad(RequiresUIThread = false)]
+#pragma warning restore RS0030 // Do not used banned APIs
+        [AppliesTo(ProjectCapability.DotNet)]
+        internal void Load()
```
davkean: This is loading for Shared Projects, which is probably not want you want. You want DotNetLanguageService for those projects just with the LanguageService.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+        [ImportingConstructor]
+        public EditorConfigFileWatcher(
+            IVsService<SVsFileChangeEx, IVsAsyncFileChangeEx> fileChangeService,
+            [Import(ContractNames.ProjectTreeProviders.FileSystemDirectoryTree)] IProjectTreeProvider fileSystemTreeProvider,
+            IUnconfiguredProjectCommonServices projectServices,
+            IUnconfiguredProjectTasksService projectTasksService,
+            IActiveConfiguredProjectSubscriptionService activeConfiguredProjectSubscriptionService)
+            : base(projectServices.ThreadingService.JoinableTaskContext)
+        {
+            Requires.NotNull(fileChangeService, nameof(fileChangeService));
```
davkean: Remove null checks, we don't check for null for MEF constructors.

davkean:
I'm think the design of the watcher needs some thought; right now you have a component that lives in every single configuration for every single project that registers for and watches (likely) the same set files and then goes and tells every single configuration to reevaluate multiple times (because each configuration goes and tells every other configuration in the same project to reevaluate). It's seems unintentional that the project reevaluates multiple times, but in a 1000 project solution that's a bunch of file watcher registrations calls for basically a handful of files.

I would instead take the same approach we took for targets files; a central solution-wide ("ProjectService") service that each project registers with that is responsible for watching any editor configs that we might be interested in. That way we only have a single file watcher registration for each possible location of an editor config. See ImportReloadManager inside CPS for an example.

davkean:
>  Changes to .editorconfig files within the project already trigger evaluation/build and we don't need to do anything special for that.

Alternatively, why do we treat files above the project code specially, what mechanism are we using this to make this work? And why can't we use it here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
```diff
@@ -0,0 +1,271 @@
...
+                {
+                    _filesBeingWatched = new Dictionary<string, uint>(StringComparers.Paths);
+                    _taskDelaySchedulerCancellationTokenSource = CreateLinkedCancellationTokenSource();
+                    _taskDelayScheduler = new TaskDelayScheduler(
+                        s_notifyDelay,
+                        _projectServices.ThreadingService,
+                        _taskDelaySchedulerCancellationTokenSource.Token);
+
+                    // The tree source to get changes to the tree so that we can identify *when* the set of .editorconfig files changes.
+                    ProjectDataSources.SourceBlockAndLink<IProjectVersionedValue<IProjectTreeSnapshot>> treeSource = _fileSystemTreeProvider.Tree.SyncLinkOptions();
```
davkean: Looking at usage here, it seems like you are using data flow just to grab the directory of the project?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerConfigItemHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+
+using Microsoft.VisualStudio.ProjectSystem.Logging;
+
+namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
+{
+    /// <summary>
+    ///     Handles changes to the &lt;EditorConfigFiles/&gt; items during design-time builds.
```
davkean: Is the item actually named "EditorConfigFiles", if so, it should be "EditorConfigFile" -> items are not plural.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
davkean: Projects are never moved, the can be renamed, but their directory always remains the same.

I'm not sure I understand the created/opened scenario, maybe we can talk tomorrow to discuss what you are trying to do? TBH, it's not clear why adding a file in a parent directory is different from adding file under a project.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerConfigItemHandler.cs
davkean: Short term this is okay. When I move AdditionalFiles to also respect evaluation, I'll move this one too. 

The reason you want to respect evaluation is for user experience; you drop an editorconfig in the project or a directory - you probably want Roslyn to see the file immediately, not later on when the design-time build has finished.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
davkean: Yes, it should just link both DisposalToken/UnloadCancellationToken. I see you copied this code from ProjectFileWatcher but it's incorrect, down to the null checks.

davkean:
The language-service changes look good, and I'll recommend you separate them into a different PR so that we can merge it in.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/EditorConfigFileWatcher.cs
tmeschter: Experimentally, dropping a .editorconfig file into project's directory (or a sub-directory) causes a re-evaluation/design-time build without me having to do anything else. I've been assuming CPS is just watching for any changes under the project directory.

davkean: Only if its included in a glob.

davkean: Let's sync up on this so that we can chat about the design.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AnalyzerConfigItemHandler.cs
tmeschter: It is in fact EditorConfigFiles.

davkean: Can we fix that? Items are not plural.

tmeschter: Maybe--we'll need to get the compiler team to buy off on that. I can't say it's at the top of my priority list at the moment, though. :-)

tmeschter:
The language services changes are now split off into https://github.com/dotnet/project-system/pull/4637.

davkean:
@tmeschter Going to close this as I don't believe it's relevant anymore. Feel free to reopen it if is.

PR #4620, opened 2019-02-27
Clean and sort rules

davidwengier:
Apologies for the size of the review, but this baselines all of the item type XAML rules to be sorted, allowing for much easier future comparisons.

What was done:

* Sort nodes:
  * Metadata appears first (ie, nodes named [Parent].[Thing])
  * Child elements sorted alphabetically by Name attribute if it exists
  * Remaining child elements sorted by element name
* Sort attributes:
  * Name attribute output first
  * Remaining attributes sorted alphabetically 
* Misc
  * EnumProperty child nodes not sorted to maintian UI order
  * Blank line between each 2nd level node

I wanted to get the attributes to be one-per-line and wrapped nicely but it turns out none of the XML writers in .NET support any whitespace between attributes. I could change the tool I used to write out the XML manually to do this if people feel strongly about it.

If its easier to review the tool that made the changes its here: https://github.com/davidwengier/XamlRuleCleaner/blob/master/Program.cs

In future I'm tempted to make this validate more rules and perhaps put something in place as a PR check so we don't regress things. Things like not having localizable strings in non-BrowseObject or invisible properties would be straight forward enough.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AppDesigner.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AppDesigner.xaml
@@ -1,27 +1,37 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
-<Rule
-    Name="AppDesigner"
-    PageTemplate="generic"
-    xmlns="http://schemas.microsoft.com/build/2009/properties">
-
+<Rule Name="AppDesigner"
+      PageTemplate="generic"
+       xmlns="http://schemas.microsoft.com/build/2009/properties">
   <Rule.DataSource>
-    <DataSource Persistence="ProjectFile" Label="Configuration" SourceOfDefaultValue="AfterContext" />
+    <DataSource Label="Configuration"
+                Persistence="ProjectFile"
+                SourceOfDefaultValue="AfterContext" />
   </Rule.DataSource>
 
-  <StringProperty Name="FolderName" Visible="false" Default="Properties">
-    <StringProperty.DataSource>
-      <DataSource Persistence="ProjectFile" Label="Configuration" HasConfigurationCondition="false"
-                  PersistedName="AppDesignerFolder" SourceOfDefaultValue="AfterContext" />
-    </StringProperty.DataSource>
-  </StringProperty>
-
-
-  <BoolProperty  Name="ContentsVisibleOnlyInShowAllFiles" Visible="false" Default="false" ReadOnly="true">
+  <BoolProperty Name="ContentsVisibleOnlyInShowAllFiles"
+                Default="false"
+                ReadOnly="true"
+                Visible="false">
     <BoolProperty.DataSource>
-      <DataSource Persistence="ProjectFile" Label="Configuration" HasConfigurationCondition="false"
-                  PersistedName="AppDesignerFolderContentsVisibleOnlyInShowAllFiles" SourceOfDefaultValue="AfterContext" />
+      <DataSource HasConfigurationCondition="false"
+                  Label="Configuration"
+                  PersistedName="AppDesignerFolderContentsVisibleOnlyInShowAllFiles"
+                  Persistence="ProjectFile"
+                  SourceOfDefaultValue="AfterContext" />
     </BoolProperty.DataSource>
   </BoolProperty>
+
+  <StringProperty Name="FolderName"
+                  Default="Properties"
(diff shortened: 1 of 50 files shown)


davkean:
Is there a reason this only appears to be item types and not the other rules? Can we sort the other rules as well?

davkean:
> 
I wanted to get the attributes to be one-per-line and wrapped nicely but it turns out none of the XML writers in .NET support any whitespace between attributes. I could change the tool I used to write out the XML manually to do this if people feel strongly about it

`XmlWriterSettings.NewLineOnAttributes` controls this. IMO, it would be nicer.

davidwengier:
> Is there a reason this only appears to be item types and not the other rules? Can we sort the other rules as well?

No big reason, I just had no need to compare those. Easy enough to run the tool on them though.

> XmlWriterSettings.NewLineOnAttributes

Yes, but using the inbuilt indentation means writing whitespace of any kind can't be done, which means the file ends up looking like this which I don't personally think is nicer, but its subjective so ¯\_(ツ)_/¯

```
<Rule
  Name="Compile"
  PageTemplate="generic"
  PropertyPagesHidden="true" xmlns="http://schemas.microsoft.com/build/2009/properties">
  <Rule.DataSource>
    <DataSource
      HasConfigurationCondition="false"
      ItemType="Compile"
      Label="Configuration"
      Persistence="ProjectFile"
      SourceOfDefaultValue="AfterContext" />
  </Rule.DataSource>
  <DynamicEnumProperty
    Name="{}{ItemType}"
    EnumProvider="ItemTypes" />
  <BoolProperty
    Name="AutoGen"
    Visible="false" />
  <EnumProperty
    Name="CopyToOutputDirectory">
    <EnumValue
      Name="Never" />
    <EnumValue
      Name="Always" />
    <EnumValue
      Name="PreserveNewest" />
  </EnumProperty>
```

My ideal I was trying to achieve would be this:

```
<Rule Name="Compile"
      PageTemplate="generic"
      PropertyPagesHidden="true" xmlns="http://schemas.microsoft.com/build/2009/properties">
  <Rule.DataSource>
    <DataSource HasConfigurationCondition="false"
                ItemType="Compile"
                Label="Configuration"
                Persistence="ProjectFile"
                SourceOfDefaultValue="AfterContext" />
  </Rule.DataSource>
  <DynamicEnumProperty Name="{}{ItemType}"
                       EnumProvider="ItemTypes" />
  <BoolProperty Name="AutoGen"
                Visible="false" />
  <EnumProperty Name="CopyToOutputDirectory">
    <EnumValue Name="Never" />
    <EnumValue Name="Always" />
    <EnumValue Name="PreserveNewest" />
  </EnumProperty>
```

davkean:
Do you mean because there's no whitespace between the lines or because the attribute isn't on the first line? The latter meh, I think that's still better than current.

davidwengier:
Mainly the attribute not being on the same line, especially for the one-liners. I just find it easier to parse.

Might have another go at it tomorrow, otherwise I'll just go with the auto indentation.

wjk:
FWIW, I find the current (every tag on one line) indentation style to be quite usable. I always have line wrapping turned on in VS. While I do agree @davidwengier's "ideal" style is much better, I can totally relate to tooling difficulties forcing compromises on code style. After all, I face this every day with my .NET Core WinForms/WPF projects. 😄

davidwengier:
I worked out some reflection hacks and managed to get things formatted almost perfectly, except for one extra space in front of `xmlns` for some reason I'm deciding to ignore. Will update with all XAML files reformatted.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AppDesigner.xaml
```diff
@@ -1,27 +1,37 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
-<Rule
-    Name="AppDesigner"
-    PageTemplate="generic"
-    xmlns="http://schemas.microsoft.com/build/2009/properties">
-
+<Rule Name="AppDesigner"
+      PageTemplate="generic"
+       xmlns="http://schemas.microsoft.com/build/2009/properties">
```
davkean: There's an extra space here that's going to get removed on reformat.

davidwengier: Fixed and reformatted all

etbyrd:
I agree with @drewnoakes - two space indent makes working on the WPF property pages a little more difficult. I made a PR for only applying the indent to the Rules: https://github.com/dotnet/project-system/pull/4630

Issue #4622, opened 2019-02-28
NullReferenceException when trying to save values in the NET Core Debug prop page


etbyrd:
**Visual Studio Version**: d16.0stg 

**Summary:** Trying to save properties in the NET Core property pages causes a NullReferenceException 

**Steps to Reproduce**:

1. Edit Application Arguments in the Debug prop page

2. Save

**Expected Behavior**: Saves the value to the launchSettings.json

**Actual Behavior**: NullReferenceException, VS hangs and crashes

**User Impact**: Upset user 

**Cause**: 

At https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs#L599, `HandleLaunchSettingsFileChangedAsync` is returning null and the `IProjectFaultHandlerService.RegisterFaultHandler` throws a NullReferenceException

etbyrd:
This may be a regression from https://github.com/dotnet/project-system/commit/1e0868519fcb884a259db8c358aeca7e12ea452a @drewnoakes

davkean:
This will be a regression from my change to observe tasks that were previously unobserved.

etbyrd:
Fix #4623 merged

PR #4623, opened 2019-02-28
Make HandleLaunchSettingsFileChangedAsync return Task.CompletedTask

etbyrd:
Fix for: https://github.com/dotnet/project-system/issues/4622

Even if the logic in HandleLaunchSettingsFileChangedAsync was regressed and a different change is needed, we should probably still have this check.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Debug/LaunchSettingsProviderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Debug/LaunchSettingsProviderTests.cs
@@ -458,7 +458,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
                 // Set the ignore flag. It should be ignored.
                 provider.LastSettingsFileSyncTimeTest = DateTime.MinValue;
                 provider.SetIgnoreFileChanges(true);
-                Assert.Null(provider.LaunchSettingsFile_ChangedTest());
+                Assert.Equal(provider.LaunchSettingsFile_ChangedTest(), Task.CompletedTask);
                 Assert.Null(provider.CurrentSnapshot);
 
                 // Should run this time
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -596,7 +596,11 @@ protected async Task CheckoutSettingsFileAsync()
         /// </summary>
         private void LaunchSettingsFile_Changed(object sender, FileSystemEventArgs e)
         {
-            _projectFaultHandler.Forget(HandleLaunchSettingsFileChangedAsync(), _project);
+            Task task = HandleLaunchSettingsFileChangedAsync();
```
davkean: Async methods should never return null, change HandleLaunchSettingsFileChangedAsync to return Task.CompletedTask.

davkean:
@jjmew @Pilchie This crash prevents the user from making changes to the Debug page, we'll want this in 16.0. @etbyrd Please retarget this to dev16.x.

Pilchie:
Per our call, can you verify that the original break is in dev16.0.x, and not just in master?

davkean:
Actually, ignore me. @Pilchie Reminded me that this broke in master not dev16.x. This branch is correct.

etbyrd:
Sure - I also verified that this does fix it

jmarolf:
@etbyrd 	you are clear to merge!

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Debug/LaunchSettingsProviderTests.cs
```diff
@@ -458,7 +458,7 @@ public async Task LaunchSettingsFile_TestIgnoreFlag()
                 // Set the ignore flag. It should be ignored.
                 provider.LastSettingsFileSyncTimeTest = DateTime.MinValue;
                 provider.SetIgnoreFileChanges(true);
-                Assert.Null(provider.LaunchSettingsFile_ChangedTest());
+                Assert.Equal(provider.LaunchSettingsFile_ChangedTest(), Task.CompletedTask);
```
drewnoakes: I don't think equality checks really make sense for tasks. I'd either make this `Assert.Same`, or test the returned task's state via its properties.

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Debug/LaunchSettingsProviderTests.cs
```diff
@@ -482,7 +482,7 @@ public async Task LaunchSettingsFile_TestTimeStampFlag()
                 // Write new file, but set the timestamp to match
                 moqFS.WriteAllText(provider.LaunchSettingsFile, JsonStringWithWebSettings);
                 provider.LastSettingsFileSyncTimeTest = moqFS.LastFileWriteTime(provider.LaunchSettingsFile);
-                Assert.Null(provider.LaunchSettingsFile_ChangedTest());
+                Assert.Equal(provider.LaunchSettingsFile_ChangedTest(), Task.CompletedTask);
```
drewnoakes: Ditto.

Issue #1125, opened 2017-01-10
.NET Core App project Debug properties are missing "enable native code debugging" checkbox


tmat:
Desktop projects have 
![image](https://cloud.githubusercontent.com/assets/41759/21829650/3911d5a6-d74c-11e6-9943-aa923cc76744.png)

I can't find this option in .NET Core project. I need to F5 with mixed mode debugging in order to debug some interop code.

tmat:
Oh, I guess I know why:

![image](https://cloud.githubusercontent.com/assets/41759/21829702/9872be0c-d74c-11e6-86f8-88e999c0137a.png)

Seems like the debugger doesn't support interop debugging with CoreCLR :(

tmat:
@noahfalk I hear this is CoreCLR issue but I couldn't find an issue in the coreclr repo that would be tracking it.

noahfalk:
@mikem8361 might be able to point you at something or you can create one now

mikem8361:
The latest coreclr master branch (.NET Core 2.0) has mixed mode debugging enabled for Windows.

axel-habermaier:
Since this checkbox will apparently not be implemented in VS2017, is there any workaround apart from starting the program, and attaching the debugger manually? (Which is cumbersome as I have to guess the right "dotnet" process in the list?)

axel-habermaier:
I tried adding ``<EnableUnmanagedDebugging>true</EnableUnmanagedDebugging>`` to a ``PropertyGroup`` of the csproj file, but unfortunately, it had no effect. Would it be possible to implement support for the functionality even without exposing it in the UI, if there is not enough time to do both?

davkean:
Hmm, looks like we don't currently respect the legacy debugger settings with the new debugger provider that we wrote. This is where we control it: https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectDebuggerProvider.cs#L76, that's really unfortunate, because the provider that this replaces does support it. 

@srivatsn Given interop debugging is supported in 2.0, we should enable this for 2.0.

onovotny:
Keep in mind that this also affects SDK-style projects that only target .NET Desktop. I have a .NET Framework exe using the SDK project system and have a need to do mixed mode debugging. I need a cumbersome hack (use a start program from a legacy project pointing to the output of the SDK project)

davidmatson:
[...]

PR #4624, opened 2019-02-28
Add Native Code Debugging CheckBox to the Debug prop page

etbyrd:
Fix for https://github.com/dotnet/project-system/issues/1125
Assuming I am reading the issue correctly, this only changes "nativeDebugging" in launchSettings.json for the currently selected profile 

Adds the check box to the bottom:
new_debug
Like non-Core projects:
old_debug

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
@@ -296,6 +296,38 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
             }
         }
 
+        public bool NativeCodeDebugging
+        {
+            get
+            {
+                if (!IsProfileSelected)
+                {
+                    return false;
+                }
+
+                if (SelectedDebugProfile.OtherSettings.Keys.Contains("nativeDebugging"))
+                {
+                    return (bool) SelectedDebugProfile.OtherSettings["nativeDebugging"];
+                }
+
+                return false;
+            }
+            set
+            {
+                //Unlike other properties that have default values, nativeDebugging may not be set yet. Because false is the default behavior adding it has no affect
+                if (!SelectedDebugProfile.OtherSettings.ContainsKey("nativeDebugging"))
+                {
+                    SelectedDebugProfile.OtherSettings.Add("nativeDebugging", false);
+                }
+
+                if (SelectedDebugProfile != null && (bool) SelectedDebugProfile.OtherSettings["nativeDebugging"] != value)
+                {
+                    SelectedDebugProfile.OtherSettings["nativeDebugging"] = value;
+                    OnPropertyChanged(nameof(NativeCodeDebugging));
+                }
+            }
+        }
+
         public bool SupportsExecutable
         {
             get
@@ -503,6 +535,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                 OnPropertyChanged(nameof(ExecutablePath));
                 OnPropertyChanged(nameof(LaunchPage));
                 OnPropertyChanged(nameof(HasLaunchOption));
+                OnPropertyChanged(nameof(NativeCodeDebugging));
                 OnPropertyChanged(nameof(WorkingDirectory));
 
                 UpdateLaunchTypes();
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
@@ -138,6 +138,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                         OnPropertyChanged(nameof(SupportsWorkingDirectory));
                         OnPropertyChanged(nameof(SupportsLaunchUrl));
                         OnPropertyChanged(nameof(SupportsEnvironmentVariables));
+                        OnPropertyChanged(nameof(SupportNativeDebugging));
                         OnPropertyChanged(nameof(ActiveProviderUserControl));
                         OnPropertyChanged(nameof(DoesNotHaveErrors));
                     }
@@ -325,6 +326,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
                     SelectedDebugProfile.OtherSettings["nativeDebugging"] = value;
                     OnPropertyChanged(nameof(NativeCodeDebugging));
                 }
+            }
+        }
+
+        public bool SupportNativeDebugging
+        {
+            get
+            {
+                return ActiveProviderSupportsProperty(UIProfilePropertyName.NativeDebugging);
             }
         }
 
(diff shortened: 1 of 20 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -296,6 +296,38 @@ public bool HasLaunchOption
...
+            get
+            {
+                if (!IsProfileSelected)
+                {
+                    return false;
+                }
+
+                if (SelectedDebugProfile.OtherSettings.Keys.Contains("nativeDebugging"))
+                {
+                    return (bool) SelectedDebugProfile.OtherSettings["nativeDebugging"];
```
davidwengier: Nit: I don't think we have spaces after casts.. could be missing from .editorconfig?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.resx
```diff
@@ -246,4 +246,7 @@
   <data name="OKBtnHelpText" xml:space="preserve">
     <value>Create the profile and return to the debug property page with the new profile selected</value>
   </data>
+  <data name="chkNativeCodeDebuggingText" xml:space="preserve">
+    <value>Enable native code debugging</value>
```
davidwengier: Accelerator key?

Though now that I think about it, can we reuse the same resource from non-Core? Then the translations will be the same? Or are those resources still in the other DLL in the VS tree?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -296,6 +296,38 @@ public bool HasLaunchOption
...
+        public bool NativeCodeDebugging
+        {
+            get
+            {
+                if (!IsProfileSelected)
+                {
+                    return false;
+                }
+
+                if (SelectedDebugProfile.OtherSettings.Keys.Contains("nativeDebugging"))
```
davidwengier: ContainsKey would be better:
```suggestion
                if (SelectedDebugProfile.OtherSettings.ContainsKey("nativeDebugging"))
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.resx
etbyrd: I asked Tom about this and he said not to worry about it - the loc folks should have a database to pull from

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
davkean: Use TryGetValue

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -296,6 +296,38 @@ public bool HasLaunchOption
...
+            }
+            set
+            {
+                //Unlike other properties that have default values, nativeDebugging may not be set yet. Because false is the default behavior adding it has no affect
+                if (!SelectedDebugProfile.OtherSettings.ContainsKey("nativeDebugging"))
+                {
+                    SelectedDebugProfile.OtherSettings.Add("nativeDebugging", false);
+                }
+
+                if (SelectedDebugProfile != null && (bool) SelectedDebugProfile.OtherSettings["nativeDebugging"] != value)
```
davkean: It seems simpler if you just write the following, also avoids "dirtying" the in memory model when the value is set to false and it's not present in the settings:

``` C#
            SelectedDebugProfile.OtherSettings.TryGetValue("nativeDebugging", out object previousValue);

            if ((bool)previousValue != value)
            {
                SelectedDebugProfile.OtherSettings["nativeDebugging"] = value;
                OnPropertyChanged(nameof(NativeCodeDebugging));
            }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.resx
etbyrd: As for the accelerator keys - nothing on the page has them so I will make a separate PR to add those

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
```diff
@@ -6,4 +6,5 @@ static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResourc
 static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResources.DeleteBtnHelpText.get -> string
 static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResources.NewBtnHelpText.get -> string
 static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResources.OKBtnHelpText.get -> string
-static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResources.chkLaunchBrowserHelpText.get -> string
\ No newline at end of file
+static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResources.chkLaunchBrowserHelpText.get -> string
+static Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.PropertyPageResources.chkNativeCodeDebuggingText.get -> string
```
drewnoakes: Not specific to this PR but it seems weird to have the UI elements as public fields.

davkean: Agreed, was looking at this earlier - why are these public API?

drewnoakes: Never mind. Missed that this was a resource.

davkean: But why is this public?

etbyrd: Not sure, but it is generated as Public when you edit the resx

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
etbyrd: Good point - changed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
drewnoakes: Is making these internal out of the question?

etbyrd: I don't believe it is out of the question - I will make a PR for this and we can discuss there

etbyrd:
@davkean @davidwengier Can I get another review pass with the new commits?

PR #4629, opened 2019-03-01
Prevent ANE when removing item with empty OriginalItemSpec

drewnoakes:
This could be triggered by changing the TFM from netcoreapp2.2 to net4.5. In that case:

- A DTB fires off a change
- `removedItem` is "Microsoft.NETCore.App"
- The removed rule is "ResolvedSdkReference"
- The "OriginalItemSpec" item property is an empty string
- `GetStringProperty` therefore returns `null` which is assigned to `dependencyId`
- `shouldProceed` is a delegate referencing `IReadOnlyDictionary<,>.ContainsKey`, and throws ANE for the null key

Although this exception would be thrown back into the Dataflow block, changing the TFM causes new Dataflow subscriptions to be created, so the UI stabilises eventually, with or without this exception.

Using `removedItem` in place of the OriginalItemSpec allows the removal to continue correctly.


drewnoakes:
Thanks to @davkean for reporting this.

Investigation suggests this is not a regression from 15.9. I believe the user impact is low enough that there's no need to consider this for 16.0.

PR #4630, opened 2019-03-01
Only Apply Two Space Indent on Rules

etbyrd:
Two space indent makes sense for the rules, but I think four spaces is easier to work with - especially for WPF.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,9 +17,6 @@ charset = utf-8-bom
 [*.xaml]
 indent_size = 4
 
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -15,6 +15,9 @@ insert_final_newline = true
 charset = utf-8-bom
 
 [*.xaml]
+indent_size = 4
+
+[src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/*.xaml]
```
davkean: Does this apply to subfolders? @davidwengier just added a sub folder.

etbyrd:
@davidwengier Great idea. Updated the PR to reflect this.

PR #4631, opened 2019-03-01
Update Package Version to get the correct property

etbyrd:
Fixes https://github.com/dotnet/project-system/issues/4466

Summary: The ``Package Version`` field on the Package property page was picking up the ``Version`` property instead of the ``Package Version`` property.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
@@ -371,7 +371,7 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <StringProperty Name="Version"
+  <StringProperty Name="PackageVersion"
                   Visible="False">
     <StringProperty.DataSource>
       <DataSource HasConfigurationCondition="False"
--- a/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
@@ -238,7 +238,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                     datalist.Add(data)
                     data = New PropertyControlData(101, "PackageId", PackageId, ControlDataFlags.None, New Control() {PackageIdLabel})
                     datalist.Add(data)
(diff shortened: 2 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
```diff
@@ -371,7 +371,7 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <StringProperty Name="Version"
+  <StringProperty Name="PackageVersion"
```
drewnoakes: Are we 100% sure it's ok to remove `Version` on `GeneralBrowseObject`?

//cc: @davidwengier @davkean

nguerrera:
@davkean PackageVersion will default to Version, and AssemblyVersion will default to Version (major.minor.patch of Version.0). But you are free to change AssemblyVersion and PackageVersion independently.

davkean:
I'll need to go back through history on Monday to understand this - I have a feeling this a deliberate design, and want to understand what's changed around that design or if the original design was broken.

davkean:
The SDK is changing the version in targets? THis might be the same thing as #define constants where we changed the context of the property.

nguerrera:
Played with this and the current design seems not completely unreasonable. After you save the package version change it is reflected in the assembly version. If you want a different assembly version, you can change that down below.

But it's maybe not perfectly intuitive that the label "Package version" means "package version (also implies a corresponding assembly version unless you specify an explicit assembly version. Also, if you manually set PackageVersion in your project file, this won't change the package version)."

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
davkean: In the future be aware that this is an API change - these properties are available to consumers inside VS.

etbyrd:
Because it seems like this can pick up non-intuitive values as in #4466, we should either change the label or change the property it is mapped to, right? As @nguerrera said, this won't even pick up changes to PackageVersion, which is very non-intuitive.

jjmew:
@etbyrd I am moving this to 16.2 I am not sure why this is on 16.1. Please let me know if we need to keep it on 16.1

Issue #182, opened 2016-05-24
Prevent <SubType> from being written to the project file


davkean:
csproj/vbideprj both burn in `<SubType>` when they discover a component or a form that needs to have a different icon and be opened in the designer. Talking with some folks, this was done for performance reasons. We should find a new way to represent this same information _without_ putting it in the project file.

davidwengier:
https://github.com/dotnet/project-system/pull/4635#issuecomment-477311344

davkean:
Starting in 16.1, this has been fixed for new project system (currently used for .NET Core) - SubType is now written to the project user file which is not committed to source control.

LunarLambda:
Hi

This is causing issues for us in VS 2019 because when opening the project after a clean git checkout (or when pulling new files) because VS doesn't properly detect the Forms as such and refuses to let us open the designer unless we add the SubType back to the main .csproj file.

Do we need to update VS or is there something else we're missing?

jnm2:
@LunarLambda It's just very slow sometimes.

LunarLambda:
How slow? I spent a good 20 minutes trying to figure it out before realising I can edit the .csproj files to make it work (though I *then* realised VS seemed to only care that .csproj had been touched at all, since I initially copy-pasted the wrong file path and it still caused the form to show up correctly).

jnm2:
@LunarLambda The slowness was a bug was fixed in VS 2022, but I see in https://github.com/dotnet/winforms/issues/3619 that there's a report that VS 2019 still has the bug. I guess that repository would be the place to raise this issue. Maybe file VS feedback linking it too?

LunarLambda:
Will do, thank you.

PR #4635, opened 2019-03-05
Normalize rules against none

davidwengier:
Now that the rule files are sorted and neat, I was able to easily see what was missing from various files. This PR brings all file rules up to date with `None.xaml` (including SubType which was missing from `None.xaml` itself) and all browse object rules up to date with `None.BrowseObject.xaml`.

Part of #4105

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
@@ -127,6 +127,15 @@
     </StringProperty.DataSource>
   </StringProperty>
 
+  <EnumProperty Name="SubType"
+                Visible="false">
+    <EnumValue Name="Designer" />
+    <EnumValue Name="Component" />
+    <EnumValue Name="Control" />
+    <EnumValue Name="Form" />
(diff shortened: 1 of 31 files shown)


davidwengier:
Oops, XLF commit got lost in cherry picking. Updating!

drewnoakes:
Is it feasible to add unit tests that enforce consistency across these files to prevent them getting out of sync in future?

davidwengier:
Good idea about the tests. I was thinking long term (I think I mentioned it in another PR) about making some kind of xaml rule validator as a PR gate, similar to Tom's localization one, but now that you mention it a unit test should be capable of doing everything, so will add it to the list.

davkean:
tag @lifengl can you also take a look?

davidwengier:
Ping @lifengl

lifengl:
I think to store SubType to the .user file, you need update the browse rule in this way: (maybe a separated PR?)
```xml
    <StringProperty Name="SubType" Visible="false">
        <StringProperty.DataSource>
            <DataSource Persistence="DesignTimeItemPropertiesStorage" ItemType="Compile" PersistedName="SubType" HasConfigurationCondition="false"/>
        </StringProperty.DataSource>
    </StringProperty>
```

davidwengier:
Would rather leave that to #182 and just use this to get things equal so I can start putting in place tests to ensure correctness.

davidwengier:
Merged via #4641

PR #4640, opened 2019-03-07
Fix F#'s project selector registration

davkean:
F#'s project selector registration was pointing to the old GUID of the managed project system package, instead of the new one. This broke in https://github.com/dotnet/project-system/pull/4613.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/811564.


davidwengier:
Should this have been re-targeted so it goes into 16.1? I think thats where #4613 went.

@jmarolf we don't have a 16.1 branch yet, but I _think_ that VS master (which this is being inserted into) is 16.2 isn't it?

jmarolf:
@davidwengier 

master is now 16.2 I've updated the readme to reflect this.

We do have a [16.1](https://github.com/dotnet/project-system/tree/dev16.1.x) branch now (I made it today). PR https://github.com/dotnet/project-system/pull/4644 was the last thing to make it into lab/d16.0stg according to this file [here](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2F.corext%2FConfigs%2Fdotnetprojectsystem-components.json&version=GBlab%2Fd16.1stg) which corresponds to this build [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2492037).

Any PRs that we want to take for 16.1 should be merged into the dev16.1.x branch

jmarolf:
Ignore what I said.  master should stay 16.1 I read the schedule wrong

PR #4641, opened 2019-03-08
Rule validation tests

davidwengier:
Thanks to Drew for the suggestion, these tests validate the rules for rules, and will ensure we don't regress things. Plan is to add more as more things are fixed up.

Includes #4635 so ignore the first 4 commits, and/or just ignore all of the `.xaml` files if its easier since this also includes some fixing of things found by these tests.

Part of #4105

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
@@ -1,0 +1,142 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Xml;
+using Xunit;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Rules
+{
+    public class XamlRuleTests
+    {
+        [Theory]
+        [MemberData(nameof(GetBrowseObjectItemRules))]
+        public void InvisiblePropertiesShouldntBeLocalized(string ruleName, string fullPath)
+        {
+            XmlDocument rule = LoadXamlRule(fullPath);
+
+            // TODO:
+        }
+
+        [Theory]
+        [MemberData(nameof(GetFileItemRules))]
+        public void FileRulesShouldntBeLocalized(string ruleName, string fullPath)
+        {
+            XmlDocument rule = LoadXamlRule(fullPath);
+            
+            // TODO:
+        }
+
+
+        [Theory]
+        [MemberData(nameof(GetBrowseObjectItemRules))]
+        public void HidePropertyPagesForBrowseObjectRules(string ruleName, string fullPath)
+        {
+            XmlDocument rule = LoadXamlRule(fullPath);
+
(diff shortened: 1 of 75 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
```diff
@@ -0,0 +1,289 @@
...
+
+        [Theory]
+        [MemberData(nameof(GetAllRules))]
+        public void NonVisiblePropertiesShouldntBeLocalized(string ruleName, string fullPath)
+        {
+            XmlDocument rule = LoadXamlRule(fullPath);
+
+            foreach (XmlNode node in rule.DocumentElement.ChildNodes)
+            {
+                if (node is XmlElement element && element.HasAttribute("Visible") && element.Attributes["Visible"].Value.Equals("false", StringComparison.OrdinalIgnoreCase))
```
drewnoakes: Is a property that omits the `Visible` attribute considered visible?

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
```diff
@@ -0,0 +1,289 @@
...
+            // First fix up the Name and DisplayName as we know they'll differ.
+            rule.DocumentElement.Attributes["Name"].Value = "None";
+            rule.DocumentElement.Attributes["DisplayName"].Value = "General";
+
+            AssertXmlEqual(none, rule);
+        }
+
+        [Theory]
+        [MemberData(nameof(GetAllRules))]
+        public void NonVisiblePropertiesShouldntBeLocalized(string ruleName, string fullPath)
```
drewnoakes: Do we need another test that verifies visible properties have `DisplayName` (and possibly other fields)?

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
davidwengier: Yes, the default is true.

https://github.com/Microsoft/msbuild/blob/e70a3159d64f9ed6ec3b60253ef863fa883a99b1/src/Framework/XamlTypes/BaseProperty.cs#L52

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
davidwengier: Yes, and many more to boot

drewnoakes: Would be good to capture your ideas for such tests in a tracking issue, if they're not already.

davidwengier: I'm generally tracking everything in #4105

PR #4642, opened 2019-03-08
Revert change

etbyrd:
Was working on creating a PR to add my License fix to the dev16.0.x branch and accidentally pushed to the branch.

I apologize - I updated my remotes and this won't happen again.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.Designer.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.Designer.vb
@@ -274,6 +274,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             '
             resources.ApplyResources(Me.LicenseFileNameTextBox, "LicenseFileNameTextBox")
             Me.LicenseFileNameTextBox.Name = "LicenseFileNameTextBox"
+            Me.LicenseFileNameTextBox.ReadOnly = True
             '
             'LicenseBrowseButton
             '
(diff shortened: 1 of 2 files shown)


Pilchie:
It is.  I'll protect the 16.0 branch.

The other thing I tend to do is `git remote set-url --push upstream no_push`.  (Assuming your upstream is called that.  Might also be `dotnet` or whatever.   This makes it so that the push url doesn't resolve and pushes fail.

drewnoakes:
Ethan and I set the push URL for his upstream remote to gibberish a few minutes back :)

So we have to remember to enable this protection on new branches in future too?

Pilchie:
Yes :(  There doesn't seem to be rule templates.  `dev15.0.x` and `dev15.9.x` already had PR required, but `master` and `dev16.0.x` didn't.

@jmarolf / @davidwengier do we have docs on setting up new branches?  Can this be added to them?

drewnoakes:
I don't see any difference in config between `master`,  `dev15.0.x`, `dev15.9.x`, or `dev16.0.x` branches.

This looks like the setting to use, with an empty list of users.

image

Actually the branch config seems to be rule based. Perhaps it's possible to set this up on all branches with a single rule. Do we ever need to push directly to branches?

Pilchie:
> I don't see any difference in config between master, dev15.0.x, dev15.9.x, or dev16.0.x branches.

That's because I already changed them.

My read was that this option covered it...

image

Pilchie:
But - I guess I missed the "and submitted via pull request"...

drewnoakes:
We can set this up for all branches with a wildcard rule if we want to.

image

drewnoakes:
@Pilchie I think you're right. That looks like a better setting.

Pilchie:
Changed to a single `*` rule with both for now.  We will see if it breaks everything ;)

drewnoakes:
Created #4643 to track this.

drewnoakes:
So to test this out I just pushed a new branch `integration-tests` which I now cannot delete :)

drewnoakes:
I had to delete the rule in order to delete the branch. So a downside of this approach is that you can create a branch, but not delete it.

Issue #4632, opened 2019-03-01
Packages property page has a scroll bar even though there are no controls off screen


davkean:
Unlike all the other pages, we always show a scroll bar on this page even though there's nothing to scroll to:

![image](https://user-images.githubusercontent.com/1103906/53671357-1c78fa80-3cd2-11e9-856d-59439a6bc827.png)

etbyrd:
Fixed in #4645

PR #4645, opened 2019-03-11
Package property sizing fixes

etbyrd:
Simple fix for: https://github.com/dotnet/project-system/issues/4632

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
@@ -705,6 +705,9 @@
   <data name="&gt;&gt;LicenseLineLabel.ZOrder" xml:space="preserve">
     <value>17</value>
   </data>
+  <data name="LicenseUrlWarning.Dock" type="System.Windows.Forms.DockStyle, System.Windows.Forms">
+    <value>Fill</value>
+  </data>
   <data name="LicenseUrlWarning.Location" type="System.Drawing.Point, System.Drawing">
     <value>1154, 373</value>
   </data>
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
```diff
@@ -1828,7 +1828,7 @@
     <value>22</value>
   </data>
   <data name="TableLayoutPanel.Size" type="System.Drawing.Size, System.Drawing">
-    <value>2011, 917</value>
+    <value>1160, 900</value>
```
Pilchie: This doesn't seem right to me.  Shouldn't we be using auto-sizing?  Something like setting `DockMode` to `Dock.Fill`?

drewnoakes: Agree. @etbyrd let's have a look at this one together.

etbyrd:
Added some changes to fix not just the size of the page, but the formatting problem with the assembly version boxes.

<details>
  <summary>Whole page showing where the scroll bar starts now and assembly version fixes</summary>
  image1
</details>

<details>
  <summary>Approximately same content size with both horizontal and vertical scroll bars visible and nonuniform gap between assembly version boxes </summary>
 image2
</details>

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
```diff
@@ -705,14 +705,17 @@
   <data name="&gt;&gt;LicenseLineLabel.ZOrder" xml:space="preserve">
     <value>17</value>
   </data>
+  <data name="LicenseUrlWarning.Dock" type="System.Windows.Forms.DockStyle, System.Windows.Forms">
+    <value>Fill</value>
+  </data>
```
drewnoakes: Is this necessary?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
drewnoakes: @Pilchie I believe this has been addressed now.

PR #4646, opened 2019-03-13
Opt into operation progress

davkean:
**Background**

In the old day of the legacy project system, after solution load had completed - the project was guaranteed to have finished running a IntelliSense & references design-time build. Also modifications to the projects also were guaranteed to have finished running design-time builds by the time the API that modified the project had returned. 

This was a simple life and meant designers could open and query the project for stuff that required a design-time (such as CodeModel or references) without worrying if the data was present. Either it was present and the design-time build completed, or it wasn't and the design-time build failed. There was no querying an in-between "in progress" state.

In turns out design-time builds are slow, this resulted in large solution load times and delays while the UI blocked waiting for the build to complete. We made some advancement over the years, by reading cached results if nothing had changed between solution loads and by being smarter around how we group builds, but in the end, builds are an open system and large solutions tend to, over time, trend towards long design-time builds as custom and NuGet package targets are added that contribute to build time.

**Async Design-Time Builds**
In the new project system (based on CPS), to avoid these UI delays and long solution load times, we stopped blocking on them. Solution load and modifications to the project, no longer guarantee that design-time builds have completed.

This presents a problem; how to consumers of this project data made aware when the design-time build has finished and they can query the project? For features that are aware of CPS projects can opt into CPS-only APIs to query this data asynchronously, but the vast majority of features are project-system agnostic and this presents a large burden on the consumer. On top of this, legacy project system is planning on moving to a similar mechanism to CPS and stop blocking on design-time builds - it doesn't have any asynchronous APIs, would be need to add the same set of APIs that CPS has?

**Operation Progress**
Enter operation progress. Operation progress is a new set of APIs (entry point via IVsOperationProgress/IVsOperationProgress2) that lets consumers easily query whether a solution in currently "in-progress" running design-time builds or other operations that need to be finished before a project can be queried for results. These APIs are currently very simplified, are in a prototype state and not designed for consumers that need very specific data. For now, a consumer can ask very simple questions; "is IntelliSense ready for the solution?" Roslyn IntelliSense, for example, will use this result to show an indication to the user that IntelliSense results are not complete, or block operations such as rename, etc.  Long term, we'll see more granular APIs that let us answer questions such as "Is IntelliSense ready for this project and its dependencies?" or "Is IntelliSense ready for this project and its consumers?" but for the prototype we're focusing on solution wide.

**CPS's version of operation progress** 

CPS provides its own set of APIs that let producers of "operations" that enable IntelliSense contribute to operation progress. It does it in very natural and CPS-like manner. Most services in CPS consume data from a set of inputs via data flow. Given these inputs have versions, "in-progress" state is easy to calculate; "is the version that I've processed, up-to-date with the latest version of that data"? This is all tracked via the IDataProgressTrackerService service. A given service registers a "data source" and simply notifies the returned registration every time that they process a version of their inputs and it worries about publishing progress/completion state to the solution-wide `IVsOperationProgress` service.

This is exactly what we've implemented for the language service; this PR informs (indirectly) informs operation progress about the state of pushing evaluation/design-time build results to Roslyn. 

This is in no way handles restore state, ie in unrestored state it takes two design-time builds for "System.Runtime/mscorlib" to be pushed to Roslyn. Designers, etc don't care about the in-between state after the first design-time build,  this will be implemented in a followup.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextHostInstanceTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextHostInstanceTests.cs
@@ -255,12 +255,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             applyChangesToWorkspaceContext = applyChangesToWorkspaceContext ?? IApplyChangesToWorkspaceContextFactory.Create();
 
             return new WorkspaceProjectContextHostInstance(project,
-                                                    threadingService,
-                                                    tasksService,
-                                                    projectSubscriptionService,
-                                                    workspaceProjectContextProvider,
-                                                    activeWorkspaceProjectContextTracker,
-                                                    ExportFactoryFactory.ImplementCreateValueWithAutoDispose(() => applyChangesToWorkspaceContext));
+                                                           threadingService,
+                                                           tasksService,
+                                                           projectSubscriptionService,
+                                                           workspaceProjectContextProvider,
+                                                           activeWorkspaceProjectContextTracker,
+                                                           ExportFactoryFactory.ImplementCreateValueWithAutoDispose(() => applyChangesToWorkspaceContext));
         }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextHostInstanceTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextHostInstanceTests.cs
@@ -253,6 +253,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             activeWorkspaceProjectContextTracker = activeWorkspaceProjectContextTracker ?? IActiveEditorContextTrackerFactory.Create();
             workspaceProjectContextProvider = workspaceProjectContextProvider ?? IWorkspaceProjectContextProviderFactory.ImplementCreateProjectContextAsync(IWorkspaceProjectContextAccessorFactory.Create());
             applyChangesToWorkspaceContext = applyChangesToWorkspaceContext ?? IApplyChangesToWorkspaceContextFactory.Create();
+            IDataProgressTrackerService dataProgressTrackerService = IDataProgressTrackerServiceFactory.Create();
 
             return new WorkspaceProjectContextHostInstance(project,
                                                            threadingService,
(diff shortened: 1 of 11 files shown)


davkean:
tag @lifengl

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextHostInstanceTests.cs
```diff
@@ -255,12 +255,12 @@ private WorkspaceProjectContextHostInstance CreateInstance(ConfiguredProject pro
             applyChangesToWorkspaceContext = applyChangesToWorkspaceContext ?? IApplyChangesToWorkspaceContextFactory.Create();
 
             return new WorkspaceProjectContextHostInstance(project,
-                                                    threadingService,
-                                                    tasksService,
-                                                    projectSubscriptionService,
-                                                    workspaceProjectContextProvider,
-                                                    activeWorkspaceProjectContextTracker,
-                                                    ExportFactoryFactory.ImplementCreateValueWithAutoDispose(() => applyChangesToWorkspaceContext));
+                                                           threadingService,
```
Pilchie: This is why I prefer to put the first parameter on the next line indented by 4 when wrapping parameters.  That way it never has to move if the method is renamed.

jmarolf: The diff still looks rather reasonable with `?w=1` turned on.

Pilchie: Yeah, but history/annotate view is littered with people adjusting whitespace. ?w=1 is a bandaid.

adrianvmsft:
Change looks good to me.
Can't comment much on how well it works - it depends on the input data source versions - that they are present in the data source registry so that the Data Tracker Service can find them.

davkean:
I've tested this via `set CPSEnableOperationProgress=true`, and it seems to work, "IntelliSense" isn't considered completed until the design-time build results arrive. There's a bit of a experience gab in opening large solutions that @lifengl is looking at.

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextHostInstanceTests.cs
davkean: Do you suggest that we put the first parameter on a newline? Or just the second parameter and later?

drewnoakes: I interpreted @Pilchie's comment as meaning if you wrap, there's a newline before the first parameter. Likewise for argument lists. I might be biased because that's my preferred way of formatting such lists (also to minimise white space churn and maintenance).

It's also consistent when nesting calls (in rare cases where that's clearer than using temporaries) such as:

```c#
Method1(
    arg1,
    Method2(
        arg2,
        arg3),
    arg4);
```

Pilchie: What @drewnoakes said.  That way if the method is renamed, qualified with this, etc, the indentation of all the parameter/argument lines doesn't need to be adjusted.

davkean: Yep sounds good. Roslyn has a refactoring that does this, so its even easier.

PR #4652, opened 2019-03-18
Make outputs of NuGet nomination immutable

davkean:
In preparation for moving NuGet nomination to operation progress, I'm moving it to a data source/snapshot model. In preparation for that, made all outputs immutable.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/Snapshots/ImmutablePropertyCollection.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
+{
+    /// <summary>
+    ///     Abstract immutable collection that supports lookup by index and name.
+    /// </summary>
+    internal abstract class ImmutablePropertyCollection<T> : IEnumerable where T : class
```
davidwengier: Any reason not to implement `IEnumerable<T>` instead?

davkean: Extra implementation that will never be called; the `IVsXXX` collections only use IEnumerable.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/ProjectRestoreInfoBuilder.cs
```diff
@@ -119,21 +107,20 @@ private static IVsReferenceItems GetReferences(IImmutableDictionary<string, IImm
 
         private static IVsReferenceItems GetProjectReferences(IImmutableDictionary<string, IImmutableDictionary<string, string>> projectReferenceItems)
         {
-            IVsReferenceItems referenceItems = GetReferences(projectReferenceItems);
+            // NuGet expects each "<ProjectReference />" to have metadata item "ProjectFileFullPath" that represents
+            // the full path of the project reference, just use the built-in "FullPath" metadata for that.
+            IImmutableDictionary<string, IImmutableDictionary<string, string>> replacementProjectReferenceItems = projectReferenceItems;
```
drewnoakes: Using a builder here would likely involve fewer allocations. Might not matter if there aren't likely to be many changes.

davkean: This code is deleted in a latter branch anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/ProjectRestoreInfoBuilder.cs
```diff
@@ -26,8 +26,8 @@ internal static IVsProjectRestoreInfo Build(IEnumerable<IProjectVersionedValue<I
 
             string msbuildProjectExtensionsPath = null;
             string originalTargetFrameworks = null;
-            var targetFrameworks = new TargetFrameworks();
-            var toolReferences = new ReferenceItems();
+            var targetFrameworks = ImmutableDictionary.Create<string, IVsTargetFrameworkInfo>();
+            var toolReferences = ImmutableDictionary.Create<string, IVsReferenceItem>();
```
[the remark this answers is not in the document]
davkean: Good call on comparer, I was just trying to maintain (incorrect) behavior, but I could have done that with an explicit comparer.

This class is basically thrown away in latest code I'm working in.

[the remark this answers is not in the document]
davkean: I'm going to do this in a followup.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/ProjectRestoreInfoBuilder.cs
davkean: Going to do this in a followup.

PR #4657, opened 2019-03-22
Return values for VSHPROPID_DesignerFunctionVisibility and VSHPROPID_DesignerVariableNaming

### LINKED ISSUE #4478: Return values for VSHPROPID_DesignerFunctionVisibility and VSHPROPID_DesignerVariableNaming
To mimic legacy, we should return values for VSHPROPID_DesignerFunctionVisibility and VSHPROPID_DesignerVariableNaming which control the naming/visibility of WinForm generated controls.

VB:
```
VSHPROPID_DesignerFunctionVisibility -> VSDFV_Friend
VSHPROPID_DesignerVariableNaming -> VSDVN_VB
```
C#:
```
VSHPROPID_DesignerFunctionVisibility -> VSDFV_Private
VSHPROPID_DesignerVariableNaming -> VSDVN_Camel
```

drewnoakes:
~First pass at fixing #4478. My goal is to learn more about properties.~

Fixes #4478.

Sets value of two known properties used by designer code generation that determine the visibility and naming style of generated fields.

Can be merged once https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/180589 completes.

### DIFF
--- a/docs/well-known-project-properties.md
+++ b/docs/well-known-project-properties.md
@@ -188,6 +188,8 @@ Specifies the designer function access level (for example, `InitializeComponent(
 | _Private_ | The designer function has private visibility.                 |
 | _Public_  | The designer has public visibility (currently not supported). |
 
+These values map to members of the [`VSDESIGNER_FUNCTIONVISIBILITY`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.vsdesigner_functionvisibility) enum in the VS SDK.
+
 ##### __Example__
 ``` XML
 <PropertyGroup>
@@ -208,6 +210,8 @@ Specifies the naming convention used by the designer.
 |---------| ---------------|
 | _Camel_ | Names use camel case (e.g. `checkBox1`).       |
 | _VB_    | Names use VB / Pascal case (e.g. `CheckBox1`). |
+
+These values map to members of the [`VSDESIGNER_VARIABLENAMING`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.vsdesigner_variablenaming) enum in the VS SDK.
 
 ##### __Example__
 ``` XML
--- a/docs/well-known-project-properties.md
+++ b/docs/well-known-project-properties.md
@@ -122,14 +122,14 @@ Specifies whether `CoreCompile` should skip compiler execution.
 
 This property is helpful when used with the [ProvideCommandLineArgs](#providecommandlineargs_(bool)) property.
 
-
-
 ##### __Example__
 ``` XML
   <PropertyGroup>
     <RunPostBuildEvent>Always</RunPostBuildEvent>
   </PropertyGroup>
 ```
+
+---
 
 ## __Design-Time Properties, Items and Item Metadata__
 These properties, items and item metadata are used for solely for Visual Studio and design-time purposes, and have no influence on the resulting build.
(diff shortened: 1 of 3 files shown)


davidwengier:
As far as your goal of learning about properties, this is a good start, and you've defined the properties correctly and the documentation update is a very nice touch I honestly wouldn't have thought of.

Unfortunately it's largely in the wrong repo 😊 These new properties need to exist so that we can communicate back to CPS, because CPS answers the call for them [from the hierarchy](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.VS.Implementation%2FPackage%2FNodes%2FProjectNode.Hierarchy.cs&version=GBmaster&line=386&lineStyle=plain&lineEnd=390&lineStartColumn=1&lineEndColumn=1). If we were planning on providing UI for these then the situation would be different, but I think these are pretty low-value as far as customizing goes, so really the only thing we need to do is set their values in our targets files. If you look at one of the other properties defined in this spot in our design time targets, like `CmdUIContextGuid` for example, you'll see no references to them in our repo.

Where the properties do need to exist though is in the [ConfigurationGeneral rule in CPS](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2Fshared%2FRules%2Fgeneral.xaml&version=GBmaster) so that it can read their values at the above mentioned spot in ProjectNode.Hierarchy.cs. A good example to look at on the CPS side would be the [TargetRuntime](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2Fshared%2FRules%2Fgeneral.xaml&version=GBmaster&line=66&lineStyle=plain&lineEnd=71&lineStartColumn=1&lineEndColumn=1) property and [where it is mapped to a VS SDK enum](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.VS.Implementation%2FPackage%2FNodes%2FProjectNode.Hierarchy.cs&version=GBmaster&line=1135&lineStyle=plain&lineEnd=1158&lineStartColumn=1&lineEndColumn=1) (though in this case I don't think you need to worry about `ProjectCapabilitiesContext`).

Strictly speaking these could also just be string properties and we could specify the values as per the enum in the SDK itself, as they're unlikely to be customized by anyone (I don't think they could be in legacy?), but it doesn't really cost anything to define some nicer values, and certainly doing string comparisons is better than Enum.Parse for perf.

Finally, its probably good to link to the VS SDK doco from our doco so people know what the values map to.

drewnoakes:
Thanks @davidwengier, that's really useful and explains why I couldn't see how such properties would be wired up to the hierarchy. I'll use your points and dig into CPS as time permits during my travels over the coming week or so.

drewnoakes:
CPS PR is up: https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/180589

drewnoakes:
This is ready for review @davidwengier @davkean.

docs/well-known-project-properties.md
```diff
@@ -173,6 +173,49 @@ Specifies whether the contents of the _Application Designer_ folder are only vis
 <PropertyGroup>
 ```
 
+#### __DesignerFunctionVisibility (enum)__
```
davidwengier: Consider linking to the equivalent VS SDK enums from docs.microsoft.com, and mentioning that these values map directly to them.

Issue #1607, opened 2017-02-23
Create dependencies node integration tests


johnbeisner:
Ask @abpiskunov about what scenarios needs to be tested.

One thing we need to test is that analyzers and diagnostics show properly when multiple projects are in the same directory. We broke this in https://github.com/dotnet/roslyn/pull/18888 and fixed it in https://github.com/dotnet/roslyn/pull/19188.

srivatsn:
@abpiskunov you are working on this - correct?

abpiskunov:
Actually no, i am working unit tests right now and then will do corresponding changes for web package nodes. Will not have time for integration tests.. Could  you assign to some one else on your team please? I can suggest scenarios etc ..

MattGertz:
Mondo escrow triage:  test-only, moving to 15.4 to get it off the radar,

tmeschter:
We're in the process of re-evaluating our integration test framework. This will be addressed in 15.6 at the earliest.

drewnoakes:
I've started writing integration tests for the dependencies node.

This required some changes to Apex, submitted in https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/169060.

That PR has been merged, but the process that pushes packages to the MyGet feed we consume broke as part of a migration to cloud build (as I understand it). @javierdlg is looking into this.

drewnoakes:
The Apex changes are now available on myget. Draft PR #4658 includes them and will close this issue once completed.

drewnoakes:
Closing this as #5251 was merged.

PR #4658, opened 2019-03-22
Dependency node integration tests

drewnoakes:
First cut of integration tests for the dependencies node. Relates to #1607.

These tests build up a solution/project(s)/global.json on disk, load the solution, then inspect the dependencies nodes.

See `DependencyNodeIntegrationTests` for how such tests are authored.

~🚧 Requires changes to Apex which have been merged but are not yet available via public package. This can be tested manually using the hand-built, unsigned package at \\mlangfs1\public\drnoakes~


drewnoakes:
Thanks to @javierdlg for publishing my Apex changes via [myget](https://vside.myget.org/feed/vs-impl/package/nuget/Microsoft.Test.Apex.VisualStudio).

jmarolf:
@drewnoakes resolve this build dependency issues and I think this looks good!

jmarolf:
@drewnoakes please retarget to feature/integration-tests

drewnoakes:
@jmarolf I just rebased on your branch. Let's see how this goes.

PR #4666, opened 2019-03-28
Convert PackageRestoreInitiator usage into -> data sources

davkean:
## Overview

In preparation for operation progress, this converts PackageRestoreIniatiator into data sources that publish via data flow. This is just prep work and does not yet opt into operation progress which will come in a future PR.

PackageRestoreInitiator/PackageRestoreInitiatorInstance has been split into three components;

**PackageRestoreConfiguredDataSource**: Which contains logic for producing restore information from a single configuration/configured project
**PackageRestoreUnconfiguredDataSource**: Which contains logic for merging the restore information from above data sources using "implicitly active" configs and resolving conflicts.
**PackageRestoreInitiator**: Responsible for taking Unconfigured data source and pushing it to NuGet, blocking build, rename, unload until restore has finished.

I've tried to make this readable as possible (as much as you can with data flow), with comments around usage I found to be a bit confusing. Please tell if you want me to add more, or confused about a particular line.

## Behavioral changes

- We now report "user" faults (yellow bar) when configurations have conflicting changes, such a DotNetCliToolReference with different versions, or different TargetFrameworks property.
- When data flow is falling behind, such as when significant changes have been made to a project (say I switch back and forth between the two branches) , we just push the latest data. Previously we'd publish every single version in between, resulting in unneeded restores.

## TODO

- [x] Tests - dataflow usage is really hard to test, trying to figure out a great approach for this.
- [x] Fix the synchronization of configs so that agree on a view of the "current project"; currently there is none, we just grab whatever version they've last published which they might not agree on.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreUnconfiguredDataSource.cs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        private static IVsTargetFrameworks GetAllTargetFrameworks(IEnumerable<ProjectRestoreUpdate> updates)
+        {
+            IEnumerable<IVsTargetFrameworkInfo> targetFrameworks = updates.SelectMany(t => t.RestoreInfo.TargetFrameworks.Cast<IVsTargetFrameworkInfo>());
+
+            return new TargetFrameworks(targetFrameworks);
+        }
+
+        //// We are taking source blocks from multiple configured projects and creating a SyncLink to combine the sources.
```
davkean: This is tracked by one of the TODOs above.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/Snapshots/RestoreLogger.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.VisualStudio.ProjectSystem.Logging;
+using NuGet.SolutionRestoreManager;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
+{
+    internal static class RestoreLogger
+    {
+        public static void BeginNominateRestore(IProjectLogger logger, string fullPath, IVsProjectRestoreInfo projectRestoreInfo)
```
davkean: This is a copy and paste, with exception to the types, I changed usage from concrete types -> interfaces.

davkean:
tag @lifengl

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/IPackageRestoreConfiguredDataSource.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
+{
+    /// <summary>
+    ///     Represents the data source of metadata needed for restore operations for an individual <see cref="ConfiguredProject"/>.
+    ///     This will be later combined with other implicitly active <see cref="ConfiguredProject"/> instances within a 
+    ///     <see cref="UnconfiguredProject"/> to provide enough data to restore the entire project.
+    /// </summary>
+    internal interface IPackageRestoreConfiguredDataSource : IProjectValueDataSource<ProjectRestoreUpdate>
```
lifengl: do you want to include ProjectContractAttribute?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DataFlowUtilities.cs
```diff
@@ -205,6 +205,92 @@ public static IDisposable LinkToAsyncAction<T>(this ISourceBlock<T> source, Func
...
+        ///     The transformed source block and a disposable value that terminates the link.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="source"/> is <see langword="null"/>.
+        ///     <para>
+        ///         -or-
+        ///     </para>
+        ///     <paramref name="transform"/> is <see langword="null"/>.
+        /// </exception>
+        public static DisposableValue<ISourceBlock<TOut>> TransformWithNoDelta<TOut>(this ISourceBlock<IProjectVersionedValue<IProjectSubscriptionUpdate>> source, Func<IProjectVersionedValue<IProjectSubscriptionUpdate>, TOut> transform, bool suppressVersionOnlyUpdates = true, IEnumerable<string> ruleNames = null)
```
lifengl: it is easy to make a mistake to turn on "suppressVersionOnlyUpdates" to cause dead locks and other problems (with operation progress feature).  My suggestion is that we don't provide the default value (which is true), and force the developer to think about it and make the right decision whenever they are using it.  It also makes it easier to code review (so you will review the place where it is used to see whether this option is set correctly.)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreConfiguredDataSource.cs
```diff
@@ -0,0 +1,82 @@
...
+            get { return _project.UnconfiguredProject; }
+        }
+
+        protected override IDisposable LinkExternalInput(ITargetBlock<RestoreUpdate> targetBlock)
+        {
+            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _projectSubscriptionService.JointRuleSource;
+
+            // Transform the changes from evaluation/design-time build -> restore data
+            DisposableValue<ISourceBlock<RestoreUpdate>> transformBlock = source.SourceBlock
+                                                                                .TransformWithNoDelta(update => update.Derive(u => CreateRestoreUpdate(u.CurrentState)),
```
lifengl: You cannot skip version only changes.  I guess....

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreUnconfiguredDataSource.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                ReportDataSourceUserFault(new Exception(string.Format(CultureInfo.CurrentCulture, VSResources.Restore_DuplicateToolReferenceItems, existingReference.Name)),
+                                          ProjectFaultSeverity.LimitedFunctionality,
+                                          ContainingProject);
+            }
+        }
+
+        private static IVsTargetFrameworks GetAllTargetFrameworks(IEnumerable<ProjectRestoreUpdate> updates)
+        {
+            IEnumerable<IVsTargetFrameworkInfo> targetFrameworks = updates.SelectMany(t => t.RestoreInfo.TargetFrameworks.Cast<IVsTargetFrameworkInfo>());
```
lifengl: do we want to retain a LinQ delegation inside our data structure? this can produce more complex in memory structure. and you rerun the queue every time the property is used. maybe getting the result out can be a simpler solution?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/Snapshots/ReferenceItemEqualityComparer.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
+{
+    internal class ReferenceItemEqualityComparer : EqualityComparer<IVsReferenceItem>
+    {
+        public readonly static ReferenceItemEqualityComparer Instance = new ReferenceItemEqualityComparer();
+
+        public override bool Equals(IVsReferenceItem x, IVsReferenceItem y)
+        {
+            if (!StringComparers.ItemNames.Equals(x.Name, y.Name))
```
lifengl: handle null ref?

lifengl:
a few comments... like to see the version story after you finish the "merging' work.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreConfiguredDataSource.cs
davkean: Why? I can ship version changes up until I start trying to coordinate the versions for operation progress or for the Unconfigured merge.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreUnconfiguredDataSource.cs
davkean: The LINQ query is realized in the the TargetFrameworks constructor, it doesn't outlive this method.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreConfiguredDataSource.cs
davkean: Ah, of course not. I just realized why not - we're trying to coordinate these versions upstream and unless config publishes every version, they may never agree.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DataFlowUtilities.cs
davkean: Fixed this - we are used to be ending of the chain so my initial feeling was that this should be true by default, but this is a transform - and hence likely in the middle of a chain. Agree and fixed this.

davkean:
@lifengl This is done, I've fixed the merging here: https://github.com/dotnet/project-system/pull/4666/files#diff-2d96e0155a799f1b5dd0fd6c836e28c9R28. Note, this just merges all other data sources except ConfiguredProjectVersion/Identity, which is our previous behavior. I think this long term we'll want to coordinate all inputs that make up an evaluation; ie all ProjectXmlVersion of all props/targets (not just the project file) and probably other things (timestamps of project.assets.json, etc).

davkean:
Hmm, testing this end-to-end it needs some tweaks. The new behavior causes us to over nominate NuGet meaning a single change might result in us nominating them 10 times in a row, which they'll happily restore 10 times.

davkean:
I'm seeing multiple versions of the same effective data source coming through:
```
  | Name | Value | Type
-- | -- | -- | --
▶ | [0] | {[2 (ProjectService), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [1] | {[3 (VsIntegration), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [2] | {[4 (ActiveProjectConfiguration), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [3] | {[7 (AdditionalRuleDefinitionsVersion), 2]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [4] | {[9 (InconsistentDataVersion), 0]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [5] | {[10 (GlobalProjectCollectionGlobalProperties), 3]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [6] | {[47 (CapabilitiesInheritedByProject), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [7] | {[51 (ProjectCapabilitiesFromImportTargetProvider), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [8] | {[52 (ProjectExtensionsCapabilitiesProvider), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [9] | {[56 (BuildSnapshotServiceRequestVersion), 7]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [10] | {[57 (DataSourceImportor), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [11] | {[59 (UnwrapChainedProjectValueDataSourceBase), 4]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [12] | {[60 (ProjectConfigurationsDataSource), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [13] | {[61 (LoadedConfiguredProjectsDataSource), 3]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [14] | {[63 (BuildSnapshotServiceRequestVersion), 3]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable> <!-- Why ?
▶ | [15] | {[64 (DataSourceImportor), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [16] | {[68 (BuildSnapshotServiceRequestVersion), 4]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable> <!-- Why?
▶ | [17] | {[69 (DataSourceImportor), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [18] | {[73 (BuildSnapshotServiceRequestVersion), 3]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>
▶ | [19] | {[74 (DataSourceImportor), 1]} | System.Collections.Generic.KeyValuePair<Microsoft.VisualStudio.ProjectSystem.NamedIdentity, System.IComparable>

```

davkean:
Oh that's just each config's version of their BuildSnapshotServiceRequestVersion - it's given a unique identity.

davkean:
Okay, after running this through the mile there's still a few changes to make:

- Because we're no longer relying on [AnyChanges](http://index/?query=PackageRestoreInitiator&rightProject=Microsoft.VisualStudio.ProjectSystem.Managed.VS&file=ProjectSystem%5CVS%5CNuGet%5CProjectRestoreInfoBuilder.cs&line=22) we are nominating way more than we were previously - NuGet actually does work in that case every those the data is the same. We should only nominate if the output is actually different.
- We're prevent overlapping of NuGet restore by blocking NominateProjectAsync preventing new data to come through.  This is causes us to wait on an outdated restore even though we know we have new data that might invalidate it. We should cancel any previous nominations when new data comes in.
- We seem to have a bug where one of the configurations we have when switching a project from single target framework -> multiple target framework while open causes TargetFrameworkMoniker to be empty resulting in a user fault to be reported.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/Snapshots/RestoreComparer.ProjectPropertyEqualityComparer.cs
```diff
@@ -0,0 +1,30 @@
...
+
+                if (!StringComparers.PropertyNames.Equals(x.Name, y.Name))
+                    return false;
+
+                return StringComparers.PropertyValues.Equals(x.Value, y.Value);
+            }
+
+            public override int GetHashCode(IVsProjectProperty obj)
+            {
+                return obj.Name.GetHashCode();
```
davidwengier: Just confirming.. this means GetHashCode can return the same value for two different objects but Equals will return false for those same two objects...

davkean: Which is a "valid" GetHashCode implementation. GetHashCode determines the bucket you end up in, after which maps/hashtables call Equals on each item in the bucket to find the actual item. If Name clashes were common, then this would be perf issue, the expectation is that they are not.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreUnconfiguredDataSource.DropConfiguredProjectVersionDataSource.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS.NuGet
+{
+    internal partial class PackageRestoreUnconfiguredDataSource
+    {
+        /// <summary>
+        ///     A <see cref="IProjectValueDataSource{T}"/> that drops <see cref="ProjectDataSources.ConfiguredProjectIdentity"/> and 
+        ///     <see cref="ProjectDataSources.ConfiguredProjectVersion"/> versions from each value of the original
+        ///     <see cref="IProjectValueDataSource{T}"/>.
+        /// </summary>
+        private sealed class DropConfiguredProjectVersionDataSource<T> : ChainedProjectValueDataSourceBase<T>
```
lifengl: We have a cheaper implementation (TransformLinkTo inside CPS), but haven't made it public. For a minor version transform, it makes sense to use that one (instead of creating a full size dataflow block).  Let me know if you need, and we can update our side.

PR #4667, opened 2019-03-28
smallest possible change to fix race in DotNetVersionCompatibility

jmarolf:
Fixes internal bugs
- [819419](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/819419)
- [739377](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/739377)

Now that our package is initialized in the background there is no gurnatee that we will be initalized as soon as the solution is opened.  We may be initialized at any point after the solution is opened.

This change makes sure that as soon as we are initialized, if the solution has finished opening, we check version compatibilty.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
@@ -102,7 +102,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
                 isFullyLoadedBool)
             {
                 _solutionOpened = true;
-                await CheckCompatibilityAsync();
+                _threadHandling.Value.RunAndForget(() => CheckCompatibilityAsync(), unconfiguredProject: null);
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
@@ -5,7 +5,6 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Threading.Tasks;
 
@@ -103,6 +102,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
                 isFullyLoadedBool)
             {
                 _solutionOpened = true;
+                await CheckCompatibilityAsync();
             }
         }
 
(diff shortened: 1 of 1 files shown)


davkean:
If I understand correctly, this changes the behavior to:

- If our package happens to get loaded before solution load, we block solution load on this checking. 
- If it does not, we block our package load on this. 
- If we go down the latter path, no longer prevent the solution from being unloaded underneath us.

This will going to:

1) Slow down solution load - the check is not parallelized 
2) Slow down our package load (I'm not sure why we do this?)
3) Probably crash VS if the solution is unloaded while we're checking.

If this is for 16.1, I'd prefer a different approach to move this to an individual check during a project that makes use of IVsRetargetProject, so that we make use of parallelized loading of projects.

davkean:
It also looks like this code has a race where projects can be added while we're checking during our package load that we can miss. IVsRetargetProject would not run into that situation.

davkean:
The other thing, if we our package load misses solution load, won't we run into the same problem?

jmarolf:
> If I understand correctly, this changes the behavior to:
> - If our package happens to get loaded before solution load, we block solution load on this checking.
> - If it does not, we block our package load on this.
> - If we go down the latter path, no longer prevent the solution from being unloaded underneath us.

This is not correct, there is no change in behavior here for when our package loads and the solution has not yet finished loading. The only change is that when out package is loaded we block out package load on this change.  I can fix that to be fire-and-forget though

@davkean I am 100% on board with re-writing all of this code.  I think most of the assumptions for how `DotNetCoreProjectCompatibilityDetector` was supposed to work are not true any longer.  Is there anything wrong with just having this check occur per-project-load in CPS instead of haveing a service try and correctly synchronize VS solution load events across various async operations?

davkean:
@jmarolf Apologies, my bad I read the diff wrong - I thought we changed from RunAsync -> Run in solution load. If we change package load to fire and forgot, I'm good with the change. 

As part of the SDK acquisition code, I imagine we'll rewrite the code to use IVsRetargetProject infrastructure (which has prerequisite checking) which is per project.

jmarolf:
@davkean please take a look when you have the chance.  I changed the compatibility check to ge fire-and-forget.

davidwengier:
@jmarolf I'm seeing an error when opening ProjectSystem.sln in d16.1stg, 28802.115. Not sure if its caused by this or fixed by this.

image

Stack is:
```
SetSite failed for package [ManagedProjectSystemPackage]Source: &apos;Microsoft.VisualStudio.ProjectSystem.Implementation&apos; Description: CancelOnUnload cannot be satisfied without a project context provided.
Parameter name: options
System.ArgumentException: CancelOnUnload cannot be satisfied without a project context provided.
Parameter name: options
   at Microsoft.VisualStudio.ProjectSystem.ProjectMultiThreadedService.Fork(Func`1 asyncAction, JoinableTaskFactory factory, UnconfiguredProject unconfiguredProject, ConfiguredProject configuredProject, ErrorReportSettings watsonReportSettings, ProjectFaultSeverity faultSeverity, ForkOptions options)
   at Microsoft.VisualStudio.ProjectSystem.FaultExtensions.RunAndForget(IProjectThreadingService threadingService, Func`1 asyncAction, UnconfiguredProject unconfiguredProject, ProjectFaultSeverity faultSeverity, ForkOptions options)
   at Microsoft.VisualStudio.ProjectSystem.VS.DotNetCoreProjectCompatibilityDetector.&lt;InitializeAsync&gt;d__27.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Packaging.ManagedProjectSystemPackage.&lt;InitializeAsync&gt;d__7.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Shell.AsyncPackage.&lt;&gt;c__DisplayClass19_0.&lt;&lt;Microsoft-VisualStudio-Shell-Interop-IAsyncLoadablePackageInitialize-Initialize&gt;b__1&gt;d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Threading.JoinableTask.&lt;JoinAsync&gt;d__68.MoveNext()
```

davkean:
This would have introduced it, please file a bug.

PR #4668, opened 2019-03-28
Do not warn user of incompatible SDK version if they are on a preview

jmarolf:
fixes internal bug
- [819411](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/819411)

If the user is on a Preview version of VS, do not warn them about Preview SDK useage

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
@@ -32,6 +32,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         private const string SupportedLearnMoreFwlink = "https://go.microsoft.com/fwlink/?linkid=868064";
         private const string UnsupportedLearnMoreFwlink = "https://go.microsoft.com/fwlink/?linkid=866797";
         private const string SuppressDotNewCoreWarningKey = @"ManagedProjectSystem\SuppressDotNewCoreWarning";
+        private const string UsePreviewSdkKey = @"ManagedProjectSystem\UsePreviewSdk";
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -228,6 +230,17 @@ private async Task WarnUserOfIncompatibleProjectAsync(CompatibilityLevel compatL
                     if (settingsManager != null)
                     {
                         suppressPrompt = settingsManager.GetValueOrDefault(SuppressDotNewCoreWarningKey, defaultValue: false);
+
+                        // Do not warn users about pre-release versions if they have checked "Use Preview SDKs"
+                        suppressPrompt |= settingsManager.GetValueOrDefault(UsePreviewSdkKey, defaultValue: false);
```
Pilchie: I don't think this is what we want here in general.  I think we still want to be able to guide people to use something newer, even if they have checked the checkbox.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -228,6 +230,17 @@ private async Task WarnUserOfIncompatibleProjectAsync(CompatibilityLevel compatL
...
+                    }
+
+                    // If this is a pre-release version do not show the prompt for version compatibility
+                    var vsSetupConfig = new SetupConfiguration();
+                    ISetupInstance setupInstance = vsSetupConfig.GetInstanceForCurrentProcess();
+                    if (setupInstance is ISetupInstanceCatalog setupInstanceCatalog)
```
Pilchie: Similarly, if the user is on 16.1 Preview, but we want them to update to 16.2 Preview, I think we still want this.

Instead, I think we want `IsPrerelease()` to be an input into the json file.

jmarolf:
I am going to take a different approach here

PR #4670, opened 2019-03-28
Fix Service Fabric Application package crash

davidwengier:
When packaging or F5ing a Service Fabric Application VS errors complaining about COM definitions:

```
Type library exporter encountered an error while processing 'Microsoft.VisualStudio.ProjectSystem.VS.Interop.IVsAppId, Microsoft.VisualStudio.ProjectSystem.Managed.VS'. Error: Type 'IVsAppId' and type 'SVsAppId' both have the same UUID.
```

This resolves the problem, and since the two IIDs were the same, there was no need to have two anyway.

Fixes internal bug [823377](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/823377)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Shipped.txt
```diff
@@ -105,7 +105,6 @@ Microsoft.VisualStudio.ProjectSystem.VS.Interop.IVsAppId.Initialize() -> int
 Microsoft.VisualStudio.ProjectSystem.VS.Interop.IVsAppId.SetGuidProperty(int propid, ref System.Guid rguid) -> int
 Microsoft.VisualStudio.ProjectSystem.VS.Interop.IVsAppId.SetProperty(int propid, object var) -> int
 Microsoft.VisualStudio.ProjectSystem.VS.Interop.IVsAppId.SetSite(Microsoft.VisualStudio.OLE.Interop.IServiceProvider pSP) -> int
-Microsoft.VisualStudio.ProjectSystem.VS.Interop.SVsAppId
```
davidwengier: I don't know why these are public, but is it safe to assume nobody is depending on us?

davkean:
@davidwengier This is fixing the symptom, this is an extremely common pattern across the product - why is this failing now?

davidwengier:
Updated.

Root cause was that these interfaces were public, so when this DLL was called via COM the CLR tried to create a type library for it, which failed because of the duplicate IIDs. Previously the language specific DLLs were the only things loaded via COM, so this wasn't a problem.

PR #4671, opened 2019-03-28
Add an IFileWatchProviderDataSource

tmeschter:
Add an `IFileWatchProviderDataSource` for potential .editorconfig files.

Note that this is dependent on the changes in https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/171765 so you may want to review that first.

### DIFF
--- a/build/Packages.targets
+++ b/build/Packages.targets
@@ -97,8 +97,8 @@
     <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.1.163-pre" />
 
     <!-- Roslyn -->
-    <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="2.11.0-beta2-63529-05" />
-    <PackageReference Update="Microsoft.CodeAnalysis" Version="2.11.0-beta2-63529-05" />
+    <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="3.0.0" />
+    <PackageReference Update="Microsoft.CodeAnalysis" Version="3.0.0" />
     <PackageReference Update="Microsoft.VisualStudio.IntegrationTest.Utilities" Version="2.6.0-beta1-62113-02" />
 
     <!-- Analyzers -->
--- a/build/Packages.targets
+++ b/build/Packages.targets
@@ -13,6 +13,7 @@
       https://dotnet.myget.org/F/project-system-interop/api/v3/index.json;
       https://vside.myget.org/F/vs-impl/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
+      C:\Users\tomescht\Desktop\LocalPackageSource
     </RestoreSources>
   </PropertyGroup>
 
@@ -93,8 +94,8 @@
     <PackageReference Update="Microsoft.Test.Apex.VisualStudio" Version="16.0.28117-pre" />
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK" Version="16.1.39-pre-gd9225dddee" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.0.201-pre-g7d366164d0" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK" Version="16.1.59-pre-g5b7f85631b" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.1.59-pre-g5b7f85631b" />
 
     <!-- Roslyn -->
     <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="2.11.0-beta2-63529-05" />
--- a/build/Packages.targets
+++ b/build/Packages.targets
@@ -13,7 +13,6 @@
       https://dotnet.myget.org/F/project-system-interop/api/v3/index.json;
       https://vside.myget.org/F/vs-impl/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
-      C:\Users\tomescht\Desktop\LocalPackageSource
     </RestoreSources>
   </PropertyGroup>
 
@@ -94,8 +93,8 @@
     <PackageReference Update="Microsoft.Test.Apex.VisualStudio" Version="16.0.28117-pre" />
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK" Version="16.1.62-pre-gff7146ed84" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.1.62-pre-gff7146ed84" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK" Version="16.1.163-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.1.163-pre" />
 
     <!-- Roslyn -->
     <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="2.11.0-beta2-63529-05" />
@@ -116,10 +115,10 @@
     <PackageReference Update="NuGet.VisualStudio" Version="4.9.1" />
 
     <!-- MSBuild -->
-    <PackageReference Update="Microsoft.Build" Version="15.8.166"/>
-    <PackageReference Update="Microsoft.Build.Utilities.Core" Version="15.8.166"/>
-    <PackageReference Update="Microsoft.Build.Engine" Version="15.8.166"/>
-    <PackageReference Update="Microsoft.Build.Tasks.Core" Version="15.8.166"/>
+    <PackageReference Update="Microsoft.Build" Version="16.0.452"/>
+    <PackageReference Update="Microsoft.Build.Utilities.Core" Version="16.0.452"/>
+    <PackageReference Update="Microsoft.Build.Engine" Version="16.0.452"/>
+    <PackageReference Update="Microsoft.Build.Tasks.Core" Version="16.0.452"/>
 
     <!-- Libraries -->
     <PackageReference Update="Newtonsoft.Json" Version="9.0.1"/>
--- a/build/Packages.targets
+++ b/build/Packages.targets
@@ -94,8 +94,8 @@
     <PackageReference Update="Microsoft.Test.Apex.VisualStudio" Version="16.0.28117-pre" />
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK" Version="16.1.59-pre-g5b7f85631b" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.1.59-pre-g5b7f85631b" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK" Version="16.1.62-pre-gff7146ed84" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers" Version="16.1.62-pre-gff7146ed84" />
 
     <!-- Roslyn -->
     <PackageReference Update="Microsoft.VisualStudio.LanguageServices" Version="2.11.0-beta2-63529-05" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
@@ -5,16 +5,18 @@ using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS
 {
     [Export(typeof(IFileWatchProviderDataSource))]
+    [AppliesTo(ProjectCapability.DotNetLanguageService)]
     internal class PotentialEditorConfigDataSource : ProjectValueDataSourceBase<IFileWatchProvider>, IFileWatchProviderDataSource
     {
         private readonly ConfiguredProject _project;
 
         private int _version;
-        private IDisposable _projectRuleSourceLink;
+        private DisposableBag _disposables;
 
         /// <summary>
         /// The block that receives updates from the active tree provider.
@@ -28,7 +30,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         [ImportingConstructor]
         public PotentialEditorConfigDataSource(ConfiguredProject project)
-            : base(project.Services)
+            : base(project.Services, synchronousDisposal: true)
         {
             _project = project;
         }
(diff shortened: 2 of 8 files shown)


tmeschter:
This is the dotnet/project-system change that corresponds to https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/171765. The `IFileWatchProvider` and `IFileWatchProviderDataSources` and the service that imports/uses them are defined on the CPS side. This change contains the implementation of these interfaces to support potential .editorconfig files.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    [Export(typeof(IFileWatchProviderDataSource))]
```
tmeschter: Something is wrong either with how I'm exporting this type or importing it in the CPS change. The service that consumes this type imports it like so:

``` C#
[ImportMany(ExportContractNames.Scopes.ConfiguredProject, typeof(IFileWatchProviderDataSource))]
private OrderPrecedenceImportCollection<IProjectValueDataSource<IFileWatchProvider>> FileWatchProviderDataSources { get; set; }
```

but it always ends up with an empty collection and I'm not sure why.

davkean: You are missing an AppliesTo, this should be marked with DotNetLanguageService must likely.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                var fileWatchProvider = new PotentialEditorConfigFileWatchProvider(this, ruleSnapshot.Items.Keys);
+                _version++;
+                ImmutableDictionary<NamedIdentity, IComparable> dataSources = ImmutableDictionary<NamedIdentity, IComparable>.Empty.Add(DataSourceKey, DataSourceVersion);
+
+                await _broadcastBlock.SendAsync(new ProjectVersionedValue<IFileWatchProvider>(fileWatchProvider, dataSources));
+            }
+        }
+
+        private class PotentialEditorConfigFileWatchProvider : IFileWatchProvider
```
davkean: Nested classes go in separate files.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -185,6 +185,10 @@
       <Context>BrowseObject</Context>
     </PropertyPageSchema>
 
+    <PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)PotentialEditorConfigFiles.xaml">
+      <Context>File</Context>
```
davkean: "ProjectSubscriptionService", "File" will work - but that context is supposed to be for the project tree.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -335,6 +339,7 @@
   <!-- List of external files that trigger design-time builds when they are modified -->
   <ItemGroup>
     <AdditionalDesignTimeBuildInput Include="$(ProjectAssetsFile)" />   <!-- NuGet assets file -->
+    <AdditionalDesignTimeBuildInput Include="@(PotentialEditorConfigFiles)" /> <!-- Places .editorconfig files may appear -->
```
davkean: I think the conclusion that this wasn't needed right? 

Alternatively, we have a watcher watch "AdditionalDesignTimeBuildInputs" and we don't need one to watch PotentialEditorConfigFiles.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
```diff
@@ -0,0 +1,90 @@
...
+        private IReceivableSourceBlock<IProjectVersionedValue<IFileWatchProvider>> _publicBlock;
+
+        [ImportingConstructor]
+        public PotentialEditorConfigDataSource(ConfiguredProject project)
+            : base(project.Services)
+        {
+            _project = project;
+        }
+
+        public override NamedIdentity DataSourceKey { get; } = new NamedIdentity(nameof(PotentialEditorConfigDataSource));
```
davkean: @lifengl Are all data sources expected to produce their own name/version?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
davkean: This is what I expected to to move our assets file watch over to.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PotentialEditorConfigFiles.xaml
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Rule Name="PotentialEditorConfigFiles"
+      Description="Potential .editorconfig files"
```
davkean: Remove, this file will never appear in UI.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PotentialEditorConfigFiles.xaml
```diff
@@ -0,0 +1,15 @@
...
+      DisplayName="NamespaceImport"
```
davkean: Remove, this file will never appear in UI.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public override IComparable DataSourceVersion => _version;
+
+        public override IReceivableSourceBlock<IProjectVersionedValue<IFileWatchProvider>> SourceBlock => _publicBlock;
```
davkean: You are missing a call to EnsureInitialized() otherwise, your class never gets initialized.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public FileChangeKinds ChangeKinds => FileChangeKinds.Added | FileChangeKinds.Removed;
+
+        protected override void Initialize()
+        {
+            base.Initialize();
+
+            _projectRuleSourceLink = _project.Services.ProjectSubscription.ProjectRuleSource.SourceBlock.LinkToAsyncAction(ProcessProjectChanged, ruleNames: PotentialEditorConfigFiles.SchemaName);
```
davkean: I think you'll need to JoinUpstreamDataSources on this ProjectRuleSource, otherwise, if anyone ever blocks on UI thread on you and it needs to switch to the UI thread to complete, it will result in a deadlock. @lifengl?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
```diff
@@ -0,0 +1,90 @@
...
+        private IBroadcastBlock<IProjectVersionedValue<IFileWatchProvider>> _broadcastBlock;
+
+        /// <summary>
+        /// The public facade for the broadcast block.
+        /// </summary>
+        private IReceivableSourceBlock<IProjectVersionedValue<IFileWatchProvider>> _publicBlock;
+
+        [ImportingConstructor]
+        public PotentialEditorConfigDataSource(ConfiguredProject project)
+            : base(project.Services)
```
davkean: synchronousDisposable:true. You have are taking no locks in Dispose.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
tmeschter: I assume you mean I need to add an `AppliesTo` attribute to the `PotentialEditorConfigDataSource` type. I did so, but it is still never being instantiated. Where do I go from here to diagnose that issue?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
tmeschter: Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
tmeschter: I removed this for now.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PotentialEditorConfigFiles.xaml
tmeschter: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PotentialEditorConfigFiles.xaml
tmeschter: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
tmeschter: Added.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
tmeschter: I've add a `JoinUpstreamDataSources(_project.Services.ProjectSubscription.ProjectRuleSource)`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
tmeschter: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
tmeschter: On the CPS side I tried changing 
``` C#
[ImportMany(ExportContractNames.Scopes.ConfiguredProject, typeof(IFileWatchProviderDataSource))]
```
to just
``` C#
[ImportMany(typeof(IFileWatchProviderDataSource))]
```

and that worked. I don't understand what that argument to `ImportMany` does, however.

davkean: Ah, I missed the import definition.  You will still need [AppliesTo] regardless.

The first argument is the "contract name", the second argument is the "contract type" both have to be the same on both sides for the "contract to match". We pass the scopes (UnconfiguredProject, ConfiguredProject) as a "contract name" when a given contract type (IFileWatchProviderDataSource) might live in [multiple scopes](https://github.com/Microsoft/VSProjectSystem/blob/1c0a47aba5a22d3eb071dc097b73851bdeaf68db/doc/overview/scopes.md).

For example, CPS has a service called `IProjectAsynchronousTasksService` that has two different implementations; one that lives in every configuration, and one that lives in every project. They are distinguished by scope arguments to "contract name".

If the service `IFileWatchProviderDataSource` will live in different scopes, then it should have a scope argument, if not, then it does not need one.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PotentialEditorConfigDataSource.cs
tmeschter: No longer an issue; `IFileWatchProvider` has been replaced with a concrete `FileWatchProvider` at the CPS level.

build/Packages.targets
```diff
@@ -13,6 +13,7 @@
       https://dotnet.myget.org/F/project-system-interop/api/v3/index.json;
       https://vside.myget.org/F/vs-impl/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
+      C:\Users\tomescht\Desktop\LocalPackageSource
```
drewnoakes: Intentional?

tmeschter: Intentional but temporary. The current change depends on CPS packages that only exist on my local system. :-)

Once the CPS side is merged in to the proper VS branch I can update the change to use the real package version, and then I will remove this.

tmeschter:
@Pilchie This is ready for your approval.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -189,7 +189,7 @@
     <PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)Page.BrowseObject.xaml">
       <Context>BrowseObject</Context>
     </PropertyPageSchema>
-
+
```
Pilchie: Nit: Whitespace

Issue #4355, opened 2018-12-06
Support for COM References in .NET Core project system


AaronRobinsonMSFT:
The 'Add References' operation in Visual Studio has a tab for COM References that is presently not visible for .NET Core projects. For .NET Core 3.0, support for COM References will be included. Having tooling for adding COM references is desirable.

Additional details:
 - PIA - not embedded
 - No-PIA - embedded
 - .NET Core 3.0 support only
 - Target stake holder is Office

See https://github.com/dotnet/samples/tree/master/core/extensions/ExcelDemo for the .NET Core official COM example.

Additional context: dotnet/coreclr#20497, Microsoft/MSBuild#3986

cc @jeffSchwMSFT

jjmew:
@etbyrd Could you take a look at this on this sprint? Our partners would like this for .NET Core 3.0.

tmeschter:
@etbyrd Is there more work to be done for this or can it be closed?

etbyrd:
Sorry about that - the work for this was completed and merged. Closing.

PR #4677, opened 2019-04-02
Support Data Source Window for .NET Framework and .NET Core 3.0

### LINKED ISSUE #4479: Enable the Data Source Window for SDK projects
We need to enable the Data Source Window to work with SDK projects. Some of its functionality won't light up until the TempPE work is done, but not all of it, so its worth enabling separately.

We need to merge https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/160023?_a=overview before turning this on or the data binding UI will break.

davidwengier:
The CPS bug that blocked this has been resolved so no point not merging this.

Fixes #4479

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -118,6 +118,7 @@
     <!-- Capabilities for WPF and WinForms -->
     <ProjectCapability Include="WPF" Condition="'$(UseWPF)' == 'true'" />
     <ProjectCapability Include="WindowsForms" Condition="'$(UseWindowsForms)' == 'true'" />
+    <ProjectCapability Include="DataSourceWindow" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or '$(_TargetFrameworkVersionWithoutV)' >= '3.0'"/>
   </ItemGroup>
   
   <!-- Common Project System rules that override rules defined in msbuild. These are exact copy of the rules defined in msbuild. -->
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -118,7 +118,7 @@
     <!-- Capabilities for WPF and WinForms -->
     <ProjectCapability Include="WPF" Condition="'$(UseWPF)' == 'true'" />
     <ProjectCapability Include="WindowsForms" Condition="'$(UseWindowsForms)' == 'true'" />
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -118,6 +118,7 @@
     <!-- Capabilities for WPF and WinForms -->
     <ProjectCapability Include="WPF" Condition="'$(UseWPF)' == 'true'" />
     <ProjectCapability Include="WindowsForms" Condition="'$(UseWindowsForms)' == 'true'" />
+    <ProjectCapability Include="DataSourceWindow" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or '$(_TargetFrameworkVersionWithoutV)' >= '3.0'"/>
```
davkean: That will enable it for:

UWP
Portable
Anything over version 3.0.

davidwengier: I'm not familiar with how those work. Do they have different `TargetFrameworkIndentifier`s, or is there some other way to detect them?

Or should I just restrict this to `.NETFramework Or (.NETCoreApp And >= 3.0)`?

davkean: The latter.

davkean: I think we'll want to start having some structure coordination to these TFMs checks so that's easier to see what's enabled for a given TFM/version and to reduce the amount of conditions that need to be evaluated.

davidwengier: Yes, I agree, I wouldn't want to add another one of these. I think @etbyrd needs this same condition for #4355 so converting to a `<Choose>` probably makes sense at that point.

ssa3512:
How do I track the release of this feature? Datasources are the last item holding up the conversion of some of my projects to the new SDK format.

davidwengier:
@ssa3512 Unfortunately there isn't a great way to track the exact release date of individual GitHub issues but in general this will be available in a version 16.1 of Visual Studio.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
Pilchie: Late, but should this also be supported for any .NETStandard TFMs, or does it require stuff that's only in .NETCoreApp?

davidwengier: As far as I know the Data Sources Window is only useful as a tool to automatically create controls on a designer surface, so wouldn't make sense in .NET Standard

PR #4683, opened 2019-04-03
Replace ITelemetryService.PostFault with IProjectFaultHandlerService.ReportFaultAsync

davkean:
Make sure we send all faults throughs IProjectFaultHandlerService.ReportFaultAsync so that they are bucketed correctly (under VS's "VisualStudioNonFatalErrors2"), and provide a gold bar to avoid this back and forth: https://twitter.com/Nick_Craver/status/1113261109467910144.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
```diff
@@ -66,8 +65,9 @@ public async Task ReleaseProjectContextAsync(IWorkspaceProjectContextAccessor ac
             {
                 accessor.Context.Dispose();
             }
-            catch (Exception ex) when (_telemetryService.PostFault(TelemetryEventName.LanguageServiceInitFault, ex))
```
tmeschter: Are there any situations where `PostFault` would return `false` and we would let the exception escape?

Really what I'm asking is: are we sure we're preserving the same behavior around which exceptions we catch and which escape?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
```diff
@@ -20,20 +19,20 @@ internal partial class WorkspaceProjectContextProvider : IWorkspaceProjectContex
...
+        private readonly IProjectFaultHandlerService _faultHandlerService;
         private readonly ISafeProjectGuidService _projectGuidService;
         private readonly Lazy<IWorkspaceProjectContextFactory> _workspaceProjectContextFactory;
 
         [ImportingConstructor]
         public WorkspaceProjectContextProvider(UnconfiguredProject project,
                                                IProjectThreadingService threadingService,
                                                ISafeProjectGuidService projectGuidService,
-                                               ITelemetryService telemetryService,
+                                               IProjectFaultHandlerService faultHandlerService,
```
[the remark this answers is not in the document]
davkean: Great questions.

It's a CPS service, however, we influence it via https://github.com/dotnet/project-system/blob/53349cf0e738ab34164e8373224d2308ab87bff4/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/FaultExtensions.cs#L18.
Yes it does. We have zero hits on the current usage, this moves it to how the majority of the product is tracking NFW, under a single event, but with a customized component. The watsons are grouped by buckets.
We have a customized component that lets us distinguish our faults from other faults. Similar to crashes we distinguish language service interaction from other kinds of faults in our code base by stacks, not by events.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
davkean: No, it always return true.

However, there is a subtle behavior change here; before we'd PostFault before stack was unwound, which meant if we got a dump we'd have no run any finallys at this point. The thing is though, PostFault doesn't actually ever create a dump - so that was pointless anyway.

I want to way to come up with a way that ProjectFaultHandlerService can be used in an exception filter, but it requires thought because it's all async all the way down.

PR #4684, opened 2019-04-03
Infra/move project system to new pools

jmarolf:
dnc-eng has changed how they do pools.  This moves the yaml file for CI into the `build` folder and makes the needed changes to the pipeline so we run on the correct VM pool


jmarolf:
**NOTE:** If this merged we will need to update which queues are required for PRs to merge

davkean:
>  If this merged we will need to update which queues are required for PRs to merge

Why?

jmarolf:
@davkean I tried several different schemes but github does not consider `unit-tests.debug` to be the same as `dotnet.project-system (Windows Debug)` even if they are named the same thing.  We can change the name of the queues to be whatever we wish once github settings are updated

jmarolf:
This is what CI should look like on PRs now:

image

Let me know if you would prefer a different name

davkean:
What is the first one?

codecov.yml
```diff
@@ -1,16 +0,0 @@
-# https://docs.codecov.io/docs/codecov-yaml
```
davkean: Why was this removed?

jmarolf:
@davkean that is the `global` run indicating whether the entire run is considered to have passed or failed. Useless to us but AzDO does not appear to have a way to hide it.

Alternatively I could have three separate YAML files checked in. That would fix it.

codecov.yml
jmarolf: Its dead code.  We haven't been able to get code coverage results since we switched to AzDO.  I can add it back.

davkean:
I suppose I'm confused about what the change does; the PR says that it changes pools but it appears to do a bunch of other things. The diff isn't helping me, I missing the jump from "changing pools" to "this means we have this extra build showing up".

codecov.yml
davkean: It's unrelated to the change, I was kinda expecting that we would bring Code Cov back, I could have sworn you had a PR that was attempting to do this.

jmarolf:
@davkean fair enough :)

When we moved from Jenkins to AzDO I tried to make the change touch as little as possible for the developer workflow.  In order to do that I created six queues (three for PR validation and three for rolling builds). I then manually configured which pool of machines each queue would use and then had each queue use the single `.vsts-dnceng.yml`.

dnceng has disabled the ability to use this approach.  Build queues now need to be configured explicitly in the YAML files themselves

This is done here: https://github.com/dotnet/project-system/blob/5e36b17931bf0eb628d86f5a87aa48ace44cb714/build/azure-pipelines.yml#L20

and here: 
https://github.com/dotnet/project-system/blob/5e36b17931bf0eb628d86f5a87aa48ace44cb714/build/azure-pipelines.yml#L72

This last one is the reason for the change.  To specify the spanish build queue you now need to use two arguments: the `name` and the `queue`
```yaml
- job: Spanish
  pool:
    name: NetCorePublic-Int-Pool
    queue: BuildPool.Windows.10.Amd64.ES.VS2017.Open
```

This is something that AzDO does not allow you to do from the UI for pipeline configurations

codecov.yml
jmarolf: I tried... No one has figured out how to do it for AzDO yet.

I'll bring this back anyways.  You are right that its unrelated

davkean:
I still don't really understand and will just accept this will what it does. This doesn't regress/remove anything we had before? I'm still waiting on VS integration and Code Coverage to come back for the AzureDevOps world.

jmarolf:
> This doesn't regress/remove anything we had before?

You will see the additional useless check that I posted in the screen shot above but everything else should be the same. Builds _may_ get faster.

> I'm still waiting on VS integration and Code Coverage to come back for the AzureDevOps world.

Yeah, it would be nice I haven't had the time to re-invent the wheel for AzDO yet.

PR #4694, opened 2019-04-04
delete correct yaml file

jmarolf:
This will fix the signed build


davidwengier:
Well I'm glad I didn't just change the Azure DevOps config on a whim :)

PR #4696, opened 2019-04-04
Revert changes from PR3602 to remove usePreviewSdk

dmonroym:
We're moving the "Use previews of .NET Core SDK" option to the new Preview Tab

This is dependent on [this](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/VS/pullrequest/173450) PR which moves it.

Related to internal bug [818730](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/818730)


src/Microsoft.VisualStudio.Editors/PublicAPI.Shipped.txt
```diff
@@ -261,14 +261,11 @@ Microsoft.VisualStudio.Editors.MyApplication.MyApplicationProperties.SingleInsta
 Microsoft.VisualStudio.Editors.MyApplication.MyApplicationProperties.SplashScreen() -> String
 Microsoft.VisualStudio.Editors.MyApplication.MyApplicationProperties.SplashScreen(value As String) -> Void
 Microsoft.VisualStudio.Editors.OptionPages.GeneralOptions
-Microsoft.VisualStudio.Editors.OptionPages.GeneralOptions.CanChangeUsePreviewSdk() -> Boolean
```
jjmew: I am not familiar with how this file gets used but this looks like we removed a public API. Is that a public API as in externals (to Microsoft) can access it? David - Do you know?

jjmew:
@jmarolf is added to the review. #Closed

src/Microsoft.VisualStudio.Editors/PublicAPI.Shipped.txt
jjmew: Maybe Jonathon know @ 

---
In reply to: [272756491](https://github.com/dotnet/project-system/pull/4696#discussion_r272756491) [](ancestors = 272756491)

Pilchie: I wouldn't worry about it - this was new in 15.9 anyway.

PR #4707, opened 2019-04-08
Allow running a task on a TaskDelayScheduler immediately

davidwengier:
Part of #4163

This adds a mechanism to TaskDelayScheduler to run a task without delay, cancelling any pending tasks that hadn't run yet.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/ITaskDelayScheduler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/ITaskDelayScheduler.cs
@@ -28,5 +28,11 @@ namespace Microsoft.VisualStudio.Threading.Tasks
         /// </remarks>
         /// <exception cref="ObjectDisposedException">The object has been disposed.</exception>
         JoinableTask ScheduleAsyncTask(Func<CancellationToken, Task> operation, CancellationToken token = default);
+
+        /// <summary>
+        /// Runs an asynchronous operation immediately, cancelling any previously scheduled tasks.
+        /// See <see cref="ScheduleAsyncTask(Func{CancellationToken, Task}, CancellationToken)"/> for more information.
+        /// </summary>
+        JoinableTask RunAsyncTask(Func<CancellationToken, Task> operation, CancellationToken token = default);
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/ITaskDelayScheduler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/ITaskDelayScheduler.cs
@@ -33,6 +33,7 @@ namespace Microsoft.VisualStudio.Threading.Tasks
         /// Runs an asynchronous operation immediately, cancelling any previously scheduled tasks.
         /// See <see cref="ScheduleAsyncTask(Func{CancellationToken, Task}, CancellationToken)"/> for more information.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
```diff
@@ -30,8 +30,19 @@ public TaskDelayScheduler(TimeSpan taskDelayTime, IProjectThreadingService threa
             _cancellationSeries = new CancellationSeries(originalSourceToken);
         }
 
+        /// <inheritdoc />
+        public JoinableTask RunAsyncTask(Func<CancellationToken, Task> operation, CancellationToken token = default)
+        {
+            return ScheduleAsyncTaskInternal(operation, true, token);
```
drewnoakes:
```suggestion
            return ScheduleAsyncTaskInternal(operation, immediate: true, token);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
```diff
@@ -30,8 +30,19 @@ public TaskDelayScheduler(TimeSpan taskDelayTime, IProjectThreadingService threa
...
+        }
+
         /// <inheritdoc />
         public JoinableTask ScheduleAsyncTask(Func<CancellationToken, Task> operation, CancellationToken token = default)
+        {
+            return ScheduleAsyncTaskInternal(operation, false, token);
```
drewnoakes:
```suggestion
            return ScheduleAsyncTaskInternal(operation, immediate: false, token);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/ITaskDelayScheduler.cs
```diff
@@ -28,5 +28,11 @@ internal interface ITaskDelayScheduler : IDisposable
         /// </remarks>
         /// <exception cref="ObjectDisposedException">The object has been disposed.</exception>
         JoinableTask ScheduleAsyncTask(Func<CancellationToken, Task> operation, CancellationToken token = default);
+
+        /// <summary>
+        /// Runs an asynchronous operation immediately, cancelling any previously scheduled tasks.
+        /// See <see cref="ScheduleAsyncTask(Func{CancellationToken, Task}, CancellationToken)"/> for more information.
+        /// </summary>
+        JoinableTask RunAsyncTask(Func<CancellationToken, Task> operation, CancellationToken token = default);
```
drewnoakes: `RunAsyncTask` can also throw `ObjectDisposedException`. Consider copying the `<exception>` documentation element from `ScheduleAsyncTask` above.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskDelayScheduler.cs
```diff
@@ -43,13 +54,16 @@ public JoinableTask ScheduleAsyncTask(Func<CancellationToken, Task> operation, C
...
+                    {
+                        await Task.Delay(_taskDelayTime, nextToken);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        return;
+                    }
                 }
 
                 if (!nextToken.IsCancellationRequested)
```
drewnoakes: Not a biggie but it might read better if this check was moved inside your new `if` block. You only need to check the token a second time if `immediate` is `false`.

PR #4710, opened 2019-04-09
ensure that the cache file is downloaded

jmarolf:
small fix that ensures we queue a download of the setting file if one does not exist on startup

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
@@ -103,7 +103,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             {
                 _solutionOpened = true;
                 // do not block package initialization on this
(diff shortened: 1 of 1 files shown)


Pilchie:
https://github.com/dotnet/project-system/blob/823302ee6d072c4b8ca41fcc3e45190d331e3a4a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs#L366

This comment seems to indicate that the current behavior was the design?

jmarolf:
@Pilchie good catch.  I should update the comment.  The implementation as it existed would lazily download the cache file and not guarantee that a message would be shown if a project was unsupported. That seems to not be the behavior that we want.  We want to ensure that the user of a project targeting an unsupported sdk is always notified.  I think its rather inconsistent to notify the user that a version is unsupported some-of-the-time.

The compromise I made with this change is that we kick off work in the back ground that will complete _eventually_ but, for example, on first solution load that eventually might be a long time.  VS kicks off a lot of work on load and we are lowest priority so it may be tens of seconds until we get run and then tens of seconds to download the file. 

With this change the behavior is: If Visual Studio is left open and connected to the internet we will always warn the user eventually.

Without this change the behavior is: If the cache file is not there, continue on, queuing its download, and assume that at some future point the stars will align such that the use is warned about this.

PR #4711, opened 2019-04-09
adding more tests for version compatibility

jmarolf:
Before making more changes to this class I wanted to add several tests to spec out the existing behavior


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.CompatibilityLevel.cs
```diff
@@ -4,7 +4,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 {
     internal sealed partial class DotNetCoreProjectCompatibilityDetector
     {
-        private enum CompatibilityLevel
+        internal enum CompatibilityLevel
```
jmarolf: Made this internal so the tests can use this type

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -150,8 +157,7 @@ public int OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)
 
                         // We need to check if this project has been newly created. Our projects will implement IProjectCreationState -we can 
                         // skip any that don't
-                        IProjectCreationState projectCreationState = project.Services.ExportProvider.GetExportedValueOrDefault<IProjectCreationState>();
```
jmarolf: This change was made because I could find no easy way to mock out the export provider,  To make this testable I added the `IProjectCreationInfoService` interface. The implementation of which is in `ProjectCreationInfoService.cs`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -41,24 +41,30 @@ internal sealed partial class DotNetCoreProjectCompatibilityDetector : IDotNetCo
...
-        private Version _ourVSVersion;
         private uint _solutionCookie = VSConstants.VSCOOKIE_NIL;
-        private bool _solutionOpened;
-        private CompatibilityLevel _compatibilityLevelWarnedForThisSolution = CompatibilityLevel.Recommended;
         private DateTime _timeCurVersionDataLastUpdatedUtc = DateTime.MinValue; // Tracks how often we meed to look for new data
-        private VersionCompatibilityData _curVersionCompatibilityData;
         private IVsSolution _vsSolution;
 
+        // These are internal for unit testing
+        internal bool SolutionOpen { get; private set; }
```
jmarolf: these properties where added so tests could introspect into what is happening in this class.  I hope to remove them someday after this class has been refactored

jmarolf:
@davidwengier thanks I'll give that a try

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
jmarolf: I've take @davidwengier suggestion and just overridden the member in a test-only class

PR #4713, opened 2019-04-09
Do not warn user of incompatible SDK version if they are on a preview

jmarolf:
fixes internal bug
- [819411](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/819411)

This change now makes it so there is an additional `preview` version that is taken into account (if the user is on a preview release of VS or has chosen to use a preview SDK).

built on top of 
- https://github.com/dotnet/project-system/pull/4712
- https://github.com/dotnet/project-system/pull/4711

review commits https://github.com/dotnet/project-system/commit/527a14969ae39df5adc08c99fc534547b3f7188d to https://github.com/dotnet/project-system/commit/04a6c06bc983c1110d550ada980d1ac95255c44b

Open question: do we want different behavior for visual studio previews vs. the use of preview sdks?


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -335,35 +335,61 @@ private static CompatibilityLevel GetCompatibilityLevelFromVersion(Version versi
...
+            if (supportedPreviewVersion is object && IsPreview() && version <= supportedPreviewVersion)
+            {
+                return CompatibilityLevel.Recommended;
+            }
+
+            return GetCompatibilityLevelFromSupportedn(version, supportedVersion, unsupportedVersion);
+        }
 
-                return CompatibilityLevel.NotSupported;
+        private static CompatibilityLevel GetCompatibilityLevelFromSupportedn(Version version, Version supportedVersion, Version unsupportedVersion)
```
Pilchie: Typo: Supportedn

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -335,35 +335,61 @@ private static CompatibilityLevel GetCompatibilityLevelFromVersion(Version versi
...
+        {
+            // a supported version exists and the version is less than the supported version
+            if (supportedVersion is object && version < supportedVersion)
+            {
+                return CompatibilityLevel.Recommended;
             }
 
-            // Only has an unsupported version
-            if (compatData.UnsupportedVersion != null)
+            // ther version is not unsupported and exactly matches the supported version
```
Pilchie: Spelling: "ther"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -335,35 +335,61 @@ private static CompatibilityLevel GetCompatibilityLevelFromVersion(Version versi
...
             {
-                if (version < compatData.SupportedVersion)
-                {
-                    return CompatibilityLevel.Recommended;
-                }
-                else if (version == compatData.SupportedVersion || (compatData.UnsupportedVersion != null && version < compatData.UnsupportedVersion))
-                {
-                    return CompatibilityLevel.Supported;
-                }
+                // no restrictions
```
Pilchie: Nit: throught - normally we capitalize the first letter of comments (applies throughout)

Pilchie:
For the open question, I think what we wanted was to be able to have a message only on Release channel of VS to say to check the checkbox, even though the VS Version and SDK Version might be identical.

jmarolf:
> 
> 
> For the open question, I think what we wanted was to be able to have a message only on Release channel of VS to say to check the checkbox, even though the VS Version and SDK Version might be identical.

@Pilchie this should be able to do that.  You provide an Preview SDK version that is supported and on preview releases of VS / VS RTM *with* the checkbox checked we do not show the user a message.  In VS RTM we will

Issue #4249, opened 2018-11-09
Light up VB WinForms/WPF property pages


davkean:
WinForms/WPF have different property pages for VB, we should light them up based on capabilities as called out in https://github.com/dotnet/project-system/issues/4248.

davkean:
It also does for C#: https://github.com/dotnet/project-system/issues/70.

Pilchie:
See also #70.

Pilchie:
Doh, same as @davkean's comment from Dec 11th.

davidwengier:
I've started looking into this and it seems like the first thing to solve is the MyApplication property, which we don't currently support and can't be represented in a XAML rule.

davkean:
Even when we have custom data sources applied to a rule?

davidwengier:
Thats what I'm looking at right now. Looks like we can either define it as an EnumProperty and use a custom provider, or define a StringProperty and specify a custom persistence provider and have that do the work. Either way is a bit hacky sounding.

davkean:
I don't know what property we're talking about, so hard to provide input.

davidwengier:
The `MyApplication` property, part of `VBProjectProperties3`. Returned [here](http://index/?query=%60GetMyAppProperties&rightProject=VisualStudio(Repo)&file=vsproject%5Cvsprojectprops%5CVsProjectProperties.cpp&line=5246) in legacy, its an instance of our `MyApplicationManager`

davkean:
How would an enum property help here? From a quick look, this would be a data source that returns an instance of IVsMyAppManager.

davidwengier:
It might not, just thinking out loud :)

melytc:
Closing as this is taken care of in the Project Properties UI.

PR #4714, opened 2019-04-10
Track MyApplication property inside app designer rather than through project system

### LINKED ISSUE #4708: Support MyApplication property for VB projects
Part of #4249

Legacy returns a value for the MyApplication property via automation, and calls into our Editors DLL to create the MyApplicationProperties object and return it.

We can't do the same thing, as our BrowseObject only supports properties that are ultimately strings, so we need an alternate system for supporting MyApplication. We can essentially do this automatically instead of using the property, we just need the MyApplicationProperties object to be:

1. Created on demand when first needed
2. Saved when the project file is saved
3. Closed when the project is closed

davidwengier:
Fixes #4708, part of #4249.

**The Situation:**
VB projects have an application framework they can opt in to which is enabled by the property pages through the [`MyApplication` property](https://docs.microsoft.com/en-us/dotnet/api/vslangproj80.vbprojectproperties3.myapplication?view=visualstudiosdk-2017) of the `BrowseObject`. This object has [8 properties](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb#L104) and stores its data in a `.myapp` file under the `My Project` folder. This file is then used to drive a [generator](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb) that generates source code that is included in the users project.

Currently the legacy implementation of MyApplication simply creates an instance of our [MyApplicationManager](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb#L136), initializes it, and returns the result of calling `GetProperties()`. It then calls `Save` when the project file saves, and `Close` when the project closes.

If the MyApplication property returns a value (ie, the system is in place) it also enables the bottom half of the VB win forms property page which contains a bunch of extra controls most of which work.

**The Problem:**
CPS currently doesn't have any way (that I can see) to have a property on the BrowseObject that is not reduced to a `string` at some point.

**The Solution:**
Since we essentially control everything around MyApplication anyway, my solution is to take full control over its lifetime without worrying about the project system in use at all. This approach has been discussed with David Kean and unless I misunderstood something, is what we discussed. Essentially we have a class that tracks the lifetime of projects, and supplies the appropriate properties object for each when requested, saves when the project is saved, and closes when the project closes. The trackers lifetime is global (its a static/Shared) but it's only initialized when first used. It's used by both legacy and CPS project systems so is independent of both.

At the moment this has been done in a way that only works in legacy (we opt out with a capability) but once the rest of the features of the property pages work we would remove this capability and this would just work by itself as a feature of the property pages. The legacy implementation of the MyApplication property could also be removed since it no longer works with this PR.

**The Issues:**
It is a fair assumption to make that we're the only consumers of these property pages (managed and legacy), and certainly from a search on GitHub and //index it seems that nobody calls the MyApplication property, this essentially breaks the contract of being a project system because we don't implement the `VBProjectProperties3` interface (not that we ever did). Even the example [in the code comments](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb#L41-L57) wouldn't work (not that it ever would for CPS).

Since we own the generator and all associated code for this an alternate approach would be to just store the properties on the `MyApplicationProperties` class in the project file and have the generator honour them. There are only 8, they're all strings (or convertible to strings), and I assume this is not an actively developed area of the product so are unlikely to change. This would mean having separate property pages for us and legacy though, so might be not ideal, though we'd use inheritance to maximise the code reuse.


src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerWindowPane.vb
```diff
@@ -436,7 +436,8 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                             hr = rdt.FindAndLockDocument(CUInt(_VSRDTFLAGS.RDT_NoLock), ProjectFullName, Hierarchy, ItemId, punkDocData, dwCookie)
                             If VSErrorHandler.Succeeded(hr) Then
                                 'We only want to save the project file itself, not any other its children (any of the other files in the project)
-                                hr = rdt.SaveDocuments(CUInt(__VSRDTSAVEOPTIONS.RDTSAVEOPT_SaveIfDirty Or __VSRDTSAVEOPTIONS.RDTSAVEOPT_SaveNoChildren),
+                                'but we do want to force it so that the MyApplication file will be saved is necessary
+                                hr = rdt.SaveDocuments(CUInt(__VSRDTSAVEOPTIONS.RDTSAVEOPT_ForceSave Or __VSRDTSAVEOPTIONS.RDTSAVEOPT_SaveNoChildren),
```
davidwengier: This actually fixes a bug in the current system, when changing something on a property page that changes the .myapp file and not the project file, we mark the properties tab as dirty but when save is pressed the `SaveIfDirty` flag means it doesn't actually save, so the .myapp file doesn't save either. If the user then closes VS they get asked to save it, but from experience I can tell you that when VS asks you to save a file that you didn't manually change, you get used to clicking "Don't Save", and in this case you would lose changes. Forcing a save here means at least the running document table events will fire.

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageVBWinForms.vb
```diff
@@ -1189,7 +1199,10 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         ''' <returns></returns>
         ''' <remarks></remarks>
         Public Overrides Function WriteUserDefinedProperty(PropertyName As String, Value As Object) As Boolean
-            If PropertyName = Const_EnableVisualStyles Then
+            If PropertyName = Const_MyApplication Then
+                Dim x = MyApplicationProperties
```
davidwengier: This is probably unnecessary, but its a lazy loading property so just in case? ¯\\\_(ツ)_/¯

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.VisualBasic.DesignTime.targets
```diff
@@ -34,7 +34,8 @@
      <ProjectCapability Include="
                           VB;
                           Managed;
-                          ClassDesigner;" />
+                          ClassDesigner;
+                          PreventAutomaticMyApplication;" />
```
davidwengier: This is what opts us _out_ of the new system, so we don't break everyones existing SDK-style VB projects by showing them property page controls that may not work 100%

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProjectLifetimeTracker.vb
```diff
@@ -0,0 +1,135 @@
...
+
+        Public Shared Function Track(projectHierarchy As IVsHierarchy) As MyApplicationProperties
+            Return s_instance.TrackInternal(projectHierarchy)
+        End Function
+
+        Private Function TrackInternal(projectHierarchy As IVsHierarchy) As MyApplicationProperties
+            Dim cookie = GetProjectFileCookie(projectHierarchy)
+
+            ' Because GetOrAdd always creates a new value, even if using a factory, we just create an object and only
+            ' initialize it later, relying on the face that MyApplicationProperties will not initialize itself twice
```
Pilchie: spelling: face => fact

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProjectLifetimeTracker.vb
```diff
@@ -0,0 +1,135 @@
...
+        End Sub
+
+        Public Shared Function Track(projectHierarchy As IVsHierarchy) As MyApplicationProperties
+            Return s_instance.TrackInternal(projectHierarchy)
+        End Function
+
+        Private Function TrackInternal(projectHierarchy As IVsHierarchy) As MyApplicationProperties
+            Dim cookie = GetProjectFileCookie(projectHierarchy)
+
+            ' Because GetOrAdd always creates a new value, even if using a factory, we just create an object and only
```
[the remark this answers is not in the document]
davidwengier: I could have sworn I looked at this, and saw that valueFactory was always called, but looking now you're right it isn't called if the key exists. I guess I could change this to use a Func then, but I'd still want to keep the `Init` call out of it just in case, as that does a bunch of interop stuff and calls into the hierarchy that its worth avoiding IMO. I could also change this to use a `Lazy`.. too many options 😛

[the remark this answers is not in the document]
davidwengier: Changed this to a lambda, and corrected the comment.

PR #4724, opened 2019-04-17
Changes to move dependencies model to the Core (non-VS) layer

lifengl:
This work is to pull the core dependencies tree logic to the core layer, so it can be loaded outside of VS environment.

I am not intended to merge this change anytime soon, but want to work with your side to eventually make this change back to the product. 
1, I shuffled the code from VS layer to the core layer to get this to work. That includes the majority part of the dependencies tree logic (except the graph node layer), some utilities.  In a few cases, I also moved code from core back to VS layer, if it depends on CPS VS layer.

2, some MEF imports now allows the component not to exist, mostly the telemetry service, as long as it retains the core function.

3, I did reference icons in the core layer now, similar to what CPS core has already done so far. Those libraries largely contains guids+ids, so it is not an issue to load it.  Of course, we can do additional work, but so far I don't feel it is worth the effort.

4, type forwarding was added to make WEB project to work as it is (in VS), we can delete them later.
The product works with those changes, and also all existing tests. The core assembly with this change works in new prototype exercises.


davkean:
Nice! I think we should aim to merge this work in for 16.2.

davkean:
@drewnoakes Can you have a look over this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/TypeForwards.cs
```diff
@@ -0,0 +1,13 @@
...
+
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.Extensibility.IProjectExportProvider))]
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.ManagedImageMonikers))]
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.DependenciesChangedEventArgs))]
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.DependencyTreeFlags))]
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.IDependenciesChanges))]
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.IDependencyModel))]
+[assembly: TypeForwardedTo(destination: typeof(Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.IProjectDependenciesSubTreeProvider))]
+
+#pragma warning restore RS0016 // Do not add multiple public overloads with optional parameters
```
drewnoakes: Is the comment here correct?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFramework.cs
```diff
@@ -3,9 +3,9 @@
 using System;
 using System.Runtime.Versioning;
 
-namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
+namespace Microsoft.VisualStudio.ProjectSystem
 {
-    internal interface ITargetFramework : IEquatable<ITargetFramework>,
+    public interface ITargetFramework : IEquatable<ITargetFramework>,
```
drewnoakes: Do we have to make this type public? Can we use IVT instead?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
```diff
@@ -3,10 +3,10 @@
 using System.Collections.Generic;
 using Microsoft.VisualStudio.Composition;
 
-namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget
+namespace Microsoft.VisualStudio.ProjectSystem
 {
-    [ProjectSystemContract(ProjectSystemContractScope2.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
-    internal interface ITargetFrameworkProvider
+    [ProjectSystemContract(ProjectSystemContractScope2.Global, ProjectSystemContractProvider.Host, Cardinality = ImportCardinality.ExactlyOne)]
+    public interface ITargetFrameworkProvider
```
drewnoakes: Ditto (public).

davkean: Agreed, we don't want this public.

lifengl: The contract between the core layer and VS layer must follow one of the two conditions below:
 1, the contract is public;
 2, the contract is optional, and not essential to use the core layer alone.

Because this is required MEF component to make the core layer to work, so it must be public, or anyone cannot use the core layer alone.  Another choice is to move the implementation inside the core layer (without depending on any VS function).  Of course, we can also copy the implementation from NuGet VS to your codebase (which is still a shim layer to wire to the NuGet framework assembly.)  I tried this, but it would require to load the NuGet assembly, which is not listed in the VS searching path at this moment, so it requires additional product change (+ more duplicated code.)

Of course, that will be your call to decide which router we want to go...

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFramework.cs
lifengl: see the comments below.  what is IVT?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/TypeForwards.cs
lifengl: I fixed it.

lifengl:
Hey, @drewnoakes : I merged changes from master, and resolved conflicts.  Can you take another look?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFramework.cs
filipnavara: IVT = InternalsVisibleTo attribute

lifengl: I don't think it makes sense, you have to white list all consumers here.

lifengl:
Removed some duplicated resource strings and resolved merge conflicts.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,78 @@
...
\ No newline at end of file
+Microsoft.VisualStudio.ProjectSystem.ITargetFramework
+Microsoft.VisualStudio.ProjectSystem.ITargetFramework.FrameworkName.get -> System.Runtime.Versioning.FrameworkName
+Microsoft.VisualStudio.ProjectSystem.ITargetFramework.FriendlyName.get -> string
+Microsoft.VisualStudio.ProjectSystem.ITargetFramework.FullName.get -> string
+Microsoft.VisualStudio.ProjectSystem.ITargetFramework.ShortName.get -> string
+Microsoft.VisualStudio.ProjectSystem.ITargetFrameworkProvider
+Microsoft.VisualStudio.ProjectSystem.ITargetFrameworkProvider.GetNearestFramework(Microsoft.VisualStudio.ProjectSystem.ITargetFramework targetFramework, System.Collections.Generic.IEnumerable<Microsoft.VisualStudio.ProjectSystem.ITargetFramework> otherFrameworks) -> Microsoft.VisualStudio.ProjectSystem.ITargetFramework
+Microsoft.VisualStudio.ProjectSystem.ITargetFrameworkProvider.GetTargetFramework(string shortOrFullName) -> Microsoft.VisualStudio.ProjectSystem.ITargetFramework
+Microsoft.VisualStudio.ProjectSystem.VS.Extensibility.IProjectExportProvider
```
drewnoakes: Lots of types moved to the core layer still have VS in their namespaces. Can we update these and maintain back compat with type forwarding?

drewnoakes: Never mind. I see type forwarding only applies to matching FQNs.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
drewnoakes: I see. I don't see a way around this @davkean. It's only adding two new types to the API surface: `ITargetFrameworkProvider` and `ITargetFramework`.

davkean: Let me think about this; this is code I had planned on throwing away.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PublicAPI.Unshipped.txt
davkean: Did we had type forwarding for all the types we moved?

drewnoakes: See https://github.com/dotnet/project-system/pull/4724/files#diff-2cde465cdf265cc84c5eded3515f7f2f. That's not all moved types. Only the public ones afaict.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
lifengl: I think we can keep them as internal for now, if you need throw it away in the future. The separation of two layers are more for long term future, instead of being required immediately.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PublicAPI.Unshipped.txt
lifengl: i guess we only need forward public API types.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
lifengl: Let me know what is your decision. I can revert it to keep them internal, and we can resolve the problem for now, and move on for other changes.

davkean: Yep, let's do that and file a bug against me to fix this "internal" relationship.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PublicAPI.Unshipped.txt
drewnoakes: All types forwarded in this PR are public.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
drewnoakes: Reverted accessibility in fa14ef8875b714e9fce8a482dc79f3ecaed7e1a8 and opened #4886 to track removing those interfaces.

drewnoakes:
I will merge this once CI checks pass.

PR #4726, opened 2019-04-17
ensure we always call GetInstanceForCurrentProcess on the UI thread

jmarolf:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -175,18 +175,29 @@ public int OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)
         // This method is overridden in test code
         protected virtual async Task<bool> IsPreviewSDKInUseAsync()
         {
-            var vsSetupConfig = new SetupConfiguration();
-            ISetupInstance setupInstance = vsSetupConfig.GetInstanceForCurrentProcess();
-            if (setupInstance is ISetupInstanceCatalog setupInstanceCatalog &&
-                setupInstanceCatalog.IsPrerelease())
+            ISetupInstanceCatalog setupInstanceCatalog = await TryGetSetupInstanceAsync();
+            if (setupInstanceCatalog?.IsPrerelease() == true)
```
etbyrd: Is the equality check needed?

jmarolf: yes `Nullable<bool>` is not convertible to a boolean expression in C# (it is in VB). By comparing it to a bool we invoke the implicit conversion operator from `Nullable<bool>` to `bool`

PR #4729, opened 2019-04-17
remove reference to Microsoft.VisualStudio.Utilities

jmarolf:
This is unused


dmonroym:
Microsoft.VisualStudio.Shell.Framework has a dependency on it; we need to update the version to 16.0.28729

PR #4731, opened 2019-04-18
Update package versions

dmonroym:
Updating package versions to latest - Primarily to update Microsoft.VisualStudio.Utilities so we can switch from DpiHelper to DpiAwareness

- **Disabled:** VSTHRD003 for **all projects**
- **Re-added** Microsoft.VisualStudio.Utilities with updated version
- **Updated** all the dependencies of Microsoft.VisualStudio.Utilities

### DIFF
--- a/build/Default.ruleset
+++ b/build/Default.ruleset
@@ -173,6 +173,7 @@
     <Rule Id="VSTHRD010" Action="None" />                       <!-- Visual Studio service should be used on main thread explicitly.-->
     <Rule Id="VSTHRD103" Action="None" />                       <!-- Call async methods when in an async method.-->
     <Rule Id="VSTHRD108" Action="None" />                       <!-- Thread affinity checks should be unconditional.-->
+    <Rule Id="VSTHRD003" Action="None" />                       <!-- Avoid awaiting or returning a Task representing work that was not started within your context as that can lead to deadlocks-->
   </Rules>
   <Rules AnalyzerId="Microsoft.VisualStudio.SDK.Analyzers" RuleNamespace="Microsoft.VisualStudio.SDK.Analyzers">
     <Rule Id="VSSDK006" Action="None" />                        <!-- Check whether the result of GetService calls is null -->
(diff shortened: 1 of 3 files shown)


build/Default.ruleset
```diff
@@ -173,6 +173,7 @@
     <Rule Id="VSTHRD010" Action="None" />                       <!-- Visual Studio service should be used on main thread explicitly.-->
     <Rule Id="VSTHRD103" Action="None" />                       <!-- Call async methods when in an async method.-->
     <Rule Id="VSTHRD108" Action="None" />                       <!-- Thread affinity checks should be unconditional.-->
+    <Rule Id="VSTHRD003" Action="None" />                       <!-- Avoid awaiting or returning a Task representing work that was not started within your context as that can lead to deadlocks-->
```
davidwengier: Is this a new diagnostic that we have code in violation of?

dmonroym: Yes, when updating the version of Microsoft.VisualStudio.Threading I got these errors (just to list a couple):
ProjectSystem\VS\Build\LanguageServiceErrorListProvider.cs(72,28): error VSTHRD003: Avoid awaiting or returning a Task representing work that was
not started within your context as that can lead to deadlocks. [G:\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\Microsoft.Vi
sualStudio.ProjectSystem.Managed.VS.csproj]
ProjectSystem\VS\Build\LanguageServiceErrorListProvider.cs(79,24): error VSTHRD003: Avoid awaiting or returning a Task representing work that was
not started within your context as that can lead to deadlocks. [G:\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\Microsoft.Vi
sualStudio.ProjectSystem.Managed.VS.csproj]

After talking with @jmarolf he suggested this solution but if there's a better way of fixing these or if this shouldn't be suppressed let me know and I'll look further into it.

davidwengier: I don't know about this specific rule, but at the very least it might be worth logging an issue to come back and evaluate the violations, and either fix the code if its a problem or, and this looks more likely in the cases you've mentioned here, use GlobalSuppressions to suppress the specific examples (and possibly log an issue for the analyzer about false positives) so that we can at least have the analyzer active and prevent new code from having problems.

dmonroym: Created Issue #4732

jmarolf: [VSTHRD003](https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD003.md) is about wanting everything to be a JTF task.  I agree with the sentiment of the analysis that mixing the two can lead to complex-to-diagnose bugs. In reality it warns every-time you do `Task.FromResult` which is never a problem and something we do a lot of places. Until this analyzer is updated to correctly do flow analysis and determine that you are actually mixing contexts I think it is just noise.

davkean: Agreed, looked at the results, every single usage is noise.

PR #4734, opened 2019-04-18
Extend up to date check log message

drewnoakes:
An internal customer reported seeing this message when building twice in a row. This extension to the log message would help identify why this seemingly incorrect behaviour occurs.


drewnoakes:
Just realised there's a test for this message. Fix incoming.

davidwengier:
> Just realised there's a test for this message

The build seemed to work ¯\_(ツ)_/¯

drewnoakes:
> The build seemed to work ¯\\_(ツ)_/¯

It failed before I force pushed a fix.

PR #4738, opened 2019-04-20
fix COM object exception

jmarolf:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -185,17 +185,29 @@ protected virtual async Task<bool> IsPreviewSDKInUseAsync()
...
+        private async Task<bool> TryQueryCOMObject(ISetupInstanceCatalog setupInstanceCatalog)
+        {
+            try
+            {
+                await _threadHandling.Value.SwitchToUIThread();
+                return setupInstanceCatalog.IsPrerelease();
+            }
+            catch (Exception)
+            {
+                // COM object may not be queryable yet.
```
davidwengier: What isn't the COM object queryable? If its early on in the VS process, don't this mean this check won't work for the first project someone opens, ever, but will work for subsequent loads?

Sorry if those are silly questions, but I don't fully understand why this COM error is occurring.

jmarolf: Yeah, I will admit to not fully understanding some of what this cause the COm query to fail in the startup. If this happens it is possible for the project file to not notify immediately

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -185,17 +185,29 @@ protected virtual async Task<bool> IsPreviewSDKInUseAsync()
...
         }
 
-        private async Task<ISetupInstanceCatalog> TryGetSetupInstanceAsync()
+        private Task<bool> IsPrereleaseAsync()
         {
-            await _threadHandling.Value.SwitchToUIThread();
             var vsSetupConfig = new SetupConfiguration();
-            ISetupInstance setupInstance = vsSetupConfig.GetInstanceForCurrentProcess();
-            if (setupInstance is ISetupInstanceCatalog setupInstanceCatalog)
+            if (vsSetupConfig.GetInstanceForCurrentProcess() is ISetupInstanceCatalog setupInstanceCatalog)
```
Pilchie: It looks to me like the exception is was actually happening in the pattern match for the cast to `ISetupInstanceCatalog`, which would imply that the exception would still escape.  Also, there is no notion of things not being queryable yet that I know of in VS.  This doesn't seem like the right fix here. At best it papers over the problem such that things will default to seeming to be the release channel even if they aren't.

This needs to be revisited to understand why the cast is actually failing.

jmarolf: > It looks to me like the exception is was actually happening in the pattern match for the cast to ISetupInstanceCatalog, which would imply that the exception would still escape.

That is not what happened based on debugging.  The pattern match succeeds and then the exception is thrown on the call to `IsPrerelease`.

> At best it papers over the problem such that things will default to seeming to be the release channel even if they aren't.

I agree this is not the right fix. This sort of this (to my naive way of thinking) should always be safe to do. I'll need to ensure that this exception doesn't happen.

davkean: What was the follow up on this? What was the exception on IsPrerelease?

davkean: I see an internal thread that states the following:

```
System.AggregateException: One or more errors occurred. ---> System.InvalidCastException: Unable to cast COM object of type 'System.__ComObject' to interface type 'Microsoft.VisualStudio.Setup.Configuration.ISetupInstanceCatalog'. This operation failed because the QueryInterface call on the COM component for the interface with IID '{9AD8E40F-39A2-40F1-BF64-0A6C50DD9EEB}' failed due to the following error: No such interface supported (Exception from HRESULT: 0x80004002 (E_NOINTERFACE)).
   at System.StubHelpers.StubHelpers.GetCOMIPFromRCW(Object objSrc, IntPtr pCPCMD, IntPtr& ppTarget, Boolean& pfNeedsRelease)
   at Microsoft.VisualStudio.Setup.Configuration.ISetupInstanceCatalog.IsPrerelease()
   at Microsoft.VisualStudio.ProjectSystem.VS.DotNetCoreProjectCompatibilityDetector.<IsPreviewSDKInUseAsync>d__43.MoveNext()
```

This is occurring because the instance returned from `GetInstanceForCurrentProcess` is failing the QI for `ISetupInstanceCatalog`. That's the issue and we should investigate that. The runtime is only doing the actual QI on the call to IsPrerelease due to changes around WinRT timeframe.

jmarolf: I'm going to investigate this, this morning and see if I can put out the correct fix.

Pilchie: @davkean a point of curiosity - how does the runtime do the `is` check without doing a QI?

davkean: I think it lies, it was only supposed to happen for WinRT interfaces but I think it bled out into traditional COM, it's been 7 years so I can't recall the exact details - see: https://github.com/dotnet/coreclr/blob/fdc9998609263d9f531079e2cbe0f3f31eba3382/src/vm/stubhelpers.cpp#L571.

PR #4740, opened 2019-04-23
Create TempPEBuildManager to compile TempPEs from dataflow updates.

davidwengier:
Part of #4163 

This is the TempPEBuildManager that subscribes to dataflow updates and compiles TempPE DLLs. Most of this code has been seen before.

Sorry, this is a pretty big file but I couldn't really break this up any more. Coming in future PRs are tests, and file change watching. Some structural quirks here could be to support those.

Note that merging this PR doesn't turn the functionality on yet.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Get the list of design time monikers that need to have TempPE libraries created
+        /// </summary>
+        string[] GetTempPEMonikers();
+
+        /// <summary>
+        /// Get the XML used by the type resolution service that describes the TempPE library for the specified input moniker, and triggers the compilation of the library if necessary
+        /// </summary>
+        /// <param name="moniker">On of the monikers return from <see cref="GetTempPEMonikers"/> for which a TempPE library should be produced.</param>
```
jmarolf: >return [](start = 53, length = 6)

returned?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+            await UnconfiguredProjectServices.ThreadingService.SwitchToUIThread();
+
+            if (shouldCompile)
+            {
+                DesignTimeInputsItem inputs = AppliedValue.Value;
+                if (inputs.TaskSchedulers.TryGetValue(moniker, out ITaskDelayScheduler scheduler))
+                {
+                    HashSet<string> files = GetFilesToCompile(moniker, inputs.SharedInputs);
+                    string outputFileName = GetOutputFileName(inputs.OutputPath, moniker);
+                    _projectFaultHandlerService.Forget(scheduler.ScheduleAsyncTask(token => CompileTempPEAsync(files, outputFileName, token)).Task, UnconfiguredProject);
```
jmarolf: Seems like there is a lot of scheduling logic both here and in `GetTempPEDescriptionXmlAsync`, perhaps this could be consolidated into a single helper method?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -77,6 +77,8 @@
   <StringProperty Name="IntDir"
                   Visible="false" />
 
+  <StringProperty Name="IntermediateOutputPath" />
```
davkean: Visible=false

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+            { }
+            catch (UnauthorizedAccessException)
+            { }
+
+            return false;
+        }
+
+        private static string GetOutputFileName(string outputPath, string moniker)
+        {
+            // All monikers are project relative paths by defintion (anything else is a link), meaning the only invalid
```
davkean: I don't under the why about "anything else is a link" it seems like there's an assumption here not called out, are linked files excluded?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+                catch (IOException)
+                { }
+                catch (UnauthorizedAccessException)
+                { }
+            }
+        }
+
+        protected virtual HashSet<string> GetFilesToCompile(string moniker, ImmutableHashSet<string> sharedInputs)
+        {
+            // This is a HashSet because we allow files to be both inputs and shared inputs, and we don't want to compile the same file twice
```
davkean: I think I would add comment here about why you are adding shared inputs and the input together, ie some background.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+                                  ITempPECompiler compiler,
+                                  IFileSystem fileSystem,
+                                  ITelemetryService telemetryService,
+                                  IProjectFaultHandlerService projectFaultHandlerService)
+             : base(threadingService.JoinableTaskContext)
+        {
+            UnconfiguredProjectServices = unconfiguredProjectServices;
+            _activeWorkspaceProjectContextHost = activeWorkspaceProjectContextHost;
+            _projectSubscriptionService = projectSubscriptionService;
+            _buildManager = new Lazy<VSBuildManager>(() => (VSBuildManager)projectEvents.BuildManagerEvents);
```
davkean: This looks like a implicit dependency not explicitly called out in an API, someone could change the return value of ProjectEvents.BuildManagerEvents and have no idea you have this dependency.  Introduce a new interface for this or implement VSBuildManager directly.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    [Export(typeof(ITempPEBuildManager))]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasicLanguageService)]
+    internal partial class TempPEBuildManager : UnconfiguredProjectHostBridge<IProjectVersionedValue<InputTuple>, TempPEBuildManager.DesignTimeInputsDelta, IProjectVersionedValue<TempPEBuildManager.DesignTimeInputsItem>>, ITempPEBuildManager
+    {
+        private static readonly TimeSpan s_compilationWaitTime = TimeSpan.FromMilliseconds(500);
+
+        protected readonly IUnconfiguredProjectCommonServices UnconfiguredProjectServices;
```
davkean: Why protected?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+using Microsoft.VisualStudio.Threading.Tasks;
+
+using Task = System.Threading.Tasks.Task;
+
+using InputTuple = System.Tuple<Microsoft.VisualStudio.ProjectSystem.IProjectSubscriptionUpdate, Microsoft.VisualStudio.ProjectSystem.IProjectSubscriptionUpdate>;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    [Export(typeof(ITempPEBuildManager))]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasicLanguageService)]
```
davkean: Is the expectation that the importer will check the capabilities of the project before using this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// This method is where we tell data flow which blocks we're interested in receiving updates for
+        /// </summary>
+        protected override IDisposable LinkExternalInput(ITargetBlock<IProjectVersionedValue<InputTuple>> targetBlock)
+        {
+            return ProjectDataSources.SyncLinkTo(
```
davkean: I think you'll need to JoinUpstreamDataSources on SourceItemsRuleSource and ProjectRuleSource, so that if someone blocks on the UI thread for your result and these need to finish on the UI thread to complete, that be allowed to proceed because there's a relationship between yourself and them.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+                {
+                    addedDesignTimeSharedInputs.Add(item);
+                }
+                else if (wasDesignTimeShared && !designTimeShared)
+                {
+                    removedDesignTimeSharedInputs.Add(item);
+                }
+            }
+
+            bool namespaceChanged = configChanges.Difference.ChangedProperties.Contains(ConfigurationGeneral.RootNamespaceProperty);
```
davkean: Does this work for VB? They have a slightly different representation here that I cannot remember.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+                item.TryGetValue(Compile.DesignTimeProperty, out string designTimeString);
+                item.TryGetValue(Compile.DesignTimeSharedInputProperty, out string designTimeSharedString);
+
+                if (linkString != null && linkString.Length > 0)
+                {
+                    // Linked files are never used as TempPE inputs
+                    return (false, false);
+                }
+
+                return (StringComparers.PropertyValues.Equals(designTimeString, bool.TrueString), StringComparers.PropertyValues.Equals(designTimeSharedString, bool.TrueString));
```
davkean: PropertyLiteralValues

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+            return new TaskDelayScheduler(s_compilationWaitTime, UnconfiguredProjectServices.ThreadingService, DisposalToken);
+        }
+
+        protected override async Task DisposeCoreAsync(bool initialized)
+        {
+            if (initialized)
+            {
+                DisposeTaskSchedulers();
+            }
+            await base.DisposeCoreAsync(initialized);
```
davkean: return.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+        }
+
+        /// <summary>
+        /// InitializeInnerCoreAsync is responsible for setting an initial AppliedValue. This value will be used by any UI thread calls that may happen
+        /// before the first data flow blocks have been processed. If this method doesn't return a value then the system will block until the first blocks
+        /// have been applied but since we are initialized on the UI thread, that blocks for us, so we must return an initial state.
+        /// </summary>
+        protected override Task InitializeInnerCoreAsync(CancellationToken cancellationToken)
+        {
+            AppliedValue = new ProjectVersionedValue<DesignTimeInputsItem>(new DesignTimeInputsItem(), ImmutableDictionary.Create<NamedIdentity, IComparable>());
```
davkean: Is this right? If I understand, by default it has zero monikers until you've processed your first changes? What if someone queries straight after load?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,550 @@
...
+        private static string GetOutputFileName(string outputPath, string moniker)
+        {
+            // All monikers are project relative paths by defintion (anything else is a link), meaning the only invalid
+            // filename characters possible are path separators so we just replace them
+            return Path.Combine(outputPath, moniker.Replace('\\', '.') + ".dll");
+        }
+
+        protected virtual async Task CompileTempPEAsync(HashSet<string> filesToCompile, string outputFileName, CancellationToken token)
+        {
+            _telemetryService.PostProperty(TelemetryEventName.TempPECompilation, TelemetryPropertyName.TempPECompilationOutputFileName, outputFileName);
```
davkean: This has PII in the file name.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Yes, that's correct. This is a rock-and-a-hard-place situation, where we can't initialize ProjectHostBridge on the UI thread (there was a thread on this with Lifeng in December) but we didn't want to initialize the TempPE manager unless it was needed by a designer.

Potentially we could change this now that we will have WinForms capabilities across project systems, which reduces the risk of unnecessary creation.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Oops, good point. I'll remove all of the telemetry, it was mainly to make my debugging easier.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: 🤦‍♂️ 

This probably explains the analyzer complaining, I thought it was a false positive.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/ITempPEBuildManager.cs
davidwengier: Thanks. Looks like "On" should also be "One" :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Yes, linked files aren't TempPE inputs. It's called out in a comment below, but I'll make it clearer.
https://github.com/dotnet/project-system/pull/4740/files#diff-8cf178a6054d65774e94e9f40a9a353eR492

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: No analyzer will still complain.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: How does someone right code that correctly queries this code at the right time?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Yes, VSBuildManager will import this, and will cope with it being null

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: This was probably used for tests at some point. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davkean: It won't be null. Capabilities doesn't affect what MEF sees, they filters exports via OrderPrecedenceImportCollection (if you look inside that collection you see that all exports are present underneath regardless of capabilities). Singular imports need to be manually filtered by importing Lazy<T, IAppliesToMetadataView> filtering against the current configured project or unconfigured project.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: Discussed offline, hopefully will be able to move to a model where the initial value is correct without deadlocking.

davidwengier:
I've addressed most of the PR feedback and resolved the conversations where applicable. The code around initalization is still WIP.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
davidwengier: VB handles RootNamespace differently in the compiler, which is what https://github.com/dotnet/roslyn/pull/33653 fixed, but I didn't run into any problems here in the testing that led to that PR.

davidwengier:
All PR feedback has been addressed, and for now this service will load with the project, but only for Windows Forms projects. In future we'll address the perf issues by fixing the bug in ProjectHostBridge (#4752).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -0,0 +1,568 @@
...
+            return AppliedValue.Value.Inputs.ToArray();
+        }
+
+        /// <summary>
+        /// Called privately to actually fire the TempPE events and optionally recompile the TempPE library for the specified input
+        /// </summary>
+        private async Task FireTempPEDirtyAsync(string moniker, bool shouldCompile)
+        {
+            // Not using use the ThreadingService property because unit tests
+            await _unconfiguredProjectServices.ThreadingService.SwitchToUIThread();
```
jmarolf: You mention doing this for unit tests, but I don't see any tests added in this PR. Is that change coming later?

davidwengier: Yes, just trying to reduce the amount of code to review.

PR #4748, opened 2019-04-24
correct fix for com exception

jmarolf:
turns out `is`/`as` patterns don't work on COM objects ¯\_(ツ)_/¯


davkean:
I'm investigating this offline, I want to know the exact cause of this and why this is being treated differently.

PR #4749, opened 2019-04-24
Add WPF and WinForms item menu handlers

### LINKED ISSUE #137: Project System: Add Item menu support for Winforms and WPF
_From @davkean on January 14, 2016 21:34_

This is tracking Add Item support including WinForms and WPF menu items:

image
image

_Copied from original issue: dotnet/roslyn#7965_

davidwengier:
Fixes #137 

This adds "Add Item" menu support for WPF and WinForms, pulling strings from their current packages so that everything works in foreign languages.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Input/VisualStudioStandard2KCommandId.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.VisualStudio.Input
+{
+    internal static class VisualStudioStandard2KCommandId
+    {
+        public const long AddForm = (long)VSConstants.VSStd2KCmdID.ADDWFCFORM;
```
davkean: Can we also add the VB ones?   

ADDMODULE = 402,
ADDMODULECTX = 403,

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WPFAddItemCommandHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        // from VS: src\vsproject\vb\vbprj\vbprjstr.h
+        protected enum LegacyVBStrings : uint
+        {
+            IDS_VSDIR_CLIENTPROJECTITEMS = 3081,
+        }
+
+        // from VS: src\vsproject\fidalgo\WPF\Flavor\WPFFlavor\Guids.cs
+        private enum CommandIds : long
```
davkean: Can we move this alongside VisualStudioStandard2KCommandId.cs?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WPFAddItemCommandHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    [ExportCommandGroup(CommandGroup.WPF)]
+    [AppliesTo(ProjectCapability.WPF)]
+    internal partial class WPFAddItemCommandHandler : AbstractAddItemCommandHandler
+    {
+        private static readonly Guid s_wpfPackage = new Guid("{b3bae735-386c-4030-8329-ef48eeda4036}");
+
+        // from VS: src\vsproject\cool\coolpkg\resource.h
+        protected enum LegacyCSharpStrings : uint
```
davkean: Why protected?

Can we move these to a separate file? When I open the file you want to see meat of the class not the boilerplate enums/types that the thing uses.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WPFAddItemCommandHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+            { (long)CommandIds.WPFPage, new CommandDetails(LegacyCSharpPackageGuid, (uint)LegacyCSharpStrings.IDS_TEMPLATE_DIRLOCALITEMS, s_wpfPackage, (uint)TemplateNames.WPFPage) },
+            { (long)CommandIds.AddWPFUserControl, new CommandDetails(LegacyCSharpPackageGuid, (uint)LegacyCSharpStrings.IDS_TEMPLATE_DIRLOCALITEMS, s_wpfPackage, (uint)TemplateNames.WPFUserControl) },
+            { (long)CommandIds.WPFUserControl, new CommandDetails(LegacyCSharpPackageGuid, (uint)LegacyCSharpStrings.IDS_TEMPLATE_DIRLOCALITEMS, s_wpfPackage, (uint)TemplateNames.WPFUserControl) },
+            { (long)CommandIds.AddWPFResourceDictionary, new CommandDetails(LegacyCSharpPackageGuid, (uint)LegacyCSharpStrings.IDS_TEMPLATE_DIRLOCALITEMS, s_wpfPackage, (uint)TemplateNames.WPFResourceDictionary) },
+            { (long)CommandIds.WPFResourceDictionary, new CommandDetails(LegacyCSharpPackageGuid, (uint)LegacyCSharpStrings.IDS_TEMPLATE_DIRLOCALITEMS, s_wpfPackage, (uint)TemplateNames.WPFResourceDictionary) },
```
davkean: Can we make this into a table with columns to make it easier to read?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// </summary>
+    internal abstract partial class AbstractAddItemCommandHandler : IAsyncCommandGroupHandler
+    {
+        protected static readonly Guid LegacyCSharpPackageGuid = new Guid("{FAE04EC1-301F-11d3-BF4B-00C04F79EFBC}");
+        protected static readonly Guid LegacyVBPackageGuid = new Guid("{164B10B9-B200-11d0-8C61-00A0C91E29D5}");
+
+        private readonly IPhysicalProjectTree _projectTree;
+        private readonly IUnconfiguredProjectVsServices _projectVsServices;
+        private readonly IVsUIService<IVsAddProjectItemDlg> _addItemDialog;
+        private readonly IVsService<SVsShell, IVsShell> _vsShell;
```
davkean: IVsService<IVsShell>, only need to specify SVShell at the import.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddClassProjectCommand.cs
```diff
@@ -39,27 +39,9 @@ protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree
             }
         }
```
davkean: Why do we still need this class? I expected it be merged with the "uber" class and follow the same pattern.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,131 @@
...
+                    return await ShowAddProjectItemDialog(node, dirName, templateName, _projectTree, _projectVsServices, _addItemDialog);
+                }
+            }
+
+            return false;
+        }
+
+        private bool TryGetCommandDetails(long commandId, out CommandDetails result)
+        {
+            IProjectCapabilitiesScope capabilities = _projectVsServices.ActiveConfiguredProject.Capabilities;
```
davkean: This service lives at the "configuredproject" scope (look at the [ProjectSystemContract] applied to it), it should be importing ConfiguredProject and use that to filter, not the active configuration. Only unconfigured-project scoped things should use the active config.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,131 @@
...
+            Requires.NotNull(nodes, nameof(nodes));
+
+            if (nodes.Count == 1)
+            {
+                if (TryGetCommandDetails(commandId, out CommandDetails result))
+                {
+                    IProjectTree node = nodes.First();
+
+                    // Get the strings from the legacy package
+                    IVsShell vsShell = await _vsShell.GetValueAsync();
```
davkean: While commands are likely to be called on the UI thread, you should also explicitly switch to the UI thread anyway before touching the shell.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,131 @@
...
+                                                                (uint)uiFlags,
+                                                                dirName,
+                                                                templateName,
+                                                                ref strBrowseLocations,
+                                                                ref strFilter,
+                                                                out _);
+
+            // Return true here regardless of whether or not the user clicked OK or they clicked Cancel. This ensures that some other
+            // handler isn't called after we run.
+            return res == VSConstants.S_OK || res == VSConstants.OLE_E_PROMPTSAVECANCELLED;
```
davkean: I just changed this code to use HResult, please do the same.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WinFormsAddItemCommandHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    [ExportCommandGroup(CommandGroup.VisualStudioStandard2k)]
+    [AppliesTo(ProjectCapability.WindowsForms)]
+    internal class WinFormsAddItemCommandHandler : AbstractAddItemCommandHandler
+    {
+        // from VS: src\vsproject\cool\coolpkg\resource.h
+        protected enum LegacyCSharpStrings : uint
```
davkean: You have two copies of this, can we just merge into a single enum instead of two enums with the same name?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WinFormsAddItemCommandHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            IDS_TEMPLATE_NEWWFCWIN32FORM = 2237,
+            IDS_TEMPLATE_NEWWFCCOMPONENT = 2246,
+            IDS_TEMPLATE_NEWUSERCONTROL = 2295,
+            IDS_PROJECTITEMTYPE_STR = 2346,
+        }
+
+        // from VS: src\vsproject\vb\vbprj\vbprjstr.h
+        protected enum LegacyVBStrings : uint
```
davkean: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.CommandDetails.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    internal partial class AbstractAddItemCommandHandler
+    {
+        protected class CommandDetails
```
davkean: This is really "TemplateDetails" as that is what it is describing.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
davidwengier: I certainly didn't check this, or think to, so my bad, but I think I might have fallen arse backwards into being correct anyway?

https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IUnconfiguredProjectVsServices.cs#L11

davidwengier:
All PR feedback has been addressed, bar one that I think might be okay?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
davidwengier: @davkean Can you confirm if this needs to be changed or not?

davkean: Yes it still is a problem, not sure what you are pointing me to?

davkean: Ah, I think I understand the confusion - I think you want to read up on scopes, and this in particular: https://github.com/Microsoft/VSProjectSystem/blob/master/doc/overview/scopes.md#scopes-at-odds.

davidwengier: My bad, thought by "this service" you meant `_projectVsServices`. Will update as appropriate.

davidwengier: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WPFAddItemCommandHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    [ExportCommandGroup(CommandGroup.WPF)]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal partial class WPFAddItemCommandHandler : AbstractAddItemCommandHandler
+    {
+        private static readonly Guid s_wpfPackage = new Guid("{b3bae735-386c-4030-8329-ef48eeda4036}");
+
+        protected override Dictionary<long, List<TemplateDetails>> GetTemplateDetails() => new Dictionary<long, List<TemplateDetails>>
```
davkean: This returns a new value every time, please store this in a static field.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IUnconfiguredProjectVsServicesFactory.cs
```diff
@@ -35,6 +35,12 @@ public static IUnconfiguredProjectVsServices Implement(Func<IVsHierarchy> hierar
                     .Returns(projectCreator());
             }
 
+            if (configuredProjectCreator != null)
```
davidwengier: Didn't end up needing this in the end, but I don't think it causes any harm being here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS2kAddItemCommandHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.VisualStudio.Input;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    [ExportCommandGroup(CommandGroup.VisualStudioStandard2k)]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class VS2kAddItemCommandHandler : AbstractAddItemCommandHandler
+    {
+        protected override Dictionary<long, List<TemplateDetails>> GetTemplateDetails() => new Dictionary<long, List<TemplateDetails>>
```
davkean: Returns a new value every time.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS97AddItemCommandHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.VisualStudio.Input;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    [ExportCommandGroup(CommandGroup.VisualStudioStandard97)]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class VS97AddItemCommandHandler : AbstractAddItemCommandHandler
+    {
+        protected override Dictionary<long, List<TemplateDetails>> GetTemplateDetails() => new Dictionary<long, List<TemplateDetails>>
```
davkean: Ditto, new value every time.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+                    // Return true here regardless of whether or not the user clicked OK or they clicked Cancel. This ensures that some other
+                    // handler isn't called after we run.
+                    return res == HResult.OK || res == HResult.Ole.PromptSaveCancelled;
+                }
+            }
+
+            return false;
+        }
+
+        private bool TryGetCommandDetails(long commandId, out TemplateDetails result)
```
davkean: TryGetTemplateDetails

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.TemplateDetails.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
+{
+    internal partial class AbstractAddItemCommandHandler
+    {
+        internal class TemplateDetails
+        {
+            public string CapabilityCheck;
```
davkean: This terminology throughout VS is "AppliesTo"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.TemplateDetails.cs
```diff
@@ -0,0 +1,28 @@
...
+            public Guid DirNamePackageGuid;
```
davkean: All these fields can be readonly.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WPFAddItemCommandHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+           // Command Id                                Capability                      ExtraCapability         DirNamePackageGuid          DirNameResourceId                                           TemplateName    TemplateNameResourceId
+           //                                                                                                                                                                   PackageGuid
+            { WPFCommandId.AddWPFWindow,                ProjectCapability.CSharp,       ProjectCapability.WPF,  LegacyCSharpPackageGuid,    LegacyCSharpStringResourceIds.IDS_TEMPLATE_DIRLOCALITEMS,   s_wpfPackage,   WPFTemplateNames.WPFWindow              },
```
davkean: I don't understand what I'm looking at, where is this ExtraCapability stored?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+    internal abstract partial class AbstractAddItemCommandHandler : IAsyncCommandGroupHandler
+    {
+        protected static readonly Guid LegacyCSharpPackageGuid = new Guid("{FAE04EC1-301F-11d3-BF4B-00C04F79EFBC}");
+        protected static readonly Guid LegacyVBPackageGuid = new Guid("{164B10B9-B200-11d0-8C61-00A0C91E29D5}");
+        private readonly ConfiguredProject _configuredProject;
+        private readonly IPhysicalProjectTree _projectTree;
+        private readonly IUnconfiguredProjectVsServices _projectVsServices;
+        private readonly IVsUIService<IVsAddProjectItemDlg> _addItemDialog;
+        private readonly IVsUIService<IVsShell> _vsShell;
+        private readonly Lazy<Dictionary<long, List<TemplateDetails>>> _commandMap;
```
davkean: These never change, I would type them as immutable types to prevent them from ever being changed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS2kAddItemCommandHandler.cs
davkean: Understand that this is cached, but we basically have a representation of this object per configuration per project, when it never changes. Store a stack field, throw away the lazy and just return the static field.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS2kAddItemCommandHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+           // Command Id                                        Capability                      AdditionalCapability                DirNamePackageGuid          DirNameResourceId                                       TemplateNameResourceId
+            { VisualStudioStandard2KCommandId.AddForm,          ProjectCapability.CSharp,       ProjectCapability.WindowsForms,     LegacyCSharpPackageGuid,    LegacyCSharpStringResourceIds.IDS_PROJECTITEMTYPE_STR,  LegacyCSharpStringResourceIds.IDS_TEMPLATE_NEWWFCWIN32FORM  },
```
davkean: TBH, I find the extension methods make it very hard to see what's going on here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.TemplateDetails.cs
davidwengier: Hah, this is my sense of neatness taking over. The two Guids are passed by ref so can't be properties or readonly fields, so I just made them all the same. I'll make the other three properties :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/WPFAddItemCommandHandler.cs
davidwengier: As discussed, its just a convenience of the helper method, to make the table look nicer. The two get combined (with &) in the helper here: https://github.com/dotnet/project-system/pull/4749/files#diff-aa7528047255ba61bc869434cf8fd4a7R18

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.TemplateDetails.cs
davkean: From a correctness point of view these values should never change, I think I would copy them to a field before passing them to a method.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS2kAddItemCommandHandler.cs
davkean: In that I couldn't connect how this was related to the TemplateDetails.

davidwengier: I may have misunderstood your request to make this into a table.

You prefer:
```
{ VisualStudioStandard2KCommandId.AddForm, new TemplateDetails(ProjectCapability.CSharp,       ProjectCapability.WindowsForms,     LegacyCSharpPackageGuid,    LegacyCSharpStringResourceIds.IDS_PROJECTITEMTYPE_STR,  LegacyCSharpStringResourceIds.IDS_TEMPLATE_NEWWFCWIN32FORM) }
```
?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+            return GetCommandStatusResult.Unhandled;
+        }
+
+        public async Task<bool> TryHandleCommandAsync(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut)
+        {
+            Requires.NotNull(nodes, nameof(nodes));
+
+            if (nodes.Count == 1)
+            {
+                if (TryGetCommandDetails(commandId, out TemplateDetails result))
```
davkean: Can you merge the two if blocks or move to early exit to avoid unneeded nesting.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+                        return false;
+                    }
+
+                    __VSADDITEMFLAGS uiFlags = __VSADDITEMFLAGS.VSADDITEM_AddNewItems |
+                                               __VSADDITEMFLAGS.VSADDITEM_SuggestTemplateName |
+                                               __VSADDITEMFLAGS.VSADDITEM_AllowHiddenTreeView;
+
+                    string strBrowseLocations = _projectTree.TreeProvider.GetAddNewItemDirectory(node);
+                    string strFilter = string.Empty;
+                    Guid addItemTemplateGuid = Guid.Empty;  // Let the dialog ask the hierarchy itself
```
davkean: I find it easier to parse if these declarations are closer to where they are actually used, ie below the LoadPackageString.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS2kAddItemCommandHandler.cs
davidwengier: > I couldn't connect how this was related to the TemplateDetails.

Yes, I can see how that could be confusing. I may have gone too far with the feedback of just wanting to see the table of commands, and not have to worry about the implementation details.

davkean: A static method to still simplify the add would help alleviate this confusion.

davkean: Or constructor overload.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.TemplateDetails.cs
davidwengier: Good idea, solves both problems.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/VS97AddItemCommandHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+           // Command Id                                Capability                      DirNamePackageGuid          DirNameResourceId                                       TemplateNameResourceId
```
davkean: TemplateNamePackageGuid

davidwengier: No, this is actually TemplateNameResourceId. This is another convenience overload for when both resources come from the same package, but clearly its more inconvenient than it appears so I'll just specify things twice :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                    await _projectVsServices.ThreadingService.SwitchToUIThread();
+
+                    // Get the strings from the legacy package
```
davkean: This is not necessarily the legacy package. I would add  comment on _why_ we're doing this. It's now obvious to both of us based on past discussions, but not so obvious to maintainers.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
davkean: _commandToTemplateMap? Or _templateMap? It doesn't contain commands, it contains templates.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+                    ErrorHandler.ThrowOnFailure(_vsShell.Value.LoadPackageString(ref result.DirNamePackageGuid, result.DirNameResourceId, out string dirName));
```
davkean: Would consider an extension method that removes the "COM-ness" to make it easier to read/consume.

davidwengier:
Thanks for all of the feedback! I've pushed an update to address it, and as a lot is readability related, let me know if there is anything I've misunderstood.

jmarolf:
:shipit: hard to reason about what command handling is doing in VS in general but looks good to me!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,128 @@
...
+                                           __VSADDITEMFLAGS.VSADDITEM_AllowHiddenTreeView;
+
+                string strBrowseLocations = _projectTree.TreeProvider.GetAddNewItemDirectory(node);
+
+                await _projectVsServices.ThreadingService.SwitchToUIThread();
+
+                // Look up the resources from each package to get the strings to pass to the Add Item dialog.
+                // These strings must match what is used in the template exactly, including localized versions. Rather than relying on
+                // our localizations being the same as the VS repository localizations we just load the right strings using the same
+                // resource IDs as the templates themselves use.
```
davkean: 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractAddItemCommandHandler.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        protected static readonly Guid LegacyCSharpPackageGuid = new Guid("{FAE04EC1-301F-11d3-BF4B-00C04F79EFBC}");
+        protected static readonly Guid LegacyVBPackageGuid = new Guid("{164B10B9-B200-11d0-8C61-00A0C91E29D5}");
+        private readonly ConfiguredProject _configuredProject;
+        private readonly IPhysicalProjectTree _projectTree;
+        private readonly IUnconfiguredProjectVsServices _projectVsServices;
+        private readonly IVsUIService<IVsAddProjectItemDlg> _addItemDialog;
+        private readonly IVsUIService<IVsShell> _vsShell;
+
+        [ImportingConstructor]
```
davkean: NIT: This attribute is doing nothing.

PR #4753, opened 2019-04-25
.NET 4.8 compatibility: Preventing throwing an error from GetHashCode functions

M-Lipin:
.NET 4.8 Accessibility changes expect that ComboBox items provide hash codes and do not throw an exception when calling its GetHashCode function so removing overriden GetHashCode functions (which simpy throw NotImplemented exception) from ApplicationPropPageVBWPF.StartupUri​, StartupObject and StartupObjectOrUri to restore default object's behavior when providing hash code (generating an integer).

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
+++ b/src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
@@ -3,7 +3,7 @@
   <Import Project="..\VisualStudioDesigner.props"/>
   <PropertyGroup>
     <!-- TODO: Function doesn't return a value on all code paths (https://github.com/dotnet/project-system/issues/2592) -->
-    <NoWarn>$(NoWarn);42353</NoWarn>
+    <NoWarn>$(NoWarn);42353;NU5125</NoWarn>
     <StrongNameKeyId>Microsoft</StrongNameKeyId>
     <TargetFramework>net46</TargetFramework>
     <!-- Nuget -->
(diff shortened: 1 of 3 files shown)


jmarolf:
@dmonroym I believe you took a look at this before.  Can you pull this down and confirm that this fixes the issue on 4.8?

jmarolf:
@M-Lipin is it intentional that you want to target 15.9 with this change?

M-Lipin:
> I believe you took a look at this before.  Can you pull this down and confirm that this fixes the issue on 4.8?

Yes, there is an issue filed in .NET 4.8, tested and confirmed that this is due to throwing an exception by ComboBox item.

M-Lipin:
> @M-Lipin is it intentional that you want to target 15.9 with this change?

This is intentional, we have an issue filed with 15.9.x target: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/855137/

Also there is similar issue for 16.0.x

jmarolf:
@M-Lipin do we need to back port to 15.0 as well?

jmarolf:
@M-Lipin do we need to back port to 15.0 as well?

jmarolf:
@M-Lipin also, ignore CI results for this. The queues that 15.9 CI were written under no longer exist

jmarolf:
@M-Lipin I pushed a change to ignore NU5125

M-Lipin:
> @M-Lipin do we need to back port to 15.0 as well?

It seems to me we do not need this, as this is not opted to use accessibility improvements introduced in .NET 4.8: there is no "Switch.UseLegacyAccessibilityFeatures.3=false"

`<AppContextSwitchOverrides value="Switch.System.Threading.ThrowExceptionIfDisposedCancellationTokenSource=false;Switch.System.Windows.Forms.DoNotSupportSelectAllShortcutInMultilineTextBox=false;Switch.UseLegacyAccessibilityFeatures=false;Switch.UseLegacyAccessibilityFeatures.2=false" />`

M-Lipin:
@jmarolf, @dmonroym, @davidwengier, thank for helping with this PR and for the code review!

jmarolf:
I think we want to take https://github.com/dotnet/project-system/pull/4758 into 15.9, let that flow to out dev16.0.x and then insert that change into 16.0 instead of merging this PR

M-Lipin:
> I think we want to take #4758 into 15.9, let that flow to out dev16.0.x and then insert that change into 16.0 instead of merging this PR

Thank you, I'll cancel this PR.

PR #4754, opened 2019-04-25
Dev16.0 - .NET 4.8 compatibility: Preventing throwing an error from GetHashCode functions

M-Lipin:
.NET 4.8 Accessibility changes expect that ComboBox items provide hash codes and do not throw an exception when calling its GetHashCode function so removing overriden GetHashCode functions (which simpy throw NotImplemented exception) from ApplicationPropPageVBWPF.StartupUri​, StartupObject and StartupObjectOrUri to restore default object's behavior when providing hash code (generating an integer).

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/WPF/ApplicationPropPageVBWPF.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/WPF/ApplicationPropPageVBWPF.vb
@@ -934,10 +934,6 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages.WPF
                 Return False
             End Function
 
-            Public Overrides Function GetHashCode() As Integer
-                Throw New NotImplementedException()
-            End Function
-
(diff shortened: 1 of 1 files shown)


davkean:
This current branch will require us to be re shipped in 16.0.x. Is that the intention? cc @jjmew

jjmew:
> This current branch will require us to be re shipped in 16.0.x. Is that the intention? cc @jjmew

We need this change for 15.9 and 16.0. @Pilchie right?

Pilchie:
Yes, the current plan is to service this in 16.0.

jmarolf:
I think we want to take https://github.com/dotnet/project-system/pull/4758 into 15.9, let that flow to out dev16.0.x and then insert that change into 16.0 instead of merging this PR

M-Lipin:
Thank you for the review and for the alternative fix!

M-Lipin:
Closing current PR to use alternative fix.

PR #4758, opened 2019-04-26
implement GetHashCode instead of throwing

jmarolf:
alternative to https://github.com/dotnet/project-system/pull/4754

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/WPF/ApplicationPropPageVBWPF.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/WPF/ApplicationPropPageVBWPF.vb
@@ -930,7 +930,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages.WPF
             End Function
 
             Public Overrides Function GetHashCode() As Integer
-                Throw New NotImplementedException()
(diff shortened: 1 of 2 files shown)


jmarolf:
Confirmed that [this](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2627112) signed build succeeded with these changes. @Pilchie for approval to merge.

src/Microsoft.VisualStudio.Editors/PropPages/WPF/ApplicationPropPageVBWPF.vb
```diff
@@ -930,7 +930,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages.WPF
             End Function
 
             Public Overrides Function GetHashCode() As Integer
-                Throw New NotImplementedException()
+                Return -1937169414 + EqualityComparer(Of String).Default.GetHashCode(Value.ToLowerInvariant)
```
Pilchie: I think you want `StringComparer.OrdinalIgnoreCase.GetHashCode(Value)`  My understanding is that `OrdinalIgnoreCase` and `ToLowerInvariant` are different

jmarolf: yeah, thats the correct fix.

Issue #4742, opened 2019-04-23
Operation Progress contributions from WorkspaceProjectContextHostInstance.Evaluation don't complete


adrianvmsft:
**Visual Studio Version**:
d16.1stg
**Summary:**
Loading Roslyn solution shows tasks in TSC that never complete

**Steps to Reproduce**:
1. Load Roslyn solution
2. Wait for a while and click the Task Status Center in bottom left corner

**Expected Behavior**:
Intellisense progress in TSC to eventually complete

**Actual Behavior**:
1 project never finishes loading "Intellisense" stage:
D:\Repos\roslyn\src\Workspaces\Remote\Core\Microsoft.CodeAnalysis.Remote.Workspaces.csproj

**User Impact**:
Operations waiting for intellisense completion may never get executed.
Solution is displayed as perpetually in-progress

0:056> !DumpObj /d 35b1865c
Name:        Microsoft.VisualStudio.ProjectSystem.DataProgressTrackerService+OutputDataSourceInfo
MethodTable: 0daa31d4
EEClass:     0d8b4e18
Size:        48(0x30) bytes
       c:\program files (x86)\microsoft visual studio\2019\d16.1stg\common7\ide\commonextensions\microsoft\project\Microsoft.VisualStudio.ProjectSystem.Implementation.dll
Fields:
      MT    Field   Offset                 Type VT     Attr    Value Name
0b57b190  4000305        4 ...easableSyncObject  0 instance 00000000 syncObject
72e035d4  4000306        8       System.Boolean  1 instance        0 synchronousDisposal
0b46c488  4000307        9          System.Byte  1 instance        0 lifecycle
72e035d4  4000308        a       System.Boolean  1 instance        0 wasInitialized
0daa2fdc  4000982        c ...essTrackerService  0 instance 11b769cc dataProgressTrackerService
1819a9e8  4000983       10 ...TasksMonitorRelay  0 instance 35b1870c tasksMonitor
7291a6ec  4000984       14   System.IDisposable  0 instance 00000000 taskDependency
72dc0994  4000985       28         System.Int32  1 instance        0 isUpToDate
00000000  4000986       18                       0 instance 1486604c inputDataSources
0dae3cc0  4000987       1c ...rOutputDataSource  0 instance 35b18644 <DataSource>k__BackingField
00000000  4000988       20                       0 instance 14770404 <LatestDataSourceVersions>k__BackingField
[...]

PR #4760, opened 2019-04-29
Hook up operation progress before dataflow

davkean:
Fixes: #4742

There was a race where we'd occasionally get called back on dataflow before the registration occured and we'd null reference.


adrianvmsft:
👍

Issue #4765, opened 2019-04-29
Replace deprecated DpiHelper with DpiAwareness


etbyrd:
Visual Studio is moving from a system-level DPI awareness `DpiHelper` to a per monitor DPI awareness `DpiAwareness`. We need to convert as well.

etbyrd:
Fixed in #4841

PR #4766, opened 2019-04-29
Replacing deprecated DpiHelper with DpiAwareness

etbyrd:
This PR replaces all uses of the deprecated DpiHelper and uses DpiAwareness instead to provide per-monitor DPI awareness. Some of the methods had 1:1 conversions and others required a little more refactoring. I am still testing this, but everything looks good so far.

Fixes: https://github.com/dotnet/project-system/issues/4765

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
+++ b/src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
@@ -244,16 +244,16 @@ Namespace Microsoft.VisualStudio.Editors.Common
             End Set
         End Property
 
-        Private Shared Sub PaintArrow(g As Graphics, dropDownRect As Rectangle)
+        Private Sub PaintArrow(g As Graphics, dropDownRect As Rectangle)
             Dim middle = New Point(Convert.ToInt32(dropDownRect.Left + dropDownRect.Width / 2), Convert.ToInt32(dropDownRect.Top + dropDownRect.Height / 2))
 
             ' if the width is odd - favor pushing it over one pixel right.
             middle.X += (dropDownRect.Width Mod 2)
 
-            Dim leftOffset = DpiHelper.LogicalToDeviceUnitsX(3)
-            Dim rightOffset = DpiHelper.LogicalToDeviceUnitsX(3)
-            Dim topOffset = DpiHelper.LogicalToDeviceUnitsY(1)
-            Dim bottomOffset = DpiHelper.LogicalToDeviceUnitsY(2)
+            Dim leftOffset = DpiAwareness.LogicalToDeviceUnits(Handle, 3)
+            Dim rightOffset = DpiAwareness.LogicalToDeviceUnits(Handle, 3)
+            Dim topOffset = DpiAwareness.LogicalToDeviceUnits(Handle, 1)
+            Dim bottomOffset = DpiAwareness.LogicalToDeviceUnits(Handle, 2)
 
             Dim arrow = New Point(2) {
                 New Point(
--- a/src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
+++ b/src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
@@ -6,6 +6,7 @@ Imports System.Windows.Forms
 Imports System.Windows.Forms.VisualStyles
 
 Imports Microsoft.VisualStudio.PlatformUI
+Imports Microsoft.VisualStudio.Utilities
 
 Namespace Microsoft.VisualStudio.Editors.Common
 
@@ -29,7 +30,7 @@ Namespace Microsoft.VisualStudio.Editors.Common
             MyBase.New()
 
             AccessibleRole = AccessibleRole.SplitButton
-            _pushButtonWidth = DpiHelper.LogicalToDeviceUnitsX(_pushButtonWidth)
+            _pushButtonWidth = DpiAwareness.LogicalToDeviceUnits(Handle, _pushButtonWidth)
         End Sub
 
         Protected Overrides Sub OnPaint(pevent As PaintEventArgs)
@@ -42,8 +43,8 @@ Namespace Microsoft.VisualStudio.Editors.Common
 
             ButtonRenderer.DrawButton(g, bounds, State)
 
-            Dim singlePixelWidth = DpiHelper.LogicalToDeviceUnitsX(1)
-            Dim dividerLineVerticalPadding = DpiHelper.LogicalToDeviceUnitsY(4)
+            Dim singlePixelWidth = DpiAwareness.LogicalToDeviceUnits(Handle, 1)
+            Dim dividerLineVerticalPadding = DpiAwareness.LogicalToDeviceUnits(Handle, 4)
 
             _dropDownRectangle = New Rectangle(
                 x:=bounds.Right - _pushButtonWidth - singlePixelWidth,
@@ -108,8 +109,8 @@ Namespace Microsoft.VisualStudio.Editors.Common
 
             If Focused Then
                 bounds.Inflate(
-                    width:=DpiHelper.LogicalToDeviceUnitsX(-4),
-                    height:=DpiHelper.LogicalToDeviceUnitsY(-4))
+                    width:=DpiAwareness.LogicalToDeviceUnits(Handle, -4),
+                    height:=DpiAwareness.LogicalToDeviceUnits(Handle, -4))
 
                 ControlPaint.DrawFocusRectangle(g, bounds)
             End If
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ReferencePathsPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ReferencePathsPropPage.vb
@@ -6,7 +6,9 @@ Imports System.Windows.Forms
 
 Imports Microsoft.VisualStudio.Editors.Common
 Imports Microsoft.VisualStudio.PlatformUI
+Imports Microsoft.VisualStudio.Utilities
 Imports Microsoft.Win32
+Imports Microsoft.VisualStudio.Imaging
 
 Imports VSLangProj80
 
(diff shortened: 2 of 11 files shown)


jmarolf:
@etbyrd this looks good to me, but could we also make sure to ban use of the `DpiHelper` type so we don't re-introduce bugs here?

etbyrd:
Talked with @jmarolf offline and will add `DpiHelper` to the banned symbols in a separate PR

src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
```diff
@@ -29,7 +30,7 @@ Namespace Microsoft.VisualStudio.Editors.Common
             MyBase.New()
 
             AccessibleRole = AccessibleRole.SplitButton
-            _pushButtonWidth = DpiHelper.LogicalToDeviceUnitsX(_pushButtonWidth)
+            _pushButtonWidth = DpiAwareness.LogicalToDeviceUnits(Handle, _pushButtonWidth)
```
davkean: Can you point me to docs here? I want to know why the Handle is interesting. For example, here, this control isn't actually visible anywhere and you are force loading the Handle (I think it gets added to a hidden window) so might not work as expected.

davkean:
Can we squash this on merge? As they are all the same effective change.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
```diff
@@ -446,15 +447,19 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
                 Dim assemblyImage As Bitmap = GetManifestBitmapTransparent("assembly.bmp", Color.FromArgb(255, 0, 255))
                 Dim namespaceImage As Bitmap = GetManifestBitmapTransparent("namespace.bmp", Color.FromArgb(255, 0, 255))
                 Dim objectImage As Bitmap = GetManifestBitmapTransparent("object.bmp", Color.FromArgb(255, 0, 255))
+                'Scale the images for per monitor DPI
+                Handle.LogicalToDeviceUnits(assemblyImage.Width)
```
davkean: What exactly are these lines doing? You aren't using the result.

src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
jmarolf: I forwarded you the *Impactful change notification* that went out.  Don't know if there are other docs here.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
davkean: Looking at the internal email it states:

> You should use the Visual Studio image service to replace your usage of DpiHelper’s image scaling methods.

src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
davkean: I've asked offline to make sure this works as expected.

src/Microsoft.VisualStudio.Editors/PropPages/ReferencePathsPropPage.vb
```diff
@@ -33,23 +35,26 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
             _moveUpImageOriginal = MoveUp.Image
             _moveDownImageOriginal = MoveDown.Image
             _removeFolderImageOriginal = RemoveFolder.Image
 
-            ' Rescale images
-            DpiHelper.LogicalToDeviceUnits(_moveUpImageOriginal)
-            DpiHelper.LogicalToDeviceUnits(_moveDownImageOriginal)
-            DpiHelper.LogicalToDeviceUnits(_removeFolderImageOriginal)
+            'Rescale the images for high DPI
+            Handle.LogicalToDeviceUnits(_moveUpImageOriginal.Width)
```
davkean: As below, this isn't actually resizing anything.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
jmarolf: I believe the recommendation is to use image monikers throughout instead of scaling bitmaps manually.

jmarolf: If we can find an existing bitmap here: https://glyphlist.azurewebsites.net/knownmonikers/ we can just use its moniker instead

src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
```diff
@@ -6,6 +6,7 @@ Imports System.Windows.Forms
 Imports System.Windows.Forms.VisualStyles
 
 Imports Microsoft.VisualStudio.PlatformUI
```
dmonroym: you can remove this

dmonroym:
> 

These should be changed as well

---
Refers to: src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerView.vb:471 in 4131caa. [](commit_id = 4131caa636e901c6739787ee48d9e9347ae9b90f, deletion_comment = False)

davkean:
Can we ban in the same PR so that you are sure you've caught all usages? Looks like some are missing.

src/Microsoft.VisualStudio.Editors/PropPages/ReferencePathsPropPage.vb
```diff
@@ -6,7 +6,9 @@ Imports System.Windows.Forms
 
 Imports Microsoft.VisualStudio.Editors.Common
 Imports Microsoft.VisualStudio.PlatformUI
+Imports Microsoft.VisualStudio.Utilities
 Imports Microsoft.Win32
+Imports Microsoft.VisualStudio.Imaging
```
dmonroym: I don't believe Imaging nor PlatformUI are necessary here

src/Microsoft.VisualStudio.Editors/Common/SplitButton.vb
davkean: Conclusion internally:

> None of our WinForms code should be running in PMA, only System. Which means relying on the Handle before parented should be fine, assuming we wrap dialogs in [a SystemAware scope](https://github.com/dotnet/project-system/issues/4493). This means we don’t need to react to DPI changes either.

@etbyrd Can we make sure that that all the UI we’re touching is in System-Aware DPI mode via winspy++? Drag the target to the control and then click the DPI tab.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
drewnoakes: @jmarolf that's a great link, thanks!

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
```diff
@@ -446,15 +447,19 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
...
+                Handle.LogicalToDeviceUnits(objectImage.Width)
+                Handle.LogicalToDeviceUnits(objectImage.Height)
 
                 Dim treeViewIcons As ImageList = New ImageList()
                 treeViewIcons.Images.Add(assemblyImage)
                 treeViewIcons.Images.Add(namespaceImage)
                 treeViewIcons.Images.Add(objectImage)
 
-                'Scale the imagelist for High DPI
-                DpiHelper.LogicalToDeviceUnits(treeViewIcons)
```
drewnoakes: @etbyrd I suspect this method was resizing the images within the `ImageList`. I'm not sure if image monikers can be used with a regular tree view. If not, you may need to find or write an equivalent helper method for `ImageList`. Note that `ImageList` removes all its contents when you change its size, so if you're going to specify a size on the `ImageList` you must do so before adding images.

I wrote some similar code for GitExtensions a while back which may be helpful:

https://github.com/gitextensions/gitextensions/blob/c4764500400604a18869d6d3b78eda9271727210/GitUI/BranchTreePanel/RepoObjectsTree.cs#L84-L129

(Using that project's [`DpiUtil`](https://github.com/gitextensions/gitextensions/blob/master/GitExtUtils/GitUI/DpiUtil.cs) class)

davkean: The expectation is that we move to image monikers to avoid the bitmap stretching.

drewnoakes: Sounds good. I don't know how this is done with a WinForms `TreeView`. AFAIK you can only set icons using `ImageList`.

davkean: They draw using vectors, and then give a bitmap of that size, rather than resizing a fixed bitmap.

davkean:
I would separate this into a few PRs so we can get some changes in that just work:

- Replacing all usages of DpiHelper.LogicalToDeviceUnits except when used for images
- Move images over to use ImageMonikers where there's a drop in replacement
- Figure out what to do with images where there's not existing image monikers

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
drewnoakes: I would presume there are hand-drawn versions for 96dpi as scaling vector art to 16/24/32px usually looks terrible ([see](https://www.pushing-pixels.org/2011/11/04/about-those-vector-icons.html)). Regardless, I agree image monikers are the better solution.

Do you know what API can be used to create an `ImageList` from a list of image monikers?

If this has to support PMA then would the `ImageList` need to be recreated on changing monitors?

davkean: See my comment here: https://github.com/dotnet/project-system/pull/4766#discussion_r279569328. All of our WinForms code should be System-Aware and not PMA, so we don't need to react to changing monitors.

etbyrd:
Moving this to separate PR's per @davkean's suggestion. First one up here: https://github.com/dotnet/project-system/pull/4780

PR #4768, opened 2019-04-29
Add richnav support for PRs

dmonroym:
More info at https://aka.ms/IntelliNavDocs & https://code.visualstudio.com/blogs/2018/12/04/rich-navigation

### DIFF
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -70,7 +70,7 @@ jobs:
     _configuration: Debug
   timeoutInMinutes: 240
   steps:
-    - script: $(Build.SourcesDirectory)\build\script\CIRestore.cmd -configuration $(_configuration) -prepareMachine
+    - script: $(Build.SourcesDirectory)\build.cmd /build /ci /configuration ${{ parameters.configuration }}
     - task: RichCodeNavIndexer@0
       continueOnError: true
       inputs:
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -61,3 +61,21 @@ jobs:
     pool: 
       name: NetCorePublic-Pool
       queue: BuildPool.Windows.Amd64.VS2019.Pre.ES.Open
+
+- job: RichNav
+  pool:
+    name: NetCorePublic-Pool
+    queue: BuildPool.Windows.10.Amd64.VS2019.Pre.Open
+  variables:
+    _configuration: Debug
+  timeoutInMinutes: 240
+  steps:
+    - script: $(Build.SourcesDirectory)\build\script\CIRestore.cmd -configuration $(_configuration) -prepareMachine
+    - task: RichCodeNavIndexer@0
+      continueOnError: true
+      inputs:
+        serviceConnection: 'rich-nav'
+        nugetServiceConnection: 'rich-nav-nuget'
+        githubServiceConnection: 'dotnet'
+        languages: 'csharp'
+        serviceEndpoint: 'https://prod.richnav.vsengsaas.visualstudio.com'
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -72,9 +72,10 @@ jobs:
   steps:
     - script: $(Build.SourcesDirectory)\build.cmd /no-build /no-test /no-deploy
     - task: UseDotNet@2
-        inputs:
-            packageType: sdk
-            version: 2.1.503
+      displayName: 'Use .NET Core sdk'
+      inputs:
+        packageType: sdk
+        version: 2.1.503
     - task: RichCodeNavIndexer@0
       continueOnError: true
       inputs:
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -72,7 +72,6 @@ jobs:
   steps:
     - script: $(Build.SourcesDirectory)\build.cmd /no-build /no-test /no-deploy
     - task: UseDotNet@2
-        displayName: 'Use .NET Core sdk'
         inputs:
             packageType: sdk
             version: 2.1.503
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -62,12 +62,6 @@ jobs:
       name: NetCorePublic-Pool
       queue: BuildPool.Windows.Amd64.VS2019.Pre.ES.Open
 
-- task: UseDotNet@2
-  displayName: 'Use .NET Core sdk'
-  inputs:
-    packageType: sdk
-    version: 2.1.503
-
 - job: RichNav
   pool:
     name: NetCorePublic-Pool
@@ -77,6 +71,11 @@ jobs:
   timeoutInMinutes: 240
   steps:
     - script: $(Build.SourcesDirectory)\build.cmd /no-build /no-test /no-deploy
+    - task: UseDotNet@2
+        displayName: 'Use .NET Core sdk'
+        inputs:
+            packageType: sdk
+            version: 2.1.503
     - task: RichCodeNavIndexer@0
       continueOnError: true
       inputs:
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -62,6 +62,12 @@ jobs:
       name: NetCorePublic-Pool
       queue: BuildPool.Windows.Amd64.VS2019.Pre.ES.Open
 
+- task: UseDotNet@2
+  displayName: 'Use .NET Core sdk'
+  inputs:
+    packageType: sdk
+    version: 2.1.503
+
 - job: RichNav
   pool:
     name: NetCorePublic-Pool
@@ -70,7 +76,7 @@ jobs:
     _configuration: Debug
   timeoutInMinutes: 240
   steps:
-    - script: $(Build.SourcesDirectory)\build.cmd /build /ci /configuration $(_configuration)
+    - script: $(Build.SourcesDirectory)\build.cmd /no-build /no-test /no-deploy
     - task: RichCodeNavIndexer@0
       continueOnError: true
       inputs:
--- a/build/ci/unit-tests.yml
+++ b/build/ci/unit-tests.yml
@@ -70,7 +70,7 @@ jobs:
     _configuration: Debug
   timeoutInMinutes: 240
   steps:
-    - script: $(Build.SourcesDirectory)\build.cmd /build /ci /configuration ${{ parameters.configuration }}
+    - script: $(Build.SourcesDirectory)\build.cmd /build /ci /configuration $(_configuration)
     - task: RichCodeNavIndexer@0
       continueOnError: true
       inputs:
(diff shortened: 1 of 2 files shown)


build/azure-pipelines.yml
```diff
@@ -65,6 +65,14 @@ jobs:
         publishRunAttachments: true
       continueOnError: true
       condition: always()
+    - task: RichCodeNavIndexer@0
```
jmarolf: I assume we don't want to run this for both debug *and* release.  probably should have a condition so that it only runs on debug?

jmarolf:
Looking at the failed test run I don't think this feature is enabled in our AzDO instance
```console
The pipeline is not valid. 
Job Windows: 
  Step RichCodeNavIndexer input serviceConnection references service connection rich-nav which could not be found. 
  The service connection does not exist or has not been authorized for use. 
  For authorization details, refer to https://aka.ms/yamlauthz.,Job Windows: Step RichCodeNavIndexer input nugetServiceConnection references service connection rich-nav-nuget which could not be found. 
  The service connection does not exist or has not been authorized for use. 
  For authorization details, refer to https://aka.ms/yamlauthz.
```

dmonroym:
Responding to both comments as they're related. I didn't fully configure the task correctly as I'm waiting for @chcosta to enable these service connections - for now it's just the default of what's asked from richnav team. The extension is enabled but both rich-nav and rich-nav-nuget are not (and depending on the naming he gives I'll have to correct them accordingly)

build/azure-pipelines.yml
```diff
@@ -66,6 +66,23 @@ jobs:
...
       condition: always()
 
+- job: RichNav
+  pool:
+    vmImage: 'vs2017-win2016'
+  variables:
+    _configuration: Debug
+  timeoutInMinutes: 20
+  steps:
+    - script: $(Build.SourcesDirectory)\build\CIBuild.cmd -configuration $(_configuration) -prepareMachine
```
jmarolf: Reading through the docs I don't think we need to build

build/azure-pipelines.yml
```diff
@@ -66,6 +66,23 @@ jobs:
...
+    - task: RichCodeNavIndexer@0
+      continueOnError: true
+      inputs:
+        serviceConnection: 'rich-nav'
+        nugetServiceConnection: 'rich-nav-nuget'
+        githubServiceConnection: 'dotnet'
+        languages: 'csharp'
```
jmarolf: Have we filed a bug on them for supporting VB?

build/azure-pipelines.yml
jmarolf: reading through _the code_ I know we don't need to build. but we do need to restore the project.

Today `CIBuild.cmd` does the following by default
```console
-restore -build -test -pack -ci
```

We only need to do
```console
-restore -ci
```

So we should probably have a new script named `CIRestore.cmd` whose contents look like this
```console
echo off
powershell -ExecutionPolicy ByPass %~dp0Build.ps1 -restore -ci %*
exit /b %ErrorLevel%
```

which we call in the yaml file like so:
```yaml
  steps:
    - script: $(Build.SourcesDirectory)\build\CIRestore.cmd -configuration $(_configuration) -prepareMachine
```

build/azure-pipelines.yml
jmarolf: I'll talk to @heejaechang tomorrow about this. Looks trivial

davkean:
@dmonroym Is there still stuff blocking this?

dmonroym:
The team owning this still doesn't seem to be ready. I'll close this now so it's not polluting our workspace.

---
In reply to: [496380159](https://github.com/dotnet/project-system/pull/4768#issuecomment-496380159) [](ancestors = 496380159)

dmonroym:
Testing this out again to see if disk space is no longer an issue since we've moved images to dnceng hosted VMs

build/ci/unit-tests.yml
```diff
@@ -61,3 +61,21 @@ jobs:
     pool: 
       name: NetCorePublic-Pool
       queue: BuildPool.Windows.Amd64.VS2019.Pre.ES.Open
+
+- job: RichNav
```
davkean: What is this job sharing from unit-tests? It seems like this should be separate pipeline.

build/ci/unit-tests.yml
```diff
@@ -61,3 +61,21 @@ jobs:
...
+  pool:
+    name: NetCorePublic-Pool
+    queue: BuildPool.Windows.10.Amd64.VS2019.Pre.Open
+  variables:
+    _configuration: Debug
+  timeoutInMinutes: 240
+  steps:
+    - script: $(Build.SourcesDirectory)\build.cmd /build /ci /configuration $(_configuration)
```
davkean: Does it actually use binaries or just source code? Above is building the tree.

build/script/CIRestore.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
```
davkean: Was this a bad merge that this script still exists?

build/ci/unit-tests.yml
dmonroym: Yes, I'll move it to its own. As of now I was just testing if it would work - doesn't seem to be the case

build/ci/unit-tests.yml
```diff
@@ -61,3 +61,21 @@ jobs:
...
       name: NetCorePublic-Pool
       queue: BuildPool.Windows.Amd64.VS2019.Pre.ES.Open
+
+- job: RichNav
+  pool:
+    name: NetCorePublic-Pool
+    queue: BuildPool.Windows.10.Amd64.VS2019.Pre.Open
+  variables:
+    _configuration: Debug
+  timeoutInMinutes: 240
```
davkean: 4 hours, is that right?

build/script/CIRestore.cmd
dmonroym: I initially created this script as part of supporting richnav - it's no longer needed. Overall this PR is just for testing purposes as of now to see if this feature could work. I'm in contact with the team to see what work's been done on their side

davkean: Do you need a concept of restore only?

build/ci/unit-tests.yml
jmarolf: I believe this is correct (it is _very_ slow)

build/ci/unit-tests.yml
jmarolf: It needs to be able to instantiate a `MSBuildWorkspace` (or the moral equivalent) so I believe a restore is all that is necessary.

jmarolf:
@dmonroym 	I believe you will need to install the 2.1 dotnet sdk in order for this to work as RichNav is a global tool. project-system does not install the SDK as it has no dependencies on it. see [dotnet-install-scripts](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) (you will need to use powershell for this though, don't tell @davkean)

davkean:
Why do you have to manually install the SDK? We do have a dependency on the SDK and one is always installed.

build/ci/unit-tests.yml
davkean: I think we'd probably only want to trigger this once a day, or avoid overlapping jobs.

jmarolf:
>Why do you have to manually install the SDK? We do have a dependency on the SDK and one is always installed.

@davkean 	where do we install it? I can't make heads or tails of the `build` folder now-a-days

davkean:
The image already has it installed with the .NET Core workload.

jmarolf:
>The image already has it installed with the .NET Core workload.

Right, my mistake. We need 2.1 for the global tool installation to succeed? Taggin @wli3 	as I thought global tools were supposed to "just work" on newer sdks

```log
[command]"C:\Program Files\dotnet\dotnet.exe" F:\workspace\_work\_temp\RichCodeNav\.store\RichCodeNav.ClientTool\0.1.958-alpha-gf5741143e3\richcodenav.clienttool\0.1.958-alpha-gf5741143e3\tools\netcoreapp2.1\any\RichCodeNav.ClientTool.dll create-snapshot https://github.com/dotnet/project-system/pull/4768 --setting ServiceBaseUrl=https://prod.richnav.vsengsaas.visualstudio.com -f File --gitProviderPAT *** --vsclkToken ***
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '2.1.0' was not found.
  - The following frameworks were found:
      3.0.0-preview7-27912-14 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
```

build/ci/unit-tests.yml
arjun27: (I'm from the rich nav team) Yes, thanks for calling it out. Improving time to build is a critical requirement for us, and we have been working on revamping our internals to improve incrementality, so that indexing does not take this long. We're testing these changes in dev currently. cc @ BertanAygun

build/ci/unit-tests.yml
arjun27: The indexer does not use the binaries

arjun27:
Thanks @dmonroym, @davkean, @jmarolf! I'm from the rich nav team, and yes you're right to point out the dependency on 2.1 of the SDK. Looping in @jepetty and @BertanAygun to confirm if we can support 3.x

build/ci/unit-tests.yml
davkean: Okay so we need a restore concept in the build.

davkean:
@jmarolf Where did you hit that? On the build machine? I didn't think we had any preview builds except spanish.

build/ci/unit-tests.yml
davkean: I'm working on it.

jmarolf:
>Where did you hit that? On the build machine? I didn't think we had any preview builds except spanish.

which version of the sdk do we expect to be installed with 16.3? Thats the version that comes with our VMs apparently

jmarolf:
@arjun27 	I believe you will want to package your global tool to roll forward automatically (see [here](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md)).  

Here is an example PR that will added for one of our global tools: https://github.com/dotnet/format/pull/22

build/ci/unit-tests.yml
davkean: Only after this: https://github.com/dotnet/project-system/pull/5327, you can now do a restore only by doing the following:

```
build /no-build /no-test /no-deploy
```

build/ci/unit-tests.yml
jmarolf: Last time I looked at the perf for intellinav most of the time was taken up serializing json...

dmonroym:
RichNav is not ready - still over 4hrs, runs on 2.1 SDK, and it requires VS2017

davkean:
Yeah that's a blocker.

davkean:
@jepetty @BertanAygun @arjun27 We just moved the http://index indexing over to our official build: https://github.com/dotnet/project-system/pull/5374/files and it added 1 minute to the build time. Please use that as your baseline.

PR #4775, opened 2019-04-30
Add .xaml file to setup authoring

tmeschter:
We depend on the PotentialEditorConfigFiles rule, so we need to remember
to ship the corresponding .xaml file.


Pilchie:
We've forgotten this a bunch of times - can we figure out how to avoid doing this in the future?

davidwengier:
> We've forgotten this a bunch of times - can we figure out how to avoid doing this in the future?

I'll see if I can add something to the rest of the tests I'm building around rules,

davkean:
I'd consider just auto-generating this setup file.

PR #4780, opened 2019-05-01
Update LogicalToDeviceUnits calls with DpiAwareness 

etbyrd:
Splitting https://github.com/dotnet/project-system/pull/4766 into separate parts. This PR will update the LogicalToDeviceUnits calls with the corresponding DpiAwareness method, except those regarding images.


jjmew:
Relates to #4765

PR #4783, opened 2019-05-02
Watch for file changes to TempPE inputs

davidwengier:
Part of #4163 

Hooks up file watching to TempPE so changes to inputs are responded to.


davkean:
I'm a little concerned by how much work this pushes to the UI thread in the case of a large branch switch, why is compilation on the UI thread required?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPEBuildManager.cs
```diff
@@ -373,9 +387,46 @@ public Task UnloadAsync()
...
+        {
+            IVsAsyncFileChangeEx fileChangeService = await _fileChangeService.GetValueAsync();
+
+            await fileChangeService.UnadviseFileChangeAsync(cookie);
+        }
+
         protected override async Task DisposeCoreAsync(bool initialized)
         {
             DisposeTaskSchedulers();
+            await UnregisterFileWatchersAsync();
```
davkean: There's a race; you could be in the middle of ApplyAsync when Dispose is called, this will unregister the file watchers but then ApplyAsync will add more.

PR #4785, opened 2019-05-03
Remove unneeded <SubType> and <DependentUpon>

davkean:
`<SubType>` is now stored in the user file, `<DependentUpon>` is now implicitly determined.


sharwell:
:hooray:

PR #4791, opened 2019-05-06
LOC CHECKIN | dotnet\project-system master | 20190506

v-mdriml:


tmeschter:
@Pilchie FYI.

Pilchie:
I take this is needed for 16.1?  If so, we should try to get it inserted by 4pm before the snap.

tmeschter:
@Pilchie We can try, but six hours is tight where RPS is involved...

PR #4796, opened 2019-05-07
Use ProjectSystemContractScope.Global

### LINKED ISSUE #4626: Add ProjectSystemContractScope.Global to represent the global scope
We cannot correctly represent our scopes using [ProjectSystemContractScope] because Global is missing. We should add this to CPS:

- [ ] Add Global scope
- [ ] Remove ProjectSystemContractScope2.Global in our tree
- [ ] Remove hack in ImportsMustMatchExportScope to exclude IProjectServiceAccessor

sharwell:
Closes #4626


jmarolf:
Ignore the duplicate `unit-tests (Spanish)` looks like and AzDO bug

davkean:
Thanks @sharwell !

PR #4803, opened 2019-05-08
Revert DependentUpon/SubType changes

davkean:
RESX generation has a dependency on DependentUpon to determine the name of the resource embedded in the dll. 

Fixed: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/888081.

I'll do this in 16.2 another way and we need to think about this given that project system doesn't require it anymore. This just fixes the regression.


davkean:
tag @jjmew

PR #4830, opened 2019-05-17
Remove existing rename type to match file

jmarolf:
simplifying rename implementation to prepare for more changes


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -78,18 +83,131 @@ public async void OnWorkspaceChangedAsync(object sender, WorkspaceChangeEventArg
...
+            if (renamedSolution == null)
+                return;
+
+            await _threadingService.SwitchToUIThread();
+            bool renamedSolutionApplied = _roslynServices.ApplyChangesToSolution(project.Solution.Workspace, renamedSolution);
+
+            if (!renamedSolutionApplied)
+            {
+                string failureMessage = string.Format(CultureInfo.CurrentCulture, VSResources.RenameSymbolFailed, oldNameBase);
+                await _threadingService.SwitchToUIThread();
```
tmeschter: Is this redundant?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -78,18 +83,131 @@ public async void OnWorkspaceChangedAsync(object sender, WorkspaceChangeEventArg
...
+            }
+
+            return solutionToRename;
+        }
+
+        private static Document GetDocument(Project project, string filePath) =>
+            (from d in project.Documents where StringComparers.Paths.Equals(d.FilePath, filePath) select d).FirstOrDefault();
+
+        private static Task<SyntaxNode> GetRootNode(Document newDocument) =>
+            newDocument.GetSyntaxRootAsync();
```
tmeschter: This method seems unnecessary.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -78,18 +83,131 @@ public async void OnWorkspaceChangedAsync(object sender, WorkspaceChangeEventArg
...
+
+        private async Task<bool> CheckUserConfirmation(string oldFileName)
+        {
+            if (_userPromptedOnce)
+            {
+                return _userConfirmedRename;
+            }
+
+            await _threadingService.SwitchToUIThread();
+            bool userNeedPrompt = _environmentOptions.GetOption("Environment", "ProjectsAndSolution", "PromptForRenameSymbol", false);
```
tmeschter: Use a named parameter to clarify what `false` represents.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
jmarolf: No, tests will fail if this is not present

jmarolf: I can try and find a way to unify this logic though

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
jmarolf: This was in the previous implementation.  I will remove/simplify this logic in a future PR

PR #4841, opened 2019-05-20
Update LogicalToDeviceUnits calls with DpiAwareness Part 2

etbyrd:
Part 2 for https://github.com/dotnet/project-system/pull/4780 which will be the final fix for https://github.com/dotnet/project-system/issues/4765

This PR removes the last calls to DPIHelper and also removes the bitmaps in the Reference Paths prop page and the Type Picker prop page in favor of getting image monikers from the image service. There were good monikers to choose from:

ReferencePathsNewMonikers

Changed:
-Move Up Folder
-Move Down Folder
-Remove Folder (The image is not cut off, that is the actual image)

TypePickerNewMonikers

-Assembly
-Namespace
-Object (Was replaced with KnownMonikers.ClassPublic)


src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypePickerDialog.vb
```diff
@@ -443,18 +444,18 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
                 PathSeparator = "."
                 Sorted = True
 
-                Dim assemblyImage As Bitmap = GetManifestBitmapTransparent("assembly.bmp", Color.FromArgb(255, 0, 255))
-                Dim namespaceImage As Bitmap = GetManifestBitmapTransparent("namespace.bmp", Color.FromArgb(255, 0, 255))
-                Dim objectImage As Bitmap = GetManifestBitmapTransparent("object.bmp", Color.FromArgb(255, 0, 255))
+                'Scale the imagelist for High DPI
+                Dim newSize = DpiAwareness.LogicalToDeviceSize(Handle, New Size(96, 96))
```
jmarolf: We talked a little about this in person.  I am not certain this is the correct thing to do here.  Can you try this out with a monitor set to a different DPI setting and confirm that dragging from one to the other redraws correctly?

etbyrd: Yes, it seemed to be drawn correctly across monitors

Issue #4745, opened 2019-04-24
OpenWithSpecific throws "GetNestedHierarchy failed"


jmarolf:
**Visual Studio Version**: 16.1stg + latest project system master

**Summary:**

found while testing https://github.com/dotnet/project-system/pull/4743

**Steps to Reproduce**:
1. Open the property pages in a debug build

```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
GetNestedHierarchy failed



   at Microsoft.VisualStudio.Editors.ApplicationDesigner.ApplicationDesignerView.InitView()

   at Microsoft.VisualStudio.Editors.ApplicationDesigner.ApplicationDesignerWindowPane.PopulateView()

   at Microsoft.VisualStudio.Editors.ApplicationDesigner.ApplicationDesignerWindowPane.ActivateLogicalView(Guid& rguidLogicalView)

   at Microsoft.VisualStudio.Shell.Interop.IVsUIShellOpenDocument.OpenSpecificEditor(UInt32 grfOpenSpecific, String pszMkDocument, Guid& rguidEditorType, String pszPhysicalView, Guid& rguidLogicalView, String pszOwnerCaption, IVsUIHierarchy pHier, UInt32 itemid, IntPtr punkDocDataExisting, IServiceProvider pSPHierContext, IVsWindowFrame& ppWindowFrame)

   at Microsoft.VisualStudio.ProjectSystem.VS.Implementation.Package.ProjectFileDocumentManager.OpenWithSpecific(SimpleHierarchyNode projectItem, UInt32 editorFlags, Guid& editorType, String physicalView, Guid& logicalView, IntPtr docDataExisting, IVsWindowFrame& frame, WindowFrameShowAction windowFrameActi......

<truncated>
---------------------------
Abort   Retry   Ignore   
---------------------------
```

drewnoakes:
Fixed in #4843.

PR #4843, opened 2019-05-20
Remove the Debug.Fail for already having the hierarchy

etbyrd:
Fix for https://github.com/dotnet/project-system/issues/4745

Currently, this seems to be *always* failing when opening the property pages. 

At least one of the reasons `GetNestedHierarchy` is failing here is because the `ItemID` is `VSConstants.VSITEMID_ROOT` and it will always fail if that is the case.

I don't believe this is valuable to have, especially if we hit it every time with no issues with the `Hierarchy`


src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
```diff
@@ -200,7 +200,6 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                         '(Note: we don't expect this code path anymore, as I believe the project designer 
                         '  is always set up using the solution hierarchy because of the fact that it's
                         '  an editor on the project file.)
-                        Debug.Fail("GetNestedHierarchy failed")
```
davidwengier: May as well just remove the whole `Else` branch.

etbyrd: Good point, done.

Issue #4854, opened 2019-05-22
Package restore should be using NuGetTargetMoniker if its exists over TargetFrameworkMoniker


davkean:
For all projects except UWP, `NuGetTargetMoniker` is equivalent to `TargetFrameworkMoniker`. However, in UWP, `NuGetTargetMoniker` represents the `TargetPlatformMoniker`: https://github.com/dotnet/sdk/blob/bb764dea3b6b7173522688fffe0f48640d714833/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets#L46.

This is what gets passed to NuGet in command-line restore.

In 16.1 we started using TargetFrameworkMoniker vs TargetFramework so that we can support projects that are not-Microsoft.NET.SDK based, this broke folks using MSBuild.Extras. We should change this to report `NuGetTargetMoniker` if it exists followed by `TargetFrameworkMoniker`.

MSiccDev:
Any ideas to work around this until it gets fixed and rolled out? This completely blocks me from working on my current project.

1Cor125:
Also blocked on this.

davkean:
I'm not aware of a workaround at the moment, for those with paid subscriptions can access previous versions of VS via: https://my.visualstudio.com/Downloads?q=Visual%20Studio%202019.

davkean:
We are planning on shipping an update to 16.1 soon with a fix for this.

davkean:
This fix will appear in 16.1.2.

dotMorten:
Thank you @davkean for being awesome as usual 👍

WonyoungChoi:
@davkean I'd like to know more about why you use the TargetFrameworkMoniker instead of the TargetFramework. 
In fact, this change is affecting the Tizen.NET projects.  Tizen.NET projects typically have the following csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>tizen40</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Tizen.NET" Version="4.0.0" />
  </ItemGroup>
</Project>
```
Of course, Microsoft.NET.Sdk doesn't handle "tizen TFM.", So `TargetFrameworkIdentifier` is defined in the build/props of Tizen.NET nuget package. 

[...]

PR #4859, opened 2019-05-23
Use NuGetTargetMoniker instead of TFM if its exists

davkean:
Fixes: https://github.com/dotnet/project-system/issues/4854.

Note: This has localization impact that I believe is okay; if we do not ship localized values - then the user will continue to get the old string.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/PackageRestore/Snapshots/RestoreBuilderTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/PackageRestore/Snapshots/RestoreBuilderTests.cs
@@ -63,6 +63,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
         }
 
         [Fact]
+        public void ToProjectRestoreInfo_RespectsNuGetTargetMonikerIfPresent()
+        {
+            var update = IProjectSubscriptionUpdateFactory.FromJson(@"
+{
+    ""CurrentState"": {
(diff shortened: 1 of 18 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/PackageRestore/Snapshots/RestoreBuilderTests.cs
```diff
@@ -62,6 +62,27 @@ public void ToProjectRestoreInfo_WhenNuGetRestoreRuleMissing_ReturnsEmpty()
             Assert.Empty(targetFramework.Properties);
         }
 
+        [Fact]
+        public void ToProjectRestoreInfo_RespectsNuGetTargetMonikerIfPresent()
```
drewnoakes: Minor: Cross ref GitHub issue here for future reference: https://github.com/dotnet/project-system/issues/4854

davkean: You don't think history gives you that?

drewnoakes: Sure it does. Just not as convenient.

tmeschter:
Non-english users will continue to see the older (but still localized) messages, yes.

jamesmontemagno:
Is this in 16.1.1 or future?

davkean:
16.1.2

PR #4863, opened 2019-05-23
Don't prompt to reload when moving from old format -> SDK-based format

davkean:
ProjectReloaderIntecepter attempts to detect and prompt user to reload when changing from:

``` XML
<Project>
  <PropertyGroup>
    <TargetFramework>net45</TargetFramework>
  </PropertyGroup>
</Project>
```
to

``` XML
<Project>
  <PropertyGroup>
    <TargetFrameworks>net45;netcoreapp2.2</TargetFrameworks>
  </PropertyGroup>
</Project>
```

However, it did it in a way to cause it to prompt in the following:

``` XML
<Project>
  <PropertyGroup>
    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
  </PropertyGroup>
</Project>
```

to:

``` XML
<Project>
  <PropertyGroup>
    <TargetFrameworks>net45</TargetFrameworks>
  </PropertyGroup>
</Project>
```

Despite these being equivalent. Change the implementation to actually consult the dimension providers and avoid it having any knowledge how configurations are determined, which handles above.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Reload/ProjectReloadInterceptor.cs
```diff
@@ -1,62 +1,62 @@
...
         }
 
-        private static bool NeedsForcedReload(ImmutableArray<ProjectPropertyElement> oldProperties, ImmutableArray<ProjectPropertyElement> newProperties)
+        public ProjectReloadResult InterceptProjectReload(ImmutableArray<ProjectPropertyElement> oldProperties, ImmutableArray<ProjectPropertyElement> newProperties)
         {
-            // If user added or removed TargetFramework/TargetFrameworks property, then force a full project reload.
-            (bool hasTargetFramework, bool hasTargetFrameworks) = ComputeProjectTargets(oldProperties);
-            (bool hasTargetFramework, bool hasTargetFrameworks) newTargets = ComputeProjectTargets(newProperties);
+            IEnumerable<string> oldDimensionsNames = GetDimensionsNames(oldProperties);
+            IEnumerable<string> newDimensionsNames = GetDimensionsNames(newProperties);
```
davidwengier: Consider ordering the dimension names before comparing the sequences. A re-order, particularly if the first framework doesn't change, is presumably not a force-reload-worthy event.

davkean: The order actually matters (its used in the name of the config), and it is currently guaranteed to be [Configuration, Platform, TargetFramework] in that exact order.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/IProjectConfigurationDimensionsProvider4.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+
+using Microsoft.Build.Construction;
+
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    internal interface IProjectConfigurationDimensionsProvider4 : IProjectConfigurationDimensionsProvider3
```
davidwengier: Since its only used by us, should this name be something different, lest CPS decide they need a version 4 of the interface?

davkean: This is my lazy ass not putting this into CPS where it belongs. I'll file a bug.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Reload/ProjectReloadInterceptor.cs
```diff
@@ -1,62 +1,62 @@
...
 
-            return hasTargetFramework != newTargets.hasTargetFramework || hasTargetFrameworks != newTargets.hasTargetFrameworks;
+            // If we have same dimensions, no need to reload
+            if (oldDimensionsNames.SequenceEqual(newDimensionsNames, StringComparers.ConfigurationDimensionNames))
+                return ProjectReloadResult.NoAction;
+
+            // We no longer have same dimensions so we need to reload all configurations by reloading the project.
+            // This catches when we switch from [Configuration, Platform] ->  [Configuration, Platform, TargetFramework] or vice versa,
```
davidwengier: Thank you for the extra comments :D

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Reload/ProjectReloadInterceptor.cs
davidwengier: Thanks for the clarification, I wasn't 100% sure on it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Reload/ProjectReloadInterceptor.cs
```diff
@@ -1,62 +1,62 @@
...
 using System.ComponentModel.Composition;
+using System.Linq;
 
 using Microsoft.Build.Construction;
 
 namespace Microsoft.VisualStudio.ProjectSystem
 {
     /// <summary>
-    /// C#/VB specific project reload interceptor.
+    ///     Reloads a project if the number of configuration dimensions has changed.
```
drewnoakes:
```suggestion
    ///     Reloads a project if its configuration dimensions change.
```

Seems that reload can occur even if the number of dimensions is the same.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Reload/ProjectReloadInterceptor.cs
```diff
@@ -1,62 +1,62 @@
...
-                if (property.Name.Equals(ConfigurationGeneral.TargetFrameworkProperty, StringComparison.OrdinalIgnoreCase))
+                foreach (string name in provider.GetBestGuessDimensionNames(properties))
                 {
-                    targets.hasTargetFramework = true;
-                }
-
-                if (property.Name.Equals(ConfigurationGeneral.TargetFrameworksProperty, StringComparison.OrdinalIgnoreCase))
-                {
-                    targets.hasTargetFrameworks = true;
+                    yield return name;
```
drewnoakes: Could avoid the generator and make this method only:

```c#
return DimensionProviders
    .Select(v => v.Value)
    .OfType<IProjectConfigurationDimensionsProvider4>()
    .SelectMany(provider => provider.GetBesGuessDimensionNames(properties));
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Reload/ProjectReloadInterceptor.cs
davkean: In practice this will never happen, but I've made the changes here as suggested.

davkean:
Added that test case.

Issue #3575, opened 2018-05-25
Make ResX editor deterministic when saving into .resx file


martinsuchan:
**Current behavior:**
The ResX/ResW editor for resource strings has one specific feature that makes development harder than necessary. **By default when adding new strings to the resx file, new strings are appended at the end of the file, instead of sorting in those new strings alphabetically.**  
This causes all kinds of issues when 2+ developers are working in branches and adding new resources strings to the same file. Each branch than contains different new strings at the end of the file, and **merging such files causes conflicts, requiring manual edits, that are quite error prone**.

As a workaround our team uses one hidden feature of the ResX editor that can be used for sorting all strings alhabetically. If I'm in the ResX editor and press Ctrl+A to select all, Ctrl+X to cut all and Ctrl+V to paste all, and save it, it saves all strings sorted alphabetically. It's important to note that this **cutting and pasting in ResX editor is quite slow**. With resource file with about 200 strings, both cut and paste takes several seconds. This could be optimized for sure - the file is only abut 50 kB in size. Ideally this cut and paste dance should not be needed at all!

**Expected behavior:**
The ResX editor should by default sort-in all new strings alphabetically to prevent merge issues when new strings are added in different branches, and without need to use cut&paste hacks for manually sorting the strings.

Another request, the **ResXFileCodeGenerator should not generate this line in the generated .cs file:**
```
Generated: 05/10/2018 12:13:23
```
Adding this string in different branches causes merge errors that needs to be manually resolved. There is no benefit in keeping this line in the generated file.

davkean:
@martinsuchan I investigated this. The behavior is actually the opposite; by default when adding new strings to ResX, new strings will be appended in _ordinal_ order. If someone manually modifies the file and puts a string out of order, it will detect and consider that file "unordered" and new resources will be added to end of the file. The behavior is actually almost identical to how I would have implemented it; it does that to avoid reordering an entire file just when you've removed or added a single resource. I suspect this was a behavior introduced after ResX files were created. 

I could consider a slight change to the algorithm to prevent the situation where you've actually added two resources in different branches but they end up clashing because of the same prefix. Dev merges just these two resources out of order - breaking our "ordinal" check. In this case, maybe we say that if 90% of the file is ordered, then we consider it all ordered.

The relevant code is here: https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceFile.vb#L936.

davkean:
[...]

PR #4864, opened 2019-05-23
Always persist ResX files in ordinal order

davkean:
Fixes: #3575

ResX's files are persisted as "alphabetized" if they were determined to be "alphabetized" when opened. However, the persistence and determination use different comparisons, resulting in a file that was persisted as "alphabetized" to be opened as "non-alphabetized". Make them the same ("ordinal ignore case") .

The determination check was changed to match the persistence format to reduce the churn when editing the ResX in a new version of VS.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceFile.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceFile.vb
@@ -1014,9 +1014,9 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
 
                     Dim comparer As IComparer
                     If _alphabetizedOrder Then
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceFile.vb
```diff
@@ -956,7 +956,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
 
                             ' we check whether the resource item in the original file was alphabetized, we keep the style when we save it...
                             If _alphabetizedOrder Then
-                                If String.Compare(lastName, Resource.Name, StringComparison.Ordinal) > 0 Then
```
davkean: This is the bug, should be OrdinalIgnoreCase.

Issue #4860, opened 2019-05-23
KeyNotFoundException in dependencies node


drewnoakes:
Reported by @dotMorten on Twitter:

https://twitter.com/dotMorten/status/1131241962470760449?s=03

![image](https://user-images.githubusercontent.com/350947/58227234-5c013100-7d6d-11e9-9373-078552c82536.png)

---

I suspect the exception is being thrown here:

https://github.com/dotnet/project-system/blob/d179a81ff071bb53c9bd467f5849af1c7a6ed9d9/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs#L530

dotMorten:
The project I was building was https://github.com/Esri/arcgis-toolkit-dotnet

jamesmontemagno:
Same with https://github.com/xamarin/Essentials in 16.1

davkean:
@drewnoakes I've tentatively put this in 16.1 in case it's a 100% repro. I wonder if this related to some of the rule changes we've made in 16.1.

davkean:
It may be related to culture too, @jamesmontemagno Are you on en-US? or something else?

jamesmontemagno:
of VS? en-US here

davkean:
@jamesmontemagno Are you sure the error is the same for Essentials? I see:
```
System.AggregateException: One or more errors occurred. ---> NuGet.Frameworks.FrameworkException: Invalid framework identifier ''.
   at NuGet.Frameworks.NuGetFramework.GetShortFolderName(IFrameworkNameProvider mappings)
   at NuGet.SolutionRestoreManager.VsSolutionRestoreService.<>c.<ToPackageSpec>b__11_0(TargetFrameworkInformation tfi)
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at NuGet.SolutionRestoreManager.VsSolutionRestoreService.ToPackageSpec(ProjectNames projectNames, IEnumerable TargetFrameworks, String originalTargetFrameworkstr, String msbuildProjectExtensionsPath)
[...]

PR #4874, opened 2019-05-28
Prevent KeyNotFoundException

drewnoakes:
Fixes the symptom of #4860, if not the cause. Unfortunately I was not able to reproduce the exception, so the underlying problem is unclear. This will, at least, prevent the exception.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -410,11 +410,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
         {
             IDependenciesSnapshot snapshot = e.Snapshot;
 
-            if (snapshot == null)
-            {
-                return;
-            }
-
             if (_tasksService.UnloadCancellationToken.IsCancellationRequested || e.Token.IsCancellationRequested)
             {
                 return;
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -410,18 +412,13 @@ private void OnDependenciesSnapshotChanged(object sender, SnapshotChangedEventAr
         {
             IDependenciesSnapshot snapshot = e.Snapshot;
 
-            if (snapshot == null)
-            {
-                return;
-            }
```
drewnoakes: This property cannot be `null`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -527,7 +524,14 @@ public async Task<IRule> GetRuleAsync(IDependency dependency, IProjectCatalogSna
             IImmutableDictionary<string, IPropertyPagesCatalog> namedCatalogs = await GetNamedCatalogsAsync();
             Requires.NotNull(namedCatalogs, nameof(namedCatalogs));
 
-            IPropertyPagesCatalog browseObjectsCatalog = namedCatalogs[PropertyPageContexts.BrowseObject];
+            if (!namedCatalogs.TryGetValue(PropertyPageContexts.BrowseObject, out IPropertyPagesCatalog browseObjectsCatalog))
+            {
+                // Issue https://github.com/dotnet/project-system/issues/4860 suggests this code path
+                // can exist, however a repro was not found to dig deeper into the underlying cause.
+                // For now just return null as the upstream caller handles null correctly anyway.
+                return null;
```
drewnoakes: This is the change that prevents the exception.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -23,6 +23,8 @@
 
 using Task = System.Threading.Tasks.Task;
 
+#nullable enable
```
drewnoakes: I will start introducing nullability annotations throughout the dependency node.

PR #4876, opened 2019-05-28
Sync workspace events

jmarolf:
initial rough attempt at synchronizing workspace and project system during a rename

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 
 using Microsoft.CodeAnalysis;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
@@ -1,170 +1,187 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+using System;
 using System.Collections.Generic;
+using System.ComponentModel.Composition;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 using Microsoft.CodeAnalysis;
+using Microsoft.VisualStudio.LanguageServices;
+using Microsoft.VisualStudio.ProjectSystem.Rename;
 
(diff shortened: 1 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -1,170 +1,187 @@
...
+            _projectVsServices.ThreadingService.RunAndForget(
+                () => HandleRenameAsync(oldFilePath, newFilePath),
+                    unconfiguredProject: _projectVsServices.Project);
+        }
 
-            if (args.Kind == WorkspaceChangeKind.DocumentChanged && args.ProjectId == _project.Id)
+        internal async Task HandleRenameAsync(string oldFilePath, string newFilePath)
+        {
+            // Do not offer to rename types if the user changes the file extensions
+            if (!oldFilePath.EndsWith(Path.GetExtension(newFilePath), StringComparison.OrdinalIgnoreCase))
```
davidwengier: Thinking out loud here, but given that this only triggers of a rename of a `<Compile />` item, do we care about file extensions? There probably isn't a scenario where it's relevant, but in theory if something can be renamed and is still considered a compile input after the rename, then we might want to process it...

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/CompileItemHandler.cs
```diff
@@ -57,6 +61,23 @@ public void Handle(IComparable version, BuildOptions added, BuildOptions removed
...
 
+        public void HandleProjectUpdate(IComparable version, IProjectChangeDescription projectChange, bool isActiveContext, IProjectLogger logger)
+        {
+            Requires.NotNull(version, nameof(version));
+            Requires.NotNull(projectChange, nameof(projectChange));
+            Requires.NotNull(logger, nameof(logger));
+
+            VerifyInitialized();
+
+            if (projectChange.Difference.RenamedItems.Any())
```
davidwengier: Is the `.Any()` call helping here at all?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
jmarolf: Since compile is computed automatically in sdk style projects this is less of an issue, but if users have set `EnableDefaultCompileItems` to false we would still want this behavior of not offering a rename if we change the file from `.cs` to `.vb`   

---
In reply to: [288347951](https://github.com/dotnet/project-system/pull/4876#discussion_r288347951) [](ancestors = 288347951)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/CompileItemHandler.cs
jmarolf: I'll remove this

---
In reply to: [288348164](https://github.com/dotnet/project-system/pull/4876#discussion_r288348164) [](ancestors = 288348164)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
jmarolf: I believe the assumption is that if the user is renaming the file type then our language service can't be right because the language that the file is presenting as is wrong / needs to be changed.

PR #4882, opened 2019-05-29
Annotate nullability in dependencies node

drewnoakes:
Applies C# 8 nullable reference type annotations to code in the dependencies node.

Adds well-known attributes to the `System.Runtime.CompilerServices` that support the feature.

This was mostly a dogfooding exercise, which uncovered a few Roslyn bugs:

- https://github.com/dotnet/roslyn/issues/36014
- https://github.com/dotnet/roslyn/issues/36015
- https://github.com/dotnet/roslyn/issues/36016
- https://github.com/dotnet/roslyn/issues/36018

This is a draft PR as there remains one bug to file for which I have not yet found a minimal repro case, nor a workaround (beyond nowarn or 21 `#pragma` pairs).


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContext.cs
```diff
@@ -50,9 +52,10 @@ public ITargetFramework GetProjectFramework(ProjectConfiguration projectConfigur
             }
         }
 
-        public ConfiguredProject GetInnerConfiguredProject(ITargetFramework target)
+        public ConfiguredProject? GetInnerConfiguredProject(ITargetFramework target)
         {
-            return _configuredProjectByTargetFramework.FirstOrDefault((x, t) => t.Equals(x.Key), target).Value;
+            // TODO remove ! when https://github.com/dotnet/roslyn/issues/36018 is fixed
+            return _configuredProjectByTargetFramework.FirstOrDefault((x, t) => t.Equals(x.Key!), target).Value;
```
drewnoakes: This is an occurrence of dotnet/roslyn#36018.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/CrossTarget/TargetFramework.cs
```diff
@@ -56,22 +58,15 @@ public TargetFramework(string moniker)
         /// </summary>
         public override bool Equals(object obj)
         {
-            if (obj == null)
-            {
-                return false;
-            }
-
-            if (obj is ITargetFramework targetFramework)
+            return obj switch
```
drewnoakes: Replaced chained `if`s with a switch expression.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesTreeViewProvider.cs
```diff
@@ -245,16 +247,16 @@ public IProjectTree FindByPath(IProjectTree root, string path)
...
                 : null;
 
             foreach (IDependency dependency in dependencies)
             {
-                IProjectTree dependencyNode = rootNode.FindChildWithCaption(dependency.Caption);
+                IProjectTree? dependencyNode = rootNode.FindChildWithCaption(dependency.Caption);
                 bool isNewDependencyNode = dependencyNode == null;
 
-                if (!isNewDependencyNode
+                if (dependencyNode != null
```
drewnoakes: Minor workaround of limitations in the static flow analysis.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesTreeViewProvider.cs
```diff
@@ -277,7 +279,7 @@ public IProjectTree FindByPath(IProjectTree root, string path)
                     : dependencyNode.Parent;
             }
 
-            return shouldCleanup
+            return currentNodes != null // shouldCleanup
```
drewnoakes: Ditto (minor workaround of limitations in the static flow analysis).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesTreeViewProvider.cs
```diff
@@ -383,7 +387,7 @@ IProjectTree UpdateTreeNode()
                 var updatedNodeParentContext = new ProjectTreeCustomizablePropertyContext
                 {
                     ExistsOnDisk = false,
-                    ParentNodeFlags = node.Parent?.Flags ?? default
+                    ParentNodeFlags = node!.Parent?.Flags ?? default
```
drewnoakes: Minor workaround of limitations in the static flow analysis. It does not deduce that this function is only called when `node` is not null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -113,7 +115,7 @@ public async Task ObserveTreeUpdateCompletedAsync(bool hasUnresolvedDependency)
 
             if (_projectId == null)
             {
-                await InitializeProjectIdAsync();
+                _projectId = await GetProjectIdAsync();
```
drewnoakes: Reworked this to return and assign here to satisfy the analysis.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -137,26 +139,21 @@ public async Task ObserveTreeUpdateCompletedAsync(bool hasUnresolvedDependency)
...
                 else
                 {
-                    SetProjectId(_telemetryService.HashValue(_project.FullPath));
+                    return _telemetryService.HashValue(_project.FullPath);
                 }
             }
         }
 
-        // helper to support testing
-        internal void SetProjectId(string projectId)
```
drewnoakes: Removed this unused helper.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/Actions/SearchGraphActionHandler.cs
```diff
@@ -83,21 +85,19 @@ private void Search(IGraphContext graphContext)
...
-                        IDependenciesGraphViewProvider viewProvider = FindViewProvider(topLevelDependency);
+                        IDependenciesGraphViewProvider? viewProvider = FindViewProvider(topLevelDependency);
 
                         if (viewProvider == null)
                         {
                             continue;
                         }
 
-                        bool processed = viewProvider.MatchSearchResults(
+                        if (!viewProvider.MatchSearchResults(
```
drewnoakes: Inlining this temporary worked around dotnet/roslyn#36016

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphChangeTracker.cs
```diff
@@ -150,7 +152,7 @@ bool HandleChanges(IGraphContext graphContext)
                         nodeProjectPath,
                         updatedDependency,
                         inputGraphNode,
-                        updatedSnapshot.Targets[updatedDependency.TargetFramework]))
+                        updatedSnapshot!.Targets[updatedDependency.TargetFramework]))
```
drewnoakes: Asking for forgiveness with `!` here to work around https://github.com/dotnet/roslyn/issues/36014

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphChangeTracker.cs
```diff
@@ -214,7 +216,7 @@ internal Enumerator(LinkedList<WeakReference<T>> list)
                 {
                     _list = list;
                     _next = list.First;
-                    Current = null;
+                    Current = null!;
```
drewnoakes: This was slightly interesting. In normal usage, the enumerator returns only non-null values. However before `MoveNext` is called, and after it returns `false`, `Current` is invalid and therefore null. It'd be technically more correct to make `Current` nullable, however all enumerations would then have to check every value for `null` which feels overkill given the enumerator works perfectly well with `foreach` as is.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -142,18 +144,18 @@ public void BeginGetGraphData(IGraphContext context)
 
             Assumes.True(IsInitialized);
 
-            GraphNodeId newNodeId = GetTopLevelGraphNodeId(projectPath, viewModel.OriginalModel.GetTopLevelId());
+            GraphNodeId newNodeId = GetTopLevelGraphNodeId(projectPath, viewModel.OriginalModel!.GetTopLevelId());
```
drewnoakes: Asking for forgiveness with `!` here because `Requires.NotNull` is not annotated with `System.Runtime.CompilerServices.EnsuresNotNullAttribute`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/IDependenciesGraphViewProvider.cs
```diff
@@ -57,6 +60,6 @@ internal interface IDependenciesGraphViewProvider
         bool MatchSearchResults(
             IDependency topLevelDependency,
             Dictionary<string, HashSet<IDependency>> searchResultsPerContext,
-            out HashSet<IDependency> topLevelDependencyMatches);
+            [NotNullWhenTrue] out HashSet<IDependency>? topLevelDependencyMatches);
```
drewnoakes: Using an attribute to model the try-get pattern.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/PackageAnalyzerAssemblyDependencyModel.cs
```diff
@@ -54,6 +56,10 @@ internal class PackageAnalyzerAssemblyDependencyModel : DependencyModel
             {
                 DependencyIDs = ImmutableArray.CreateRange(dependenciesIDs);
             }
+            else
+            {
+                DependencyIDs = ImmutableList<string>.Empty;
```
drewnoakes: I decided to ensure `DependencyIDs` was always non-null. This is probably the closest thing to a potential bug fix from this exercise.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -273,8 +275,9 @@ private void OnSubtreeProviderDependenciesChanged(object sender, DependenciesCha
                 return;
             }
 
+            // TODO remove ! when https://github.com/dotnet/roslyn/issues/36018 is fixed
             ITargetFramework targetFramework =
-                string.IsNullOrEmpty(e.TargetShortOrFullName) || TargetFramework.Any.Equals(e.TargetShortOrFullName)
+                string.IsNullOrEmpty(e.TargetShortOrFullName) || TargetFramework.Any.Equals(e.TargetShortOrFullName!)
```
drewnoakes: Another workaround for https://github.com/dotnet/roslyn/issues/36018

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -470,9 +473,9 @@ async Task<AggregateCrossTargetProjectContext> TryUpdateCurrentAggregateProjectC
 
                 foreach (string targetFrameworkMoniker in targetFrameworkMonikers)
                 {
-                    ITargetFramework targetFramework = _targetFrameworkProvider.GetTargetFramework(targetFrameworkMoniker);
+                    ITargetFramework? targetFramework = _targetFrameworkProvider.GetTargetFramework(targetFrameworkMoniker);
 
-                    if (!previousContext.TargetFrameworks.Contains(targetFramework))
+                    if (targetFramework == null || !previousContext.TargetFrameworks.Contains(targetFramework))
```
drewnoakes: Working around a complaint that `Contains` didn't like a nullable argument.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Composition/OrderPrecedenceImportCollectionExtensions.cs
```diff
@@ -56,7 +58,7 @@ internal static IEnumerable<T> ExtensionValues<T>(this OrderPrecedenceImportColl
             }
         }
 
-        public static T FirstOrDefaultValue<T>(this OrderPrecedenceImportCollection<T> imports, Func<T, bool> predicate)
+        public static T? FirstOrDefaultValue<T>(this OrderPrecedenceImportCollection<T> imports, Func<T, bool> predicate) where T : class
```
drewnoakes: Had to add a `class` type constraint here in order to return `default`. This is fine for our current usages but feels a little unnecessarily restrictive.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Composition/OrderPrecedenceImportCollectionExtensions.cs
```diff
@@ -72,7 +74,7 @@ public static T FirstOrDefaultValue<T>(this OrderPrecedenceImportCollection<T> i
             return default;
         }
 
-        public static TImport FirstOrDefaultValue<TImport, TArg>(this OrderPrecedenceImportCollection<TImport> imports, Func<TImport, TArg, bool> predicate, TArg arg)
+        public static TImport? FirstOrDefaultValue<TImport, TArg>(this OrderPrecedenceImportCollection<TImport> imports, Func<TImport, TArg, bool> predicate, TArg arg) where TImport : class
```
drewnoakes: Ditto (type constraint).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphChangeTracker.cs
davidwengier: Might be worth a comment in the code here indicating that this is okay because of the early return from `updatedDependency == null`, as that wouldn't have been apparent to me if I hadn't just read that roslyn issue (plus it doesn't sound like this is a temporary workaround)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/DependenciesGraphProvider.cs
```diff
@@ -142,18 +144,18 @@ public void BeginGetGraphData(IGraphContext context)
...
         private GraphNode DoAddGraphNode(
             GraphNodeId graphNodeId,
             IGraphContext graphContext,
             string projectPath,
-            GraphNode parentNode,
+            GraphNode? parentNode,
             IDependencyViewModel viewModel)
         {
-            _iconCache.Register(viewModel.Icon);
+            _iconCache!.Register(viewModel.Icon);
```
davidwengier: I think for this pattern where the analysis doesn't know about `InitializeCoreAsync`, it might be nicer to define the field as `GraphIconCache _iconCache = null!;` rather than having to annotate every call site (or the first call site in every method)

drewnoakes: I debated this. In this class only one member uses the field, but I agree that in the general case this may be better.

davidwengier: Yeah, this case was fine, but since you're (one of?) the first to annotate things we should probably establish some patterns.

drewnoakes: In the main I think it's a great addition to the language. Applying it was pretty fast and painless, with only a lessons learned, pointed to in my self-review comments. We definitely become dependent upon upstream types being annotated (corefx, CPS and other VS packages, ...).

PR #4887, opened 2019-05-30
Rename code-behind files so they are nested

davkean:
Rename the code-behind files so that they are nested under their rule file in VS.

When I remove the DependentUpon on broke this nesting relationship.


davidwengier:
It's been so long since I've spent any time in the Project System solution I didn't even notice :P

davidwengier:
Wait... did you revert that change? These are still referenced

https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj#L76

davkean:
Yep, you are right.

PR #4892, opened 2019-06-04
Tidying

drewnoakes:
Smal bits from larger work on the dep node.


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -6,8 +6,6 @@
 using System.ComponentModel.Composition;
 using System.Linq;
 using System.Threading.Tasks;
-
```
davidwengier: This would seem to violate our .editorconfig `dotnet_separate_import_directive_groups` rule :P

Issue #4540, opened 2019-02-05
Remove xproj migrate


davkean:
`dotnet migrate` will be removed in the 3.0 timeframe, we should make a counter change where we prevent ourselves from calling in the xproj project factory. 

- [ ] We should have our implementation of `IVsProjectUpgradeViaFactory4.UpgradeProject_CheckOnly` in `MigrateXprojProjectFactory` return `VSPUVF_PROJECT_DEPRECATED` to indicate that these project types are not supported. 
- [ ] Delete migration code
- [ ] Remove https://github.com/dotnet/project-system/blob/master/THIRD-PARTY-NOTICES.md because we've deleted src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\Xproj\MigrationError.cs
- [ ] Doc results at https://docs.microsoft.com/en-us/visualstudio/porting/port-migrate-and-upgrade-visual-studio-projects?view=vs-2017

davkean:
Run this through compat owner, blocked until approval. Also should doc this here: https://docs.microsoft.com/en-us/visualstudio/porting/port-migrate-and-upgrade-visual-studio-projects?view=vs-2017.

drewnoakes:
PR #4895 is ready to merge once `master` is targeting 16.3.

drewnoakes:
PR https://github.com/MicrosoftDocs/visualstudio-docs/pull/3379 is open for the documentation update.

drewnoakes:
PR https://github.com/dotnet/docs/pull/12785 created to update documentation about migration tooling.

PR #4895, opened 2019-06-05
Deprecate xproj migration

drewnoakes:
Fixes some parts of #4540.

With this change, when a user loads a legacy `.xproj` project they will be notified that the project is unsupported. The project will appear in Solution Explorer unloaded with "incompatible" alongside the name.

Tested on drewnoakes/metadata-extractor-dotnet@2be0b57d9cd06b570b9f923df2774e8a904ca747

image

image

---

Marked as draft to prevent merging until we have alignment with CLI release.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Xproj/MigrateXprojProjectFactory.cs
```diff
@@ -1,360 +1,70 @@
...
 using Microsoft.VisualStudio.Shell.Flavor;
 using Microsoft.VisualStudio.Shell.Interop;
 
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
-
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Xproj
 {
     [Guid(ProjectType.LegacyXProj)]
     internal sealed class MigrateXprojProjectFactory : FlavoredProjectFactoryBase, IVsProjectUpgradeViaFactory, IVsProjectUpgradeViaFactory4
```
davkean: This is no longer "migrate"

davkean:
I expected this to also contain deletions of:

- MigrationError.cs
- MigrationReport.cs
- THIRD-PARTY-NOTICES.md

davkean:
I think you'll want to also remove the unused strings that this code no longer uses.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Xproj/MigrateXprojProjectFactory.cs
drewnoakes: I pondered this. Will rename to `XprojProjectFactory`.

davkean: case it "XProj" can ya, that case bothers me every time.

drewnoakes:
Thanks for the review @davkean. Updated.

I found a bunch more unused resource strings and fixed them in #4896.

drewnoakes:
Keeping it in draft until master is 16.3 to prevent accidental merging.

drewnoakes:
Remaining task is to update docs https://docs.microsoft.com/en-us/visualstudio/porting/port-migrate-and-upgrade-visual-studio-projects?view=vs-2017, which we need to coordinate the timing of.

davkean:
You could probably que up a PR for that too if you want to have it just ready.

drewnoakes:
Code updated to:
- Write an error to the upgrade log linking to documentation that explains how to manually migrate from xproj to csproj (@livarcocc to review and improve text)
- Remove implementation of `IVsProjectUpgradeViaFactory`, which simplifies code a fair bit
- Remove check for `xproj` file extension as this project factory should only be called for xproj files, further simplifying code

PR #4896, opened 2019-06-05
Remove unused resource strings

drewnoakes:
Removes 8 unused resource strings from `Resources.resx` and 18 from `VSResources.resx`.

Note: there is no overlap with resources removed as part of #4895.


drewnoakes:
On @tmeschter's suggestion I grepped the codebase for string occurrences of these resource names and none turned up.

PR #4900, opened 2019-06-07
Resource Editor fails to add non-string resources in .NETCore projects

### LINKED ISSUE #4806: Resource Editor fails to add non-string resources in .NETCore projects
**Visual Studio Version**:
16.0.2

**Summary:**
Resource editor is trying to activate .NETCore types in desktop process, resulting in a failure to load the assembly which doesn't exist in desktop.
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/692319

Given the resources editor only supports a small set of types that are in the framework, I don't think we need to remote this part of the designer into a .NETCore process.  Instead we could have the editor read TypeForwardedFrom on these types, then activate that identity which will be the desktop identity.  We can ensure that for the set of types the resource editor uses remain serializeable on .NETCore and support round-trip serialization with desktop (since this is already needed to support .NET Framework assembly reuse in Core).  If my assumption about "small set of types" is not correct we may need to revisit this direction.

**Steps to Reproduce**:

1. create a new .netcore 3.0 winforms project

2. open resources

3. try to add image

**Expected Behavior**:
image successfully added

**Actual Behavior**:
Error adding the image
image

**User Impact**:
Cannot add non-string resources to resx in .NETCore3.0 projects.

This is dependent on the work in MSBuild to support handling the non-string resources in core-MSBuild.  /cc @rainersigwald @DustinCampbell @dreddy-work @nguerrera @merriemcgaw

dmonroym:
Fix for https://github.com/dotnet/project-system/issues/4806


src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1630,11 +1630,11 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
         ''' file are only used to produce the .designer.cs file, which is still built against
         ''' netstandard.dll.
         ''' </remarks>
-        Private Shared Function AdjustSystemStringAssemblyQualifiedName(AssemblyQualifiedName As String) As String
-            Static SystemStringAssemblyQualifiedName As String
+        Private Shared Function AdjustAssemblyQualifiedName(AssemblyQualifiedName As String) As String
+            Static CorrectedAssemblyQualifiedName As String
 
             ' If this type definitely isn't System.String then bail out before we allocate
```
dmonroym: Need to fix these comments

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1643,11 +1643,27 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
...
+                    Return CorrectedAssemblyQualifiedName
+                End If
+                If typeName = "System.Drawing.Icon" Then
+
+                    If CorrectedAssemblyQualifiedName Is Nothing Then
+                        CorrectedAssemblyQualifiedName = GetType(Drawing.Icon).AssemblyQualifiedName
+                    End If
+
+                    Return CorrectedAssemblyQualifiedName
                 End If
```
dmonroym: I believe I'm also missing other types like audio so I need to test for that

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1630,11 +1630,11 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
         ''' file are only used to produce the .designer.cs file, which is still built against
         ''' netstandard.dll.
         ''' </remarks>
-        Private Shared Function AdjustSystemStringAssemblyQualifiedName(AssemblyQualifiedName As String) As String
-            Static SystemStringAssemblyQualifiedName As String
+        Private Shared Function AdjustAssemblyQualifiedName(AssemblyQualifiedName As String) As String
+            Static CorrectedAssemblyQualifiedName As String
```
davidwengier: This doesn't make sense to be static any more, since this method now can return multiple names. In fact, the way the code is now, the method would always return the result of the first call to it.

Perhaps a Dictionary<string, Lazy<string>> might be better to use?

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1630,11 +1630,11 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
...
 
             ' If this type definitely isn't System.String then bail out before we allocate
-            If Not AssemblyQualifiedName.StartsWith("System.String") Then
+            If Not (AssemblyQualifiedName.StartsWith("System.String") Or AssemblyQualifiedName.StartsWith("System.Drawing.Bitmap") Or AssemblyQualifiedName.StartsWith("System.Drawing.Icon")) Then
```
davkean: Agree with Dave on this - If you this a map from old type name -> new type name this effectively just becomes a single lookup, and you don't need to keep this in sync with the if statements below.

drewnoakes: Consider ditching `StartsWith`. However unlikely, it does the wrong thing for a `System.String.Foo` type.

Probably easiest to parse out `typeName` first, then look it up in a dictionary as mentioned above.

Try to avoid a code path that doesn't explicitly return anything. If the input string was `"Foo"`, this method would currently return `Nothing`. It should probably return `AssemblyQualifiedName` in cases where no change is needed.

ericstj:
I see that this only handles String, Bitmap, and Icon.  Are those the only types that currently get special handling by the resource editor?

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1630,24 +1632,28 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
...
         ''' </remarks>
-        Private Shared Function AdjustSystemStringAssemblyQualifiedName(AssemblyQualifiedName As String) As String
-            Static SystemStringAssemblyQualifiedName As String
-
-            ' If this type definitely isn't System.String then bail out before we allocate
-            If Not AssemblyQualifiedName.StartsWith("System.String") Then
-                Return AssemblyQualifiedName
-            End If
+        Private Shared Function AdjustAssemblyQualifiedName(AssemblyQualifiedName As String) As String
+            Dim CorrectedAssemblyQualifiedName As String = Nothing
```
davkean: Variables are `camelCased`.

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1630,24 +1632,28 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
...
             If indexOfFirstComma <> -1 Then
                 Dim typeName = AssemblyQualifiedName.Substring(startIndex:=0, length:=indexOfFirstComma)
-                If typeName = "System.String" Then
-
-                    If SystemStringAssemblyQualifiedName Is Nothing Then
-                        SystemStringAssemblyQualifiedName = GetType(String).AssemblyQualifiedName
+                If s_correctedAssemblyQualifiedName.TryGetValue(typeName, CorrectedAssemblyQualifiedName) Then
+                    Return CorrectedAssemblyQualifiedName
+                Else
+                    If typeName = "System.String" Then
```
davkean: To simplify this even further, just initialize the dictionary already populated with all the types and you can remove the if statements.

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1630,24 +1636,14 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
...
         ''' </remarks>
-        Private Shared Function AdjustSystemStringAssemblyQualifiedName(AssemblyQualifiedName As String) As String
-            Static SystemStringAssemblyQualifiedName As String
-
-            ' If this type definitely isn't System.String then bail out before we allocate
-            If Not AssemblyQualifiedName.StartsWith("System.String") Then
-                Return AssemblyQualifiedName
-            End If
+        Private Shared Function AdjustAssemblyQualifiedName(AssemblyQualifiedName As String) As String
+            Dim correctedAssemblyQualifiedName As String = Nothing
```
drewnoakes: nit: move this to inner scope, just before it's used

src/Microsoft.VisualStudio.Editors/GlobalSuppressions.vb
```diff
@@ -421,7 +421,7 @@
...
 <Assembly: CodeAnalysis.SuppressMessage("Style", "IDE1006:Naming Styles", Scope:="member", Target:="~M:Microsoft.VisualStudio.Editors.ResourceEditor.Resource.#ctor(System.Runtime.Serialization.SerializationInfo,System.Runtime.Serialization.StreamingContext)")>
 <Assembly: CodeAnalysis.SuppressMessage("Style", "IDE1006:Naming Styles", Scope:="member", Target:="~M:Microsoft.VisualStudio.Editors.ResourceEditor.Resource.AddFileWatcherEntry(Microsoft.VisualStudio.Editors.ResourceEditor.FileWatcher)")>
-<Assembly: CodeAnalysis.SuppressMessage("Style", "IDE1006:Naming Styles", Scope:="member", Target:="~M:Microsoft.VisualStudio.Editors.ResourceEditor.Resource.AdjustSystemStringAssemblyQualifiedName(System.String)~System.String")>
+<Assembly: CodeAnalysis.SuppressMessage("Style", "IDE1006:Naming Styles", Scope:="member", Target:="~M:Microsoft.VisualStudio.Editors.ResourceEditor.Resource.AdjustAssemblyQualifiedName(System.String)~System.String")>
```
drewnoakes: Is this because the argument `AssemblyQualifiedName` starts with a capital letter? If so, consider renaming to `assemblyQualifiedName` and removing this suppression.

dmonroym: Current naming is consistent with the rest of the file. Is it worth changing this one function's style to remove the suppression?

drewnoakes: No, probably not.

dmonroym:
Audio is also specially handled so I just added that. Files are already being handled correctly

---
In reply to: [500466929](https://github.com/dotnet/project-system/pull/4900#issuecomment-500466929) [](ancestors = 500466929)

davkean:
Can we merge this in tomorrow @dmonroym?

dmonroym:
Was only waiting for checks and final reviews. Merging now

---
In reply to: [505330083](https://github.com/dotnet/project-system/pull/4900#issuecomment-505330083) [](ancestors = 505330083)

PR #4908, opened 2019-06-11
Remove old design-time build logging info

panopticoncentral:
Visual Studio 2017 version 15.9 is the only currently supported version of VS 2017 that the tools run on, so we can remove the old instructions for earlier point releases.

### DIFF
--- a/docs/design-time-builds.md
+++ b/docs/design-time-builds.md
@@ -87,30 +87,7 @@ You can force Visual Studio to show the results of a design-time build using the
 5. Open the solution
 6. Under %TEMP%, look for [RANDOMGUID].designtime.log files, these will contain the results of the design-time build. If running Visual Studio 2015 Update 2 or higher, the name of the project and design-time target that is being called will also be included in the file name.
 
-#### Visual Studio 2017 (version 15.3 or earlier)
-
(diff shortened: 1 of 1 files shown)


docs/design-time-builds.md
```diff
@@ -87,30 +87,7 @@ You can force Visual Studio to show the results of a design-time build using the
...
 6. Under %TEMP%, look for [RANDOMGUID].designtime.log files, these will contain the results of the design-time build. If running Visual Studio 2015 Update 2 or higher, the name of the project and design-time target that is being called will also be included in the file name.
 
-#### Visual Studio 2017 (version 15.3 or earlier)
-
-In Visual Studio 2017 there are two C# and Visual Basic project systems. By default the majority of projects (.NET Framework, UWP, etc) continue to open in the same project system as previous versions of Visual Studio, and hence the instructions are the same as above. 
-
-The new .NET Core and .NET Standard project types, however, open in the new [project system](http://github.com/dotnet/project-system) and you can use the following steps to see the results of design-time builds:
-
-1. Close all instances of Visual Studio
-2. Download the following script: [DesignTimeBuildOutputPane.cmd](/docs/repo/content/DesignTimeBuildOutputPane.cmd?raw=true)
```
drewnoakes: Do we also want to remove this file?

davkean: I still use it.

Issue #3823, opened 2018-08-02
Port WinFormEditorProvider from CPS to here


davkean:
This adds support for opening a WinForms designer from a source file:

https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/134764?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.VS.Implementation%2FEditor%2FWinFormEditorProvider.cs&discussionId=1355860&_a=files

davkean:
Should put this behind a new capability describing WinForms.

wjk:
@davkean Since the milestone is 16.2, does this mean that the changes required to support the WinForms designer in a Core project will ship in the VS 16.2 update? If so, I can't wait.

Also: Are there any issues (related to WinForms designer support or otherwise) I could help out and create a PR for? Thanks!

davkean:
WinForms for .NET Framework scenarios should quasi work (sans a few bugs) in 16.1. 16.2 we fixed a few more bugs, but there's nothing really to do. This bug is just a cleanup bug for layering purposes.

WinForms for .NET Core is very far away and cannot give you a timeframe, the majority of work for this is in closed source so nothing really to help. Our help wanted issues are tracked via https://github.com/dotnet/project-system/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22.

davkean:
This is merged in, I'll leave it open to track the deletion of the CPS code.

davkean:
This is now done.

Issue #4804, opened 2019-05-08
Consider turning off component designer by default


davkean:

davkean:
This is done, in the new project system component designer does not open by default.

PR #4913, opened 2019-06-13
Enable WinForms designer to open

davkean:
Enable WinForms designer to open

Fixes: #3823, #4804

This ports WinFormEditorProvider from CPS, and changes behavior from it and legacy:

- Users can now control (using Open With... -> Set As Default) whether the designer opens by default for Forms, User Controls and Components separately from each other. Legacy/CPS have a single setting that controls this. This is helpful because the Component designer is less useful than the Form designer, so you might want the latter to open but not the former.

- This new setting roams and is shared across all languages. Legacy/CPS is per language per VS install.

- Component Designer is turned off by default. Users can open it via the View Designer command, or use above Open With to set it to open by default.

- The name of the designer editor has changed from "CSharp/Visual Basic Form Editor" for Forms, User Controls and Components in legacy to:

Type|Name
---|---
Form|Form (Windows Forms) Designer 
Component|Component (Windows Forms) Designer
UserControl|User Control (Windows Forms) Designer

I feel these names are better descriptions of what they open and are consistent with the "XAML Designer".

This is a draft PR because we're blocked on CPS producing a package with an updated `IProjectSpecificEditorProvider` implementation. tag @adrianvmsft.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
@@ -16,7 +16,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
     {
         private const string FastUpToDateEnabledSettingKey = @"ManagedProjectSystem\FastUpToDateCheckEnabled";
         private const string FastUpToDateLogLevelSettingKey = @"ManagedProjectSystem\FastUpToDateLogLevel";
-
+        
         private readonly IEnvironmentHelper _environment;
         private readonly IVsUIService<ISettingsManager> _settingsManager;
         private readonly JoinableTaskContext _joinableTaskContext;
@@ -32,10 +32,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         public bool IsProjectOutputPaneEnabled
         {
-            get
-            {
-                return IsEnabled("PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED", ref _isProjectOutputPaneEnabled);
-            }
+            get { return IsEnvironmentVariableEnabled("PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED", ref _isProjectOutputPaneEnabled); }
         }
 
         public async Task<bool> GetIsFastUpToDateCheckEnabledAsync(CancellationToken cancellationToken = default)
@@ -47,10 +44,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         public async Task<LogLevel> GetFastUpToDateLoggingLevelAsync(CancellationToken cancellationToken = default)
         {
             await _joinableTaskContext.Factory.SwitchToMainThreadAsync(cancellationToken);
+
             return _settingsManager.Value?.GetValueOrDefault(FastUpToDateLogLevelSettingKey, LogLevel.None) ?? LogLevel.None;
         }
 
-        private bool IsEnabled(string variable, ref bool? result)
+        private bool IsEnvironmentVariableEnabled(string variable, ref bool? result)
         {
             if (result == null)
             {
@@ -61,5 +59,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
             return result.Value;
         }
+
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
@@ -16,7 +16,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
     {
         private const string FastUpToDateEnabledSettingKey = @"ManagedProjectSystem\FastUpToDateCheckEnabled";
         private const string FastUpToDateLogLevelSettingKey = @"ManagedProjectSystem\FastUpToDateLogLevel";
-        
+        private const string UseDesignerByDefaultSettingKey = @"ManagedProjectSystem\UseDesignerByDefault";
+
         private readonly IEnvironmentHelper _environment;
         private readonly IVsUIService<ISettingsManager> _settingsManager;
         private readonly JoinableTaskContext _joinableTaskContext;
@@ -35,17 +36,38 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             get { return IsEnvironmentVariableEnabled("PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED", ref _isProjectOutputPaneEnabled); }
         }
 
-        public async Task<bool> GetIsFastUpToDateCheckEnabledAsync(CancellationToken cancellationToken = default)
+        public Task<bool> GetIsFastUpToDateCheckEnabledAsync(CancellationToken cancellationToken = default)
         {
-            await _joinableTaskContext.Factory.SwitchToMainThreadAsync(cancellationToken);
-            return _settingsManager.Value?.GetValueOrDefault(FastUpToDateEnabledSettingKey, true) ?? true;
+            return GetSettingValueOrDefault(FastUpToDateEnabledSettingKey, true, cancellationToken);
         }
 
-        public async Task<LogLevel> GetFastUpToDateLoggingLevelAsync(CancellationToken cancellationToken = default)
+        public Task<LogLevel> GetFastUpToDateLoggingLevelAsync(CancellationToken cancellationToken = default)
+        {
+            return GetSettingValueOrDefault(FastUpToDateLogLevelSettingKey, LogLevel.None, cancellationToken);
+        }
+
+        public Task<bool> GetUseDesignerByDefaultAsync(string designerCategory, bool defaultValue, CancellationToken cancellationToken = default)
+        {
+            return GetSettingValueOrDefault(UseDesignerByDefaultSettingKey + "\\" + designerCategory, defaultValue, cancellationToken);
+        }
+
+        public Task SetUseDesignerByDefaultAsync(string designerCategory, bool value, CancellationToken cancellationToken = default)
+        {
+            return SetSettingValueAsync(UseDesignerByDefaultSettingKey + "\\" + designerCategory, value, cancellationToken);
+        }
+
+        private async Task<T> GetSettingValueOrDefault<T>(string name, T defaultValue, CancellationToken cancellationToken)
         {
             await _joinableTaskContext.Factory.SwitchToMainThreadAsync(cancellationToken);
 
-            return _settingsManager.Value?.GetValueOrDefault(FastUpToDateLogLevelSettingKey, LogLevel.None) ?? LogLevel.None;
+            return _settingsManager.Value.GetValueOrDefault(name, defaultValue);
+        }
+
+        private async Task SetSettingValueAsync(string name, object value, CancellationToken cancellationToken)
+        {
+            await _joinableTaskContext.Factory.SwitchToMainThreadAsync(cancellationToken);
+
+            await _settingsManager.Value.SetValueAsync(name, value, isMachineLocal:false);
         }
 
         private bool IsEnvironmentVariableEnabled(string variable, ref bool? result)
@@ -59,6 +81,5 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
             return result.Value;
         }
-
     }
 }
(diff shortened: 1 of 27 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsEditorProvider.EditorInfo.cs
```diff
@@ -0,0 +1,33 @@
+﻿/********************************************************
```
davkean: Will fix this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsEditorProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// <summary>
+    ///     A project-specific editor provider that is responsible for handling two things;
+    ///     
+    ///     1) Add the Windows Forms designer the list of editor factories for a "designable" source file, and 
+    ///        determines whether it opens by default.
+    ///     
+    ///     2) Persists whether the designer opens by default when the user uses Open With -> Set As Default.
+    /// </summary>
+    [Export(typeof(IProjectSpecificEditorProvider))]
+    [AppliesTo(ProjectCapability.DotNet)]
```
davkean: Unsure yet what capability to tie this lightup to.

davkean:
tag @Tanya-Solyanik @lifengl @DustinCampbell

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
```diff
@@ -32,25 +33,44 @@ public ProjectSystemOptions(IEnvironmentHelper environment, IVsUIService<SVsSett
...
+        {
+            return SetSettingValue(UseDesignerByDefaultSettingKey + "\\" + designerCategory, value, cancellationToken);
+        }
+
+        private async Task<T> GetSettingValueOrDefault<T>(string name, T defaultValue, CancellationToken cancellationToken)
         {
             await _joinableTaskContext.Factory.SwitchToMainThreadAsync(cancellationToken);
-            return _settingsManager.Value?.GetValueOrDefault(FastUpToDateEnabledSettingKey, true) ?? true;
+
+            return _settingsManager.Value.GetValueOrDefault(name, defaultValue);
```
drewnoakes: Is there no way `_settingsManager.Value` can be `null`? (Also below.)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsEditorProvider.EditorInfo.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms
```
drewnoakes: Consider `#nullable enable` to all new files.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IProjectSystemOptions.cs
```diff
@@ -44,5 +44,15 @@ bool IsProjectOutputPaneEnabled
         ///     The level of fast up to date check logging.
         /// </value>
         Task<LogLevel> GetFastUpToDateLoggingLevelAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        ///     Gets a value indicating whether the designer is the default editor for the specified designer category.
+        /// </summary>
+        Task<bool> GetUseDesignerByDefault(string designerCategory, bool defaultValue, CancellationToken cancellationToken = default);
```
drewnoakes: `Async` suffix on this and below (mirroring implementation).

merriemcgaw:
tagging @dreddy-work and @Shyam-Gupta to take a look too.

Shyam-Gupta:
@davkean Will there be a project system tools option that user can enable to have WinForms designer open by default when user selects a component file ?

Tanya-Solyanik:
@davkean  - do I understand correctly that these changes go into project system that runs core applications only? I'm concerned about the effect of this change on the classic .NET framework applications. If the user disables the default designer for the Core Form (this project is opened by CPS), what would happen when the same user double-clicks on a .NET classic Framework Form that uses the legacy project system?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -218,4 +218,13 @@ In order to debug this project, add an executable project to this solution which
   <data name="Restore_EmptyTargetFrameworkMoniker" xml:space="preserve">
     <value>The value of the 'TargetFrameworkMoniker' and 'NuGetTargetMoniker' properties in the '{0}' configuration are both empty. This configuration will not contribute to NuGet restore, which may result in restore and build errors.</value>
   </data>
+  <data name="WindowsFormEditor_DisplayName" xml:space="preserve">
```
Tanya-Solyanik: pardon my ignorance, but why are the old names not removed?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsEditorProvider.cs
Tanya-Solyanik: `DotNet` in not in this list - https://github.com/microsoft/VSProjectSystem/blob/master/doc/overview/project_capabilities.md, what projects have it, both core and classic?

davkean: Classic does not have .NET, but it doesn't matter this code never runs in the case of a legacy project system as it is CPS specific.

davkean:
@Tanya-Solyanik This does not affect legacy at all, just CPS.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
davkean: Do you mean from legacy?

Tanya-Solyanik: yes, are they still used?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsEditorProvider.cs
Tanya-Solyanik: By default .NET classic projects use the legacy project system, however we can create a CPS-based project  manually and it works fine. So this will disable double-click in these projects and we say that this is OK because they are not supported officially and we'll face it later if needed?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
Tanya-Solyanik: Did you have "old" names in CPS before your change?

davkean: CPS lives in a different code base. That will be deleted from there.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectSystemOptions.cs
davkean: There's a theoretical case where GetService can return null during IDE shutdown. I don't think we want to handle that case.

davkean:
This PR has been updated, changes:

- I've slightly changed the description of the "designers" to match the exact format of the rest of the designers/editors.
- Introduced the ability for a feature flag to override whether the designer opens for .NET Core 3.0 projects. @merriemcgaw @Tanya-Solyanik @DustinCampbell The intention is that you'll add this feature flag to your standalone VSIX which will "light up" for .NET Core 3.0 and above.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsEditorProvider.cs
```diff
@@ -0,0 +1,176 @@
...
+                return null;
+
+            SubTypeDescriptor? descriptor = await GetSubTypeDescriptorAsync(documentMoniker);
+            if (descriptor == null)
+                return null;
+
+            // Project properties are more expensive than above, so only
+            // read them if absolutely must open the designer for the file
+            if (!await IsDesignerEnabledAsync())
+                return null;
```
drewnoakes: This method serialises three async operations. Would it reduce delay when opening a file if they were all kicked off at the same time before being awaited? Based on your comment you may wish to do this only for the first two.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/PropertyExtensions.cs
```diff
@@ -30,5 +30,27 @@ public static async Task<Guid> GetValueAsGuidAsync(this IProperty property)
...
+
+        /// <summary>
+        ///     Returns the value of the specific <see cref="IProperty"/> as an enumeration, or
+        ///     the default of <typeparamref name="TEnum"/> if it cannot be parsed.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="property"/> is <see langword="null"/>.
+        /// </exception>
+        public static async Task<TEnum> GetValueAsEnumAsync<TEnum>(this IProperty property)
+            where TEnum : struct
```
drewnoakes: You can use an enum type constraint here.

davkean: Tried that first, unfortunately the parse isn't `T : enum`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -52,6 +52,9 @@
   <StringProperty Name="Configuration"
                   Visible="false" />
 
+  <StringProperty Name="WindowsFormsDesignerMode"
```
davidwengier: Why not define this as EnumProperty with the values here, and generated enum, rather than handle the mapping yourself and creating the enum yourself?

davkean: Does it actually generate a enum? They were still strings.

davidwengier: It should generate an enum (called WindowsFormsDesignerModeValues I guess) in the .cs file in the obj directory. From memory the values still get read as strings, but the cps property providers take care of checking for valid values, so consuming code can just use `string.Equals` for a cheaper comparison. It's possible that last statement is wrong however, and that's just the pattern of how CPS does it.

davkean: i'll look into it, it actually looks a lot more complicated than just reading a string value and parsing.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/PropertyExtensions.cs
drewnoakes: I think it should work:

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQAcoKgW1QOzgBsiBjACwlIGsBLfAcwBoYRaimATEAagB8AAgCYAjAFgAUAIDMAAmGyAwrIDek2RtkRCOWQFFVsgL7rNM+QBZZAWQA8AFQB8ACgCUsgO6Uks+7JCyyDAIcKQwTPo6qqaamno6AHT2CACeAApYyBDOAEQAgjkRtAz4APZIilBZAcFwEBGlcDC+skjIxDCuANwxxpJGQA==

davkean: Ah, didn't think to put struct in there as well. Thanks!

drewnoakes: Yeah it seems redundant.

davkean:
Heads up, I dialed this PR back to everything before we turned it off for .NET Core via feature which we'll merge in a follow up. I've made a couple of minor changes since last update, and added tests. I'll merge this in and do a follow up for the feature flag.

PR #4920, opened 2019-06-14
XAML Rule review

drewnoakes:
`Rule` has both [`Name`](https://github.com/microsoft/msbuild/blob/f65f3e1e6ccd33284a7dba489914048292ff02a3/src/Framework/XamlTypes/Rule.cs#L96) and [`DisplayName`](https://github.com/microsoft/msbuild/blob/f65f3e1e6ccd33284a7dba489914048292ff02a3/src/Framework/XamlTypes/Rule.cs#L110) properties. Only `DisplayName` is localised. Therefore, rule properties that are visible but do not have a `DisplayName` specified will never be localised in the UI.

This PR adds a unit test to ensure visible properties have a display name. It also adds several missing display names.

It also ensures property `Description` text ends with a full stop, which seems to be convention in the UI.

@tmeschter please review this from a loc perspective. It may not be something we want to merge until later, or perhaps not at all.

Related to #4105.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
@@ -85,6 +85,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.Rules
 
         [Theory]
         [MemberData(nameof(GetAllRules))]
+        public void PropertyDescriptionMustEndWithFullStop(string ruleName, string fullPath)
+        {
+            XElement rule = LoadXamlRuleX(fullPath).Root;
+
+            // Ignore XAML documents for other types such as ProjectSchemaDefinitions
+            if (rule.Name.LocalName != "Rule")
+                return;
+
+            foreach (var property in GetProperties(rule))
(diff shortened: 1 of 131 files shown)


tmeschter:
This looks fine to me. I believe master is a 16.3 branch now; did you want to take this for 16.2? If so, you need to get it into the appropriate branch by tomorrow.

drewnoakes:
No rush for these. If you want to merge it, feel free. Otherwise I'll leave it for 16.3. These have been like this for a long time now.

drewnoakes:
Just re-read that master is 16.3. Merging.

davkean:
I think we want to tweak this, so that we're only localizing rules that actually show up in the UI. For example, ConfigurationGeneral/Debugger are never shown in a UI, I think PropertyPagesHidden="true" should be an indicator of that and we should consistently apply this for non-localizable rules.

davkean:
Filed https://github.com/dotnet/project-system/issues/4944.

davidwengier:
> so that we're only localizing rules that actually show up in the UI

This is why the tests I'd written for these issues so far were only for the ones that had already been split into their different contexts. Doing the rest is (was?) still on my plate with various issues, but they're at the end of an increasing list, to say the least. @drewnoakes if you want to take over ownership of the rules and assign #4105 feel free.

PR #4923, opened 2019-06-14
Permanently disable the language version combo box

tmeschter:
This builds on top of #4918 to disable the language version combo box in the advanced build settings dialog.


src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
```diff
@@ -131,4 +131,15 @@
...
   </data>
+  <data name="LanguageVersionAutomaticallySelected" xml:space="preserve">
+    <value>Automatically selected based on framework version</value>
+    <comment>Refers to the C# langauge version being automatically selected based on the target framework</comment>
+  </data>
+  <data name="CantSelectLanguageVersionLinkLabel" xml:space="preserve">
+    <value>Why can't I select a different C# version?</value>
+  </data>
+  <data name="CantSelectLanguageVersionFWLink" xml:space="preserve">
+    <value>http://blah</value>
```
tmeschter: Please note that I won't merge this in until @cartermp has created the actual link for this.

jmarolf: aka.ms?

tmeschter: Yeah, it's going to be an aka.ms link, it just hasn't been determined yet.

cartermp: `aka.ms/csharp-versions` is now live.

The URL currently points to the `LangVersion` section of our docs, but I'll update that once the actual content is written.

cartermp: Fun story, this is how I got that URL:

* Edge --> website didn't load
* Chrome --> website didn't load
* Edge --> turn off all extensions --> website didn't load
* Chrome --> turn off all extensions --> website didn't load
* Firefox --> worked great, including with all of my extensions that block everything

tmeschter: Updated.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.resx
```diff
@@ -127,13 +127,13 @@
   </data>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="lblLanguageVersion.Location" type="System.Drawing.Point, System.Drawing">
-    <value>6, 23</value>
+    <value>6, 30</value>
```
drewnoakes: Are the large number of location/size changes here related to DPI settings in the designer? If so, have you validated that the UI still looks good at 100% scaling?

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
```diff
@@ -131,4 +131,15 @@
...
   </data>
+  <data name="LanguageVersionAutomaticallySelected" xml:space="preserve">
+    <value>Automatically selected based on framework version</value>
+    <comment>Refers to the C# language version being automatically selected based on the target framework</comment>
+  </data>
+  <data name="CantSelectLanguageVersionLinkLabel" xml:space="preserve">
+    <value>Why can't I select a different C# version?</value>
+  </data>
+  <data name="CantSelectLanguageVersionFWLink" xml:space="preserve">
+    <value>https://aka.ms/csharp-versions</value>
```
drewnoakes: This document doesn't currently reflect this change. Is that work being tracked somewhere?

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
```diff
@@ -131,4 +131,15 @@
...
+    <comment>{Locked}</comment>
```
drewnoakes: Is this a special syntax for prevening a string from being translated? If so is it understood by human or machine?

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
tmeschter: That's work @cartermp is planning on doing.

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
tmeschter: Yes, it's one of a set of specialized comments for exercising some control over the translation process. This one means that the string must preserve the original value across all locales. Our human translators understand these, and they are "machine-readable" to a certain extent. This one, for example, will prevent the string from actually being placed in the .xlf files--there's no point in the translators even seeing it if we don't want it translated.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.resx
tmeschter: Hmm, good question. I'll double-check.

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
cartermp: Yep.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.resx
tmeschter: I'm going to back out these changes just to be safe.

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
drewnoakes: Nice. Are these documented somewhere?

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
```diff
@@ -352,6 +347,13 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                 End If
             Next
         End Sub
+
+        Private Sub LnkLabel_LinkClicked(sender As Object, e As LinkLabelLinkClickedEventArgs) Handles lnkLabel.LinkClicked
+            TelemetryLogger.LogAdvBuildSettingsPropPageEvent(TelemetryLogger.AdvBuildSettingsPropPageEvent.LangVersionMoreInfoLinkClicked)
```
davkean: What business decision will we make in regards to the user clicking this? If we are making a business decision on this, might this be answered by looking at the aka link numbers, or the page views?

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
```diff
@@ -71,23 +68,8 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         ''' </summary>
         Private Sub PopulateLanguageVersions()
             cboLanguageVersion.Items.Clear()
-
-            cboLanguageVersion.Items.AddRange(CSharpLanguageVersionUtilities.GetAllLanguageVersions())
```
davkean: Can CSharpLanguageVersionUtilities be deleted if this is the only usage?

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
```diff
@@ -103,6 +85,19 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+        ''' <summary>
+        ''' Sets up the link to take the user to more information on why they can't change the language
+        ''' version.
+        ''' </summary>
+        Private Sub SetupLangVersionLinkLabel()
+            Dim link = New LinkLabel.Link
+            link.LinkData = My.Resources.Strings.CantSelectLanguageVersionFWLink
+            lnkLabel.Links.Clear()
+            lnkLabel.Links.Add(link)
+            lnkLabel.Text = My.Resources.Strings.CantSelectLanguageVersionLinkLabel
```
davkean: Why does this need to be dynamic every time the page is initialized? Is there a reason this can't be done in InitializeComponent?

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
```diff
@@ -352,6 +347,13 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                 End If
             Next
         End Sub
+
+        Private Sub LnkLabel_LinkClicked(sender As Object, e As LinkLabelLinkClickedEventArgs) Handles lnkLabel.LinkClicked
+            TelemetryLogger.LogAdvBuildSettingsPropPageEvent(TelemetryLogger.AdvBuildSettingsPropPageEvent.LangVersionMoreInfoLinkClicked)
+
+            Dim url = DirectCast(e.Link.LinkData, String)
+            Process.Start(url)
```
davkean: This can throw a Win32Exception in certain circumstances, such as if the default URL handler isn't installed.

src/Microsoft.VisualStudio.Editors/PropPages/Strings.resx
tmeschter: There's an internal document. I'll send around the link.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
tmeschter: I'll take a look.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
tmeschter: I'll investigate.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
tmeschter: Noted. I'll add a try/catch.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
cartermp: (Note: I didn't request this specific point, but I'll attempt to rationalize it below)

Both page views and aka link numbers could be skewed because you can access the information outside of VS. I don't think we'd make a business decision about this since we can't for sure say if clicking on this link indicates dissatisfaction, though, so if we can't find a business decision for this specific data point I'm not going to die on a hill to keep it.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.resx
tmeschter: I backed out these changes, restarted VS at 100% scaling, and the designer put them right back. So I'm going to assume these are fine.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
tmeschter: Yes, deleted.

src/Microsoft.VisualStudio.Editors/PropPages/AdvBuildSettingsPropPage.vb
tmeschter: I simplified this code and moved it so it is called once in the constructor.

miloush:
There is no mention in https://github.com/dotnet/project-system/pull/4918 about disabling this, probably the only functionality I ever used the dialog box for.

Not only it prevents users to configure their build environment, especially when you need to downgrade, but it also results in the users not seeing anymore what language version the compiler uses. Specially, if `<LangVersion>` is set in the project, this dialog will now still claim the version is automatically selected which is misleading.

miloush:
Here is an example of an issue that suffers from disabling this option: https://github.com/dotnet/roslyn/issues/38048

khellang:
Yeah, it would be nice if it at least showed the active language version: `C# 8.0 (automatically selected based on target framework .NET Core 3.0)`

cartermp:
@miloush Note that there's functionally no difference from using the now-deprecated dropdown and setting the LangVersion in your project file manually. So there's a workaround to the Roslyn bug today, it just doesn't involve UI.

miloush:
@cartermp true (and why I used suffer, not block) - the same way there is functionally no difference by writing the whole project file manually. The point of IDE is to make these changes easy in few seconds, without requiring and extra knowledge (and unloading the whole project). I just don't see why it is better to waste customers' time doing this manually.

cartermp:
In this case, since .NET Framework doesn't get new features, C# 8 is for .NET Core 3 (and .NET Standard 2.1). You can read about this decision here: https://devblogs.microsoft.com/dotnet/net-core-is-the-future-of-net/

> All future investment in .NET will be in .NET Core. This includes: Runtime, JIT, AOT, GC, BCL (Base Class Library), C#, VB.NET, F#, ASP.NET, Entity Framework, ML.NET, WinForms, WPF and Xamarin. 

This also includes new language features, hence the current defaults.

miloush:
I got that, but that doesn't explain why the UI needs to be removed, or why anyone who needs to downgrade the language version now has to do it manually. Again, there are no benefits to the user from this design change.

I think I have said all I could to make it clear I don't see the UI deprecation as rational and justified, and there are no new arguments on either side. I hope it goes well for the users.

Issue #4762, opened 2019-04-29
FrameworkReference support in Dependencies node


drewnoakes:
Tracking issue for FrameworkReference support in the Dependencies node.

- [x] Project system changes (PR #4928)
- [x] MSBuild `ResolveFrameworkReferences` target stub (PR https://github.com/microsoft/msbuild/pull/4471)
- [x] SDK `ResolveFrameworkReferences` target (PR https://github.com/dotnet/sdk/pull/3355)
- [ ] Verify FrameworkReference browse object (#5070, PR #5230)

drewnoakes:
Duplicated by: 

- https://dev.azure.com/devdiv/DevDiv/_workitems/edit/754156/
- https://dev.azure.com/devdiv/DevDiv/_workitems/edit/820329/
- https://dev.azure.com/devdiv/DevDiv/_workitems/edit/857810/
- https://github.com/dotnet/project-system/issues/4716

drewnoakes:
Note: this issue is about unbreaking the dependencies node experience for .NET Core 3. Issue #4362 is tracking improvements to the UX.

drewnoakes:
Adding `FrameworkReference` items to the tree without any additional filtering logic gives:

![image](https://user-images.githubusercontent.com/350947/59603115-e90c8f80-914b-11e9-9029-025522fe5ec2.png)

Questions at this point:

- [ ] Do framework references expand to show child dependencies?
- [ ] Does it make sense to show both the `Microsoft.WindowsDesktop.App` SDK _and_ Framework?
- [ ] Should the NuGet package used to deliver the framework be hidden, as is done for SDKs today?
- [ ] Is there a better icon for framework references?

//cc: @dsplaisted @nguerrera @cartermp @davkean

drewnoakes:
A more recent screenshot of the same project with the latest .NET Core SDK 3.0.100 preview:

![image](https://user-images.githubusercontent.com/350947/60181630-fa4d4e80-9865-11e9-98ee-6cea9b6d407b.png)

drewnoakes:
Once the MSBuild and SDK PRs are merged and available we'll be able to remove our own stub target and see actual resolved item properties in the browse object.

Issue #4444, opened 2019-01-09
Show assemblies within shared framework targeting packs


dsplaisted:
For .NET Core 3.0, we will be using [targeting packs](https://github.com/dotnet/designs/pull/50/files?short_path=ebf265d#diff-ebf265d0476041dd8cab82fa24ab5b0e) to deliver references for shared frameworks.

We've discussed that we would like these to be displayed in the solution explorer dependencies tree, something like this:

```
Dependencies
    Frameworks
        Microsoft.NETCore.App
            System.Runtime.dll
            System.Collections.Generic.dll
            ...
        Microsoft.WindowsDesktop.App
            System.Windows.Forms.dll
            ...
```

We'll need to decide what metadata should be set on these references to indicate which shared framework they are a part of, add that metadata in the SDK, and modify the project system to populate the dependencies tree based on that data.

dsplaisted:
FYI @davidwengier @Pilchie

drewnoakes:
Has some overlap with #4362.

cartermp:
This should be mostly covered by #4363 but it's good to have this be a distinct issue. We currently went with "references" as the name, but that's definitely up for discussion. As far as how they're displayed, the tree view that we have today for packages-that-are-really-shared-framework-references looks like it already has the desired structure.

jjmew:
Tracked by #AB827445

drewnoakes:
@dsplaisted how do you envisage this working in the case that two FrameworkReferences bring in the same reference?

dsplaisted:
Conflict resolution should choose the reference that's used, and that reference should have metadata on it that indicates which shared framework it came from.

Another case is when there are multiple framework references referring to different profiles of the same shared framework.  I'd suggest displaying that like this:

- Microsoft.WindowsDesktop.App
  - WindowsForms
    - System.Windows.Forms.dll
    - ...
  - WPF
    - System.PresentationFramework
    - ...
  - Microsoft.Win32.Registry.dll
  - ...

drewnoakes:
This is very similar to #1413. I commented there:

> Another approach to achieve the same result would be to visit the targeting pack on disk and discover the assemblies there directly and add them as graph nodes.
> 
> This data could be cached across all projects in the solution, allowing graph node reuse.
> 
> This might potentially provide less metadata on those referenced assemblies.

dsplaisted:
[...]

Issue #4362, opened 2018-12-10
Dependencies Node UX improvements


cartermp:
This is a tracking issue for features related to the UX of the dependencies node:

- [x] Show target frameworks in their own group, not under SDK (PR #4928)
- [ ] Only show dependency warning icon when there's a problem (#5044)
- [ ] Diagnostics not shown under packages in the dependencies node (#3780)
- [ ] Diagnostic severity not shown under packages in the dependencies node (#2582)
- [ ] Group package/project dependencies by provider type (#1255)
- [ ] Tree/graph nodes don't collapse on update (#6012)
- [ ] Provide feedback that dependencies node is populating (#5045)
- [ ] Use tool tip for information about dependencies node warnings (#4357)

drewnoakes:
I've seen a spinner for folders within projects, though don't know how that happens yet.

DamianEdwards:
What is the semantic difference that's worth highlighting in this case between Framework References and Shared Frameworks? I'd argue they're the same thing when it comes to visualizing them here. Also, can we make the nodes interactive at all, e.g. supporting deletion, selection, property editing via the grid, etc.

drewnoakes:
- #2079 warning icon while restoring
- #2257 #2653 #4357 general confusion about warning icons not showing tooltips or error list items

cartermp:
@DamianEdwards 

> What is the semantic difference that's worth highlighting in this case between Framework References and Shared Frameworks? I'd argue they're the same thing when it comes to visualizing them here.

In this case, I'm calling for Shared Framework and Framework Reference to be distinct sub-nodes. If a given project type gets its references from only one or the other, we should only show that sub-node unless one is added. Otherwise, both would have the same tree view for displaying their contents.

> Also, can we make the nodes interactive at all, e.g. supporting deletion, selection, property editing via the grid, etc.

I think that's a separate issue that I'll have to think about. Having a way to manage these various kinds of dependencies beyond the project file (and in NuGet's case, the NuGet UI) may be helpful. What sorts of items would you feel are good for CRUD operations?

DamianEdwards:
@cartermp still not really following. What is the difference? A FrameworkReference **is** a reference to a shared framework, so I'm not sure why they'd be separate at all.

cartermp:
[...]

PR #4883, opened 2019-05-29
Fast up-to-date check support for .editorconfig files

tmeschter:
1. Add EditorConfigFiles.xaml containing an item rule for <EditorConfigFile ... /> MSBuild items. This informs the project system that these are files we need to know about.
2. Add a corresponding EditorConfigFiles.BrowseObject.xaml file so that we can show basic info in the Properties window when clicking on a .editorconfig file in Solution Explorer. Not strictly necessary for fast up-to-date support, but nice to have.
3. Add PropertyPageSchema items for the new .xaml files to Microsoft.Managed.DesignTime.targets. This seems to be necessary for the project system to see that changes have been made in .editorconfig files.
4. Update CSharp.ProjectItemsSchema.xaml and VisualBasic.ProjectItemsSchema.xaml to effectively indicate that EditorConfigFiles items are input source files, like Compile and AdditionalFiles items.

PR #4928, opened 2019-06-17
FrameworkReference support

drewnoakes:
Relates to #4362 / #4762 / #4444.

Work in progress.

Outstanding questions:

- [x] Do framework references expand to show child dependencies? (deferred until later, tracked by #4444)
- [x] Should the NuGet package used to deliver the framework be hidden, as is done for SDKs today?
- [x] Does it make sense to show both the `Microsoft.WindowsDesktop.App` SDK _and_ Framework?
- [x] Is there a better icon for framework references?
- [x] Should we create a new `ResolvedFrameworkReference` rather than re-using `CollectedFrameworkReference`?

Outstanding tasks:

- [x] Update .swr files
- [x] Determine additional metadata for `FrameworkReference` item browse objects


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/FrameworkReference.xaml
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Rule Name="FrameworkReference"
```
davidwengier: New rule files need to be listed in CommonFiles.swr too

drewnoakes: Reviewing #4883 is on my TODO list, which will cover this.

cartermp:
> Do framework references expand to show child dependencies?

Is there a notable performance implication for this? I'd rather not pile onto any existing issues if possible here, and if there's nothing actionable to be done about seeing the child dependencies, I'd be fine with not showing them. If we get feedback about this later, we could revisit.

> Does it make sense to show both the `Microsoft.WindowsDesktop.App` SDK _and_ Framework?

I think showing Framework is sufficient here, since calling it an SDK is a bit of a misnomer (yes, this is a part of the literal SDK attribute on the project file).

I think there was a desire to be rid of `SDK` as a note altogether, since all it contained was items that were already shown elsewhere, and its contents were really just shared frameworks. Seems like we could put 'em both under `Frameworks`.

> Should the NuGet package used to deliver the framework be hidden, as is done for SDKs today?

I think so, yes.

> Is there a better icon for framework references?

Perhaps. But we can iterate on that and chat with some UX folks separately 🙂

drewnoakes:
@cartermp 

> Is there a notable performance implication for [showing child dependencies]?

I'm looking into this. I think it'd be useful and educational to see what's inside. Unless there's a strong reason not to, I'd like to allow this. Though it's lower priority and can slip beyond 16.3 if need be.

> I think showing Framework is sufficient here

I'll do this.

> I think there was a desire to be rid of SDK as a note altogether, since all it contained was items that were already shown elsewhere, and its contents were really just shared frameworks.

@davkean pointed out that "Extension SDKs" can appear in this node, and so we cannot lose the node altogether. We can, however, look to remove MSBuild SDKs from it, though this will require changes within the SDK itself to stop reporting them, as I don't think the project system can differentiate between the different kinds of SDK in any way. @davkean please correct me if I'm mistaken here.

drewnoakes:
Framework dependencies are now nested underneath, and the NuGet package is now hidden:

image

cartermp:
Gotcha, makes sense to show Extension SDKs if we have them. I think the desire I was referring to was for first-party "SDKs" (that were really shared framework references).

dsplaisted:
As @nguerrera has mentioned, it looks like you're using a very old version of the .NET Core SDK.  Current 3.0 previews shouldn't have any `PackageReference` items for the WindowsDesktop FrameworkReference.  Targeting and runtime packs are downloaded via `PackageDownload` items, which should not show up in the tree.  Currently, we do still have a `PackageReference` for Microsoft.NETCore.App if you have any other `PackageReference` items, but that will go away as part of https://github.com/dotnet/sdk/issues/3325.

> Do framework references expand to show child dependencies?

I think it would be a good idea if they expand to show the reference assemblies (mscorlib, System, System.Core, PresentationFramework, etc) coming from the targeting pack.  Here's an example of the metadata currently applied to the references resolved from targeting packs:

```
C:\Program Files\dotnet\packs\Microsoft.WindowsDesktop.App.Ref\3.0.0-preview7-27817-03\ref\netcoreapp3.0\PresentationFramework.dll
    AssemblyVersion = 4.0.0.0
    ExternallyResolved = true
    FileVersion = 4.800.19.31411
    FrameworkName = Microsoft.WindowsDesktop.App.WPF
    FrameworkVersion = 3.0.0-preview7-27817-03
    IsSystemReference = true
    NuGetPackageId = Microsoft.WindowsDesktop.App.Ref
    NuGetPackageVersion = 3.0.0-preview7-27817-03
    Private = false
    PublicKeyToken = 31bf3856ad364e35
    ReferenceGrouping = Microsoft.WindowsDesktop.App.WPF
    ReferenceGroupingDisplayName = Microsoft.WindowsDesktop.App.WPF
    ResolvedFrom = TargetingPack
    Visible = false
    WinMDFile = false
```

These `FrameworkName` or `ReferenceGrouping` metadata could be used to determine which shared framework to put these under.  We can also update the metadata if something else is needed.

> Should the NuGet package used to deliver the framework be hidden, as is done for SDKs today?

As mentioned, I don't think you should need to do anything special to hide this with the latest SDKs.

> Does it make sense to show both the Microsoft.WindowsDesktop.App SDK and Framework?

No, I think we should not have anything in the SDKs node for .NET Core 3.0 projects.  Personally, I think it would be ideal that for prior versions, we would show the implicit `PackageReference` items under the Frameworks node instead of under SDKs.

> Is there a better icon for framework references?

No opinion from me here.

> Can we use evaluated FrameworkReference items in PackageRestoreConfiguredDataSource?
> Should we create a new ResolvedFrameworkReference rather than re-using CollectedFrameworkReference?

I don't know about the details of these.

We also haven't accounted for the profiles of the WindowsDesktop shared framework.  The following are valid `FrameworkReferences`:

- `Microsoft.WindowsDesktop.App.WPF` - Only references assemblies needed for WPF
- `Microsoft.WindowsDesktop.App.WindowsForms` - Only references assemblies needed for Windows Forms
- `Microsoft.WindowsDesktop.App` - References both Windows Forms and WPF assemblies, as well as the integration DLL

Generally, which of these is referenced will depend on whether the `UseWPF` and/or `UseWindowsForms` properties are set to true in the project.  However, these references flow transitively, so if you are getting the `FrameworkReference` items via a design time build (which runs `AddTransitiveFrameworkReferences`), you may have more than one of these FrameworkReferences in the same project.

It might be nice to show subnodes for WPF and WindowsForms under the Microsoft.WindowsDesktop.App node.  On the other hand, it may be a lot simpler if you just get the FrameworkReferences from evaluation, in which case you would normally not have multiple WindowsDesktop FrameworkReferences, and you can just show the full name including WPF or WindowsForms under the Frameworks node.

drewnoakes:
> using a very old version of the .NET Core SDK

Indeed. My PC just arrived from the UK and I didn't realise how old the SDK was on it. Updating gives better results.

For project:

```xml
<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
  <PropertyGroup>
    <OutputType>WinExe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
    <UseWPF>true</UseWPF>
  </PropertyGroup>
</Project>
```

The whole tree looks like:

image

The tree doesn't currently change based upon combinations of `UseWPF` and `UseWindowsForms`.

> These FrameworkName or ReferenceGrouping metadata could be used to determine which shared framework to put these under

That sounds good, though in the example you show the framework is `Microsoft.WindowsDesktop.App.WPF` not `Microsoft.WindowsDesktop.App`. I'll be looking into this today but at first glance it seems like it might be an issue.

> ...if you are getting the FrameworkReference items via a design time build

The intention is to populate all top-level items via evaluation, and disallow adding top-level items in design time builds.

What additional data would you expect to be provided by a design time build about a `FrameworkReference`? Currently I assume:

- Contained assemblies (via the mechanism described above)
- Unresolved state (by absence from design time build items)

> ...if you just get the FrameworkReferences from evaluation, in which case you would normally not have multiple WindowsDesktop FrameworkReferences, and you can just show the full name including WPF or WindowsForms under the Frameworks node.

How would this work exactly? The evaluated item will be `Microsoft.WindowsDesktop.App`. How should the dependencies node know to change its display name? Is there a general mechanism in the design for this, or would it involve hard coding special knowledge of the frameworks into the project system?

davkean:
> I think there was a desire to be rid of SDK as a note altogether, since all it contained was items that were already shown elsewhere, and its contents were really just shared frameworks.

> We can, however, look to remove MSBuild SDKs from it, 

We're mixing concepts here, MSBuild SDKs are a completely different concept. We don't show them anywhere or do anything with them. You are referring to what I'm going to call "NuGet-based SDKs". We should not touch these at all, we do not show items that they represent elsewhere, the contents are only shown under SDKs. Let's treat this as a very separate concept to lighting up FrameworkReference.

drewnoakes:
This is now ready for review.

davidwengier:
Looks like there is an erroneous file here: https://github.com/drewnoakes/project-system/blob/framework-references/new

Also TIL GitHub doesn't allow file level comments

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Models/FrameworkDependencyModel.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+
+using Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot;
+using Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.RuleHandlers;
+
+#nullable enable
```
davidwengier: Nit: This is now not required

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/FrameworkRuleHandler.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+
+using Microsoft.VisualStudio.Imaging.Interop;
+using Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.CrossTarget;
+using Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Models;
+
+#nullable enable
```
davidwengier: Ditto, etc.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Models/FrameworkDependencyModel.cs
drewnoakes: Fixed. There was one in my PR from earlier today as well which I've removed here.

drewnoakes:
> Looks like there is an erroneous file here

Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -375,6 +383,11 @@
   <Target Name="CollectSDKReferencesDesignTime" />
   <Target Name="CollectResolvedSDKReferencesDesignTime" />
 
+  <Target
+    Name="ResolveFrameworkReferencesDesignTime"
+    Returns="@(FrameworkReference)"
+    DependsOnTargets="ResolveFrameworkReferences" />
```
drewnoakes: Once https://github.com/microsoft/msbuild/pull/4471 and https://github.com/dotnet/sdk/pull/3355 are merged, this will be changed to return `@(ResolvedFrameworkReference)`. Should it be moved somewhere else though?

davkean: This target isn't needed as you aren't doing anything additional.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/FrameworkReference.xaml
```diff
@@ -0,0 +1,27 @@
...
+      xmlns="http://schemas.microsoft.com/build/2009/properties">
+  
+  <Rule.DataSource>
+    <DataSource HasConfigurationCondition="False"
+                ItemType="FrameworkReference"
+                Persistence="ProjectFile"
+                SourceOfDefaultValue="AfterContext" />
+  </Rule.DataSource>
+
+  <StringProperty Name="OriginalItemSpec"
```
davkean: I would not expect this property to exist.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ResolvedFrameworkReference.xaml
```diff
@@ -0,0 +1,57 @@
...
+
+  <StringProperty Name="TargetingPackVersion"
+                  DisplayName="Targeting Pack Version"
+                  ReadOnly="True" />
+
+  <StringProperty Name="Profile"
+                  DisplayName="Profile"
+                  ReadOnly="True" />
+
+  <StringProperty Name="RuntimePackPath"
```
davkean: As per @dsplaisted RuntimeXXX properties don't exist all the time, so I would remove them to reduce confusion.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ResolvedFrameworkReference.xaml
```diff
@@ -0,0 +1,57 @@
...
+
+  <StringProperty Name="RuntimePackName"
+                  DisplayName="Runtime Pack Name"
+                  ReadOnly="True" />
+
+  <StringProperty Name="RuntimePackVersion"
+                  DisplayName="Runtime Pack Version"
+                  ReadOnly="True" />
+
+  <StringProperty Name="IsImplicitlyDefined"
```
davkean: We're expecting both of these properties (IsImplicitlyDefined/PrivateAssets) to be copied to the ResolvedFrameworkReference, but they aren't currently: https://github.com/dotnet/sdk/pull/3355/files#r297966595.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
drewnoakes: Isn't it needed until we have the actual target in MSBuild? In this case we're synthesizing the `ResolvedFrameworkReference`.

Once the `ResolveFrameworkReferences` task returns `ResolvedFrameworkReference` then won't this target become:

```xml
<Target name="ResolveFrameworkReferencesDesignTime" />
```

cartermp:
🎉

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ResolvedFrameworkReference.xaml
drewnoakes: https://github.com/dotnet/sdk/pull/3355 was updated to include `IsImplicitlyDefined` but not `PrivateAssets`. It seems like your conversation with @dsplaisted was unfinished in this regard, and I'm not sure what the better path forward here is. I'll make a note on #4762 to track this.

PR #4929, opened 2019-06-17
Turn on discovery of .editorconfig files

### LINKED ISSUE #4933: Remove deprecated analyzers
In https://github.com/dotnet/roslyn-analyzers/issues/1917#issuecomment-503293520 and https://github.com/dotnet/roslyn-analyzers/issues/1918#issuecomment-503290765 @mavasani points out that `Microsoft.CodeQuality.Analyzers.Exp` is deprecated.

In 92dbbc15e63961f9159b11fdec1c7ae6bcd8fa5f I updated our usage of Roslyn Analyzers from version 2.6.3-beta1.19059.2 to version 2.9.3 and discovered the following packages were not available at that newer version:

- `Text.Analyzers`
- `Microsoft.CodeQuality.Analyzers.Exp`
- ~`Roslyn.Diagnostics.Analyzers`~

@mavasani are all three of these deprecated? Will we be losing anything by removing references to them?

tmeschter:
1. Update to a newer version of Microsoft.Net.Compilers that understands
.editorconfig files.
2. Turn on the automatic discovery of .editorconfig files via the
`DiscoverEditorConfigFiles` property in src/Directory.Build.props. The
latest Microsoft.Net.Compilers understands .editorconfig but does not
yet search for them by default.
3. Replace our .ruleset files with equivalent settings in the .editorconfig file.

Fixes #4933.

Remaining work:
- [x] Fix nullability annotations such as done by @drewnoakes in #4957.
- [x] Update our CI builds to use an image with a recent VS2019. It looks like our current CI systems use VS2017 images, and those don't have the MSBuild changes needed to pass .editorconfig files to the compiler. Otherwise the CI builds will fail.
- [x] Wait for the fix made in https://github.com/dotnet/roslyn/pull/36670 to make it into d162.stg builds. Otherwise you'll end up with lots of spurious warnings and errors in the dotnet/project-system code in the IDE.
- [x] Test for and identify other issues.


drewnoakes:
> The latest Microsoft.Net.Compilers understands .editorconfig but does not yet search for them by default

If `DiscoverEditorConfigFiles` is `false` will Roslyn only consider `.editorconfig` project items?

tmeschter:
@drewnoakes The CoreCompile target in MSBuild will always pass along any `EditorConfigFiles` items to the compiler/IDE, so yes, any explicit `EditorConfigFiles` items in the project will be considered regardless of the `DiscoverEditorConfigFiles` property. This provides a way to turn off the automatic discovery if it is buggy or causing performance problems without completely losing support for .editorconfig files in the compiler.

drewnoakes:
Fixes #4933.

.editorconfig
```diff
@@ -47,8 +47,8 @@ dotnet_separate_import_directive_groups = true
 # Suggest more modern language features when available
 dotnet_style_object_initializer = true:suggestion
 dotnet_style_collection_initializer = true:suggestion
-dotnet_style_coalesce_expression = true:suggestion
-dotnet_style_null_propagation = true:suggestion
+dotnet_style_coalesce_expression = true:error
```
davkean: Why are these and other suggestions being changed and what relationship do they have with https://github.com/dotnet/project-system/issues/4933?

src/Directory.Build.props
```diff
@@ -51,6 +48,8 @@
     <CopyVsixManifestToOutput>false</CopyVsixManifestToOutput>
     
     <NoWarn>$(NoWarn);NU5125</NoWarn>
+
+    <DiscoverEditorConfigFiles>true</DiscoverEditorConfigFiles>
```
davkean: Why isn't this on by default? Are users expected to opt into this to avoid editor configs from being ignored?

davkean:
Given the build errors, looks like these editor configs aren't be respected.

.editorconfig
drewnoakes: @davkean the relationship to #4933 was edited in by me as I created that issue today and later noticed @tmeschter created this PR, which coincidentally fixes it.

davkean: Okay so my question is; I don't see the relationship between the commits/comments and these changes. Why are we now erroring on the lack usage of null propagation?

src/Directory.Build.props
tmeschter: It isn't turned on yet. It will be turned on by default for 16.2 GA.

.editorconfig
tmeschter: The .ruleset files set the corresponding IDE diagnostics (IDE0029, IDE0030, etc.) to be errors which would have overridden the "suggestion" specified in the .editorconfig. I saw no point in setting the severity twice within the same file via different mechanism, so where possible I've updated the existing "nice" name rather than adding another `dotnet_diagnostics.FOO1234.severity = master".

tmeschter:
Yeah, I'm going to need to figure out what's different between my local build and the CI build.

tmeschter:
Ah, the CI build is using images with VS2017 installed, and so the MSBuild we use won't have the required .editorconfig support. @jmarolf what are my options for moving us to a build queue with VS2019?

davkean:
I disabled some new rules for tests due to their noise and pushed it.

When I use this branch, it appears that the editorconfig isn't being respected, what's the trick to turn above on?

.editorconfig
davkean: Fair enough.

drewnoakes:
I merged #4957 to master, then merged master here which got all builds running except the Spanish one, as we don't yet have the necessary VS2019 ES image. @MattGal is on the case.

davkean:
This is falling behind changes, are there plans to make progress on this?

tmeschter:
@davkean Yes, I'm looking at this again. However, due to bugs you'll need to be on a 16.3 build to see everything working properly when working inside ProjectSystem.sln.

PR #4939, opened 2019-06-20
Don't hide/remove empty TFMs from the Dependencies node

### LINKED ISSUE #4937: Don't hide/remove empty TFMs
**Visual Studio Version**: 16.2.0-pre.2.0+29006.145

**Summary**:

TFMs with no dependencies are removed from the dependencies node, which can be confusing.

**Steps to Reproduce**:

1. With SDK 3.0.100 installed open this project:

   ```xml
     <PropertyGroup>
       <OutputType>Exe</OutputType>
       <TargetFrameworks>netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
     </PropertyGroup>
   </Project>
   ```

2. The Dependencies node looks like this:

   image

3. Add:

   ```xml
   <ItemGroup>
     <PackageReference Include="MetadataExtractor" Version="2.1.0" />
   </ItemGroup>
   ```

4. The Dependencies node now looks like:

   image

**Expected Behavior**:

Don't mislead the user into thinking that a package is used in all TFMs. We should show an empty ".NETCoreApp 3.0" node in the first screenshot and dock those children under ".NETCoreApp 2.0".

**Actual Behavior**:

See screenshot.

**User Impact**:

User can be misled.

drewnoakes:
Fixes #4937.

This PR is basically all in 89b57bbcec0831652773484a8827c40ff79ac500.

Previously any target framework not containing a dependency was removed from the snapshot. This was presumably to clean up when a target framework was removed from the project.

This change passes an explicit list of target frameworks through for the snapshot update, so that the tree can always show all target frameworks in multi-targeting projects.

Without this change, it's possible to have a misleading presentation in the UI (see #4937). With SDK 3.0.100, the MSBuild SDK no longer appears as a dependency, so it's much more likely we'd have target frameworks with no dependencies at all.

This also adds validation that the active target framework is present in related collections.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
@@ -175,17 +175,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
 
             if (activeTargetFramework.Equals(TargetFramework.Empty))
             {
-                if (dependenciesByTargetFramework.Count != 0)
-                {
-                    Requires.Fail($"If {nameof(activeTargetFramework)} is empty, {nameof(dependenciesByTargetFramework)} must be empty.");
-                }
+                Requires.Argument(
+                    dependenciesByTargetFramework.Count == 0,
+                    nameof(dependenciesByTargetFramework),
+                    $"Must be empty when {nameof(activeTargetFramework)} is empty.");
             }
             else
             {
-                if (!dependenciesByTargetFramework.ContainsKey(activeTargetFramework))
-                {
-                    Requires.Fail($"{nameof(activeTargetFramework)} must be present in {nameof(dependenciesByTargetFramework)}.");
-                }
+                Requires.Argument(
+                    dependenciesByTargetFramework.ContainsKey(activeTargetFramework),
+                    nameof(dependenciesByTargetFramework),
+                    $"Must contain {nameof(activeTargetFramework)} ({activeTargetFramework.FullName}).");
             }
 
             ProjectPath = projectPath;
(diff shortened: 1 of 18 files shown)


davkean:
Can you clarify the impact of this on a single TFM? For multi-TFM does this add a non-expandable node? If so, should we consider putting a child node under it saying that there's no dependencies?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Snapshot/DependenciesSnapshot.cs
```diff
@@ -148,27 +155,42 @@ bool RemoveTargetFrameworksWithNoDependencies()
...
         {
             Requires.NotNullOrEmpty(projectPath, nameof(projectPath));
-            Requires.NotNull(activeTarget, nameof(activeTarget));
-            Requires.NotNull(targets, nameof(targets));
+            Requires.NotNull(activeTargetFramework, nameof(activeTargetFramework));
+            Requires.NotNull(dependenciesByTargetFramework, nameof(dependenciesByTargetFramework));
+
+            if (activeTargetFramework.Equals(TargetFramework.Empty))
+            {
+                if (dependenciesByTargetFramework.Count != 0)
```
davkean: Requires.Argument with the condition would simplify these checks.

drewnoakes: Done.

I considered this at the time but because the validation is on the relationship between two arguments it wasn't clear which one to blame.

davkean: For the BCL, we typically just picked the first one in the relationship.

drewnoakes:
> Can you clarify the impact of this on a single TFM?

There is no change for single TFM.

> For multi-TFM does this add a non-expandable node?

Yes, for any TFMs with no dependencies.

> If so, should we consider putting a child node under it saying that there's no dependencies?

We don't do this anywhere else though. Eg. a P2P reference where the referenced project has no dependencies.

Personally I'm ok with the way it is now but will take feedback and am happy if you want to file an issue about it.

PR #4943, opened 2019-06-20
 Use Microsoft.Build.NoTargets SDK for documentation project

drewnoakes:
This dogfoods a custom MSBuild SDK project in our solution ([`Microsoft.Build.NoTargets`](https://github.com/microsoft/MSBuildSdks/tree/master/src/NoTargets)).

> The `Microsoft.Build.NoTargets` MSBuild project SDK allows project tree owners the ability to define projects that do not compile an assembly. This can be useful for utility projects that just copy files, build packages, or any other function where an assembly is not compiled.

### DIFF
--- a/docs/Documentation.csproj
+++ b/docs/Documentation.csproj
@@ -1,7 +1,11 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project>
 
+  <!-- See https://github.com/microsoft/MSBuildSdks/tree/master/src/NoTargets -->
+  <Sdk Name="Microsoft.Build.NoTargets" Version="1.0.73" />
+  
   <ItemGroup>
     <None Include="**/*.md" />
+    <PackageReference Remove="@(PackageReference)" />
   </ItemGroup>
 
 </Project>


docs/Documentation.csproj
```diff
@@ -1,7 +1,11 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project>
 
+  <!-- See https://github.com/microsoft/MSBuildSdks/tree/master/src/NoTargets -->
+  <Sdk Name="Microsoft.Build.NoTargets" Version="1.0.73" />
+  
   <ItemGroup>
     <None Include="**/*.md" />
+    <PackageReference Remove="@(PackageReference)" />
```
drewnoakes: This removes our default references (via `Directory.Build.props`) to make the Dependencies node nicer here. It's not required.

drewnoakes:
This is failing because the MSBuild project SDK doesn't include a `Test` target. I've opened https://github.com/microsoft/MSBuildSdks/issues/109 to see whether it should.

davkean:
Is this SDK in the box? Or is it using https://github.com/microsoft/msbuild/issues/4025? If the latter, it would be preferable if we did not use it given we're giving guidance to community folks to also not use it.

drewnoakes:
I don't think it's in the box, no. Your call between trading off dogfooding the feature vs. giving the right impression.

PR #4946, opened 2019-06-21
Save 230,912 bytes via lossless image optimisations

drewnoakes:
PNG files on disk reduced from 237,955 bytes down to 7,290 bytes (a 97% reduction).

In a Release build, this commit reduces the size of `Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll` from 900,096 bytes to 669,184 bytes (a 26% reduction).

In both cases the delta is 230,912 bytes as the files are copied into the assembly unchanged.

---

This PR, combined with the removal of unused strings in #4896, brings the in-memory module size of `Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll` down from 1.05 MB to 0.68 MB.

### Before

image

### After

image


davkean:
Nice!

PR #4947, opened 2019-06-21
Save 68,609 bytes by replacing icon with XAML

drewnoakes:
In a Release build, `Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll` is reduced from 900,096 bytes to 831,488 bytes (an 8% reduction).

---

This compounds with reductions from #4946 and #4896 to help bring down the installation size and working set.


davkean:
I think this is good? But I don't know enough about how this affects DPI, I assume this automatically scales?

drewnoakes:
> how this affects DPI, I assume this automatically scales?

The .ico file was so large because it had 17 different BMP files inside it at different sizes (from 16x16 to 256x256) and bit depths (8bpp and 32bpp). That's useful when you need the icon to appear quite small, as you can hand align pixels to a coarse 96dpi screen to keep things crisp. It was also useful in the old days when 32bpp colour wasn't ubiquitous.

That dialog shows this image at 32x32 display units (equivalent to ~1/3 inch on screen) which is large enough that the default XAML rendering will look crisp enough at 96dpi.

At higher DPIs it will scale automatically and look identical.

Issue #4951, opened 2019-06-21
Add RuntimeIdentifierGraphPath to nomination


nkolev92:
**Visual Studio Version**: 16.2

**Summary**:

Original ask [here](https://github.com/NuGet/Home/issues/7351). 

The SDK will provide the RuntimeIdentifierGraphPath, that NuGet will treat that as the project provided runtime identifier graph.
This will effectively replace the platforms package. 

cc @davkean 

**Steps to Reproduce**:
N/A

**Expected Behavior**:
N/A
**Actual Behavior**:
N/A
**User Impact**:
Affets .NET Core 3.0 customers.

davkean:
This has been merged.

PR #4952, opened 2019-06-21
Pass RuntimeIdentifierGraphPath during nomination

nkolev92:
Fixes https://github.com/dotnet/project-system/issues/4951

Original ask [here](https://github.com/NuGet/Home/issues/7351).
Spec at https://github.com/NuGet/Client.Engineering/blob/master/Specs/RuntimeJsonFromProject.md. 

The SDK will provide the RuntimeIdentifierGraphPath, that NuGet will treat that as the project provided runtime identifier graph.
This will effectively replace the platforms package.

cc @davkean @jjmew

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
@@ -89,6 +89,10 @@
                   ReadOnly="True"
                   Visible="False" />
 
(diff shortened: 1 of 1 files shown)


davkean:
@jjmew This in 16.2, what do we need to take this for approval?

jjmew:
> @jjmew This in 16.2, what do we need to take this for approval?

A regular insertion to d16.2stg before 7/1 will be enough. It just needs to go in the right branch. 
CC: @jmarolf

nkolev92:
Thanks @davkean, @jjmew

PR #4955, opened 2019-06-23
Change Control subtype to UserControl

davidwengier:
Fix the enum value for user controls.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
@@ -140,7 +140,7 @@
     </EnumProperty.DataSource>
     <EnumValue Name="Designer" />
     <EnumValue Name="Component" />
-    <EnumValue Name="Control" />
+    <EnumValue Name="UserControl" />
     <EnumValue Name="Form" />
     <EnumValue Name="Code" />
   </EnumProperty>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/ApplicationDefinition.BrowseObject.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/ApplicationDefinition.BrowseObject.xaml
@@ -140,7 +140,7 @@
     </EnumProperty.DataSource>
     <EnumValue Name="Designer" />
     <EnumValue Name="Component" />
-    <EnumValue Name="Control" />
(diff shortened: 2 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
```diff
@@ -140,7 +140,7 @@
     </EnumProperty.DataSource>
     <EnumValue Name="Designer" />
     <EnumValue Name="Component" />
-    <EnumValue Name="Control" />
+    <EnumValue Name="UserControl" />
     <EnumValue Name="Form" />
```
davkean: This is also wrong, this doesn't exist.

davkean:
Is this the underlying cause of https://github.com/dotnet/roslyn/issues/36609?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
davidwengier: `Form`? Okay, I shall remove.

davidwengier:
Yes and no. @lifengl identified three issues in an email, and this is one of them, however he's also changed CPS so the values here don't matter any more, if I understand correctly.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/AdditionalFiles.BrowseObject.xaml
davidwengier: Where is the official list of subtypes? `Form` is used in [our own code](https://github.com/dotnet/project-system/blob/afa02d302e08df1ffe8689dbcd6b9d65c6f59c2f/src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj#L50) in a couple of spots, and is part of legacy: http://index/?query=%60VSSUBTYPE_FORM&rightProject=VS(Repo)&file=src%5Cvsproject%5Cinc%5Csubtypes.h&line=25

davkean: Yeah my bad, I even handled this in https://github.com/dotnet/project-system/pull/4913. I was originally under the impression that Form/Designer were the same, but they aren't: http://index/?query=Form&rightProject=System.Windows.Forms&file=System%5CWinForms%5CForm.cs&line=49. Let me find which control uses "Designer".

PR #4957, opened 2019-06-23
Update nullability annotations

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/GraphViewProviderBase.cs
```diff
@@ -121,7 +122,7 @@ IEnumerable<string> GetExistingChildren()
         public virtual bool MatchSearchResults(
             IDependency topLevelDependency,
             Dictionary<string, HashSet<IDependency>> searchResultsPerContext,
-            out HashSet<IDependency>? topLevelDependencyMatches)
+            [MaybeNullWhen(returnValue: false)] out HashSet<IDependency>? topLevelDependencyMatches)
```
davidwengier: Does this mean that the out param definitely won't be null when the method returns true?

drewnoakes: The above construct is incorrect. Pushed a correction.

The four attributes (AllowNull/DisallowNull/NotNull/MaybeNull) and the conditional form above all override the type's declared nullability. So in the above example the HashSet should be non-null with the attribute specifying a null return when false (or a nullable type with NotNullWhen(true)).

davidwengier: Yep, the update makes much more sense. Couldn't work out what "maybe null" meant when the parameter type was already nullable.

drewnoakes:
I think this needs a newer compiler. I've looked but can't see how to update the reference.

davkean:
`<MicrosoftNetCompilersVersion>` in Versions.props controls this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/GraphNodes/ViewProviders/IDependenciesGraphViewProvider.cs
```diff
@@ -60,6 +60,6 @@ internal interface IDependenciesGraphViewProvider
         bool MatchSearchResults(
             IDependency topLevelDependency,
             Dictionary<string, HashSet<IDependency>> searchResultsPerContext,
-            [NotNullWhenTrue] out HashSet<IDependency>? topLevelDependencyMatches);
+            [MaybeNullWhen(false)] out HashSet<IDependency> topLevelDependencyMatches);
```
davkean: If it "may be" null, then why isn't it marked nullable?

drewnoakes: It actuall was until my force push. You use annotations when you want to override the declared type's nullability. Another example would be a `string?` property where the setter cannot be null (i.e. annotate the setter with `[DisallowNull]`). It's also useful on generic types such as `IEquatable<T>` (which reminds me, I should update one of the annotations for that interface in the dependencies node.)

tmeschter:
We need to be careful that we don't conflict with #4929 as that also updates the Microsoft.Net.Compiler package.

PR #4962, opened 2019-06-25
Add SubType/ItemType support to project tree parser

davkean:
Add SubType/ItemType support to project tree parser, and a few other changes.

Need this to test the WindowsFormsEditorProvider.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/ProjectSystem/ProjectTreeWriterOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/ProjectSystem/ProjectTreeWriterOptions.cs
@@ -10,10 +10,10 @@ namespace Microsoft.VisualStudio.ProjectSystem
         None = 0,
         Tags = 1,
         FilePath = 2,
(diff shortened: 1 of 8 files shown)


davkean:
Putting this in 16.2 as need this for the WindowsFormsEditorProvider, which is also going into it. This is a test only change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/ProjectSystem/ProjectTreeWriterOptions.cs
```diff
@@ -10,9 +10,11 @@ internal enum ProjectTreeWriterOptions
...
         Tags = 1,
         FilePath = 2,
-        Capabilities = 4,
+        Flags = 4,
         Visibility = 8,
         Icons = 16,
-        AllProperties = FilePath | Visibility | Capabilities | Icons
+        ItemType = 32,
+        SubType = 64,
+        AllProperties = FilePath | Visibility | Flags | Icons | ItemType
```
davidwengier: Missing SubType

PR #4966, opened 2019-06-26
Turn on nullable by default

davkean:
- "Baseline" existing files by opt'd them out at the file level
- Turn on nullable by default at the project level 
- Removed (now) unneeded `#nullable enable` directives.

This "stops the bleeding" by opt'ing all new files into nullable.


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/IDependenciesChanges.cs
```diff
@@ -4,6 +4,8 @@
 
 #nullable enable    
 
+#nullable disable
+
```
drewnoakes: Actually is this correct?

drewnoakes: GitHub's choking on this PR. Will fetch locally and re-check.

davkean: That looks like the only one, the extra whitespace after enable threw off the RegEx.

davkean:
```
E:\project-system2>findstr /sip /c:"nullable enable" *.cs

E:\project-system2>
```

Issue #4944, opened 2019-06-21
Stop localizing/adding localizable values in rules that never show up in the UI


davkean:
In https://github.com/dotnet/project-system/pull/4920, we added some awesome new validation to make sure that we're using rules consistently and property. However, in doing so we've started localizing more properties that are really never ever show up in the UI. We've started using PropertyPagesHidden="true" on some of those rules, and we should consistently apply this across all rules that never end up in as a browse object.

davkean:
@drewnoakes I've assigned to you for now, but feel free to throw to someone else. One reason I assigned to you was because the Reference rules are in a fuzzy territory - it's not clear to me which of the rules are the "browse object" and hence shown in the properties windows .

drewnoakes:
@davkean I'm happy to take this one. Reducing the LOC load and in-memory footprint sounds like pure goodness.

Once I'm further along with `FrameworkReference` support I'd like to revisit the Dependencies node rule files, especially around BrowseObject support.

drewnoakes:
@davkean these are the files that currently define `PropertyPagesHidden="true"`.

- `GeneralBrowseObject.xaml`
- `GeneralConfiguredBrowseObject.xaml`
- `Items/AdditionalFiles.BrowseObject.xaml`
- `Items/AdditionalFiles.xaml`
- `Items/ApplicationDefinition.BrowseObject.xaml`
- `Items/ApplicationDefinition.xaml`
- `Items/Compile.BrowseObject.xaml`
- `Items/Compile.xaml`
- `Items/Content.BrowseObject.xaml`
- `Items/Content.xaml`
- `Items/EditorConfigFiles.BrowseObject.xaml`
- `Items/EditorConfigFiles.xaml`
- `Items/EmbeddedResource.BrowseObject.xaml`
- `Items/EmbeddedResource.xaml`
- `Items/Folder.xaml`
- `Items/None.BrowseObject.xaml`
- `Items/None.xaml`
- `Items/Page.BrowseObject.xaml`
- `Items/Page.xaml`
- `Items/Resource.BrowseObject.xaml`
- `Items/Resource.xaml`
- `Items/XamlAppDef.BrowseObject.xaml`
- `Items/XamlAppDef.xaml`

Are you sure none of these should be translated?

davkean:
Ugh, your right. My bad - that PropertyPagesHidden is really a bad indication of this.

davkean:
[...]

PR #4967, opened 2019-06-26
Add missing translations to setup file and add test to prevent such problems in the future

drewnoakes:


drewnoakes:
This is failing with:

> D:\a\1\s\setup\Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles\CommonFiles.swr(97,2): error SWIX1108: Could not find file at path: D:\a\1\s\artifacts\Debug\VSSetup\Rules\cs\NuGetRestore.xaml. [C:\Users\VssAdministrator\.nuget\packages\roslyntools.repotoolset\1.0.0-beta-62705-01\tools\VisualStudio.SetupPackage.swixproj]

(Repeated for each culture.)

`NuGetRestore` is translated (due to `DisplayName`s) but those translations are not included in the install.

The only difference I can see between `NuGetRestore` and other XAML rule files is how they're included in `Microsoft.Managed.*.targets` files:

```xml
<PropertyPageSchema Include="$(ManagedXamlNeutralResourcesDirectory)NuGetRestore.xaml">
```

vs

```xml
<PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)PackageReference.xaml">
```

That is, `ManagedXamlResourcesDirectory` vs `ManagedXamlNeutralResourcesDirectory`.

If this is the case:
- all properties in `NuGetRestore.xaml` should be `Visible="false"` so that the translation is not performed for that file, and
- we have a technique for identifying which files should not be translated for #4944.

//cc: @davkean

drewnoakes:
> we have a technique for identifying which files should not be translated for #4944

Just realised this is what @davkean was suggesting in https://github.com/dotnet/project-system/issues/4944#issuecomment-504316505.

davidwengier:
`Neutral` in the variable name in this case means "Language neutral", ie it doesn't need to be localized.

PR #4973, opened 2019-07-01
DesignTimeInputs data source and tests

davidwengier:
This is the first little piece of the latest rewrite of TempPE, this time using data flow. I wouldn't be surprised if this changes (in particular the shape of `DesignTimeInputs`) but I'm putting it up early because it's my first ever use of data flow.

Also, excitingly, it's the first set of tests I've been able to write that isn't tied horribly to the internal implementation details of the class under test! Well except all of those mocks that are used by it and its base classes, and the specific parts of the project changes that I specified in JSON, but it's close to good at least! 😛

Also note this is going into a feature branch.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal class DesignTimeInputs
+    {
+        public string[] Inputs = Array.Empty<string>();
+        public string[] SharedInputs = Array.Empty<string>();
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal class DesignTimeInputs
+    {
+        public string[] Inputs = Array.Empty<string>();
+        public string[] SharedInputs = Array.Empty<string>();
```
drewnoakes: Consider making this type immutable. I've only seen immutable data passed around with Dataflow and assume mutable types can lead to corruption if a consumer starts modifying the data.

Similarly consider using `ImmutableArray<string>`.

Can `DesignTimeInputs` be a `readonly struct`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
```
drewnoakes: Missing copyright header in this and other files.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
davidwengier: Yeah, as mentioned in the description the shape of this will change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
davidwengier: Thanks, I _always_ forget this. I wish there was an analyzer for it.

drewnoakes: `StyleCop.Analyzers` offers this in [SA1636](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1636.md).

davidwengier: Yeah, CPS has that on, but my experience with StyleCop Analyzers does not enamour me to turn them on.

drewnoakes: It's slow in that solution because there are so many hidden rules and thousands of diagnostics. If that was all cleaned up, I don't think you'd notice a perf impact. It's also no fun with overly prescriptive rules.

PR #4974, opened 2019-07-01
Add waiter service

jmarolf:
allows the user to cancel a long running rename operation


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IVsThreadedWaitDialogFactoryFactory.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+
+using Microsoft.VisualStudio.ProjectSystem.VS;
+using Microsoft.VisualStudio.Shell.Interop;
+
+using Moq;
+
+using Xunit;
+
+namespace Microsoft.VisualStudio.Mocks
```
davkean: This namespace always matches where the type it's extending lives so that it just shows up in intellisense.

PR #4978, opened 2019-07-02
Force .cs and .vb to use crlf

drewnoakes:
Several unit tests include string literals for expected values. Without this setting, a repo checked out with alternative line ending settings will have failing tests. This can occur, for example, if you clone the repo using WSL.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,8 +12,8 @@
 # default for csharp files.
 # Note: This is only used by command line
 ###############################################################################
-*.cs     diff=csharp text 
-*.vb     text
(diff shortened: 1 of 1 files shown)


drewnoakes:
Quite right. `eol=crlf` is what I wanted. I did test this, but my approach was flawed as I'd also set a global `core.autocrlf` variable that made things work. I've cleared that config setting, updated `.gitattributes`, removed the cached file and hard reset. Things are working correctly now. Thanks.

davidwengier:
Just in case you want to be really pedantic about having a clean git history, the commit message lists `.cs` twice 😛

PR #4980, opened 2019-07-02
Snapshot tracks only visible unresolved items

drewnoakes:
Allowing non-visible items to turn this flag 'true' can produce a "Dependencies" node with a warning icon, but no children with a corresponding warning to drill into.

Relates to #4362.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyExtensionsTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyExtensionsTests.cs
@@ -50,31 +50,31 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot
                 KnownMonikers.AboutBox,
                 KnownMonikers.Abbreviation);
 
-            var dependencyResolved = IDependencyFactory.FromJson(@"
-{
-    ""ProviderType"": ""Yyy"",
-    ""Id"": ""tfm1\\yyy\\dependencyExisting"",
-    ""Name"":""dependencyExisting"",
-    ""Caption"":""DependencyExisting"",
-    ""HasUnresolvedDependency"":""false"",
-    ""SchemaName"":""MySchema"",
-    ""SchemaItemType"":""MySchemaItemType"",
-    ""Priority"":""1"",
-    ""Resolved"":""true""
-}", iconSet: iconSet);
+            var dependencyResolved = new TestDependency
+            {
+                ProviderType = "Yyy",
+                Id = "tfm1\\yyy\\dependencyExisting",
+                Name = "dependencyExisting",
+                Caption = "DependencyExisting",
+                SchemaName = "MySchema",
+                SchemaItemType = "MySchemaItemType",
+                Priority = 1,
+                Resolved = true,
+                IconSet = iconSet
+            };
 
-            var dependencyUnresolved = IDependencyFactory.FromJson(@"
-{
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Tree/Dependencies/Snapshot/DependencyExtensionsTests.cs
```diff
@@ -50,31 +50,31 @@ public void ToViewModel()
                 KnownMonikers.AboutBox,
                 KnownMonikers.Abbreviation);
 
-            var dependencyResolved = IDependencyFactory.FromJson(@"
-{
-    ""ProviderType"": ""Yyy"",
-    ""Id"": ""tfm1\\yyy\\dependencyExisting"",
-    ""Name"":""dependencyExisting"",
-    ""Caption"":""DependencyExisting"",
-    ""HasUnresolvedDependency"":""false"",
```
drewnoakes: This change gets rid of `HasUnresolvedDependency`, which isn't even on `IDependency`. For this simple test type I prefer the strongly typed approach.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesTreeViewProvider.cs
```diff
@@ -78,7 +78,7 @@ internal sealed class DependenciesTreeViewProvider : IDependenciesTreeViewProvid
 
             dependenciesTree = CleanupOldNodes(dependenciesTree, currentTopLevelNodes);
 
-            ProjectImageMoniker rootIcon = _viewModelFactory.GetDependenciesRootIcon(snapshot.HasUnresolvedDependency).ToProjectSystemType();
+            ProjectImageMoniker rootIcon = _viewModelFactory.GetDependenciesRootIcon(snapshot.HasVisibleUnresolvedDependency).ToProjectSystemType();
```
drewnoakes: This is where the icon is determined at the root level. This PR means only visible and unresolved descendents of the root can cause the root to have a yellow triangle.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -438,7 +438,7 @@ private void OnDependenciesSnapshotChanged(object sender, SnapshotChangedEventAr
                         {
                             dependenciesNode = await viewProvider.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
 
-                            await _treeTelemetryService.ObserveTreeUpdateCompletedAsync(snapshot.HasUnresolvedDependency);
+                            await _treeTelemetryService.ObserveTreeUpdateCompletedAsync(snapshot.HasVisibleUnresolvedDependency);
```
drewnoakes: Note that this change will impact telemetry, in keeping with what's presented on screen.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -183,9 +183,9 @@ void Add(AddDependencyContext context, IDependencyModel dependencyModel)
                     string.Equals(id, dependency.Id),
                     "dependenciesWorld dictionary entry keys must match their value's ids.");
 
-                if (!dependency.Resolved)
+                if (!dependency.Resolved && dependency.Visible)
```
drewnoakes: Here is the additional constraint.

PR #4981, opened 2019-07-02
Add a stub for target ResolveFrameworkReferences

drewnoakes:
The stub immediately below this declares it depends upon `ResolveFrameworkReferences` yet that target may not exist.


davkean:
Can we file a bug to remove this and the DesignTime targets?

drewnoakes:
> Can we file a bug to remove this and the DesignTime targets?

This is being tracked in #4762.

davkean:
👍

PR #4995, opened 2019-07-05
Design time inputs file watcher

davidwengier:
A dataflow block that listens for design time inputs and shared inputs, and watches those files, and then sends output whenever a file is changed.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/TestHelpers/TestIVsAsyncFileChangeEx.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/TestHelpers/TestIVsAsyncFileChangeEx.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.Shell
+{
+    public class TestIVsAsyncFileChangeEx : IVsAsyncFileChangeEx
+    {
+        private uint _lastCookie;
+        private readonly Dictionary<uint, string> _watchedFiles = new Dictionary<uint, string>();
+
+        public IEnumerable<string> WatchedFiles => _watchedFiles.Values;
+
+        public Task<uint> AdviseFileChangeAsync(string filename, _VSFILECHANGEFLAGS filter, IVsFreeThreadedFileChangeEvents2 sink, CancellationToken cancellationToken = default)
+        {
+            uint cookie = _lastCookie++;
+            _watchedFiles.Add(cookie, filename);
+            return System.Threading.Tasks.Task.FromResult(cookie);
+        }
+
+        public Task<string> UnadviseFileChangeAsync(uint cookie, CancellationToken cancellationToken = default)
+        {
+            string file = _watchedFiles[cookie];
+            _watchedFiles.Remove(cookie);
+            return System.Threading.Tasks.Task.FromResult(file);
+        }
+
+        public Task<uint> AdviseDirChangeAsync(string directory, bool watchSubdirectories, IVsFreeThreadedFileChangeEvents2 sink, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
+        }
+
+        public System.Threading.Tasks.Task IgnoreDirAsync(string directory, bool ignore, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
+        }
+
+        public System.Threading.Tasks.Task IgnoreFileAsync(uint cookie, string filename, bool ignore, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
+        }
+
+        public System.Threading.Tasks.Task SyncFileAsync(string filename, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task<string> UnadviseDirChangeAsync(uint cookie, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
@@ -4,12 +4,12 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
-using System.Threading;
 using System.Threading.Tasks.Dataflow;
 
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Threading;
+
 using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
@@ -25,7 +25,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
         private readonly IDesignTimeInputsDataSource _designTimeInputsDataSource;
         private readonly IVsService<IVsAsyncFileChangeEx> _fileChangeService;
         private ImmutableDictionary<string, uint> _fileWatcherCookies = ImmutableDictionary<string, uint>.Empty.WithComparers(StringComparers.Paths);
-        private readonly SemaphoreSlim _disposeGate = new SemaphoreSlim(initialCount: 1);
 
         private int _version;
         private IDisposable? _dataSourceLink;
@@ -39,6 +38,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
         /// The public facade for the broadcast block.
         /// </summary>
         private IReceivableSourceBlock<IProjectVersionedValue<string[]>>? _publicBlock;
+
+        /// <summary>
+        /// The block that actually does our processing
+        /// </summary>
+        private ITargetBlock<IProjectVersionedValue<DesignTimeInputs>>? _actionBlock;
 
         [ImportingConstructor]
         public DesignTimeInputsFileWatcher(ConfiguredProject project,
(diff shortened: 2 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+            _disposables.AddDisposable(join);
+
+            _broadcastBlock = DataflowBlockSlim.CreateBroadcastBlock<IProjectVersionedValue<string>>(nameFormat: nameof(DesignTimeInputsFileWatcher) + "Broadcast {1}");
+            _publicBlock = _broadcastBlock.SafePublicize();
+        }
+
+        private async Task ProcessDesignTimeInputs(IProjectVersionedValue<DesignTimeInputs> input)
+        {
+            // We don't want to watch any new files while we're disposing
+            if (IsDisposing)
```
davkean: But if you're already disposed its okay? What if you get disposed while this method is running?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            IVsAsyncFileChangeEx vsAsyncFileChangeEx = await _fileChangeService.GetValueAsync();
+
+            // we don't care about the difference between types of inputs, so we just construct one hashset for fast comparisons later
+            var allFiles = new HashSet<string>(StringComparers.Paths);
+            allFiles.AddRange(designTimeInputs.Inputs);
+            allFiles.AddRange(designTimeInputs.SharedInputs);
+
+            // Remove any files we're watching that we don't care about any more
+            foreach (string file in _fileWatcherCookies.Keys.ToArray())
```
davkean: The synchronization in this method does not look correct, it's possible to both watch and unwatch the same file if two threads enter this method. I don't think you are expecting this method to overlap, right? If so, just avoid usage of Interlocked as it confuses the sitaution.

davkean: Why do you need the ToArray? `_fileWatcherCookies` is immutable, it's not going to change underneath you.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _broadcastBlock?.Complete();
+                _disposables.Dispose();
+
+                _threadingService.ExecuteSynchronously(async () =>
```
davkean: This can overlap with ProcessDesignTimeInputs you need to protect both methods with a lock.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+                    }
+                });
+            }
+
+            base.Dispose(disposing);
+        }
+
+        public int FilesChanged(uint cChanges, string[] rgpszFile, uint[] rggrfChange)
+        {
+            for (int i = 0; i < cChanges; i++)
```
davkean: I'd consider merging these results into a single notification.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                PostToOutput(rgpszFile[i]);
+            }
+
+            return HResult.OK;
+        }
+
+        public int DirectoryChanged(string pszDirectory)
+        {
+            throw new NotImplementedException();
```
davkean: Return `HResult.NotImplemented` and below, you can't throw from a COM interface as you are breaking the contract and managed consumers will need to handle both HRs and exceptions.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
davkean: See my comment below in dispose to handle this, both methods should be protected by a lock.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        protected override void Initialize()
+        {
+            base.Initialize();
+
+            IDisposable dataSourceLink = _designTimeInputsDataSource.SourceBlock.LinkToAsyncAction(ProcessDesignTimeInputs);
+
+            IDisposable join = JoinUpstreamDataSources(_designTimeInputsDataSource);
```
davkean: You don't need to capture the result here, it's already added to a list and disposed by the base: http://index/?query=ProjectValueDataSource&rightProject=Microsoft.VisualStudio.ProjectSystem&file=Core%5CDataflow%5CProjectValueDataSourceBase.cs&line=140.

This will reduce the need to have a DisposableBag, just store the link and dispose it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <summary>
+        /// The public facade for the broadcast block.
+        /// </summary>
+        private IReceivableSourceBlock<IProjectVersionedValue<string>>? _publicBlock;
+
+        [ImportingConstructor]
+        public DesignTimeInputsFileWatcher(ConfiguredProject project,
+                                           IDesignTimeInputsDataSource designTimeInputsDataSource,
+                                           IVsService<SVsFileChangeEx, IVsAsyncFileChangeEx> fileChangeService)
+             : base(project.Services, synchronousDisposal: true, registerDataSource: false)
```
davkean: Once you've introduced a lock to protect Dispose from overlapping with dataflow, you'll need to turn  `synchronousDisposal` off to avoid deadlocks, see: http://index/?query=ProjectValueDataSource&rightProject=Microsoft.VisualStudio.ProjectSystem&file=Utilities%5CDisposables%5COnceInitializedOnceDisposed.cs&line=46.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
davidwengier: Thanks. Going to use a CancellationTokenSource to cancel any processing that is happening when Dispose is called.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
davidwengier: Thanks

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
davidwengier: Actually, that would still leave a small window for leaking a token. A semaphore probably works better, though is more likely to result in watching files, then unwatching them immediately.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/TestHelpers/TestIVsAsyncFileChangeEx.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.Shell
+{
+    public class TestIVsAsyncFileChangeEx : IVsAsyncFileChangeEx
```
davkean: This is a mock, can we name it as such and move it?

davidwengier: Yep. I wasn't sure about this when I made it, but I guessed that mocks were exclusively for Moq mocks :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            JoinUpstreamDataSources(_designTimeInputsDataSource);
+
+            _broadcastBlock = DataflowBlockSlim.CreateBroadcastBlock<IProjectVersionedValue<string[]>>(nameFormat: nameof(DesignTimeInputsFileWatcher) + "Broadcast {1}");
+            _publicBlock = _broadcastBlock.SafePublicize();
+        }
+
+        private async Task ProcessDesignTimeInputs(IProjectVersionedValue<DesignTimeInputs> input)
+        {
+            await _disposeGate.ExecuteWithinLockAsync(JoinableCollection, JoinableFactory, async () =>
```
davkean: This doesn't fix the race. A thread can enter this method and still have its state disposed underneath it. Both Dispose and this method need to share the same lock.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                return _publicBlock!;
+            }
+        }
+
+        protected override void Initialize()
+        {
+            base.Initialize();
+
+            _dataSourceLink = _designTimeInputsDataSource.SourceBlock.LinkToAsyncAction(ProcessDesignTimeInputs);
```
davkean: This has a race, your callback can be called before the public/broadcast block exist.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
davidwengier: Thanks, I copied this usage from elsewhere but I guess that had difference concerns. I was hoping Dispose would wait for the lock to be released. Should have checked :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+    /// Produces output whenever a design time input changes
+    /// </summary>
+    [Export(typeof(IDesignTimeInputsFileWatcher))]
+    [AppliesTo(ProjectCapability.CSharpOrVisualBasicLanguageService)]
+    internal class DesignTimeInputsFileWatcher : ProjectValueDataSourceBase<string[]>, IVsFreeThreadedFileChangeEvents2, IDesignTimeInputsFileWatcher
+    {
+        private readonly IProjectThreadingService _threadingService;
+        private readonly IDesignTimeInputsDataSource _designTimeInputsDataSource;
+        private readonly IVsService<IVsAsyncFileChangeEx> _fileChangeService;
+        private ImmutableDictionary<string, uint> _fileWatcherCookies = ImmutableDictionary<string, uint>.Empty.WithComparers(StringComparers.Paths);
```
davidwengier: If we trust Dataflow to not overlap calls, which we pretty much do in this class, this doesn't really need to be an immutable dictionary. Thoughts?

dmonroym: Might be better to use a ConcurrentDictionary then

---
In reply to: [302824562](https://github.com/dotnet/project-system/pull/4995#discussion_r302824562) [](ancestors = 302824562)

davidwengier: I've changed to use a normal dictionary. Dataflow will ensure that there is no overlap with this method.

PR #4996, opened 2019-07-08
Annotate up to date check

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -372,9 +370,12 @@ private bool CheckGlobalConditions(BuildAction buildAction, BuildUpToDateCheckLo
...
         {
-            logger.Verbose("Adding project file inputs:");
-            logger.Verbose("    '{0}'", _msBuildProjectFullPath);
-            yield return _msBuildProjectFullPath;
+            if (_msBuildProjectFullPath != null)
+            {
+                logger.Verbose("Adding project file inputs:");
+                logger.Verbose("    '{0}'", _msBuildProjectFullPath);
+                yield return _msBuildProjectFullPath;
+            }
```
drewnoakes: This prevents misleading log output. The consumer noop'd the null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -420,7 +421,10 @@ private IEnumerable<string> CollectInputs(BuildUpToDateCheckLogger logger)
             if (_customInputs.Count != 0)
             {
                 logger.Verbose("Adding " + UpToDateCheckInput.SchemaName + " inputs:");
+                // TODO remove pragmas when https://github.com/dotnet/roslyn/issues/37040 is fixed
+#pragma warning disable CS8622
                 foreach (string input in _customInputs.Select(_configuredProject.UnconfiguredProject.MakeRooted))
+#pragma warning restore CS8622
```
drewnoakes: Filed https://github.com/dotnet/roslyn/issues/37040 for a few instances of this problem.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -445,26 +452,32 @@ private IEnumerable<string> CollectOutputs(BuildUpToDateCheckLogger logger)
...
             foreach (string input in inputs)
             {
                 DateTime? time = GetTimestampUtc(input, timestampCache);
 
                 if (time > latest)
                 {
+                    // TODO remove pragmas when https://github.com/dotnet/roslyn/issues/37039 is fixed
+#pragma warning disable CS8629 // Nullable value type may be null
                     latest = time.Value;
+#pragma warning restore CS8629
```
drewnoakes: Filed https://github.com/dotnet/roslyn/issues/37039 for this problem.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -512,10 +525,12 @@ private IEnumerable<string> CollectOutputs(BuildUpToDateCheckLogger logger)
         private bool CheckOutputs(BuildUpToDateCheckLogger logger, IDictionary<string, DateTime> timestampCache)
         {
             // We assume there are fewer outputs than inputs, so perform a full scan of outputs to find the earliest
-            (DateTime? outputTime, string outputPath) = GetEarliestOutput(CollectOutputs(logger), timestampCache);
+            (DateTime? outputTime, string? outputPath) = GetEarliestOutput(CollectOutputs(logger), timestampCache);
 
             if (outputTime != null)
             {
+                Assumes.NotNull(outputPath);
```
drewnoakes: Would be nice if this taught the below code that `outputPath` can no longer be `null`. I don't believe there's an attribute for that either.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
davidwengier: That's a good catch, I would have assumed that any warning there would be because CPS is un-annotated.

drewnoakes: The NRT feature is designed to work when consuming un-annotated code. In general if you're seeing warnings that you don't agree with, it's worth digging in (at least while the feature is young).

Issue #4958, opened 2019-06-24
Feature: Operation progress support for NuGet


davkean:
Tracking the operation progress support for NuGet coming in 16.3.

PR #4999, opened 2019-07-08
Add restore operation progress support

davkean:
Fixes: https://github.com/dotnet/project-system/issues/4958.

This is ~a draft of~ the operation progress support for package restore. This causes a project's IntelliSense stage to be considered in progress until restore has finished. 

See the comments for PackageRestoreProgressTracker for more information on how this works under the covers.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreConfiguredDataSource.cs
```diff
@@ -49,29 +49,50 @@ protected override UnconfiguredProject ContainingProject
...
+            IProjectValueDataSource<IProjectSnapshot> projectSource = _projectSubscriptionService.ProjectSource;
+            IProjectValueDataSource<IProjectSubscriptionUpdate> jointRuleSource = _projectSubscriptionService.JointRuleSource;
 
             // Transform the changes from evaluation/design-time build -> restore data
-            DisposableValue<ISourceBlock<RestoreUpdate>> transformBlock = source.SourceBlock
-                                                                                .TransformWithNoDelta(update => update.Derive(u => CreateRestoreUpdate(_project, u.CurrentState)),
-                                                                                                      suppressVersionOnlyUpdates: false,    // We need to coordinate these at the unconfigured-level
-                                                                                                      ruleNames: s_rules);
+            IPropagatorBlock<IProjectVersionedValue<Tuple<IProjectSnapshot, IProjectSubscriptionUpdate>>, RestoreUpdate> transformBlock = DataflowBlockSlim.CreateTransformBlock<IProjectVersionedValue<Tuple<IProjectSnapshot, IProjectSubscriptionUpdate>>, RestoreUpdate>(
```
jmarolf: no var use with data flow continues to make me sad :(

jmarolf:
This looks great!

Unrelated to this change, but something that is on my mind: if we have situation where restore isn't up-to-date after we retry cotinuously.  Couldn't we use this `PackageRestoreProgresTracker` to just say that the restore failed after a certain number of DTBs (saw two) so the users machine doesn't heat up?

davkean:
This went through some pretty big changes since last update:

- I did not like how the PackageRestoreInitiator talked to the trackers, it felt very icky and didn't have them participate in "progress" until the design-time build had completed. This could result in a race where we'd miss the sliding window of progress during startup. This has been fixed by introducing IPackageRestoreService and the progress tracker now initializes during startup.
- I ended moving the sync-linking of the project snapshot in the trackers themselves instead of the ConfiguredDataSource. The latter didn't use or care about that at all, and was not used in the data we sent to NuGet.
- The asset file + date is now part of the output of restore, only a single instance needs to read the file on disk.

PR #5000, opened 2019-07-08
Validate top level rule DataSource properties

drewnoakes:
Validates:

1. When `SourceType` is `TargetResults` that `MSBuildTarget` is present
1. That `MSBuildTarget` is only present when `SourceType` is `TargetResults`
1. When `SourceType` is `Item` that `ItemType` is present
1. That `ItemType` is only present when `SourceType` is `Item`

This resulted in removing a bunch of redundant `ItemType` attributes from rule data sources.


davkean:
@drewnoakes Good change! Love seeing these sorts of things.

PR #5003, opened 2019-07-09
Annotate most of Microsoft.VisualStudio.ProjectSystem.Managed

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed/ImmutableCollectionLinqExtensions.cs
```diff
@@ -41,6 +40,7 @@ internal static class ImmutableCollectionLinqExtensions
             return false;
         }
 
+        [return: MaybeNull]
         public static T FirstOrDefault<T, TArg>(this ImmutableArray<T> immutableArray, Func<T, TArg, bool> predicate, TArg arg)
```
drewnoakes: This is the recommended approach for indicating that an output with unconstrained generic type can be null.

drewnoakes: This is the recommended approach for indicating that an output with unconstrained generic type can be null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ImmutableCollectionLinqExtensions.cs
```diff
@@ -49,7 +49,7 @@ internal static class ImmutableCollectionLinqExtensions
                     return obj;
             }
 
-            return default;
+            return default!;
```
drewnoakes: Unfortunately the attribute above doesn't influence the member body, so we need to `!` force this here. Tracked by https://github.com/dotnet/roslyn/issues/36039.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -31,8 +29,8 @@ internal class ApplyChangesToWorkspaceContext : OnceInitializedOnceDisposed, IAp
         private readonly ConfiguredProject _project;
         private readonly IProjectLogger _logger;
         private readonly ExportFactory<IWorkspaceContextHandler>[] _workspaceContextHandlerFactories;
-        private IWorkspaceProjectContext _context;
-        private ExportLifetimeContext<IWorkspaceContextHandler>[] _handlers;
+        private IWorkspaceProjectContext? _context;
+        private ExportLifetimeContext<IWorkspaceContextHandler>[] _handlers = Array.Empty<ExportLifetimeContext<IWorkspaceContextHandler>>();
```
drewnoakes: In some cases it is easier to use singleton empty collections rather than `null`. Doing so can avoid lots of null checks or `!`, especially when there's an implicit relationship between being initialised and non-null (as happens a lot in this repo).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -137,30 +135,34 @@ protected override void Dispose(bool disposing)
...
             foreach (ExportLifetimeContext<IWorkspaceContextHandler> handler in _handlers)
             {
-                handler.Value.Initialize(_context);
+                handler.Value.Initialize(_context!);
             }
         }
 
         private void ProcessProjectBuildFailure(IProjectRuleSnapshot snapshot)
         {
+            Assumes.NotNull(_context);
```
drewnoakes: Unfortunately these guard methods are not annotated to indicate that the argument is not null after this point. There is a `DoesNotReturnIf(bool)` attribute, but not a `DoesNotReturnIfNull(string parameterName)` attribute. It seems we have to `!` the first usage after this point. I filed https://github.com/dotnet/roslyn/issues/37071 to see if we could just `!` the argument to the guard method directly, which would be nicer.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/BuildOptions.cs
```diff
@@ -5,13 +5,11 @@
...
 using Microsoft.CodeAnalysis;
 
-#nullable disable
-
 namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
 {
     internal class BuildOptions
     {
-        private static readonly PropertyInfo s_analyzerConfigPathsProperty = typeof(CommandLineArguments).GetProperty("AnalyzerConfigPaths");
+        private static readonly PropertyInfo? s_analyzerConfigPathsProperty = typeof(CommandLineArguments).GetProperty("AnalyzerConfigPaths");
```
drewnoakes: `Type.GetProperty` is not annotated. I saw a later `null` test and annotated this. If the `?` here was removed, it would still compile without warning.

I mention this as there are probably other places like this in the repo.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/CompileItemHandler.cs
```diff
@@ -31,6 +30,7 @@ public CompileItemHandler(UnconfiguredProject project)
         }
 
         [ImportMany]
+        [NotNull, AllowNull]
         private readonly IEnumerable<IFileRenameHandler> _fileRenameHandlers = null;
```
drewnoakes: This one was interesting. Here `AllowNull` says the field can be set to null (as done in the initializer), yet `NotNull` says that consumers of the field will not receive a null. In this case that's true enough as the property's set by MEF after the ctor completes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Logging/ProjectLoggerExtensions.ProjectLoggerBatch.cs
```diff
@@ -45,43 +43,22 @@ public void WriteLine(in StringFormat format)
             {
                 if (IsEnabled)
                 {
-                    Init();
-                    WriteNewLineIfNeeded();
-                    WriteIndent();
-                    Write(format);
```
drewnoakes: I inlined these methods as it allows the compiler to flow information about the null state of `_builder` through all methods.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/ProjectRuleSnapshotExtensions.cs
```diff
@@ -4,6 +4,8 @@
 
 #nullable disable
 
+// TODO annotate this file once NotNullIfNotNull is supported https://github.com/dotnet/roslyn/issues/31549
```
drewnoakes: This file still disables the nullable feature. Without `NotNullIfNotNull` support, it's hard to get this right. https://github.com/dotnet/roslyn/issues/35816 is tracking that work.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
```diff
@@ -107,7 +105,11 @@ public async Task<string> GetFileAsync(SpecialFiles fileId, SpecialFileFlags fla
             }
 
             // We haven't found the file but return the default file path as that's the contract.
-            IProjectTree rootNode = _projectTree.CurrentTree;
+            IProjectTree? rootNode = _projectTree.CurrentTree;
+
+            if (rootNode == null)
+                return null;
```
drewnoakes: This is a behaviour change. Instead of NRE it returns null. There were other ways this could return null however, so I don't think it's a problem.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppDesignerFolderSpecialFileProvider.cs
```diff
@@ -30,17 +28,12 @@ public AppDesignerFolderSpecialFileProvider(Lazy<IPhysicalProjectTree> projectTr
             _properties = properties;
         }
 
-        // For unit tests
-        protected AppDesignerFolderSpecialFileProvider()
-        {
-        }
```
drewnoakes: Removed unused constructor, as it gave warnings about non-null fields not being initialised.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppDesignerFolderSpecialFileProvider.cs
```diff
@@ -56,22 +49,29 @@ public virtual async Task<string> GetFileAsync(SpecialFiles fileId, SpecialFileF
...
         }
 
-        private async Task<string> GetDefaultAppDesignerFolderPathAsync()
+        private async Task<string?> GetDefaultAppDesignerFolderPathAsync()
         {
-            string rootPath = _projectTree.Value.TreeProvider.GetRootedAddNewItemDirectory(_projectTree.Value.CurrentTree);
+            IProjectTree? currentTree = _projectTree.Value.CurrentTree;
+
+            if (currentTree == null)
+                return null;
```
drewnoakes: Behaviour change. Instead of ANE, return null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UnconfiguredProjectTasksService.cs
```diff
@@ -49,7 +46,7 @@ public Task OnProjectFactoryCompleted()
             }
             else
             {
-                _projectLoadedInHost.TrySetResult(false);
+                _projectLoadedInHost.TrySetResult(null);
```
drewnoakes: We were boxing a bool as an object here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
```diff
@@ -69,7 +67,7 @@ internal Enumerator(in LazyStringSplit split)
                 _index = 0;
                 _input = split._input;
                 _delimiter = split._delimiter;
-                Current = null;
+                Current = null!;
```
drewnoakes: I think this makes sense for enumerators. Technically `Current` can be null, but 99.99% of consumers are `foreach`, in which case you won't see that state of the enumerator. You can still see nulls if the sequence has null elements though (which it won't here).

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
```diff
@@ -75,7 +75,7 @@ void HandleChangesForRule(bool resolved, IProjectChangeDescription projectChange
                 foreach (string removedItem in projectChange.Difference.RemovedItems)
                 {
                     string dependencyId = resolved
-                        ? projectChange.Before.GetProjectItemProperties(removedItem).GetStringProperty(ResolvedAssemblyReference.OriginalItemSpecProperty) ?? removedItem
+                        ? projectChange.Before.GetProjectItemProperties(removedItem)!.GetStringProperty(ResolvedAssemblyReference.OriginalItemSpecProperty) ?? removedItem
```
drewnoakes: Assuming the changed item has properties here, hence `!`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
```diff
@@ -109,10 +109,10 @@ void HandleChangesForRule(bool resolved, IProjectChangeDescription projectChange
 
                 IDependencyModel CreateDependencyModelForRule(string itemSpec, IProjectRuleSnapshot projectRuleSnapshot)
                 {
-                    IImmutableDictionary<string, string> properties = projectRuleSnapshot.GetProjectItemProperties(itemSpec);
+                    IImmutableDictionary<string, string> properties = projectRuleSnapshot.GetProjectItemProperties(itemSpec)!;
 
                     string originalItemSpec = resolved
-                        ? properties.GetStringProperty(ResolvedAssemblyReference.OriginalItemSpecProperty)
+                        ? properties.GetStringProperty(ResolvedAssemblyReference.OriginalItemSpecProperty) ?? itemSpec
```
drewnoakes: Behaviour change. if somehow the original item spec property was null, use the item spec.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Utilities/MetadataExtensions.cs
```diff
@@ -37,7 +36,7 @@ internal static class MetadataExtensions
         /// <param name="key">The key that identifies the property to look up.</param>
         /// <param name="stringValue">The value of the string if found and non-empty, otherwise <see langword="null"/>.</param>
         /// <returns><see langword="true"/> if the property was found with a non-empty value, otherwise <see langword="false"/>.</returns>
-        public static bool TryGetStringProperty(this IImmutableDictionary<string, string> properties, string key, out string stringValue)
+        public static bool TryGetStringProperty(this IImmutableDictionary<string, string> properties, string key, [NotNullWhen(returnValue: true)] out string? stringValue)
```
drewnoakes: `TryGet*` pattern.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppDesignerFolderSpecialFileProvider.cs
drewnoakes: Turns out this is needed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/FaultExtensions.cs
```diff
@@ -37,7 +35,7 @@ internal static class FaultExtensions
         /// <param name="project">
         ///     The project related to the failure, if applicable. Can be <see langword="null"/>.
         /// </param>
-        public static Task ReportFaultAsync(this IProjectFaultHandlerService faultHandlerService, Exception ex, UnconfiguredProject project, ProjectFaultSeverity severity = ProjectFaultSeverity.Recoverable)
+        public static Task ReportFaultAsync(this IProjectFaultHandlerService faultHandlerService, Exception ex, UnconfiguredProject? project, ProjectFaultSeverity severity = ProjectFaultSeverity.Recoverable)
```
jmarolf: may want to wrap these like you did in the other files

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
jmarolf: 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
jmarolf: can we file an issue on the VSSDK folks for this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/FSharp/FSharpCommandLineParserService.cs
```diff
@@ -22,7 +20,7 @@ internal class FSharpCommandLineParserService : ICommandLineParserService
         private const string LongReferencePrefix = "--reference:";
 
         [ImportMany]
-        private readonly IEnumerable<Action<string, ImmutableArray<CommandLineSourceFile>, ImmutableArray<CommandLineReference>, ImmutableArray<string>>> _handlers = null;
+        private readonly IEnumerable<Action<string, ImmutableArray<CommandLineSourceFile>, ImmutableArray<CommandLineReference>, ImmutableArray<string>>>? _handlers = null;
```
jmarolf: should this be an empty array?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/CompileItemHandler.cs
jmarolf: Should this set of attributes be added for all MEF import collections? Is this the general pattern we want to use?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
```diff
@@ -120,7 +118,7 @@ private async Task<T> ExecuteUnderLockCoreAsync<T>(Func<CancellationToken, Task<
 
                 try
                 {
-                    T result = default;
+                    T result = default!;
```
jmarolf: I assume we want to say that T may be null here?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -30,11 +28,11 @@ internal partial class WorkspaceProjectContextHostInstance : OnceInitializedOnce
...
 
-            private IDataProgressTrackerServiceRegistration _evaluationProgressRegistration;
-            private IDataProgressTrackerServiceRegistration _projectBuildProgressRegistration;
-            private DisposableBag _disposables;
-            private IWorkspaceProjectContextAccessor _contextAccessor;
-            private ExportLifetimeContext<IApplyChangesToWorkspaceContext> _applyChangesToWorkspaceContext;
+            private IDataProgressTrackerServiceRegistration? _evaluationProgressRegistration;
+            private IDataProgressTrackerServiceRegistration? _projectBuildProgressRegistration;
+            private DisposableBag? _disposables;
+            private IWorkspaceProjectContextAccessor? _contextAccessor;
```
jmarolf: Shouldn't this be non-null?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
drewnoakes: Can do, but there's nothing they can do until the compiler supports it. As soon as there's compiler support, I'll send them a PR if they don't get to it first.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/CompileItemHandler.cs
drewnoakes: I'm on the fence here. It works, and is a 'correct' usage of the attributes, but it looks pretty weird.

An alternative is just to assign an empty array here. It involves more work/allocation than just assigning null (which is a no-op) but the overhead is trivial and it's easier to see that it's correct.

The way it is now, it's valid to set it to null elsewhere in the class, which is obviously not what we want. I think I'll assign `Array.Empty<T>()` here, and would vote for doing that elsewhere too.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
drewnoakes: We can't unfortunately, as `T` is an unconstrained generic type.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
drewnoakes: It's not assigned in the constructor. If we want it to be non-null, we either need a non-null placeholder (as in the case of empty collections) or to `#pragma warning disable` the ctor, which doesn't feel great. I don't think there's a good placeholder here.

This happens a fair bit in this repo, where things are initialised after construction.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ImmutableCollectionLinqExtensions.cs
davkean: Ugh - that is not great.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
davkean: I agree in some cases, but in this case it would be a bug that this is null that is no longer caught.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/FSharp/FSharpCommandLineParserService.cs
```diff
@@ -22,7 +21,7 @@ internal class FSharpCommandLineParserService : ICommandLineParserService
         private const string LongReferencePrefix = "--reference:";
 
         [ImportMany]
-        private readonly IEnumerable<Action<string, ImmutableArray<CommandLineSourceFile>, ImmutableArray<CommandLineReference>, ImmutableArray<string>>> _handlers = null;
+        private readonly IEnumerable<Action<string, ImmutableArray<CommandLineSourceFile>, ImmutableArray<CommandLineReference>, ImmutableArray<string>>> _handlers = Array.Empty<Action<string, ImmutableArray<CommandLineSourceFile>, ImmutableArray<CommandLineReference>, ImmutableArray<string>>>();
```
davkean: I would prefer these are not filled in as it would be a bug if they are null by the time they are touched.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
davkean: We should probably change these methods to return empty to behave identiical to MSBuild; if a property doesn't exist its "empty". Probably need to rationalize these extension methods with this `ProjectRuleSnapshotExtensions`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/FSharp/FSharpCommandLineParserService.cs
drewnoakes: Good point. The alternative is to attribute them with `[NotNull, AllowNull]` and leave off the `?` modifier. I'll do that before merging.

drewnoakes: I can't actually get this to work with attributes alone. Leaving the type non-nullable and assigning `null!` works. I've made the change.

An observation while searching for instances of this: most `ImportMany` fields use `OrderPrecedenceImportCollection` which is always initialised and would hide any import bugs. I haven't checked but suspect MEF would assign an empty enumerable here anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
drewnoakes: Actually it's possible to do this with the current attributes:

```c#
void AssertNotNull([NotNull] object? o) { }

string M(object? o)
{
    AssertNotNull(o);
    return o.ToString(); // no diagnostic here
}
```

davkean:
We're want to merge this in quickly to avoid it fall behind. @jmarolf Are you good with the change?

jmarolf:
Yep, lets not hold this one up

PR #5004, opened 2019-07-09
Ignore test deployment & coverage files

davkean:
In 16.2, deployments & coverage files are no longer deployed to TestResults - ignore their new locations.


davkean:
I've asked internally if this is by-design.

Issue #1437, opened 2017-02-02
We should be respecting $(ProjectAssetsFile)


davkean:
We find the assets files ourselves in the ProjectLockFileWatcher. We shouldn't be doing that, we should respecting $(ProjectAssetsFile).
See: https://github.com/NuGet/Home/issues/4463#issuecomment-276832985

nguerrera:
Something just occurred to me. @dsplaisted  made changes so that SDK will not default ProjectAssetsFile and instead use what restore writes to generated temp props. Would this create a chicken-and-egg problem for initiating the watch when restore hasn't been done yet?

davkean:
@nguerrera Yes, this is a chicken and egg. If we can't find $(ProjectAssetsFile) we can't tell CPS that it needs to reevaluate. We were also planning to move to a model where CPS itself monitored the file based on https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L281 but sounds like this would also be broken by that.

nguerrera:
:(

dsplaisted:
I don't know how feasible this is, but I think the "right" thing to do would be to monitor the `MSBuildProjectExtensionsPath`.  NuGet restore will drop .props and .targets which should be imported there, as well as the assets.json file.  So when the new .props and .targets show up there the project should be re-evaluated.  The .props file will then set the `ProjectAssetsFile` property, and you can then additionally watch for changes in that file if it's necessary.

davkean:
We already monitor the props/targets when we reevaluate and they appear, also wouldn't play nicely with shared intermediates.

Was this change to remove logic from SDK? If so, can we add it back to NuGet targets?

davkean:
- Assets file changed
- Reevaluate
- New props/targets are picked up

We don't monitor for props/targets existence under they appear in evaluation - there's thousand of locations they could appear.

dzmitry-lahoda:
Not sure if right thread - but some tooling is wrong...

See the line with `ProjectAssetsFile`? If I remove it - my `csproj` fails to build. 
[...]

PR #5006, opened 2019-07-09
Replace ProjectAssetFileWatcher with IFileWatchDataSource

davkean:
IFileWatchDataSource is a new feature that was added to unify the logic of re-revaluating a project based on external files (such as .editorconfig and project.assets.json). Plug in `<AdditionalDesignTimeBuildInput />` items, of which project.assets.json is the only one.

This also has the side-effect of playing nicely with other features that trigger reevaluation to avoid unnecessarily reevaluations when others have triggered it or during branch switches.

Fixes: #1437


davkean:
@drewnoakes The new tests caught that I missed the setup portion!

drewnoakes:
Good to hear 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AdditionalDesignTimeBuildInputDataSource.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            return transformBlock;
+        }
+
+        private FileWatchData CreateFileWatch(IImmutableDictionary<string, IProjectRuleSnapshot> update)
+        {
+            IProjectRuleSnapshot snapshot = update.GetSnapshotOrEmpty(AdditionalDesignTimeBuildInput.SchemaName);
+
+            var fullPaths = snapshot.Items.Select(item => item.Value)
+                                          .Select(metadata => metadata.GetValueOrDefault(AdditionalDesignTimeBuildInput.FullPathProperty))
```
drewnoakes: You could merge these two `Select`s, unless this was for style reasons.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AdditionalDesignTimeBuildInput.xaml
```diff
@@ -0,0 +1,23 @@
...
+                ItemType="AdditionalDesignTimeBuildInput"
+                Persistence="ProjectFile"
+                SourceOfDefaultValue="AfterContext" />
+  </Rule.DataSource>
+
+  <StringProperty Name="FullPath"
+                  ReadOnly="true"
+                  Visible="false">
+    <StringProperty.DataSource>
+      <DataSource ItemType="AdditionalDesignTimeBuildInput"
```
drewnoakes: An aside: I haven't looked much into per-property data sources. Can you think of validation rules we can apply on them? Do they behave the same way as top-level rule data sources? My guess was not, so #5000 only validates the top level occurrences.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AdditionalDesignTimeBuildInputDataSource.cs
```diff
@@ -0,0 +1,71 @@
+﻿ // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
drewnoakes: Is there a leading space here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/ProjectAssetFileWatcher.ProjectAssetFileWatcherInstance.cs
```diff
@@ -1,317 +0,0 @@
...
-            }
-
-            private static byte[] GetFileHashOrNull(string path)
-            {
-                byte[] hash = null;
-
-                try
-                {
-                    using var hasher = SHA256.Create();
-                    using FileStream file = File.OpenRead(path);
```
drewnoakes: I sure won't miss hitting this during debugging.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AdditionalDesignTimeBuildInput.xaml
davkean: I've respecified the values, it would be good to know if these need to be respecified, and either way enforced. They have at least one additional rule, that ItemType needs to match the parent data source ItemType.

davkean:
We can use this to watch global.json and reevaluate.

PR #5008, opened 2019-07-09
Unify data sources

davkean:
This includes https://github.com/dotnet/project-system/pull/5006. so just review https://github.com/dotnet/project-system/commit/c457cbaff147a9e3f57125829b15ef39ac05f580 onwards.

This unifies the `<AdditionalDesignTimeBuildInput />` and `<PotentialEditorConfigFiles>` data sources into a single base class. The editor config did a few things that were unnecessarily:

- Lived in VS layer, I've pushed it down to core, nothing about editor config is VS specific.
- Had a unique key/version, even though its entirely based off evaluation results
- Used DefiningProjectDirectory/MakeRooted to root the file when it can be solely based on FullPath.
- Created an ActionBlock that transformed the value that pushed the result to a broadcast block. This could just be replaced with a transform.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractItemFileWatchDataSource.cs
```diff
@@ -0,0 +1,77 @@
...
+        protected override UnconfiguredProject ContainingProject
+        {
+            get { return _project.UnconfiguredProject; }
+        }
+
+        protected override IDisposable LinkExternalInput(ITargetBlock<IProjectVersionedValue<FileWatchData>> targetBlock)
+        {
+            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _projectSubscriptionService.ProjectRuleSource;
+
+            // Transform the changes from evaluation/design-time build -> restore data
```
davidwengier: This comment seems incorrect now (restore data)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractItemFileWatchDataSource.cs
```diff
@@ -0,0 +1,77 @@
...
+            DisposableValue<ISourceBlock<IProjectVersionedValue<FileWatchData>>> transformBlock = source.SourceBlock
+                                                                                                        .TransformWithNoDelta(update => update.Derive(u => CreateFileWatch(u.CurrentState)),
+                                                                                                            suppressVersionOnlyUpdates: true,
+                                                                                                            ruleNames: new[] { ItemRuleName });
+
+            // Set the link up so that we publish changes to target block
+            transformBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
```
davidwengier: Do you need to dispose the link returned here? I've noticed there is a lot of inconsistencies around this in general.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AdditionalDesignTimeBuildInputDataSource.cs
```diff
@@ -1,71 +1,26 @@
...
-
-            // Transform the changes from evaluation/design-time build -> restore data
-            DisposableValue<ISourceBlock<IProjectVersionedValue<FileWatchData>>> transformBlock = source.SourceBlock
-                                                                                                        .TransformWithNoDelta(update => update.Derive(u => CreateFileWatch(u.CurrentState)),
-                                                                                                            suppressVersionOnlyUpdates: true,
-                                                                                                            ruleNames: new[] { AdditionalDesignTimeBuildInput.SchemaName });
-
-            // Set the link up so that we publish changes to target block
-            transformBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+        public override string ItemRuleName => AdditionalDesignTimeBuildInput.SchemaName;
```
davidwengier: Perhaps this property should be renamed `SchemaName` for consistency?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractItemFileWatchDataSource.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            return transformBlock;
+        }
+
+        private FileWatchData CreateFileWatch(IImmutableDictionary<string, IProjectRuleSnapshot> update)
+        {
+            IProjectRuleSnapshot snapshot = update.GetSnapshotOrEmpty(ItemRuleName);
+
+            var fullPaths = snapshot.Items.Select(item => item.Value)
+                                          .Select(metadata => metadata.GetValueOrDefault(Compile.FullPathProperty))
```
davidwengier: Using `Compile` here is cheating. It would be better if this came from an abstract property, then new implementations are less likely to forget to add the `FullPath` property to their rule.

davidwengier: hmm.. maybe this isn't really here any more? I can see its different in #5006

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AdditionalDesignTimeBuildInputDataSource.cs
davkean: It's RuleName everywhere else in CPS, the rule itself is inconsistent.

davkean: But meh, I've updated it along with FullPath so its consistent with the rule.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractItemFileWatchDataSource.cs
davkean: You need to dispose the first link, it's a bit confusing...

Issue #5010, opened 2019-07-09
For WPF and Winforms .NETCore projects, versions lower than 3.0 should not be shown on the Application property page 


etbyrd:
**Summary**:
For .NETCore WPF and Winforms projects the Application property page Target Framework ComboBox should not show versions lower than 3.0 as they are not supported. 

**Expected Behavior**:
Only versions higher than 3.0 should be available

**Actual Behavior**:
All versions are shown

**User Impact**: 
Could cause confusion of supported frameworks

**Internal Issue**
[888249](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/888249)

etbyrd:
Fixed and tracked in: https://github.com/dotnet/sdk/issues/3438

PR #5011, opened 2019-07-09
Correctly show supported frameworks for .NETCore WPF and WinForms projects

etbyrd:
Fix for: https://github.com/dotnet/project-system/issues/5010

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
@@ -6,6 +6,7 @@ Imports System.Windows.Forms
 
 Imports Microsoft.VisualStudio.Editors.Common
 Imports Microsoft.VisualStudio.Shell.Interop
+Imports Microsoft.VisualStudio.Shell
 
 Imports NativeMethods = Microsoft.VisualStudio.Editors.Interop.NativeMethods
 
@@ -126,7 +127,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Try
                 Dim siteServiceProvider As OLE.Interop.IServiceProvider = Nothing
                 VSErrorHandler.ThrowOnFailure(ProjectHierarchy.GetSite(siteServiceProvider))
-                Dim sp As New Shell.ServiceProvider(siteServiceProvider)
+                Dim sp As New ServiceProvider(siteServiceProvider)
                 Dim vsFrameworkMultiTargeting As IVsFrameworkMultiTargeting = TryCast(sp.GetService(GetType(SVsFrameworkMultiTargeting).GUID), IVsFrameworkMultiTargeting)
                 ' TODO: Remove IsTargetFrameworksDefined check after issue #800 is resolved.
                 If (TargetFrameworksDefined() = False And vsFrameworkMultiTargeting IsNot Nothing) Then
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
```diff
@@ -126,15 +127,21 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
                 If (TargetFrameworksDefined() = False And vsFrameworkMultiTargeting IsNot Nothing) Then
 
                     Dim supportedTargetFrameworksDescriptor = GetPropertyDescriptor("SupportedTargetFrameworks")
                     Dim supportedFrameworks As IEnumerable(Of TargetFrameworkMoniker) = TargetFrameworkMoniker.GetSupportedTargetFrameworkMonikers(vsFrameworkMultiTargeting, DTEProject, supportedTargetFrameworksDescriptor)
 
+                    Dim isWPForWindowsForms = ProjectHierarchy.IsCapabilityMatch("WPF") Or ProjectHierarchy.IsCapabilityMatch("WindowsForms")
+
                     For Each supportedFramework As TargetFrameworkMoniker In supportedFrameworks
+                        Dim frameworkName As New FrameworkName(supportedFramework.Moniker)
+                        If isWPForWindowsForms And frameworkName.Version.Major < 3 And String.Equals(frameworkName.Identifier, ".NETCoreApp") Then
```
davidwengier: I don't know if the framework identifier is case sensitive, but its worth finding out and passing in the right StringComparer.

Also, using `AndAlso` instead of `And` will short circuit and be infinitesimally faster :)

davkean: @nguerrera @dsplaisted Can we please introduce handshake between us and the SDK so that we don't have business logic in the UI layer?

dsplaisted: Happy to work on the right handshake, but we need to understand what information the project system will need.  FWIW, I can retarget a .NET Framework WPF app to .NET Framework 2.0 in Visual Studio, so the equivalent check doesn't exist for .NET Framework and I'm not sure that we necessarily need it for .NET Core.

dsplaisted: Actually, .NET Framework 2.0 does appear in the dropdown, but when you try to retarget it fails with the following dialog:

> Attempted re-targeting of the project has been canceled. Required assemblies 'WindowsBase', 'PresentationCore', and 'PresentationFramework' are missing from the target framework.

nguerrera: The listis populated from items in the sdk, right. Should we exclude/remove as appropriate in sdk?

davkean: That check from legacy only runs in WPF flavor, but uses IVsFrameworkMultiTargeting to figure if those assemblies are present in the target framework. We need a data driven approach to avoid hardcoding these specific target frameworkism's in the project-system/UI layer, ie "Does this FrameworkReference exist in this TFM?"

davkean: Another approach, simialr to what we did we project references, is give up UI validation at all, and just have the build tell you that these references cannot be referenced in the target.

etbyrd: From a light investigation it seems that the supported TFM's are added [here](https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedTargetFrameworks.props) but I am not sure it is possible to determine if it is a WPF or WinForms project at that time, however

etbyrd: Created an issue here https://github.com/dotnet/sdk/issues/3438 to discuss this further

Issue #4538, opened 2019-02-05
Rename file from solution explorer might loose changes and undo doesn't work


vsfeedback:
Rename file from solution explorer might loose changes.

I noticed several times adding a new file, writing some code, then for whatever reason renaming the file, I end up with an empty class definition.

I just now do that with an existing file and choosing yes on the prompt to also rename the definition. It'll remove the change I just made. So this could also be the refactoring using an old version (from disk) to do the rename.

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/61984/rename-file-from-solution-explorer-might-loose-cha.html
VSTS ticketId: 441845_
_These are the original issue comments:_

Radames Cruz Moreno [MSFT] on 5/30/2017, 05:18 PM (615 days ago): <p>Thank you for your feedback! For us to investigate this further, could you please tell us:</p><ul><li>What type of project are you working on? Desktop application? Web application?</li><li>What programming language are you using?</li><li>How often does this happen?</li></ul><p><br></p><p>We look forward to hearing from you!</p>

[...]

Issue #2663, opened 2017-07-26
Renamer can null-reference


davkean:
.NET Crashes in: https://developercommunity.visualstudio.com/content/problem/67068/unit-tests-take-a-very-long-time-appear-in-the-tes.html

```
Exception Info: System.NullReferenceException
   at Microsoft.VisualStudio.ProjectSystem.VS.Renamer+<GetStrategyAsync>d__14.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at Microsoft.VisualStudio.ProjectSystem.VS.Renamer+<RenameAsync>d__13.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(System.Threading.Tasks.Task)
   at Microsoft.VisualStudio.ProjectSystem.VS.Renamer+<OnWorkspaceChangedAsync>d__12.MoveNext()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__6_0(System.Object)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(System.Delegate, System.Object, Int32)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(System.Object, System.Delegate, System.Object, Int32, System.Delegate)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(System.Object)
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
[...]

PR #5015, opened 2019-07-10
Add undo for rename type to match file

jmarolf:
fixes https://github.com/dotnet/project-system/issues/4538
fixes https://github.com/dotnet/project-system/issues/3527
fixes https://github.com/dotnet/project-system/issues/2663

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
@@ -21,7 +21,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
 {
     [Export(typeof(IFileRenameHandler))]
     [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -21,11 +21,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
 {
     [Export(typeof(IFileRenameHandler))]
     [AppliesTo(ProjectCapability.CSharpOrVisualBasic)]
-    internal sealed class Renamer : IFileRenameHandler
+    internal sealed partial class Renamer : IFileRenameHandler
     {
         private readonly IUnconfiguredProjectVsServices _projectVsServices;
         private readonly IUnconfiguredProjectTasksService _unconfiguredProjectTasksService;
         private readonly Workspace _workspace;
+        private readonly IVsService<Shell.Interop.SDTE, EnvDTE.DTE> _dte;
```
davidwengier: This doesn't need the `SDTE` bit, only the import needs both type arguments.

PR #5022, opened 2019-07-10
ItemType is allowed when SourceType != "Item"

### LINKED ISSUE #5007: Class item template instantiates with no namespace
**Visual Studio Version**: 16.2.0-pre.4.0+29104.9.d16.2

**Summary**:

Class item template instantiates with no namespace.

**Steps to Reproduce**:

1. Run the project system at master (16702928b28ed5704efdaae06811c52c6496f5bc)

2. Create a new .NET Core Console App

3. Right-click project, Add, Class...

4. Click _Add_

5. Class appears with no namespace:

```c#
using System;
using System.Collections.Generic;
using System.Text;

namespace
{
    class Class1
    {
    }
}
```

drewnoakes:
Reverts part of #5000.

Fixes #5007 and broken property pages.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/XamlRuleTests.cs
@@ -195,11 +195,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Rules
                     // An item type must be specified
                     Assert.NotNull(itemType);
                 }
-                else
-                {
-                    // Item type has no purpose on other source types
-                    Assert.Null(itemType);
-                }
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AnalyzerReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AnalyzerReference.xaml
@@ -7,6 +7,7 @@
       xmlns="http://schemas.microsoft.com/build/2009/properties">
   <Rule.DataSource>
     <DataSource HasConfigurationCondition="False"
+                ItemType="Analyzer"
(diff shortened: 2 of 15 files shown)


drewnoakes:
> Should it be marked as required for design time build rules?

I'm not sure how we would identify these. Naming convention?

davkean:
When MSBuildTarget is non-null.

drewnoakes:
> When MSBuildTarget is non-null

Done in f097151b8f172c1e0471324c6627b9f87a8ac19a.

drewnoakes:
@davkean can you help me parse [this doc comment](https://github.com/microsoft/msbuild/blob/0cd0e92a7243088977d31b56626b56d6116de016/src/Framework/XamlTypes/DataSource.cs#L119-L124) on `DataSource.ItemType`?

> If a `BaseProperty` is an item definition metadata or item metadata, this field specified the item type of the item definition or the item, respectively. For common properties field must not be set.

This makes a distinction between item definition metadata/item metadata and common properties. I don't know the difference here. Perhaps it's helpful for a rule on `ItemType` usage.

davkean:
I think its referring to ConfigurationGeneral, and other properties that are at the project/config level.

davidwengier:
> I'd like to understand why it is ok to omit ItemType for UpToDateCheckBuilt, UpToDateCheckInput and UpToDateCheckOutput, but not for the other rules.

For the specific regression we just had: Because they're not in the BrowseObject context, and the code in question that cared about having an ItemType only looked at the BrowseObject context.

I'm not sure there is an objective rule you can put in place around ItemType usage because it depends on the consumers of the rules. If nobody uses a rule in a scenario where ItemType is important, then ItemType isn't important. It's also not specific to BrowseObject, or even rule contexts at all. As Dave pointed out offline, ItemType is important in [ReferencesHostBridge](http://index/?query=ReferencesHostBridge&rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Package%5CReferencesHostBridge.cs&line=38) and it is explicitly loading rules by name. So for rules that ReferencesHostBridge loads by name, ItemType is important. Its difficult to make an objective rule out of that.

PR #5032, opened 2019-07-11
Remove the need for .NET Core 2.1 to be installed

davkean:
16.3 includes 3.0 by default which we automatically work against, but there's no 'Or' concept in .vsconfig so we cannot represent that.


davkean:
@dotnet/project-system I'm going to merge this in today, tell me if you have any feedback.

drewnoakes:
We only specify VS2019 in the [getting started](https://github.com/dotnet/project-system/blob/master/docs/repo/getting-started.md) docs. If we require 16.3, we should update those docs.

davkean:
We don't require 16.3, we just have no way of saying "2.0" or "3.0".

PR #5034, opened 2019-07-11
Update analyzers to the latest versions

davkean:


src/Microsoft.VisualStudio.AppDesigner/PublicAPI.Shipped.txt
```diff
@@ -351,6 +351,22 @@ Microsoft.VisualStudio.Editors.Constants.New() -> Void
 Microsoft.VisualStudio.Editors.IVBPackage
 Microsoft.VisualStudio.Editors.IVBPackage.GetLastShownApplicationDesignerTab(projectHierarchy As Microsoft.VisualStudio.Shell.Interop.IVsHierarchy) -> Integer
 Microsoft.VisualStudio.Editors.IVBPackage.SetLastShownApplicationDesignerTab(projectHierarchy As Microsoft.VisualStudio.Shell.Interop.IVsHierarchy, tab As Integer) -> Void
+Microsoft.VisualStudio.Editors.MyApplication.IVsMyApplicationProperties.AuthenticationMode() -> Integer
```
davkean: This was a bug fix in the public API analyzer.

build/Versions.props
```diff
@@ -21,8 +21,6 @@
     <MicrosoftDotNetIBCMergeVersion>4.7.1-alpha-00001</MicrosoftDotNetIBCMergeVersion>
     <MicrosoftNetCompilersVersion>3.3.0-beta1-19327-04</MicrosoftNetCompilersVersion>
     <MicrosoftDiaSymReaderPdb2PdbVersion>1.1.0-beta1-62624-01</MicrosoftDiaSymReaderPdb2PdbVersion>
-    <RoslynAnalyzersPackagesVersion>2.9.3</RoslynAnalyzersPackagesVersion>
```
davkean: These properties are supposed to be used when we use them in targets files and project files, I've just inlined the versions similar to other package references.

src/Microsoft.VisualStudio.AppDesigner/PublicAPI.Shipped.txt
davidwengier: The old "forget VB exists" bug? :)

Issue #5051, opened 2019-07-14
DependencyIDs create a lot of long, duplicate strings


panopticoncentral:
I don't have exact numbers, but looking at the GC heap of a large CPS solution, there are a lot of duplicate strings showing up that are DependencyIDs. The problem is that DependencyIDs tend to be very long and are generated on the fly, so common child IDs end up wasting a lot of space. (In a solution with 520 projects, one common dependency ID is duplicated 4k times and wastes nearly 1Mb all on its own.)

We should find a way to intern these so that we're not duplicating them all over the place.

davkean:
Do we need to create these ids at all? @drewnoakes Do you know what we use them for?

panopticoncentral:
A fair question... The big waste is the fact that Dependency objects use the IDs to connect parent/child Dependencies. If there's no reason to have things be disconnected, direct references would be much more efficient.

drewnoakes:
It may be possible to avoid creating them altogether for graph nodes (which allow composite IDs). I took a stab at that a long while back but didn't quite get it working. I'm more familiar with the code now so may have better results. I can take a look, probably in the next sprint.

davkean:
I would imagine that's where the majory of duplicated strings would be created, but I do wonder how they are being created if they are created on demand.

drewnoakes:
This situation has been improved significantly in #6155. Pending task is to inspect new heaps and see whether this is still an issue.

drewnoakes:
> The big waste is the fact that Dependency objects use the IDs to connect parent/child Dependencies. 

Since #6155 this is no longer the case.

Analysis of recent dumps shows that dependency IDs are no longer contributing in a significant fashion to heap consumption.

panopticoncentral:
So awesome! Thanks so much for your hard work on this!

drewnoakes:
#6276 further reduces the size of the dependency IDs by removing the TFM prefix from all of them.

drewnoakes:
#6282 eliminates composite dependency IDs altogether.

PR #5052, opened 2019-07-15
Intern dependencyid string

jmarolf:
potential fix for https://github.com/dotnet/project-system/issues/5051

This uses the same strategy for string interning that roslyn does in the parser.

A few notes:

1. checks if the `char` array present in the `PooledStringBuilder` has already been interned, this should be the main source of improvements ([here](https://github.com/jmarolf/project-system/blob/d8432c42b42f3a9cc34baffe31dc2a343dce0b1c/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTable.cs#L195))
2. Uses our pooled objects class so if every thread needs to intern a string we will have enough objects without needing to lock.  Using this approach, under heavy contention some strings will be interned multiple times. May re-write this to just use a single global `StringTable` that is held under lock.  Will need to investigate perf via ETL trace.
3. ~~The set of interned strings is never cleared.  I should probably add an implicitly active component that releases all strings once no .NET projects are loaded in VS.~~ fixed this in https://github.com/dotnet/project-system/pull/5052/commits/4805b75161df4c1067850e07bd01adf0d91ebd15

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/PooledStringBuilder.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/PooledStringBuilder.cs
@@ -52,6 +52,15 @@ namespace Microsoft.VisualStudio.Buffers.PooledObjects
             return result;
         }
 
+        public string InternStringAndFree()
+        {
+            var st = StringTable.GetInstance();
+            string result = st.Add(_builder);
+            Free();
+
+            return result;
+        }
+
         // global pool
         private static readonly ObjectPool<PooledStringBuilder> s_poolInstance = CreatePool();
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTable.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTable.cs
@@ -85,8 +85,6 @@ namespace Microsoft.VisualStudio.Buffers.PooledObjects
         public void Free()
         {
             // leave cache content in the cache, just return it to the pool
-            // Array.Clear(this.localTable, 0, this.localTable.Length);
-            // Array.Clear(sharedTable, 0, sharedTable.Length);
 
             _pool?.Free(this);
         }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTable.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTable.cs
@@ -1,0 +1,880 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using System.Threading;
+
+namespace Microsoft.VisualStudio.Buffers.PooledObjects
+{
+    internal class StringTable
+    {
+        // entry in the caches
+        private struct Entry
+        {
+            // hash code of the entry
+            public int HashCode;
+
+            // full text of the item
+            public string Text;
+        }
+
+        // Size of local cache.
+        private const int LocalSizeBits = 11;
+        private const int LocalSize = (1 << LocalSizeBits);
+        private const int LocalSizeMask = LocalSize - 1;
+
+        // max size of shared cache.
+        private const int SharedSizeBits = 16;
+        private const int SharedSize = (1 << SharedSizeBits);
+        private const int SharedSizeMask = SharedSize - 1;
+
+        // size of bucket in shared cache. (local cache has bucket size 1).
+        private const int SharedBucketBits = 4;
+        private const int SharedBucketSize = (1 << SharedBucketBits);
+        private const int SharedBucketSizeMask = SharedBucketSize - 1;
+
+        // local (L1) cache
+        // simple fast and not threadsafe cache 
+        // with limited size and "last add wins" expiration policy
+        //
+        // The main purpose of the local cache is to use in long lived
+        // single threaded operations with lots of locality (like parsing).
+        // Local cache is smaller (and thus faster) and is not affected
+        // by cache misses on other threads.
+        private readonly Entry[] _localTable = new Entry[LocalSize];
+
+        // shared (L2) threadsafe cache
+        // slightly slower than local cache
+        // we read this cache when having a miss in local cache
+        // writes to local cache will update shared cache as well.
+        private static readonly Entry[] s_sharedTable = new Entry[SharedSize];
+
+        // essentially a random number 
+        // the usage pattern will randomly use and increment this
+        // the counter is not static to avoid interlocked operations and cross-thread traffic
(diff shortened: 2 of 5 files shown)


drewnoakes:
I'd like to review this tomorrow.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTableLifetimeService.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.ProjectSystem;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.Buffers.PooledObjects
+{
+    /// <summary>
```
davkean: Can you explain what this is trying to acheive? Are you seeing that we're not sharing a lot of strings between configurations?

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/PooledStringBuilder.cs
```diff
@@ -52,6 +52,15 @@ public string ToStringAndFree(int startIndex, int length)
             return result;
         }
 
+        public string InternStringAndFree()
+        {
+            var st = StringTable.GetInstance();
```
davkean: We grab a string table and never release it?

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTable.cs
```diff
@@ -0,0 +1,882 @@
...
+        }
+
+        public void Free()
+        {
+            _pool?.Free(this);
+        }
+
+        public void ClearAll()
+        {
+            s_staticPool = CreatePool();
```
davkean: Can you help me understand the relationship between an instance of "StringTable" and the `static pool` that its clearing here?

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTableLifetimeService.cs
davkean: Also, this object lives per implicitly active configuration.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTableLifetimeService.cs
```diff
@@ -0,0 +1,68 @@
...
+            return _instance;
+        }
+
+        internal class Instance : IMultiLifetimeInstance
+        {
+            public StringTable? StringTable { get; private set; }
+
+            public Task InitializeAsync()
+            {
+                StringTable = StringTable.GetInstance();
```
davkean: I don't understand why it grabs this StringTable. No one appears to be using this instance at all, except to clear the static pool below.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTableLifetimeService.cs
panopticoncentral: Not sure if you're asking a general or specific question here. The general observation is that a lot of the space in Dependency objects is wasted because Dependency IDs are so long and generated on the fly rather than centralized (so we don't have much of any sharing). I haven't reviewed this PR yet in detail, it might be overkill since it would be a huge improvement just to intern within a project (I think in some cases I was looking at, there were 8 copies of a particular Dependency ID per project.)

drewnoakes:
`StringPool` appears to be designed for use where consumers allocate an instance of the pool per hardware thread and don't share across threads. This consumption does not enforce that.

The [FNV hash function](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) is designed to work with 8-bit characters. Any non-ASCII strings are not pooled at all. Given part of the ID (model ID) comes from the original item spec, it's possible that non-Latin customers would not have the same performance characteristics.

The hash table implementation uses a fixed number of buckets and quadratic probing. As it reaches capacity it starts evicting tenants at random.

The dependency ID strings are created by concatenating three different strings. With this pool (or one like it), we allocate the string, look it up, then release a reference to the newly created string if its unused. With more control over the hashing we could avoid allocating the temporary (similar to `Dependency.IdEquals`).

I've been developing [an object pool](https://github.com/drewnoakes/string-theory/blob/pool-samples/StringTheory.PoolExamples/WeakPool.cs) based upon `WeakReference<T>` and was contemplating using it here. It'd avoid having to manually track pool lifetimes as the GC is still able to clean up unreferenced strings. I haven't yet determined whether the overhead of weak references is suitable.

Do you have any data to validate the performance of the approach in this PR?

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/PooledStringBuilder.cs
drewnoakes: Agree that this looks incorrect.

panopticoncentral:
@drewnoakes I've been thinking it would be good to have a weak interning table type available. Did you look at how `ConditionalWeakTable` handles the collection problem?

drewnoakes:
@panopticoncentral I looked at consuming `ConditionalWeakTable` and decided it wasn't a good fit, but I didn't look at the implementation. Thanks for the suggestion.

It's using dependent handles rather than weak handles to create a dependency between the lifetimes of keys and values, even if values have strong references back to keys. It seems that managing dependent handles requires much more work by the runtime than weak references, though I'm no expert here.

Internally the hashing structure in [`ConditionalWeakTable<TKey,TValue>`](https://github.com/dotnet/corefx/blob/1acaa2017cfc64ef1aabcf3b19d0e097af692616/src/Common/src/CoreLib/System/Runtime/CompilerServices/ConditionalWeakTable.cs) is different to what I currently have in [`WeakPool<T>`](https://github.com/drewnoakes/string-theory/blob/pool-samples/StringTheory.PoolExamples/WeakPool.cs). CWT avoids heap allocating hash table chain objects, using a single struct array instead with 'next' indicies (similar to `HashSet<T>` and `Dictionary<TKey,TValue>`). I'll look into this, and it might have nicer performance characteristics, but it won't help if having many weak references on the heap is an issue.

src/Microsoft.VisualStudio.ProjectSystem.Managed/PooledObjects/StringTableLifetimeService.cs
davkean: I was referring to this class exactly. I understand what we're trying to solve with the interning.

jmarolf:
I had wanted to get some perf traces of before/after here before closing this but I think we want an entirely different approach here.  @drewnoakes 	has pointed out most of the issues with this approach.  Roslyn takes a speed over memory approach and can constrain the types of characters they need to worry about due to the set of valid uni-code characters that can be identifiers.

Issue #4606, opened 2019-02-22
add search keywords for .NET Core options in VS Tools -> Options


jimmylewis:
**Visual Studio Version**: VS2019 Preview 4 (dogfood)

**Summary:**
When searching in Tools -> Options for the settings on the `Projects and Solutions -> .NET Core` page, none of the options appear based on keyword searches.

**Steps to Reproduce**:

1. Open Tools -> Options
2. Search for any of these terms related to the options on this page: msbuild, project, logging, previews, sdk, etc.

![image](https://user-images.githubusercontent.com/3976973/53273952-6a767880-36aa-11e9-86a5-5dfea2e02da9.png)

**Expected Behavior**:
Appropriate keywords should find this option page from search

**Actual Behavior**:
None of these work.  Searching for .net or core works because they match the category name, but there's no related keywords associated.

**User Impact**:
Harder to find these options, especially if the user doesn't already know about or where this specific page is in the tree view.

etbyrd:
We don't own the search dialog, but I am investigating if there is a way for us to add keywords or something similar to make it easier to search for those options.

Pilchie:
There is - it's an attribute on the page registration indicating a resource.  See the `keywordList` parameter [here](http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices.CSharp/CSharpPackage.cs,64)

etbyrd:
Fixed in https://github.com/dotnet/project-system/pull/5054

PR #5054, opened 2019-07-15
Add option page keywords

etbyrd:
Fix for https://github.com/dotnet/project-system/issues/4606

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
+++ b/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
@@ -283,4 +283,4 @@
 [$RootKey$\ToolsOptionsPages\Projects\NETCore]
 "Package"="{67909B06-91E9-4F3E-AB50-495046BE9A9A}"
 "Page"="{2E6DB64B-DA09-4B9F-A334-37A86FECDA6A}"
-@="#1500"
+"Keywords"="#1600"
--- a/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
+++ b/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
@@ -281,6 +281,7 @@
 [$RootKey$\ToolsOptionsPages\Projects]
 
 [$RootKey$\ToolsOptionsPages\Projects\NETCore]
+@="#1500"
 "Package"="{67909B06-91E9-4F3E-AB50-495046BE9A9A}"
 "Page"="{2E6DB64B-DA09-4B9F-A334-37A86FECDA6A}"
 "Keywords"="#1600"
--- a/src/Microsoft.VisualStudio.Editors/VSPackage.resx
+++ b/src/Microsoft.VisualStudio.Editors/VSPackage.resx
@@ -161,4 +161,14 @@
     <value>.NET Core</value>
     <comment>Used for the Tools|Options|Projects and Solutions page.</comment>
   </data>
+  <data name="1600" xml:space="preserve">
+    <value>up to date checks; 
+Don't call MSBuild if a project appears to be up to date; 
+logging level; 
+futd; 
+incremental; 
+up-to-date; 
+build;</value>
+    <comment>.NET Core SDK-Style Projects option page keywords</comment>
+  </data>
 </root>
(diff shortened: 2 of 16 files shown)


etbyrd:
I've been trying to test this by using `devenv /updateconfiguration` but I still can't get the changes to deploy

src/Microsoft.VisualStudio.Editors/VSPackage.resx
```diff
@@ -120,4 +120,8 @@
   <data name="1001" xml:space="preserve">
     <value>.NET Core</value>
   </data>
+  <data name="1002" xml:space="preserve">
+    <value>up to date checks; Don't call MSBuild if a project appears to be up to date; logging level; futd; incremental; up-to-date;</value>
```
drewnoakes: Maybe add "build"?

```suggestion
    <value>up to date checks; Don't call MSBuild if a project appears to be up to date; logging level; futd; incremental; up-to-date;build</value>
```

davkean:
F5 should work for this - is it not?

davkean:
I had a look, I don't think that attribute registration actually being used which is why its probably not working, I think the registration is here: https://github.com/dotnet/project-system/blob/df940c8971c15554e1a5cfb90c60fdb4073c5528/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef#L287. The attribute is just a friend way of writing the latter. We've run into this before - perhaps we need to delete the attribute to avoid this confusion? @panopticoncentral you added both of these, do you remember what happened here?

etbyrd:
That may be the case. It is also worth noting that I can F5 changes to Roslyn's option pages:
https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/CSharp/Impl/CSharpPackage.cs#L54

src/Microsoft.VisualStudio.Editors/VSPackage.resx
etbyrd: Added!

etbyrd:
Requires https://github.com/dotnet/project-system/pull/5069 and updated with an unused resource ID

Also note that https://github.com/dotnet/project-system/pull/5071 will not require an update to the search keywords because the title of the tab is always a search keyword

src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
```diff
@@ -26,7 +26,7 @@ Namespace Microsoft.VisualStudio.Editors
     '* In the future, we should consider moving to a RegPkg.exe model
     <
         Guid("67909B06-91E9-4F3E-AB50-495046BE9A9A"),
-        ProvideOptionPage(GetType(GeneralOptionPage), "Projects and Solutions", ".NET Core", 1400, 1400, True),
+        ProvideOptionPage(GetType(GeneralOptionPage), "Projects and Solutions", ".NET Core", 1400, 1400, True, 1600),
```
etbyrd: This actually wasn't required because we have auto generating our .pkgdef turned off, but I added it just to clear any future confusion

etbyrd: I really wish we could get that enabled like Roslyn does, it makes it easier - but we only have one option page so it isn't much of a priority

etbyrd: @davkean do you know why we don't generate out pkgdef?

src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
```diff
@@ -283,4 +283,4 @@
 [$RootKey$\ToolsOptionsPages\Projects\NETCore]
 "Package"="{67909B06-91E9-4F3E-AB50-495046BE9A9A}"
 "Page"="{2E6DB64B-DA09-4B9F-A334-37A86FECDA6A}"
-@="#1500"
```
davkean: Please add this back, this is where we get the name of the page from.

src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
davkean: Please sync with my changes and avoid changing this - this only adds confusion as its not used. This project hasn't been setup to auto-gen the pkgdef, there's a bug on it.

PR #5057, opened 2019-07-16
Add MyType property

### LINKED ISSUE #5046: BrowseObject needs to implement VBProjectProperties3
For MyApplication support (part of #4249) our BrowseObject needs to support VBProjectProperties3, or the property pages need to allow reading the `MyType` property without requiring a cast to that interface.

davidwengier:
For the VB MyApplication model we need to support the `MyType` property from our browse object. This is currently retrieved by the property pages via the `VBProjectProperties3` interface, which CPS doesn't implement, so this PR adds a fallback to use `ICustomTypeDescriptor`.

Legacy only uses `VBProjectProperties3` for the `MyType` and `MyApplication` properties, and I couldn't find any other consumers, so actually doing the work to allow CPS to implement various project properties interfaces is overkill at the moment (The `MyApplication` property support was already added for #4708 and just needs to be turned on for CPS projects when ready).

Fixes #5046

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb
+++ b/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb
@@ -1397,6 +1397,20 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
                     Return MyApplicationProperties.ApplicationTypeFromOutputType(
                                 CUInt(props3.OutputType),
                                 props3.MyType) = ApplicationTypes.WindowsApp
+                Else
+                    ' CPS projects return DynamicTypeBrowseObject for its BrowseObject, which doesn't implement VBProjectProperties3 (or anything else) so we just try to get these properties by ICustomTypeDescriptor
+                    Dim customTypeProvider As ICustomTypeDescriptor = TryCast(obj, ICustomTypeDescriptor)
+                    If customTypeProvider IsNot Nothing Then
+                        Dim properties As PropertyDescriptorCollection = customTypeProvider.GetProperties()
+                        Dim myTypeProperty As PropertyDescriptor = properties.Find(NameOf(props3.MyType), ignoreCase:=False)
+                        Dim outputTypeProperty As PropertyDescriptor = properties.Find(NameOf(props3.OutputType), ignoreCase:=False)
+
+                        If myTypeProperty IsNot Nothing AndAlso outputTypeProperty IsNot Nothing Then
+                            Return MyApplicationProperties.ApplicationTypeFromOutputType(
+                                CUInt(outputTypeProperty.GetValue(obj)),
+                                CType(myTypeProperty.GetValue(obj), String)) = ApplicationTypes.WindowsApp
+                        End If
+                    End If
                 End If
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb
```diff
@@ -1397,6 +1397,20 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
                     Return MyApplicationProperties.ApplicationTypeFromOutputType(
                                 CUInt(props3.OutputType),
                                 props3.MyType) = ApplicationTypes.WindowsApp
+                Else
+                    ' CPS projects return DynamicTypeBrowseObject for its BrowseObject, which doesn't implement VBProjectProperties3 (or anything else) so we just try to get these properties by ICustomTypeDescriptor
```
davkean: This is a roundabout way of reading `DTE.Project.Properties`.

davidwengier: Yes, but its also a static method that only has the IVSHierarchy to act on. I can change the callers to pass in the DTE Project, or get the VSHPROPID_ExtObject property and cast it to `EnvDTE.Project` if you think it would be better, perhaps for reasons I'm not aware of, but based purely on code complexity it comes out about the same in my eyes.

davkean: Folks in the future are going to read this code and going to try and understand why its reading the properties via this very obscure way that only works for CPS, there is no guaranteed handshake between CPS and consumers that it implements ICustomTypeDescriptor, only that its implements IDispatch.

davkean: I would even consider just always using Project.Properties and remove the CPS-only/legacy-only paths.

davidwengier: I guess that makes sense, I essentially did that with the `MyApplication` property anyway.

davidwengier: Found some helper methods, the new code is much neater. Thank you :D

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb
```diff
@@ -1388,15 +1388,20 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+                ' Query the MyType and OutputType properties from the project properties
+                Dim project As Project = EnvDTEProject(Hierarchy)
+
+                If project Is Nothing Then
+                    Return False
                 End If
-                Dim props3 As VSLangProj80.VBProjectProperties3 = TryCast(obj, VSLangProj80.VBProjectProperties3)
-                If props3 IsNot Nothing Then
+
+                Dim myTypeProperty As [Property] = GetProjectProperty(project, NameOf(VSLangProj80.VBProjectProperties3.MyType))
```
davkean: This method is pretty inefficient, it walks every property, when it can just use the string indexer of the Properties instance.

davidwengier: Logged #5060

Issue #4736, opened 2019-04-19
*.CoreCompileInputs.cache causes projects to rebuild unnecessarily until the third build


Treit:
**Visual Studio Version**:
VS 2019 16.0.1

**Summary:**
SDK-style projects, even a trivial .NET Core project, build the entire project graph both the first and second time the project is built, even when nothing has changed between the first and second build.

While for trivial projects this is not a huge issue, for large projects with many dependencies this means that you can waste many minutes waiting for the project to build itself and all dependencies a second time.

**Steps to Reproduce**:

1. Create a new .NET Core console application project in Visual Studio 2019.

2. From the 'Build' menu, choose 'Clean Solution' to ensure you are starting from a clean state.

3. From the 'Build' menu, choose 'Build Solution' or press F6.

4. Note the 'output' Window says that it rebuilt the project, with the output:
========== Build: 1 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========

5. From the 'Build' menu, choose 'Build Solution' or press F6 for a second time.

6. Note that, once again, the project was rebuilt and the output window says:
========== Build: 1 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========

7. From the 'Build' menu, choose 'Build Solution' or press F6 for a third time.

8. Note that the output window says:
========== Build: 0 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========

**Expected Behavior**:
The project should not be rebuilt the second time, since nothing changed from the previous build. On the second build command, the output window should say:
========== Build: 0 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========

**Actual Behavior**:
The project rebuilds everything on the second build. On the second build command, the output window says:
========== Build: 1 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========

**User Impact**:
For large projects and for scenarios where you do something like change a single unit test after the initial build and then try to rebuild just the unit test project, you end up building the entire project graph unnecessarily. This can be very expensive, wasting the time of the user and leading to a significant loss in productivity.

**Notes**:
The rebuild is triggered by a "CoreCompileInputs.cache" file for the project. If you enable "FastUpToDate" logging under Tools | Options | Projects and Solutions | .NET Core you can see output like the following on the second build:

[...]

PR #5059, opened 2019-07-16
Up-to-date check uses immutable snapshot

drewnoakes:
Introduces the immutable `BuildUpToDateCheck.State` class to hold state relating to the fast up-to-date check.

This is preparatory work for #4736, which I have a local fix for that builds upon this work. Unlike this PR, that fix will change the unit tests for the up to date check.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -86,12 +86,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                 CopyReferenceInputs = copyReferenceInputs;
             }
 
-            public State Update(IProjectVersionedValue<Tuple<IProjectSubscriptionUpdate, IProjectSubscriptionUpdate, IProjectItemSchema>> e, out bool itemsChanged)
+            public State Update(
+                IProjectSubscriptionUpdate jointRuleUpdate,
+                IProjectSubscriptionUpdate sourceItemsUpdate,
+                IProjectItemSchema projectItemSchema,
+                IComparable configuredProjectVersion,
+                out bool itemsChanged)
             {
-                IProjectSubscriptionUpdate jointRuleUpdate = e.Value.Item1;
-                IProjectSubscriptionUpdate sourceItemsUpdate = e.Value.Item2;
-                IProjectItemSchema projectItemSchema = e.Value.Item3;
-
                 bool isDisabled = jointRuleUpdate.CurrentState.IsPropertyTrue(ConfigurationGeneral.SchemaName, ConfigurationGeneral.DisableFastUpToDateCheckProperty, defaultValue: false);
 
                 string msBuildProjectFullPath = jointRuleUpdate.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildProjectFullPathProperty, MSBuildProjectFullPath);
@@ -162,6 +163,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                         {
                             copyReferenceInputsBuilder.Add(item[CopyUpToDateMarker.SchemaName]);
                         }
+
                         if (!string.IsNullOrWhiteSpace(item[ResolvedCompilationReference.OriginalPathProperty]))
                         {
                             copyReferenceInputsBuilder.Add(item[ResolvedCompilationReference.OriginalPathProperty]);
@@ -238,15 +240,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                     itemsChanged = true;
                 }
 
-                IComparable lastVersionSeen = e.DataSourceVersions[ProjectDataSources.ConfiguredProjectVersion];
-
                 return new State(
                     msBuildProjectFullPath,
                     msBuildProjectDirectory,
                     markerFile,
                     outputRelativeOrFullPath,
                     newestImportInput,
-                    lastVersionSeen,
+                    lastVersionSeen: configuredProjectVersion,
                     isDisabled,
                     itemTypes,
                     itemsBuilder.ToImmutable(),
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -14,6 +14,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
     {
         private sealed class State
         {
+            public static State Empty { get; } = new State();
+
             public string? MSBuildProjectFullPath { get; }
             public string? MSBuildProjectDirectory { get; }
             public string? MarkerFile { get; }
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -142,7 +108,7 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
                 _configuredProject.Services.ProjectSubscription.JointRuleSource.SourceBlock.SyncLinkOptions(DataflowOption.WithRuleNames(ProjectPropertiesSchemas)),
                 _configuredProject.Services.ProjectSubscription.SourceItemsRuleSource.SourceBlock.SyncLinkOptions(),
                 _projectItemSchemaService.SourceBlock.SyncLinkOptions(),
-                target: DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<Tuple<IProjectSubscriptionUpdate, IProjectSubscriptionUpdate, IProjectItemSchema>>>(OnChangedAsync),
+                target: DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<Tuple<IProjectSubscriptionUpdate, IProjectSubscriptionUpdate, IProjectItemSchema>>>(OnChanged),
```
drewnoakes: The handler now works with snapshots so needn't take a lock in order to update the reference. Dataflow guarantees handler invocations won't overlap.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -56,32 +54,14 @@ internal sealed class BuildUpToDateCheck : OnceInitializedOnceDisposedUnderLockA
         private readonly ITelemetryService _telemetryService;
         private readonly IFileSystem _fileSystem;
 
+        private State _state = State.CreateEmpty();
+
         private IDisposable? _link;
-        private IComparable? _lastVersionSeen;
 
-        private bool _isDisabled = true;
         private bool _itemsChangedSinceLastCheck = true;
```
drewnoakes: This is one of two remaining fields that are mutated by `IsUpToDateAsync`. The fix for #4736 will remove this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -56,32 +54,14 @@ internal sealed class BuildUpToDateCheck : OnceInitializedOnceDisposedUnderLockA
...
-        private string? _msBuildProjectFullPath;
-        private string? _msBuildProjectDirectory;
-        private string? _markerFile;
-        private string? _outputRelativeOrFullPath;
-        private string? _newestImportInput;
 
         internal DateTime LastCheckTimeUtc { get; private set; } = DateTime.MinValue;
```
drewnoakes: This is the other remaining fields that is mutated by `IsUpToDateAsync`. I am not sure it's needed and may try to remove it in future.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -735,18 +581,20 @@ async Task<bool> IsUpToDateInternalAsync(CancellationToken token)
 
                 try
                 {
-                    if (!CheckGlobalConditions(buildAction, logger))
+                    State state = _state;
```
drewnoakes: Here we copy a reference to the state and pass it around on the stack, so it doesn't matter if Dataflow triggers an update to the state while we're calculating the result here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -0,0 +1,266 @@
...
+                BuiltOutputs = builtOutputs;
+                CopiedOutputFiles = copiedOutputFiles;
+                AnalyzerReferences = analyzerReferences;
+                CompilationReferences = compilationReferences;
+                CopyReferenceInputs = copyReferenceInputs;
+            }
+
+            public static State CreateEmpty()
+            {
+                return new State();
```
davidwengier: How many of these are created? Could just be a static readonly Empty instance.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -0,0 +1,266 @@
...
+            }
+
+            public State Update(IProjectVersionedValue<Tuple<IProjectSubscriptionUpdate, IProjectSubscriptionUpdate, IProjectItemSchema>> e, out bool itemsChanged)
```
davkean: Has this behavior changed at all or effectively copy and paste?

davidwengier: Very subjective, but might be nicer for this method to take the three items from the tuple, and the version, as separate parameters, just for nicety sake.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
davkean: With the lock removed, does this introduce a race for these?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
drewnoakes: It's effectively copy paste, however it works with immutable collections now rather than mutable ones.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
drewnoakes: Not many. Only on creation and on unload. I'll make an `Empty` property though.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
drewnoakes: I've broken out the args in other places too. Agree it looks nicer. Will do that here too.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: The `IsUpToDateAsync` check is still protected by a lock. Only the project subscription handler is unprotected now (though Dataflow prevents overlap).

There's as much of a race now as there was before for `_itemsChangedSinceLastCheck`.

This field is going away in the next few days anyway. Instead we'll have `LastItemChangedAtUtc` on the state snapshot.

That will leave just one mutable field, `LastCheckTimeUtc`, which is only read/written within the lock in `IsUpToDateAsync`.

PR #5071, opened 2019-07-17
Rename .NET Core options page

davkean:
This includes https://github.com/dotnet/project-system/pull/5069, so just review https://github.com/dotnet/project-system/pull/5071/commits/c940c5585702a574e405a740d1042d3498b80710

This renames ".NET Core" options page to "SDK-Style Projects". 

I considered:

.NET SDK Projects -> Ruled out because we call it "SDK-Style" on Projects and Solutions -> General.
.NET SDK-Style Projects -> Ruled out because it was too long and looked weird.

image

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
+++ b/src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
@@ -24,10 +24,14 @@ Namespace Microsoft.VisualStudio.Editors
     '*   from here for the moment.
     '*
     '* In the future, we should consider moving to a RegPkg.exe model
-    <
-        Guid("67909B06-91E9-4F3E-AB50-495046BE9A9A"),
-        ProvideOptionPage(GetType(GeneralOptionPage), "Projects and Solutions", ".NET Core", 1400, 1400, True),
-        CLSCompliant(False)
+    <Guid("67909B06-91E9-4F3E-AB50-495046BE9A9A"),
+    ProvideOptionPage(GetType(GeneralOptionPage),
+                     "Projects and Solutions",
+                     "SDK-Style Projects",
+                     0,        ' categoryResourceID: Not used, we don't own parent category
+                     0,        ' pageNameResourceID: Not used, set in Microsoft.VisualStudio.Editors.pkgdef, 
+                     True),    ' supportsAutomation
+    CLSCompliant(False)
     >
     Friend Class VBPackage
         Inherits Shell.Package
--- a/src/Microsoft.VisualStudio.Editors/Menus.vsct
+++ b/src/Microsoft.VisualStudio.Editors/Menus.vsct
@@ -1,0 +1,943 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- 
+  Useful references:
+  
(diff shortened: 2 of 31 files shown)


src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
```diff
@@ -24,10 +24,14 @@ Namespace Microsoft.VisualStudio.Editors
...
-        Guid("67909B06-91E9-4F3E-AB50-495046BE9A9A"),
-        ProvideOptionPage(GetType(GeneralOptionPage), "Projects and Solutions", ".NET Core", 1400, 1400, True),
-        CLSCompliant(False)
+    <Guid("67909B06-91E9-4F3E-AB50-495046BE9A9A"),
+    ProvideOptionPage(GetType(GeneralOptionPage),
+                     "Projects and Solutions",
+                     "SDK-Style Projects",
+                     0,        ' categoryResourceID: Not used, we don't own parent category
+                     0,        ' pageNameResourceID: Not used, set in Microsoft.VisualStudio.Editors.pkgdef, 
+                     True),    ' supportsAutomation
```
drewnoakes: Named arguments might be better here, assuming they work on attribute constructors in VB. e.g. `categoryResourceID := 0`.

davkean: I couldn't figure out VB's named argument syntax, and it hit a compiler bug that I've not yet filed. Do you know what the syntax is? I thought it was `categoryResourceId::=` but compiler does not like that.

davkean:
What's the documentation work? Do we refer to this somewhere?

drewnoakes:
We at least refer to it here: https://github.com/dotnet/project-system/blob/master/docs/repo/up-to-date-check.md#debugging

Looking around online there are quite a few references to the ".NET Core" page, but they're all directing users to the "Use preview versions" checkbox which isn't there anymore anyway. I likely overestimated the documentation work, which is always good.

src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
jmarolf: it is `:=` but you can't use that here.  Because VB is case-insensitive and `ProvideOptionPageAttribute` has public properties that only differ from its constructor argument names by case it is ambiguous.

dmonroym:
Could we just name this page "Project System" instead? Whatever else we decide to add there in the future would be relevant to our team

src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
davkean: Ah, that's probably what I hit and resulted in an extraordinary large number of errors.

davkean:
@dmonroym We can choose a more generalized name when we replace legacy, otherwise, I think it would only cause confusion with it.

drewnoakes:
This option should probably move to that page now:

image

However the last time we moved it we saw a large number of feedback items and questions on Stack Overflow etc about the fact it wasn't where some document said it was.

davidwengier:
lol that page is where that option started life. I don't think enough time has passed since the trauma for me to stomach another move :P

davkean:
Let us never speak of that suggestion again. :)

dmonroym:
Moving checkboxes... Never as simple as it seems :P

PR #5075, opened 2019-07-18
Remove resources

davkean:
This resource was a used by a now defunt options page that was used for simplified "options".

The registration is being removed here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/192167.

Also removed other resources that I've now confirmed are being used nowhere.


drewnoakes:
What's going on with the status on this PR?

PR #5077, opened 2019-07-18
Change integration test to use operation progress

davkean:
This changes the integration test to use operation progress instead of the NuGet restore service. This is a better indication of when the project system is done.


build/Packages.targets
```diff
@@ -122,7 +122,7 @@
     <PackageReference Update="Microsoft.Build.Tasks.Core" Version="16.0.461"/>
 
     <!-- Libraries -->
-    <PackageReference Update="Newtonsoft.Json" Version="9.0.1"/>
+    <PackageReference Update="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)"/>
```
davkean: This is needed as Apex has dependency on a higher version than VS ships.

drewnoakes:
I'm pretty sure the integration tests branch doesn't need those deleted files.

src/HostAgnostic.props
```diff
@@ -2,6 +2,7 @@
 <Project>
   <PropertyGroup>
     <BannedSymbolsOptOut Condition="$(IsTestProject)">true</BannedSymbolsOptOut>
+    <NewtonsoftJsonVersion>9.0.1</NewtonsoftJsonVersion>
```
davidwengier: As of 16.2, VS ships with 12.0.2 and we're supposed to update "at your convenience", so can we just set this to 12.0.2 and not need the special case for integration tests?

davkean:
To avoid getting out of sync, can we merge that branch back into mainsteam and just not turn them on?

src/HostAgnostic.props
davkean: Ah, good to know, I'll do a followup change. I need to make sure we don't have a handshake with other components here.

drewnoakes:
> can we merge that branch back into mainsteam and just not turn them on?

Fine with me. @jmarolf what do you think?

src/HostAgnostic.props
davkean: https://github.com/dotnet/project-system/pull/5079

Issue #4802, opened 2019-05-08
Remove double-writes in build


davkean:
We have the following [double-writes](https://github.com/KirillOsenkov/MSBuildStructuredLog/wiki/Double-write-detection) in our build, which means that we're dependent on different binaries in the tree.

We try very hard to avoid this: 
![image](https://user-images.githubusercontent.com/1103906/57399315-d9318100-7185-11e9-9b74-2f27b84706dd.png)

jmarolf:
@davkean how can we ensure that this fails the build when we regress this?

drewnoakes:
> how can we ensure that this fails the build when we regress this?

Filed https://github.com/microsoft/msbuild/issues/4524.

jmarolf:
Roslyn uses a tool called [BuildBoss](https://github.com/dotnet/roslyn/tree/master/src/Tools/BuildBoss) which reads the binlog after a build and fails if there is a double-write

PR #5080, opened 2019-07-18
Remove double-writes in the build

davkean:
VSSDK.DTE/VSSDK.TemplateWizardInterface are non-official Microsoft packages that NuGet.VisualStudio has taken a dependency on. These packages contain the same dlls as other packages. I've asked them to remove the dependency but in the meantime avoid using assets from those packages.

Fixes: https://github.com/dotnet/project-system/issues/4802.


drewnoakes:
Are these the only double writes reported in the binlog?

davkean:
Yes, double-writes are gone after above:

image

PR #5081, opened 2019-07-18
Don't fire a change when a new file is being watched.

davidwengier:
The compiler (which you haven't seen yet) makes a lot more sense when the file watcher just watches.


davidwengier:
Guess not :(

PR #5082, opened 2019-07-18
Create a TempPE Compiler Manager

davidwengier:
Contains #5081 so skip the first two commits (or don't they're really small).

This PR contains a class that manages the compilation of TempPE DLLs. It has two dataflow subscriptions:

* One for the design time inputs, and project information, which give it the state that it will be working with
* One for file changes, so it knows when to recompile

The basic idea is that as a design time input is changed or added, the manager collects them in a queue, and then after a short delay it starts running through the whole queue on a background thread. As needed, when more files change, that work is stopped and a new processing task is scheduled. Additionally there is a facility for "I need this now!" which will cause a compile (presumably on the UI thread) if required, without interrupting the queue.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
@@ -212,21 +212,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 
         protected override async Task DisposeCoreAsync(bool initialized)
         {
-            if (initialized)
+            // This will stop our blocks taking any more input
+            _inputsActionBlock?.Complete();
+            _fileWatcherActionBlock?.Complete();
+
+            if (_inputsActionBlock != null)
             {
-                // This will stop our blocks taking any more input
-                _inputsActionBlock?.Complete();
-                _fileWatcherActionBlock?.Complete();
-
-                if (_inputsActionBlock != null)
-                {
-                    await Task.WhenAll(_inputsActionBlock.Completion, _fileWatcherActionBlock!.Completion);
-                }
-
-                // By waiting for completion we know that the following dispose will cancel any pending compilations, and there won't be any more
-                _scheduler.Dispose();
-                _disposables.Dispose();
+                await Task.WhenAll(_inputsActionBlock.Completion, _fileWatcherActionBlock!.Completion);
             }
+
+            // By waiting for completion we know that the following dispose will cancel any pending compilations, and there won't be any more
+            _scheduler.Dispose();
+            _disposables.Dispose();
         }
 
         private async Task ProcessCompileQueue(CancellationToken token)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
@@ -13,7 +13,6 @@ using System.Threading.Tasks.Dataflow;
 using Microsoft.VisualStudio.IO;
 using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
-using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.ProjectSystem.Utilities;
 using Microsoft.VisualStudio.Threading;
 using Microsoft.VisualStudio.Threading.Tasks;
@@ -205,13 +204,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 
         protected override async Task DisposeCoreAsync(bool initialized)
         {
-            // This will stop our blocks taking any more input
-            _inputsActionBlock?.Complete();
-            _fileWatcherActionBlock?.Complete();
-
             if (_inputsActionBlock != null)
             {
-                await Task.WhenAll(_inputsActionBlock.Completion, _fileWatcherActionBlock!.Completion);
+                // This will stop our blocks taking any more input
+                _inputsActionBlock.Complete();
+                _fileWatcherActionBlock!.Complete();
+
+                await Task.WhenAll(_inputsActionBlock.Completion, _fileWatcherActionBlock.Completion);
             }
 
             // By waiting for completion we know that the following dispose will cancel any pending compilations, and there won't be any more
@@ -252,8 +251,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
         /// <summary>
         /// Gets the XML that describes a TempPE DLL, including building it if necessary
         /// </summary>
-        /// <param name="fileName"></param>
-        /// <returns></returns>
+        /// <param name="fileName">A project relative path to a source file that is a design time input</param>
+        /// <returns>An XML description of the TempPE DLL for the specified file</returns>
         public async Task<string> GetDesignTimeInputXmlAsync(string fileName)
         {
             // Make sure we're not being asked to compile a random file
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
@@ -1,0 +1,372 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+
+using Microsoft.VisualStudio.IO;
+using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
+using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal class TempPECompilerManager : OnceInitializedOnceDisposedAsync
+    {
+        private static readonly TimeSpan s_compilationDelayTime = TimeSpan.FromMilliseconds(500);
+
+        private readonly ConfiguredProject _project;
+        private readonly IProjectSubscriptionService _projectSubscriptionService;
+        private readonly IActiveWorkspaceProjectContextHost _activeWorkspaceProjectContextHost;
+        private readonly IDesignTimeInputsDataSource _inputsDataSource;
+        private readonly IDesignTimeInputsFileWatcher _fileWatcher;
+        private readonly ITempPECompiler _compiler;
+        private readonly IFileSystem _fileSystem;
+        private readonly TaskDelayScheduler _scheduler;
+
+        private readonly DisposableBag _disposables = new DisposableBag();
+
+        private ImmutableArray<string> _designTimeInputs;
+        private ImmutableArray<string> _sharedDesignTimeInputs;
+        private ImmutableDictionary<string, bool> _filesToCompile = ImmutableDictionary<string, bool>.Empty.WithComparers(StringComparers.Paths); // Key is filename, value is whether to ignore the last write time check
+        private string? _outputPath;
+        private readonly TaskCompletionSource<bool> _receivedProjectRuleSource = new TaskCompletionSource<bool>();
+
+        private ITargetBlock<IProjectVersionedValue<Tuple<DesignTimeInputs, IProjectSubscriptionUpdate>>>? _inputsActionBlock;
+        private ITargetBlock<IProjectVersionedValue<string[]>>? _fileWatcherActionBlock;
+
+        [ImportingConstructor]
+        public TempPECompilerManager(ConfiguredProject project,
+                                     IProjectSubscriptionService projectSubscriptionService,
+                                     IActiveWorkspaceProjectContextHost activeWorkspaceProjectContextHost,
+                                     IProjectThreadingService threadingService,
+                                     IDesignTimeInputsDataSource inputsDataSource,
+                                     IDesignTimeInputsFileWatcher fileWatcher,
+                                     ITempPECompiler compiler,
+                                     IFileSystem fileSystem)
+            : base(threadingService.JoinableTaskContext)
+        {
+            _project = project;
+            _projectSubscriptionService = projectSubscriptionService;
+            _activeWorkspaceProjectContextHost = activeWorkspaceProjectContextHost;
+            _inputsDataSource = inputsDataSource;
+            _fileWatcher = fileWatcher;
+            _compiler = compiler;
+            _fileSystem = fileSystem;
+
+            _scheduler = new TaskDelayScheduler(s_compilationDelayTime, threadingService, CancellationToken.None);
+        }
+
+        protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
+        {
+            // Create an action block process the design time inputs and configuration general changes
+            _inputsActionBlock = DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<Tuple<DesignTimeInputs, IProjectSubscriptionUpdate>>>(ProcessDataflowChanges);
+
+            IDisposable projectLink = ProjectDataSources.SyncLinkTo(
+                   _inputsDataSource.SourceBlock.SyncLinkOptions(
+                       linkOptions: new StandardRuleDataflowLinkOptions
+                       {
+                           PropagateCompletion = true,
+                       }),
+                   _projectSubscriptionService.ProjectRuleSource.SourceBlock.SyncLinkOptions(
+                      linkOptions: new StandardRuleDataflowLinkOptions
+                      {
+                          RuleNames = Empty.OrdinalIgnoreCaseStringSet.Add(ConfigurationGeneral.SchemaName),
+                          PropagateCompletion = true,
+                      }),
+                   _inputsActionBlock,
+                   DataflowOption.PropagateCompletion,
+                   cancellationToken: _project.Services.ProjectAsynchronousTasks.UnloadCancellationToken);
+
+            // Create an action block to process file change notifications
+            _fileWatcherActionBlock = DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<string[]>>(ProcessFileChangeNotification);
+            IDisposable watcherLink = _fileWatcher.SourceBlock.LinkTo(_fileWatcherActionBlock, DataflowOption.PropagateCompletion);
+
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+            _disposables.AddDisposable(projectLink);
+            _disposables.AddDisposable(watcherLink);
+
+            return Task.CompletedTask;
+        }
+
+        internal void ProcessFileChangeNotification(IProjectVersionedValue<string[]> arg)
+        {
+            // Ignore any updates until we've received the first set of design time inputs (which shouldn't happen anyway)
+            // That first update will queue all of the files so we're not losing anything
```
davkean: Why should this never happen?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                    foreach (string file in inputs.Inputs.Except(_designTimeInputs, StringComparers.Paths))
+                    {
+                        ImmutableInterlocked.TryAdd(ref _filesToCompile, file, /* ignoreWriteTime */ false);
+                    }
+                }
+            }
+
+            // Make sure we have the up to date list of inputs
+            _designTimeInputs = inputs.Inputs;
+            _sharedDesignTimeInputs = inputs.SharedInputs;
```
davkean: These fields are not updated atomically which is a problem because ProcessFileChangeNotification and ProcessDataflowChanges can overlap, causing the former to see incomplete state.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+            // That first update will queue all of the files so we're not losing anything
+            if (_designTimeInputs == null)
+            {
+                return;
+            }
+
+            foreach (string changedFile in arg.Value)
+            {
+                // if a shared input changes, we recompile everything
+                if (_sharedDesignTimeInputs.Contains(changedFile))
```
davkean: This field can be changed underneath it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                {
+                    // A normal design time input, so just add it to the queue
+                    ImmutableInterlocked.TryAdd(ref _filesToCompile, changedFile, /* ignoreWriteTime */ false);
+                }
+            }
+
+            QueueCompilation();
+        }
+
+        internal void ProcessDataflowChanges(IProjectVersionedValue<Tuple<DesignTimeInputs, IProjectSubscriptionUpdate>> input)
```
davkean: I would consider having DesignTimeInputs push "full paths" from metadata rather than having this class "MakeRooted" on the result. On top of that, this code is de-duping the paths, but doing on a non-canonical path, which means that the following won't be de-dup'd:

```
<Foo Include="...\Foo.cs"/>
<Foo Include="C:\Project\Foo.cs"/>
```

The other added benefit is that the string will be shared for other consumers that use FullPath.

davkean: We should do the same everywhere - just have everyone work off FullPath so that we start sharing the strings.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                }
+            }
+
+            // Make sure we have the up to date list of inputs
+            _designTimeInputs = inputs.Inputs;
+            _sharedDesignTimeInputs = inputs.SharedInputs;
+
+            // Make sure we have the up to date output path
+            string basePath = configChanges.After.Properties[ConfigurationGeneral.ProjectDirProperty];
+            string objPath = configChanges.After.Properties[ConfigurationGeneral.IntermediateOutputPathProperty];
```
davkean: This path can be bad.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                _scheduler.ScheduleAsyncTask(ProcessCompileQueue, _project.Services.ProjectAsynchronousTasks.UnloadCancellationToken);
+            }
+        }
+
+        protected override async Task DisposeCoreAsync(bool initialized)
+        {
+            if (initialized)
+            {
+                // This will stop our blocks taking any more input
+                _inputsActionBlock?.Complete();
```
davkean: Can just move these into the null checks and skip the '?'

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
+using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal class TempPECompilerManager : OnceInitializedOnceDisposedAsync
```
davkean: Who actually "initializes" this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                throw new ArgumentException("Can only get XML snippets for design time inputs", nameof(fileName));
+            }
+
+            // Remove the file from our todo list, in case it was in there.
+            ThreadingTools.ApplyChangeOptimistically(ref _filesToCompile, fileName, (s, f) => s.Remove(f));
+
+            string outputFileName = await GetOutputFileName(fileName);
+            // make sure the file is up to date
+            await _activeWorkspaceProjectContextHost.OpenContextForWriteAsync(async accessor =>
+            {
```
davkean: You can remove "async" and just return the result of CompileDesignTimeInput to reduce the unneeded state machine.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        /// <summary>
+        /// Gets the XML that describes a TempPE DLL, including building it if necessary
+        /// </summary>
+        /// <param name="fileName"></param>
+        /// <returns></returns>
+        public async Task<string> GetDesignTimeInputXML(string fileName)
+        {
+            // Make sure we're not being asked to compile a random file
```
davkean: I think you wanted to initialize here?

How does the caller know when they can call this method?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                }
+            });
+        }
+
+        /// <summary>
+        /// Gets the XML that describes a TempPE DLL, including building it if necessary
+        /// </summary>
+        /// <param name="fileName"></param>
+        /// <returns></returns>
+        public async Task<string> GetDesignTimeInputXML(string fileName)
```
davkean: Doesn't need to be async.

davkean: GetDesignTimeInputXmlAsync.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+        private ImmutableArray<string> _sharedDesignTimeInputs;
+        private ImmutableDictionary<string, bool> _filesToCompile = ImmutableDictionary<string, bool>.Empty.WithComparers(StringComparers.Paths); // Key is filename, value is whether to ignore the last write time check
+        private string? _outputPath;
+        private readonly TaskCompletionSource<bool> _receivedProjectRuleSource = new TaskCompletionSource<bool>();
+
+        private ITargetBlock<IProjectVersionedValue<Tuple<DesignTimeInputs, IProjectSubscriptionUpdate>>>? _inputsActionBlock;
+        private ITargetBlock<IProjectVersionedValue<string[]>>? _fileWatcherActionBlock;
+
+        [ImportingConstructor]
+        public TempPECompilerManager(ConfiguredProject project,
```
davkean: As I was reading the PR, I found it a little hard to keep track of when the mutable state of this class is assigned and can be accessed. I would consider merging it into a single field representing your "current", that is assigned all at once atomically.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+            _inputsActionBlock = DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<Tuple<DesignTimeInputs, IProjectSubscriptionUpdate>>>(ProcessDataflowChanges);
+
+            IDisposable projectLink = ProjectDataSources.SyncLinkTo(
+                   _inputsDataSource.SourceBlock.SyncLinkOptions(
+                       linkOptions: new StandardRuleDataflowLinkOptions
+                       {
+                           PropagateCompletion = true,
+                       }),
+                   _projectSubscriptionService.ProjectRuleSource.SourceBlock.SyncLinkOptions(
+                      linkOptions: new StandardRuleDataflowLinkOptions
```
davkean: DataflowOption.WithRuleNames

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
+        {
+            // Create an action block process the design time inputs and configuration general changes
+            _inputsActionBlock = DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<Tuple<DesignTimeInputs, IProjectSubscriptionUpdate>>>(ProcessDataflowChanges);
+
+            IDisposable projectLink = ProjectDataSources.SyncLinkTo(
+                   _inputsDataSource.SourceBlock.SyncLinkOptions(
+                       linkOptions: new StandardRuleDataflowLinkOptions
```
davkean: DataflowOption.PropagateCompletion

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+    codebase = ""{Path.GetFileName(outputFileName)}""
+    name = ""{fileName}""
+    version = ""0.0.0.0""
+    snapshot_id = ""1""
+    replaceable = ""True""
+  />
+</root>";
+        }
+
+        private async Task CompileDesignTimeInput(IWorkspaceProjectContext context, string designTimeInput, string outputFileName, bool ignoreFileWriteTime, CancellationToken token = default)
```
davkean: "Async"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                _disposables.Dispose();
+            }
+        }
+
+        private async Task ProcessCompileQueue(CancellationToken token)
+        {
+            await _activeWorkspaceProjectContextHost.OpenContextForWriteAsync(async accessor =>
+            {
+                // Grab the next file to compile off the queue
+                (string fileName, bool ignoreFileWriteTime) = _filesToCompile.FirstOrDefault();
```
davkean: I think this method will be easier to read if you come up with a way to avoid the double-lookup of the collection, once here, and then once below when you "Remove" it .

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                    await Task.WhenAll(_inputsActionBlock.Completion, _fileWatcherActionBlock!.Completion);
+                }
+
+                // By waiting for completion we know that the following dispose will cancel any pending compilations, and there won't be any more
+                _scheduler.Dispose();
+                _disposables.Dispose();
+            }
+        }
+
+        private async Task ProcessCompileQueue(CancellationToken token)
```
davkean: Can remove two state machines by remove async /await completely form this method and its body.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+                    // Grab another file off the queue
+                    (fileName, ignoreFileWriteTime) = _filesToCompile.FirstOrDefault();
+                }
+            });
+        }
+
+        /// <summary>
+        /// Gets the XML that describes a TempPE DLL, including building it if necessary
+        /// </summary>
+        /// <param name="fileName"></param>
```
davkean: Remove the empty params/returns elements.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -0,0 +1,372 @@
...
+            // Make sure we're not being asked to compile a random file
+            if (!_designTimeInputs.Contains(fileName, StringComparers.Paths))
+            {
+                throw new ArgumentException("Can only get XML snippets for design time inputs", nameof(fileName));
+            }
+
+            // Remove the file from our todo list, in case it was in there.
+            ThreadingTools.ApplyChangeOptimistically(ref _filesToCompile, fileName, (s, f) => s.Remove(f));
+
+            string outputFileName = await GetOutputFileName(fileName);
```
davkean: This call will deadlock (in `await _receivedProjectRuleSource.Task;`) if the following is true:

1. This method is entered on the UI thread
2. The data sources for ProcessDataflowChanges need the UI thread to finish. 

In that case, the UI thread will not let step 2. proceed as it doesn't know that it needs to let the await on the UI thread to proceed.

davidwengier:
Only one comment about threading!! 🎉 🎉 🎉 


(though I would have never thought of that myself...)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
davidwengier: It _shouldn't_, but as you say in a below comment, the calls could overlap so a bit of extra protection didn't seem like a terrible idea.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
davidwengier: I'll have to think about this. At the moment compilation is the only thing in the chain that cares about full paths, so I'm not sure whether using full paths will just mean the other blocks needing to call `MakeRelative` all the time, eg to return the list of inputs, fire change events, subscribe to file watching etc. I think there would be more sharing happening now.

Your point about de-duping the wrong things is valid, though its worth pointing out that anything with a path like in your example will be considered as a linked file, so isn't a valid DesignTimeInput anyway. I need to think about whether there is a way to express the same file with different monikers and have them both be valid inputs. I also think the risk is pretty low, as duplicate files in the compilation queue mean after the first entry gets compiled, the second entry wouldn't pass the file write check (though granted, sometimes that is skipped).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
davidwengier: Nothing yet, I'm trying to do this in reasonable chunks, as the whole reason for the rewrite was that the previous version was too big to review.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
davidwengier: As above, this is still to come.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
davkean: >  though its worth pointing out that anything with a path like in your example will be considered as a linked file, 

That's only for files outside of the project cone. Regardless, you can write relative paths in many ways.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
davidwengier: I must be missing something, how do I remove two? Are you suggesting using a Forget for the actual compilation, so we start compiling in parallel?

davidwengier:
Have addressed all of the feedback that doesn't need further discussion.

davidwengier:
Merging this to unblock things. Further review/discussion welcome if anyone finds the time.

PR #5087, opened 2019-07-18
Move to Microsoft.NetCore.Analyzers

davkean:
Fixes: https://github.com/dotnet/project-system/issues/4377

This was a lot harder than expected and the value is questionable. We spend more time hitting bugs, suppressing warnings then actually fixing real analysis. 

- First all .NET analysis has been moved to a package confusioningly called Microsoft.NetCore.Analyzers which I originally thought was NET Core rules until I dug through the source code.  This meant these previous rules stopped running when we updated one of the other analyzers packages we had updated.
- Had to re-suppress analysis that we've already suppressed because the IDs changed when they moved
- Baselined the VB code for String.Empty, the analysis is wrong and the recommendation fix will end up introducing null references
- Ran into other random bugs with the rules and disabled or suppressed

tag @mavasani


mavasani:
> .NET analysis has been moved to a package confusioningly called Microsoft.NetCore.Analyzers which I originally thought was NET Core rules until I dug through the source code. This meant these previous rules stopped running when we updated one of the other analyzers packages

`Microsoft.NetCore.Analyzers` has existed as a part of `Microsoft.CodeAnalysis.FxCopAnalyzers` package for a few years now. The repo [readme](https://github.com/dotnet/roslyn-analyzers#the-following-are-subpackages-or-nuget-dependencies-that-are-automatically-installed-when-you-install-the-microsoftcodeanalysisfxcopanalyzers-package) documents the rules contained in each sub-package of `Microsoft.CodeAnalysis.FxCopAnalyzers` package. It is also expected that clients directly reference `Microsoft.CodeAnalysis.FxCopAnalyzers` package, unless there is a specific reason to choose individual sub-packages, such as they need different versions or wan't to explicitly choose a strict subset of the sub-packages, etc., none of which seem to be applicable based on the current package references in this repo.

> Had to re-suppress analysis that we've already suppressed because the IDs changed when they moved

Which rule IDs? The only ID changes happened a few years ago when bunch of RSxxxx rules original written in `Roslyn.Diagnostics.Analyzers` were found to be generic enough to be promoted to CAxxxx rules that would benefit all customers. Other then that, all the CA rule IDs should be identical to the legacy FxCop rule implementations.

> Baselined the VB code for String.Empty, the analysis is wrong and the recommended fix will end up introducing null references

I believe the analysis is correct, it is finding cases where you have a comparison with empty string. We have implemented one possible code fix, which is to directly compare the length to zero without a null check. Another potential fix that can be offered would be to have a null check followed by string compare. A fix all is still going to be tricky here as it is very likely you want a mix of these fixes applied based on prior null checks in the code.

> Ran into other random bugs with the rules and disabled or suppressed

Note that https://github.com/dotnet/roslyn-analyzers/issues/2683 that you filed has already been fixed, but we haven't yet release a new stable package with the fix. Were there any other bugs?

Issue #266, opened 2016-06-24
Codeanalysis property page.


srivatsn:

ruant:
Is this planned to be done?

prbans:
https://github.com/MicrosoftDocs/visualstudio-docs/issues/3129

> @mavasani when researching editorconfig, I wasn't able to find how to enforce it during CI or produce build time errors across all files. I was only able to detect errors on files that were currently open in VS2017. I believe the documentation also states that editorconfig does not support static code analysis. Do you know if either of these features will eventually make it to editorconfig files?
I believe that implementing dotnet/project-system#266 may help bridge the gap until editorconfig is ready

drewnoakes:
This was addressed in #5094.

We will port this experience to the new Project Properties UI, which is tracked in #7242.

PR #5094, opened 2019-07-19
Add Code Analysis property page with Roslyn Analyzers section

mavasani:
The UI implemented in this PR has been iterated over and approved by the UX team, see [this work item](https://devdiv-design.visualstudio.com/D3%20Studio/_workitems/edit/7294) for details.

image

image

This page includes a new Roslyn Analyzers section with the following controls:
1. Hyperlink for overview documentation on Roslyn analyzers
2. Check boxes to selectively disable analyzer execution during build and/or live analysis via project properties `RunRoslynAnalyzersDuringBuild` and `RunRoslynAnalyzersDuringLiveAnalysis`. Default values are 'true' for both of these. Addresses https://github.com/dotnet/roslyn/issues/23591. These properties will be respected once the targets change in https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/190487?path=%2Fsrc%2Fedev%2FStaticAnalysis%2Ffxcoptask%2FMicrosoft.CodeAnalysis.targets&_a=files makes it in.
3. Text box showing the installed FxCopAnalyzers NuGet package version, if any.
4. Following buttons:
   1. `Install`: Installs the latest stable FxCopAnalyzers NuGet package (currently 2.9.3).
   2. `...`: Opens NuGet package manager with "Microsoft.CodeAnalysis.FxCopAnalyzers" as the search string for custom installation.
   3. `Uninstall`: Uninstalls the currently installed FxCopAnalyzers NuGet package.
   4. `Refresh`: Refreshes the UI, in case user has done offline install/uninstall/upgrade of NuGet packages.

Addresses part of #266. Note that in very near future we plan to enhance this page with information about any project level editorconfig file and ability to create or view editorconfig file from this property page (similar to the ruleset selection and opening from Code Analysis page in legacy project system).

https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/190487 adds the corresponding section to the existing Code Analysis property page in the legacy project system.

Tagging @mikadumont who helped drive the UX review. Also tagging other folks involved in the design thread @jinujoseph @vatsalyaagrawal @jjmew @cartermp

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
@@ -80,8 +80,8 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Return HelpKeywords.VBProjPropAssemblyInfo
         End Function
 
-        Private Sub RefreshInstalledFxCopAnalyzersVersionAndButtons()
-            If Not TryGetInstalledFxCopAnalyzersVersion(_installedFxCopAnalyzersVersionString) Then
+        Private Sub RefreshInstalledFxCopAnalyzersVersionAndButtons(Optional restorePackages As Boolean = False)
+            If Not TryGetInstalledFxCopAnalyzersVersion(restorePackages, _installedFxCopAnalyzersVersionString) Then
                 RefreshInstallFxCopAnalyzersButtons(reset:=True)
                 Return
             End If
@@ -108,38 +108,40 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         End Function
 
         ' Returns False if attempt to GetInstalledPackages did not succeed
-        Private Function TryGetInstalledFxCopAnalyzersVersion(ByRef versionString As String) As Boolean
+        Private Function TryGetInstalledFxCopAnalyzersVersion(restorePackages As Boolean, ByRef versionString As String) As Boolean
             versionString = Nothing
 
             If DTEProject Is Nothing Then
                 Return False
             End If
 
-            Using session = _threadedWaitDialogFactory.StartWaitDialog(My.Resources.Strings.RestoringPackagesMessage)
-                _packageRestorer.RestorePackages(DTEProject)
+            If restorePackages Then
+                Using session = _threadedWaitDialogFactory.StartWaitDialog(My.Resources.Strings.RestoringPackagesMessage)
+                    _packageRestorer.RestorePackages(DTEProject)
 
-                session.UserCancellationToken.ThrowIfCancellationRequested()
+                    session.UserCancellationToken.ThrowIfCancellationRequested()
 
-                ' Add a small delay to ensure GetInstalledPackages invocation below gets updated packages.
-                Thread.Sleep(2000)
+                    ' Add a small delay to ensure GetInstalledPackages invocation below gets updated packages.
+                    Thread.Sleep(2000)
+                End Using
+            End If
 
-                Dim installedPackages As IEnumerable(Of IVsPackageMetadata)
-                Try
-                    installedPackages = _packageInstallerServices.GetInstalledPackages(DTEProject)
-                Catch
-                    Return False
-                End Try
+            Dim installedPackages As IEnumerable(Of IVsPackageMetadata)
+            Try
+                installedPackages = _packageInstallerServices.GetInstalledPackages(DTEProject)
+            Catch
+                Return False
+            End Try
 
-                For Each package As IVsPackageMetadata In installedPackages
-                    If String.Equals(package.Id, FxCopAnalyzersPackageId, StringComparison.OrdinalIgnoreCase) OrElse
-                       s_childPackageIds.Contains(package.Id) Then
-                        versionString = package.VersionString
-                        Exit For
-                    End If
-                Next
+            For Each package As IVsPackageMetadata In installedPackages
+                If String.Equals(package.Id, FxCopAnalyzersPackageId, StringComparison.OrdinalIgnoreCase) OrElse
+                   s_childPackageIds.Contains(package.Id) Then
+                    versionString = package.VersionString
+                    Exit For
+                End If
+            Next
 
-                Return True
-            End Using
+            Return True
         End Function
 
         Private Sub InstallLatestVersionButton_Click(sender As Object, e As EventArgs) Handles InstallLatestVersionButton.Click
@@ -151,7 +153,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                 _installedFxCopAnalyzersVersionString = Nothing
             End Try
 
-            RefreshInstalledFxCopAnalyzersVersionAndButtons()
+            RefreshInstalledFxCopAnalyzersVersionAndButtons(restorePackages:=True)
 
             If _installedFxCopAnalyzersVersionString Is Nothing Then
                 DesignerFramework.DesignUtil.ShowMessage(ServiceProvider, My.Resources.Strings.FxCopAnalyzersInstallFailedMessage, DesignerFramework.DesignUtil.GetDefaultCaption(ServiceProvider), MessageBoxButtons.OK, MessageBoxIcon.Error)
@@ -168,7 +170,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Catch
             End Try
 
-            RefreshInstalledFxCopAnalyzersVersionAndButtons()
+            RefreshInstalledFxCopAnalyzersVersionAndButtons(restorePackages:=True)
 
             If _installedFxCopAnalyzersVersionString IsNot Nothing Then
                 DesignerFramework.DesignUtil.ShowMessage(ServiceProvider, String.Format(My.Resources.Strings.FxCopAnalyzersUninstallFailedMessage, versionText), DesignerFramework.DesignUtil.GetDefaultCaption(ServiceProvider), MessageBoxButtons.OK, MessageBoxIcon.Error)
@@ -178,7 +180,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         Private Sub InstallCustomVersionButton_Click(sender As Object, e As EventArgs) Handles InstallCustomVersionButton.Click
             Debug.Assert(_nugetPackage IsNot Nothing)
 
-            ' Reset and disable all the buttons (except Refresh) as we are about to navigate
+            ' Reset and disable all the buttons as we are about to navigate
             ' away from this page to the NuGet Package manager.
             RefreshInstallFxCopAnalyzersButtons(reset:=True)
 
@@ -197,13 +199,17 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             task.Start()
         End Sub
 
-        Private Sub RefreshButton_Click(sender As Object, e As EventArgs) Handles RefreshButton.Click
-            RefreshInstalledFxCopAnalyzersVersionAndButtons()
-        End Sub
-
         Private Sub RoslynAnalyzersLabel_LinkClicked(sender As Object, e As LinkLabelLinkClickedEventArgs) Handles RoslynAnalyzersHelpLinkLabel.LinkClicked
             RoslynAnalyzersHelpLinkLabel.LinkVisited = True
             Process.Start(RoslynAnalyzersDocumentationLink)
+        End Sub
+
+        Protected Overrides Sub OnPageActivated(activated As Boolean)
+            MyBase.OnPageActivated(activated)
+
+            If activated Then
+                RefreshInstalledFxCopAnalyzersVersionAndButtons()
+            End If
         End Sub
 
         Private NotInheritable Class SearchQuery
--- a/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
@@ -1,0 +1,246 @@
+﻿' Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+Imports System.Diagnostics
+Imports System.Threading
+Imports System.Windows.Forms
+Imports Microsoft.VisualStudio.ComponentModelHost
+Imports Microsoft.VisualStudio.ProjectSystem
+Imports Microsoft.VisualStudio.Shell
+Imports Microsoft.VisualStudio.Shell.Interop
+Imports NuGet.VisualStudio
+
+Namespace Microsoft.VisualStudio.Editors.PropertyPages
+
+    Friend NotInheritable Class CodeAnalysisPropPage
+        Inherits PropPageUserControlBase
+
+        Private Const NugetFeed As String = "https://packages.nuget.org/api/v2"
+        Private Const FxCopAnalyzersPackageId As String = "Microsoft.CodeAnalysis.FxCopAnalyzers"
+
+        Private Shared ReadOnly s_latestStableVersion As Version = New Version(2, 9, 3)
+        Private Shared ReadOnly s_childPackageIds As HashSet(Of String) = New HashSet(Of String)(
+            {"Microsoft.CodeQuality.Analyzers", "Microsoft.NetCore.Analyzers", "Microsoft.NetFramework.Analyzers"},
+            StringComparer.OrdinalIgnoreCase)
+
+        Private _packageInstallerServices As IVsPackageInstallerServices
+        Private _packageInstaller As IVsPackageInstaller2
+        Private _packageUninstaller As IVsPackageUninstaller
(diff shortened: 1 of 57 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
```diff
@@ -389,4 +389,23 @@
     </StringProperty.DataSource>
   </StringProperty>
 
+  <BoolProperty Name="RunRoslynAnalyzersDuringBuild"
```
davidwengier: For easier future review, can we keep the properties in this file in alphabetical order please.

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+                Return False
+            End If
+
+            Using session = _threadedWaitDialogFactory.StartWaitDialog(My.Resources.Strings.RestoringPackagesMessage)
+                _packageRestorer.RestorePackages(DTEProject)
+
+                session.UserCancellationToken.ThrowIfCancellationRequested()
+
+                ' Add a small delay to ensure GetInstalledPackages invocation below gets updated packages.
+                Thread.Sleep(2000)
```
davidwengier: Rather than a thread sleep, can this method be made async, and have the below check implemented with a timeout? Presumably for a project that is already fully restored this is 2 seconds of waiting (on the UI thread?) for no real reason. It doesn't look like this method actually needs to return its value synchronously, it can just enable/disable buttons whenever its finished.

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+            shell?.LoadPackage(nugetGuid, _nugetPackage)
+
+            RefreshInstalledFxCopAnalyzersVersionAndButtons()
+        End Sub
+
+        Protected Overrides ReadOnly Property ControlData() As PropertyControlData()
+            Get
+                If m_ControlData Is Nothing Then
+                    m_ControlData = New PropertyControlData() {
+                        New PropertyControlData(1, "RunRoslynAnalyzersDuringBuild", RunRoslynAnalyzersDuringBuild, ControlDataFlags.None),
```
davidwengier: Is there a PR coming for the legacy project system to support these properties, or was it intended that this property page only apply to SDK-style projects?

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+
+            Dim componentModel = CType(ServiceProvider.GetService(GetType(SComponentModel)), IComponentModel)
+            _packageInstallerServices = componentModel.GetExtensions(Of IVsPackageInstallerServices)().SingleOrDefault()
+            _packageInstaller = componentModel.GetService(Of IVsPackageInstaller2)()
+            _packageUninstaller = componentModel.GetService(Of IVsPackageUninstaller)()
+            _packageRestorer = componentModel.GetService(Of IVsPackageRestorer)()
+            _threadedWaitDialogFactory = CType(ServiceProvider.GetService(GetType(SVsThreadedWaitDialogFactory)), IVsThreadedWaitDialogFactory)
+
+            Dim shell = CType(ServiceProvider.GetService(GetType(SVsShell)), IVsShell)
+            Dim nugetGuid = New Guid("5fcc8577-4feb-4d04-ad72-d6c629b083cc")
```
davidwengier: Nit: Would be good for all magic strings to be constants at the top of the file.

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+
+            ' We're able to launch the package manager (with an item in its search box) by
+            ' using the IVsSearchProvider API that the NuGet package exposes.
+            '
+            ' We get that interface for it and then pass it a SearchQuery that effectively
+            ' wraps the package name we're looking for. The NuGet package will then read
+            ' out that string and populate their search box with it.
+
+            Dim extensionProvider = CType(_nugetPackage, IVsPackageExtensionProvider)
+            Dim extensionGuid = New Guid("042C2B4B-C7F7-49DB-B7A2-402EB8DC7892")
```
davidwengier: Another magic string

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+        End Sub
+
+        Private Sub InstallCustomVersionButton_Click(sender As Object, e As EventArgs) Handles InstallCustomVersionButton.Click
+            Debug.Assert(_nugetPackage IsNot Nothing)
+
+            ' Reset and disable all the buttons (except Refresh) as we are about to navigate
+            ' away from this page to the NuGet Package manager.
+            RefreshInstallFxCopAnalyzersButtons(reset:=True)
+
+            ' We're able to launch the package manager (with an item in its search box) by
```
davidwengier: Neat trick 👍

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+            task.Start()
+        End Sub
+
+        Private Sub RefreshButton_Click(sender As Object, e As EventArgs) Handles RefreshButton.Click
+            RefreshInstalledFxCopAnalyzersVersionAndButtons()
+        End Sub
+
+        Private Sub RoslynAnalyzersLabel_LinkClicked(sender As Object, e As LinkLabelLinkClickedEventArgs) Handles RoslynAnalyzersHelpLinkLabel.LinkClicked
+            RoslynAnalyzersHelpLinkLabel.LinkVisited = True
+            Process.Start("https://docs.microsoft.com/visualstudio/code-quality/roslyn-analyzers-overview")
```
davidwengier: Another magic string

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+                Dim installedPackages As IEnumerable(Of IVsPackageMetadata)
+                Try
+                    installedPackages = _packageInstallerServices.GetInstalledPackages(DTEProject)
+                Catch
+                    Return False
+                End Try
+
+                For Each package As IVsPackageMetadata In installedPackages
+                    If String.Equals(package.Id, FxCopAnalyzersPackageId, StringComparison.OrdinalIgnoreCase) OrElse
+                       s_childPackageIds.Contains(package.Id, StringComparer.OrdinalIgnoreCase) Then
```
davidwengier: No need to specify a comparer here if its already specified on the hashset

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -0,0 +1,246 @@
...
+
+        Private Sub UninstallAnalyzersButton_Click(sender As Object, e As EventArgs) Handles UninstallAnalyzersButton.Click
+            Debug.Assert(DTEProject IsNot Nothing)
+            Debug.Assert(_installedFxCopAnalyzersVersionString IsNot Nothing)
+            Dim versionText = _installedFxCopAnalyzersVersionString
+
+            Try
+                _packageUninstaller.UninstallPackage(DTEProject, FxCopAnalyzersPackageId, removeDependencies:=True)
+            Catch
+            End Try
```
davidwengier: Would it be useful to display the error to the user, or some how log it, rather than just show a generic failure message, or is there no extra detail provided? Same question for installation failures I guess.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
davkean: "Roslyn" was deliberately not used in any API name as per original API review. I would consider calling this `RunAnalyzersDuringBuild` or just `RunAnalyzers`.

davkean:
Why does this page need to live in our respository? CPS has extension points for adding property pages.

davkean:
Property pages represent the live state of the project, when state changes underneath them - they refresh automatically.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
mavasani: > "Roslyn" was deliberately not used in any API name as per original API review

Yes, that is true. However, the analyzers have become popular and well known as `Roslyn analyzers`, so much so that even [docs pages](https://docs.microsoft.com/en-us/visualstudio/code-quality/code-analysis-for-managed-code-overview?view=vs-2019) uses the term `Roslyn analyzers` almost everywhere. During the UX review almost everyone unanimously felt that we should use `Roslyn analyzers` as that is the term that developer community relates to much more then any other term.

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
mavasani: Yes, hopefully soon. It turned out to be lot more non-trivial in the legacy project system then I anticipated (I looked at the logic for existing properties on Code Analysis property page). Probably there is an easier way that I am just not aware about...

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
mavasani: Yes, that sounds reasonable. I can make that change, thanks!

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
mavasani: Sure.

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
mavasani: I am hoping this would not happen often, if at all. I can include the message but that might clutter the dialog and may not be really helpful. We can probably wait for feedback on this one..

davkean:
RE: Refresh

I'm sure I follow why the background you've provided prevents us from using the live state of the project?

src/Microsoft.VisualStudio.Editors/PropPages/xlf/CodeAnalysisPropPage.de.xlf
```diff
@@ -0,0 +1,67 @@
...
+        <note />
+      </trans-unit>
+      <trans-unit id="RefreshButton.Text">
+        <source>&amp;Refresh</source>
+        <target state="new">&amp;Refresh</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="RoslynAnalyzersLineLabel.Text">
+        <source>Label1</source>
+        <target state="new">Label1</target>
```
davkean: This does not have to be localized.

davkean:
Maybe I don't understand what Refresh does? I'm making the assumption that its indicating the version of the package installed?

mavasani:
> Maybe I don't understand what Refresh does? I'm making the assumption that its indicating the version of the package installed?

Sorry, I thought your concern was around having the install/uninstall buttons. Refresh is primarily for the scenario where the user clicks the browse (`...`) button to navigate away from the property page and open up NuGet package manager to install/update/uninstall the analyzer package and then comes back to the property page. Current implementation disables all the other buttons (install/uninstall/browse) when user clicks on browse, and clicking on Refresh invokes the NuGet VS API to get installed packages and refresh the installed version and enable the appropriate buttons. Let me know if there is a better/faster way to fetch installed NuGet packages for a project, so we can auto refresh whenever the property page gets focus.

src/Microsoft.VisualStudio.Editors/PropPages/xlf/CodeAnalysisPropPage.de.xlf
mavasani: The resx file for the property page seems to be auto-generated by the designer. I can remove this resx entry manually, but it may again be auto-generated in future.

tmeschter:
With the exception of the .xaml file changes to pick up the properties you need, I believe all of this could just as well be in the dotnet/roslyn repo. We _do_ have a lot of property pages in dotnet/project-system, but that's more an accident of history than anything else--we shouldn't own all the ones we have, and we don't want to end up owning another.

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
mavasani: @davidwengier I gave this a shot but the attempt to set the `Text` field below does not succeed even if we switch back to the UI thread:

```vb
         Task.Run(
                Function()
                    ' Add a small delay to ensure GetInstalledPackages invocation below gets updated packages.
                    Thread.Sleep(2000)

                    Dim installedPackages As IEnumerable(Of IVsPackageMetadata)
                    Try
                        installedPackages = _packageInstallerServices.GetInstalledPackages(DTEProject)
                    Catch
                        RefreshInstallFxCopAnalyzersButtons(reset:=True)
                        Return Tasks.Task.CompletedTask
                    End Try

                    _installedFxCopAnalyzersVersionString = Nothing
                    For Each package As IVsPackageMetadata In installedPackages
                        If String.Equals(package.Id, FxCopAnalyzersPackageId, StringComparison.OrdinalIgnoreCase) OrElse
                           s_childPackageIds.Contains(package.Id) Then
                            _installedFxCopAnalyzersVersionString = package.VersionString
                            Exit For
                        End If
                    Next

                    ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync()
                    ' Below update does not succeed and the line after that to refresh button states is never executed
                    InstalledVersionTextBox.Text = If(_installedFxCopAnalyzersVersionString, My.Resources.Strings.NotInstalledText)
                    RefreshInstallFxCopAnalyzersButtons(reset:=False)
                End Function)
        End Sub
```

I am going to stick with the current approach (note that the delay is inside a threaded wait dialog so the current UI experience is not bad).

mavasani:
@davkean - I was able to get rid of the Refresh button and use auto-refresh with https://github.com/dotnet/project-system/pull/5094/commits/4b9986e7f5bf98b1e51dda752799a7f11858035f. Hope that addresses your concern.

@tmeschter - I have added you to an internal thread on this. The Roslyn IDE/analysis team will own this page even in this repo and is primarily being added here as seemed to fell under the same core property pages bucket that includes Build, Package, Debug, etc and we felt it will be easier to maintain here along with other property pages.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
mavasani: Based on the offline discussions, we have decided to not use "Roslyn" in the property pages as well as the documentation. Fixed with https://github.com/dotnet/project-system/pull/5094/commits/5173428a2f3a0d48a376664a90153061bb5f7b90

mavasani:
Ping.. any further feedback?

etbyrd:
Is there a way to force a tab selection on the Nuget Package Manager when you create the search query? It defaults to the last tab visited or on a new project will default to the "Installed" tab. If there is a way to always start at the "Browse" tab if the package is not installed that would be ideal - but it is a minor concern.

I am not a huge fan of Debug.Assert but that is just a personal preference. 

Everything looks good to me. Could you also add me to that internal thread?

mavasani:
> Is there a way to force a tab selection on the Nuget Package Manager when you create the search query? It defaults to the last tab visited or on a new project will default to the "Installed" tab. If there is a way to always start at the "Browse" tab if the package is not installed that would be ideal - but it is a minor concern.

I am not aware about such a feature, but someone from the NuGet team should likely be the ones to ask. Let me follow-up on a separate offline thread with them.

> Everything looks good to me. Could you also add me to that internal thread?

Thanks, will do.

By the way, I do not have merge permissions on this repo, so someone else would have to merge the PR on my behalf whenever we are fine with it.

jjmew:
> Why does this page need to live in our respository? CPS has extension points for adding property pages.

From a private conversation we decided this was okay.

PR #5100, opened 2019-07-20
Enable inline test failures during build

davkean:
This brings back inline test failures that we used to have before we adopted RoslynTools.RepoToolset:

## Before
image

## After
image


davidwengier:
> before we adopted RoslynTools.RepoToolset

Wouldn't it be better to fix roslyn-tools?

davkean:
The build toolset in RoslynTools.RepoToolset is dead and replaced with Arcade. I don't want to spend any energy in making above work in its complicated handling of multi-architecture/multi-TFM when we're basically the only folks left using it.

Issue #5092, opened 2019-07-19
Integration Tests TODO:


davkean:
- [x] Test run deployment issues ("Failed to get the file for deployment item 'Context.xml': System.IO.FileNotFoundException: Could not find file 'C:\project-system\artifacts\Debug\bin\IntegrationTests\Context.xml', etc)
- [x] Tests don't run twice in a row (An item with the specified name C:\project-system\artifacts\Debug\TestResults\IntegrationTestTemp already exists)
- [x] VS hive keeps getting reset after running
- [x] VS crashes badly during runs
- [x] RunSettings in ~temp~ intermediate?
- [X] ~create suite.json file (Upload results to AzureDevOps (see #2c594a96)~ Not used anymore.
- [x] Why is xUnit Test Adapter running? (IntegrationTests probably inherits xunit console runner from RoslynToolset)
- [x] Failed to create the Test Contract deployment directory 'C:\Program Files (x86)\Microsoft Visual Studio\2019\master\Common7\IDE\CommonExtensions\TestContracts' in all test runs
- [x] Include `-optimize` flag (per https://github.com/dotnet/project-system/pull/5118/files#r306073290)
- [x] MEF errors when a test fails about trying to resolve attributes from System.Runtime
- [x] Errors around failing to elevate the process for codemarkers and failuring to find ElevateClient.
- [x] Video recording (Apex already does screenshots)
- [x] Validate double writes again (seems xunit is triggering this)
- [x] Fix hive name setting (support both env var and settings file)
- [x] `/diagnostic` option increases log verbosity
- [x] Run a single configuration for integration tests
- [x] Merge and delete `feature/integration-tests` branch
- [x] Fix VS shutdown so that it only shuts down after the last test
- [x] Lower msbuild verbosity on `/diagnostic` once we're happy
- [x] Update VSSDK version (fixes hang due to `/Embedding` flag) #5186
- [x] File bug for broken TRX viewer in dev16.x
- [x] Add `/no-build` option (PR #5190)
- [x] Tests can run from a normal (non-VS) command prompt
- [x] Move `.vsts-ci.yml` into `build` folder and update signed build configuration
- [x] Honour the `RunningUnderCI` property
- [x] Change SetVSEnvironment to limit to 16 once Spanish in on 16.0 (blocked by https://github.com/dotnet/project-system/issues/5176)
- [ ] Suppress first-start dialog (it blocks tests from making progress)
[...]

PR #5102, opened 2019-07-20
New integration test infrastructure

davkean:
This is the initial start of the new integration test infrastructure from myself and @drewnoakes, biggest changes from previous:

- Logic has been moved from build.ps1 -> msbuild
- We reuse VSIX deployment that occurs during the build, where as previously we'd uninstall/reinstall the VSIXes
- The logic has been simplified, less copying files around, no more "mstest deployment" - everything just runs from output location and results are output directly to where we want them.
- The integration test logic is tied to a project, similar to how build/test already works. It makes the logic much simpler as we can reuse all the evaluation state of the build.
- Integration tests can now be run multiple times in a row 
- A new VS instance no longer crashes if you run them while VS is open
- VS is no longer torn down while if you run them while it is open
~- Settings are no longer thrown away after they are run~ This still seems to be occuring
- The output is much quieter with less noise being output to the console.

The current state of these tests are ready for "personal run" but not CI runs. CI run logic is being tracked by : https://github.com/dotnet/project-system/issues/5092.

Questions for @jmarolf:

- What is suite.json used for? See commit #2c594a96.
- Why do we need to convert TRX -> xUnit?


jmarolf:
> What is suite.json used for? See commit #2c594a96.

this file is needed if you want to insert these tests as a nupkg at VS insertion time and run them against the VS you are insertion against.  This would allow other teams in VS to run our integration tests. For example, if we ran our tests on all PR in VS and if an SDK insertion would break project system the insertion PR would fail. That was the idea at least. This infrastructure on the VS side appears to have changed and this file is no longer needed? [here](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki?pagePath=%2FEngineering%20System%20%26%20Tools%2FTest%20Docs%2FIntegration%20Tests%2FTest%20Run%20Authoring%2FTools%2FTest%20Deployment&pageId=884&wikiVersion=GBwikiMaster) is the latest documentation on how test assets need to be organized.

jmarolf:
> Why do we need to convert TRX -> xUnit?

Once upon a time there was a desire to track test result statistics across all of github.com/dotnet. From the latest meetings it looks like this tracking requires arcade now so there is no point.

davkean:
We're merging this in, this state is good for personal integration test execution.

Devs can run this via:

```
build /integrationtests
```
-or-
```
Microsoft.VisualStudio.ProjectSystem.IntegrationTests>msbuild /t:IntegrationTest
```

PR #5112, opened 2019-07-22
Use dedicated comparer for IDependencyModel

drewnoakes:
`IDependencyModel` is an extensibility point and we should not be requiring implementors to add correct `Equals`/`GetHashCode` methods. There's no good way to express this on the interface, but we don't want the implementation to vary based on provider anyway.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/IDependencyModelEqualityComparer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/IDependencyModelEqualityComparer.cs
@@ -9,10 +9,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
     {
         public static IDependencyModelEqualityComparer Instance { get; } = new IDependencyModelEqualityComparer();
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/IDependencyModelEqualityComparer.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
+{
+    internal sealed class IDependencyModelEqualityComparer : IEqualityComparer<IDependencyModel>
+    {
+        public static IDependencyModelEqualityComparer Instance { get; } = new IDependencyModelEqualityComparer();
+
+        public bool Equals(IDependencyModel x, IDependencyModel y)
+        {
+            Requires.NotNull(x, nameof(x));
```
davidwengier: Aren't Equals methods supposed to deal with nulls gracefully?

drewnoakes: True. It'll never happen in current usages, but I've updated it all the same.

PR #5113, opened 2019-07-22
A few small bits pulled off old branches/stashes

drewnoakes:


src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
```diff
@@ -1,3 +1,4 @@
+; TODO generate this file https://github.com/dotnet/project-system/issues/68
```
davkean: This is the comment syntax?

drewnoakes: From https://github.com/madskristensen/PkgdefLanguage:

image

PR #5111, opened 2019-07-22
Use build.cmd for integration tests

drewnoakes:

PR #5114, opened 2019-07-22
Fix restore failure

davkean:
These package references are never used, and/or given a version which breaks restore.

This caused one of the failures in https://github.com/dotnet/project-system/pull/5111.

### DIFF
--- a/build/Toolset.proj
+++ b/build/Toolset.proj
@@ -10,12 +10,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="RoslynTools.RepoToolset" />
-    <PackageReference Include="RoslynTools.VsixExpInstaller" />
     <PackageReference Include="Codecov" />
     <PackageReference Include="OpenCover" />
     <PackageReference Include="Microsoft.DiaSymReader.Pdb2Pdb" />
-    <PackageReference Include="Microsoft.DevDiv.Validation.MediaRecorder" />
-    <PackageReference Include="Microsoft.DevDiv.Validation.Logging.ScreenshotCollector" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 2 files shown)


build/Toolset.proj
```diff
@@ -10,12 +10,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="RoslynTools.RepoToolset" />
-    <PackageReference Include="RoslynTools.VsixExpInstaller" />
     <PackageReference Include="Codecov" />
     <PackageReference Include="OpenCover" />
     <PackageReference Include="Microsoft.DiaSymReader.Pdb2Pdb" />
-    <PackageReference Include="Microsoft.DevDiv.Validation.MediaRecorder" />
```
jmarolf: Note: without these you will not have any screenshots or videos output on CI.  I have no idea how you will debug CI in these cases

jmarolf: See the temp artifacts directory on this [build](https://dev.azure.com/dnceng/public/_build/results?buildId=274165)

davkean: These were removed previously - see https://github.com/dotnet/project-system/issues/5092 that is tracking readding them.

This is what we found with both of these existing ones:

1) Screenshots are already taken by Apex or one of its installed plugins. The screen shot taken by the above collector never actually captured the screen at the failure.
2) We couldn't actually get media recorder to work at all, we plan in debugging why it wasn't working and readding it.

davkean: Thanks, I do see a screen recording, could never getting that working will retry tomorrow. 

In that folder you'll see some screenshots with "Fail.[randomchars].png" that is the built-in/already installed screenshot take, the OnTestCaseFail.png is the one above. We felt it added nothing (in fact in some cases worse) then the other one so we've removed it.

jmarolf: alright, well it was recording video as recently as that build I pointed out

davkean: Does screen recording happen on your box? It doesn't on mine - I wonder if its machine specific, I did check and it seems to bail on dual screens - so wonder if bails on other cases.

jmarolf: I am perfectly alright with not having the additional screen shots

PR #5117, opened 2019-07-22
Cleanup references

davkean:
- Remove unneeded reference
- Promote "implicit" package refs
   - These packages are ones contained in RoslynTools.RepoToolset package and require our repo to double-restore before it has downloaded all the packages. This one of the causes that prevents from Visual Studio opening a non-restored repository and "just working". Promote these to real references which also gives us a good indication of what our repository depends on.
- Cleanup package sources


build/Packages.targets
```diff
@@ -13,19 +13,26 @@
       https://dotnet.myget.org/F/project-system-interop/api/v3/index.json;
       https://vside.myget.org/F/vs-impl/api/v3/index.json;
       https://vside.myget.org/F/vssdk/api/v3/index.json;
+      https://api.nuget.org/v3/index.json;
```
jmarolf: why remove this?

build/Packages.targets
```diff
@@ -13,19 +13,26 @@
...
+    <PackageReference Update="XliffTasks"                                                             Version="$(XliffTasksVersion)" />
     <PackageReference Update="RoslynDependencies.ProjectSystem.OptimizationData"                      Version="$(RoslynDependenciesProjectSystemOptimizationDataVersion)" />
     <PackageReference Update="Microsoft.DotNet.IBCMerge"                                              Version="[$(MicrosoftDotNetIBCMergeVersion)]" />
     <PackageReference Update="Microsoft.DiaSymReader.Pdb2Pdb"                                         Version="$(MicrosoftDiaSymReaderPdb2PdbVersion)" />
+    <PackageReference Update="Microsoft.Net.Compilers"                                                Version="3.3.0-beta2-19357-02" />
     <PackageReference Update="OpenCover"                                                              Version="$(OpenCoverVersion)" />
     <PackageReference Update="Codecov"                                                                Version="$(CodecovVersion)" />
+    <PackageReference Update="MicroBuild.Core"                                                        Version="$(MicroBuildCoreVersion)" />
```
jmarolf: Why do we need to update these?

build/Packages.targets
davkean: I didn't, this is an addition.

build/Packages.targets
davkean: See the update, I've explicitly set their version. I want to remove the double-restore - so I've moved them into our dependency graph.

build/Packages.targets
jmarolf: Sorry, just waking up

build/Packages.targets
jmarolf: All the micro-build package should be restore-able only on a signed build machine. I don't understand how setting the version of packages that will fail to restore / aren't restored fixes things

davkean: > All the micro-build package should be restore-able only on a signed build machine
>  I don't understand how setting the version of packages that will fail to restore

Explain what you mean by this. This doesn't fail to restore.

jmarolf: I see, `MicroBuild.Core` was uploaded to the dotnet-buildtools myget feed [here](https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/MicroBuild.Core). So restore will work. I think this won't break the signed build (which uses version 1.42) but you should check.

jmarolf: And now that I checked on the signed build, it looks like that was broken in a recent PR...

davkean: This one: https://github.com/dotnet/project-system/pull/5107/files would be one cause @drewnoakes.

davkean: See: %USERPROFILE\.nuget\packages\roslyntools.repotoolset\1.0.0-beta-62705-01\tools\Tools.proj

We have these dependencies, we use it to build the Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles project.

davkean: @jmarolf Can we modify .vsts-ci.yml to call something new, or does this require us to push that to every branch?

davkean: Jon answered with yes, .vsts-ci.yml is per branch.

davkean:
This passed the sign build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2871911.

PR #5118, opened 2019-07-22
fix signed build

jmarolf:


davkean:
Thanks for fixing this. Is this XAML used for all branches, or is per branch?

.vsts-ci.yml
```diff
@@ -27,10 +27,10 @@ steps:
     feed: '8f470c7e-ac49-4afe-a6ee-cf784e438b93'
 
 - task: BatchScript@1
-  displayName: Run script CIBuild.cmd
+  displayName: Run script build.cmd
   inputs:
-    filename: 'build\CIBuild.cmd'
-    arguments: '-configuration $(BuildConfiguration) -optimize -sign'
```
drewnoakes: The current `build.cmd` doesn't include `-optimise`. We probably need this too.

PR #5130, opened 2019-07-23
/ci no longer implies /diagnostic

davkean:
It was confusing that ci implies other switches, explicitly set /diagnostic on unit tests & signed builds.


davkean:
As we modify the builds, we've found it very hard to understand what behavior the CI has by reading the scripts. The `no-` does two things; 1) it turns off options that are on by default,  2) it documents the behavior of the yml.

PR #5131, opened 2019-07-23
Fix optimization for signed builds

drewnoakes:
f341eab removed the `-optimize` argument for signed builds, resulting in two changes:

1. Output binaries were not being optimised
2. IBC data was not being applied (when available)

The "optimize" parameter of `Build.ps1` was serving double duty.

This PR:

- adds a new `ibc` parameter
- sets `ibc` on signed builds only
- removes overriding the MSBuild `Optimize` property so that optimisation occurs for Release builds even if we don't want IBC

### DIFF
--- a/build/azure-pipelines.yml
+++ b/build/azure-pipelines.yml
@@ -36,7 +36,7 @@ jobs:
         _configuration: Release
   timeoutInMinutes: 20
   steps:
-    - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /no-deploy /no-integration /configuration $(_configuration)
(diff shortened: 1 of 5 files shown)


build/azure-pipelines.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         _configuration: Release
   timeoutInMinutes: 20
   steps:
-    - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /no-deploy /no-integration /configuration $(_configuration)
+    - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /no-deploy /no-integration /no-ibc /configuration $(_configuration)
```
davkean: You missed one in this file, there's two command-lines.

drewnoakes: Thanks, fixed.

PR #5136, opened 2019-07-23
Reduce MEF composition errors in build log

davkean:
This reduces the MEF composition errors in the build log, there's a few remaining that require changes to Apex itself.


src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/TestBase.cs
```diff
@@ -23,20 +23,11 @@ protected TestBase()
             SuppressReloadPrompt = false;
         }
 
+        protected override bool IncludeReferencedAssembliesInHostComposition => false; // Do not things we reference to the MEF Container
```
drewnoakes: Did you mean?

```suggestion
        protected override bool IncludeReferencedAssembliesInHostComposition => false; // Do not add things we reference to the MEF Container
```

PR #5146, opened 2019-07-24
Simplify how we find media recorder

davkean:
It gets deployed to output and we just load it from there.


drewnoakes:
We've worked out that this approach doesn't deploy `VSTestVideoRecorder.exe` and therefore the media recorder fails at runtime. Another approach inbound.

PR #5148, opened 2019-07-24
Make media recording actually work

davkean:
We already setup the correct paths in IntegrationTests.targets to load the adapter directly from the package.


davkean:
@jmarolf We figured this out - we did break this when we turned off MSTest deploy.

PR #5152, opened 2019-07-24
Configure CI integration tests on master

drewnoakes:
This brings config from the `feature/integration-tests` branch over to run on `master`.


build/IntegrationCIBuild.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
```
davkean: We don't need this.

drewnoakes: This PR is just bringing over what's in the `feature/integration-tests` branch. I'll make another PR once this merges that includes what's currently in #5111.

drewnoakes:
This successfully triggered a build after merge https://dev.azure.com/dnceng/public/_build/results?buildId=277134

PR #5153, opened 2019-07-24
Log VS and MSBuild versions/paths

drewnoakes:


build/build.ps1
```diff
@@ -95,16 +95,19 @@ function InstallVSIX([string] $vsixExpInstalleExe, [string] $rootsuffix, [string
 
 function LocateVisualStudio {
   if ($InVSEnvironment) {
+    Write-Host "Using Visual Studio from VSINSTALLDIR environment variable: $env:VSINSTALLDIR"
```
jmarolf: This is going to print out something like `C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\` for that variable if we are executing from a VS command prompt.  Don't know how useful that will be for people.

drewnoakes: The intent is to show exactly what version is running locally so you can compare with the CI machine when trying to repro a build that's failing under CI. If there's a way to also include the version number for a given path, we can add that too.

jmarolf: I guess I would change the message to explicitly state that we know because of this variable that we are likely executing from within a VS command prompt

drewnoakes:
Merging as I want to see the version running under CI.

build/build.ps1
jmarolf: > If there's a way to also include the version number for a given path, we can add that too.

vswhere is the only commandline tool I am aware of that can calculate this (I believe it needs to read from the registry to be able to do this). You could parse the output of `vswhere -all -prerelease -format json` and find the version (`installationVersion`) for a given `installationPath` that matches the `VSINSTALLDIR` variable

PR #5154, opened 2019-07-24
Updates to CI logging

drewnoakes:


drewnoakes:
@davkean we talked earlier about wanting to see a dump. The build for this PR has one, though it's a minidump.

https://dev.azure.com/dnceng/public/_build/results?buildId=277253

> 10>EXEC : error : (5:15.434) [Platform:DumpCollectorService] **A dump file for the 'devenv' process has been written to the following location** 'F:\workspace\_work\1\s\artifacts\Debug\tmp\Dumps\devenv_5464.dmp'.
> 10>EXEC : error : (5:15.606) [Platform:Testcase] The test failed due to an exception, but the stack trace was unavailable. Refer to the mstest log if necessary.

This dump was published under artifact _Debug Temporary Directory_ archive. I'm not sure there's anything else of interest in the `tmp` folder. I think we could just zip up the _Dumps_ folder.

It shows a call to `System.Windows.Window.ShowDialog`, which we would expect based on the video recording seen earlier today.

davkean:
Where did it come from? Was this a test?

drewnoakes:
(Already discussed offline, but for completeness) no was generated by what we already have on master, probably as part of Apex.

PR #5158, opened 2019-07-24
File watcher needs to watch full paths, not relative ones.

davidwengier:
I plan to revisit paths across the board later, once all of the bits are in and working. For now this gets bits working.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
@@ -100,8 +100,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 
             foreach (string changedFile in arg.Value)
             {
+                string relativeFilePath = _project.MakeRelative(changedFile);
+
                 // if a shared input changes, we recompile everything
-                if (_sharedDesignTimeInputs.Contains(changedFile))
+                if (_sharedDesignTimeInputs.Contains(relativeFilePath))
                 {
                     foreach (string file in _designTimeInputs)
                     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -100,8 +100,10 @@ internal void ProcessFileChangeNotification(IProjectVersionedValue<string[]> arg
 
             foreach (string changedFile in arg.Value)
             {
+                string relativeFilePath = _project.MakeRelative(changedFile);
+
                 // if a shared input changes, we recompile everything
-                if (_sharedDesignTimeInputs.Contains(changedFile))
+                if (_sharedDesignTimeInputs.Contains(relativeFilePath))
```
davkean: Are these actually results actually guaranteed to be in this same format? 

ie the following files are consider the same file, but are considered as different files here:

```
Foo\..\Foo\Foo.cs
Foo\Foo.cs
```

Doesn't SharedDesignTimeInputs come from the project file so might be in the format of ".Foo\..\Foo\Foo.cs"?

davidwengier: No, no guarantees anywhere. As per the description, this is just to get things working so I can test the various possibilities in VS.

davkean: Do you have a list of things that remaining so that we don't lose track of these issues?

davidwengier: Yeah, I keep the emails from the PR comments that still need to be actioned, discussed or thought about :)

PR #5159, opened 2019-07-24
Add telemetry to track compilation queue info

davidwengier:
Telemetry to track things like queue size, how often cancellations are, how often things are actually compiled, etc.

May be refined in future.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -231,21 +238,51 @@ private Task ProcessCompileQueueAsync(CancellationToken token)
...
+                            {
+                                compileCount++;
+                            }
+                        }
+                        catch (OperationCanceledException)
+                        {
+                            LogTelemetry(true);
+                            return;
+                        }
+                        count++;
```
drewnoakes: Undeclared symbol?

davidwengier: Yeah, that line should have gone. VS Code leaves a lot to be desired when it comes to staging parts of a file, but its the only thing I've got ¯\\\_(ツ)_/¯

I even installed the GitExtensions extension today hoping I could do it from VS, but then discovered it doesn't work without GitExtensions installed. Pity.

PR #5160, opened 2019-07-24
Handle when vstest.console does not produce trx

davkean:
Also pulled out reporting of the errors into a separate target so that we can (in a followup PR) run a target in between it and breaking the build.


davkean:
Closed in lieu of https://github.com/dotnet/project-system/pull/5161.

PR #5169, opened 2019-07-25
Refactor hive/environment setting

davkean:
Pull this logic our into separate classes.


src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/Environment/AssemblyLifetime.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.VisualStudio
+{
+    [TestClass] // Just to make sure AssemblyInitialize get called
+    public static class AssemblyLifetime
```
davkean: This will become more important in future PRs where it starts calling a lot more components.

PR #5171, opened 2019-07-25
Move files

davkean:
This includes other PRs, so just review https://github.com/dotnet/project-system/commit/bbe981f63206d8c484530637443c3350ba6a580b.


jmarolf:
Can we file an issue to make `TestContext` a public api in APEX so we don't need to use reflection?

davkean:
@jmarolf We're rethinking the logger situation after the last update of vstest - the whole thing has got a lot worse and we're going to throw away this code.

PR #5172, opened 2019-07-25
More TempPECompilerManager Tests

davidwengier:
Fixes an issue where a compilation being cancelled would leave a 0 byte file on disk, and adds a couple of other test cases that didn't have coverage.

Also include #5158 so only review https://github.com/dotnet/project-system/commit/4b4b9d7b4634f725c9d8a7ca2ad4487282ba581e


davidwengier:
Feels good man...
image

Now to work out why the spanish and release legs failed, but not debug 😕

PR #5179, opened 2019-07-25
Move Setup/DeployXXX projects 

davkean:
Move projects around so that things that are modified often are left top-level and lesser used projects are moved under nested folders. Also moved CommonFiles -> source tree; it was the only project in the solution that lived outside of src. Also no longer need to duplicate directory.build.prop/targets for it.

- Move setup\Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles -> src\setup\Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles
- Move src\ProjectSystemSetup -> src\setup\ProjectSystemSetup
- Move src\VisualStudioEditorsSetup-> src\setup\VisualStudioEditorsSetup
- Move src\DeployTestDependencies -> src\deploy\DeployTestDependencies
- Move src\DeployIntegrationDependencies -> src\deploy\DeployIntegrationDependencies

The new solution looks like:

image


ProjectSystem.sln
```diff
@@ -3,7 +3,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.28301.62
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ProjectSystemSetup", "src\ProjectSystemSetup\ProjectSystemSetup.csproj", "{9FBD7EF2-9449-486D-9FDD-FA56160AA8BB}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Setup", "Setup", "{AC8DB8AE-AC9F-4503-83C6-255B66C318C6}"
```
davkean: This was deliberately moved to the top of the solution; when you first open the solution, the default F5 project is based on the first project in the solution based on a depth first walk of the "nested" folders.

src/setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
```diff
@@ -15,4 +15,9 @@
   <ItemGroup>
     <SwrFile Include="*.swr" />
   </ItemGroup>
+
+  <Target Name="Build" DependsOnTargets="ResolveProjectReferences">
+    <MakeDir Directories="$(IntermediateOutputPath)" ContinueOnError="True"/>
+  </Target>
+
```
tmeschter: What's the purpose behind this?

davkean: This was in Directory.Build.targets and moved to the project, I've not changed it. I think its intentions is to ensure that its project dependencies are built and that its intermediates are created for when we generate vsman/vsix for it.

ProjectSystem.sln
drewnoakes: TIL 👍

drewnoakes:
Do we still need `Directory.Build.props` and `Directory.Build.targets` in the repo root now, or can they merge into those in `src`?

drewnoakes:
Do we need the RepoToolset stuff under the src branch? If not we can move some code from the root `.props` file into `build/*.props`.

davidwengier:
While things are being shuffled, I've been meaning to ask, is there a reason we don't use the `src` and `tests` style folder layout?

davkean:
> Do we need the RepoToolset stuff under the src branch? If not we can move some code from the root .props file into build/*.props

I'm trying to figure this out - I spent a little time last night trying to understand how this works and do plan to centralize this.

>While things are being shuffled, I've been meaning to ask, is there a reason we don't use the src and tests style folder layout?

I'm not a fan of it because I modify both constantly - it added an extra folder that you need to navigate. Is there a reason you like it?

davidwengier:
>  Is there a reason you like it? 

Just for ease of navigation. 5 out of 9 of our projects are test related, would be nice to get them out of the way when browsing around looking for a file.

davkean:
So that would probably be the reverse of the above proposal right?

```
ProjectSystem.sln
test\
src\
deploy\
setup\
```

Is that what you mean? If there's agrement on that, I'm okay with it.

PR #5185, opened 2019-07-26
Avoid faulting block when PackageReference is missing

davkean:
Telemetry shows that we're hitting configurations without "PackageReference" defined, filter them out from NuGet restore to prevent us throwing from GetExportedValue.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
```diff
@@ -45,7 +45,9 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreInfo> targe
 
             var restoreConfiguredInputSource = new UnwrapCollectionChainedProjectValueDataSource<IReadOnlyCollection<ConfiguredProject>, PackageRestoreConfiguredInput>(
                 _project.Services, 
-                projects => projects.Select(project => GetProjectRestoreDataSource(project)),
+                projects => projects.Select(project => project.Services.ExportProvider.GetExportedValueOrDefault<IPackageRestoreConfiguredInputDataSource>())
+                                    .Where(dataSource => dataSource != null)    // Filter our those without PackageReference
```
drewnoakes:
```suggestion
                                    .Where(dataSource => dataSource != null)    // Filter out those without PackageReference
```

PR #5186, opened 2019-07-26
Update VSSDK Build Tools version

drewnoakes:
This should fix a hang we're seeing on the CI box.

### DIFF
--- a/build/Packages.targets
+++ b/build/Packages.targets
@@ -35,7 +35,7 @@
     <PackageReference Update="Microsoft.DevDiv.Validation.MediaRecorder"                              Version="15.0.199" />
 
     <!-- VS SDK -->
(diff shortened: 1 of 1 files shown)


ProjectSystem.sln
```diff
@@ -39,6 +39,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Build", "Build", "{DD7EF107
 		.vsts-ci.yml = .vsts-ci.yml
 		build\azure-pipelines-integration.yml = build\azure-pipelines-integration.yml
 		build\azure-pipelines.yml = build\azure-pipelines.yml
+		build\Packages.targets = build\Packages.targets
```
davkean: This is already under Dependencies.

drewnoakes: Reverted that line.

PR #5190, opened 2019-07-26
Add /no-build option

drewnoakes:
This will be useful now while investigating test failures where we don't want anything to build. It might be removed later, but is pretty harmless to leave in.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -54,9 +54,8 @@ exit /b %ERRORLEVEL%
 echo Usage: %BatchFile% [options]
 echo.
 echo   Build targets:
-echo     /build                    Perform a build (default)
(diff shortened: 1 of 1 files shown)


build.cmd
```diff
@@ -55,6 +56,7 @@ echo.
 echo   Build targets:
 echo     /build                    Perform a build (default)
 echo     /rebuild                  Perform a clean, then build
+echo     /no-build                 Don't build at all (useful if running tests)
```
davkean: Should we merge with `/build` so it becomes `/no-build`.

drewnoakes:
Build error is unrelated to this change. Merging.

Issue #5180, opened 2019-07-25
App Designer - Debug page - TAB order starts at "delete" button


WilliamXieMSFT:
**Visual Studio Version**:
D16.3 Preview 2

**Summary**:
For Asp.NET Core and .NET Core Console applications, opening the project's App Designer Debug page and hitting TAB starts at the "Delete" button rather than the ComboBox for Profile.

**Steps to Reproduce**:

1. Open an Asp.NET Core or .NET Core Console app in Visual Studio

2. Right Click Project > Properties, Navigate to the Debug section & click on the page to set focus.

3. Hit tab and observe the keyboard focus.

**Expected Behavior**:
Focus should be on the Profile Combobox as it's the first item in the page:
![image](https://user-images.githubusercontent.com/29215642/61895216-03fabe00-aec7-11e9-84b3-84e3906ec755.png)

**Actual Behavior**:
Focus is on the "Delete" button:
![image](https://user-images.githubusercontent.com/29215642/61895242-12e17080-aec7-11e9-9c3b-cb6e22995d19.png)

**User Impact**:
This bug was found from an internal UX board review: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/951777

etbyrd:
@jmarolf This seems to be a potential regression from this fix: https://github.com/dotnet/project-system/pull/3782 The tab order still has the profile combo box at 0, so I am not sure how this behavior cropped up again.

etbyrd:
It seems that at least part of the issue is that the Profile ComboBox is disabled when the page is loaded, and the tab is put onto the first control that is not disabled. https://github.com/dotnet/project-system/blob/7d06fb8c592377af422e83235c9031c0bf57372c/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageControl.xaml#L74-L80

OneWay and EnvironmentVariablesValid both are false because they haven't been initialized yet. I will get a PR out for this.

WilliamXieMSFT:
Great, thanks @etbyrd !

davkean:
@etbyrd I've moved this into 15.3 as it affects accessibility.

etbyrd:
Fixed in https://github.com/dotnet/project-system/pull/5191

PR #5191, opened 2019-07-26
Fix Starting Tab Selection on Debug Property Page

etbyrd:
Fix for: https://github.com/dotnet/project-system/issues/5180

This ensures that when tabbing onto the WPF Debug property page, the first item is consistently selected. The issue was that not all of the properties had been initialized - HasProfiles was returning false and the Profile ComboBox was disabled when the page placed the tab selection on the first enabled control. This fix just assumes that if the property has not been initialized, enable the ComboBox. If, after initialization, the property is still false, the tab selection will go to the correct control and the ComboBox will be disabled. 

Note: I am not sure why the page is not showing the selection around the ComboBox until you tab off and back onto it, but I confirmed via the Narrator that it is selected and will respond to key commands.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
@@ -414,11 +414,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
             OnPropertyChanged(nameof(NewProfileEnabled));
         }
 
+        private bool _debugTargetsCoreInitialized = false;
         public bool HasProfiles
         {
             get
             {
-                return CurrentLaunchSettings != null && CurrentLaunchSettings.Profiles.Count > 0;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -712,6 +723,8 @@ internal virtual void InitializeDebugTargetsCore(ILaunchSettings profiles)
                 {
                     _firstSnapshotCompleteSource.TrySetResult(true);
                 }
+
+                _debugTargetsCoreInitialized = true;
```
davkean: It looks like this field has the same meaning as _firstSnapshotCompleteSource?

etbyrd: Doesn't seem to be the case. It looks like _firstSnapshotCompleteSource is only used with `DebugPageViewModel` which is only used for tests. _firstSnapshotCompleteSource is always null otherwise.

etbyrd:
I haven't really explored the implementation, but it is definitely on my wish list. I think I should look into it next sprint.

PR #5193, opened 2019-07-27
do not trigger signed build on doc changes

jmarolf:


jmarolf:
> Awesome! Can we do this for the real build too, so no insertions are kicked off?

This PR is for official builds (verified this doesn't break the signed build [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2889231))

jmarolf:
[This](https://github.com/dotnet/project-system/pull/5192) PR is for CI

davidwengier:
Lol thanks.. Read one notification, must have clicked the other :)

.vsts-ci.yml
```diff
@@ -1,5 +1,22 @@
 # Responsible for producing the signed product build for insertion into the Visual Studio repository
 
+pr: none
+# Branches that trigger a build on commit
+trigger:
+  branches:
+    include:
+    - master
```
davkean: Do we have to list these branches?

jmarolf: If we don't then we will need to override the set of branches on the build defintion.  I think its better to list these out in the file instead of having to wrestle with a web UI

PR #5198, opened 2019-07-30
Fix toolbar missing in Resource/Settings editor

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/951702/.

Turns out when I testing this, it was pulling the menus from a cache instead EditorsUI.dll, so I was testing the wrong thing. This fixes embedding of the Menus.ctmenu and references to it. I've repro'd the bug, and with below changes under Exp it no longer repros.

This is currently blocked on a build with https://github.com/dotnet/xliff-tasks/pull/150 in it, and will fail to build.


nguerrera:
Looks like the build issues with xliff-tasks are resolved. Pushed a commit to use latest build so that we can have this verified by the time @davkean comes online.

I also sent a PR to update the xliff-tasks build badge: https://github.com/dotnet/xliff-tasks/pull/161

cc @tmeschter

PR #5199, opened 2019-07-30
Increate test timeouts to try to out-flake the flakyness

davidwengier:
I know timeouts are terrible, and I'm loathed to just try this, but I _love_ how un-compromised the implementation classes are by the tests because they actually exercise dataflow etc.

This might not work anyway ¯\\\_(ツ)_/¯


davidwengier:
Grrrr.. even with doubled timeouts... <old man yells at CI>

PR #5200, opened 2019-07-30
Move GenerateDependentAssemblyVersionFile to msbuild

davkean:
This moves GenerateDependentAssemblyVersionFile to msbuild.

While this is significantly more code than what we had before, the benefits will be clear when build.ps1 is replaced with an msbuild proj and you can use MSBuild Structured Build Logger to investigate the build.

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -106,6 +106,9 @@ function Build {
     $CodecovProj = Join-Path $PSScriptRoot 'Codecov.proj'
     & $MsbuildExe $CodecovProj /m /nologo /clp:Summary /nodeReuse:$nodeReuse /warnaserror /v:diag /t:Codecov /p:Configuration=$configuration /p:UseCodecov=$useCodecov /p:NuGetPackageRoot=$NuGetPackageRoot $properties
   }
+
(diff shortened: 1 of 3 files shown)


build/build.ps1
```diff
@@ -106,6 +106,9 @@ function Build {
     $CodecovProj = Join-Path $PSScriptRoot 'Codecov.proj'
     & $MsbuildExe $CodecovProj /m /nologo /clp:Summary /nodeReuse:$nodeReuse /warnaserror /v:diag /t:Codecov /p:Configuration=$configuration /p:UseCodecov=$useCodecov /p:NuGetPackageRoot=$NuGetPackageRoot $properties
   }
+
+  $GenerateDependentAssemblyVersionsProj = Join-Path $PSScriptRoot 'Scripts\GenerateDependentAssemblyVersionFile.targets'
+  & $MsbuildExe $GenerateDependentAssemblyVersionsProj /m /nologo /clp:Summary /nodeReuse:$nodeReuse /warnaserror /p:Configuration=$configuration
```
jmarolf: can we ensure that this produces a binlog on ci? I love turning things into msbuild but I thought the whole point was better logging?

davkean: It does in a followup, this is temporary while build.ps1 exists.

PR #5207, opened 2019-07-30
Annotations

drewnoakes:
More nullability annotations.

After this, there are still 561 occurrences of `#nullable disable` under `src` (mostly in test code though).

image

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowProjectLogger.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowProjectLogger.cs
@@ -6,8 +6,6 @@ using System.ComponentModel.Composition;
 using Microsoft.VisualStudio.ProjectSystem.Logging;
 using Microsoft.VisualStudio.Shell.Interop;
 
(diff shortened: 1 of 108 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowProjectLogger.cs
```diff
@@ -44,9 +42,9 @@ public void WriteLine(in StringFormat format)
                 // as via build), then we'll need to implement a better queueing mechanism.
                 _threadingService.RunAndForget(async () =>
                 {
-                    IVsOutputWindowPane pane = await _outputWindowProvider.GetOutputWindowPaneAsync();
+                    IVsOutputWindowPane? pane = await _outputWindowProvider.GetOutputWindowPaneAsync();
```
davkean: This looks wrong - this should return non-null, I think?

drewnoakes: `ProjectOutputWindowPaneProvider.CreateOutputWindowPaneAsync` can return `null` (in command line builds), which surfaces here.

Issue #5197, opened 2019-07-29
Cannot switch from "Icon and Manifest" to "Resource File"


nguerrera:
**Visual Studio Version**:

VisualStudio.16.IntPreview/16.3.0-pre.2.0+29128.8.master

**Summary**:

It's difficult to change from "Icon and Manifest" to "Resource File" in .NET Core project because it switches back to "Icon and Manifest" before I can click browse after selecting "Resource File"

Repro is not 100%, and eventually it will settle down and stay on  Resource File. But usually I can hit this following the steps below, as can @wli3

**Steps to Reproduce**:

1. File -> New Project -> .NET Core Console App

2. Solution Explorer -> Right click project -> properties

3. Application tab, under resources -> select "Resource file"

![cantpickresourcefile](https://user-images.githubusercontent.com/75470/62087316-36355400-b216-11e9-97e4-a25e09285d99.gif)

**Expected Behavior**:
It stays on "Resource file" so I can browse or type in the path to resource file.

**Actual Behavior**:
It reverts back to Icon and Manifest

**User Impact**:
Frustrating UI.

drewnoakes:
I can repro this in 16.2.0 (GA).

1. New .NET Core project
2. Open properties
3. Click _Resource file_
  - The _Icon and manifest_ item is reselected
  - The `.csproj` file now contains:
    ```xml
    <ApplicationIcon />
    <Win32Resource />
    ```
4. Clicking _Resource file_ a second time (for me at least) causes the selection to stick

drewnoakes:
@etbyrd can you take a look at this?

etbyrd:
I see why this is happening. I will try to get a PR out for this tomorrow.

drewnoakes:
@etbyrd I've added it to this sprint as it's a regression.

davkean:
@etbyrd @drewnoakes Do we have an understanding of what regressed here?

etbyrd:
Looking at just the blame, I am not sure if this is even a regression.

davidwengier:
Closing this as a duplicate of #4519

I've confirmed that if we don't set property values to their current values this doesn't repro. The real fix for this will be done in CPS (see https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1025549)

davidwengier:
In case anyone is watching this issue, the linked duplicate has been fixed.

Issue #4519, opened 2019-01-31
Changing project properties adds unnecessary empty ApplicationIcon and StartupObject


nguerrera:
**Visual Studio Version**:

```
Microsoft Visual Studio Enterprise 2019 Int Preview
Version 16.0.0 Preview 3.0 [28529.153.d16.0]
VisualStudio.16.IntPreview/16.0.0-pre.3.0+28529.153.d16.0
```

**Summary:**

Changing project properties adds unnecessary empty ApplicationIcon and StartupObject properties to project file.

**Steps to Reproduce**:

1. File -> New Project -> Console App (.NET Core)

2. Inspect project file

3. Right click project, properties

4. Change output type to Windows Application

5. Inspect project file again

**Expected Behavior**:

Only change in project file is `s/Exe/WinExe/`

**Actual Behavior**:

Two extra lines are added to project file setting ApplicationIcon and StartupObject to empty.

``` diff
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
-   <OutputType>Exe</OutputType>
+   <OutputType>WinExe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
+   <ApplicationIcon />
+   <StartupObject />
  </PropertyGroup>

</Project>
```

**User Impact**:

Unwanted, unnecessary content in csproj file

drewnoakes:
These empty elements would override imported properties too.

nguerrera:
Oh that's worse. I thought it was merely persisting the evaluated empty value and hoping if it were non-empty from imports it would reflect that.

davidwengier:
CPS is tracking this here: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1025549

davidwengier:
Internal PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/215694 has been completed.

PR #5215, opened 2019-07-30
Fix resource and icon radio button behavior, clean up code

etbyrd:
Fix for https://github.com/dotnet/project-system/issues/5197

The setters for the radio buttons were always returning true because `SetIconAndWin32ResourceFile` was always returning true. I moved some of the code out of that method and call it when the radio buttons are changed, so it is only updated once. 

I verified all of the behavior matched, but I will give it another pass.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
@@ -299,39 +299,6 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             propApplicationManifest = GetPropertyDescriptor("ApplicationManifest")
             propWin32ResourceFile = GetPropertyDescriptor("Win32ResourceFile")
 
-            obj = TryGetNonCommonPropertyValue(propApplicationIcon)
-            If Not PropertyControlData.IsSpecialValue(obj) Then
-
-                stApplicationIcon = TryCast(obj, String)
-
-                If (Trim(stApplicationIcon) = "") Then
-                    If (ProjectProperties.OutputType <> VSLangProj.prjOutputType.prjOutputTypeLibrary) Then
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
```diff
@@ -510,9 +457,65 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            Win32ResourceFile.Enabled = False
+            Win32ResourceFileBrowse.Enabled = False
+            Win32ResourceRadioButton.Checked = False
+            EnableControl(ApplicationIconLabel, ApplicationIconSupported())
+            EnableControl(ApplicationIcon, ApplicationIconSupported())
+            EnableControl(AppIconBrowse, ApplicationIconSupported())
+            If (ProjectProperties.OutputType <> VSLangProj.prjOutputType.prjOutputTypeLibrary) Then
+                ApplicationManifest.Text = stApplicationManifest
+                EnableControl(ApplicationManifestLabel, ApplicationManifestSupported())
+                EnableControl(ApplicationManifest, ApplicationManifestSupported())
```
drewnoakes: Have you tested changing the output type? Are the enabled states of these controls updated correctly?

davkean:
Can we separate the bug fix commit from the cleanup commit? I can't tell what the bug is.

davkean:
It would be good to understand why this is broken for .NET Core but not .NET Framework, that context would help.

davidwengier:
I wonder if its related to #4519 and those changes causing the pages to refresh.

davkean:
@etbyrd I didn't see an update to this?

davkean:
@melytc Going to close this out as its old and becoming stale, feel free to reopen/open another version if you have a fix for this.

PR #5222, opened 2019-07-31
Fix flaky tests

davidwengier:
This seems to fix the flakyness of the tests by allowing the tests to complete the data flow block for the file watcher, and to wait for the task scheduler to do its work for the compiler.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcherTests.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcherTests.cs
@@ -21,6 +21,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 {
     public class DesignTimeInputsFileWatcherTests
     {
+        private const int TestTimeoutMillisecondsDelay = 1000;
+
         public static IEnumerable<object[]> GetTestCases()
         {
             return new[]
@@ -85,6 +87,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 
             using DesignTimeInputsFileWatcher watcher = CreateDesignTimeInputsFileWatcher(fileChangeService, out ProjectValueDataSource<DesignTimeInputs> source);
 
+            watcher.AllowSourceBlockCompletion = true;
+
             // Send our input
             await source.SendAsync(new DesignTimeInputs(designTimeInputs, sharedDesignTimeInputs));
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcherTests.cs
```diff
@@ -111,8 +115,12 @@ internal async Task VerifyDesignTimeInputsWatched(string[] designTimeInputs, str
...
 
+            watcher.SourceBlock.Complete();
+
+            await watcher.SourceBlock.Completion;
+
             // The timeout here is annoying, but even though our test is "smart" and waits for data, unfortunately if the code breaks the test is more likely to hang than fail
-            if (await Task.WhenAny(finished.Task, Task.Delay(500)) != finished.Task)
+            if (await Task.WhenAny(finished.Task, Task.Delay(TestTimeoutMillisecondsDelay)) != finished.Task)
             {
                 throw new AssertActualExpectedException(fileChangeNotificationsExpected.Length, notificationCount, "Timed out after 500ms");
```
drewnoakes:
```suggestion
                 throw new AssertActualExpectedException(fileChangeNotificationsExpected.Length, notificationCount, $"Timed out after {TestTimeoutMillisecondsDelay}ms");
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/TempPECompilerManagerTests.cs
```diff
@@ -385,7 +395,7 @@ private async Task VerifyDLLsCompiled(int expectedDLLs, Func<Task> actionThatCau
             await actionThatCausesCompilation();
 
             // Sadly, we need a timeout
-            var delay = Task.Delay(TimeSpan.FromSeconds(1));
+            var delay = Task.Delay(TestTimeoutMillisecondsDelay);
 
             if (await Task.WhenAny(_compilationOccurredCompletionSource.Task, delay) == delay)
             {
```
drewnoakes: There's another incorrect error message below here.

davidwengier:
Grrr... maybe I have to give up on testing data flow  :(

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/TempPECompilerManager.cs
```diff
@@ -204,7 +211,11 @@ private void QueueCompilation()
         {
             if (_filesToCompile.Count > 0)
             {
-                _scheduler.ScheduleAsyncTask(ProcessCompileQueueAsync, _project.Services.ProjectAsynchronousTasks.UnloadCancellationToken);
+                JoinableTask task = _scheduler.ScheduleAsyncTask(ProcessCompileQueueAsync, _project.Services.ProjectAsynchronousTasks.UnloadCancellationToken);
```
davkean: I would consider setting a TaskCOmpletetionSource that you can wait on instead.

davkean: Or just wait for the telemetry?

davidwengier: The tests are using a TaskCompletionSource, but for scenarios where there is not supposed to be anything compiled it doesn't help. Making this optionally synchronous seems to have fixed the flakiness in this, its really the dataflow that is giving me trouble. I have one more thing to try for that though.

PR #5223, opened 2019-07-31
Fix flaky tests

davidwengier:
Attempt Number 4.

Unfortunately the flakiness best manifests itself on a fresh PR build, sorry about the spam :)


drewnoakes:
> Unfortunately the flakiness best manifests itself on a fresh PR build

Why would it be any different to `/azp run`?

davidwengier:
No idea, I just know that re-running the failed jobs has made them pass 100% of the time, and I've only ever seen the tests fail locally once, and I've tried from VS, from command line, doing a rebuild, doing a git clean, etc.

drewnoakes:
Very strange. Is it something to do with build machine affinity? Are your tests creating files on disk which remain cached? DLLs loaded into memory (cold vs. warm timing)?

davidwengier:
Not my tests directly, but I assume its a cold vs warm thing, and on cold start the background tasks (data flow and the task scheduler) don't get a chance to run. But thats just a guess.

davkean:
What were the additional failures that caused you to re-run this? It sounds like there's still flakiness...

davidwengier:
Nah, there weren't any failures, I just thought I'd run them again to be sure that first pass wasn't a fluke. We'll know if they are still flaky though, as it affects the auto-merges into this branch every time master changes.

PR #5201, opened 2019-07-30
Framework reference browse object

drewnoakes:
Relates to #5070 and #4762.

With SDK 3.0.100-preview8, we now receive `ResolvedFrameworkReference` items with the required properties. This PR removes the workaround that was required until now to framework references correctly.

- The `Path` property is populated.
- The item's description is now "Framework Properties" rather than "Framework Reference Properties". This matches "Package Properties" for `<PackageReference />`.
- The version has been added. Although the version is available in the path itself, it's usually off screen.

### Before

image

### After

image

PR #5226, opened 2019-07-31
Improvements to reference rules

### LINKED ISSUE #4281: Properties Windows for ProjectReferences experience is broken
When clicking on a project reference in Solution Explorer we show a few properties as editable in the properties window, but Private Assets, Include Assets and Exclude Assets and Reference Output Assembly error. The three Assets properties also are free text, rather than showing a good UI.

When setting Reference Output Assembly the node breaks but that is being tracked by a separate item.

drewnoakes:
Fixes #4281 and other bugs in rule files discovered after authoring tests.

There's a fair bit of churn in the tests due to consolidation. Ultimately the most important thing in this PR to review is that the changes to the rule files themselves are correct.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ProjectReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ProjectReference.xaml
@@ -63,8 +63,7 @@
 
   <StringProperty Name="ExcludeAssets"
                   Description="Assets to exclude from this reference."
-                  DisplayName="Exclude Assets"
-                  Visible="True" />
+                  DisplayName="Exclude Assets" />
 
   <StringProperty Name="Identity"
                   Description="Security identity of the referenced assembly (see System.Reflection.Assembly.Evidence or System.Security.Policy.Evidence)."
@@ -78,16 +77,14 @@
 
   <StringProperty Name="IncludeAssets"
                   Description="Assets to include from this reference."
-                  DisplayName="Include Assets"
-                  Visible="True" />
+                  DisplayName="Include Assets" />
 
   <BoolProperty Name="LinkLibraryDependencies"
                 Visible="False" />
 
   <StringProperty Name="PrivateAssets"
                   Description="Assets that are private in this reference."
-                  DisplayName="Private Assets"
-                  Visible="True" />
+                  DisplayName="Private Assets" />
 
   <StringProperty Name="Project"
                   Visible="False" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedAssemblyReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedAssemblyReference.xaml
@@ -135,7 +135,7 @@
     <BoolProperty.DataSource>
       <DataSource HasConfigurationCondition="False"
                   ItemType="Reference"
-                  Persistence="AssemblyReference"
+                  Persistence="ProjectFile"
                   SourceOfDefaultValue="AfterContext" />
     </BoolProperty.DataSource>
   </BoolProperty>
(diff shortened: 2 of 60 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ProjectReference.xaml
```diff
@@ -63,8 +63,7 @@
 
   <StringProperty Name="ExcludeAssets"
                   Description="Assets to exclude from this reference."
-                  DisplayName="Exclude Assets"
-                  Visible="True" />
+                  DisplayName="Exclude Assets" />
```
drewnoakes: Removed properties setters set to default.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedAssemblyReference.xaml
```diff
@@ -135,7 +135,7 @@
     <BoolProperty.DataSource>
       <DataSource HasConfigurationCondition="False"
                   ItemType="Reference"
-                  Persistence="AssemblyReference"
+                  Persistence="ProjectFile"
```
drewnoakes: I think this property-level data source was copied from the rule-level one, and the `Persistence` property not updated.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedFrameworkReference.xaml
```diff
@@ -8,7 +8,7 @@
   
   <Rule.DataSource>
     <DataSource HasConfigurationCondition="False"
-                ItemType="_ResolvedFrameworkReference"
+                ItemType="FrameworkReference"
```
drewnoakes: This should probably have been included in #5201. I'm not sure why this property is needed, as it doesn't effect anything as far as I can tell. However other such property-level `ItemType` properties match the outer item type, so I've updated this too for consistency.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedProjectReference.xaml
```diff
@@ -40,7 +40,14 @@
 
   <BoolProperty Name="CopyLocalSatelliteAssemblies"
                 Description="Indicates whether the satellite assemblies of the reference target should be copied into this project's output directory."
-                DisplayName="Copy Local Satellite Assemblies" />
+                DisplayName="Copy Local Satellite Assemblies">
+    <BoolProperty.DataSource>
+      <DataSource HasConfigurationCondition="False"
+                  ItemType="ProjectReference"
+                  Persistence="ProjectFile"
```
drewnoakes: This is an example of adding a property-level data source that persists to the project file on resolved dependencies. Without this, the rule-level data source is inherited and cannot write its value anywhere. There are several instances of this throughout the PR.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/Folder.xaml
```diff
@@ -15,13 +15,13 @@
 
   <StringProperty Name="FolderName"
                   Category="Misc"
-                  Description="Name of this folder"
+                  Description="Name of this folder."
```
drewnoakes: Added trailing periods to descriptions. We already had a test for this, but it wasn't checking all rule files. This PR updates the tests to be more explicit about what rule files are included for each verification.

davkean:
Didn't https://github.com/dotnet/project-system/issues/4281 mention providing a better UI for manipulating PrivateAssets/ExcludeAssets? I don't see it in this.

drewnoakes:
I looked at that but the values are potentially many values (semicolon separated ), which `EnumProperty` doesn't support. `StringListProperty` could work here, but I don't see it adding much. Am I missing something?

PR #5233, opened 2019-08-01
Replace build.ps1 with build.proj

davkean:
This is straight replacement of build.ps1 with an MSBuild project that does the same thing. This let's us diagnose the entire build via a single Build.binlog.

I've maintained the behavior as before with one exception[1], and doesn't include changes to remove the need to double-restore toolset which will happen in followup.

[1] I've removed TEMP directory setting. I could never get this to work in the same way that PowerShell had it, I could only get MSBuild to respect it - but not child proceses it spawned. I'm unsure why we redirected temp and was probably from days of Jenkins when we had overlapping builds on the same machine.


build.cmd
```diff
@@ -5,49 +5,54 @@ set BatchFile=%0
...
+set OptIntegrationTest=false
+set OptCI=false
+set OptSign=false
+set OptIbc=false
+set OptNodeReuse=true
 
 :ParseArguments
-if    "%1" == "" goto :DoneParsing
-if /I "%1" == "/?" call :Usage && exit /b 1
-if /I "%1" == "/build"                set OptBuild=$true  && set OptRebuild=$false  && shift && goto :ParseArguments
```
davkean: Moving the `&&` prevents spaces being put in the value.

build.cmd
```diff
@@ -5,49 +5,54 @@ set BatchFile=%0
...
-if /I "%1" == "/no-ci"                set OptCI=$false && set PrepareMachine=$false && shift && goto :ParseArguments
-if /I "%1" == "/ibc"                  set OptIbc=$true                              && shift && goto :ParseArguments
-if /I "%1" == "/no-ibc"               set OptIbc=$false                             && shift && goto :ParseArguments
-if /I "%1" == "/rootsuffix"           set PropRootSuffix=/p:RootSuffix=%2           && shift && shift && goto :ParseArguments
-if /I "%1" == "/configuration"        set BuildConfiguration=%2                     && shift && shift && goto :ParseArguments
+if    "%1" == ""                                                                                    goto :DoneParsing
+if /I "%1" == "/?"                                                                                  call :Usage && exit /b 1
+if /I "%1" == "/build"                set OptBuild=true&& set OptRebuild=false&&                    shift && goto :ParseArguments
+if /I "%1" == "/no-build"             set OptBuild=false&& set OptRebuild=false&&                   shift && goto :ParseArguments
```
davkean: Moving the && prevents spaces being put in the value.

drewnoakes: Wrapping the name/value in quotes would work.

```suggestion
if /I "%1" == "/no-build"             set "OptBuild=false" && set "OptRebuild=false" &&                   shift && goto :ParseArguments
```

davkean: Yep, good point - done that instead.

jmarolf:
The insane verbosity in the logs on AzDO is unfortunate. I am never going to look at all that text. Having a summary of where the build fails is nice to know where to start when I open the binlog. I don't need the rest of it. this is a regression from build.ps1

davkean:
This change hasn't changed the verbosity of AzureDevOps, that was changed a couple of weeks ago. I'm not sure I understand this: _Having a summary of where the build fails is nice to know where to start when I open the binlog._ There's no way to adjust verbosity of a build log AFAIK?

jmarolf:
@davkean 	previously (before the several refactorings that were done) specifying `/diagnostic` did not change the build output except for the fact that it produces binlog. see the log for [this build](https://dev.azure.com/dnceng/public/_build/results?buildId=274228)

davkean:
Understood, but this isn't this PR? Can you file a bug please? We found the new log to helpful.

build/build.proj
```diff
@@ -0,0 +1,100 @@
+<Project DefaultTargets="PrepareMachine;RestoreToolset;CallToolset;GenerateDependentAssemblyVersionsFile;UploadCodeCoverage">
```
tmeschter: I take it these targets will be run in the given order?

Is there an advantage to this over having a `Build` target that depends on these five? Like so:
``` XML
<Project DefaultTargets="Build">
...
  <Target Name="Build"
          DependsOn="PrepareMachine;RestoreToolset;CallToolset;GenerateDependentAssemblyVersionsFile;UploadCodeCoverage">
  ...
  </Target>
</Project>
```

davkean: I used to do what you are suggesting, but realized that DefaultTargets can be multiple and felt it was a little simpiler.

davkean:
Sign build passed with this.

PR #5236, opened 2019-08-01
Up-to-date check bits

drewnoakes:
Some documentation and tweaks made while looking into bugs.

- Renamed "LastItemChanged" to "LastItemsChanged" to indicate it's not the last time at which an item changed, it's the last time at which the set of items changed.
- Moved `LastCheckedAtUtc` to the state object, and introduced appropriate locking.
- Improved log messages a bit.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -600,7 +600,7 @@ async Task<bool> IsUpToDateInternalAsync(CancellationToken token)
                 finally
                 {
                     LastCheckTimeUtc = DateTime.UtcNow;
-                    logger.Verbose("Up to date check completed in {0:#,##0.#} ms", sw.Elapsed.TotalMilliseconds);
```
davidwengier: Not that your change here is bad, but the old code is culture aware too (assuming logger.Verbose doesn't do something whacky and custom under the covers)...
image

drewnoakes: TIL!

PR #5248, opened 2019-08-05
XAML rule tests

drewnoakes:
- Expands test coverage of dependencies rules.
  - Ensure resolved dependencies have `OriginalItemSpec` property
  - Ensure dependencies have `IsImplicityDefined` property
  - Ensure dependencies have `Visible` property
- Refactoring of test code and splitting up into multiple classes based on area

The final commit causes a huge diff, as it splits the test file up. Other commits are easy to review. The most important part to review are changes the XAML files themselves.


davidwengier:
> as it splits the test file up

Excellent, it was due :)

PR #5250, opened 2019-08-06
Split TempPECompilerManager into two components

### LINKED ISSUE #5231: Split TempPECompilerManager into two components
One to produce changes to design time inputs, and one to compile said changes into DLLs

Part of #4163 

See that job for more info

davidwengier:
Apologies for the annoying nature of this PR, but when I was starting to implement the next piece of TempPE there was a lot of logic that had to be duplicated from TempPECompilerManager. In order to avoid that, I split its job into two pieces - one does the calculation of changed files, and one to manage the compilation of all of the changes.

I feel like the end result is actually a lot nicer too, and makes more sense, but I leave that for you to determine through the medium of PR comments.

Fixes #5231


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -0,0 +1,241 @@
...
+        private DesignTimeInputs? _latestDesignTimeInputs;
+        private string? _latestOutputPath;
+        private int _version;
+
+        /// <summary>
+        /// The block that sends out updates from this service.
+        /// </summary>
+        private IBroadcastBlock<IProjectVersionedValue<DesignTimeInputsDelta>>? _broadcastBlock;
+
+        /// <summary>
```
davkean: I don't find these comments to be helpful - they are just restating their names.

davidwengier: Thanks, removed one, expanded another.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
```diff
@@ -0,0 +1,323 @@
...
+                while (_filesToCompile.Count > 0)
+                {
+                    if (IsDisposing || IsDisposed)
+                    {
+                        return;
+                    }
+
+                    if (token.IsCancellationRequested)
+                    {
+                        LogTelemetry(true);
```
drewnoakes:
```suggestion
                        LogTelemetry(cancelled: true);
```

Ditto twice below.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
```diff
@@ -0,0 +1,323 @@
...
+            _scheduler.Dispose();
+            _changeTrackerLink?.Dispose();
+        }
+
+        private Task ProcessCompileQueueAsync(CancellationToken token)
+        {
+            int compileCount = 0;
+            int initialQueueLength = _filesToCompile.Count;
+            var compileStopWatch = Stopwatch.StartNew();
+            return _activeWorkspaceProjectContextHost.OpenContextForWriteAsync(async accessor =>
```
drewnoakes: I looked briefly and it seems possible to update `OpenContextForWriteAsync` to accept a cancellation token. It's not clear to me whether that'd be important in this scenario or not. The telemetry here might reveal that eventually.

It's possible that there are other consumers of this and related APIs that would benefit from cancellation support on what I presume to be a widely-scoped lock.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
```diff
@@ -0,0 +1,323 @@
...
+                return true;
+            }
+            return false;
+        }
+
+        private string GetOutputFileName(string designTimeInput)
+        {
+            Assumes.NotNull(_state);
+
+            // All monikers are project relative paths by defintion (anything else is a link, and linked files can't be TempPE inputs), meaning
```
drewnoakes:
```suggestion
            // All monikers are project relative paths by definition (anything else is a link, and linked files can't be TempPE inputs), meaning 
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTrackerTests.cs
```diff
@@ -0,0 +1,343 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    public class DesignTimeInputsChangeTrackerTests : IDisposable
+    {
+        private const int TestTimeoutMillisecondsDelay = 1000;
+
+        private string? _lastIntermediateOutputPath;
+        private readonly string _projectFolder = @"C:\MyProject";
+        private string _intermediateOutputPath = "MyOutput";
+        private readonly DesignTimeInputsChangeTracker _manager;
```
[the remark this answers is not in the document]
davidwengier: It really did manage things! :P

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
```diff
@@ -0,0 +1,323 @@
...
+
+                    string outputFileName = GetOutputFileName(fileName);
+                    try
+                    {
+                        if (await CompileDesignTimeInputAsync(accessor.Context, fileName, outputFileName, ignoreFileWriteTime, token))
+                        {
+                            compileCount++;
+                        }
+                        // We remove the file if the compilation wasn't cancelled, regardless of whether it really happened or not.
+                        ThreadingTools.ApplyChangeOptimistically(ref _filesToCompile, fileName, (s, f) => s.Remove(f));
```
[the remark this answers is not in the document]
davidwengier: Thanks, good call. This is leftover from when `_filesToCompare` was a HashSet.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -0,0 +1,238 @@
...
+                // If a new shared design time input is added, we need to recompile everything regardless of source file modified date
+                // because it could be an old file that is being promoted to a shared input
+                if (inputs.SharedInputs.Except(_latestDesignTimeInputs.SharedInputs, StringComparers.Paths).Any())
+                {
+                    AddAllInputsToQueue(true);
+                }
+                // If the namespace or output path inputs have changed, then we recompile every file regardless of date
+                else if (configChanges.Difference.ChangedProperties.Contains(ConfigurationGeneral.RootNamespaceProperty) ||
+                         configChanges.Difference.ChangedProperties.Contains(ConfigurationGeneral.ProjectDirProperty) ||
+                         configChanges.Difference.ChangedProperties.Contains(ConfigurationGeneral.IntermediateOutputPathProperty))
```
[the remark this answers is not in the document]
davidwengier: These all come from CPS with appropriate comparers

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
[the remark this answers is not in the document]
davidwengier: Changed these to just TryRemove

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -0,0 +1,238 @@
...
+            // Make sure the design time inputs don't change while we're processing this notification
+            DesignTimeInputs designTimeInputs = _latestDesignTimeInputs;
+
+            var changedFiles = new List<DesignTimeInputFileChange>();
+            foreach (string changedFile in arg.Value)
+            {
+                string relativeFilePath = _project.MakeRelative(changedFile);
+
+                // if a shared input changes, we recompile everything
+                if (designTimeInputs.SharedInputs.Contains(relativeFilePath))
```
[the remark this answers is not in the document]
davidwengier: It's not a real problem because all of the inputs come from the same source (the projects items), but I was going to change these to be HashSets in a future PR anyway, I guess I'll just do that now.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputFileChange.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
+{
+    internal class DesignTimeInputFileChange
+    {
+        public readonly string File;
+        public readonly bool IgnoreFileWriteTime;
```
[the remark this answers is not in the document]
davidwengier: ¯\\\_(ツ)_/¯

Will change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
davidwengier: Definitely something for another PR, and something I'm not qualified to comment on. Any thoughts @davkean?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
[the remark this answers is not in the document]
davidwengier: Wow I was putting off that change, but this new design makes it so simple :)

davidwengier:
Thank you for the thorough and helpful comments @drewnoakes, I've pushed an update addressing them, will merge assuming checks pass.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -0,0 +1,238 @@
...
+        private int _version;
+
+        private IBroadcastBlock<IProjectVersionedValue<DesignTimeInputsDelta>>? _broadcastBlock;
+
+        /// <summary>
+        /// The public facade for the broadcast block. We don't expose the broadcast block directly because we don't want to allow consumers to complete or fault us
+        /// </summary>
+        private IReceivableSourceBlock<IProjectVersionedValue<DesignTimeInputsDelta>>? _publicBlock;
+
+        [ImportingConstructor]
```
davkean: This isn't a MEF component, who imports this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        protected override void Dispose(bool disposing)
+        {
+            if (_inputsActionBlock != null)
+            {
+                // This will stop our blocks taking any more input
+                _inputsActionBlock.Complete();
+                _fileWatcherActionBlock!.Complete();
+
+                _threadingService.ExecuteSynchronously(() => Task.WhenAll(_inputsActionBlock.Completion, _fileWatcherActionBlock.Completion));
```
davkean: Why do we need this? This says "block a threadpool thread while we wait for these blocks to finish" - but you aren't using the results.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
davkean: Yeah there could be benefit here. I think I avoided it at the time because it ballooned overloads and we didn't need it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
```diff
@@ -0,0 +1,323 @@
...
+        public Task<string> GetDesignTimeInputXmlAsync(string fileName)
+        {
+            if (_state == null)
+            {
+                throw new InvalidOperationException("Can't get design time input information until project information has been received");
+            }
+
+            // Make sure we're not being asked to compile a random file
+            if (!_state.AllInputs.Contains(fileName, StringComparers.Paths))
+            {
```
davkean: Is there a race here? By the time I've read the values and passed to them this method it could be out of date and throw?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
davidwengier: We need to wait for completion to ensure that the disposal doesn't overlap with processing more input, as Complete is not a blocking call. This was the design we worked out for the file watcher, copied from somewhere in CPS (I don't remember where): https://github.com/dotnet/project-system/blob/feature/TempPE/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsFileWatcher.cs#L160

If this case is different somehow I'd appreciate any extra context you can share.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
davidwengier: Nothing yet.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
davidwengier: Do you mean that the file could be part of `_state` at the time of the check, and then not part of it when compiling later? That's certainly possible, though it shouldn't result in a throw.

In general with this class I've avoided any locking or synchronization, but instead made things such that they can cope with failure, assuming that failure cases will be rare. That may not pan out to be true, but it's worth noting that this method has no callers at the moment, so I wouldn't bet on its current implementation being 100% correct anyway. I'm doing this is pieces because previously your feedback was that the PRs were too big, too hard to review, and should be split up into components.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
davkean: Dispose doesn't do anything other than complete the blocks, what's the race? File watcher dispose uses the results that the data flow listener touches, so the case is different.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
davkean: Where does the caller get a "file name" from? What code they write that lets them call this method without throwing?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
```diff
@@ -7,13 +7,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 {
     internal class DesignTimeInputs
     {
-        public readonly ImmutableArray<string> Inputs;
-        public readonly ImmutableArray<string> SharedInputs;
+        public readonly ImmutableHashSet<string> Inputs;
+        public readonly ImmutableHashSet<string> SharedInputs;
```
drewnoakes: Properties here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
davkean: I'll whiteboard for you on Monday how this occurs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputs.cs
davidwengier: Thanks, my bad. This file wasn't in your first review.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
davidwengier: Fair enough, I took it to be a general problem to solve. Thanks for the explanation.

> I'll whiteboard for you on Monday how this occurs.

That would be awesome, there are still some comments from the previous review that I wanted to talk through with you a few weeks ago.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs
davidwengier: Ahh I think I see what you mean; If the list of monikers that the consumer of this gets is out of date when this method is called. That does sound possible. I'm also thinking that this might change to simply process the whole queue, rather than pluck out one file, on the assumption that if one file is requested, the rest are likely to follow. 

The thing that will call this, which is the next piece to come, will be ProjectHostBridge-derived, so will be doing the UI thread stuff that does, so as mentioned I wouldn't be surprised if this changes.

drewnoakes:
I read through your recent changes. They look good to me. Does the code feel more obviously thread safe to you now?

davidwengier:
Yeah, carrying more state in the dictionary entries and not in the class itself seems better.

davidwengier:
Attn @dotnet/project-system as mentioned in the team meeting I'm merging this today (once I've fixed the build failures resulting from the merge)

PR #5251, opened 2019-08-06
Dependency node integration tests

drewnoakes:
First cut of integration tests for the dependencies node. Relates to #1607.

This revitalises PR #4658 which was automatically closed when we deleted the `feature/integration-tests` branch.

These tests build up a solution/project(s)/global.json on disk, load the solution, then inspect the dependencies nodes.

See `DependencyNodeIntegrationTests` for how such tests are authored.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/Construction/Solution.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/Construction/Solution.cs
@@ -1,0 +1,74 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    /// <summary>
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/Construction/Solution.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.IO;
+using System.Text;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    /// <summary>
+    /// Defines a <c>.sln</c> file to be created when using <see cref="ProjectLayoutTestBase"/>.
+    /// </summary>
+    public sealed class Solution : IEnumerable
+    {
```
tmeschter: Does this need to implement `IEnumerable` if the `IEnumerable.GetEnumerator()` implementation just throws an exception?

drewnoakes: For some definition of "need". But yes, it's intentional. If you want to use C# collection initialisers, you need to implement `IEnumerable` and have at least one `void Add(...)` method.

drewnoakes:
These tests run locally, both in the IDE and from the command line.

They aren't yet running on CI, though that's mostly orthogonal from a code perspective.

Once running, we will (probably) have an additional "integration" leg to the build.

PR #5254, opened 2019-08-07
Avoid assuming target framework is unique across configurations

davkean:
It's perfectly valid to have a multi-TFM project where two of the <TargetFrameworks> values map to the same underlying TFM. An example of this is the Microsoft.Toolkit.Uwp.Notifications project in the https://github.com/windows-toolkit/WindowsCommunityToolkit.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/957711/

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/ImmutablePropertyCollection.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/ImmutablePropertyCollection.cs
@@ -19,8 +19,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
 
         protected ImmutablePropertyCollection(IEnumerable<T> items)
         {
+            // While the majority of elements (items, properties, metadata) are guaranteed to be unique,
+            // Target Frameworks are not, filter out duplicates - NuGet uses the int-based indexer anyway.
+
+#pragma warning disable IDE0039 // We want to cache the delegate
+            Func<T, string> keySelector = i => GetKeyForItem(i);
+#pragma warning restore IDE0039
+
             _items = ImmutableList.CreateRange(items);
-            _itemsByName = _items.ToImmutableDictionary(i => GetKeyForItem(i), StringComparers.ItemNames);
+            _itemsByName = items.Distinct(keySelector, StringComparers.ItemNames)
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/ImmutablePropertyCollection.cs
```diff
@@ -19,8 +19,15 @@ internal abstract class ImmutablePropertyCollection<T> : IEnumerable<T> where T
 
         protected ImmutablePropertyCollection(IEnumerable<T> items)
         {
+            // While the majority of elements (items, properties, metadata) are guaranteed to be unique,
+            // Target Frameworks are not, filter out duplicates - NuGet uses the int-based indexer anyway.
+#pragma warning disable CA2214 // Do not call overridable methods in constructors
+            string keySelector(T i) => GetKeyForItem(i);
```
davidwengier: Nit: By defining this as a local function, a new `Func<T, string>` will be created each time its used. If you define it as a `Func<T, string>` you'll just get one.

davkean: Yeah that's pretty damn annoying and not obvious. The analzyer recommends the change.

davidwengier: I wouldn't normally have bothered commenting, as I'm hoping a future compiler will "fix" this, but I thought since you were lazily creating the HashSet that you were allocation-resistant :)

davkean: Yep this allocation will add up.

drewnoakes: I think we can reduce allocation here quite a bit more too. I'll file a PR.

drewnoakes: Filed #5257.

PR #5255, opened 2019-08-07
Bits ahead of more annotations

drewnoakes:
A bunch of small changes that should be quick to review, split out of another wave of nullability annotations.


src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Build/BuildUtilitiesTests.cs
```diff
@@ -24,7 +24,7 @@ public void GetProperty_MissingProperty()
 </Project>";
 
             var project = ProjectRootElementFactory.Create(projectXml);
-            var property = BuildUtilities.GetProperty(project, "NonExistantProperty");
+            var property = BuildUtilities.GetProperty(project, "NonExistentProperty");
```
davidwengier: Perhaps the typo was to ensure it would really be non existent 🤔

PR #5256, opened 2019-08-07
More nullability annotations

drewnoakes:
This cuts the number of `#nullable disable` pragmas almost exactly in half.

Before:

image

After:

image

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/AssertEx.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/AssertEx.cs
@@ -26,5 +26,30 @@ namespace Xunit
         {
             CollectionLength(collection.Cast<object>(), expectedCount);
         }
+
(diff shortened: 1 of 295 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.TestServices/AssertEx.cs
```diff
@@ -26,5 +24,30 @@ public static void CollectionLength(IEnumerable collection, int expectedCount)
         {
             CollectionLength(collection.Cast<object>(), expectedCount);
         }
+
+        public static void SequenceEqual<T>(IEnumerable<T> expected, IEnumerable<T> actual)
```
davkean: Context?

drewnoakes: I rewrote part of `VisualBasicNamespaceImportsList` to work better with nullability analysis, and that broke `VisualBasicNamespaceImportsListTests.UpdateNamespaceImportListTest`, so I rewrote that test which benefited from this assertion method.

PR #5257, opened 2019-08-08
Reduce allocations in ImmutablePropertyCollection<T>

drewnoakes:
- Previous implementation would enumerate input items twice. Most callers of this constructor would lazily construct objects, meaning twice the allocation. This implementation enumerates once.

  image

- Previous implementation would allocate a `HashSet<string>` to track unique keys, then produce an immutable dictionary as the backing store for this object. This implementation uses the backing store collection directly while it is being built, avoiding the need for the temporary set.

- Previous implementation would store data using immutable list and dictionary, both of which use many GC tracked node objects to store items. This implementation maintains immutability through the API, but uses mutable collections internally as they have lower memory consumption, and less fragmentation leading to better locality of reference.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/ProjectProperties.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/ProjectProperties.cs
@@ -12,10 +12,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
     internal class ProjectProperties : ImmutablePropertyCollection<IVsProjectProperty>, IVsProjectProperties
     {
         public ProjectProperties(IEnumerable<IVsProjectProperty> items)
-            : base(items)
+            : base(items, item => item.Name)
         {
         }
-
-        protected override string GetKeyForItem(IVsProjectProperty value) => value.Name;
     }
 }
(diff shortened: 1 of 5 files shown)


drewnoakes:
This build failed due to the virtual method call in the constructor. In the general case, this is a valid concern, however in this case no subtypes (currently) do any work in their constructors. This problem existed in the old implementation too, but the analysis didn't follow the logic through the indirection of the lambda.

I will update the implementation so that subtypes pass a lambda to the base to access keys, avoiding this inspection and preventing issues in future if subtypes change their behaviour somehow.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/ProjectProperties.cs
```diff
@@ -12,10 +12,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
     internal class ProjectProperties : ImmutablePropertyCollection<IVsProjectProperty>, IVsProjectProperties
     {
         public ProjectProperties(IEnumerable<IVsProjectProperty> items)
-            : base(items)
+            : base(items, item => item.Name)
```
davkean: This is cached once, right?

davkean: The removal of the virtual also avoids capture class & persumably caches the delegate.

drewnoakes: Appears to be, yes.

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQAcoKgW1QOzgBsiBjACwlIGsBLfAcwB8ABAJgEYBYAKF5YDMAAnZCAynGBCQQgEJRkEXgG9eQkcInAAFAEppQ4AojbkQgLwA+Icn3KAvr0d8egkWznGVaoegQB7GCoggBNPRW0WAFYAHhYOAAYAGniE6zA4fFI7Z3sgA

PR #5261, opened 2019-08-08
Bumping versions to unify them

### LINKED ISSUE #5209: Unify versions between Editors/AppDesigner and Managed/Managed.VS
These dlls have different versions for no actual reason. Just unify so that we have one version (choose the higher version to avoid binding redirect breaks):

https://github.com/dotnet/project-system/blob/master/build/Versions.props#L3-L6

dmonroym:
Fixes #5209


build/Versions.props
```diff
@@ -1,7 +1,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <VisualStudioVersion>16.3.0</VisualStudioVersion>
-    <ProjectSystemVersion>3.3.0</ProjectSystemVersion>
+    <ProjectSystemVersion>16.3.0</ProjectSystemVersion>
```
tmeschter: If they are going to have the same value, can't we just get rid of one entirely?

build/Versions.props
```diff
@@ -1,9 +1,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VisualStudioVersion>16.3.0</VisualStudioVersion>
-    <ProjectSystemVersion>3.3.0</ProjectSystemVersion>
-    <VersionBase Condition="'$(UseVisualStudioVersion)' == 'true'">$(VisualStudioVersion)</VersionBase>
```
davkean: Can we remove this UseVisualStudioVersion throughout the code base?

build/Versions.props
davkean: You'll need to update usage throughout the tree, for example: GenerateDependentAssemblyVersionFile.proj.

dmonroym: @davkean What is GenerateDependentAssemblyVersionFile.proj? I can't seem to find that anywhere

davkean: https://github.com/dotnet/project-system/blob/master/build/Scripts/GenerateDependentAssemblyVersionFile.proj

davkean: What I do when I change/remove a variable is the following:

```
project-system>findstr /sip /c:"VisualStudioVersion" *
```

dmonroym: Ah that's very useful. Somehow when I do a search for this file it's not indexed - seems to be true for most files under the /build directory.

davkean: Another thing to be aware of, is that you also need to search the toolset under:

```
C:\Users\[user]\.nuget\packages\roslyntools.repotoolset
```

To see if it consumes it.

davkean:
Note these files moved in https://github.com/dotnet/project-system/pull/5277.

davkean:
I think you'll want to rebase this change.

build/Versions.props
dmonroym: roslyntools.repotoolset doesn't consume it

build/Versions.props
dmonroym: `UseVisualStudioVersion` is no longer used

src/setup/VisualStudioEditorsSetup/Directory.Build.props
```diff
@@ -1,11 +1,8 @@
 <!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
```
davkean: These empty Build.props files can be deleted now and the projects will automatically pick up the parent.

PR #5270, opened 2019-08-09
Dependency node cleanup

drewnoakes:
Small changes pulled out of larger work on the dependencies node.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
@@ -112,22 +112,20 @@ namespace Microsoft.VisualStudio.ProjectSystem
         {
             Requires.NotNull(action, nameof(action));
 
-            using (var source = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, DisposalToken))
+            using var source = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, DisposalToken);
+            CancellationToken jointCancellationToken = source.Token;
+
+            try
             {
-                CancellationToken jointCancellationToken = source.Token;
+                T result = default!;
+                await _semaphore.ExecuteAsync(async () => { result = await action(jointCancellationToken); }, jointCancellationToken);
 
-                try
-                {
-                    T result = default!;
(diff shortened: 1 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/OnceInitializedOnceDisposedUnderLockAsync.cs
```diff
@@ -112,29 +114,29 @@ private async Task<T> ExecuteUnderLockCoreAsync<T>(Func<CancellationToken, Task<
...
+                if (!TryTranslateException(ex, cancellationToken, DisposalToken))
+                    throw;
             }
 
             throw Assumes.NotReachable();
         }
 
         private async Task ExecuteUnderLockCoreAsync(Func<CancellationToken, Task> action, CancellationToken cancellationToken = default)
         {
+            Requires.NotNull(action, nameof(action));
```
davkean: Why?

drewnoakes: Per 890e3b9fafad516597d102454c32e25cfa79c806, makes it consistent with the other methods in this class. Either they should all be removed, or all added. I am less bullish than you on the idea that NRT means we don't need runtime checks, for now at least.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Snapshot/AggregateDependenciesSnapshotProvider.cs
```diff
@@ -36,10 +36,7 @@ internal class AggregateDependenciesSnapshotProvider : IAggregateDependenciesSna
         /// <inheritdoc />
         public void RegisterSnapshotProvider(IDependenciesSnapshotProvider snapshotProvider)
         {
-            if (snapshotProvider == null)
-            {
-                return;
-            }
+            Requires.NotNull(snapshotProvider, nameof(snapshotProvider));
```
drewnoakes: @davidwengier I assume you're referring to this? If so then it's fine, no caller passes `null`.

davidwengier: Yep, that's the one.

PR #5271, opened 2019-08-09
Move props/targets into scripts

davkean:


davkean:
Kinda regretting calling this "scripts", I'm thinking maybe splitting out the props/targets and the projects ."proj" Thoughts?

drewnoakes:
Agree `proj` sounds better than `scripts`.

tmeschter:
I see two distinct categories of files here: .props and .targets that are imported into other projects, and .proj files that are "standalone" projects in their own right. What if the former went into an `imports` directory, and the latter went into `proj`?

davkean:
@tmeschter I went down that route and have merged that in https://github.com/dotnet/project-system/pull/5277.

PR #5277, opened 2019-08-12
New build folder layout

davkean:
This introduces a new build folder layout.

The easiest way to view the changes is to compare the folders:

Old: https://github.com/dotnet/project-system/tree/master/build
New: https://github.com/dotnet/project-system/tree/d674b2336f9b4ca3a8d27eb68812d8c56b28fbbf/build


build/proj/Build.proj
```diff
@@ -71,6 +71,7 @@
         Test=$(Test);
         Sign=$(Sign);
         IntegrationTest=$(IntegrationTest);
+        VersionsPropsPath=$(RepoRoot)build\import\Versions.props;
```
drewnoakes: This is required by the toolset? I don't see any other use of `VersionsPropsPath` in the PR.

.vsts-ci.yml
```diff
@@ -28,7 +28,7 @@ steps:
 - task: NuGetRestore@1
   displayName: Restore internal toolset
```
drewnoakes: Should this display name change too?

build/proj/Build.proj
davkean: Yes, it's needed as the default has changed and used by the toolset.

PR #5278, opened 2019-08-12
Collapse XLF files in PRs by default

drewnoakes:


.gitattributes
```diff
@@ -5,6 +5,9 @@
 *.vb linguist-vendored
 *.reg text
 
+# Collapse XLF files in PRs by default
+*.xlf linguist-generated=true
```
tmeschter: Is there a way to expand them all through the UI?

davkean: Yes.

PR #5285, opened 2019-08-12
Doc pipeline builds

davkean:
- Add names/urls to pipelines so that we can find them
- Rename integration.yml -> integration-tests.yml to be consistent with name of the build and existing unit-tests.yml.


build/ci/official.yml
```diff
@@ -1,5 +1,8 @@
+# Name: DotNet-Project-System
+# URL: https://dev.azure.com/devdiv/DevDiv/_build?definitionId=9675
```
dmonroym: Might be worth mentioning this is only available to FTEs

PR #5289, opened 2019-08-13
Improve dependency node MSBuild rule data handling

drewnoakes:
Steps towards larger refactoring of how the dependency node gets data from MSBuild.

Easiest to review commit by commit. Of note are:

- 7a04e2d5cf7175477fc3b8e32dca731bdf9d73dd which replaces the last usages of a .NET event with the Dataflow block @lifengl added a while back.
- 5394ef890ee6d172c71722bab874d03e41fda792 which removes redundant operations as joint rule name sets contain evaluated rule names
- 5b325d2c0d98a69550d2cb2cf6b6ec2a95534d46 which pulls duplicated code into a method and calls it twice
- 48b0d370e9ba314fb97f78650146f75e3ccd904c which avoids slicing the set of rules before calling handlers, as handlers only query items they're interested in anyway, which saves some allocation

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
@@ -190,7 +190,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscription
                 // Ensure the project doesn't unload during the update
                 await _tasksService.LoadedProjectAsync(async () =>
                 {
-                    // TODO pass _tasksService.UnloadCancellationToken into handler to reduce redundant work on unload
+                    // TODO pass _tasksService.UnloadCancellationToken into HandleAsync to reduce redundant work on unload
 
                     // Ensure the project's capabilities don't change during the update
                     using (ProjectCapabilitiesContext.CreateIsolatedContext(configuredProject, capabilities))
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
```diff
@@ -229,23 +216,18 @@ private IReadOnlyCollection<string> GetWatchedRules(RuleHandlerType handlerType)
...
-                // Slice project changes to include only rules the handler claims an interest in.
-                var projectChanges = projectUpdate.ProjectChanges
-                    .Where(x => handlerRules.Contains(x.Key))
-                    .ToImmutableDictionary();
-
-                if (projectChanges.Any(x => x.Value.Difference.AnyChanges))
-                {
-                    handler.Value.Handle(projectChanges, targetFrameworkToUpdate, changesBuilder);
-                }
+                handler.Value.Handle(projectUpdate.ProjectChanges, targetFrameworkToUpdate, changesBuilder);
```
davkean: Do handles expect this change behavior (previously only got changes related to them, now get everything)?

drewnoakes: Yes, handlers look up via their rules directly and ignore extra members, so it's equivalent in that regard. The behaviour change here is that _all_ handlers are now called regardless of whether they have any changes to process. The handlers themselves deal with this. Currently they allocate enumerators for added/changed/removed, but in a future change I'll protect against that by checking for zero counts before each `foreach`.

Currently there are two handler implementations. One in `DependenciesRuleHandlerBase` and an override of that which is specific to packages in `PackageRuleHandler`. They are very similar methods and I'm currently looking to merge them.

It's this handler where the logic we discussed on Monday will most likely live, hence my wanting to get all the bits leading up to this nice and clean.

drewnoakes:
Thanks @davkean for the review.

I'd usually leave this a bit longer for others to look at but I want to submit another PR on top of these changes. If anyone else has any feedback about this PR, leave it here and I'll address it.

PR #5291, opened 2019-08-13
Reduce allocation in DisposableBag and remove locking

drewnoakes:
- The cancellation support was unused so I removed it. It makes the type much smaller and simpler.
- The 'observable disposable' concept was unused, so has been removed.
- `AddDisposable` was renamed to `Add`, and object initializer support enabled. `RemoveDisposable` was also shortened. This looks better to me in usages.
- Removed `lock (this)` (which is a bad idea anyway as `this` is public which allows deadlocks if others locked on the instance too) from the implementation. Replaced with lock-free operations to guarantee safety and correctness.

Further questions:

- Should this be a stack, rather than a set, so that disposal is LIFO? I'm not sure it matters with our current usages, but in general it might be more correct. The current implementation gives no guarantee about disposal order, and it will in fact differ between executions. This could make certain classes of bug hard to find.
- Currently the implementation allows calling `Add`/`Remove` after the bag is disposed. `Add` will just dispose the argument immediately. I feel like throwing `ObjectDisposedException` in such cases would be better, as in general we shouldn't be using objects after disposing them.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/DisposableBag.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/DisposableBag.cs
@@ -53,7 +53,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Utilities
         /// Adds a value to be disposed of when this collection is disposed of.
         /// </summary>
         /// <param name="disposable">The value to be disposed of later. May be <c>null</c>.</param>
-        internal void AddDisposable(IDisposable? disposable)
+        public void AddDisposable(IDisposable? disposable)
         {
             if (disposable == null)
             {
@@ -82,7 +82,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Utilities
         /// <summary>
         /// Adds values to be disposed of when this collection is disposed of.
         /// </summary>
-        internal void AddDisposables(IEnumerable<IDisposable?> disposables)
+        public void AddDisposables(IEnumerable<IDisposable?> disposables)
         {
             Requires.NotNull(disposables, nameof(disposables));
 
@@ -96,7 +96,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Utilities
         /// Removes a disposable value from the collection.
         /// </summary>
         /// <param name="disposable">The value to remove. May be <c>null</c>.</param>
-        internal void RemoveDisposable(IDisposable? disposable)
+        public void RemoveDisposable(IDisposable? disposable)
         {
             if (disposable == null)
             {
@@ -113,7 +113,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Utilities
         /// Calls <see cref="IDisposable.Dispose"/> on all elements in a sequence,
         /// allowing the sequence itself or elements inside it to be null.
         /// </summary>
-        internal static void DisposeAllIfNotNull(IEnumerable<IDisposable?>? sequence, bool cacheSequence = false)
+        private static void DisposeAllIfNotNull(IEnumerable<IDisposable?>? sequence, bool cacheSequence = false)
         {
             if (sequence != null)
             {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/DisposableBag.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/DisposableBag.cs
@@ -4,45 +4,18 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
-using System.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.Utilities
 {
     /// <summary>
-    /// A class that tracks a set of disposable objects and a cancellation token for purposes
-    /// of managing the lifetime of a version-sync'd block join.
+    /// A class that tracks a set of disposable objects and disposes them collectively.
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/DisposableBag.cs
```diff
@@ -121,80 +71,39 @@ internal void AddDisposable(IDisposable? disposable)
...
-                    sequence = sequence.ToList();
-                }
-
-                foreach (IDisposable? item in sequence)
-                {
-                    item?.Dispose();
-                }
-            }
-        }
+        IEnumerator IEnumerable.GetEnumerator() => throw new NotSupportedException();
```
davidwengier: Externally this looks a lot like a collection (Add and Remove methods), is named like a collection (Bag), and implemented IEnumerable, so this feels like a trap that doesn't really help any. It wouldn't be a big deal for this to just be the following, would it?

```suggestion
        IEnumerator IEnumerable.GetEnumerator() => _disposables.GetEnumerator();
```

drewnoakes:
@davidwengier I can see it both ways. Throwing would draw out bugs where a type's dispose isn't synchronising properly with its other operations. For example, an object creating new dataflow subscriptions _after_ is has been disposed. Silently disposing the subscription isn't necessarily wrong, but creating subscriptions after being disposed would suggest that some async code should be acquiring the same lock used during dispose and testing `IsDisposed`. There may be other bugs hiding in that operation &mdash; hooking events (that would never unhook), registering for other kinds of notifications/services. Having an exception could help us find bugs like that.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/DisposableBag.cs
drewnoakes: What should it do if the bag is disposed? `_disposables` would be null in such a case.

davidwengier: `_disposables?.GetEnumerator() ?? Enumerator.Empty // you know what I mean :P` ?

It's probably not worth doing if its not trivial, but in 30 years if the .NET industry treats all enumerables with suspicion I reserve the right to say "I told you so" :P

lifengl:
interesting, this class was copied from CPS side, and we also did some work to reduce the memory usage (including getting rid of the cancellation token).  we didn't remove lock(this), which I don't like, but haven't got a good way to remove it.  I switched to use a List<>, instead of an immutable hashset on our side, so the lock free doesn't work directly.  I think the immutable set uses almost 10 times memory, and make it hard to check values in WinDBG.  (I constantly need enumerate those disposable bags to find dataflow links.)

drewnoakes:
@lifengl it's true that the `ImmutableHashSet<>` uses a lot more memory. I hadn't considered the benefit of having a contiguous array of items in WinDBG. Here's a lock-free implementation of `DisposableBag` that's backed by an array. Do you think it might be helpful?

```c#
internal sealed class DisposeBag : IDisposable
{
    private IDisposable[] _disposables = new IDisposable[8];
    private int _count;
    private int _state;

    private const int StateUnlocked = 0;
    private const int StateLocked = 1;
    private const int StateDisposed = 2;

    public void Add(IDisposable disposable)
    {
        var spin = new SpinWait();

        while (true)
        {
            var prior = Interlocked.CompareExchange(ref _state, StateLocked, StateUnlocked);

            switch (prior)
            {
                case StateUnlocked:
                {
                    // Acquired lock
                    if (_count == _disposables.Length)
                    {
                        // Resize
                        var larger = new IDisposable[_disposables.Length * 2];
                        Array.Copy(_disposables, larger, _disposables.Length);
                        _disposables = larger;
                    }

                    _disposables[_count++] = disposable;
                    Volatile.Write(ref _state, StateUnlocked);
                    break;
                }

                case StateLocked:
                {
                    // Lock held elsewhere, spin and try again
                    spin.SpinOnce();
                    break;
                }

                case StateDisposed:
                {
                    throw new ObjectDisposedException(nameof(DisposeBag));
                }
            }
        }
    }

    public void Remove(IDisposable disposable)
    {
        var spin = new SpinWait();

        while (true)
        {
            var prior = Interlocked.CompareExchange(ref _state, StateLocked, StateUnlocked);

            switch (prior)
            {
                case StateUnlocked:
                {
                    // Acquired lock
                    var index = Array.IndexOf(_disposables, disposable, 0, _count);

                    if (index != -1)
                    {
                        _count--;
                        if (index == _count)
                        {
                            _disposables[index] = null;
                        }
                        else
                        {
                            _disposables[index] = _disposables[_count];
                        }
                    }

                    Volatile.Write(ref _state, StateUnlocked);
                    break;
                }

                case StateLocked:
                {
                    // Lock held elsewhere, spin and try again
                    spin.SpinOnce();
                    break;
                }

                case StateDisposed:
                {
                    throw new ObjectDisposedException(nameof(DisposeBag));
                }
            }
        }
    }

    public void Dispose()
    {
        var spin = new SpinWait();

        while (true)
        {
            var prior = Interlocked.CompareExchange(ref _state, StateDisposed, StateUnlocked);

            switch (prior)
            {
                case StateDisposed:
                {
                    // Already disposed.
                    return;
                }

                case StateUnlocked:
                {
                    // Successfully transitioned to final state
                    foreach (var disposable in _disposables)
                    {
                        disposable.Dispose();
                    }

                    _disposables = null;
                    break;
                }

                default:
                {
                    // Lock held elsewhere, spin and try again
                    spin.SpinOnce();
                    break;
                }
            }
        }
    }
}
```

PR #5297, opened 2019-08-14
Annotations

drewnoakes:
Another wave of annotations, mostly through test code.

Before:

image

After:

image

eeb2abc679cc1b882848a3335abd63a8fbc17b39 contains all annotation changes in _production_ code and should be looked at individually. In particular, some additions to type constraints on internal interfaces related to service location, and annotating all service lookups as potentially returning `null`. This highlights places where we make an assumption that a given service exists (i.e. most of the other annotations in this commit).


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IPackageRestoreUnconfiguredInputDataSourceFactory.cs
```diff
@@ -33,13 +33,12 @@ public PackageRestoreUnconfiguredDataSource(IReceivableSourceBlock<IProjectVersi
 
             public IComparable DataSourceVersion { get; }
 
-            ISourceBlock<IProjectVersionedValue<object>> IProjectValueDataSource.SourceBlock { get; }
+            ISourceBlock<IProjectVersionedValue<object>> IProjectValueDataSource.SourceBlock => SourceBlock;
```
drewnoakes: Added implementation here, as it was never being set, which NRT analysis pointed out.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IProjectTreeCustomizablePropertyContextFactory.cs
```diff
@@ -4,34 +4,24 @@
...
+            string? itemType = null,
             bool isFolder = false,
             ProjectTreeFlags flags = default,
-            IImmutableDictionary<string, string> metadata = null)
+            IImmutableDictionary<string, string>? metadata = null)
         {
             var mock = new Mock<IProjectTreeCustomizablePropertyContext>();
             mock.Setup(x => x.ItemName).Returns(itemName ?? string.Empty);
-            mock.Setup(x => x.ItemType).Returns(itemType);
+            mock.Setup<string?>(x => x.ItemType).Returns(itemType);
```
drewnoakes: This happens a fair bit in our repo. The interface which defines `ItemType` is not in our repo and is not annotated. We know it can be `null`, so have to add the type explicitly here to avoid warnings in `Returns`.

drewnoakes:
I think the compile error here is a Roslyn bug. Investigating.

drewnoakes:
Firstly, the diagnostic is not showing up in VS, only in build output. This is probably due to using a specific version of Roslyn via a package, versus the version used in VS for analysis. Maybe.

Secondly, here's a minimal repro of what's happening:

```c#
string s = "Not null";
Assert.False(s.Equals(null));
Assert.Equal("NOT NULL", s.ToUpper()); // warns that s can be null
```

Reducing further:

```c#
string s = "Not null";
bool b = s.Equals(null);
s.ToUpper(); // warns
```

The analysis considers the comparison with `null` as suggesting the reference can be null, even though in our original example we're making an assertion about the behaviour of the equals method implementation.

In this case it's plain to see that it definition does _not_ equal null, as it's assigned within the same method.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IProjectTreeCustomizablePropertyContextFactory.cs
davidwengier: I'm getting an IDE0001 "Name can be simplified" in the error list for another spot where `<string?>` is specified like this, for this reason. Do you not get that here? Can you add a `#pragma disable` perhaps?

drewnoakes: On the changes in this PR? Can you give a file and line number?

davidwengier: No, from a previous PR presumably. Its here: https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptingPropertyValueProviderBase.cs#L29

Removing the `<string?>` brings back the nullable warning, as you would expect.

This is using VS master, 29213.130

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/VsShellUtilitiesHelper.cs
```diff
@@ -30,9 +30,9 @@ public VsShellUtilitiesHelper(IProjectThreadingService threadingService)
         {
             await _threadingService.SwitchToUIThread();
 
-            IVsAppId vsAppId = await vsAppIdService.GetValueAsync();
+            IVsAppId? vsAppId = await vsAppIdService.GetValueAsync();
 
-            if (ErrorHandler.Succeeded(vsAppId.GetProperty((int)VSAPropID.VSAPROPID_ProductSemanticVersion, out object oVersion)) &&
+            if (ErrorHandler.Succeeded(vsAppId!.GetProperty((int)VSAPropID.VSAPROPID_ProductSemanticVersion, out object oVersion)) &&
```
davidwengier: Whats the logic of making GetValueAsync() return a nullable, and then immediately using the damnit operator?

drewnoakes: This PR makes all the `GetService` calls (and related APIs) nullable. However we use those services in a _lot_ of places, and I don't think we are realistically going to add null checks everywhere, especially for services so integral to VS that if they're absent we've got bigger problems. The `!` here reflects that assumption.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IProjectTreeCustomizablePropertyContextFactory.cs
drewnoakes: I don't see this in 16.2.2 but I do in newer builds. Filed https://github.com/dotnet/roslyn/issues/37980.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/VsShellUtilitiesHelper.cs
davidwengier: I get that.. but wouldn't it make more sense to mark the API as not nullable? If we don't expect people to null check, then shouldn't the stated return type reflect that?

drewnoakes: I guess it's subjective. There are cases where we do check whether the returned service is null. In future the analyzer may change to suggest removing redundant null checks, which would then break working code. My preference (at this early stage of using NRT) is to try and reflect reality in the type system, like the rest of the type system.

PR #5298, opened 2019-08-14
Use shared project for build files

drewnoakes:
This avoids having to keep the list of solution items up to date.

Having these items in the solution means their contents will be searched by find-in-files.

image

### DIFF
--- a/build/build.shproj
+++ b/build/build.shproj
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>41d1fa55-b719-4e8d-9d79-f7610004d496</ProjectGuid>
+    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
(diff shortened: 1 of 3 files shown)


build/build.shproj
```diff
@@ -0,0 +1,13 @@
...
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>41d1fa55-b719-4e8d-9d79-f7610004d496</ProjectGuid>
+    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
+  <PropertyGroup />
+  <Import Project="build.projitems" Label="Shared" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
```
drewnoakes: I suspect we may be able to remove some of these imports.

drewnoakes:
I'll follow this PR with two others:

1. Convert the documentation project to shproj
2. Convert `Microsoft.VisualStudio.ProjectSystem.Managed.TestServices` to a shared project and consume it in three test assemblies (so that we are actually dogfooding shared projects in our solution.)

davidwengier:
Do shared projects have any advantages over the current Documentation project?

I get it for the one that will actually be shared of course :P

drewnoakes:
> Do shared projects have any advantages over the current Documentation project?

- Consistency with this `build` project (though of course `build.csproj` would have been an option too)
- Not having (as far as I know) targets for build/test/clean/etc

Issue #4186, opened 2018-10-28
Can't use msbuild /restore against this repo


Pilchie:
With VS 2017 Version 15.9Preview 4 and current master (https://github.com/dotnet/project-system/commit/c2e466aabd03b83de4a3b0f44a488cc6d13ce790)

1. `nuget.exe locals all -clear`
2. `git clean -xdf`
3. `MSBuild.exe /nologo /bl /clp:Summary /m /restore /v:m .\ProjectSystem.sln`

Result:
Build failed.

Related to #3595.

Build warnings/errors (binlog attached)
```
Build FAILED.

C:\code\project-system\src\DeployTestDependencies\DeployTestDependencies.csproj : warning NU1604: Project dependency Microsoft.NET.Test.Sdk does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [C:\code\project-system\ProjectSystem.sln]
C:\code\project-system\src\DeployTestDependencies\DeployTestDependencies.csproj : warning NU1604: Project dependency xunit.runner.visualstudio does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [C:\code\project-system\ProjectSystem.sln]
C:\code\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests\Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests.csproj : warning NU1602: Microsoft.VisualStudio.ProjectSystem.Managed.VS does not provide an inclusive lower bound for dependency Microsoft.VSSDK.BuildTools. An approximate best match of Microsoft.VSSDK.BuildTools 12.0.21005 was resolved. [C:\code\project-system\ProjectSystem.sln
]
C:\code\project-system\src\DeployTestDependencies\DeployTestDependencies.csproj : warning NU1602: Microsoft.VisualStudio.ProjectSystem.CSharp.VS does not provide an inclusive lower bound for dependency Microsoft.VSSDK.BuildTools. An approximate best match of Microsoft.VSSDK.BuildTools 12.0.21005 was resolved. [C:\code\project-system\ProjectSystem.sln]
[...]

PR #5300, opened 2019-08-14
Remove need to triple-restore during build

davkean:
Fixes: https://github.com/dotnet/project-system/issues/4186.

Previously we would need to triple restore during build:

1) Restore RoslynTools.RepoToolset to get Build.proj
2) Call Build.proj and restore Tools.proj
3) Restore ProjectSystem.sln

This prevented us from opening ProjectSystem.sln on a clean enlistment and have it "just work" because required targets would be missing.

This simplifies above, by:

- Porting build.proj into the repro ("called CoreBuild.proj") and simplifying it by removing concepts we don't use (such as build orchestration) 
- Make sure that all the projects have a package dependency on everything that requires to build the tree
- Continue to use RoslynTools.RepoToolset, but not require it to open the solution

In the end, the change is not as huge as the effort but I've done it in individual steps (via commits) if you'd like to understand how I ended up with below.


src/Directory.Build.props
```diff
@@ -1,8 +1,17 @@
 <!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
 <Project>
+  <PropertyGroup>
+    <IsUnitTestProject>false</IsUnitTestProject>
+    <IsUnitTestProject Condition="$(MSBuildProjectName.EndsWith('.UnitTests')) or $(MSBuildProjectName.EndsWith('.Tests'))">true</IsUnitTestProject>    
+    <IsIntegrationTestProject>false</IsIntegrationTestProject>
+    <IsIntegrationTestProject Condition="$(MSBuildProjectName.EndsWith('.IntegrationTests'))">true</IsIntegrationTestProject>
+    <IsTestProject>false</IsTestProject>
+    <IsTestProject Condition="'$(IsUnitTestProject)' == 'true' or '$(IsIntegrationTestProject)' == 'true' or '$(IsPerformanceTestProject)' == 'true'">true</IsTestProject>
```
drewnoakes: `IsPerformanceTestProject`?

build/import/Versions.props
```diff
@@ -15,9 +15,12 @@
 
     <!-- Toolset -->
     <RoslynToolsRepoToolsetVersion>1.0.0-beta-62705-01</RoslynToolsRepoToolsetVersion>
+    <RoslynToolsModifyVsixManifestVersion>1.0.0-beta-62327-02</RoslynToolsModifyVsixManifestVersion>
+    <RoslynToolsSignToolVersion>1.0.0-beta-62414-01</RoslynToolsSignToolVersion>
     <RoslynDependenciesProjectSystemOptimizationDataVersion>2.6.0-beta1-62205-02</RoslynDependenciesProjectSystemOptimizationDataVersion>
     <MicrosoftDotNetIBCMergeVersion>4.7.1-alpha-00001</MicrosoftDotNetIBCMergeVersion>
     <MicrosoftDiaSymReaderPdb2PdbVersion>1.1.0-beta1-62624-01</MicrosoftDiaSymReaderPdb2PdbVersion>
+    <MicroBuildPluginsSwixBuildVersion>1.0.147</MicroBuildPluginsSwixBuildVersion>
```
drewnoakes: Why was this pulled out to a property? I don't see any other usage of it in this PR.

Though I can't see usages of `RoslynToolsModifyVsixManifestVersion` or `RoslynToolsSignToolVersion` either.

drewnoakes:
> we no longer need to run build.cmd before you open ProjectSystem.sln

👍 👍 👍

src/Directory.Build.props
davkean: I can delete that usage, I've removed this.

build/import/Versions.props
davkean: These are used by RoslynTools.RepoToolset and without it the project fails to build because it can't find the targets files from it.

PR #5303, opened 2019-08-15
 Allow signing on VS 2019

davkean:
This allows signing on VS 2019, instead of limiting to VS 2017.

Originally we were going to consume this by upgrading to the latest version of RoslynTools.RepoToolset. I spent 2 days on that port and gave up; it's a significant work item due to a large number of breaking changes. I timeboxed and ran out of time.

Instead, I've gone down the approach of copying the Sign.proj from the toolset, and slightly changing it. Both are in two separate commits.

### DIFF
--- a/build/proj/Sign.proj
+++ b/build/proj/Sign.proj
@@ -17,9 +17,7 @@
   <Target Name="Sign">
 
     <PropertyGroup>
-      <!-- VS2019 and later place msbuild.exe under "MSBuild\Current", while VS2017 places it under "MSBuild\15.0". -->
-      <_MSBuildExePath Condition="Exists('$(VSINSTALLDIR)\MSBuild\Current\Bin\msbuild.exe')">$(VSINSTALLDIR)\MSBuild\Current\Bin\msbuild.exe</_MSBuildExePath>
-      <_MSBuildExePath Condition="'$(_MSBuildExePath)' == ''">$(VSINSTALLDIR)\MSBuild\15.0\Bin\msbuild.exe</_MSBuildExePath>
+      <_MSBuildExePath>$(VSINSTALLDIR)\MSBuild\$(MSBuildToolsVersion)\Bin\msbuild.exe</_MSBuildExePath>
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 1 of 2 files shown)


build/proj/Sign.proj
```diff
@@ -0,0 +1,39 @@
...
+  <Import Project="..\import\Versions.props" />
+  <Import Project="..\import\NuGet.props" />
+  <Import Project="..\import\RepoLayout.props" />
+
+  <Target Name="Sign">
+
+    <PropertyGroup>
+      <!-- VS2019 and later place msbuild.exe under "MSBuild\Current", while VS2017 places it under "MSBuild\15.0". -->
+      <_MSBuildExePath Condition="Exists('$(VSINSTALLDIR)\MSBuild\Current\Bin\msbuild.exe')">$(VSINSTALLDIR)\MSBuild\Current\Bin\msbuild.exe</_MSBuildExePath>
+      <_MSBuildExePath Condition="'$(_MSBuildExePath)' == ''">$(VSINSTALLDIR)\MSBuild\15.0\Bin\msbuild.exe</_MSBuildExePath>
```
davidwengier: I think you can remove the line above, and change this to use `$(MSBuildToolsVersion)` to be simpler and work for both versions...
```suggestion
      <_MSBuildExePath Condition="'$(_MSBuildExePath)' == ''">$(VSINSTALLDIR)\MSBuild\$(MSBuildToolsVersion)\Bin\msbuild.exe</_MSBuildExePath>
```

davkean: Much cleaner.

build/proj/Sign.proj
```diff
@@ -0,0 +1,37 @@
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project DefaultTargets="Sign">
+  <!--
+    Required parameters:
+      ConfigFile                    SigntToolData.json path.
```
drewnoakes:
```suggestion
      ConfigFile                    SignToolData.json path.
```

drewnoakes:
Does this work on VS2017 too?

davkean:
MSBuildToolsVersion should do so.

davkean:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2945280 is the sign build that is verifying this change.

PR #5304, opened 2019-08-15
Move official build to VS 2019

davkean:
This will enable us to use some of the new VS 2019 features, such as .editorconfig support.


davkean:
Here's the build verifying this: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2945300.

PR #5305, opened 2019-08-15
Defer editor creation for Partial Load Mode

davidwengier:
Tested all editors with a delayed design time build, and both Settings and Resource editors rely on build results, and the App Designer includes the Resource Editor.

Fixes [AB#952010](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/952010)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
+++ b/src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.pkgdef
@@ -135,6 +135,7 @@
 "DisplayName"="#1300"
 "EditorTrustLevel"=dword:00000001
 "Package"="{67909B06-91E9-4F3E-AB50-495046BE9A9A}"
+"DeferUntilIntellisenseIsReady"=dword:00000001
 @="Project Designer"
 
 [$RootKey$\Editors\{04b8ab82-a572-4fef-95ce-5222444b6b64}\LogicalViews]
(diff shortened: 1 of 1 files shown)


davkean:
@davidwengier I think we'll want to raise this with the operation progress group, preventing the entire app designer from opening for a couple of pages is not ideal.

davkean:
Given that the Resource Editor is just a hosted editor, what is the behavior if you disable resource editor but not the appdesigner?

davidwengier:
If the resource file exists, the editor just blocks the UI until the design time build is complete, so its not actually the worst experience in the world (depending on how long the DTB takes of course). Creating the resource file however, which the property pages can do if it doesn't exist, caused errors.

> I think we'll want to raise this with the operation progress group, preventing the entire app designer from opening for a couple of pages is not ideal.

I agree, and I can reply to the thread, but I suspect they'll simply tell us to opt in (which this PR does) and then we can opt out once we've fixed our editor to be resilient to design time build results being delayed. The fact that the editor hosts multiple pages or user controls I would think is an implementation detail of ours.

I don't think its worth holding up this PR though, as all it does is apply the default behavior which these editors will get anyway, whilst protecting us from that default changing.

davidwengier:
Attn @dotnet/project-system as mentioned in the team meeting I'm merging this today.

davkean:
Can we open a bug to track the open issue around AppDesigner?

davkean:
> Creating the resource file however, which the property pages can do if it doesn't exist, caused errors.

Why does this error? I would expect this to work.

davidwengier:
> Can we open a bug to track the open issue around AppDesigner?

Logged #5317

> Why does this error? I would expect this to work.

In hindsight that may have been my build. I now get the same error as mentioned in the issue created, on Friday I saw a different Assert but I possibly had a debug build of CPS at the time.

PR #5310, opened 2019-08-16
Force VS 2019 for building repository

davkean:
We are moving across to consuming VS 2019-only features such as compiler .editorconfig suppport.


davkean:
We originally had this but removed it due Spanish/Signed builds still using VS 2017. These have now moved to VS 2019.

davkean:
Signed build version of this: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2949166.

PR #5312, opened 2019-08-16
Split unit-tests.yml into template

davkean:
This splits common logic from unit-test into a parameter-driven template file.

This unifies the build logic between Spanish and Windows runs, by changing the following:

1) NuGet Cache is cleared in Windows runs, it was only running on Spanish previously which was an oversight. This caches when our sources don't contain a package that we've referred to.

2) Spanish runs also test sign, similar to Windows.

### DIFF
--- a/build/ci/unit-tests-template.yml
+++ b/build/ci/unit-tests-template.yml
@@ -1,0 +1,49 @@
+# Template used by unit-tests.yml
+jobs:
+- job: ${{ parameters.name }}
+  pool: ${{ parameters.pool }}
+  timeoutInMinutes: 20
+  steps:
+    - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /clearnugetcache /no-deploy /no-integration /no-ibc /configuration ${{ parameters.configuration }}
+    - task: PublishBuildArtifacts@1
+      inputs:
+        PathtoPublish: '$(Build.SourcesDirectory)\artifacts\${{ parameters.configuration }}\log'
+        ArtifactName: '${{ parameters.configuration }} unit test logs'
+        publishLocation: Container
+      continueOnError: true
+      condition: failed()
+
+    - task: PublishBuildArtifacts@1
+      inputs:
+        PathtoPublish: '$(Build.SourcesDirectory)\artifacts\${{ parameters.configuration }}\bin'
+        ArtifactName: '${{ parameters.configuration }} bin folder'
+        publishLocation: Container
+      continueOnError: true
+      condition: failed()
+
+    - task: PublishBuildArtifacts@1
+      inputs:
+        PathtoPublish: '$(Build.SourcesDirectory)\artifacts\${{ parameters.configuration }}\VSSetup'
+        ArtifactName: '${{ parameters.configuration }} VSSetup folder'
+        publishLocation: Container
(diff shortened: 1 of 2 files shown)


build/ci/unit-tests.yml
```diff
@@ -38,102 +38,23 @@ pr:
...
 
-- job: Spanish
-  pool:
-    name: NetCorePublic-Pool
-    queue: BuildPool.Windows.Amd64.VS2019.Pre.ES.Open
-  variables:
-    _configuration: Debug
-  timeoutInMinutes: 20
-  steps:
-    - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /diagnostic /no-sign /no-deploy /no-integration /no-ibc /clearnugetcache /configuration $(_configuration)
```
drewnoakes: This used to pass `/no-sign` but will now use `/sign`.

build/ci/unit-tests.yml
```diff
@@ -38,102 +38,23 @@ pr:
...
-  strategy:
-    maxParallel: 4
-    matrix:
-      Debug:
-        _configuration: Debug
-      Release:
-        _configuration: Release
-  timeoutInMinutes: 20
-  steps:
-    - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /no-deploy /no-integration /no-ibc /configuration $(_configuration)
```
drewnoakes: This used to not pass `/clearnugetcache`, but will now have that option.

build/ci/unit-tests-template.yml
```diff
@@ -0,0 +1,48 @@
...
+      continueOnError: true
+      condition: failed()
+    - task: PublishTestResults@2
+      inputs:
+        testRunner: 'xUnit'	
+        testResultsFiles: '**/*.xml' 	
+        searchFolder: '$(Build.SourcesDirectory)\artifacts\${{ parameters.buildConfiguration }}\TestResults'	
+        configuration: '${{ parameters.buildConfiguration }}'	
+        publishRunAttachments: true	
+        failTaskOnFailedTests: true
```
drewnoakes: This used to not be specified for the Spanish build.

build/ci/unit-tests-template.yml
```diff
@@ -0,0 +1,48 @@
+# Template used by unit-tests.yml
+
+jobs:
+- job: ${{ parameters.jobName }} ${{ parameters.buildConfiguration }}
```
drewnoakes: This is no longer building debug for the en-US locale.

build/ci/unit-tests.yml
davkean: See description.

build/ci/unit-tests.yml
davkean: See description.

build/ci/unit-tests-template.yml
davkean: It was an oversight.

build/ci/unit-tests-template.yml
davkean: I don't follow - that's what the first job does in unit-tests.yml.

drewnoakes: My mistake. The first item appears higher up in between two large deleted chunks.

PR #5313, opened 2019-08-16
Consistently call build.cmd 

davkean:
This includes https://github.com/dotnet/project-system/pull/5312 so just review https://github.com/dotnet/project-system/commit/25dac9a10c473f35767a16c1ae11c6781bc07c6e.

his moves official build from BatchScript -> CmdLine ("script"). The docs are not clear on the differences and cannot find the source code for BatchScript, but this consistently brings it inline with unit-tests and integration-tests.


davkean:
This hit the AV.

/azp run

PR #5318, opened 2019-08-19
Fix DependenciesSnapshotProvider leak

### LINKED ISSUE #5315: Memory leak: entire project leaked after openning/closing solution quickly
Memory leak: open Roslyn.sln and close it quickly, the entire solution memory leaked through this path:

```
->  121f9988 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.AggregateDependenciesSnapshotProvider
            ->  121f99a4 System.Collections.Generic.Dictionary`2[[System.String, mscorlib],[System.ValueTuple`2[[Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.IDependenciesSnapshotProvider, Microsoft.VisualStudio.ProjectSystem.Managed],[System.IDisposable, mscorlib]], mscorlib]]
            ->  208c0b54 System.Collections.Generic.Dictionary`2+Entry[[System.String, mscorlib],[System.ValueTuple`2[[Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.IDependenciesSnapshotProvider, Microsoft.VisualStudio.ProjectSystem.Managed],[System.IDisposable, mscorlib]], mscorlib]][]
            ->  24f48674 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependenciesSnapshotProvider
            ->  1169a224 Microsoft.VisualStudio.ProjectSystem.UnconfiguredProjectCommonServices
            ->  1167dd40 Microsoft.VisualStudio.ProjectSystem.UnconfiguredProjectImpl
```

The leak appears to be a race condition.  A DependenciesSnapshotProvider registers itself to a global table in the constructor, but the clean-up logic was registered during the initialization code.  When I open Roslyn and close it quickly, the provider never has a chance to run the initialization code, and it leaks under that condition.

drewnoakes:
Fixes #5315 and [AB#956763](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/956763).

@lifengl identified a memory leak from `AggregateDependenciesSnapshotProvider` to `DependenciesSnapshotProvider` to `UnconfiguredProjectImpl` which was keeping the solution in memory even after closing.

`AggregateDependenciesSnapshotProvider` is global in scope, hence doesn't unload with the project. `DependenciesSnapshotProvider` instances register and unregister themselves with that global instance. It is important that this unregistration works reliably.

The problem was that registration occurred in the `DependenciesSnapshotProvider` constructor, and deregistration occurred in `Dispose`. If project load is then cancelled, the initialize method is not called, and neither is dispose, resulting in a dangling registration.

Further investigation shows several other similar pathways through which the type may fail to clean up after itself in various ways. Reasoning about the behaviour of this class is quite difficult as `DependenciesSnapshotProvider` is a moderately complex class with several competing responsibilities, and various lifetimes and locks. 

This PR breaks it into smaller components and makes lifetimes, locks and responsibilities more explicit.

Previously the shutdown pathways had implicit relationships with the initialization pathways that create stuff to unwind/release. This PR makes these relationships explicit by using a `DisposableBag` and putting `DisposableDelegate` instances into it whenever work that must be unwound occurs.

I've also removed a global lock from the common read-only pathways through `AggregateDependenciesSnapshotProvider`.

The diff is pretty horrendous in full but I've done my best to make each commit small and clear.


davidwengier:
I've read through this, and don't have any comments, but I don't think I can actually comment on what this is doing. I can see how https://github.com/dotnet/project-system/pull/5318/commits/3e77f49c2aea231acb68f28571d7255e0ae69016 moves registration out of the constructor, which presumably fixes the leak, and that looks good, but overall I think the PR is too big for me to reason about without a much deeper understanding of the code. Sorry :)

drewnoakes:
Thanks @davidwengier. I spent the whole of yesterday banging my head against this. It was hard to unravel the code into logical chunks. I understand it's difficult to review.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -160,15 +177,18 @@ public Task OnProjectFactoryCompletedAsync()
...
-                // This host object subscribes to configured project evaluation data for its own purposes.
-                SubscribeToConfiguredProjectEvaluation(
-                    _activeConfiguredProjectSubscriptionService,
-                    OnActiveConfiguredProjectEvaluatedAsync);
-
-                // Each of the host's subscribers are initialized.
-                return Task.WhenAll(
-                    Subscribers.Select(
-                        subscriber => subscriber.InitializeSubscriberAsync(this, _activeConfiguredProjectSubscriptionService)));
+                lock (_lock)
```
davkean: This is a little worrying previously it wouldn't call 3rd parties under a lock now it does. What is this protecting against?

drewnoakes: `IDependencyCrossTargetSubscriber` is internal and not an extension point for 3rd party code. It has two implementations, neither of which do anything particularly heavyweight. This lock is scoped to the unconfigured project so I expect contention to be low.

The lock here synchronises calls to `AddSubscriptions` with `ReleaseSubscriptions`. Given this is an init path, it may be possible to remove it. I'll investigate.

drewnoakes: Looking at this again that lock can be removed. It was intended to protect the initialisation of subscribers (which can happen via one of many pathways) but that protection occurs in the `Subscribers` property directly anyway, so this longer-lived lock is removed in 9e5c22ea7d39f84e706a368d38a66f79df8625c3.

PR #5321, opened 2019-08-20
Fix paths to pipelines on solution file

dmonroym:


drewnoakes:
@dmonroym https://github.com/dotnet/project-system/pull/5298 is intended to fix this kind of thing. I'll merge it now.

davidwengier:
How does that solve this?

drewnoakes:
@davidwengier the shared project is supposed to include everything on disk without needing to edit the solution file (which is very disruptive when switching between branches/commits). If it's not achieving that we probably need to update the globs.

davidwengier:
My mistake, it does do that, but only when the project is loaded. It doesn't respond to live changes on disk.

davkean:
@davidwengier Neither does the solution.

PR #5323, opened 2019-08-20
Annotations

drewnoakes:
The only behaviour change is in 6d45d37e2348daf1b31517764795882d1965dd2d, where a null value will now throw earlier.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IActiveConfiguredProjectProviderFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IActiveConfiguredProjectProviderFactory.cs
@@ -18,14 +18,14 @@ namespace Microsoft.VisualStudio.ProjectSystem
         }
 
         public static IActiveConfiguredProjectProvider Create(
-            Func<ProjectConfiguration>? getActiveProjectConfiguration = null,
-            Func<ConfiguredProject>? getActiveConfiguredProject = null)
+            Func<ProjectConfiguration?>? getActiveProjectConfiguration = null,
+            Func<ConfiguredProject?>? getActiveConfiguredProject = null)
         {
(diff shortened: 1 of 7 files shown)


drewnoakes:
This requires part of #5322 to build.

src/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IActiveConfiguredProjectProviderFactory.cs
```diff
@@ -18,14 +18,14 @@ public static IActiveConfiguredProjectProvider ImplementActiveProjectConfigurati
...
 
         public static IActiveConfiguredProjectProvider Create(
-            Func<ProjectConfiguration>? getActiveProjectConfiguration = null,
-            Func<ConfiguredProject>? getActiveConfiguredProject = null)
+            Func<ProjectConfiguration?>? getActiveProjectConfiguration = null,
+            Func<ConfiguredProject?>? getActiveConfiguredProject = null)
         {
             var mock = new Mock<IActiveConfiguredProjectProvider>();
-            mock.SetupGet(p => p.ActiveProjectConfiguration)
+            mock.SetupGet<ProjectConfiguration?>(p => p.ActiveProjectConfiguration)
```
davkean: Why was this needed?

drewnoakes: You'll see several of the mock setup methods annotated in this way.

The problem is that the interfaces they're mocking aren't annotated, so we have to tell the compiler when mocked members have some kind of nullability.

In this particular case, a caller of `Create` was using `!` on the `getActiveProjectConfiguration` argument because its `Func` returned a nullable value. The docs say `ActiveProjectConfiguration` can be null, so I updated this mock construction to reflect that.

If `IActiveConfiguredProjectProvider` was annotated, the type argument here would be redundant.

PR #5324, opened 2019-08-20
Bump Roslyn analyzers to 2.9.4

drewnoakes:


build/import/Packages.targets
```diff
@@ -115,13 +115,13 @@
...
-    <PackageReference Update="Microsoft.CodeAnalysis.PublicApiAnalyzers"                              Version="2.9.3" />
-    <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="2.9.4-beta1.final" />
+    <PackageReference Update="Microsoft.CodeAnalysis.Analyzers"                                       Version="2.9.4" />
+    <PackageReference Update="Microsoft.CodeQuality.Analyzers"                                        Version="2.9.4" />
+    <PackageReference Update="Microsoft.NetFramework.Analyzers"                                       Version="2.9.4" />
+    <PackageReference Update="Microsoft.NetCore.Analyzers"                                            Version="2.9.4" />
+    <PackageReference Update="Microsoft.CodeAnalysis.BannedApiAnalyzers"                              Version="2.9.4" />
+    <PackageReference Update="Microsoft.CodeAnalysis.PublicApiAnalyzers"                              Version="2.9.4" />
+    <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="2.9.4" />
```
dmonroym: Would it be better to define this number in Versions.props so we only have to change it once next time?

davkean: We've only used versions inside Versions.props for things that are absolutely required and used elsewhere, typically in RoslynTools.RepoToolset. If I had a choice there would be zero properties driving packages, so much easier to reason about if you see it all inline. The new solution wide package design looks almost identical to what we've got in this file.

PR #5326, opened 2019-08-20
Design time inputs build manager bridge

### LINKED ISSUE #5232: Create DesignTimeInputsBuildManager that consumes the design time inputs changes
This should inherit from UnconfiguredProjectHostBridge and fire events.

Part of #4163

davidwengier:
Creates a class that bridges between the VSLangProj/UI thread components, and the design time inputs/data flow components. In an ideal world, our VSBuildManager itself would probably do this job, but that can't be done as the base class (UnconfiguredProjectHostBridge) does all of the good work, and VSBuildManager inherits from ConnectionPointContainer.

This completes the wire up for all TempPE work. I'm going to do another PR to follow up and work out what is happening with file paths (whether to use relative or full) but thought it would be easier to keep the PRs separate.

This includes initialization of the system (ie, turning it on), which is currently done lazily when the first designer asks for the first file. In a "standard" windows forms app this means that the first compilations will happen on the UI thread, and from then on the background compilations will keep up with changes. The first background compilation ends up having no work to do, because the UI thread has beaten it to the punch.

Fixes #5232


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        /// <summary>
+        /// Gets the XML that describes a TempPE DLL, including building it if necessary
+        /// </summary>
+        public async Task<string> GetDesignTimeInputXmlAsync(string file)
+        {
+            // Not using use the ThreadingService property because unit tests
+            await _threadingService.SwitchToUIThread();
+
+            if (!SkipInitialization)
```
davkean: Why can't you initialize this in tests?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
```diff
@@ -0,0 +1,141 @@
...
+                await InitializeAsync();
+            }
+
+            DesignTimeInputsDelta value = AppliedValue.Value;
+
+            return await _designTimeInputsCompiler.GetDesignTimeInputXmlAsync(file, value.TempPEOutputPath, value.SharedInputs);
+        }
+
+        /// <summary>
+        /// ApplyAsync is called on the UI thread and its job is to update AppliedValue to be correct based on the changes that have come through data flow after being processed
```
davkean: Is this correct? It says "called on the UI thread", but the first thing it does is switch to the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
```diff
@@ -0,0 +1,141 @@
...
+        protected override Task DisposeCoreAsync(bool initialized)
+        {
+            if (initialized)
+            {
+                _designTimeInputsCompiler.Dispose();
+            }
+            return base.DisposeCoreAsync(initialized);
+        }
+
+        /// <summary>
```
davkean: Are these unique comments or just copied from the base? Do we need to copy them? Can't we just look at the base definitions of them?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
```diff
@@ -0,0 +1,141 @@
...
+        protected override Task InitializeInnerCoreAsync(CancellationToken cancellationToken)
+        {
+            return Task.CompletedTask;
+        }
+
+        protected override Task DisposeCoreAsync(bool initialized)
+        {
+            if (initialized)
+            {
+                _designTimeInputsCompiler.Dispose();
```
davkean: This is an import, are you deliberately disposing this? MEF components don't own the lifetime of their imports.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridgeTests.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        internal class TestDesignTimeInputsBuildManagerBridge : DesignTimeInputsBuildManagerBridge
+        {
+            public TestDesignTimeInputsBuildManagerBridge(IProjectThreadingService threadingService, IDesignTimeInputsChangeTracker designTimeInputsChangeTracker, IDesignTimeInputsCompiler designTimeInputsCompiler, VSBuildManager buildManager)
+                : base(threadingService, designTimeInputsChangeTracker, designTimeInputsCompiler, buildManager)
+            {
+            }
+
+            public Task TestProcessAsync(DesignTimeInputsDelta delta)
```
davkean: This name is not descriptive - can you just call this ApplyAsync?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -30,13 +32,16 @@ internal VSBuildManager(IProjectThreadingService threadingService, IUnconfigured
...
-
             Project = new OrderPrecedenceImportCollection<VSLangProj.VSProject>(projectCapabilityCheckProvider: unconfiguredProjectServices.Project);
+            DesignTimeInputsBuildManagerBridge = new OrderPrecedenceImportCollection<IDesignTimeInputsBuildManagerBridge>(projectCapabilityCheckProvider: unconfiguredProjectServices.Project);
         }
 
         [ImportMany(ExportContractNames.VsTypes.VSProject)]
         internal OrderPrecedenceImportCollection<VSLangProj.VSProject> Project { get; }
 
+        [ImportMany]
+        internal OrderPrecedenceImportCollection<IDesignTimeInputsBuildManagerBridge> DesignTimeInputsBuildManagerBridge { get; }
```
davkean: The contract this inteface is marked with is marked as ExactlyOne but you import it as a collection?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -78,6 +91,14 @@ public object DesignTimeOutputMonikers
         /// </summary>
         public string BuildDesignTimeOutput(string bstrOutputMoniker)
         {
+            _threadingService.VerifyOnUIThread();
```
davkean: I don't believe this and above is legacy behavior, and its very different to most of the VS APIs that CPS implements. Legacy will never throw off UI-thread, and will always marshal back to the UI thread. CPS almost always lets you call these APIs on a background and will switch to UI thread if it needs to.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridgeTests.cs
```diff
@@ -0,0 +1,142 @@
...
+            _bridge = new TestDesignTimeInputsBuildManagerBridge(threadingService, changeTracker, compilerMock.Object, _buildManager);
+            _bridge.SkipInitialization = true;
+        }
+
+        public void Dispose()
+        {
+            ((IDisposable)_bridge).Dispose();
+        }
+
+        internal class TestDesignTimeInputsBuildManagerBridge : DesignTimeInputsBuildManagerBridge
```
davkean: Does this class only exist to provide a helper method for ApplyAsync?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
davidwengier: Just avoiding mocking heaps of CPS stuff that's used by the base class. I can if you think it's important.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
davidwengier: There is an analyser that triggers if you use VerifyOnUIThread in an async method

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridgeTests.cs
davidwengier: Yes, it only exists because ApplyAsync is protected

davidwengier:
Appreciate the review, will look at the rest of the feedback tomorrow.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: This is probably just my lack of knowledge of MEF. I can't import the class directly because it would be a circular reference, and the capabilities of IDesignTimeInputsBuildManagerBridge are more restricted than VsBuildManager, so I was under the impression I need to use OrderPrecedenceImportCollection to do the capabilities check. Is there an alternative approach?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: I guess I misunderstood the "always need to be on the UI thread to call VS services" to mean we could rely on callers to us doing the same, and that since these methods are part of VSLangProj, that applied. I can change this to a UI thread switch inside ExecuteSynchronously.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
davidwengier: These are my comments, explaining what each does in my words, mostly written when I was first coming to terms with how this all works.

I can remove them if you think its a problem, but for new or external developers I'd like to think there is value here. "LinkExternalInput" means something once you know what you're doing, but wasn't obvious to me at least.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridgeTests.cs
davidwengier: Good idea. This used to do more, but the name is wrong now.

PR #5327, opened 2019-08-20
Remove "pack" and "restore" concepts from CoreBuild.proj

davkean:
This removes the ability inside CoreBuild.proj to turn off pack/restore. Build.proj (which calls CoreBuild) always passed true for these. 

- Restore always occurs, there's no way to turn it off. If folks decide they want this ability, we can always add it back.
- Pack always occurs when you `/build` or `/rebuild`.

This lets https://github.com/dotnet/project-system/pull/4768 do a restore only build via:

```
build /no-build /no-test /no-deploy
```

### DIFF
--- a/build/proj/CoreBuild.proj
+++ b/build/proj/CoreBuild.proj
@@ -13,7 +13,6 @@
     Deploy                          "true" to deploy assets (e.g. VSIXes)
     Test                            "true" to run tests
     IntegrationTest                 "true" to run integration tests
-    Pack                            "true" to build NuGet packages and VS insertion manifests
     Sign                            "true" to sign built binaries
     SignType                        "real" to send binaries to signing service, "test" to only validate signing configuration.
   -->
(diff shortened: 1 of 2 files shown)


build/proj/CoreBuild.proj
```diff
@@ -62,11 +59,10 @@
 
   <Target Name="Build">
     <ItemGroup>
-      <SolutionBuildTarget Include="Rebuild" Condition="'$(Rebuild)' == 'true'" />
-      <SolutionBuildTarget Include="Build" Condition="'$(Build)' == 'true' and '$(Rebuild)' != 'true'" />
+      <SolutionBuildTarget Include="Rebuild;Pack" Condition="'$(Rebuild)' == 'true'" />
+      <SolutionBuildTarget Include="Build;Pack" Condition="'$(Build)' == 'true' and '$(Rebuild)' != 'true'" />
```
davidwengier: Why not just change the condition on the Pack SolutionBuildTarget to refer to `BuildOrRebuild`? Is there some other reason why the ordering change is advantageous?

davkean: The property didn't exist when I wrote that. I've updated it to use it.

PR #5330, opened 2019-08-20
Minor changes in CodeCov

davkean:
These are some minor changes that don't actually make it run yet but want to get them in to make review easier as I update this.


build/import/Coverage.targets
```diff
@@ -0,0 +1,46 @@
+﻿<Project>
+
+  <Target Name="ExtractWindowsPdb" BeforeTargets="Test">
```
davkean: This is a straight port and have made no changes.

davkean:
Hit AV again.

PR #5336, opened 2019-08-21
Move to new repo layout

### LINKED ISSUE #5276: Change our repository layout
Our current repository layout combines a lot of things into the `src` directory and relies on a lot of logic in `Directory.Build.props` files that don't have a single purpose.

Our proposed new layout is:

```
\
\artifacts
\build
\docs
\resources
\tests
\src
\setup
\deploy  -- maybe
```

This issue tracks that work.

- [ ] Move logic out of Directory.Build.props to specific purpose props files under build
- [ ] Move logic out of Directory.Build.targets to specific purpose targets files under build
- [ ] Move test projects to `tests`
- [ ] Move setup projects to `setup`
- [ ] Move deploy projects to `deploy` (or `tests`?)

davidwengier:
This is probably hell to PR, but hopefully I've done the commits in a sensible way. Ignore the first commit, which is just file moves and no content changes, and hopefully it will be okay.

This moves stuff out of our src directory and gives us three top level folders for projects:

/
/setup
/src
/tests

I got rid of deploy and just put them in tests.

Feedback on all of this appreciated. I'm still validating this with some clean builds, and will kick off a signed build from this PR to validate that, but this at least builds in command line and in Visual Studio happily. There could be more things moved around to other files etc. but its certainly easier to do follow up PRs, in case other people move other stuff around.

Fixes #5276


tests/Directory.Build.props
```diff
@@ -0,0 +1,31 @@
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+  <PropertyGroup>
+    <IsUnitTestProject>false</IsUnitTestProject>
```
davkean: Pretty sure most of this file is needed in the root. We use IsUnitTestProject to determine packages,, and below the rulset changes affect source and test.

davkean: Oh, I've seen how you've done this - let me read a bit more of this first to have an opinion.

davidwengier: Yeah, I broke it up to make reviewing easy, but also means commenting on individual commits is probably harder. Not sure what is better ¯\\\_(ツ)_/¯

src/Directory.Build.props
```diff
@@ -2,79 +2,18 @@
...
-    <TestArchitectures>x86</TestArchitectures>
-
-    <!-- Use IBC optimization data if available -->
-    <IbcOptimizationDataDir Condition="$(EnableIbc) == 'true'">$(NuGetPackageRoot)\RoslynDependencies.ProjectSystem.OptimizationData\$(RoslynDependenciesProjectSystemOptimizationDataVersion)\content\OptimizationData\</IbcOptimizationDataDir>
-
-    <UseCommonOutputDirectory Condition="'$(UseCommonOutputDirectory)' == ''">true</UseCommonOutputDirectory>
+    <CodeAnalysisRuleSet>$(RepoRoot)build\ruleset\Default.ruleset</CodeAnalysisRuleSet>
 
     <!-- Prevent vsix manifests from being copied to the shared output directory, makes build more deterministic -->
     <CopyVsixManifestToOutput>false</CopyVsixManifestToOutput>
```
davkean: I think this is a setup only affecting property - but you can check by seeing if any extension.manifest get copied.

davidwengier: Thats what I thought too, but I got build failures when I removed it from here. I haven't looked into it deeply, but it seemed like if this wasn't here, it didn't embed the manifest.

davkean: Did setup projects get the same property?

davkean:
You'll want to kick off a signed build against this.

Can you paste a screenshot of what Solution Explorer looks like?

setup/Directory.Build.props
```diff
@@ -0,0 +1,22 @@
...
+<Project>
+  <PropertyGroup>
+    <IsUnitTestProject>false</IsUnitTestProject>
+    <IsIntegrationTestProject>false</IsIntegrationTestProject>
+    <IsTestProject>false</IsTestProject>
+    <OutDirName Condition="'$(OutDirName)' == ''">Dlls</OutDirName>
+  </PropertyGroup>
+
+  <Import Project="..\Directory.Build.props"/>
+  <Import Project="..\build\import\VisualStudio.XamlRules.props" />
```
davkean: Instead of duplicating this import, can we just move this property in the file into RepoLayout.props and delete this file.

src/Directory.Build.props
```diff
@@ -2,79 +2,18 @@
 <Project>
   <PropertyGroup>
     <IsUnitTestProject>false</IsUnitTestProject>
```
davkean: Maybe just default this in the parent? and remove all the sets except actual test projects?

src/Directory.Build.props
```diff
@@ -2,79 +2,18 @@
...
-    <CodeAnalysisRuleSet Condition="'$(IsTestProject)' == 'true'">$(RepoRoot)build\ruleset\Test.ruleset</CodeAnalysisRuleSet>
-    <CodeAnalysisRuleSet Condition="'$(IsTestProject)' != 'true'">$(RepoRoot)build\ruleset\Default.ruleset</CodeAnalysisRuleSet>
-
-    <TestArchitectures>x86</TestArchitectures>
-
-    <!-- Use IBC optimization data if available -->
-    <IbcOptimizationDataDir Condition="$(EnableIbc) == 'true'">$(NuGetPackageRoot)\RoslynDependencies.ProjectSystem.OptimizationData\$(RoslynDependenciesProjectSystemOptimizationDataVersion)\content\OptimizationData\</IbcOptimizationDataDir>
-
-    <UseCommonOutputDirectory Condition="'$(UseCommonOutputDirectory)' == ''">true</UseCommonOutputDirectory>
+    <CodeAnalysisRuleSet>$(RepoRoot)build\ruleset\Default.ruleset</CodeAnalysisRuleSet>
```
davkean: Maybe just default and have tests the one that overrides it?

davidwengier:
Solution Explorer:

image

If anyone has a good term to use to put the build and Documentation projects into, I'm all ears.

davkean:
I think the intention is that they will stay as they are, as they now represent "top-level" folders.

davidwengier:
Have kicked off a signed build https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2965557&view=results

src/Directory.Build.props
davidwengier: Just to follow up on this, yes this is in setup projects too. Once we work out (in a later task) why removing this fails the build, and merge in Tom's editorconfig branch, this whole `src\Directory.Build.props` file can be deleted.

PR #5339, opened 2019-08-21
Replace .ruleset files with .editorconfig

tmeschter:
Replace the .ruleset files by updating our .editorconfig with equivalent settings.


.editorconfig
```diff
@@ -210,3 +210,214 @@ csharp_prefer_braces = false:none
...
+dotnet_diagnostic.VSTHRD001.severity = none     # Await JoinableTaskFactory.SwitchToMainThreadAsync() to switch to the UI thread instead of APIs that can deadlock or require specifying a priority.
+dotnet_diagnostic.VSTHRD002.severity = none     # Synchronously waiting on tasks or awaiters may cause deadlocks. Use JoinableTaskFactory.Run instead.
+dotnet_diagnostic.VSTHRD010.severity = none     # Visual Studio service should be used on main thread explicitly.
+dotnet_diagnostic.VSTHRD103.severity = none     # Synchronously blocks. Await ThrowsAsync instead.
+dotnet_diagnostic.VSTHRD110.severity = none     # Observe result of async calls
+dotnet_diagnostic.VSTHRD200.severity = none     # Naming stylesNaming styles:  Before: Task Open()  After: Task OpenAsync()
+
+# xunit.analyzers
+dotnet_diagnostic.xUnit1026.severity = none     # Theory methods must use all parameters
+dotnet_diagnostic.xUnit1004.severity = none     # Test methods should not be skipped
```
sharwell: 💡 This one can be helpful if left as a refactoring.

```suggestion
dotnet_diagnostic.xUnit1004.severity = silent     # Test methods should not be skipped
```

.editorconfig
```diff
@@ -210,3 +210,214 @@ csharp_prefer_braces = false:none
...
+# XmlDocumentationComments.Analyzers
+dotnet_diagnostic.RS0010.severity = warning
+
+# Microsoft.CodeAnalysis.CSharp.Features
+                                                # Name:                                         Before:                                             After:
+dotnet_diagnostic.IDE0001.severity = error      # Simplify names                                System.Version version;                             Version version;
+dotnet_diagnostic.IDE0002.severity = error      # Simplify (member access)                      System.Version.Equals("1", "2");                    Version.Equals("1", "2");
+dotnet_diagnostic.IDE0005.severity = error      # Using directive is unnecessary                using System.Text;
+dotnet_diagnostic.IDE0030.severity = error      # Use coalesce expression (nullable)            int? y = x.HasValue ? x.Value : 0                   int? y = x ?? 0;
+dotnet_diagnostic.IDE0030WithoutSuggestion.severity = error
```
sharwell: ❔ Can you explain the need for this one?

.editorconfig
```diff
@@ -210,3 +210,214 @@ csharp_prefer_braces = false:none
...
+dotnet_diagnostic.IDE1006.severity = error      # Naming styles                                 Task Open()                                         Task OpenAsync()
+dotnet_diagnostic.IDE1006WithoutSuggestion.severity = suggestion
```
sharwell: ❔ Can you explain why this one is needed, and also why the severity differs from IDE1006?

.editorconfig
tmeschter: Nope. This was a largely mechanical translation from the .ruleset file. I'm not trying to reason about _why_ we have the settings we do. 😄

.editorconfig
tmeschter: As with the other one, no idea.

tmeschter:
@davidwengier This is likely to conflict with #5336.

davidwengier:
Feel free to merge this @tmeschter 	I'll deal with the conflicts. Should be pretty minor.

tmeschter:
@davidwengier Is there a way to run a signed build of this change before merging it in? Seeing as that where things went south the last time I worked on this...

davidwengier:
@tmeschter Go to this build: https://devdiv.visualstudio.com/DevDiv/_build?definitionId=9675
Click "Queue" (or "Run Pipeline" if you've got the new UI ¯\\\_(ツ)_/¯)
Enter "refs/pull/5339/head" as the branch

.editorconfig
davkean: It looks [unintentional](https://github.com/dotnet/project-system/commit/a24890a0ced2a64938a71cf0555ba8b57b0b9a58) that this one is an error, but `IDE1006WithoutSuggestion` isn't, Still looking to what these `WithoutSuggestion` items mean.

davkean: Here's the description of what these mean: https://github.com/dotnet/roslyn/blob/1f0ce4646b5d4471ea6d1bdac79495b22653e462/src/CodeStyle/Core/Analyzers/AbstractCodeStyleDiagnosticAnalyzer.cs#L32, I'm guessing so that it underlines/warns all instances of the violation, not just the first one on a line?

.editorconfig
davkean: What does silent mean?

.editorconfig
davkean: I think we can follow up with another PR if we want to remove them.

davidwengier:
@tmeschter I beat you to merging my PR, so I've merged in master and fixed the conflicts in your branch, and updated a couple of files that weren't there when you made this.

Were there no differences between `Default.ruleset` and `Test.ruleset`, or did you just skip the differences for ease? Because it should now be possible to put a .editorconfig file in the tests folder and get back the differences, if they're worth it.

.editorconfig
jmarolf: It will show three dots under the diagnostic location indicating that there is a code fix at this location but will not show anything in the error list

davkean:
.editorconfig already has exclusions for tests, looks like rules are being enforced for tests that were not previously hence the build breaks.

tmeschter:
@davidwengier Test.ruleset pulled in Default.ruleset and then made further, test-project-specific modifications. The .editorconfig does something similar; on line [215](https://github.com/dotnet/project-system/pull/5339/files#diff-1e70daafb475c0ce3fef7d2728279182R215) I specify that the following rules apply to all .cs and .vb files, and then on line [387](https://github.com/dotnet/project-system/pull/5339/files#diff-1e70daafb475c0ce3fef7d2728279182R387) I specify that the following rules only apply to .cs and .vb files under the *Tests directories. I expect that needs to be tweaked to reflect the new repo structure.

That being said, with all the test projects in the same top-level folder it may now make sense for me to split the test-specific rules into their own .editorconfig file in the test directory.

tmeschter:
I've fixed up the .editorconfig file to reflect the new location of the test projects, and I've queued up a signed build to validate the changes. If all that passes I'm going to merge this in.

PR #5340, opened 2019-08-22
Force a build when packing

davidwengier:
When the build output is up to date, the Pack command currently does nothing. This ensures it always does a build, and therefore a pack.

Fixes [AB#935258](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/935258)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IVsSolutionBuildManager2Factory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IVsSolutionBuildManager2Factory.cs
@@ -39,8 +39,7 @@ namespace Microsoft.VisualStudio.Shell.Interop
                     return VSConstants.S_OK;
                 }
 
-                uint dwFlags = (uint)(VSSOLNBUILDUPDATEFLAGS.SBF_SUPPRESS_SAVEBEFOREBUILD_QUERY | VSSOLNBUILDUPDATEFLAGS.SBF_OPERATION_BUILD);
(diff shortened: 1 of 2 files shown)


davkean:
This means [forces a rebuild](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fenv%2Fmsenv%2Fcore%2Fslnupd2.cpp&version=GBmaster&line=1315&lineStyle=plain&lineEnd=1316&lineStartColumn=1&lineEndColumn=1), which doesn't seem like the intention. 

Try this:

```
> dotnet new
> dotnet build
> dotnet pack <!-- Does not rebuild
```

davidwengier:
I wasn't sure about this, but you've been "in a call" all day so I guessed :)

The current implementation does a pack by just doing a build and setting GeneratePackageOnBuild to true, so forcing a build was the path of least resistance. The other option seemed to be to change the code to match legacy and provide a PackageAction property to tell CPS what build action to execute for it. Would that be the preferred approach?

davkean:
By matching legacy here, what do you mean? What UI gesture in legacy?

davidwengier:
Sorry, legacy might be the incorrect term. In the bug on Azure DevOps Davi linked to [this code](https://dev.azure.com/devdiv/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2FPackageAndDeploy%2FPackagingService%2FCommands%2FPackaging.cs&version=GBmaster&line=616&lineEnd=616&lineStartColumn=13&lineEndColumn=78&lineStyle=plain) as being the correct approach. It calls a different method on the solution build manager than our code did.

PR #5361, opened 2019-08-23
Updating README to reflect master is now 16.4

dmonroym:


dmonroym:
The checks for the unit-tests keep showing: Expected — Waiting for status to be reported. Looking into this

dmonroym:
Ah of course. The pipelines aren't running because we exclude README changes

PR #5362, opened 2019-08-23
Split .editorconfig

tmeschter:
Split out the test-project-specific settings into their own
.editorconfig file.

### DIFF
--- a/tests/.editorconfig
+++ b/tests/.editorconfig
@@ -1,0 +1,45 @@
+# EditorConfig is awesome:http://EditorConfig.org
+
+[*.cs]
+# We allow usage of "var" inside tests as it reduces churn as we remove/rename types
+csharp_style_var_for_built_in_types = true:none
+csharp_style_var_elsewhere = true:none
+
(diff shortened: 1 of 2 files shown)


tests/.editorconfig
```diff
@@ -0,0 +1,45 @@
...
+
+[*.cs]
+# We allow usage of "var" inside tests as it reduces churn as we remove/rename types
+csharp_style_var_for_built_in_types = true:none
+csharp_style_var_elsewhere = true:none
+
+# Dotnet code analysis settings:
+[*.{cs,vb}]
+# Microsoft.CodeQuality.Analyzers
+# For tests, the ConfigureAwait(true) is good enough. Either they are already running on a thread pool
```
davkean: This comment seems out of date and applicable to none of these CAXXX rules - was it in a different location before the ruleset -> editorconfig?

davkean: We don't enforce ConfigureAwait anywhere - so the comment is kindof meaningless now.

PR #5374, opened 2019-08-26
Take ownership of http://index indexing

davkean:
Previously this was located in another repository and kept being out of sync with our build, moving it to the official build.


davkean:
Official build completed successfully.

drewnoakes:
Does this need to be removed from wherever was doing this before?

davkean:
@KirillOsenkov Will handle that - he'll delete the build.

PR #5378, opened 2019-08-27
Add IPhysicalProjectTreeStorage.AddFileAsync

davkean:
This is breakout from a larger PR where I'm adding licenses special file provider support, this adds a new method called `IPhysicalProjectTreeStorage.AddFileAsync` which creates a zero'd file on disk and publishes it to Solution Explorer before it has returned.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IPhysicalProjectTreeStorage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IPhysicalProjectTreeStorage.cs
@@ -60,7 +60,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
         /// <exception cref="NotSupportedException">
         ///     <paramref name="path"/> contains a colon character (:) that is not part of a drive label ("C:\").
         /// </exception>
-        Task<IProjectItem?> CreateEmpyFileAsync(string path);
+        Task<IProjectItem?> CreateEmptyFileAsync(string path);
 
         /// <summary>
         ///     Creates a folder on disk, adding it add to the physical project tree.
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IPhysicalProjectTreeStorage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IPhysicalProjectTreeStorage.cs
@@ -25,9 +25,6 @@ namespace Microsoft.VisualStudio.ProjectSystem
         /// <param name="path">
         ///     The path of the folder to create, can be relative to the project directory.
         /// </param>
-        /// <returns>
-        ///     The created <see cref="IProjectItem"/>.
-        /// </returns>
         /// <remarks>
         ///     This method will automatically publish the resulting tree to <see cref="IProjectTreeService.CurrentTree"/>.
         /// </remarks>
@@ -60,7 +57,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
         /// <exception cref="NotSupportedException">
         ///     <paramref name="path"/> contains a colon character (:) that is not part of a drive label ("C:\").
         /// </exception>
-        Task<IProjectItem?> CreateEmptyFileAsync(string path);
+        Task CreateEmptyFileAsync(string path);
 
         /// <summary>
         ///     Creates a folder on disk, adding it add to the physical project tree.
@@ -68,9 +65,6 @@ namespace Microsoft.VisualStudio.ProjectSystem
         /// <param name="path">
         ///     The path of the folder to create, can be relative to the project directory.
         /// </param>
-        /// <returns>
-        ///     The created <see cref="IProjectTree"/>.
-        /// </returns>
         /// <remarks>
         ///     This method will automatically publish the resulting tree to <see cref="IProjectTreeService.CurrentTree"/>.
         /// </remarks>
@@ -103,6 +97,6 @@ namespace Microsoft.VisualStudio.ProjectSystem
         /// <exception cref="NotSupportedException">
         ///     <paramref name="path"/> contains a colon character (:) that is not part of a drive label ("C:\").
         /// </exception>
-        Task<IProjectTree?> CreateFolderAsync(string path);
+        Task CreateFolderAsync(string path);
     }
 }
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/PhysicalProjectTreeStorage.cs
```diff
@@ -22,34 +23,52 @@ internal class PhysicalProjectTreeStorage : IPhysicalProjectTreeStorage
...
+        public Task<IProjectTree?> CreateFolderAsync(string path)
         {
-            _fileSystem.Value.CreateDirectory(fullPath);
+            Requires.NotNullOrEmpty(path, nameof(path));
+
+            return AddToProjectAsync(path, waitForFileSystemUpdates: false, async fullPath =>
+            {
+                _fileSystem.Value.CreateDirectory(fullPath);
+
+                await _folderManager.Value.IncludeFolderInProjectAsync(fullPath, recursive: false);
```
davidwengier: Old code, but should be able to remove this async state machine too

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/PhysicalProjectTreeStorage.cs
```diff
@@ -22,34 +23,52 @@ internal class PhysicalProjectTreeStorage : IPhysicalProjectTreeStorage
...
             Requires.NotNullOrEmpty(path, nameof(path));
 
-            if (_treeService.Value.CurrentTree == null)
-                throw new InvalidOperationException("Physical project tree has not yet been published.");
+            return await AddToProjectAsync(path, waitForFileSystemUpdates: true, async fullPath =>
+            {
+                using (_fileSystem.Value.Create(fullPath)) { }
 
-            string fullPath = _unconfiguredProject.MakeRooted(path);
+                await _sourceItemProvider.Value.AddAsync(fullPath);
```
davidwengier: Can remote this async state machine, and maybe the outer one, depending on that cast.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IPhysicalProjectTreeStorage.cs
```diff
@@ -14,11 +14,54 @@ namespace Microsoft.VisualStudio.ProjectSystem
...
     /// <remarks>
     ///     This interface provides a simple facade over the <see cref="IFileSystem"/>, <see cref="IFolderManager"/>, 
-    ///     <see cref="IProjectTreeService"/> and <see cref="IProjectItemProvider"/> interfaces.
+    ///     <see cref="IProjectTreeService"/>, <see cref="IProjectItemProvider"/> and <see cref="IProjectItemProvider"/> interfaces.
     /// </remarks>
     [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
     internal interface IPhysicalProjectTreeStorage
     {
+        /// <summary>
+        ///     Creates a zeroed file on disk, adding it add to the physical project tree.
```
drewnoakes: Zeroed or empty? To me a zeroed file has non-zero length and contains all zeroes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/PhysicalProjectTreeStorage.cs
```diff
@@ -13,45 +13,58 @@ internal class PhysicalProjectTreeStorage : IPhysicalProjectTreeStorage
...
+
+            }) as IProjectItem;
+        }
+
+        public Task<IProjectTree?> CreateFolderAsync(string path)
+        {
+            Requires.NotNullOrEmpty(path, nameof(path));
 
-            return AddToProjectAsync(fullPath);
+            return AddToProjectAsync(path, waitForFileSystemUpdates: false, async fullPath =>
```
drewnoakes: `CreateEmpyFileAsync` awaits the `AddToProjectAsync` task, whereas this returns it directly. Perhaps they should be consistent. Awaiting in both will create clearer stack traces.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractFindByNameSpecialFileProvider.cs
```diff
@@ -29,8 +28,8 @@ internal abstract class AbstractFindByNameSpecialFileProvider : ISpecialFileProv
 
         protected AbstractFindByNameSpecialFileProvider(
             IPhysicalProjectTree projectTree,
-            [Import(ExportContractNames.ProjectItemProviders.SourceFiles)] IProjectItemProvider sourceItemsProvider,
-            [Import(AllowDefault = true)] Lazy<ICreateFileFromTemplateService>? templateFileCreationService,
+            IProjectItemProvider sourceItemsProvider,
+            Lazy<ICreateFileFromTemplateService>? templateFileCreationService,
```
drewnoakes: This was only nullable due to `AllowDefault = true`:

```suggestion
            Lazy<ICreateFileFromTemplateService> templateFileCreationService,
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/PhysicalProjectTreeStorage.cs
davkean: Not convinced by this - one needs the result, one doesn't, and would just add unneeded state machinary.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/IPhysicalProjectTreeStorage.cs
davkean: Yep, good call.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/PhysicalProjectTreeStorage.cs
drewnoakes: Ah, I missed the `as` cast at the end.

PR #5387, opened 2019-08-28
Introduce new AbstractSpecialFileProvider base

davkean:
This will become the root base class for special file providers, and moved AppDesignerFolderSpecialFileProvider over to it. Other providers forthcoming.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
+{
+    /// <summary>
+    ///     Provides the base class for all <see cref="ISpecialFileProvider"/> instances.
+    /// </summary>
+    internal abstract class AbstractSpecialFileProvider : ISpecialFileProvider
+    {
+        private readonly IProjectTreeService _treeService;
+
+        protected AbstractSpecialFileProvider(IProjectTreeService treeService)
+        {
+            _treeService = treeService;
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppDesignerFolderSpecialFileProvider.cs
```diff
@@ -1,82 +1,55 @@
...
 
 namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
 {
     /// <summary>
-    ///     Provides a <see cref="ISpecialFileProvider"/> that handles the AppDesigner, 
+    ///     Provides a <see cref="ISpecialFileProvider"/> that handles the AppDesigner folder;
     ///     called "Properties" in C# and "My Project" in Visual Basic.
     /// </summary>
     [ExportSpecialFileProvider(SpecialFiles.AppDesigner)]
     [Export(typeof(AppDesignerFolderSpecialFileProvider))]
```
davidwengier: Nit: Looks like this export could be removed

davkean: Yep, you are right.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+    internal abstract class AbstractSpecialFileProvider : ISpecialFileProvider
+    {
+        private readonly IProjectTreeService _treeService;
+
+        protected AbstractSpecialFileProvider(IProjectTreeService treeService)
+        {
+            _treeService = treeService;
+        }
+
+        public virtual async Task<string?> GetFileAsync(SpecialFiles fileId, SpecialFileFlags flags, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
davkean: MEF Metadata on the type tells CPS what special file the handler handles - it could be many, ours are 1 to 1, though. We could add a field to check this, but it will just restate what the metadata states and would solely be for an assert or argument throw that would never occur.

PR #5388, opened 2019-08-28
Force generate package on build to false when packing

### LINKED ISSUE #1339: Pack from VS UI should call msbuild /t:pack instead of msbuild /p:GeneratePackageOnBuild=true
There are a couple of issues with the current approach:

1) Not consistent with command line : calling ```dotnet pack``` builds the project by default, but only if ```GeneratePackageOnBuild``` is not set to ```true```. However, invoking Pack from the UI always builds the project and there is no way to opt-out.

2) Confusing for user : User asked to pack the project, not build it, and therefore the entry point should be the Pack target, and not build.

CC: @rrelyea @srivatsn

davidwengier:
Previously for the Pack command we would set GeneratePackageOnBuild to true, and then do a build. This causes a problem as we would see the project as up to date even if the package output didn't exist.

This fixes the problem by setting GeneratePackageOnBuild to false, and then calling the Pack target, which will create build outputs and/or package outputs as necessary. Just calling Pack isn't enough, as it has a condition that it doesn't do a build if GeneratePackageOnBuild is true, which according to the Nuget team is by design.

Fixes #1339
Fixes [AB#935258](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/935258)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
@@ -112,13 +112,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
 
         public int UpdateSolution_Done(int fSucceeded, int fModified, int fCancelCommand)
         {
-            _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(false);
+            _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(null);
             return HResult.OK;
         }
 
         public int UpdateSolution_Cancel()
         {
-            _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(false);
+            _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(null);
             return HResult.OK;
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
@@ -93,12 +93,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
                 var projectVsHierarchy = (IVsHierarchy)Project.Services.HostObject;
                 ErrorHandler.ThrowOnFailure(_buildManager!.SaveDocumentsBeforeBuild(projectVsHierarchy, (uint)VSConstants.VSITEMID.Root, 0 /*docCookie*/));
 
-                // Enable generating package on build ("GeneratePackageOnBuild") for all projects being built.
-                _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(true);
+                // Turn off "GeneratePackageOnBuild" because otherwise the Pack target will not do a build, even if there is no built output
+                _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(false);
 
-                // Kick off the build.
                 uint dwFlags = (uint)(VSSOLNBUILDUPDATEFLAGS.SBF_SUPPRESS_SAVEBEFOREBUILD_QUERY | VSSOLNBUILDUPDATEFLAGS.SBF_OPERATION_BUILD);
-                ErrorHandler.ThrowOnFailure(_buildManager.StartSimpleUpdateProjectConfiguration(projectVsHierarchy, null, null, dwFlags, 0, 0));
+
+                // We tell the Solution Build Manager to Package, which will call the Pack target, which will build if necessary.
+                uint[] buildFlags = new[] { VSConstants.VS_BUILDABLEPROJECTCFGOPTS_PACKAGE };
+                ErrorHandler.ThrowOnFailure(_buildManager.StartUpdateSpecificProjectConfigurations(1, new[] { projectVsHierarchy }, null, null, buildFlags, null, dwFlags, 0));
             }
 
             return true;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
@@ -100,7 +100,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands
 
                 // We tell the Solution Build Manager to Package, which will call the Pack target, which will build if necessary.
                 uint[] buildFlags = new[] { VSConstants.VS_BUILDABLEPROJECTCFGOPTS_PACKAGE };
-                ErrorHandler.ThrowOnFailure(_buildManager.StartUpdateSpecificProjectConfigurations(1, new[] { projectVsHierarchy }, null, null, buildFlags, null, dwFlags, 0));
+                ErrorHandler.ThrowOnFailure(_buildManager.StartUpdateSpecificProjectConfigurations(cProjs: 1,
+                                                                                                   rgpHier: new[] { projectVsHierarchy },
+                                                                                                   rgpcfg: null,
+                                                                                                   rgdwCleanFlags: null,
+                                                                                                   rgdwBuildFlags: buildFlags,
+                                                                                                   rgdwDeployFlags: null,
+                                                                                                   dwFlags: dwFlags,
+                                                                                                   fSuppressUI: 0));
             }
 
             return true;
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
```diff
@@ -93,12 +93,14 @@ protected override async Task<bool> TryHandleCommandAsync(IProjectTree node, boo
...
-                _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(true);
+                // Turn off "GeneratePackageOnBuild" because otherwise the Pack target will not do a build, even if there is no built output
+                _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(false);
 
-                // Kick off the build.
                 uint dwFlags = (uint)(VSSOLNBUILDUPDATEFLAGS.SBF_SUPPRESS_SAVEBEFOREBUILD_QUERY | VSSOLNBUILDUPDATEFLAGS.SBF_OPERATION_BUILD);
-                ErrorHandler.ThrowOnFailure(_buildManager.StartSimpleUpdateProjectConfiguration(projectVsHierarchy, null, null, dwFlags, 0, 0));
+
+                // We tell the Solution Build Manager to Package, which will call the Pack target, which will build if necessary.
+                uint[] buildFlags = new[] { VSConstants.VS_BUILDABLEPROJECTCFGOPTS_PACKAGE };
```
davkean: @drewnoakes Looks like we should probably consider implementing up-to-date for this now?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
```diff
@@ -93,12 +93,14 @@ protected override async Task<bool> TryHandleCommandAsync(IProjectTree node, boo
...
+                ErrorHandler.ThrowOnFailure(_buildManager.StartUpdateSpecificProjectConfigurations(1, new[] { projectVsHierarchy }, null, null, buildFlags, null, dwFlags, 0));
             }
```
davkean: Is there an entry-point that "package" (this is solution "publish", right?) can be executed elsewhere via a command or menu item? If so, this code won't be running and we won't override the global property.

davidwengier: This is that command and menu item. There isn't anything in VS itself that I could find that uses the package option

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
davidwengier: It's worth noting that all of the regular up-to-date checks still apply here, and repeated runs of this command don't modify any files on disk, but yes Fast Up To Date does explicitly opt out of this due to the different build action.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
```diff
@@ -93,12 +93,14 @@ protected override async Task<bool> TryHandleCommandAsync(IProjectTree node, boo
...
+                // Turn off "GeneratePackageOnBuild" because otherwise the Pack target will not do a build, even if there is no built output
+                _generatePackageOnBuildPropertyProvider.OverrideGeneratePackageOnBuild(false);
 
-                // Kick off the build.
                 uint dwFlags = (uint)(VSSOLNBUILDUPDATEFLAGS.SBF_SUPPRESS_SAVEBEFOREBUILD_QUERY | VSSOLNBUILDUPDATEFLAGS.SBF_OPERATION_BUILD);
-                ErrorHandler.ThrowOnFailure(_buildManager.StartSimpleUpdateProjectConfiguration(projectVsHierarchy, null, null, dwFlags, 0, 0));
+
+                // We tell the Solution Build Manager to Package, which will call the Pack target, which will build if necessary.
+                uint[] buildFlags = new[] { VSConstants.VS_BUILDABLEPROJECTCFGOPTS_PACKAGE };
+                ErrorHandler.ThrowOnFailure(_buildManager.StartUpdateSpecificProjectConfigurations(1, new[] { projectVsHierarchy }, null, null, buildFlags, null, dwFlags, 0));
```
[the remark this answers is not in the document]
davidwengier: I feel like for the `1` it makes sense, I feel ambivalent about the `null`s and `0` since I think they inherently mean "who cares", but since you asked nicely I feel this is a simple change to make :)

davidwengier:
Could this potentially be fixed at a lower layer (e.g. in the relevant MSBuild targets) by noting the package output as an output if (and only if) GeneratePackageOnBuild is true? It seems like then MSBuild would do the right thing.

Yes, possibly. I feel like this change is more honest though (the Pack UI command now runs the Pack msbuild target) and it doesn't preclude Fast Up To Date from being changed to help in future (by dealing with the Package build action). In both cases, knowing about the nupkg as an output is necessary, so would probably require the same changes to targets anyway. Also, and I don't know the details of Fast Up To Date in depth, but I don't think that it would be an easy change, as if you do a Build, then Pack, then Build, fast up to date would essentially see three builds, with different outputs in each. Having it see two Builds and a Package seems more straight forward to me.

Changing the targets so that Pack always does a build regardless of GeneratePackageOnBuild is also technically possible, though the Nuget team have said that the current behaviour is by design. If that ever changes, this solution will still work anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
davkean: Can you create an AppXPackage producing project and reference this project and try publishing that? I'm pretty sure that will skip this global property registering.

davkean:
Can you please investigate if we also need to do the right thing around "Clean" & package? Legacy handles this: https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2Fvsproject%2FCVsProjBuildableProjectCfg.cpp&version=GBmaster&line=362&lineStyle=plain&lineEnd=362&lineStartColumn=31&lineEndColumn=65.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
davkean: The way I found this was searched for all usages of VS_BUILDABLEPROJECTCFGOPTS_PACKAGE.

davkean:
Looks like there's three targets you need to worry about (I don't know if CPS handles the last two but legacy does):

```
PackageAction
CleanPackageAction
RebuildPackageAction
```

These feel like they should be owned/defined by the pack targets.

davkean:
I'm not convinced this that only calling package is the correct behavior here. While Pack might call "Build" on the current project - it does so outside of the normal Solution build hooks which means:

1) Dependencies won't be built.
2) Any solution hooks that run on build (such as NuGet restore) might not be called.

Can you expand on why you don't call build? I know there was some sort of issues but I'm looking through the commits and I see no context as to why you don't? The other usage of this [inside VS does call build](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2FPackageAndDeploy%2FPackagingService%2FCommands%2FPackaging.cs&version=GBmaster&line=616&lineStyle=plain&lineEnd=616&lineStartColumn=43&lineEndColumn=77)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/AbstractGenerateNuGetPackageCommand.cs
davidwengier: Yeah, I searched for the same thing and also found the AppX packager, which is also linked in the AzDo issue. Maybe I'm missing something, but I don't see how it's related? It only calls the Package target on the first project in the selection, not on dependencies (see [this comment](https://dev.azure.com/devdiv/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2FPackageAndDeploy%2FPackagingService%2FCommands%2FPackaging.cs&version=GBmaster&line=621&lineStyle=plain&lineEnd=622&lineStartColumn=1&lineEndColumn=1)). Even if it did do it on dependencies, it wouldn't work today because we don't currently supply a value for PackageAction, so this change only improves things. And finally, if you were creating a package for a project, you wouldn't want individual packages for dependant projects anyway, you'd want the dependencies included in your package.

davidwengier:
Confirmed the above scenario doesn't work, so updated this to build dependency projects.

davidwengier:
Logged https://github.com/NuGet/Home/issues/8522 to make the solution more ideal

davidwengier:
Logged https://devdiv.visualstudio.com/DevDiv/_workitems/edit/974321 for the missing CPS options

davidwengier:
Merging without approval. May git have mercy on my soul.

PR #5391, opened 2019-08-29
Use DisposableBag for subscriptions

drewnoakes:
Part of an initiative to consolidate duplicated code across the dependencies node.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
@@ -5,7 +5,6 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependencySharedProjectsSubscriber.cs
```diff
@@ -56,12 +58,10 @@ public void AddSubscriptions(AggregateCrossTargetProjectContext projectContext)
 
         public void ReleaseSubscriptions()
         {
-            foreach (IDisposable link in _subscriptionLinks)
-            {
-                link.Dispose();
-            }
+            // Prevent double-dispose
+            DisposableBag? value = Interlocked.Exchange(ref _subscriptions, null);
```
davkean: That's exactly what DisposableBag does under the covers to prevent double-disposing.

drewnoakes: The comment is perhaps incomplete. DisposableBag is a one-time thing. We need to recreate the instance for future subscriptions, done in another interlocked operation elsewhere in this PR.

drewnoakes: I removed the interlocked stuff here as it was incomplete and the caller is currently synchronising calls anyway.

This will be revisited as I work towards consolidating `DependencyRulesSubscriber` and `DependencySharedProjectsSubscriber` which both contain similar code.

PR #5392, opened 2019-08-29
OriginalItemSpec cannot be null

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Models/DependencyModel.cs
```diff
@@ -21,7 +21,7 @@ private enum DependencyFlags : byte
 
         protected DependencyModel(
             string path,
-            string? originalItemSpec,
+            string originalItemSpec,
```
davkean: What if the rule doesn't include this property?

drewnoakes: In the current implementation it will never be null. If the `OriginalItemSpec` metadata is not present then `ItemSpec` is used.

PR #5395, opened 2019-08-29
Fix up-to-date false negative on project load

### LINKED ISSUE #5386: [Regression] VS always considers a project out of date if its reopened and nothing has changed
**Visual Studio Version**:

**Summary**:

**Steps to Reproduce**:

1. Create Console App (.NET Core) project
2. Build
3. Build

-- Notice that its up-to-date --

3. Close Solution
4. Reopen solution
5. Build
6. Build again

**Expected Behavior**:

For it to be up-to-date

**Actual Behavior**:
The project is out-of-date, and will never be up-to-date until you _Clean_ or _Rebuild_ or make a change to the project.

```
1>FastUpToDate: Adding UpToDateCheckBuilt outputs: (ConsoleApp64)
1>FastUpToDate:     'C:\Users\davkean\source\repos\ConsoleApp64\ConsoleApp64\bin\Debug\netcoreapp3.0\ConsoleApp64.dll' (ConsoleApp64)
1>FastUpToDate:     'C:\Users\davkean\source\repos\ConsoleApp64\ConsoleApp64\obj\Debug\netcoreapp3.0\ConsoleApp64.pdb' (ConsoleApp64)
1>FastUpToDate:     'C:\Users\davkean\source\repos\ConsoleApp64\ConsoleApp64\obj\Debug\netcoreapp3.0\ConsoleApp64.dll' (ConsoleApp64)
1>FastUpToDate:     'C:\Users\davkean\source\repos\ConsoleApp64\ConsoleApp64\bin\Debug\netcoreapp3.0\ConsoleApp64.pdb' (ConsoleApp64)
1>FastUpToDate: The set of project items was changed more recently (8/28/2019 12:56:04 PM) than the earliest output 'C:\Users\davkean\source\repos\ConsoleApp64\ConsoleApp64\bin\Debug\netcoreapp3.0\ConsoleApp64.dll' (8/28/2019 12:55:49 PM), not up to date. (ConsoleApp64)
1>FastUpToDate: Up to date check completed in 0.5 ms (ConsoleApp64)
1>------ Build started: Project: ConsoleApp64, Configuration: Debug Any CPU ------
1>You are using a preview version of .NET Core. See: https://aka.ms/dotnet-core-preview
1>ConsoleApp64 -> C:\Users\davkean\source\repos\ConsoleApp64\ConsoleApp64\bin\Debug\netcoreapp3.0\ConsoleApp64.dll
========== Build: 1 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========

```

**User Impact**:
Wasted time waiting for redundant builds.

drewnoakes:
Fixes #5386.

A data point used by the fast up-to-date check in VS is the last time the set of inputs changed. This is used to address issues described in #4736 and #3245.

After #5089 that time stamp was updated when the project loaded. This meant that loading a project which is up-to-date (from MSBuild's perspective) would always have `LastItemsChangedAtUtc` after the last output time, but building would not update the outputs and the project would be considered out of date until it was actually rebuilt (either due to clean/rebuild/input change).

This change only updates `LastItemsChangedAtUtc` when we already have a set of items. This means the property maintains its default value (`DateTime.MinValue`) after project load and is only updated when the set of items actually changes.

(This PR targets `dev16.3.x` and once merged I'll cherry-pick it to `master` as well.)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -294,6 +294,29 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                     copyReferenceInputs,
                     lastItemsChangedAtUtc,
                     LastCheckedAtUtc);
+
+                static CopyToOutputDirectoryType GetCopyType(IImmutableDictionary<string, string> itemMetadata)
+                {
+                    if (itemMetadata.TryGetValue(CopyToOutputDirectory, out string value))
+                    {
+                        if (string.Equals(value, Always, StringComparison.OrdinalIgnoreCase))
+                        {
+                            return CopyToOutputDirectoryType.CopyAlways;
+                        }
+
+                        if (string.Equals(value, PreserveNewest, StringComparison.OrdinalIgnoreCase))
+                        {
+                            return CopyToOutputDirectoryType.CopyIfNewer;
+                        }
+                    }
+
+                    return CopyToOutputDirectoryType.CopyNever;
+                }
+
+                static string? GetLink(IImmutableDictionary<string, string> itemMetadata)
+                {
+                    return itemMetadata.TryGetValue(Link, out string link) ? link : null;
+                }
             }
 
             public State WithLastCheckedAtUtc(DateTime lastCheckedAtUtc)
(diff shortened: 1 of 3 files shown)


davkean:
I don't get it - can you explain more?

drewnoakes:
> Presumably when the last update is datetime min, it's not considered up to date, but doing a build then sets it because we receive output items? 

`LastItemsChangedAtUtc` represents the last time the set of items in the project changed. If it's `MinValue` then it won't be greater than any of the outputs and so won't contribute towards suggesting a project is out of date (i.e. the false negative this PR fixes).

> And presumably that happens regardless of whether any files were written by the build?

Before this change we would have required an output timestamp to be greater than the time of project load, as we set `LastItemsChangedAtUtc` to project load time. With this change, that requirement no longer exists and we let the other fast up-to-date checks decide.

drewnoakes:
Put another way, when the project loads we should not assume that the set of items has changed.

PR #5416, opened 2019-09-03
Options layout

drewnoakes:
- Adds some vertical space between options
- Removes left margin from the "Logging Level" text so the left column is flush aligned
- Moves the combo box to the same horizontal offset as is used in other property pages (e.g. _VB Defaults_)
- Makes the group box align to top, as in other property pages (e.g. _Performance_)
- Reduces the height of the combo box to match other property pages

Before:

image

After:

image


davkean:
I wonder why the old name ".NET Core" is showing?

drewnoakes:
> I wonder why the old name ".NET Core" is showing?

The screenshot is from 16.2.3. It's fine in 16.3.0.

PR #5420, opened 2019-09-03
Extract base class for rules subscribers

drewnoakes:
The dependencies node defines interface `IDependencyCrossTargetSubscriber`, implementations of which subscribe to project Dataflow sources and emit events when data changes.

There are currently two implementations of this interface, `DependencyRulesSubscriber` and `DependencySharedProjectsSubscriber`. Both of these shared a considerable amount of code yet differed in various ways.

For example:
- They used different base classes (`OnceInitializedOnceDisposedUnderLockAsync` vs. `OnceInitializedOnceDisposed` respectively).
- The former protected against project unload and capabilities changes while processing updates, while the latter did not.
- The former suppressed version-only updates while the latter did not.

This PR extracts a new common base class `DependencyRulesSubscriberBase<T>` to standardise the handling of Dataflow subscriptions, updates and instance lifecycle events.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/CrossTarget/IDependenciesRuleHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/CrossTarget/IDependenciesRuleHandler.cs
@@ -1,6 +1,5 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Immutable;
 
 using Microsoft.VisualStudio.Composition;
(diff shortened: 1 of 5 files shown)


davidwengier:
A random question, assuming I know nothing about dependencies node etc. but..

> The former protected against project unload and capabilities changes while processing updates, while the latter did not.

This makes sense because shared projects don't reload themselves silently so capabilities can't change out from under them. Given that that might be changing with other work we do, I wonder aloud if it even makes sense to have hard-coded support for these, or if somehow there could be capability based logic (ie, does the project have the Reload capability) in the node providers.

If that even makes any sense...

davkean:
> This makes sense because shared projects don't reload themselves silently so capabilities can't change out from under them. 

Haven't look at the change, but I suspect this doesn't affect Shared Project (only the depedencies node for Shared Projects), but Shared Project capabilities do and can change, CSharp for example is a capability that changes during the lifetime of the project because it comes from targets and hence isn't present before the active configuration has been determined.

drewnoakes:
> Given that that might be changing with other work we do, I wonder aloud if it even makes sense to have hard-coded support for these

Do you mean support for shared projects? Unlike `DependencyRulesSubscriber`, `DependencySharedProjectsSubscriber` is also subscribes to `SharedFoldersSource` for information about the shared projects in the solution. Other dependency types are able to source data from `ProjectSource` and `JointSource`.

PR #5431, opened 2019-09-04
LOC CHECKIN | dotnet/project-system dev16.3.x  | 20190904

v-zbsail:
fix for bug https://devdiv.visualstudio.com/DevDiv/_workitems/edit/976541

Fixes [AB#976541](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/976541)


tmeschter:
@davkean Note that even loc-only changes require QB approval for 16.3 at this point.

davkean:
Ah, mb.

PR #5433, opened 2019-09-04
Remove multi-lifetime support from VisualStudioRefactorNotifyService

davkean:
This class was an implicitly-active service that lived in configured-project scope when it had no initialization or shutdown logic. Make it a plain service.


jmarolf:
Thanks @davkean 	!

PR #5434, opened 2019-09-04
Keep version number in line with Visual Studio

davidwengier:
I'm guessing the intention when changing version numbers to 16.3.0 was to keep them in line with VS, so...


davidwengier:
> Do we have a document somewhere that should capture this?

Yep, I updated the checklist in OneNote too

PR #5435, opened 2019-09-05
Remove async from renamer

davkean:
This includes https://github.com/dotnet/project-system/pull/5433, so only review https://github.com/dotnet/project-system/commit/6475aa759c8688fc71e8550b28542850ab231707.

These code paths were async because they used IVsService<T>, moved to IVsUIService as they are using only UI-bound services.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Refactor/VisualStudioRefactorNotifyService.cs
```diff
@@ -11,50 +12,124 @@
...
-        public Task DeactivateAsync() => UnloadAsync();
+        public void OnBeforeGlobalSymbolRenamed(string projectPath, IEnumerable<string> filePaths, string rqName, string newName)
+        {
+            IVsHierarchy? projectHierarchy = GetProjectHierarchy(projectPath);
+            if (projectHierarchy == null)
+            {
+                return;
+            }
+
+            if (!(projectHierarchy is IVsHierarchyRefactorNotify refactorNotify))
```
jmarolf: couldn't this just be a single if with an `||`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Refactor/VisualStudioRefactorNotifyService.cs
```diff
@@ -11,50 +12,124 @@
...
+
+        public void OnAfterGlobalSymbolRenamed(string projectPath, IEnumerable<string> filePaths, string rqName, string newName)
+        {
+            IVsHierarchy? projectHierarchy = GetProjectHierarchy(projectPath);
+            if (projectHierarchy == null)
+            {
+                return;
+            }
+
+            if (!(projectHierarchy is IVsHierarchyRefactorNotify refactorNotify))
```
jmarolf: same question

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Refactor/VisualStudioRefactorNotifyService.cs
davkean: I didn't change this code.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Refactor/VisualStudioRefactorNotifyService.cs
davkean: I didn';t change this code.

PR #5441, opened 2019-09-05
Annotation new warnings

davkean:
New warnings appeared latest master, annotate types correctly.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -454,7 +454,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
 
                 if (!cancellationToken.IsCancellationRequested)
                 {
-                    dependenciesNode = await viewProvider.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -454,7 +454,7 @@ private async Task OnDependenciesSnapshotChangedAsync(SnapshotChangedEventArgs e
 
                 if (!cancellationToken.IsCancellationRequested)
                 {
-                    dependenciesNode = await viewProvider.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
+                    dependenciesNode = await viewProvider!.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
```
drewnoakes: This seems like a regression in the compiler.

davkean: I was unsure about that, @agocke @jcouv Thoughts?

Null checked here: https://github.com/dotnet/project-system/blob/8f558fe5d4f6891164601017b48d15a7bdb83ea6/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs#L427
Had to `!` here: https://github.com/dotnet/project-system/blob/8f558fe5d4f6891164601017b48d15a7bdb83ea6/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs#L457.

PR #5462, opened 2019-09-10
Avoid NGEN'ing resource dlls

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/980771. This SDK has a fix to avoid NGEN'ing resource dlls for Editors and AppDesigner.


jjmew:
AB#980771

PR #5463, opened 2019-09-10
Check for invalid path characters in license file path

etbyrd:
Probable fix for [972272](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/972272/)

This makes it impossible to set a license file path with invalid path characters

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
@@ -489,18 +489,20 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             If _unconfiguredProject Is Nothing OrElse _configuredProject Is Nothing Then
                 GetProjectsAndProvider()
             End If
-            'The TextBox needs to have the relative path, so the property isn't linked to the TextBox. It must be set manually.
-            AddOrChangeLicenseItem(_previousLicenseFileName, relativeFilePath)
-            LicenseFileNameTextBox.Text = relativeFilePath
-            _previousLicenseFileName = LicenseFileNameTextBox.Text
-            'If the user has changed the directory on their PackageLicenseFile, we should keep it there
-            Dim PackageLicenseFileSet = TryCast(TryGetNonCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile")), String)
-            If PackageLicenseFileSet IsNot Nothing AndAlso PackageLicenseFileSet IsNot "" Then
-                Dim currentPackageLicenseFileDirectory = Path.GetDirectoryName(PackageLicenseFileSet)
-                Dim potentialFullString = Path.Combine(currentPackageLicenseFileDirectory, Path.GetFileName(relativeFilePath))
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
```diff
@@ -489,18 +489,20 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
-            _previousLicenseFileName = LicenseFileNameTextBox.Text
-            'If the user has changed the directory on their PackageLicenseFile, we should keep it there
-            Dim PackageLicenseFileSet = TryCast(TryGetNonCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile")), String)
-            If PackageLicenseFileSet IsNot Nothing AndAlso PackageLicenseFileSet IsNot "" Then
-                Dim currentPackageLicenseFileDirectory = Path.GetDirectoryName(PackageLicenseFileSet)
-                Dim potentialFullString = Path.Combine(currentPackageLicenseFileDirectory, Path.GetFileName(relativeFilePath))
-                SetCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile"), potentialFullString)
-            Else
-                SetCommonPropertyValue(GetPropertyDescriptor("PackageLicenseFile"), Path.GetFileName(relativeFilePath))
+            If relativeFilePath = String.Concat(relativeFilePath.Split(Path.GetInvalidFileNameChars())) Then
```
drewnoakes: Does this work for a path like `Foo\Bar\Baz.cs` or `Foo/Bar/Baz.cs`? The slashes are invalid file name characters, but are valid in paths.

etbyrd: No - that was a mistake from testing. It should have been InvalidPathChars.

drewnoakes: Can you use [`String.IndexOfAny`](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexofany?view=netframework-4.8#System_String_IndexOfAny_System_Char___) instead?

drewnoakes:
This PR fixes the symptom but doesn't really help the user understand why their license file isn't being added to the project. How hard would it be to validate the text box's content and give feedback to the user when they enter something invalid?

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
etbyrd: Sure - Would that be significantly faster?

etbyrd:
I agree - I am going to submit a separate PR with similar behavior to the resource file text box on the Application property page.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.vb
drewnoakes: It would avoid extra string allocations, but mostly it'd be clearer.

drewnoakes:
> I am going to submit a separate PR with similar behavior to the resource file text box on the Application property page. It will prompt the user with an error if the file does not exist or has invalid characters.

PR #5467, opened 2019-09-11
Be resilent when AppDesigner folder is missing or a file

davkean:
This makes the special file provider for AppDesigner folder (and hence AbstractSpecialFileProvider) more resilent against being:

- Excluded from the project
- Included in the project, but missing on disk
- An actual file on disk

This mimics legacy extremely closely - except that CPS has a bug where it silently ignores failures from special file handlers, which will be fixed later.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
@@ -1,7 +1,11 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System.Globalization;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.Interop;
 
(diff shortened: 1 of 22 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractSpecialFileProvider.cs
```diff
@@ -11,37 +15,97 @@ namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
...
+
+        private Task AddFileAsync(string path)
+        {
+            return _isFolder ? _storage.AddFolderAsync(path) : _storage.AddFileAsync(path);
+        }
+
+        private Task VerifyStateAsync(IProjectTree node, string path)
+        {
+            if (_isFolder != node.IsFolder)
+                throw new IOException(string.Format(CultureInfo.CurrentCulture, Resources.SpecialFileProvider_FileOrFolderAlreadyExists, node.Caption), Win32Interop.HResultFromWin32(Win32Interop.ERROR_FILE_EXISTS));
```
davkean: This is supposed to be shown to the user, but the CPS is currently hiding errors from special file handlers: http://index/?query=ProjectNode.GetFile&rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Package%5CNodes%5CProjectNode.cs&line=2651.

Issue #4938, opened 2019-06-20
Third-party Windows Forms controls that use LicenseManager crash the designer with E_NOTIMPL


jnm2:
This is the last thing I know of that blocks SDK-style csproj from being used for .NET Framework Windows Forms projects in Visual Studio 2019 Update 1.

My only reluctance to file this is that I passionately dislike the LICX paradigm. The constantly-generated LICX cache file causes CI build failures and failures between developer machines due to outdated strongly-named versions of assemblies never being removed from the file when newer versions are referenced. This has been so much pain over the years that we banned it from source control. Any other way to license at all, or no licensing, would be extremely welcome for other reasons besides enabling the simplified csproj format.

Therefore, either of these two outcomes is ideal for me:

- DevExpress stops using LicenseManager as part of their licensing (no LICX!) ❤
- The new project system implements what LicenseManager needs, **but** LicenseManager and/or the new project system is changed so that the LICX cache is gone or in the .vs folder, is no longer an embedded resource in the compilation, and does not affect the csproj. ❤

These outcomes would be sad:

- Neither Microsoft or DevExpress changes anything. It is never possible to use the SDK-style csproj for .NET Framework Windows Forms applications. 😞
- The new project system implements what LicenseManager needs exactly the way the old project system does. We're back to the LICX baggage. 😭

See also: https://www.devexpress.com/Support/Center/Question/Details/T740667/designer-fails-to-load

**Steps to Reproduce**:

Create a control with the following code:

```cs
[LicenseProvider(typeof(LicFileLicenseProvider))]
public class SomeThirdPartyControl : Control
{
    private readonly License license;

    public SomeThirdPartyControl()
    {
        license = LicenseManager.Validate(typeof(SomeThirdPartyControl), this);
    }

    protected override void Dispose(bool disposing)
    {
        if (disposing) license?.Dispose();
        base.Dispose(disposing);
    }
}
```

Build the project and then add a new form, open its designer, and add this control using the toolbox.

**Expected Behavior**: The designer should open.

**Actual Behavior**: 

Designer crashes with E_NOTIMPL.

```
at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
at Microsoft.VisualStudio.Design.Serialization.LicenseService.GetLicenseDocData(FileAccess access)
at Microsoft.VisualStudio.Design.Serialization.LicenseService.FillLicensedTypes()
[...]

PR #5468, opened 2019-09-11
Add licenses.licx special file provider

davkean:
Fixes: https://github.com/dotnet/project-system/issues/4938.

This adds a new implementation of AbstractFindByNameSpecialFileProvider called AbstractFindByNameSpecialFileProvider2 that derives from the new AbstractSpecialFileProvider. The original will be removed & replaced in a later PR


davkean:
This works well if you opt out of globs, with globs enabled we need to implement: https://github.com/dotnet/sdk/issues/3631.

PR #5486, opened 2019-09-13
Check for invalid characters in resource file 

etbyrd:
Adds a check for invalid characters in the name of the resource file on the Application property page. It fails validation instead of a warning because the file cannot exist.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
@@ -611,6 +611,9 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                         If Trim(Win32ResourceFile.Text).Length = 0 Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_NeedResFile
                             Return ValidationResult.Warning
+                        ElseIf Not Win32ResourceFile.Text.IndexOfAny(Path.GetInvalidPathChars()) = -1 Then
+                            message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_InvalidCharactersInFilePath
+                            Return ValidationResult.Failed
                         ElseIf Not File.Exists(Win32ResourceFile.Text) Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_ResourceFileNotExist
                             Return ValidationResult.Warning
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
@@ -608,12 +608,12 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                     End If
                 Case VsProjPropId80.VBPROJPROPID_Win32ResourceFile
                     If Win32ResourceRadioButton.Checked Then
+                        Dim FirstInvalidCharacter As Integer = Win32ResourceFile.Text.IndexOfAny(Path.GetInvalidPathChars())
                         If Trim(Win32ResourceFile.Text).Length = 0 Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_NeedResFile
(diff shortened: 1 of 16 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
```diff
@@ -611,6 +611,9 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                         If Trim(Win32ResourceFile.Text).Length = 0 Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_NeedResFile
                             Return ValidationResult.Warning
+                        ElseIf Not Win32ResourceFile.Text.IndexOfAny(Path.GetInvalidPathChars()) = -1 Then
+                            message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_InvalidCharactersInFilePath
```
tmeschter: Is the set of characters returned by `Path.GetInvalidPathChars()` generally small? If so, consider updating the message to list the characters that may not be used.

etbyrd: It looks to be around 36 characters. Do you think that is too many to display?

etbyrd: Or maybe the error could show the invalid characters that are being used?

etbyrd:
Screenshot of error:
invalid

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
drewnoakes: @etbyrd I agree. Showing which characters the user must remove in order for the value to be accepted seems like the sweet spot here.

tmeschter: Hmm. 36 is too many. In that case, either include just the first invalid character in the message, or all the invalid characters. Just the first will be easier and entirely sufficient, I think.

tmeschter: Showing all of them will involve de-duplication and figuring out how to localize a list of arbitrary length. I'm not convinced that level of effort is worth it.

etbyrd: I agree. I think just showing the first invalid character would be enough here.

melytc: I added the first invalid character to the message, what do you think?
Screenshot (8)

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.vb
```diff
@@ -611,6 +611,10 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                         If Trim(Win32ResourceFile.Text).Length = 0 Then
                             message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_NeedResFile
                             Return ValidationResult.Warning
+                        ElseIf Not Win32ResourceFile.Text.IndexOfAny(Path.GetInvalidPathChars()) = -1 Then
+                            Dim FirstInvalidCharacter As String = Path.GetInvalidPathChars().First
+                            message = My.Resources.Microsoft_VisualStudio_Editors_Designer.PropPage_InvalidCharactersInFilePath & vbCrLf & "Please remove the following character: " & FirstInvalidCharacter
```
davidwengier: I don't think this does what you think it does. This will always show a double quote to the user, as that is what is first in `Path.GetInvalidChars()`. What you probably want to do is get the value of `Win32ResourceFile.Text.IndexOfAny(Path.GetInvalidPathChars())` before the `If` block, and then if that value is not -1, add it to the message via `Win32ResourceFile.Text.Substring(theCharacter, 1)`.

melytc: Fixed this, thanks!

PR #5492, opened 2019-09-15
Merge feature/TempPE into master

### LINKED ISSUE #4163: Provide an implementation of BuildManager to produce TempPE assembly during type resolution

davidwengier:
Fixes #4163

In theory this has all been reviewed before.


davkean:
Can we rebase this against master? The majority of commits in this are merge comments.

davidwengier:
Sure thing

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -77,6 +95,18 @@ public object DesignTimeOutputMonikers
...
         {
+            if (DesignTimeInputsBuildManagerBridge?.AppliesTo(_unconfiguredProjectServices.Project.Capabilities) == true)
+            {
+                IDesignTimeInputsBuildManagerBridge bridge = DesignTimeInputsBuildManagerBridge.Value;
+
+                return _threadingService.ExecuteSynchronously(async () =>
+                {
+                    await _threadingService.SwitchToUIThread();
+
+                    return await bridge.GetDesignTimeInputXmlAsync(bstrOutputMoniker);
```
davkean: Rememinder me, why is this async but required on the UI thread?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -97,19 +127,25 @@ void IEventSource<_dispBuildManagerEvents>.OnSinkRemoved(_dispBuildManagerEvents
         /// </summary>
         internal virtual void OnDesignTimeOutputDeleted(string outputMoniker)
         {
-            _threadingService.VerifyOnUIThread();
+            _threadingService.ExecuteSynchronously(async () =>
```
davkean: This is always called on the UI thread, why did we change the assert?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davkean: This can be followup, but it looks like there's some enforced UI-threadiness in this case that isn't needed. The rules you have to follow are:

1) If I'm on the UI thread, I have a guarantee that the state of the project doesn't change underneath me
2) I can be freely called on a background thread (switching to the UI thread if required)

Above in BuildDesignTimeOutput/DesignTimeOutputMonikers I'm forced to the UI thread for what appears to be unneeded reasons. I think you can still have above guarantees without the forced switch to the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: All of the `VerifyOnUIThread` was removed from this class in response to https://github.com/dotnet/project-system/pull/5326#discussion_r317933815 in the last PR.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
davidwengier: This is in response to the same comment as below (https://github.com/dotnet/project-system/pull/5326#discussion_r317933815) regarding switching to the UI thread if necessary, rather than verifying (throwing) if not on the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VsBuildManager.cs
```diff
@@ -29,13 +32,16 @@ internal VSBuildManager(IProjectThreadingService threadingService, IUnconfigured
...
+            _unconfiguredProjectServices = unconfiguredProjectServices;
             Project = new OrderPrecedenceImportCollection<VSLangProj.VSProject>(projectCapabilityCheckProvider: unconfiguredProjectServices.Project);
         }
 
         [ImportMany(ExportContractNames.VsTypes.VSProject)]
         internal OrderPrecedenceImportCollection<VSLangProj.VSProject> Project { get; }
 
+        [Import]
+        internal Lazy<IDesignTimeInputsBuildManagerBridge, IAppliesToMetadataView>? DesignTimeInputsBuildManagerBridge { get; private set; }
+
```
davkean: Why can't this be an import?

davidwengier: It's a circular reference. Will add a comment in a follow up

PR #5494, opened 2019-09-16
Make sure AppDesigner exists before adding template

davkean:
After we moved the template-based special file providers over to AbstractFindByNameUnderAppDesignerSpecialFileProvider, we stopped ensuring that the AppDesigner existed before we added the template, this caused template creation to fail.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppManifestSpecialFileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppManifestSpecialFileProvider.cs
@@ -24,9 +24,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
             _properties = properties;
         }
 
-        protected override Task CreateFileAsync(string path)
+        protected override async Task CreateFileCoreAsync(string path)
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AppManifestSpecialFileProvider.cs
```diff
@@ -24,9 +24,9 @@ internal class AppManifestSpecialFileProvider : AbstractFindByNameUnderAppDesign
             _properties = properties;
         }
 
-        protected override Task CreateFileAsync(string path)
+        protected override async Task CreateFileCoreAsync(string path)
         {
-            return _templateFileCreationService.CreateFileAsync("ResourceInternal.zip", path);
+            await _templateFileCreationService.CreateFileAsync("ResourceInternal.zip", path);
```
drewnoakes: Why add the state machine here but not in `AppSettingsSpecialFileProvider.cs` etc?

davkean: Ah I originally made a change here that required it, and then removed it without removing this change.

PR #5500, opened 2019-09-16
Turn off "Separate using directives groups"

davkean:
Just review: https://github.com/dotnet/project-system/pull/5500/commits/9b6131126e2f0cd504cba55331e620d3b5de1b12


sharwell:
@davkean It sounds like this setting is not enforced by the code style tools regardless of whether it's set to true or false. Can you provide more information about how the project benefits from the change in this pull request?

davkean:
@sharwell 

The commit lists the three bugs we kept running into:

dotnet/roslyn#25003
dotnet/roslyn#27834
dotnet/roslyn#36984

We kept fighting the tools just to stay in the state where the usings were separated and we've given up.

sharwell:
Thanks! We are in the process of updating dotnet-format to support organizing usings. I'll discuss with @JoeRobich about fixing those at the same time.

PR #5501, opened 2019-09-16
Hide broken windows forms templates

### LINKED ISSUE #5398: Issues with Inherited Form or Inherited Control templates
Two problems: 
* At the moment when in a non-WinForms project  the Inherited Control and Inherited Form show in the Add Item dialog when they shouldn't
* The Inheritance Picker window that the templates open up doesn't work in .NET Core projects

The code for these is located here: https://devdiv.visualstudio.com/DevDiv/_git/VS.ref_archive?path=%2Fsrc%2Fcsharp%2FWizards%2FVSZ%2Fcsharpprojectitems&version=GBmaster
The Inheritance Picker window code is here: https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fvsip%2FPackages%2FWin%2FInheritanceDialog.cs&version=GBmaster

These use the older "vsz" template format so don't support capabilities. We need to work out what to do with these:
* Move them to vstemplate?
* Hide them via a manaully written Add Item template filter that explicitly calls these templates out?

davidwengier:
Fixes #5398 

We filter Windows Forms templates by capabilities but there are two that don't support that, so this hides them too.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
@@ -1,0 +1,46 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms
+{
+    /// <summary>
+    /// A filter for the Add New Item dialog that filters out Windows Forms items from non-Windows Forms projects.
+    /// </summary>
+    [ExportProjectNodeComService(typeof(IVsFilterAddProjectItemDlg))]
+    [AppliesTo("!" + ProjectCapability.WindowsForms)]
+    internal class WindowsFormsAddItemFilter : IVsFilterAddProjectItemDlg
+    {
+        public int FilterTreeItemByLocalizedName(ref Guid rguidProjectItemTemplates, string pszLocalizedName, out int pfFilter)
+        {
+            pfFilter = 0;
+            return HResult.NotImplemented;
+        }
+
+        public int FilterTreeItemByTemplateDir(ref Guid rguidProjectItemTemplates, string pszTemplateDir, out int pfFilter)
+        {
+            pfFilter = 0;
+            // Most of the templates for Windows Forms items are filtered by capabilities in the .vstemplate but there are a couple
+            // that use an older .vsz tmeplate format that doesn't support capabilities, so we filter them out here.
+            // The AppliesTo on this class ensures this code only runs for Windows Forms projects
+            if (pszTemplateDir.EndsWith("\\Windows Forms", StringComparisons.Paths))
+            {
+                pfFilter = 1;
+            }
+            return HResult.OK;
+        }
+
+        public int FilterListItemByLocalizedName(ref Guid rguidProjectItemTemplates, string pszLocalizedName, out int pfFilter)
+        {
+            pfFilter = 0;
+            return HResult.NotImplemented;
+        }
+
+        public int FilterListItemByTemplateFile(ref Guid rguidProjectItemTemplates, string pszTemplateFile, out int pfFilter)
+        {
+            pfFilter = 0;
+            return HResult.NotImplemented;
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


tmeschter:
Is there a plan to fix the templates so that they respect the capabilities?

davidwengier:
> Is there a plan to fix the templates so that they respect the capabilities?

Not at this stage. To do so would mean rewriting the templates to the .vstemplate format, and then we'd have to rewrite the wizard so that it works with .NET Core before lighting them up anyway. I need to look at the telemetry to see how often these are used in legacy, but I suspect not often (and I have no idea how to look at the telemetry just yet :P)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms
+{
+    /// <summary>
+    /// A filter for the Add New Item dialog that filters out Windows Forms items from non-Windows Forms projects.
+    /// </summary>
+    [ExportProjectNodeComService(typeof(IVsFilterAddProjectItemDlg))]
+    [AppliesTo("!" + ProjectCapability.WindowsForms)]
```
davkean: This is too broad and will affect every single project-system, please combine _DotNet_.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+            return HResult.NotImplemented;
+        }
+
+        public int FilterTreeItemByTemplateDir(ref Guid rguidProjectItemTemplates, string pszTemplateDir, out int pfFilter)
+        {
+            pfFilter = 0;
+            // Most of the templates for Windows Forms items are filtered by capabilities in the .vstemplate but there are a couple
+            // that use an older .vsz tmeplate format that doesn't support capabilities, so we filter them out here.
+            // The AppliesTo on this class ensures this code only runs for Windows Forms projects
+            if (pszTemplateDir.EndsWith("\\Windows Forms", StringComparisons.Paths))
```
davkean: I don't follow this, this says "only run for WinForms" but the AppliesTo says "!WindowForms". How does this filter out these templates when the project is a WinForms template?

davidwengier: Oops, the comment is worded incorrectly. It should say this "only runs for non-Windows Forms projects"...  but also, I forgot that we don't want the two templates for Winforms projects either, so this whole thing is wrong 🤦‍♂️

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
davkean: @davidwengier I just realized this won't work - "WinForms" is a dynamic capability (comes from targets) and these services need to be fixed for the lifetime of the project. This is enforced and will result in assertions/crashes in certain situations. This needs to be "DotNet" so that it is always present - and then you need to consult the project for the capabilities when the method is called.

davidwengier: >  then you need to consult the project for the capabilities when the method is called

Yep, makes sense based on your other comment too. This is how I had it yesterday and then realized I can clean it up.. guess not :D

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public int FilterTreeItemByLocalizedName(ref Guid rguidProjectItemTemplates, string pszLocalizedName, out int pfFilter)
+        {
+            pfFilter = 0;
+            return HResult.NotImplemented;
+        }
+
+        public int FilterTreeItemByTemplateDir(ref Guid rguidProjectItemTemplates, string pszTemplateDir, out int pfFilter)
+        {
+            pfFilter = 0;
```
davkean: Capture the `_project` field an a variable, and if its null, return E_UNEXPECTED so that we don't return OK on null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/WindowsFormsAddItemFilter.cs
davkean: Added a test to catch this: https://github.com/dotnet/project-system/pull/5520.

PR #5503, opened 2019-09-16
Refactor dimension providers and improve locking

drewnoakes:
Previously the property was read without a project lock.

This also adds uniform support for add/remove/rename for all dimensions.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -153,8 +153,58 @@ public IEnumerable<string> GetBestGuessDimensionNames(ImmutableArray<ProjectProp
...
+                    return ProjectAccessor.OpenProjectXmlForWriteAsync(project, msbuildProject =>
+                    {
+                        action(msbuildProject, evaluatedPropertyValue);
+                    });
+                });
+            }
+        }    
 
         /// <summary>
         /// Gets the value for the specified property of the specified project.
```
davidwengier: Can any of the protected methods below here be removed now?

drewnoakes: I just double checked and as it stands they're all used, but your question prompted me to wonder if we cannot flatten the types further, and I think we can. The override in `TargetFrameworkProjectConfigurationDimensionProvider` doesn't appear to be doing anything different from the base. Investigating.

PR #5504, opened 2019-09-16
Re-work the SDK-Style Projects options page

tmeschter:
At the suggestion of the settings team, this commit makes somes changes to our Tools | Options page to simplify it and bring it in line with standard conventions. This will also simplify the future changes needed to make it work in a cloud environment.

1. Further separate the UI and the settings. The `GeneralOptionPageControl` type (the UI itself) used to hold on to the `GeneralOptions` type (reponsible for storing, loading, and saving the options) directly. `GeneralOptions` has been replaced with `SDKStyleProjectOptionsData`, and the UI and settings and now hooked up with standard XAML binding syntax.
2. Expose the settings through `GeneralOptionPage.AutomationObject`. By overriding the `AutomationObject` property we get a couple of benefits. First, the settings infrastructure will handle loading and saving the settings for us and we can remove the dependency on `ISettingsManager`. Second, the properties on the `AutomationObject` are automatically exposed through DTE.
3. Remove a bunch of code we don't need. `GeneralOptionPage` can derive directly from `UIElementDialogPage`, and `GeneralOptionPageControl` can derive directly from `UserControl`. As such we can eliminate `OptionPage` and `OptionPageControl`.
4. Make `GeneralOptionPageResources` public again. We recently make this type internal as there didn't seem to be any reason to make it public. It turns out this broke the option page entirely as we get a XAML binding error while trying to load it.

With these changes in place making the page work in the cloud environment will be trivial:
1. Mark the page with an attribute indicating it supports the cloud environment.
2. Pass a flag down to the `UIElementDialogPage` constructor indicating that the settings should be copied to the server, rather than being local-only.

Those changes will happen in a future commit.


src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptionPageControl.xaml.vb
```diff
@@ -1,47 +1,49 @@
...
     Partial Friend NotInheritable Class GeneralOptionPageControl
-        Inherits OptionPageControl
-
-        Private ReadOnly _generalOptions As GeneralOptions
+        Inherits UserControl
 
+        ''' <summary>
+        ''' The values shown in the Log Level drop down menu.
+        ''' </summary>
         Public Shared ReadOnly FastUpToDateLogLevelItemSource As String() = {
```
tmeschter: A future change will move this into the XAML file itself.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptionPageControl.xaml.vb
```diff
@@ -1,47 +1,49 @@
...
         }
 
-        Public Sub New(serviceProvider As IServiceProvider)
-            MyBase.New()
-
-            _generalOptions = New GeneralOptions(serviceProvider)
-
-            InitializeComponent()
+        Public Sub New()
+            Dim labelStyle = New Style(GetType(Label))
```
tmeschter: A future change will move all of the styles into the XAML file itself.

src/Microsoft.VisualStudio.Editors/OptionPages/SDKStyleProjectOptionsData.vb
```diff
@@ -0,0 +1,67 @@
...
+            clonedData.CopyFrom(Me)
+            Return clonedData
+        End Function
+
+        Public Sub CopyFrom(source As SDKStyleProjectOptionsData)
+            FastUpToDateCheckEnabled = source.FastUpToDateCheckEnabled
+            FastUpToDateCheckLogLevel = source.FastUpToDateCheckLogLevel
+        End Sub
+
+        <SharedSettings("ManagedProjectSystem\FastUpToDateCheckEnabled", False)>
```
davidwengier: What is `False` here?

tmeschter: `True` for a local-only settings, `False` for a setting that roams. If VB has syntax for named parameters that you can use in attributes, I couldn't figure it out.

davidwengier: I believe its `:=`

davidwengier: Or maybe thats just for optional parameters.. ¯\_(ツ)_/¯
https://github.com/dotnet/project-system/blob/master/src/Common/ManagedCodeMarkers.vb#L23

drewnoakes: https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/concepts/attributes/#attribute-parameters

src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.vbproj
```diff
@@ -50,7 +50,7 @@
       <SubType>Designer</SubType>
     </EmbeddedResource>
     <EmbeddedResource Update="OptionPages\GeneralOptionPageResources.resx">
-      <Generator>ResXFileCodeGenerator</Generator>
+      <Generator>PublicResXFileCodeGenerator</Generator>
```
drewnoakes: Can we add a comment here explaining why this needs to be public so we don't break this again?

src/Microsoft.VisualStudio.Editors/OptionPages/SDKStyleProjectOptionsData.vb
tmeschter: For an attribute, the name can only be specified for the fields or properties of the attribute. The `machineLocal` parameter is an actual parameter of the `SharedSettingsAttribute` constructor, so I can't do this:
``` VB
<SharedSettings("ManagedProjectSystem\FastUpToDateCheckEnabled", machineLocal:=False)>
```
The compiler raises an error that `machineLocal` isn't a field or property.

PR #5516, opened 2019-09-18
Fix hang in VisualStudioWaitIndicator 

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/972274.

VisualStudioWaitIndicator was an "implicitly" active service that was being fired up per implicitly active configuration. In cases where the implicitly active configuration changed in between renaming a file and the rename actually occuring, the Renamer could hang forever waiting for VisualStudioWaitIndicator to "activate" on an inactive configuration. 

VisualStudioWaitIndicator is a simple wrapper around an interface in Visual Studio, there is no need for it to live in an "active" configuration, or any configuration at all - just convert it over to normal service that lives in the global scope.


davkean:
tag @ocallesp

PR #5518, opened 2019-09-18
Various cleanup following the TempPE branch merge

davidwengier:
Because of the change in using statement separation rule, might be easier to review by commit.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
```diff
@@ -45,24 +42,20 @@ internal class DesignTimeInputsBuildManagerBridge : UnconfiguredProjectHostBridg
         /// <summary>
         /// Get the list of design time monikers that need to have TempPE libraries created. Needs to be called on the UI thread.
         /// </summary>
-        public string[] GetTempPEMonikers()
+        public async Task<string[]> GetTempPEMonikers()
```
davkean: Nit: "Async"

Issue #5519, opened 2019-09-18
DependenciesSnapshotProvider is marked with [ProjectAutoLoad] with a dynamic capability


davkean:
DependenciesSnapshotProvider is marked with a dynamic capability "DependenciesTree" yet is marked as an auto-load component. These are in conflict, dynamic capabilities which can be added/removed at any time by switching branches, installing/uninstalling packages, etc must use IProjectDynamicLoadComponent, see: https://github.com/microsoft/VSProjectSystem/blob/master/doc/overview/dynamicCapabilities.md.

Remove the test workaround in the CertainExportsMustNotBeMarkedWithDynamicCapabilities test when this is fixed.

drewnoakes:
Potentially related to #5531.

davkean:
Why do you say that?

drewnoakes:
Because fixing this will likely fix that issue as well.

melytc:
Was this fixed by https://github.com/dotnet/project-system/pull/5520?

drewnoakes:
@melytc no, this is still outstanding.

PR #5520, opened 2019-09-18
Unban ProjectAutoLoad/ConfiguredProjectAutoLoad

davkean:
Have added a new test that verifies correct usage now, of which there is a single violation: https://github.com/dotnet/project-system/issues/5519.

Based on this: https://github.com/dotnet/project-system/pull/5501#discussion_r324947156.

### DIFF
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentVerificationTests.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentVerificationTests.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Reflection;
 using Microsoft.VisualStudio.Composition;
 using Microsoft.VisualStudio.Composition.Reflection;
+using Microsoft.VisualStudio.Packaging;
 using Xunit;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS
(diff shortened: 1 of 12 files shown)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentVerificationTests.cs
```diff
@@ -97,6 +98,69 @@ public void ExportsFromSamePartMustApplyToSameCapabilities(Type type)
...
+                return;
+
+            foreach (KeyValuePair<MemberRef, ExportDefinition> exportDefinitionPair in definition!.ExportDefinitions)
+            {
+                ExportDefinition export = exportDefinitionPair.Value;
+                if (contractsWithFixedCapabilities.Contains(export.ContractName) && 
+                    export.Metadata.TryGetValue(nameof(AppliesToAttribute.AppliesTo), out object metadata) && 
+                    metadata is string appliesTo)
+                {
+                    foreach (string capability in GetDynamicCapabilities())
```
davidwengier: Since in theory any string is a valid dynamic capability, wouldn't it be better to ensure that all of the capabilities in `appliesTo` are part of FixedCapabilities?

davkean: Yep good idea.

davkean:
@drewnoakes An equivalent is IProjectDynamicLoadComponent.

drewnoakes:
`IProjectDynamicLoadComponent` is what I tried (twice) but it doesn't allow the same level of control over timing IIRC (it was a while ago). It always seemed to fire too late, after other requests were made of the component.

PR #5521, opened 2019-09-18
Don't cache the active configured project

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/PasteOrdering.cs
```diff
@@ -18,15 +20,15 @@ internal class PasteOrdering : IPasteHandler, IPasteDataObjectProcessor
     {
         public const int OrderPrecedence = 10000;
 
-        private readonly ConfiguredProject _configuredProject;
+        private readonly ActiveConfiguredProject<ConfiguredProject> _configuredProject;
```
tmeschter: Can you update the commit message to explain how this is different from what we did before?

drewnoakes: Added some longer-form explanation.

PR #5522, opened 2019-09-18
Fix a few minor details found by analyzers

davidwengier:
Found an interesting analyzer, installed it to take a look, it found a few minor fixes.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Waiting/VisualStudioWaitIndicator.cs
```diff
@@ -33,7 +33,7 @@ public void Wait(string title, string message, bool allowCancel, Action<Cancella
         public T Wait<T>(string title, string message, bool allowCancel, Func<CancellationToken, T> action)
         {
             if (typeof(T) == typeof(Task))
-                throw new ArgumentException(nameof(T));
+                throw new ArgumentException("Type argument must not be Task", nameof(T));
```
davidwengier: I'm not in love with this message, but I figure its better than "`T`" :P

tmeschter: Much better. :-)

PR #5523, opened 2019-09-18
Rename files

tmeschter:
Rename a few files to match the name of the classes within.


dmonroym:
>             Common.Switches.TracePDPerfBegin("DesignerTabControlRenderer.PerformLayout()")

Should these be changed as well?

---
Refers to: src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ProjectDesignerTabRenderer.vb:306 in 8f9385c. [](commit_id = 8f9385c5dceea3578d70c8a6ef263c056c04c954, deletion_comment = False)

Issue #1256, opened 2017-01-19
"View In Object Browser" does nothing for packages


davkean:
1. Create a Console App (.NET Core)
2. Expand Dependencies -> NuGet
3. Right-click on Microsoft.NETCore.App and choose View In Object Browser

-- Nothing --

I suspect we want to disable this, unless we support opening all of it's assemblies in the object browser.

abpiskunov:
dupe/related of this one https://github.com/dotnet/roslyn-project-system/issues/403

davkean:
I don't think this is a dupe, at least in the assembly/project version we should just use the existing infrastructure to open them. In this case, what would we even do? Loop over every reference and view them?

abpiskunov:
sure. For package i was planning to hide it as part of #403. Let's keep this issue to split the work and simplify reviews :)

abpiskunov:
Note: after cross targeting dependencies change , View in object browser needs to be handled on roslyn side completely for any kind of dependencies: assemblies and project we support it, for packages don't show context menu at all.   CPS implementation of this command really hardcoded to IProjectItemTree , but we have IProjectItemTree items only for active target , not for others.  

We need 
- handle command ourselves to control visibility http://index/?rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Package%5cComponents%5cDefaultCommandSet2KHandler.cs&line=133
- handle actual command execution to show assembly or project in Object browser  http://index/?rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Package%5cComponents%5cDefaultVsUIHierarchyWindowCmdsHandler.cs&line=175

BenjaminHolland:
@abpiskunov The links provided are broken.

abpiskunov:
Find files and lines from links in actual CPS code instead of index please.

Thanks,
Anton

On Jun 18, 2017, at 12:59 PM, Ben Holland <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #5533, opened 2019-09-19
Suppress "Open in Object Browser" for NuGet packages

drewnoakes:
Fixes part of #1256.

There's nothing obviously correct to show for NuGet packages themselves in the Object Browser.

Instead we should in future allow the Object Browser to be launched from child nodes representing assemblies within the packages themselves (the other part of #1256).

Legacy doesn't offer Open in Object Browser for packages either, though if using `packages.config` then package assemblies are added directly as references and do work with the Object Browser.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
+{
+    /// <summary>
+    /// Command handler that provides special handling for standard commands on dependency nodes.
+    /// </summary>
+    /// <remarks>
+    /// CPS provides a handler for the same commands, however it does not know anything special about dependency nodes.
+    /// </remarks>
+    [ExportCommandGroup(VSConstants.CMDSETID.StandardCommandSet2K_string)]
+    [AppliesTo(ProjectCapability.DependenciesTree)]
+    [Order(10)]
+    internal sealed class DependenciesNodeCommandGroupHandler : ICommandGroupHandler
+    {
+        private static readonly CommandStatusResult s_suppressedResult = new CommandStatusResult(handled: true, null, CommandStatus.NotSupported | CommandStatus.Invisible);
+
+        public CommandStatusResult GetCommandStatus(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, string commandText, CommandStatus progressiveStatus)
+        {
+            var cmdId = (VSConstants.VSStd2KCmdID)commandId;
+
+            switch (cmdId)
+            {
+                case VSConstants.VSStd2KCmdID.QUICKOBJECTSEARCH: // Open in Object Browser
+                {
+                    if (nodes.Count == 1)
+                    {
+                        IProjectTree tree = nodes.First();
+
+                        if (tree.Flags.Contains(DependencyTreeFlags.NuGetPackageDependency))
+                        {
+                            // Suppress "Open in Object Browser" for NuGet packages
+                            return s_suppressedResult;
+                        }
+                    }
+
+                    break;
+                }
+            }
+
+            return CommandStatusResult.Unhandled;
+        }
+
+        public bool TryHandleCommand(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, long commandExecuteOptions, IntPtr variantArgIn, IntPtr variantArgOut)
+        {
+            return false;
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
@@ -1,53 +1,33 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Immutable;
-using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Input;
+using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
 {
-    /// <summary>
-    /// Command handler that provides special handling for standard commands on dependency nodes.
-    /// </summary>
-    /// <remarks>
-    /// CPS provides a handler for the same commands, however it does not know anything special about dependency nodes.
-    /// </remarks>
-    [ExportCommandGroup(VSConstants.CMDSETID.StandardCommandSet2K_string)]
-    [AppliesTo(ProjectCapability.DependenciesTree)]
-    [Order(10)]
-    internal sealed class DependenciesNodeCommandGroupHandler : ICommandGroupHandler
+    [ProjectCommand(VSConstants.CMDSETID.StandardCommandSet2K_string, (long)VSConstants.VSStd2KCmdID.QUICKOBJECTSEARCH)]
+    [AppliesTo(ProjectCapability.PackageReferences)]
+    [Order(ProjectSystem.Order.Default)]
+    internal sealed class SuppressObjectBrowserForPackageReferenceCommand : AbstractSingleNodeProjectCommand
     {
-        private static readonly CommandStatusResult s_suppressedResult = new CommandStatusResult(handled: true, null, CommandStatus.NotSupported | CommandStatus.Invisible);
-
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
+{
+    /// <summary>
+    /// Command handler that provides special handling for standard commands on dependency nodes.
+    /// </summary>
+    /// <remarks>
+    /// CPS provides a handler for the same commands, however it does not know anything special about dependency nodes.
+    /// </remarks>
+    [ExportCommandGroup(VSConstants.CMDSETID.StandardCommandSet2K_string)]
+    [AppliesTo(ProjectCapabilities.Cps)]
```
davkean: This is too general and will cause us to load in situations we shouldn't, this should be tied to PackageReference or DependenciesTree.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public CommandStatusResult GetCommandStatus(IImmutableSet<IProjectTree> nodes, long commandId, bool focused, string commandText, CommandStatus progressiveStatus)
+        {
+            var cmdId = (VSConstants.VSStd2KCmdID)commandId;
+
+            switch (cmdId)
+            {
+                case VSConstants.VSStd2KCmdID.QUICKOBJECTSEARCH: // Open in Object Browser
+                {
+                    if (nodes.Count == 1)
```
davkean: Why do we only disable command if a single node is selected?

davkean:
> There's nothing obviously correct to show for NuGet packages themselves in the Object Browser.

We could show all the assemblies of the package?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
drewnoakes: I'll tie it to DependenciesTree.

drewnoakes:
> We could show all the assemblies of the package?

I'm not sure this is great. Which do we select? Does it just load all of them into the Object Browser, then focus the first or last? If a package has dependencies on other packages, would we load them in too? This feels like no matter what we choose it'd surprise _someone_. I'm thinking it's easier to allow selection of specific assemblies, but I'm willing to be convinced otherwise.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
drewnoakes: This mirrors CPS's implementation which only enables "Open in Object Browser" when you have a single item selected, presumably to avoid the issues around loading multiple items into the Object Browser.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// Command handler that provides special handling for standard commands on dependency nodes.
+    /// </summary>
+    /// <remarks>
+    /// CPS provides a handler for the same commands, however it does not know anything special about dependency nodes.
+    /// </remarks>
+    [ExportCommandGroup(VSConstants.CMDSETID.StandardCommandSet2K_string)]
+    [AppliesTo(ProjectCapability.DependenciesTree)]
+    [Order(10)]
+    internal sealed class DependenciesNodeCommandGroupHandler : ICommandGroupHandler
```
davkean: We have a concept called ProjectCommand and a derived type that simplifies this ceremony.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Command handler that provides special handling for standard commands on dependency nodes.
+    /// </summary>
+    /// <remarks>
+    /// CPS provides a handler for the same commands, however it does not know anything special about dependency nodes.
+    /// </remarks>
+    [ExportCommandGroup(VSConstants.CMDSETID.StandardCommandSet2K_string)]
+    [AppliesTo(ProjectCapability.DependenciesTree)]
+    [Order(10)]
```
davkean: Order.Default.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
davkean: Are we sure this isn't ECMD_VIEWREFINOBJECTBROWSER?

davkean: That's legacy.

davkean: Ah, weirdly the managed value is named differently to its native equivalent, they are the same value.

davkean:
I think creating a new "component set" with the entire contents of the package wouldn't confuse folks, I'm not sure if that's possible. I'm okay with disabling this for now though.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
drewnoakes: `AbstractSingleNodeProjectCommand` with `ProjectCommandAttribute` looks like what you mean. That'd reduce the code a little, though makes it async. I've updated it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+    internal sealed class SuppressObjectBrowserForPackageReferenceCommand : AbstractSingleNodeProjectCommand
+    {
+        private static readonly Task<CommandStatusResult> s_suppressedResult = Task.FromResult(new CommandStatusResult(handled: true, null, CommandStatus.NotSupported | CommandStatus.Invisible));
+        
+        protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree node, bool focused, string? commandText, CommandStatus progressiveStatus)
+        {
+            if (node.Flags.Contains(DependencyTreeFlags.NuGetPackageDependency))
+            {
+                // Suppress "Open in Object Browser" for NuGet packages
+                return s_suppressedResult;
```
davkean: Can you add another property to `GetCommandStatusResult` called `NotSupported`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+         
+            return CommandStatusResult.Unhandled.AsTask();
```
davkean: GetCommandStatusResult.Unhandled.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/DependenciesNodeCommandGroupHandler.cs
drewnoakes: This isn't just `NotSupported`, it's also `Invisible`. Both are important, to ensure CPS doesn't show the item.

davkean: There's 4 different states:

- I don't handle the command
- I do handle the command
- The command is disabled
- The command is disabled and hidden

We have states for the first two, we can come up with a name that matches the last one.

drewnoakes: In 12ed9d8f09f783850238f5ffb13ab116338c300e I went with `GetCommandStatusResult.Suppressed`.

PR #5534, opened 2019-09-19
Rename method

davkean:
Give this a name that better describes it purpose.


davkean:
I went down a different approach.

PR #5535, opened 2019-09-19
Wrap IVsAddProjectItemDlg usage in a service

davkean:
Introduce a new service called IAddItemDialogService that is responsible for showing the add item and add existing item dialogs.

This is untested as of yet.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/AddItemDialogService.cs
```diff
@@ -0,0 +1,97 @@
...
+                __VSADDITEMFLAGS.VSADDITEM_SuggestTemplateName | 
+                __VSADDITEMFLAGS.VSADDITEM_AllowHiddenTreeView);
+        }
+
+        public Task<bool> ShowAddNewItemDialogAsync(IProjectTree node, string directoryLocalizedName, string templateLocalizedName)
+        {
+            Requires.NotNullOrEmpty(directoryLocalizedName, nameof(directoryLocalizedName));
+            Requires.NotNullOrEmpty(templateLocalizedName, nameof(templateLocalizedName));
+
+            return ShowDialogAsync(node,
```
davidwengier: Should pass `directoryLocalizedName` and `templateLocalizedName` through

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/AddItemDialogService.cs
```diff
@@ -0,0 +1,97 @@
...
+            return ShowDialogAsync(node, 
+                __VSADDITEMFLAGS.VSADDITEM_AddExistingItems | 
+                __VSADDITEMFLAGS.VSADDITEM_AllowMultiSelect | 
+                __VSADDITEMFLAGS.VSADDITEM_AllowStickyFilter | 
+                __VSADDITEMFLAGS.VSADDITEM_ProjectHandlesLinks);
+        }
+
+        private async Task<bool> ShowDialogAsync(IProjectTree node, __VSADDITEMFLAGS flags, string? localizedDirectoryName = null, string? localizedTemplateName = null)
+        {
+            string path = _projectTree.TreeProvider.GetAddNewItemDirectory(node);
```
davidwengier: If path can be null, this should be `string?`

PR #5538, opened 2019-09-20
Use References folder context menu

davkean:
We were using IDM_VS_CTXT_REFERENCE context menu for the Packages/Projects/References node because CPS throught they were "References". We've stopped marking them as as Reference and instead ReferenceFolder, which causes CPS to use the IDM_VS_CTXT_REFERENCEROOT context menu.

Also simplified the number of flags we were using for the nodes - they aren't being used anywhere and we can them back if we need them.

Before:
image

After:

image

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
@@ -66,7 +66,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
         /// </summary>
         public static readonly ProjectTreeFlags DependencyFlags
                 = ProjectTreeFlags.Create("Dependency")
-                + ProjectTreeFlags.Create(ProjectTreeFlags.Common.VirtualFolder.ToString()) // TODO why ToString here?
+                + ProjectTreeFlags.Create(ProjectTreeFlags.Common.VirtualFolder)
                 + ProjectTreeFlags.Create(ProjectTreeFlags.Common.BubbleUp)
                 + SupportsRuleProperties
                 + SupportsRemove;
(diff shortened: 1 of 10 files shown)


davkean:
tag @donnie-msft, you were looking into doing this, right?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
```diff
@@ -88,27 +88,23 @@ public static class DependencyTreeFlags
                 + GenericDependency;
 
         internal static readonly ProjectTreeFlags TargetNode = ProjectTreeFlags.Create("TargetNode");
-        internal static readonly ProjectTreeFlags SubTreeRootNode = ProjectTreeFlags.Create("SubTreeRootNode");
+        internal static readonly ProjectTreeFlags SubTreeRootNode = ProjectTreeFlags.Create(ProjectTreeFlags.Common.VirtualFolder) +
```
drewnoakes: This should be named as `SubTreeRootNodeFlags` for consistency with other fields.

davkean: It should be the reverse; Flags should be removed from the names. We know they are flags based on the type name. Feel free to follow up with that change - I've didn't change the name.

drewnoakes: Members with `Flags` suffix are composite. Those without are singular.

donnie-msft:
> tag @donnie-msft, you were looking into doing this, right?
I wasn't looking to modify the ProjectSystem directly, but instead to use MEF and create a `IProjectItemContextMenuProvider`. The flags of interest were `DependenciesRootNode`, `NuGetSubTreeRootNode`, `NuGetPackageDependency`, and `SdkSubTreeRootNode`. 

I'm not super familiar with PS, yet, but my understanding of your changes is that I won't have at least a couple of these flags any longer. Not immediately clear on what I need to do different. 

My goal is to add some NuGet items to various context menus, and this is the approach I was starting off with. It's a WIP, so maybe we should have a chat about this? 

//@davkean

drewnoakes:
@donnie-msft, @davkean and I worked on this PR together, and he's now on leave for a few weeks. We had your goals in mind. Ping me when you're thinking of next picking this up and we'll take a look together.

PR #5558, opened 2019-10-03
Minor updates to Code Analysis property page

mavasani:
1. Bump up the latest stable version to 2.9.5
2. Add a temporary workaround for https://github.com/NuGet/Home/issues/8616, which seems to avoid the underlying issue.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
@@ -18,7 +18,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         Private Const NuGetPackageManagerSearchProviderGuid As String = "042C2B4B-C7F7-49DB-B7A2-402EB8DC7892"
         Private Const RoslynAnalyzersDocumentationLink As String = "https://docs.microsoft.com/visualstudio/code-quality/roslyn-analyzers-overview"
 
-        Private Shared ReadOnly s_latestStableVersion As Version = New Version(2, 9, 3)
+        Private Shared ReadOnly s_latestStableVersion As Version = New Version(2, 9, 5)
         Private Shared ReadOnly s_childPackageIds As HashSet(Of String) = New HashSet(Of String)(
             {"Microsoft.CodeQuality.Analyzers", "Microsoft.NetCore.Analyzers", "Microsoft.NetFramework.Analyzers"},
             StringComparer.OrdinalIgnoreCase)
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -18,7 +18,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         Private Const NuGetPackageManagerSearchProviderGuid As String = "042C2B4B-C7F7-49DB-B7A2-402EB8DC7892"
         Private Const RoslynAnalyzersDocumentationLink As String = "https://docs.microsoft.com/visualstudio/code-quality/roslyn-analyzers-overview"
 
-        Private Shared ReadOnly s_latestStableVersion As Version = New Version(2, 9, 3)
+        Private Shared ReadOnly s_latestStableVersion As Version = New Version(2, 9, 5)
```
davidwengier: Is this going to need to be bumped all the time?

mavasani:
Thanks @davidwengier. We actually found a critical bug in 2.9.5 package, so we are going to publish a 2.9.6 package later today, and I will update this PR accordingly.

davidwengier:
Surely there is a better way to do this? Can the Nuget API be used to discover the latest stable version?

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
mavasani: Yes, unfortunately none of the NuGet APIs allow us to specify an upper bound. We cannot use the NuGet API that always installs the latest version because in future we are likely to release newer versions that require next VS release as the minimum version.

mavasani:
> Surely there is a better way to do this? Can the Nuget API be used to discover the latest stable version?

I replied in https://github.com/dotnet/project-system/pull/5558#discussion_r331277609. I did not find any NuGet VS API that allows us to find a latest stable version, while enforcing a strict upper bound.

mavasani:
@davidwengier If you know about such an API, I'd be very happy to change the implementation in a follow-up PR. Thanks!

src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
davidwengier: That is tough, but even if they did, you wouldn't really know what to set as the upper bound.

We recently moved to using versions that match VS, ie our DLLs are now v16.4. I don't suppose you would do the same? Though that assumes there is an API for Nuget that would let you specify a base version and have it pick up the latest patch.

mavasani: We match the major and minor version of our analyzer packages with underlying Microsoft.CodeAnalysis on which the package is built. So, our 2.x packages need at least Microsoft.CodeAnalysis 2.x, 3.x packages need Microsoft.CodeAnalysis 3.x and so on. So, if such an API was provided, we could set 4.0 as an exclusive upper bound, so any latest package earlier then 4.0 should be installed.

mavasani:
@davidwengier The PR is ready to merge whenever tests pass. Note that I do not have merge permissions on this repo, so I would appreciate if you can merge on my behalf. Thanks!

PR #5578, opened 2019-10-14
Correctly check if the user wants to rename the symbol

jmarolf:
Somewhere along the way the check go reversed so we never actually attempt to do a rename.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
@@ -91,7 +91,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
                 return;
 
             // Ask if the user wants to rename the symbol
-            if (await CheckUserConfirmation(oldName)) != true)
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -91,7 +91,7 @@ internal async Task HandleRenameAsync(string oldFilePath, string newFilePath)
                 return;
 
             // Ask if the user wants to rename the symbol
-            if (await CheckUserConfirmation(oldName))
+            if (await CheckUserConfirmation(oldName)) != true)
```
davkean: I probably broke this due to misreading this check, IMO this reads a little better:

``` C#
if (!await CheckUserConfirmation(oldName))
```

tmeschter: Or maybe

``` C#
bool userWantsToRenameSymbol = await CheckUserConfirmation(oldName);
if (!userWantsToRenameSymbol)
    ...
```

`!await` does look a little funny to me...

PR #5591, opened 2019-10-20
Remove unneeded DependenciesSnapshotProvider interfaces

panopticoncentral:
DependenciesSnapshotProvider implements and exports two internal interfaces, ICrossTargetSubscriptionsHost and IDependenciesSnapshotProvider. Since this is the only implementer and both interfaces are internal, we don't really need them, we can just export the class itself and use it.


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -20,14 +20,24 @@
...
+    /// <summary>
+    /// Provides immutable dependencies snapshot for a given project.
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    [Export(typeof(DependenciesSnapshotProvider))]
     [AppliesTo(ProjectCapability.DependenciesTree)]
-    internal sealed partial class DependenciesSnapshotProvider : OnceInitializedOnceDisposedAsync, ICrossTargetSubscriptionsHost, IDependenciesSnapshotProvider
+    internal sealed partial class DependenciesSnapshotProvider : OnceInitializedOnceDisposedAsync
     {
         public const string DependencySubscriptionsHostContract = "DependencySubscriptionsHostContract";
```
davidwengier: This can go too

Issue #5592, opened 2019-10-20
IntegratedConsoleDebugging project capability not respected 


aodl:
**Visual Studio Version**:
16.2.0

**Project Type**
ASP .Net Core 2.2

**Summary**:
There is an 'IntegratedConsoleDebugging' project capability that could be detected to decide whether or not to launch inside the integrated console, but it isn't used. See [issue 3667](https://github.com/dotnet/project-system/issues/3667) (which indicates that the 'IntegratedConsoleDebugging' should be used) and the [resolution](https://github.com/dotnet/project-system/pull/4388) (which doesn't check for the project capability)

The only way to opt out of the integrated console debugging experience appears to be to select the 'Automatically close the console when debugging stops' debug option. 

**Expected Behavior**:
Check the project capabilities to see if 'IntegratedConsoleDebugging' is included. If not, expect that the app will not be launched for debugging in the integrated console.

**Actual Behavior**:
Regardless of whether or not the app has 'IntegratedConsoleDebugging' project capability specified, it ends up being launched for debugging in the integrated console.

**User Impact**:
The 'Automatically close the console when debugging stops' debug option can't be specified at the project level and isn't very flexible. Making use of the 'IntegratedConsoleDebugging' capability as a requisite for launching in the integrated console would provide flexibility for VSIX extensions to temporarily remove the capability and bypass the integrated console for special launches (temporarily) without needing to fiddle with an IDE-wide option that actually gets persisted (and therefore would cause problems if VS suddenly crashed without undoing the change to the IDE option).

-------------

Update: Similarly `DebugLaunchOptions.IntegratedConsole` isn't specified as a debug options flag, despite the launch taking place inside the integrated console

drewnoakes:
@aodl would you like to submit a PR for this? It seems like you're clear on what's needed.

aodl:
I'd be happy to give it a try, but I'm not currently aware of where the decision to launch inside the integrated console is made, and why it's made there. Or how to actually tell VS to run an app in the integrated console (I have a feeling it's outside of CPS). @davkean, would you be able give me some pointers?

[...]

PR #5594, opened 2019-10-21
Respect IntegratedConsoleDebugging capability, issue 5592

aodl:
https://github.com/dotnet/project-system/issues/5592

Don't launch in integrated console if the project doesn't declare `IntegratedConsoleDebugging` capability

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ConsoleDebugTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ConsoleDebugTargetsProvider.cs
@@ -36,10 +36,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         private readonly ProjectProperties _properties;
         private readonly IProjectThreadingService _threadingService;
         private readonly IVsUIService<IVsDebugger10> _debugger;
-        private readonly UnconfiguredProject _project;
+        private readonly ConfiguredProject _project;
 
         [ImportingConstructor]
-        public ConsoleDebugTargetsProvider(UnconfiguredProject project,
+        public ConsoleDebugTargetsProvider(ConfiguredProject project,
                                            IDebugTokenReplacer tokenReplacer,
                                            IFileSystem fileSystem,
                                            IEnvironmentHelper environment,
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ConsoleDebugTargetsProvider.cs
```diff
@@ -336,7 +336,9 @@ private async Task<DebugLaunchSettings> GetConsoleTargetForProfile(ILaunchProfil
             bool useCmdShell = false;
             if (await IsConsoleAppAsync())
             {
-                if (await IsIntegratedConsoleEnabledAsync())
+                var isIntegratedConsoleCapable = _project.Capabilities.Contains(ProjectCapabilities.IntegratedConsoleDebugging);
```
davkean: Your first guess was on the right track, `ConsoleDebugTargetsProvider` lives in the `ConfiguredProject` scope because it [exports this](https://github.com/dotnet/project-system/blob/02514b0f9ac016b6ecc19e1fb93eed9646f0ee87/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IDebugProfileLaunchTargetsProvider.cs#L15) that means that this capabilities check should be based on `ConfiguredProject.Capabilities`.

You can get your hands on a `ConfiguredProject`, just by adding another argument `ConfiguredProject configuredProject` to the constructor

aodl: Thanks, I've addressed that now

aodl:
hey @davkean, is there anything else I can do on this for the moment? I suppose a consequence of releasing this is that various project types will need to declare `IntegratedConsoleDebugging` capability whereas they didn't need to before.

davkean:
Sorry for the late reply, the entire team has been at conferances this week. I think we need to make one more change, which is opt ourselves into this capability in [Microsoft.Managed.Designtime.targets](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L102), otherwise this will turn off the feature for us. I'll confirm this tomorrow (Australian time) and push this change to this branch if so.,

davkean:
I couldn't push to this branch due to permissions issues, I've created a new PR here that squashes your commits and adds an additional commit on top: https://github.com/dotnet/project-system/pull/5605 so that we don't break.

PR #5598, opened 2019-10-27
Pass in RunAnalyzers and RunAnalyzersDuringLiveAnalysis properties to…

mavasani:
… LanguageService

We recently added project properties for configuring analyzer execution during build and live analysis, see documentation [here](https://docs.microsoft.com/visualstudio/code-quality/disable-code-analysis). Currently, these properties are only respected for NuGet based analyzers. We plan to respect these properties even for VSIX based analyzers, including the built-in IDE code style analyzers. This PR passes the required properties down to the Roslyn langauge service, which will be updated with a Roslyn PR to consume these properties and add this support.


davidwengier:
@mavasani Just to confirm, this is going in to master, so would be 16.5. Is that okay?

mavasani:
> Just to confirm, this is going in to master, so would be 16.5. Is that okay?

Yes, that is fine.

@davidwengier Can you please merge whenever convenient?

PR #5602, opened 2019-10-29
Make sure screen readers are aware of checkable menu item

davkean:
This makes sure that the Debug profile and framework switchers are known to be checkable, which helps screen readers tell users that. After this change, Before this change, Windows Narrater says "menu item", after this change it says "menu item, unchecked" for unchecked items.

image

Fixes our part of: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/642995

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -26,7 +26,7 @@ namespace Microsoft.VisualStudio.Packaging
         new string[] { "SolutionHasProjectCapability:.NET & CPS" }
         )]
 
-    [ProvideMenuResource("Menus.ctmenu", 3)]
+    [ProvideMenuResource("Menus.ctmenu", 4)]
     internal partial class ManagedProjectSystemPackage : AsyncPackage
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -26,7 +26,7 @@ namespace Microsoft.VisualStudio.Packaging
         new string[] { "SolutionHasProjectCapability:.NET & CPS" }
         )]
 
-    [ProvideMenuResource("Menus.ctmenu", 3)]
+    [ProvideMenuResource("Menus.ctmenu", 4)]
```
drewnoakes: Why did this number change?

davkean: It's a version number of the file.

drewnoakes: Is it tracked externally?

davkean: Yeah, it's used for caching purposes I believe. I'm not entirely sure how it works TBH.

PR #5618, opened 2019-11-01
Avoid P2P reference notification on solution close

### LINKED ISSUE #2640: Dependency node is doing lots of work on unload
The dependency node hooks onto Project unload and then goes and notifies every single project in the solution - that's consumes large amounts of time/memory. We should be smarter about this, especially in solution close case.

drewnoakes:
Fixes #2640.

Currently when a project is closed, the dependencies node notifies any projects that depend upon it to update their dependency tree as that project is no longer resolved.

This is useful when, for example, a subset of projects are unloaded.

However when a solution is being closed there's no need for a project referrer to update its tree in response to unloading projects. Doing so creates a lot of redundant tree updates at a time when we should just be tearing everything down.

This PR avoids doing that redundant work.

After this change, looking in PerfView at an ETL trace, I see almost no activity within the dependencies node during solution close, whereas beforehand there was a moderate amount.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -33,6 +33,7 @@ namespace Microsoft.VisualStudio.Packaging
         public const string PackageGuid = "860A27C0-B665-47F3-BC12-637E16A1050A";
 
         private IDotNetCoreProjectCompatibilityDetector? _dotNetCoreCompatibilityDetector;
+        private SolutionService? _solutionService;
         private IVsRegisterProjectSelector? _projectSelectorService;
         private uint _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
 
@@ -52,6 +53,8 @@ namespace Microsoft.VisualStudio.Packaging
 
             var componentModel = (IComponentModel)(await GetServiceAsync(typeof(SComponentModel)));
             Lazy<DebugFrameworksDynamicMenuCommand> debugFrameworksCmd = componentModel.DefaultExportProvider.GetExport<DebugFrameworksDynamicMenuCommand>();
+            _solutionService = (SolutionService)componentModel.GetService<ISolutionService>();
+            _solutionService.StartListening();
 
             var mcs = (OleMenuCommandService)await GetServiceAsync(typeof(IMenuCommandService));
             mcs.AddCommand(debugFrameworksCmd.Value);
@@ -75,10 +78,15 @@ namespace Microsoft.VisualStudio.Packaging
 
         protected override void Dispose(bool disposing)
         {
-            if (disposing && _projectSelectorCookie != VSConstants.VSCOOKIE_NIL)
+            if (disposing)
             {
-                _projectSelectorService?.UnregisterProjectSelector(_projectSelectorCookie);
-                _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
+                if (_projectSelectorCookie != VSConstants.VSCOOKIE_NIL)
+                {
+                    _projectSelectorService?.UnregisterProjectSelector(_projectSelectorCookie);
+                    _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
+                }
+
+                _solutionService?.Dispose();
             }
 
             base.Dispose(disposing);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -33,7 +33,6 @@ namespace Microsoft.VisualStudio.Packaging
         public const string PackageGuid = "860A27C0-B665-47F3-BC12-637E16A1050A";
 
         private IDotNetCoreProjectCompatibilityDetector? _dotNetCoreCompatibilityDetector;
-        private SolutionService? _solutionService;
         private IVsRegisterProjectSelector? _projectSelectorService;
         private uint _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
 
@@ -53,8 +52,8 @@ namespace Microsoft.VisualStudio.Packaging
 
             var componentModel = (IComponentModel)(await GetServiceAsync(typeof(SComponentModel)));
             Lazy<DebugFrameworksDynamicMenuCommand> debugFrameworksCmd = componentModel.DefaultExportProvider.GetExport<DebugFrameworksDynamicMenuCommand>();
-            _solutionService = (SolutionService)componentModel.GetService<ISolutionService>();
-            _solutionService.StartListening();
+            var solutionService = (SolutionService)componentModel.GetService<ISolutionService>();
+            solutionService.StartListening();
 
             var mcs = (OleMenuCommandService)await GetServiceAsync(typeof(IMenuCommandService));
             mcs.AddCommand(debugFrameworksCmd.Value);
@@ -85,8 +84,6 @@ namespace Microsoft.VisualStudio.Packaging
                     _projectSelectorService?.UnregisterProjectSelector(_projectSelectorCookie);
                     _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
                 }
-
-                _solutionService?.Dispose();
             }
 
             base.Dispose(disposing);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
@@ -1,0 +1,102 @@
+﻿// Copyright(c) Microsoft.All Rights Reserved.Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.Composition;
+using System.Threading;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    /// <inheritdoc cref="ISolutionService"/>
+    [Export(typeof(ISolutionService))]
+    internal sealed class SolutionService : ISolutionService, IVsSolutionEvents, IVsPrioritizedSolutionEvents, IDisposable
+    {
+        private const int StateUninitialized = 0;
+        private const int StateListening = 1;
+        private const int StateDisposed = 2;
+
+        private readonly IVsUIService<IVsSolution> _solution;
+        private readonly JoinableTaskContext _joinableTaskContext;
+
+        private int _state = StateUninitialized;
+        private uint _cookie = VSConstants.VSCOOKIE_NIL;
+        
+        /// <inheritdoc />
+        public bool IsSolutionClosing { get; private set; }
+
+        [ImportingConstructor]
+        public SolutionService(IVsUIService<SVsSolution, IVsSolution> solution, JoinableTaskContext joinableTaskContext)
+        {
+            _solution = solution;
+            _joinableTaskContext = joinableTaskContext;
+        }
+
+        public void StartListening()
+        {
+            Assumes.True(_joinableTaskContext.IsOnMainThread, "Must be called on the UI thread.");
+
+            if (Interlocked.CompareExchange(ref _state, StateListening, StateUninitialized) != StateUninitialized)
+            {
+                return;
+            }
+
+            IVsSolution? solution = _solution.Value;
+            Assumes.Present(solution);
+
+            Verify.HResult(solution.AdviseSolutionEvents(this, out _cookie));
+        }
+
+        public int OnAfterOpenSolution(object pUnkReserved, int fNewSolution)                                 => UpdateClosing(false);
+        public int OnBeforeCloseSolution(object pUnkReserved)                                                 => UpdateClosing(true);
+        public int OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)                                    => HResult.OK;
+        public int OnQueryCloseProject(IVsHierarchy pHierarchy, int fRemoving, ref int pfCancel)              => HResult.OK;
+        public int OnBeforeCloseProject(IVsHierarchy pHierarchy, int fRemoved)                                => HResult.OK;
+        public int OnAfterLoadProject(IVsHierarchy pStubHierarchy, IVsHierarchy pRealHierarchy)               => HResult.OK;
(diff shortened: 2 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
```diff
@@ -117,6 +120,8 @@ public int OnBeforeUnloadProject(IVsHierarchy pRealHierarchy, IVsHierarchy pStub
 
         public int OnAfterOpenSolution(object pUnkReserved, int fNewSolution)
         {
+            IsSolutionClosing = false;
+
             return HResult.NotImplemented;
```
swesonga: Is NotImplemented still the correct return value for these methods? Wouldn't HResult.OK be a more accurate reflection of what the code does?

drewnoakes: Perhaps. It might make sense to return `HResult.OK` from all of these as `NotImplemented` doesn't appear to receive any special treatment. There seems to be an outage for symbols right now, so I'm unable to look into `msenv.dll` to see the exact behaviour. I've tested extensively with `NotImplemented` so am reluctant to change it now as I'd like to move on to other things.

davidwengier: There is always //index: http://index/?query=%60OnAfterOpenSolution&rightProject=VS(Repo)&file=src%5Cenv%5Cmsenv%5Ccore%5Cvsslnpst.cpp&line=21425

Looks like OK might be better, as NotImplemented could cause it to not be fired again, though it actually looks like the handling of the Prioritized handlers makes the normal ones redundant as the return value is ignored from them.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
```diff
@@ -152,11 +159,15 @@ public int PrioritizedOnBeforeUnloadProject(IVsHierarchy pRealHierarchy, IVsHier
...
         public int PrioritizedOnAfterOpenSolution(object pUnkReserved, int fNewSolution)
         {
+            IsSolutionClosing = false;
+
             return HResult.NotImplemented;
         }
 
         public int PrioritizedOnBeforeCloseSolution(object pUnkReserved)
         {
+            IsSolutionClosing = true;
```
davkean: Pick Prioritized or not-Prioritized, not both. They are the same events just one (Prioritized) fires earlier than the other.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
davkean: I think you'll need to change the design of this slightly; this currently only ever listens to solution events if the project has ".NET" capability, dependencies node can be present without that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
drewnoakes: > just one (Prioritized) fires earlier than the other

I thought the same, and it's true during load, but during unload/close the order is reversed.

It's probably still safe to only used the prioritized methods, but I think the current approach is also safe so long as load/unload cannot overlap (which seems like a reasonable assumption).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
drewnoakes: @davkean where is this tied to .NET capability?

drewnoakes: Index shows that the return value only matters for `OnBeforeCloseSolution`. If `NotImplemented` is returned, it won't call again. I've changed them all to `Ok` and tested that things still work correctly.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/ProjectRuleHandler.cs
```diff
@@ -39,7 +39,8 @@ internal class ProjectRuleHandler : DependenciesRuleHandlerBase
         public ProjectRuleHandler(
             IAggregateDependenciesSnapshotProvider aggregateSnapshotProvider,
             DependenciesSnapshotProvider snapshotProvider,
-            IUnconfiguredProjectCommonServices commonServices)
+            IUnconfiguredProjectCommonServices commonServices,
+            [Import(AllowDefault = true)] ISolutionService? solutionService)
```
davkean: Why is this optional?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ISolutionService.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright(c) Microsoft.All Rights Reserved.Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    [ProjectSystemContract(ProjectSystemContractScope.ProjectService, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.OneOrZero)]
```
davkean: Why is this optional?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
davkean: @drewnoakes Look at the caller of StartListeningAsync.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/ProjectRuleHandler.cs
drewnoakes: Because of the split between VS/non-VS layers. `ISolutionService` is defined in `MS.VS.ProjectSystem.Managed`. `SolutionService` is defined in `MS.VS.ProjectSystem.Managed.VS`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ISolutionService.cs
drewnoakes: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsSolutionEventListener.cs
drewnoakes: @davkean I've updated the PR, moving the service to global scope and initialising it in the package. Please re-review.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -52,6 +53,8 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
 
             var componentModel = (IComponentModel)(await GetServiceAsync(typeof(SComponentModel)));
             Lazy<DebugFrameworksDynamicMenuCommand> debugFrameworksCmd = componentModel.DefaultExportProvider.GetExport<DebugFrameworksDynamicMenuCommand>();
+            _solutionService = (SolutionService)componentModel.GetService<ISolutionService>();
```
davkean: Should be an async lookup.

davkean: My bad, I misread this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
```diff
@@ -0,0 +1,102 @@
...
+        private const int StateDisposed = 2;
+
+        private readonly IVsUIService<IVsSolution> _solution;
+        private readonly JoinableTaskContext _joinableTaskContext;
+
+        private int _state = StateUninitialized;
+        private uint _cookie = VSConstants.VSCOOKIE_NIL;
+        
+        /// <inheritdoc />
+        public bool IsSolutionClosing { get; private set; }
```
davkean: Nit, this is a property and hence under constructor.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
```diff
@@ -0,0 +1,102 @@
...
+        public int OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)                                    => HResult.OK;
+        public int OnQueryCloseProject(IVsHierarchy pHierarchy, int fRemoving, ref int pfCancel)              => HResult.OK;
+        public int OnBeforeCloseProject(IVsHierarchy pHierarchy, int fRemoved)                                => HResult.OK;
+        public int OnAfterLoadProject(IVsHierarchy pStubHierarchy, IVsHierarchy pRealHierarchy)               => HResult.OK;
+        public int OnQueryUnloadProject(IVsHierarchy pRealHierarchy, ref int pfCancel)                        => HResult.OK;
+        public int OnBeforeUnloadProject(IVsHierarchy pRealHierarchy, IVsHierarchy pStubHierarchy)            => HResult.OK;
+        public int OnQueryCloseSolution(object pUnkReserved, ref int pfCancel)                                => HResult.OK;
+        public int OnAfterCloseSolution(object pUnkReserved)                                                  => HResult.OK;
+
+        public int PrioritizedOnAfterOpenSolution(object pUnkReserved, int fNewSolution)                      => UpdateClosing(false);
```
davkean: I still think this should pick one Before and one After. Otherwise, someone will debug this and be confused why we're setting a property twice with the same value. I see no value in this other than debugging confusion.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -75,10 +78,15 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
...
             {
-                _projectSelectorService?.UnregisterProjectSelector(_projectSelectorCookie);
-                _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
+                if (_projectSelectorCookie != VSConstants.VSCOOKIE_NIL)
+                {
+                    _projectSelectorService?.UnregisterProjectSelector(_projectSelectorCookie);
+                    _projectSelectorCookie = VSConstants.VSCOOKIE_NIL;
+                }
+
+                _solutionService?.Dispose();
```
davkean: This is a MEF component, you don't own the lifetime of this, let MEF clean it up.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            IsSolutionClosing = isClosing;
+            return HResult.OK;
+        }
+
+        public void Dispose()
+        {
+            Assumes.True(_joinableTaskContext.IsOnMainThread, "Must be called on the UI thread.");
+
+            if (Interlocked.CompareExchange(ref _state, StateDisposed, StateListening) != StateListening)
```
davkean: Why not derive from OnceInitializedOnceDisposed?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        private HResult UpdateClosing(bool isClosing)
+        {
+            IsSolutionClosing = isClosing;
+            return HResult.OK;
+        }
+
+        public void Dispose()
+        {
+            Assumes.True(_joinableTaskContext.IsOnMainThread, "Must be called on the UI thread.");
```
davkean: As a heads up, IVsService<T>.Value also does this assertion - but up to you if you want to enforce again.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
drewnoakes: 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
drewnoakes: Trying that now.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
drewnoakes: Cheers, that's simpler.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
drewnoakes: Is that a standard everywhere? Personally I put stateful properties above the ctor (along with fields and other state), and computed properties after the ctor.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
drewnoakes: My goal was to set the flag as early as possible and increase confidence in correctness in cases I might not have tested for. I will add a comment to address any potential confusion.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
davkean: Yes.

PR #5620, opened 2019-11-02
add sourcelink support

### LINKED ISSUE #5619: Add SourceLink
Microsoft.VisualStudio.ProjectSystem.Managed.dll doesn't appear to be SourceLinked. Would be super nice to add SourceLink for easier debugging.

jmarolf:
fixes #5619


KirillOsenkov:
Curious how does this work if ExcludeAssets on the package is set to `all`? How do the targets get used during the build?

KirillOsenkov:
FYI I pulled this down and didn't see SourceLink targets files used in the build. That's why I was asking above ^^

PR #5622, opened 2019-11-04
Never forgive

drewnoakes:
Removes a bunch of null-forgiveness operators.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
@@ -140,7 +140,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
 
                 Assumes.Present(dte);
 
-                using var undo = UndoScope.Create(dte, renameOperationName);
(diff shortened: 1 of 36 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -138,9 +138,9 @@ private Task<bool> RenameAsync(string oldFilePath, string newFilePath, bool isCa
 
                 DTE? dte = _dte.Value;
 
-                Assumes.NotNull(dte);
+                Assumes.Present(dte);
 
-                using var undo = UndoScope.Create(dte, renameOperationName);
+                using var _ = UndoScope.Create(dte, renameOperationName);
```
dmonroym: why was this change made?

drewnoakes: The `undo` symbol is not used anywhere. My preference is to discard anything that's unused, so that I don't spend any time looking for usages of an unused symbol.

drewnoakes:
> I don't love it

Assuming you're referring to `Strings`, neither do I. It's the lesser of two evils.

PR #5625, opened 2019-11-06
Adding OutputGroupProvider in order to supply descriptions for certain .Net Core project output groups

MSLukeWest:
This description gets surfaced in the UI during Installer Project scenarios.  This description is meant to help guide users down the correct path when creating msi installers for .NET Core 3 projects when using the VS Installer Projects extension.

This implementation is based off the VsOutputGroupProvider in the CPS repo.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroup.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Build
+{
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Diagnostics;
```
davidwengier: We have usings outside of the namespace. I'll update our .editorconfig to express this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroup.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            Name = name;
+            TargetName = targetName;
+            DisplayName = displayName;
+            Description = description;
+            Outputs = items;
+            IsSuccessful = successful;
+        }
+
+        #region IOutputGroup Members
```
davidwengier: Remove regions :)

davidwengier:
@tmeschter This has missed the LOC cut off, correct?

davidwengier:
@MSLukeWest I'm not really familiar with Output Groups, where does this display in Visual Studio?

MSLukeWest:
> @MSLukeWest I'm not really familiar with Output Groups, where does this display in Visual Studio?

This gets displayed in the "Add Project Output Group" dialog in VS.  The only way to get to this dialog is to use installer projects (https://marketplace.visualstudio.com/items?itemName=visualstudioclient.MicrosoftVisualStudio2017InstallerProjects).

tmeschter:
@davidwengier Yes, this has missed the loc cutoff.
@MSLukeWest You can retarget this to master and we can discuss taking it for 16.5, but we can't take further string changes for 16.4. Even absent the string changes I don't think this would meet the bar.

MSLukeWest:
> @davidwengier Yes, this has missed the loc cutoff.
> @MSLukeWest You can retarget this to master and we can discuss taking it for 16.5, but we can't take further string changes for 16.4. Even absent the string changes I don't think this would meet the bar.

Understood, retargeted to Master for 16.5

MSLukeWest:
FYI @SvetBonev

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroup.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            Name = name;
+            TargetName = targetName;
+            DisplayName = displayName;
+            Description = description;
+            Outputs = items;
+            IsSuccessful = successful;
+        }
+
+        /// <inheritdoc/>
```
davkean: Remove these <inheritdoc/> they are not needed, VS will automatically show docs from the inteface for them.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Build
+{
+    [Export(typeof(IOutputGroupProvider))]
+    [AppliesTo(ProjectCapabilities.VisualStudioWellKnownOutputGroups)]
+    [Order(1000)]
```
davkean: Use Order.Default.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /// <summary>
+        /// Produces a set of output groups that is a subset of all well known output groups.
+        /// </summary>
+        /// <returns>Set of all known output groups.</returns>
+        private async Task<IImmutableSet<IOutputGroup>> GetOutputGroupMetadataAsync()
+        {
+            // Start with the comment set of output groups.
+            var result = s_outputGroups;
```
davkean: Please follow var rules, there will be messages in the ErrorList telling you what to change.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Initializes a new instance of the <see cref="OutputGroupProvider"/> class.
+        /// </summary>
+        /// <param name="projectAccessor">Imported <see cref="IProjectAccessor"/>.</param>
+        /// <param name="configuredProject">Imported <see cref="ConfiguredProject"/>.</param>
+        [ImportingConstructor]
+        private OutputGroupProvider(IProjectAccessor projectAccessor, ConfiguredProject configuredProject)
+        {
+            _projectAccessor = projectAccessor;
+            _configuredProject = configuredProject;
+            _outputGroups = new AsyncLazy<IImmutableSet<IOutputGroup>>(GetOutputGroupMetadataAsync);
```
davkean: This looks wrong; this will fix the result for the entire lifetime of the project even though targets can be removed and added. Is this what CPS does?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Build
+{
+    [Export(typeof(IOutputGroupProvider))]
+    [AppliesTo(ProjectCapabilities.VisualStudioWellKnownOutputGroups)]
+    [Order(1000)]
+    internal class OutputGroupProvider : IOutputGroupProvider
```
davkean: Rename this to "PublishItemsOutputGroupProvider" because that is all it does.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Creates a new output group with metadata only.
+        /// </summary>
+        /// <param name="name">Output group name.</param>
+        /// <param name="targetName">Output group target name.</param>
+        /// <param name="displayName">Output group display name.</param>
+        /// <param name="description">Optional output group description.</param>
+        /// <return>New <see cref="IOutputGroup"/>.</return>
+        private static IOutputGroup NewGroup(string name, string targetName, string displayName, string description)
+        {
+            return new OutputGroup(name, targetName, displayName, description, ImmutableList<KeyValuePair<string, IImmutableDictionary<string, string>>>.Empty, false);
```
davkean: What does false for IsSuccesful mean?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
MSLukeWest: Done.  I didn't see these originally because they're messages instead of errors/warnings like the rest of the style rules, and I apparently had those turned off.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
MSLukeWest: Yes, this is what CPS does.  This list of targets is used for filtering out items from s_outputGroups that may not be applicable to this project.  I don't think there's a scenario where the PublishItemsOutputGroup target would get added/removed after the project is loaded.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
MSLukeWest: In CPS it's used to indicate whether or not the output group built successfully.  For our purposes it's not used at all.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/OutputGroupProvider.cs
davkean: >  I don't think there's a scenario where the PublishItemsOutputGroup target would get added/removed after the project is loaded.

Given CPS already does this at the moment, I'm okay with this, but absolutely this could occur via a simple branch switch; project starts off as say a legacy project (that we've automatically opened in CPS - coming soon) and you switch to a branch where it is an SDK-based project. The project stays loaded and we apply the diff.

PR #5626, opened 2019-11-07
Add rule for usings outside of namespace

davidwengier:
We were missing a rule

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -206,6 +206,9 @@ csharp_preserve_single_line_blocks = true
 # Code block
 csharp_prefer_braces = false:none
 
+# Using statements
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -40,6 +40,7 @@ indent_size = 2
 # Sort using and Import directives with System.* appearing first
 dotnet_sort_system_directives_first = true
 dotnet_style_require_accessibility_modifiers = always:warning
+csharp_using_directive_placement = outside_namespace:error
```
davkean: Wrong section, should be under [*.cs]

davidwengier: lol, I had it there first, but then thought it made sense to be near the using statements. I'll move it back :)

PR #5627, opened 2019-11-07
Cleanup

drewnoakes:
Small bits pulled from other work.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowPaneProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowPaneProvider.cs
@@ -52,13 +52,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Logging
         private static IVsOutputWindowPane CreateProjectOutputWindowPane(IVsOutputWindow outputWindow)
         {
             Guid paneGuid = s_projectOutputWindowPaneGuid;
-            HResult hr = outputWindow.CreatePane(ref paneGuid, VSResources.OutputWindowPaneTitle, fInitVisible: 1, fClearWithSolution: 0);
-            if (hr.Failed)
-                throw hr.Exception;
 
-            hr = outputWindow.GetPane(ref paneGuid, out IVsOutputWindowPane pane);
-            if (hr.Failed)
-                throw hr.Exception;
+            Verify.HResult(outputWindow.CreatePane(ref paneGuid, VSResources.OutputWindowPaneTitle, fInitVisible: 1, fClearWithSolution: 0));
+
+            Verify.HResult(outputWindow.GetPane(ref paneGuid, out IVsOutputWindowPane pane));
 
             return pane;
         }
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/VisualBasic/MapDynamicEnumValuesProviderTests.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/VisualBasic/MapDynamicEnumValuesProviderTests.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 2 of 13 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ExportProjectNodeComServiceAttribute.cs
```diff
@@ -24,11 +24,6 @@ public string[] Iid
             get;
         }
 
-        public static bool IncludeInherited
```
davkean: This is required, it's part of the metadata of the contact. Without it, MEF won't find these types.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowPaneProvider.cs
```diff
@@ -52,13 +52,10 @@ public ProjectOutputWindowPaneProvider(IProjectThreadingService threadingService
...
         {
             Guid paneGuid = s_projectOutputWindowPaneGuid;
-            HResult hr = outputWindow.CreatePane(ref paneGuid, VSResources.OutputWindowPaneTitle, fInitVisible: 1, fClearWithSolution: 0);
-            if (hr.Failed)
-                throw hr.Exception;
 
-            hr = outputWindow.GetPane(ref paneGuid, out IVsOutputWindowPane pane);
-            if (hr.Failed)
-                throw hr.Exception;
+            Verify.HResult(outputWindow.CreatePane(ref paneGuid, VSResources.OutputWindowPaneTitle, fInitVisible: 1, fClearWithSolution: 0));
```
davkean: I used to deliberately use the previous pattern so that the exceptional path showed up in code coverage. Undecided about this...

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/VisualBasic/MapDynamicEnumValuesProviderTests.cs
```diff
@@ -122,7 +124,7 @@ private static IEnumerable<PageEnumValue> CreateEnumValueInstances(List<Tuple<st
             }
         }
 
-        private Dictionary<string, IEnumValue> CreateEnumValueMap(List<string> keys, IEnumerable<PageEnumValue> pageEnumValues)
+        private static Dictionary<string, IEnumValue> CreateEnumValueMap(List<string> keys, IEnumerable<PageEnumValue> pageEnumValues)
```
davkean: Roslyn clearly didn't find this one?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowPaneProvider.cs
drewnoakes: You could make the same argument about all other validation we have, such as `Requires.NotNull`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ExportProjectNodeComServiceAttribute.cs
drewnoakes: Reinstated.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/VisualBasic/MapDynamicEnumValuesProviderTests.cs
drewnoakes: I'll see if I can create a minimal repro.

drewnoakes: Does Roslyn actually have this feature? I cannot get it to suggest converting a method to static on even a simple case:

```c#
class Program
{
    static void Main()
    {
        new Program().M();
    }

    private void M()
    {
        N();
    }

    private void N()
    {
    }
}
```

davidwengier: It's part of roslyn-analyzers so you need that referenced. It definitely has it though, as I was getting it erroneously last night :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Logging/ProjectOutputWindowPaneProvider.cs
davkean: Requires is different, it's a contract break, it will never happen in a normal code path

I'm okay with this.

drewnoakes: True. I can see the desire to track places we're not explicitly testing what happens for erroneous HResults. It's one commit if we want to back it out in future.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/VisualBasic/MapDynamicEnumValuesProviderTests.cs
drewnoakes: Seems it's related to being called from an xUnit fact: https://github.com/dotnet/roslyn-analyzers/issues/3019

PR #5628, opened 2019-11-07
Simplify package initialization code

drewnoakes:
- Encapsulate F# project factory registration/deregistration
- Replace casts via `GetService<TService,TInterface>()` extension method
- Reformat code slightly
- Remove redundant default ctor


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
```diff
@@ -38,5 +41,22 @@ internal static void GetProjectFactoryGuid(XDocument doc, out Guid guidProjectFa
 
             guidProjectFactory = ProjectType.LegacyFSharpGuid;
         }
+
+        public static async Task<IDisposable> RegisterAsync(ManagedProjectSystemPackage package)
```
davkean: This doesn't need a package, just IAsyncServiceProvider.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
```diff
@@ -38,5 +41,22 @@ internal static void GetProjectFactoryGuid(XDocument doc, out Guid guidProjectFa
...
+        {
+            IVsRegisterProjectSelector projectSelector = await package.GetServiceAsync<SVsRegisterProjectTypes, IVsRegisterProjectSelector>();
+
```
davkean: This needs an explicit switch to the UI thread.

davkean:
I think we have almost hit the point where we just have a concept of startup services for the package.

drewnoakes:
> I think we have almost hit the point where we just have a concept of startup services for the package.

The latest update moves all component-specific initialization from `ManagedProjectSystemPackage` behind `IPackageService`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
drewnoakes: In the latest update I've documented on the interface that all calls to `IPackageService.InitializeAsync` occur on the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
drewnoakes: In the latest update it was necessary to pass the package so it could call back for `RegisterProjectFactory`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Xproj/XprojProjectFactory.cs
```diff
@@ -38,5 +45,15 @@ protected override object PreCreateForOuter(IntPtr outerProjectIUnknown)
             // Should not be called
             throw new NotImplementedException();
         }
+
+        /// <inheritdoc />
+        public Task<IDisposable?> InitializeAsync(ManagedProjectSystemPackage package, IComponentModel componentModel)
+        {
+            IVsProjectFactory factory = new XprojProjectFactory();
```
davidwengier: Can this be `this` rather than create a new one?

drewnoakes: Thanks, fixed. Missed this during conversion from static to instance.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
```diff
@@ -1,19 +1,40 @@
...
 {
+    [Export(typeof(IPackageService))]
     [Guid("E720DAD0-1854-47FC-93AF-E719B54B84E6")]
     [ProvideObject(typeof(FSharpProjectSelector), RegisterUsing = RegistrationMethod.CodeBase)]
-    internal sealed class FSharpProjectSelector : IVsProjectSelector
+    internal sealed class FSharpProjectSelector : IVsProjectSelector, IPackageService
     {
         private const string MSBuildXmlNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
 
+        /// <inheritdoc />
```
davkean: Please remove this, it's not needed, Roslyn will show docs from the interface.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/DebuggerTraceListener.cs
```diff
@@ -0,0 +1,57 @@
...
+            var type = Type.GetType($"Microsoft.VisualStudio.ProjectSystem.TraceUtilities, {assemblyName}");
+            Assumes.NotNull(type);
+
+            FieldInfo field = type.GetField("Source", BindingFlags.NonPublic | BindingFlags.Static);
+            Assumes.NotNull(field);
+
+            var source = (TraceSource)field.GetValue(null);
+
+            source.Switch.Level = SourceLevels.Warning;
+            source.Listeners.Add(new DebuggerTraceListener());
```
davidwengier: `this`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -33,12 +32,8 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
...
             
-            JoinableTaskContext context = componentModel.GetService<JoinableTaskContext>();
-
             IEnumerable<IPackageService> packageServices = componentModel.GetExtensions<IPackageService>();
 
-            await context.Factory.SwitchToMainThreadAsync(cancellationToken);
-
             foreach (IPackageService packageService in packageServices)
             {
                 await packageService.InitializeAsync(this);
```
davidwengier: Do we actually need to wait for all of these, as opposed to fire-and-forgetting them? Presumably the order that MEF finds them is indeterminant, so having dependencies between IPackageServices would be a bad design anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/DebuggerTraceListener.cs
drewnoakes: Something about lightning.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
davkean: Yes we do, the project selector and project factory _**must be**_ registered by the time the package has finished initializing; the solution has initialized the package synchronously on the UI thread when opening a project so that they register them, if they are not registered by the time the package has finished initializing, the projects (F#, XProj) will fail to load correctly.
There will be very little value kicking off the others asynchronously anyway - they do very little.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/DebuggerTraceListener.cs
davkean: ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
```diff
@@ -1,19 +1,47 @@
...
+
+        public async Task InitializeAsync(IAsyncServiceProvider asyncServiceProvider)
+        {
+            Assumes.Null(_projectSelector);
+
+            _projectSelector = await asyncServiceProvider.GetServiceAsync<SVsRegisterProjectTypes, IVsRegisterProjectSelector>();
+
+            await _context.Factory.SwitchToMainThreadAsync();
+
+            Guid selectorGuid = typeof(FSharpProjectSelector).GUID;
```
davkean: GetType().GUID.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IPackageService.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <remarks>
+    /// Implementations must be exported in global scope.
+    /// </remarks>
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ZeroOrMore)]
+    internal interface IPackageService
+    {
+        /// <summary>
+        /// Called when the package is initializing.
+        /// </summary>
+        /// <param name="asyncServiceProvider"></param>
```
davkean: Remove the empty param or doc it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
```diff
@@ -1,19 +1,47 @@
...
-    internal sealed class FSharpProjectSelector : IVsProjectSelector
+    internal sealed class FSharpProjectSelector : IVsProjectSelector, IPackageService, IDisposable
     {
         private const string MSBuildXmlNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
 
+        private readonly JoinableTaskContext _context;
+        private IVsRegisterProjectSelector? _projectSelector;
+        private uint _cookie = VSConstants.VSCOOKIE_NIL;
+
+        public FSharpProjectSelector(JoinableTaskContext context)
```
davkean: [ImportingCosntructor]

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
```diff
@@ -1,39 +1,39 @@
...
+        private readonly JoinableTaskContext _context;
+        private IVsSolution? _solution;
         private uint _cookie = VSConstants.VSCOOKIE_NIL;
         
-        /// <inheritdoc />
         public bool IsSolutionClosing { get; private set; }
 
-        [ImportingConstructor]
-        public SolutionService(IVsUIService<SVsSolution, IVsSolution> solution)
+        public SolutionService(JoinableTaskContext context)
```
davkean: [ImportingCosntructor]

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs
```diff
@@ -1,39 +1,39 @@
...
+    internal sealed class SolutionService : ISolutionService, IVsSolutionEvents, IVsPrioritizedSolutionEvents, IPackageService
     {
-        private readonly IVsUIService<IVsSolution> _solution;
-
+        private readonly JoinableTaskContext _context;
+        private IVsSolution? _solution;
         private uint _cookie = VSConstants.VSCOOKIE_NIL;
         
-        /// <inheritdoc />
         public bool IsSolutionClosing { get; private set; }
```
davkean: Props still end up after the constructor. :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetectorInitializer.cs
```diff
@@ -0,0 +1,32 @@
...
+    [Export(typeof(IPackageService))]
+    internal sealed class DotNetCoreProjectCompatibilityDetectorInitializer : IPackageService
+    {
+        public async Task InitializeAsync(IAsyncServiceProvider asyncServiceProvider)
+        {
+            IComponentModel componentModel = await asyncServiceProvider.GetServiceAsync<SComponentModel, IComponentModel>();
+
+            // Need to use the CPS export provider to get the dotnet compatibility detector
+            IDotNetCoreProjectCompatibilityDetector dotNetCoreCompatibilityDetector = componentModel
+                .GetService<IProjectServiceAccessor>()
```
davkean: Just import "IProjectServiceAccessor" in the constructor and you can avoid this dance.

davkean:
Please test F# SDK projects, XProj and the Debug default TFM selector after these changes to make sure they still work.\

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
drewnoakes: I don't think it'd be incorrect to `await WhenAll` them. I'll try it out.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IPackageService.cs
drewnoakes: Removed. The `<returns>` doc was wrong too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
drewnoakes: Shouldn't an analyzer flag this?

drewnoakes: There is an analyzer for this, but we've turned it off.

https://github.com/dotnet/project-system/blob/master/.editorconfig#L332

Turning it on flags a bunch of diagnostics. I'll take a look at this in a follow up. Some of the diagnostics look wrong. If so I'll get some bugs filed.

drewnoakes: https://github.com/dotnet/roslyn-analyzers/issues/3025

drewnoakes:
To test:

- [x] F# SDK projects
- [x] XProj
- [ ] Debug default TFM selector

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -30,14 +33,20 @@ internal sealed class ManagedProjectSystemPackage : AsyncPackage
...
             IEnumerable<IPackageService> packageServices = componentModel.GetExtensions<IPackageService>();
 
+            var sw = Stopwatch.StartNew();
+
             foreach (IPackageService packageService in packageServices)
             {
                 await packageService.InitializeAsync(this);
             }
+
+            Debug.WriteLine($"Initialized managed language project system package services in {sw.Elapsed.TotalMilliseconds:,0.##} ms");
```
swesonga: Would the TraceUtilities class be useful for logging this statistic on release builds?

drewnoakes: Possibly. I pushed a WIP commit to continue working on this at @davkean's office yesterday (where we all met) so this change will likely be removed. Unfortunately GitHub doesn't allow converting a PR to a draft (only the other way around).

drewnoakes:
This PR is now ready for final review.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FSharp/FSharpProjectSelector.cs
```diff
@@ -1,19 +1,49 @@
...
+        {
+            Requires.NotNull(context, nameof(context));
+
+            _context = context;
+        }
+
+        public async Task InitializeAsync(IAsyncServiceProvider asyncServiceProvider)
+        {
+            Assumes.Null(_projectSelector);
+            Assumes.True(_context.IsOnMainThread, "Must be on UI thread");
```
davkean: I'm missing where we guarantee this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -20,73 +16,32 @@ namespace Microsoft.VisualStudio.Packaging
...
-        {
-        }
-
         protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
 #pragma warning disable RS0030 // Do not used banned APIs
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-
-            _projectSelectorService = await this.GetServiceAsync<SVsRegisterProjectTypes, IVsRegisterProjectSelector>();
+            Assumes.True(ThreadHelper.JoinableTaskContext.IsOnMainThread, "Must be on UI thread.");
```
davkean: This is an incorrect assertion, see: http://index/?query=AsyncPackage.InitializeAsync&rightProject=Microsoft.VisualStudio.Shell.15.0&file=AsyncPackage.cs&line=69, in particular:

```
This method is invoked from a background thread.
```

davkean: See `LoadType::Asynchronous` in vspakmgr.cpp.

drewnoakes: I didn't see this on a background thread, but indeed the code suggests it might.

> a UI context rule

Could you elaborate on that?

davkean: Packages can be initialized for a variety of reasons, due to a synchronous load due a project type or something else referring to it, an explicit load by other code, or due to UI contexts. UI contexts are trigger points that force a package to be loaded, such as "Solution has projects", "Solution has C# projects", "Debugging", etc. We have a custom UI context which you can see applied to the package that says if there are any projects with CPS or .NET capabilities.

drewnoakes: Thanks, that's great to know.

Debugging today I noticed that the package init was happening much earlier than it was last week.

PR #5636, opened 2019-11-12
Changed table layout in BuildPropPage

melytc:
In the JP-OS, the 'Advanced' button in the Build tab was being cropped due to the last row not having AutoSize as its value.

**Currently on JP-OS:**

image

**Expected:**

image


davkean:
Nice!

PR #5638, opened 2019-11-12
Replace StringComparer/StringComparison usage

davkean:
This replaces all but 7 usages of StringComparer/StringComparison with explicit properties on StringComparers/StringComparisons. Banning of these APIs will be done in a follow when these last 7 items are fixed, they are a little more challenging than below.


src/Microsoft.VisualStudio.ProjectSystem.Managed/StringComparers.cs
```diff
@@ -13,16 +15,31 @@ namespace Microsoft.VisualStudio
...
-        public static IEqualityComparer<string> Paths => StringComparer.OrdinalIgnoreCase;
-        public static IEqualityComparer<string> PropertyNames => StringComparer.OrdinalIgnoreCase;
-        public static IEqualityComparer<string> PropertyLiteralValues => StringComparer.OrdinalIgnoreCase;
-        public static IEqualityComparer<string> PropertyValues => StringComparer.Ordinal;
-        public static IEqualityComparer<string> RuleNames => StringComparer.OrdinalIgnoreCase;
-        public static IEqualityComparer<string> ConfigurationDimensionNames => StringComparer.Ordinal;
-        public static IEqualityComparer<string> DependencyProviderTypes => StringComparer.OrdinalIgnoreCase;
-        public static IEqualityComparer<string> ItemNames => StringComparer.OrdinalIgnoreCase;
-        public static IEqualityComparer<string> ItemTypes => StringComparer.OrdinalIgnoreCase;
+        public static StringComparer WorkspaceProjectContextIds => StringComparer.Ordinal;
```
davkean: These need to be typed as "StringComparer" as we use both `IEqualityComparer<string>` and `IComparer<string>`.

PR #5639, opened 2019-11-12
Display the project's imports as a tree

### LINKED ISSUE #4805: Show imported projects in tree
SDKs and user projects can `<Import>` further project files. These create a hierarchy of imports.

It would be useful to be able to view this in the tree under the Dependencies node.

This should probably only be visible when Show All Files is selected.

drewnoakes:
Fixes #4805.

This PR adds an "Imports" sub-tree to projects that shows all project imports and their structure:

image

Due to the size of this sub-tree, it is only displayed when the per-project 'show all files' feature is enabled. When disabled, the tree is removed from memory and its data subscription disposed.

Double-clicking (or pressing <kbd>Enter</kbd>) on these imports opens them in the editor.

Import tree

---

Capturing ideas for future work here (unlikely to do all these before merge so will move pending items to another issue at that time):

- [x] Open non-editable files as read only
- [x] Indicate editable/non-editable files
- [x] Preview/provisional tab support
- [ ] Option to show these at all times
- [ ] Browse object showing path, label, group label
- [ ] Investigate including imports in Find in Files
- [ ] Sync with Active Document


onovotny:
I love this, but I would like to have the option to have this node always available, not just with "Show All Files" enabled. Any chance to have an option for it?

panopticoncentral:
It does seem odd that it would be under "show all files" since that's a separate thing. Would it make sense to have this under the Dependencies node, since that conceptually fits?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,288 @@
...
+    /// <para>
+    /// This feature can be enabled and disabled via the <c>ProjectImportsTree</c> project capability.
+    /// </para>
+    /// </remarks>
+    [Export(ExportContractNames.ProjectTreeProviders.PhysicalViewRootGraft, typeof(IProjectTreeProvider))]
+    [AppliesTo(Capability)]
+    internal sealed class ProjectImportsSubTreeProvider : ProjectTreeProviderBase, IProjectTreeProvider, IShowAllFilesProjectTreeProvider
+    {
+        public const string Capability = "ProjectImportsTree";
+        private const string RootNodeCaption = "Imports";
```
davidwengier: Should this be in a resx for localization?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,288 @@
...
+                                    IProjectTree updatedTree = SyncNode(
+                                        imports: e.Value.Value,
+                                        tree: treeSnapshot.Value.Tree);
+
+                                    return Task.FromResult(new TreeUpdateResult(updatedTree, e.DataSourceVersions));
+                                });
+
+                            return;
+
+                            IProjectTree SyncNode(IReadOnlyList<IProjectImportSnapshot> imports, IProjectTree tree)
```
davidwengier: 4th level local function, a new record! 🎉 🏅

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,288 @@
...
+                }
+            }
+        }
+
+        // CPS's physical tree provider currently only uses the IShowAllFilesProjectTreeProvider.ShowAllFiles property
+        // for graft root providers. These other interface methods have safe no-op implementations in case that changes
+        // one day.
+
+        bool IShowAllFilesProjectTreeProvider.CanIncludeItems(IImmutableSet<IProjectTree> nodes)   => true;
+        bool IShowAllFilesProjectTreeProvider.CanExcludeItems(IImmutableSet<IProjectTree> nodes)   => true;
```
davidwengier: These two sound like they control menu items ("Exclude From Project"/"Include In Project") so `false` might be a safer option.

davkean:
Looks like preview isn't working?

davkean:
I think we should some sort of indication of user-editable files vs non-user editable files.

drewnoakes:
> I would like to have the option to have this node always available

Definitely doable. Consequences of this are:

1. Extra resources (memory/CPU) are required to populate and display the trees. Each project's import tree can be quite large (see below), and if you have hundreds of projects in your solution that memory adds up. The trees are unique per-project too, so we cannot easily share resources between projects.
2. Items will always appear in Solution Explorer search

> Would it make sense to have this under the Dependencies node?

Possibly. There are pros and cons. What do people think?

---

Here's an example of the import tree for a `netcoreapp3.1` default console app project, fully-expanded:

image

onovotny:
@drewnoakes I'm okay with the extra CPU cost, at least as an option. Doesn't have to be enabled by default.

I also do want things to appear in search.

khellang:
> What do people think?

Can building the trees be done lazily? I.e. always display `Imports` under the `Dependencies` node, but don't resolve a tree until top-level expansion is requested?

drewnoakes:
@onovotny can you explain why you would like to see this all the time? The CPU cost is less of a concern to me than the memory cost.

@khellang yes the tree can be built lazily but we still need to hold onto the list of imports at evaluation time.

I'll run some perf tests over the next while to get a sense of the impact here.

onovotny:
I want to easily reference the full build chain, especially when combined with @mhutch’s intellisense extension. 

What I wind up doing now is a “find in folder” of the SDK directory.

poke:
This is awesome! I'm really looking forward to this!

I would also like to see some of this permanently. In my case, I wouldn't really be interested in the full tree though. But what I would like to see are the first-level items for the project, to see what directly affects the project.

In that sense, I wouldn't want to see the full SDK tree by default but rather just a single item for the SDK, and another item for the Directory.Build and that custom import in my project etc.; just as a quick way to see what is affecting the project before I can then drill down into the details (then requiring the full tree).

drewnoakes:
@onovotny thanks that's helpful.

@poke I'm looking into differentiating between implicit (brought in via SDK/packages/etc) and explicit (`Directory.Build.*`/`<Import>`/etc) imports. The former should be opened as read only, for example. I'm considering making 'Show all Files' add implicit entries. This idea could one day also extend to the dependencies node, cutting down the number/depth of items there too.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
drewnoakes: Yes, thanks, fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
drewnoakes: In a property setter, no less.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// </para>
+    /// <para>
+    /// This feature can be enabled and disabled via the <c>ProjectImportsTree</c> project capability.
+    /// </para>
+    /// </remarks>
+    [Export(ExportContractNames.ProjectTreeProviders.PhysicalViewRootGraft, typeof(IProjectTreeProvider))]
+    [AppliesTo(Capability)]
+    internal sealed class ProjectImportsSubTreeProvider : ProjectTreeProviderBase, IProjectTreeProvider, IShowAllFilesProjectTreeProvider
+    {
+        public const string Capability = "ProjectImportsTree";
```
davkean: Please move to Capability.cs.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+        private static readonly ProjectImageMoniker s_nodeIcon = KnownMonikers.TargetFile.ToProjectSystemType();
+
+        private static ProjectTreeFlags ProjectImportsTreeRootFlags { get; } = ProjectTreeFlags.Create(
+            ProjectTreeFlags.Common.BubbleUp |              // sort to top of tree, not alphabetically
+            ProjectTreeFlags.Common.VirtualFolder |
+            ProjectTreeFlags.Common.DisableAddItemFolder);
+        
+        private static ProjectTreeFlags ProjectImportFlags { get; } = ProjectTreeFlags.Create(
+            ProjectTreeFlags.Common.SourceFile |
+            ProjectTreeFlags.Common.NonMemberItem);         // enable double-click/enter to edit
```
davkean: This comment appears wrong, it's SourceFile that enables open.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+                lock (SyncObject)
+                {
+                    if (_showAllFiles == value)
+                    {
+                        return;
+                    }
+
+                    _showAllFiles = value;
+                }
+
```
davkean: I dislike this pattern of putting these methods in a setter. Can we not do this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+                        ITargetBlock<IProjectVersionedValue<IProjectImportTreeSnapshot>> actionBlock
+                            = DataflowBlockSlim.CreateActionBlock(
+                                new Action<IProjectVersionedValue<IProjectImportTreeSnapshot>>(SyncTree));
+
+                        IReceivableSourceBlock<IProjectVersionedValue<IProjectImportTreeSnapshot>> importTreeSource
+                            = _projectSubscriptionService.ImportTreeSource.SourceBlock;
+
+                        using (TrySuppressExecutionContextFlow())
+                        {
+                            _subscription = importTreeSource.LinkTo(actionBlock, DataflowOption.PropagateCompletion);
```
davkean: LinkToAction.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+
+                        // Subscribe to data to populate the tree and keep it updated with changes
+                        ITargetBlock<IProjectVersionedValue<IProjectImportTreeSnapshot>> actionBlock
+                            = DataflowBlockSlim.CreateActionBlock(
+                                new Action<IProjectVersionedValue<IProjectImportTreeSnapshot>>(SyncTree));
+
+                        IReceivableSourceBlock<IProjectVersionedValue<IProjectImportTreeSnapshot>> importTreeSource
+                            = _projectSubscriptionService.ImportTreeSource.SourceBlock;
+
+                        using (TrySuppressExecutionContextFlow())
```
davkean: @lifengl Do we need to suppress execution flow anymore?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+                                        tree: treeSnapshot.Value.Tree);
+
+                                    return Task.FromResult(new TreeUpdateResult(updatedTree, e.DataSourceVersions));
+                                });
+
+                            return;
+
+                            IProjectTree SyncNode(IReadOnlyList<IProjectImportSnapshot> imports, IProjectTree tree)
+                            {
+                                var knownImportPaths = new HashSet<string?>(StringComparers.Paths);
```
davkean: If imports can be null, then we'll display an empty node in the tree.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,280 @@
...
+            if (disposing)
+            {
+                _subscription?.Dispose();
+            }
+
+            base.Dispose(disposing);
+        }
+
+        [Export]
+        [AppliesTo(Capability)]
```
davkean: There's a race here that can cause this AppliesTo to be out of sync with the AppliesTo on the provider. Underneath this capabiliity will be respected by the GetExportedValue call, which will respect whatever the current set of capabilities for the project are, which may or may not have changed from the set of capabilities that caused this tree to be created, causing it to return null for this component. I would just remove the capability - if the tree exists, so should this export.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
drewnoakes: If you don't include `NonMemberItem` it throws when you double-click.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
drewnoakes: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
davkean: I think you'll find the command doesn't even work without SourceFile: http://index/?query=DefaultVsUIHierarchyWindowCmdsHandler&rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Package%5CCommands%5CDefaultVsUIHierarchyWindowCmdsHandler.cs&line=80

drewnoakes: I copied the comment to both flags.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
drewnoakes: True. We are setting the path with a non-null value, but the getter is marked as nullable. I've added an `Assumes.NotNullOrEmpty` which satisfies NRT checking.

MeikTranel:
@drewnoakes, RE:
> @khellang yes the tree can be built lazily but we still need to hold onto the list of imports at evaluation time.

How would that work? When evaluating whether to import a project file at the bottom (so to speak) we first have to know property values declared on the nth level of imports before that. Many import statement in the .net sdk .targets files for example are gated behind properties set deep in the dependency tree of respective .props files at the top. You have to evaluate all project files appearing before the import statement, to know whether or not that import is supposed to go through. And at that point you already know about all imports before that... that doesn't seem really "lazy" to me.

In regards to memory (forgive me if i'm missing something really big) shouldn't the memory usage of this tree be mostly linear to the amount of project files imported, wherein each would just be representation of a filepath and some metadata like a readonly boolean and reference list of all children? And while of course the .net sdk has lots imports... we're still talking about like a 100 filepath strings++. If i have a solution with hundreds of projects that might run into issues but with that large of a solution this is to be expected.

The most desirable behavior would be an option akin to the "Track active file in solution explorer" option, where i could toggle this on or off. I'd be a little disappointed if i had to load all the bin/obj files into the solution explorer because i wanted to see imports.

This is a great feature by the way. I have been struggling lately to keep coworkers up to snuff about how all the intricate Directory.Build.props stuff comes to fruition. 👍

davkean:
@MeikTranel Drew is referring to the tree itself, not the target files themselves. We have to worry about large solutions, it's exactly these solutions that push us close to, or over the memory limit of a large address aware 32-bit process. It's not just the overhead of the nodes itself, but also the snapshots of "import data" that feed into the tree - that might otherwise not be produced without this tree.

MeikTranel:
I can understand that. In the end you folks are of course the people with the telemetry data to answer these kinds of design decisions. Going back to my suggestion: Would it be acceptable to offer an option that drives this behavior? We have other options labeled with a warning that they can have issues with large workspaces in the VS Settings. Both Explorer View and a toggle to enable this in the regular solution explorer should be problematic in the same sense in situations like that.

Maybe we should track and discuss this in a separate issue 😄  I see no reason why that decision should hold up this PR (as it stands this PR greatly improves the product either way).

drewnoakes:
@MeikTranel we're definitely considering having an option to show these at all times. It's captured in the checklist in the PR description as "Option to show these at all times".

Nirmal4G:
Enable these features via project tools extension. Then put a Button on solution explorer, say enable msbuild/project file debugging... Or something similar.

Don't add as part of normal project experience.

drewnoakes:
@Nirmal4G doesn't having it behind 'Show All Files' achieve what you're suggesting?

Nirmal4G:
@drewnoakes It changes the meaning of `Show All Files`.

Besides, these tools are only for props/targets/SDK authoring and debugging custom project files when something goes wrong, etc...

They are not needed for regular Dev User scenarios. So, my recommendation is that they should be behind `Project File Tools` extension and also with Separate button (`Enable Project File Debugging` or something similar) on the Solution Explorer.

Nirmal4G:
Also, instead of doing only imports, can we implement full binary log support for project file debuggability and the like...?

I would definitely like that. I would also want first class SDK authoring and packaging support too.

poke:
> They are not needed for regular Dev User scenarios.

I think that is very subjective. I'm personally advocating our teams to make more use of MSBuild capabilities. At that point, non-trivial projects have a lot to gain from custom project extensions. And if you don't want to hide this complexity completely (which you shouldn't), then making it more visual makes a lot of sense.

Nirmal4G:
What I meant is put it under a switch. We should use MSBuild capabilities more. Not at the expense of confusing normal users.

The main reason I said what I said is that there should be context distinction, we switch from using project to modifying and/or customizing build process.

You should only need to do these things few times over the lifetime of the project.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/ProjectImportsSubTreeProvider.cs
drewnoakes: These are specific to "Show All Files".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/ProjectImports/ImportTreeCommandGroupHandler.cs
```diff
@@ -0,0 +1,244 @@
...
+            _serviceProvider = serviceProvider;
+            _configuredProject = configuredProject;
+            _uiShellOpenDocument = uiShellOpenDocument;
+            _externalFilesManager = externalFilesManager;
+            _oleServiceProvider = oleServiceProvider;
+        }
+
+        protected abstract bool IsOpenCommand(long commandId);
+
+        protected abstract bool IsOpenWithCommand(long commandId);
```
davidwengier: Subjective, but it took me a while to understand the API here. IsOpenCommand is used to determine if the command should be handled, and IsOpenWith is only used for picking a view. it would be more logical to me if `IsOpenCommand` was renamed to `CanHandleCommand` and `IsOpenWithCommand` was changed to `GetLogicalViewForCommand(commandId`)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/ProjectImports/ImportTreeCommandGroupHandler.cs
```diff
@@ -0,0 +1,244 @@
...
+                                out windowFrame));
+
+                        RunningDocumentInfo rdtInfo = rdt.GetDocumentInfo(item.FilePath);
+
+                        // Set it as read only if necessary.
+                        bool isReadOnly = item.Flags.Contains(ProjectImportsSubTreeProvider.ProjectImportImplicit);
+
+                        if (isReadOnly)
+                        {
+                            if (rdtInfo.DocData is IVsTextBuffer textBuffer)
```
davidwengier: Nit: combine the two if statements

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/ProjectImports/ImportTreeCommandGroupHandler.cs
```diff
@@ -0,0 +1,244 @@
...
+                {
+                    throw new AggregateException(exceptions);
+                }
+            }
+        }
+
+        [ExportCommandGroup(CMDSETID.UIHierarchyWindowCommandSet_string)]
+        [AppliesTo(ProjectCapability.ProjectImportsTree)]
+        [Order(ProjectSystem.Order.BeforeDefault)]
+        private sealed class UIHierarchyWindowCommandSetGroupHandler : ProjectImportsCommandGroupHandlerBase
```
davidwengier: Nit: This and the below to separate files

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,330 @@
...
+
+                        ITargetBlock<IProjectVersionedValue<ValueTuple<IProjectImportTreeSnapshot, IProjectSubscriptionUpdate>>> actionBlock =
+                            DataflowBlockSlim.CreateActionBlock<IProjectVersionedValue<ValueTuple<IProjectImportTreeSnapshot, IProjectSubscriptionUpdate>>>(
+                                SyncTree,
+                                new ExecutionDataflowBlockOptions
+                                {
+                                    NameFormat = "Import Tree Action: {1}"
+                                });
+
+                        using (TrySuppressExecutionContextFlow())
```
davidwengier: I don't know what execution context flow is, or why it should be suppressed. Is there doco for this you know of to point me to?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.cs
```diff
@@ -0,0 +1,330 @@
...
+                            });
+
+                        // Subscribe to data to populate the tree and keep it updated with changes
+                        IReceivableSourceBlock<IProjectVersionedValue<IProjectImportTreeSnapshot>> importTreeSource
+                            = _projectSubscriptionService.ImportTreeSource.SourceBlock;
+
+                        IReceivableSourceBlock<IProjectVersionedValue<IProjectSubscriptionUpdate>> projectRuleSource
+                            = _projectSubscriptionService.ProjectRuleSource.SourceBlock;
+
+                        var intermediateBlock =
```
davidwengier: What does this block actually do?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.ImplicitProjectCheck.cs
```diff
@@ -0,0 +1,36 @@
...
+            private readonly string? _programFiles64;
+            private readonly string _windows;
+
+            public string? ProjectExtensionsPath { get; set; }
+
+            public ImplicitProjectCheck()
+            {
+                _windows        = Environment.GetFolderPath(Environment.SpecialFolder.Windows);
+                _programFiles86 = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86);
+
```
davidwengier: Does there need to be an entry here for any paths coming from the Visual Studio installation?

(i'll try to get this branch running locally so I'll be able to answer this myself :) )

davidwengier: Yes, we do. Microsoft.Common.props and others all show up as editable for me, because it comes from E:\VSInstalls\Master\MSBuild\Current\Microsoft.Common.props on my machine.
image

theo-albers:
Wouldn't it be nice to open full details of the project in a new document tab? So instead of using the logical place, the solution explorer, open a new tab. That opens a whole new opportunity of visualizing the project structure, variables, conditions, etc.
Since it will be a snapshot view, you can search efficiently. It's also fine to have a brief delay when you initiate the view from the solution explorer.

MeikTranel:
I mean there's opportunities and more opportunities here... but such a thing would definitely better fit in an extension. But that's beside the point. We're in a pull request here scope creeping the poor dude in real time. Let's move this into issues. I dont think all these comments are helping actual reviewers...  <sup><sup><sup>sorry 😄</sup></sup></sup>

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.cs
drewnoakes: This pattern originated in CPS's capabilities tree provider.

[Above](https://github.com/dotnet/project-system/pull/5639#discussion_r345527552), @davkean asked @lifengl whether this is still needed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.cs
drewnoakes: You need an intermediate block to allow filtering to specific rule names via `LinkTo` below.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/ProjectImports/ImportTreeCommandGroupHandler.cs
drewnoakes: Future extensions can add more `Is*Command` methods to be called from chained `if` blocks. As for `GetLogicalViewForCommand` I prefer to keep the logic internal, and using a `bool` keeps this abstraction tighter than a `Guid`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/ProjectImports/ImportTreeCommandGroupHandler.cs
drewnoakes: These classes are so small that I prefer to leave them here. It's easier to compare them that way.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.ImplicitProjectCheck.cs
drewnoakes: Thank you, I'll be sure to handle this. The `VSAPPIDDIR` environment variable might do the job.

drewnoakes:
Turns out we can get the "show all files" value via dataflow. I am considering removing the use of `IShowAllFilesProjectTreeProvider` (which would mean the CPS change could be reverted if we wanted to do so, @lifengl).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ProjectImportsSubTreeProvider.ImplicitProjectCheck.cs
drewnoakes: Fixed.

image

drewnoakes:
I'm planning on merging this today and moving pending items to an issue.

PR #5642, opened 2019-11-13
Reuse TPL boolean task results

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskResult.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.VisualStudio.Threading
     internal static class TaskResult
```
sharwell: 💭 Why not just remove this class?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskResult.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.VisualStudio.Threading
...
     {
         /// <summary>
-        ///     Represents a Task that's completed successfully with the result of <see langword="false"/>.
+        ///     Represents a <see cref="Task{TResult}"/> that's completed successfully with the result of <see langword="false"/>.
         /// </summary>
-        public static readonly Task<bool> False = Task.FromResult(false);
+        public static readonly Task<bool> False = TplExtensions.FalseTask;
```
sharwell:
```suggestion
        public static Task<bool> False => TplExtensions.FalseTask;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskResult.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.VisualStudio.Threading
...
 
         /// <summary>
-        ///     Represents a Task that's completed successfully with the result of <see langword="true"/>.
+        ///     Represents a <see cref="Task{TResult}"/> that's completed successfully with the result of <see langword="true"/>.
         /// </summary>
-        public static readonly Task<bool> True = Task.FromResult(true);
+        public static readonly Task<bool> True = TplExtensions.TrueTask;
```
sharwell:
```suggestion
        public static Task<bool> True => TplExtensions.TrueTask;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskResult.cs
drewnoakes: It crossed my mind but `TplExtensions` is a mouthful.

sharwell: 🤪

Issue #4215, opened 2018-11-02
Show our project file editor in the right hand side of diff view


davidwengier:
Currently when diffing a project file the right hand side is a plain text buffer (readonly, no colorization). We should make it use the our editor for a better experience.

Because the project file is always open this line is presumably a problem: 
http://index/?query=DocumentOpeningService&rightProject=Microsoft.VisualStudio.Editor.Implementation&file=DocumentOpeningService.cs&line=65

We need to ensure that the buffer passed to GetDocumentBuffer is the IVsTextLines returned from here:
http://index/?query=%60IVsTextBufferProvider&rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Editor%5CVsTextBufferProviderAggregator.cs&line=67

Also double check that legacy isn't broken because of [this](http://index/?query=%60IVsTextBufferProvider&rightProject=Microsoft.VisualStudio.AppDesigner&file=PropPageDesigner%5CPropPageDesignerDocData.vb&line=275) and that opening the diff while the app designer is open doesn't try to reuse that buffer.

davkean:
Note this was also a feedback item: https://developercommunity.visualstudio.com/content/problem/238045/editing-a-csproj-file-through-the-integrated-diffi.html.

davidwengier:
More feedback: https://developercommunity.visualstudio.com/idea/482797/project-files-cant-be-edited-in-diff-mode.html

davidwengier:
Internal PR and this repos PR have been merged.

PR #5646, opened 2019-11-18
Add necessary registrations for project file editing

davidwengier:
Part of fixing #4215, this adds the registry keys necessary to tell the diff window what to do with our project files. This is responding to the removal of the temporary registry keys in https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/214544. Since the keys deal with specific .NET project file extensions they belong in this repo.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -25,6 +25,14 @@ namespace Microsoft.VisualStudio.Packaging
         new string[] { "dotnetcore" },
         new string[] { "SolutionHasProjectCapability:.NET & CPS" }
         )]
+    // Tell the diff window that these files should use the XML editor
+    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".csproj")]
+    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".vbproj")]
+    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".fsproj")]
+    // Tell the diff window that these files are supported and should be editable. The actual content type doesn't matter for this
+    [ProvideDiffSupportedContentType(".csproj", "")]
+    [ProvideDiffSupportedContentType(".vbproj", "")]
+    [ProvideDiffSupportedContentType(".fsproj", "")]
 
     [ProvideMenuResource("Menus.ctmenu", 4)]
     internal partial class ManagedProjectSystemPackage : AsyncPackage
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
@@ -25,14 +25,6 @@ namespace Microsoft.VisualStudio.Packaging
         new string[] { "dotnetcore" },
         new string[] { "SolutionHasProjectCapability:.NET & CPS" }
         )]
-    // Tell the diff window that these files should use the XML editor
-    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".csproj")]
-    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".vbproj")]
-    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".fsproj")]
-    // Tell the diff window that these files are supported and should be editable. The actual content type doesn't matter for this
-    [ProvideDiffSupportedContentType(".csproj", "")]
-    [ProvideDiffSupportedContentType(".vbproj", "")]
-    [ProvideDiffSupportedContentType(".fsproj", "")]
 
     [ProvideMenuResource("Menus.ctmenu", 4)]
     internal partial class ManagedProjectSystemPackage : AsyncPackage
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProvideDiffSupportedContentTypeAttribute.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProvideDiffSupportedContentTypeAttribute.cs
@@ -26,7 +26,6 @@ namespace Microsoft.VisualStudio
             key.SetValue(_extension, _contentType);
         }
 
-        //
         public override void Unregister(RegistrationContext context)
         {
             context.RemoveValue(RegistryKey, _extension);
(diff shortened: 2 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProvideDiffSupportedContentTypeAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        public override void Register(RegistrationContext context)
+        {
+            using Key key = context.CreateKey(RegistryKey);
+
+            key.SetValue(_extension, _contentType);
+        }
+
+        //
```
dmonroym: nit: remove //

davidwengier: Thank you! And thanks for the reminder to log a bug with roslyn for why it inserts them when looking at a decompilation :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -25,6 +25,14 @@ namespace Microsoft.VisualStudio.Packaging
         new string[] { "dotnetcore" },
         new string[] { "SolutionHasProjectCapability:.NET & CPS" }
         )]
+    // Tell the diff window that these files should use the XML editor
+    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".csproj")]
+    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".vbproj")]
+    [ProvideLanguageExtension("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".fsproj")]
+    // Tell the diff window that these files are supported and should be editable. The actual content type doesn't matter for this
+    [ProvideDiffSupportedContentType(".csproj", "")]
```
davkean: These don't appear to be package-specific yet they are applied to the package. I think these should actually be assembly-based attributes.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
```diff
@@ -25,6 +25,14 @@ namespace Microsoft.VisualStudio.Packaging
...
+    [ProvideDiffSupportedContentType(".vbproj", "")]
+    [ProvideDiffSupportedContentType(".fsproj", "")]
```
davkean: What does empty mean here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProvideDiffSupportedContentTypeAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    internal class ProvideDiffSupportedContentTypeAttribute : RegistrationAttribute
+    {
+        private const string RegistryKey = @"Diff\SupportedContentTypes";
+
+        private readonly string _contentType;
+        private readonly string _extension;
+
+        public ProvideDiffSupportedContentTypeAttribute(string extension, string contentType)
```
davkean: ContextType is never used, just remove it and name the attribute after what "empty" means.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
davkean: I see the comment now, just rename the attribute to make this clear and then you don't need a comment.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
davkean: I would consider moving all these attrbutes to ProjectTypeRegistration, if we were to add a new project type registration, we'd want these to be kept in sync.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProvideDiffSupportedContentTypeAttribute.cs
davidwengier: ContentType is used, its the value of the registry key

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
davidwengier: Perhaps I'm misunderstanding what you mean, but I don't know what you would rename the attribute to, it follows the pattern of the other Provide* attributes. If I remove the comment, then 6 months from now someone will be wondering what an empty string means and will have to go digging through code. There are many times where I wish there would have been more explanatory comments in our repo.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
davidwengier: I just kept all of the registrations together ¯\\\_(ツ)_/¯

I'm guessing that if they're at the assembly level they're applied whether our package is loaded or not, right? In this case certainly we just want registry entries to exist so it probably doesn't matter, though I don't really know what it means to be "package specific" either. Is it just the difference between attributes that need some code in `InitializeAsync`?

davidwengier: `ProvideLanguageExtension` can't be applied to assemblies, should I leave them, or write my own version that calls the one in shell?

davkean: The object you are applying the attribute provides the "component" - indicated by http://index/?query=ProvideLanguageExtension%20&rightProject=Microsoft.VisualStudio.Shell.Framework&file=RegistrationAttribute.cs&line=151, usually this results in a `Package` name/value to be added to the key.  ProvideLanguageExtension is actually defined incorrectly - it doesn't use it at all, and therefore should be an assembly-based attribute.  Applying these to the package continues an falsehood that these need to be applied to the package and then this continues to get copied around the tree. All the attributes are package-based attributes.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs
davkean: The comment exists because the attribute name isn't descriptive enough`. ProvideDiffSupportedExtension` with no content type would remove the need for the comment and wouldn't have me questioning why we're passing empty string here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProvideDiffSupportedContentTypeAttribute.cs
davkean: It's always empty, this doesn't need to be a parameter.

Issue #5650, opened 2019-11-18
COMException when disposing SolutionService


davidwengier:
`SolutionService.Dispose()` retrieves the `Value` property of an `IVsUIService` but it doesn't do so from the UI thread so we get a `COMException` when closing the solution. This presumably leads to us leaking solution events. This should only happen when VS is shutting down so its probably not going to have any real world side effects, but worth fixing.

See here:
https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/SolutionService.cs#L82

Which hits this:
![image](https://user-images.githubusercontent.com/754264/69096998-b6274880-0aa9-11ea-999d-0907c189d56c.png)

PR #5651, opened 2019-11-19
Fix exception on shutdown

davkean:
Fixes: https://github.com/dotnet/project-system/issues/5650

OncenInitializedOnceAsync pushes onto a background thread by default, change the approach to just derive from OnceInitializedOnceDisposedAsync and explicitly switch to UI thread.


drewnoakes:
This change in undone in #5628 (which is now ready for final review) as it uses a different approach that results in dispose being called on the main thread.

PR #5652, opened 2019-11-19
Remove unneeded inheritdoc

davkean:
VS will automatically show documentation from interface and base declarations.


davkean:
This removes the `<inheritdoc />` elements from the XML files, csc doesn't actually copy the documentation from the other elements into the XML file.

PR #5655, opened 2019-11-21
Cleanup analysis

davkean:


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -439,7 +439,7 @@ private async Task OnDependenciesSnapshotChangedAsync(SnapshotChangedEventArgs e
 
                 if (!cancellationToken.IsCancellationRequested)
                 {
-                    dependenciesNode = await viewProvider.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
+                    dependenciesNode = await viewProvider!.BuildTreeAsync(dependenciesNode, snapshot, cancellationToken);
```
davkean: https://github.com/dotnet/roslyn/issues/38548

drewnoakes: This went away, then came back. It only seems to happen in the IDE, not as a build warning too. Hence I've been ignoring it.

davkean: Build is a fixed version, IDE uses latest version,.

drewnoakes: So yeah, it's a regression in Roslyn.

PR #5662, opened 2019-11-26
Enforce modifier order & static local functions

davkean:


davkean:
There's no violations of this across the tree.

PR #5664, opened 2019-11-27
Move fields/properties/events/nested types

davkean:
Make DebugPageViewModel follow same standards as rest of tree.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -507,6 +444,209 @@ public ObservableList<NameValuePair> EnvironmentVariables
...
+        }
+
+        public bool NewProfileEnabled => LaunchProfiles != null;
+
+        public ICommand NewProfileCommand
+        {
+            get
+            {
+                return LazyInitializer.EnsureInitialized(ref _newProfileCommand, () =>
+                 new DelegateCommand(state =>
```
drewnoakes: Indentation looks funny here (it was like this beforehand though).

davkean: Yeah might fix it latter.

PR #5672, opened 2019-11-28
Fix incorrect symbol removal logic

swesonga:
Fixes https://developercommunity.visualstudio.com/content/problem/812018/vs2019-preview-4-broke-conditional-compilation-sym.html caused by https://github.com/dotnet/project-system/commit/cb72e7b0


davidwengier:
I don't see any reason to hold off merging this, its the resulting insertion PR you need approval for, and you'll need to log an Azure DevOps work item for that, and fill out the Ask Mode template.

drewnoakes:
> its the resulting insertion PR you need approval for

Agreed, but if that's not approved then we need an admin for force-push this branch back to the prior commit. It could save time to verify that this meets the bar in principle.

davidwengier:
Good point.

PR #5674, opened 2019-11-28
Update path capitalisation in project file

drewnoakes:


src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.vbproj
```diff
@@ -411,23 +411,23 @@
     <Compile Update="PropPages\AdvancedServicesDialog.vb">
       <SubType>UserControl</SubType>
     </Compile>
-    <Compile Update="proppages\AdvBuildSettingsPropPage.vb">
+    <Compile Update="PropPages\AdvBuildSettingsPropPage.vb">
```
davidwengier: Do we even need these files to be listed any more?

drewnoakes: Possibly not. There's a lot of metadata on these items.

PR #5675, opened 2019-11-28
VB comment cleanup

drewnoakes:
Also removes redundant numeric suffixes.

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/interop/NativeMethods.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/interop/NativeMethods.vb
@@ -16,10 +16,10 @@ Namespace Microsoft.VisualStudio.Editors.AppDesInterop
         Private Const VB_COMPILER_GUID As String = "019971d6-4685-11d2-b48a-0000f87572eb"
         Public Shared ReadOnly VBCompilerGuid As Guid = New Guid(VB_COMPILER_GUID)
 
-        '/ <summary>
-        '/     Handle type for HDC's that count against the Win98 limit of five DC's.  HDC's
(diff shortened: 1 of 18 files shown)


src/Microsoft.VisualStudio.AppDesigner/interop/NativeMethods.vb
```diff
@@ -16,10 +16,10 @@ Namespace Microsoft.VisualStudio.Editors.AppDesInterop
...
         Public Shared ReadOnly VBCompilerGuid As Guid = New Guid(VB_COMPILER_GUID)
 
-        '/ <summary>
-        '/     Handle type for HDC's that count against the Win98 limit of five DC's.  HDC's
-        '/     which are not scarce, such as HDC's for bitmaps, are counted as GDIHANDLE's.
-        '/ </summary>
+        ' <summary>
+        '     Handle type for HDC's that count against the Win98 limit of five DC's.  HDC's
+        '     which are not scarce, such as HDC's for bitmaps, are counted as GDIHANDLE's.
+        ' </summary>
```
davkean: Should be triple slash right?

drewnoakes: I believe so, yes. Will update.

davkean: This is spread throughout the changes not this one location.

drewnoakes: The others are `<devdoc>` which are not VB standard as far as I know.

PR #5680, opened 2019-11-29
Remove empty documentation elements

drewnoakes:
The VB files are long enough as they are. This helps cut some of them down a bit.

This deletes 3606 lines across 207 files.

Done with regex:

```regex
^\s*('''|///)\s*<(\w+)>(\s*\r?\n(\s*('''|///)\s*\r?\n)*\s*('''|///)\s*)?</\2>\s*\r?\n
```

This explicitly doesn't remove empty `<param>` elements as doing so could lead to warnings in cases where not all elements are removed.

Some BOMs seem to have been updated (hence 55 insertions).


davkean:
Thanks for doing this, I attempted this a while ago and gave up.

Issue #4247, opened 2018-11-09
Rationalize WPF build actions


davkean:
WPF build actions aren't show up in CPS because of the lack of AvailableItemName:

   
``` XML
   <AvailableItemName Include="SplashScreen" />
   <AvailableItemName Include="DesignData" />
   <AvailableItemName Include="DesignDataWithDesignTimeCreatableTypes" />

```

We can either add rules for them, or [implement AvailableItemName](https://github.com/dotnet/project-system/issues/113). The later would be preferable as it lets us hide WPF build actions when not targeting WPF.

davidwengier:
Have been looking into this today, some notes:

* CPS has basically no AvaialbleItemName support
  * There is one class that reads it, BuildFiles.cs, which is solely for C++
* There are no convenient extension points that would allow Project System to implement it on top of CPS
  * CPS wants people to use the rule system to provide additional items
* There is an AdditionalItemTypeProviders extension point that could be used in more places
* I'm investigating hooking it up in places where `ShowMissingItemTypes` is handled
* This would allow Project System to provide the AvailableItemName support, exposing it via AdditionalItemTypeProviders
* Alternatively CPS could simply look up the AvailableItemName structure directly

I will keep work on the PoC and see if something relatively straightforward is feasible.

davidwengier:
On a .NET Core console application (ie, plain SDK project) this adds a couple of options to the Build Action:

Before:
![image](https://user-images.githubusercontent.com/754264/50554195-5b0e2b00-0d0a-11e9-98a6-d8873893de89.png)

After:
![image](https://user-images.githubusercontent.com/754264/50554200-682b1a00-0d0a-11e9-9a60-165d6246c605.png)

davidwengier:
CPS PR: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/157290

I haven't looked at C++ cached mode yet because I need more context, but wanted to get something up and tests running while people are away.

davidwengier:
This is being solved via #113

davkean:
We can cleanup and delete all the other ones too, such as ApplicationDefinition, Page, etc if they don't contain extra properties over None.

davidwengier:
Good point, reopening.

davidwengier:
AdditionalFiles, ApplicationDefinition, Page and Resource as all the same as None.
XamlAppDef doesn't have LastGenOutput, DesignTime, AutoGen or CustomTool.

[...]

PR #5681, opened 2019-11-29
Remove unnecessary rules

davidwengier:
Fixes https://github.com/dotnet/project-system/issues/4247

This removes Xaml rule files for items where there is an `AvailableItemName` defined. Pay particular attention to https://github.com/dotnet/project-system/pull/5681/commits/f961ae45d3064907378f395e2dae6c28a0b3e9d7 which fixes a bug in the Fast Up to Date where it assumed that rule names match item types, but auto-generated rules in CPS do not follow that convention.


davkean:
Does this remove the ability to have a "localized" sentenced cased name in the Build Actions for these generated ones?

davidwengier:
Oh good point! Yes, there is no localization, the build actions are just shown in the dropdown. We'd have to invent something to extend `AvailableItemName` to allow it to supply a display name, and then support localization of same.

davkean:
@davidwengier I think this would just bring us back to legacy compat which I'm okay with - though a followup to add English-based sentence casing would be good.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -314,8 +315,12 @@ private State()
 
                 bool itemsChanged = false;
 
-                foreach ((string itemType, IProjectChangeDescription projectChange) in sourceItemsUpdate.ProjectChanges)
+                foreach ((string schemaName, IProjectChangeDescription projectChange) in sourceItemsUpdate.ProjectChanges)
                 {
+                    // ProjectChanges is keyed by the rule name which is usually the same as the item type, but not always (eg, in auto-generated rules)
+                    string? itemType = projectCatalogSnapshot.NamedCatalogs[PropertyPageContexts.File].GetSchema(schemaName)?.DataSource.ItemType;
```
drewnoakes: Is this indexer guaranteed to find its key?

davidwengier: There is certainly lots of spots in code where it's assumed to exist, but an extra question mark here wouldn't hurt much. 👍

davidwengier:
Any Microsofties who want to play around with this, the build is here: https://vsdrop.corp.microsoft.com/file/v1/Products/DevDiv/VS/8cfd0934755bca0955616644ece196aaf317f0a7/74cd36dc-9833-f208-0707-184a070f7ef9%3Bbootstrappers/Enterprise/vs_enterprise.exe

drewnoakes:
I was just reading the [CPS doc on extending XAML rules](https://github.com/microsoft/VSProjectSystem/blob/master/doc/extensibility/extending_rules.md) which explains that use of `OverrideMode="Replace"` _can only be performed on rules imported via MSBuild, and not by extension points._ Does this PR effect this?

davidwengier:
This change would break extended or replaced rules, but I looked through GitHub and index and couldn't see any that are relevant. I think the risk is pretty low.

davkean:
@davidwengier What's blocking this from going in?

davidwengier:
The WPF team had some automation test failures from the build of this, late last year, but I haven't heard whether they'd investigated and were caused by this change or not.

I've been meaning to rebase it and get a new build for it just in case, so thanks for the reminder :)

davidwengier:
New insertion is here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/227076

davidwengier:
The automation tests all passed :tada:

PR #5688, opened 2019-12-02
Remove comment about legacy project system

drewnoakes:


davkean:
Agreed.

drewnoakes:
I misread this as being about changing legacy.

PR #5694, opened 2019-12-03
Use reference equality for type checks

drewnoakes:
These are clearer to read and theoretically faster as they avoid calling the virtual `Equals` function.

Transformations:

1. `type1.Equals(type2)` ➡️ `type1 is type2`
2. `obj1.GetType() Is TypeOf(SomeType)` ➡️ TypeOf obj1 Is SomeType`

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
@@ -12,6 +12,7 @@ Imports System.Reflection
 Imports System.Resources
 Imports System.Runtime.Serialization
 Imports System.Text
+Imports System.Windows.Forms
 
 Imports Microsoft.VisualStudio.Editors.Common.Utils
 Imports Microsoft.VisualStudio.Shell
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -1481,11 +1482,11 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
                     'Value is not Nothing
 
                     'Verify that the new value is of the same type as before.
-                    If Not GetValueType().Equals(NewResourceValue.GetType()) Then
+                    If GetValueType() IsNot NewResourceValue.GetType() Then
                         'Exception to this rule.  We need to handle Byte() <-> MemoryStream conversion automatically
-                        If NewResourceValue.GetType().Equals(GetType(Byte())) AndAlso GetValueType().Equals(GetType(MemoryStream)) Then
```
davkean: These aren't these checks and am very worried about changing 15 year old with no tests.

This will now return true for `MyMemoryStream : MemoryStream`

drewnoakes: Quite right. This is a R# bug.

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AbEA3YaAuIUArgHYA+AkgLYAO0+AzgAQDKAno/jNQLABQABWLAMASzDMAshAAmxDDGaCoEAOZQAhn37M9ykeMmsR0zWNIAKAJQD995gBEx1ZiwC8zAERe7DvX7+AMIQpIwQigB0AOpQYtwAMhYwlgAq7LQwAPIAZm7MlCxZwABWMGD4trr+zCFhETAxcYnJloyRAOIw+OmZNpEAogCOxJoYjJZdPRkpxWUV1ouBzAOksmwiAqvrMvKKQA=

PR #5696, opened 2019-12-03
IVsShellUtilitiesHelper cleanup

drewnoakes:
For a brief moment in time I was looking to extend this interface and its implementation to provide the VS installation directory. But it didn't quite do what I wanted.

Pulled these cleanup commits out to their own branch.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/IVsShellUtilitiesHelper.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/IVsShellUtilitiesHelper.cs
@@ -10,10 +10,13 @@ namespace Microsoft.VisualStudio.Shell.Interop
     internal interface IVsShellUtilitiesHelper
     {
         /// <summary>
-        /// Returns the version of VS as defined by VSVSAPROPID_ProductSemanticVersion with the trailing sem version stripped, or null on failure. 
+        /// Returns the version of VS as defined by <see cref="VSAPropID.VSAPROPID_ProductSemanticVersion"/> with the trailing sem version stripped, or <see langword="null"/> on failure.
         /// </summary>
         Task<Version?> GetVSVersionAsync(IVsService<IVsAppId> vsAppIdService);
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/IVsShellUtilitiesHelper.cs
```diff
@@ -2,18 +2,24 @@
...
 using System;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.VS;
 using Microsoft.VisualStudio.ProjectSystem.VS.Interop;
 
 namespace Microsoft.VisualStudio.Shell.Interop
 {
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
```
davkean: This part that implements this imports IProjectThreadingService which lives in ProjectService scope. The unit test has caught that.

drewnoakes: Timely, given recent discussions. Will fix.

drewnoakes:
```suggestion
    [ProjectSystemContract(ProjectSystemContractScope.ProjectService, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
```

davkean:
Unrelated, but these are basically extension methods, don't really need to be hidden behind interface, they should hide the VS interfaces, or just be converted to an extension method over the top of IVsXXX interface with the consumer switching before hand.

drewnoakes:
The docs suggest this interface exists for unit testing purposes.

davkean:
It probably was way more complicated at one point, it just wraps interfaces of which one has a default no op (switchtoUithread)

PR #5699, opened 2019-12-03
Invoke single-file generators in cloud scenarios

tmeschter:
_Note that this is incomplete and not ready for review._

### DIFF
--- a/build/import/HostAgnostic.props
+++ b/build/import/HostAgnostic.props
@@ -61,9 +61,6 @@
     <PackageReference Include="Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime" />
     <PackageReference Include="Microsoft.VisualStudio.ImageCatalog" />
     <PackageReference Include="StreamJsonRpc" />
-
-    <PackageReference Include="Microsoft.ServiceHub.Framework" />
-
   </ItemGroup>
 
   <ItemGroup Condition="'$(IsUnitTestProject)' == 'true'" >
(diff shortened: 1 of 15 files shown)


tmeschter:
I should state that this implementation is incomplete and not yet ready for review.

tmeschter:
@davidwengier The `RunCustomTool` method exists on the `VSProjectItem` interface, and there is no implementation of this on the client in a cloud-connected scenario.

Long-term we will probably always go through the new interface, even in local-only scenarios, and calls to `RunCustomTool` go away entirely. For the moment, however, I don't want to risk breaking existing scenarios.

davidwengier:
No DTE? Okay, that's worth knowing, thanks.

> Long-term we will probably always go through the new interface

If this is true, is it worth establishing as pattern that any new interface be Async first, just in case?

tmeschter:
For the moment at least there is no useful DTE. At some point there might be a "light-weight" project system on the client side, but it would exist largely to provide easier access to cloud services, and right now it isn't clear where we would draw the line between what goes on the client and what goes on the server.

Cloud services will always be async. I've wrapped it here in synchronous code simply because the existing code is synchronous.

davkean:
What's going on with this PR? Can we close this and reopen if still applicable.

tmeschter:
Yeah, I'll close it for the moment. The correct behavior depends on some minor changes on the server side but those aren't a high priority at the moment.

PR #5701, opened 2019-12-04
Stub out a new assembly

tmeschter:
Stub out a new assembly, Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client. This assembly will hold code meant to run on the client in cloud-connected scenarios.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.Packaging
+{
+    [Guid(PackageGuid)]
+    [PackageRegistration(AllowsBackgroundLoading = true, RegisterUsing = RegistrationMethod.CodeBase, UseManagedResourcesOnly = true)]
+
+    internal sealed class ManagedProjectSystemClientPackage : AsyncPackage
+    {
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.Packaging
+{
+    [Guid(PackageGuid)]
+    [PackageRegistration(AllowsBackgroundLoading = true, RegisterUsing = RegistrationMethod.CodeBase, UseManagedResourcesOnly = true)]
+
+    internal sealed class ManagedProjectSystemClientPackage : AsyncPackage
```
davkean: Why do we need a package? I think we should only add it if actuallay needed.

davkean: We don't need a package to produce a pkgdef, if that's why you added it.

dmonroym:
There should also be an associated PR in the VS repo so this gets inserted properly, right?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
tmeschter: I'm going to need it later when I start defining menu items and need a place to put `[ProvideMenuResource("Menus.ctmenu", 1)]`. At least, I think I do.

davkean: Can we wait until then before we check this in? I just deleted a bunch of pre-emptive packages.

davkean:
This doesn't look applicable anymore given https://github.com/dotnet/project-system/pull/5854, please reopen if it still is.

PR #5709, opened 2019-12-06
Add UI support for property "AuthenticationMode" to debug property page.

srdjanjovcic:


srdjanjovcic:
@davkean @tmeschter -- Can you review those changes?

davkean:
Are these new types deliberately public? Please don't add new public API here.

srdjanjovcic:
Regarding public APIs. This is only way I found that you can actually implement user-serialized properties in property pages correctly.

If property page cannot access configuration state at top top of the view (which is shared among different property pages), it cannot implement user-serialized properties which understand selected configuration/platform.

tmeschter:
@srdjanjovcic But why do they have to be public? Why won't `Friend` work?

srdjanjovcic:
@tmeschter They are in different assembly.

drewnoakes:
Can we use `InternalsVisibleTo` instead?

srdjanjovcic:
@drewnoakes I do not know -- would it be acceptable? I am OK as long as team is OK with it.

I see that there are some InternalsVisibleTo used in project-system codebase (e.g., exposing Microsoft.VisualStudio.ProjectSystem.Managed.dll internals to Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll)

Since those two are built in same code base, I guess that would be OK to do this?

drewnoakes:
IVT isn't great, but it's better than adding public API. I'd vote for it here. Anyone object?

srdjanjovcic:
I have updated PR to use InternalsVisibleTo

srdjanjovcic:
@drewnoakes. @tmeschter, @davkean -- Any further feedback? Can we get this PR accepted?

PR #5731, opened 2019-12-11
Avoid Tuple allocation for Dataflow sync link

drewnoakes:
A while back I extended CPS to allow using `ValueTuple` as well as `Tuple` for sync-link'd data. This utilises that change on our side.


drewnoakes:
@davkean seems there's a problem with the CLA bot. I no longer have admin permission. Can you merge this please?

Issue #5702, opened 2019-12-05
Project import tree improvements


drewnoakes:
Following #4805 and #5639 there are some pending items we are considering for the project import tree:

- [ ] Browse object showing path, label, group label
- [ ] Option to show these at all times
- [ ] Include imported files in "Find in Files"
- [ ] (Maybe) track "Show all files" state via Dataflow

drewnoakes:
- [ ] While the import tree is visible, drag/dropping the project's `.csproj.user` file into the editor gives:

  > 'ClassLibrary1.csproj.user' cannot be opened. The file is not supported.

drewnoakes:
- [ ] Review context menu on the "Imports" node

drewnoakes:
- [ ] Review whether ordering of children in tree reflects ordering of imports, including in the case of updates that change order
- [ ] Investigate the 'Search within file contents' option of Solution Explorer's search:
   ![image](https://user-images.githubusercontent.com/350947/70571401-efa72b80-1bf1-11ea-9396-b2c246667d05.png)

davkean:
@drewnoakes This is marked tracking but it actually has work inside of it.

drewnoakes:
@shyamnamboodiripad I looked into the ordering of tree items. They're being displayed alphabetically. Once #5736 lands, they'll be displayed in the correct evaluation order.

shyamnamboodiripad:
> They're being displayed alphabetically. Once #5736 lands, they'll be displayed in the correct evaluation order.

@drewnoakes Sounds great - Thank you!

davkean:
@drewnoakes Can you split this up into individual issues that folks might be able to bite off under "Help-Wanted"?

drewnoakes:
I will once my current bout of work on it slows down.

davkean:
Find in Files:

- It appears that the new search only searches folders that return GUID_ItemType_PhysicalFolder for VSHPROPID_TypeGuid and does not respect VSHPROPID_IsNonSearchable. The imports node is a virtual folder and is not considered GUID_ItemType_PhysicalFolder, nor do we want to for it to be considered that as that brings other behavior along with it, such as provided a file name from ProjectItem.FileNames.

PR #5736, opened 2019-12-16
Display import tree items in correct order

### LINKED ISSUE #5711: Imports tree should support duplicate node names
Originally this was an error report (see below). That error has now been fixed, but duplicate nodes are hidden. Instead, we should show all nodes so that the import tree is not misleading.

---

A project file that imports two files with the same file name causes this exception:

```
System.ArgumentException: An item with the caption "MicroBuild.Plugins.SwixBuild.props"(ProjectTreeFlags: FileOnDisk ProjectImport) already exists under parent "MicroBuild.Core.props" with the same sort order as the item being added.
   at Microsoft.VisualStudio.ProjectSystem.Designers.ProjectTree.Add(UnattachedProjectTreeNode node)
   at Microsoft.VisualStudio.ProjectSystem.Designers.ProjectTree.Add(IProjectTree subtree)
   at Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|20_8(IReadOnlyList`1 imports, IProjectTree tree)
   at Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|20_8(IReadOnlyList`1 imports, IProjectTree tree)
   at Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|20_8(IReadOnlyList`1 imports, IProjectTree tree)
   at Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|20_8(IReadOnlyList`1 imports, IProjectTree tree)
   at Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|20_8(IReadOnlyList`1 imports, IProjectTree tree)
   at Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<>c__DisplayClass20_0.<set_ShowAllFiles>b__7(IProjectVersionedValue`1 currentTree, ConfiguredProjectExports configuredProjectExports, CancellationToken token)
```

Names should be disambiguated to be made unique.

Repro: https://devdiv.visualstudio.com/DevDiv/_git/VS-Platform?path=%2Fsrc%2FEditor

drewnoakes:
Fixes #5711. 
Relates to #5754.
Relates to #5702.

Previously, import tree items would always be displayed in alphabetical order. With this change, their tree order reflects the actual order they are evaluated in. This is helpful when tracing the value of a given property, for example.

For CPS to respect the `IProjectTree2.DisplayOrder` property, the `SortByDisplayOrder` capability must be present. This was only used in F# projects before where it controlled both the display and manipulation of item order. Consequently a new `EditableDisplayOrder` capability has been added which covers the manipulation component, and has been added to F#-specific commands/exports.

### Before

Alphabetical:

image

### After

In evaluation order:

image

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AddExistingItemAboveCommand.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AddExistingItemAboveCommand.cs
@@ -9,7 +9,7 @@ using Task = System.Threading.Tasks.Task;
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
 {
     [ProjectCommand(CommandGroup.ManagedProjectSystemOrder, ManagedProjectSystemOrderCommandId.AddExistingItemAbove)]
-    [AppliesTo(ProjectCapability.SortByDisplayOrder)]
+    [AppliesTo(ProjectCapability.SortByDisplayOrder + " & " + ProjectCapability.EditableDisplayOrder)]
     internal class AddExistingItemAboveCommand : AbstractAddItemCommand
     {
         private readonly IAddItemDialogService _addItemDialogService;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -82,6 +82,12 @@
     <ProjectCapability Include="LaunchProfiles" />
     <ProjectCapability Include="NoGeneralDependentFileIcon"/>
     <!--
+      We want the tree to respect IProjectTree2.DisplayOrder when it is set.
+      This is different to capability EditableDisplayOrder, used by F# for example, which
+      allows the user to manipulate display order values.
+    -->
+    <ProjectCapability Include="SortByDisplayOrder"/>
+    <!--
       List of capabilities below is adding back common capabilities defined in imported targets.
       We disabled them with the property DefineCommonCapabilities=false to get rid of default
       References capability, but everything else we want back.
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -82,12 +82,6 @@
     <ProjectCapability Include="LaunchProfiles" />
     <ProjectCapability Include="NoGeneralDependentFileIcon"/>
     <!--
-      We want the tree to respect IProjectTree2.DisplayOrder when it is set.
-      This is different to capability EditableDisplayOrder, used by F# for example, which
-      allows the user to manipulate display order values.
-    -->
-    <ProjectCapability Include="SortByDisplayOrder"/>
-    <!--
       List of capabilities below is adding back common capabilities defined in imported targets.
       We disabled them with the property DefineCommonCapabilities=false to get rid of default
       References capability, but everything else we want back.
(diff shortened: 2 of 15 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/AddExistingItemAboveCommand.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Input.Commands.Ordering
 {
     [ProjectCommand(CommandGroup.ManagedProjectSystemOrder, ManagedProjectSystemOrderCommandId.AddExistingItemAbove)]
-    [AppliesTo(ProjectCapability.SortByDisplayOrder)]
+    [AppliesTo(ProjectCapability.SortByDisplayOrder + " & " + ProjectCapability.EditableDisplayOrder)]
```
davidwengier: It's probably worth having a constant for this since the combination is used frequently

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -81,6 +81,12 @@
     <ProjectCapability Include="ProjectImportsTree" />
     <ProjectCapability Include="LaunchProfiles" />
     <ProjectCapability Include="NoGeneralDependentFileIcon"/>
+    <!--
+      We want the tree to respect IProjectTree2.DisplayOrder when it is set.
+      This is different to capability EditableDisplayOrder, used by F# for example, which
+      allows the user to manipulate display order values.
+    -->
+    <ProjectCapability Include="SortByDisplayOrder"/>
```
davkean: This has a significant impact on performance of tree updates. This isn't designed to be enabled for large trees that you might see in C#.

davkean: F# has this as a fixed capability (comes from project type) probably for a reason, it probably cannot be dynamic (comes from targets).

davkean: @lifengl is there a way we can enable this just for a graft and not the entire tree?

drewnoakes: I'm not convinced there's a significant perf impact. The only place CPS opts in to the capability is `DisplayOrderProjectScopeUnattachedProjectTreeNodeFactory` which derives from `ProjectScopeUnattachedProjectTreeNodeFactory` with higher precedence, customising two things:

1. Modifing the `IComparer<UnattachedProjectTreeNode>` used for sets of child nodes to be [`DisplayOrderProjectTreeSort`](http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Implementation&file=Designers%5CDisplayOrderProjectTreeSort.cs) rather than `ProjectTreeSort`. This comparer's overhead is negligible.
1. Overriding [`IndexOf`](http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Implementation&file=Designers%5CDisplayOrderProjectScopeUnattachedProjectTreeNodeFactory.cs&line=42) for looking up a child by caption. This override allocates less memory but has slightly worse algorithmic complexity than the base it replaces (O(N) vs O(4 log N)). I've filed https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/219245 which removes allocation by the override altogether.

drewnoakes:
> needs work to make it so

Please elaborate on this so I can do something about it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
drewnoakes: Have made this a fixed capability.

drewnoakes:
c0296d88a6a8ebf5da7afceaa81aa2e2de2e91cd removes `SortByDisplayOrder` from the default project capabilities and introduces a workaround that addresses #5711 until we're able to support display orders only for specific root graft subtrees.

PR #5743, opened 2019-12-16
WIP Changed table layout panel in BuildPropPage

melytc:
Fixes 'Advanced' button not displaying properly. Apparently, #5636 still had the button truncated.

DO NOT MERGE

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
@@ -1318,7 +1318,7 @@
     <value>18</value>
   </data>
   <data name="overarchingTableLayoutPanel.Size" type="System.Drawing.Size, System.Drawing">
-    <value>539, 531</value>
+    <value>539, 800</value>
   </data>
   <data name="overarchingTableLayoutPanel.TabIndex" type="System.Int32, mscorlib">
     <value>0</value>


drewnoakes:
@melytc I added "WIP" to your PR title so that the WIP bot will prevent merging. You can remove it when you're ready for it to be reviewed/merged.

src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
```diff
@@ -1318,7 +1318,7 @@
     <value>18</value>
   </data>
   <data name="overarchingTableLayoutPanel.Size" type="System.Drawing.Size, System.Drawing">
-    <value>539, 531</value>
+    <value>539, 800</value>
```
drewnoakes: How did you come up with this value? If I'm not mistaken, these are just hints to WinForms for the initial layout. If the `overarchingTableLayoutPanel` is docked, it'll resize according to its container.

That said, the proof's in the pudding :) Do you have before/after screenshots?

dmonroym: This whole PR is just a trial to get a VS build to test it in the Japanese OS, the number is an exaggeration to try and see if it would actually do something

melytc: So this is just a test to see if incrementing the size would fix the problem. Since the bug only appears in my VM with a particular OS, I'm trying to create this build in AzDO and see if the change reflects on the VM.

melytc:
Re-routed bug to corresponding team, closing this PR.

PR #5763, opened 2019-12-23
Add project file editing integration tests

davidwengier:
Been meaning to do this for a while. It was surprisingly painless.

### DIFF
--- a/tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/ProjectFileEditorTests.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/ProjectFileEditorTests.cs
@@ -1,0 +1,256 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Test.Apex.VisualStudio.Search;
+using Microsoft.Test.Apex.VisualStudio.Solution;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using VSLangProj;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    [TestClass]
+    public sealed class ProjectFileEditorTests : IntegrationTestBase
+    {
+        [TestMethod]
+        public void ProjectFileEditor_DoubleClick_OpensEditor()
+        {
+            var solution = VisualStudio.ObjectModel.Solution;
+
+            ProjectTestExtension consoleProject;
+
+            using (Scope.Enter("Create Project"))
+            {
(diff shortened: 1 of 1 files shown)


tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/ProjectFileEditorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+            using (Scope.Enter("Verify Text Buffer Changed"))
+            {
+                var editor = VisualStudio.ObjectModel.WindowManager.ActiveDocumentWindowAsTextEditor;
+                editor.Editor.Verify.ContainsText("<OutputType>Library</OutputType>");
+            }
+
+            using (Scope.Enter("Verify Project Unchanged"))
+            {
+                // project reload is on a one second delay
+                Services.Synchronization.WaitFor(TimeSpan.FromSeconds(1), () => consoleProject.Properties.Verify.PropertyValueIs(ProjectProperty.OutputType, prjOutputType.prjOutputTypeLibrary));
```
drewnoakes: Why wait this way rather than, say, `await Task.Delay(...)`?

davidwengier: Mostly a guess that this was "idiomatic" apex, and partly because if the `DoubleClick()` call takes a while (and apex doesn't seem speedy), this might not have to wait at all.

tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/ProjectFileEditorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+                consoleProject.DoubleClick();
+            }
+
+            using (Scope.Enter("Verify Text Buffer Changed"))
+            {
+                var editor = VisualStudio.ObjectModel.WindowManager.ActiveDocumentWindowAsTextEditor;
+                editor.Editor.Verify.ContainsText("<OutputType>Library</OutputType>");
+            }
+
+            using (Scope.Enter("Verify Project Unchanged"))
```
davidwengier:
```suggestion
            using (Scope.Enter("Verify Project Changed"))
```

tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/ProjectFileEditorTests.cs
drewnoakes: Oh, this polls for the condition?

davidwengier: Yeah, the first parameter is a timeout but it will exit early if the condition passes.

PR #5764, opened 2020-01-01
Consider existing child captions when preventing duplicates

drewnoakes:
Improves the workaround for #5711 added in #5736.

Longer term we'll want to enable display order within the tree so that items with duplicate captions appear correctly. Without that, the tree is misleading. For example, an ASP.NET Core project has four sibling `Sdk.props` files:

image


drewnoakes:
> Longer term we'll want to enable display order within the tree so that items with duplicate captions appear correctly.

PR for the CPS side of this is up: https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/220752

PR #5766, opened 2020-01-06
Fix broken link in documentation

drewnoakes:


dmonroym:
@drewnoakes the unit-tests will never run since this is in the docs directory. Let me know if you want me to merge in these changes.

drewnoakes:
@dmonroym yes please. I would have committed this directly to `master` but no longer have permission to do so.

PR #5770, opened 2020-01-08
Don't dispose PropPageDesignerDocData when designer is closing

### LINKED ISSUE #5723: no localregistry service message popping up
_This issue has a corresponding [ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/717378/no-localregistry-service-message-popping-up.html). Please vote and comment there to make sure your voice is heard._

_Copied from [AB#991802](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/991802)_

---
fresh install of VS 2019 Preview, was updating nuget properties (within the project view) and after changing release notes this pop up started showing up, preventing other operations in VS.. had to kill the process

**Detailed repro steps:**

1. Open VS and create a new Class Library project.
2. Open the project's Properties page.
3. Change the Target Framework under the Application tab. For example, change from .NET Standard 2.0 to 2.1
4. Save.
5. In the Team Explorer window, undo the changes. (You'll have to link it to a git project)
6. Confirm that you want to undo the changes in the pop message.

---
#### Visual Studio Feedback System on 30/08/2019, 07:44 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

#### Neil Highley on 02/09/2019, 07:33 PM: 

<p>Just got this after attempting to change the .net version from a projects properties window. Browsing to a source code file allowed me to restart visual studio. After which, the error didn't appear.</p>

#### Robert Balsley on 11/09/2019, 03:53 AM: 

<p>Received this error too when trying to increment the File and Assembly version numbers in Visual Studio version 16.2.4. Closing the solution and reopening it worked for me.</p>

#### Eric Brunnett on 11/09/2019, 09:40 AM: 

<p>I am receiving this error as well. I just hit debug (after about the 50th time today) and all of sudden this popped up. Any click inside visual studio causes the error to resurface, including just clicking on the VS icon on the menu bar so I couldn't quit VS without killing the process. I had multiple instances running, only one instance was affected, even while that instance was still running, I could use the others without getting the error.</p>

#### Visual Studio Feedback System on 26/09/2019, 01:38 AM: 

<p>This issue is currently being investigated. Our team will get back to you if either more information is needed, a workaround is available, or the issue is resolved.</p>

#### Melissa Trevino [MSFT] on 02/10/2019, 07:06 AM: 

<p>Hello! Thank your for reporting this issue.<br>
In order to understand what triggered that pop-up message, we will need more information; could you send us a dump of the VS session?<br>
Here are some steps you could follow: <a target='_blank' href="https://docs.microsoft.com/en-us/visualstudio/ide/report-a-problem-minidumps?view=vs-2017">https://docs.microsoft.com/en-us/visualstudio/ide/report-a-problem-minidumps?view=vs-2017</a>.<br>
If that does not work, another option is to open Task Manager (open ‘More details’ option), open the VS processes, right-click on the VS window where that pop-up message happened, and select ‘Create dump file’.</p>

#### Jesper Staun Hansen on 02/10/2019, 11:40 PM: 

<p>I met this when I used right click file -&gt; "Undo Changes" in "Team Explorer - Changes"</p><p>It happened like this:</p><p>1. Open a project properties and change it from netstandard 2.0 to 2.1 (and keep properties open and in focus throughout)</p><p>2. Save All</p><p>3. In "Team Explorer - Changes" find the .csproj file in the "Changes" section.</p><p>4. Right click and "Undo changes"</p><p>5. Confirm</p><p>Visual Studio will get a bit freaky and maybe give the popup (a lot of times in a row).</p>

#### vaclav.haken on 03/10/2019, 00:43 AM: 

(private comment, text removed)
#### Scott Johnson on 03/10/2019, 00:46 PM: 

<p>Same problem here, Thanks; for the browsing another source file.  That allowed me to at least quit VS.</p>

#### Michael Lücke on 08/10/2019, 07:11 PM: 

<p>The same popup for me when I switched the target platform of my project from x86 to AnyCPU. Fortunately I was able to continue working.</p>

#### vaclav.haken on 08/10/2019, 08:44 PM: 

<p>I have given the dump to MS (the message with it) is marked as private (and visible to MS employees only then) so hopefully it will allow them to fix it</p>

#### Basim Kadhim on 16/10/2019, 10:11 AM: 

<p>Also encountering this problem periodically. Once it starts happening, lots of things trigger it to keep coming up. Also used the trick of opening another source file just to be able to shut down VS.</p>

#### Melissa Trevino [MSFT] on 17/10/2019, 08:10 AM: 

<p>Hello! We are still working on fixing this problem. </p><p>I found that this happens specifically when you revert changes while having the Properties Page open. To avoid it, <strong>close the Properties Page before undoing your changes</strong>. That should not trigger the message and will let you continue working.</p>

#### Visual Studio Feedback System on 18/10/2019, 10:22 AM: 

<p>This issue is currently being investigated. Our team will get back to you if either more information is needed, a workaround is available, or the issue is resolved.</p>

#### Antonin Prochazka on 26/10/2019, 01:54 AM: 

<p>Same here, version 16.3.5.</p>

#### Mike Harris on 29/10/2019, 03:45 AM: 

<p>getting the same error anytime i try to change a value on the project publish page.</p><p>vs 2019 v16.3.6</p>

#### Ross Gardiner on 12/11/2019, 08:03 PM: 

<p>I also see this problem in VS 16.3.7.</p>

#### Kaine Varley on 12/11/2019, 10:10 PM: 

<p>I was forced to End Task using Task Manager, delete my local repo and re-clone...</p><p>(VS 16.3.8)</p>

#### Antonin Prochazka on 15/11/2019, 08:27 PM: 

<p><a rel="user" href="https://developercommunity.visualstudio.com/users/183166/6856edb2-54d2-464e-a8d6-e071057145d4.html" nodeid="183166">@Kaine Varley</a> maybe deleting the .vs folder in your working folder would have been enough. It solves various weird situations in Visual Studio for us. Definitely doing hard git clean is faster then deleting and re-cloning the repo and it has the same effect.</p>

#### Kaine Varley on 15/11/2019, 08:41 PM: 

<p>Thanks <a rel="user" href="https://developercommunity.visualstudio.com/users/55050/f3aea587-be8e-4922-b043-771f253f44b1.html" nodeid="55050">@Antonin Prochazka</a>. </p><p>As discussed earlier in the thread, but wasn't clear to me at the time, the problem appears to be when the property page of a project (in the solution) is open. </p><p>A fix for this is sorely needed...</p><p>- Kaine</p>

---
#### john.dinnen solved on 13/11/2019, 04:12 AM, 1 votes: 

<p>I unloaded the solution and reloaded it again from the Solution Explorer.</p>

#### Nicholas Teeter on 27/11/2019, 11:56 PM: 

<p>Same issue happened to me when modifying the post-build event.  I had to unload and reload the project from the Solution Explorer.</p>

#### Michael Lücke solved on 08/10/2019, 07:16 PM, 0 votes: 

<p>... but just one click more forced me to quit VS via task manager.</p>

#### Scott Johnson solved on 03/10/2019, 00:42 PM, 0 votes: 

<p>Same problem here.  Thanks for the browsing another source file; I was stuck until then.</p>

davidwengier:
Fixes #5723

A change was made to the shell to fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/684877 which means that while reopening document windows, all of the doc data are kept in the running document table (see code [here](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fenv%2Fmsenv%2Fcore%2Fvsshlsvc.cpp&version=GC6ff311afae3a6bfbd18cd01097c84fa8aa03f8e5&line=4107&lineEnd=4108&lineStartColumn=1&lineEndColumn=1&lineStyle=plain)). This broke the property pages because they always created a new doc data for each property page tab as its being created, but if the doc data was already in the RDT, that new one never got in there so not only was the new one orphaned (never initialized), worse the doc data that is in the RDT continued to be used by other components in Visual Studio but it had been disposed.

This fixes the issue by changing the `CreateDesigner` function to check for and re-use doc data in the RDT already, and makes the `PropPageDesignerDocData` take care of disposing itself when it is removed from the RDT.

There is a slight change to the public API here, removing a method that now has no function, but its easy enough to put that back as a no-op if anyone thinks its an issue.

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerPanel.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerPanel.vb
@@ -170,13 +170,13 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                 Return _customViewProvider
             End Get
             Set
-                If _customViewProvider Is value Then
+                If _customViewProvider Is Value Then
                     Exit Property
                 End If
 
                 'Close anything currently showing
                 CloseFrame()
-                _customViewProvider = value
+                _customViewProvider = Value
                 If Visible Then
                     ShowDesigner()
                 End If
@@ -797,13 +797,13 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                 Return _tabTitle
             End Get
             Set
-                _tabTitle = value
+                _tabTitle = Value
 
                 ' We set Window.Text to be the page Title to help screen reader.
-                If String.IsNullOrEmpty(value) Then
+                If String.IsNullOrEmpty(Value) Then
                     Text = String.Empty
                 Else
-                    Text = My.Resources.Designer.GetString(My.Resources.Designer.APPDES_PageName, value)
+                    Text = My.Resources.Designer.GetString(My.Resources.Designer.APPDES_PageName, Value)
                 End If
 
                 _pageNameLabel.Text = Text
@@ -818,7 +818,7 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
                 Return _tabAutomationName
             End Get
             Set
-                _tabAutomationName = value
+                _tabAutomationName = Value
             End Set
         End Property
 
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerPanel.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerPanel.vb
@@ -275,6 +275,13 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
 
                         If DocData IsNot Nothing Then
                             ExistingDocDataPtr = Marshal.GetIUnknownForObject(DocData)
+                        Else
+                            ' See if there is doc data in the RDT
+                            Dim rdt As New Shell.RunningDocumentTable(_serviceProvider)
+                            Dim rdtInfo As Shell.RunningDocumentInfo = rdt.GetDocumentInfo(MkDocument)
+                            If rdtInfo.DocData IsNot Nothing Then
+                                ExistingDocDataPtr = Marshal.GetIUnknownForObject(rdtInfo.DocData)
+                            End If
                         End If
                         OleServiceProvider = CType(_serviceProvider.GetService(GetType(OLE.Interop.IServiceProvider)), OLE.Interop.IServiceProvider)
                         Debug.Assert(OleServiceProvider IsNot Nothing, "Unable to get OleServiceProvider")
--- a/src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerDocData.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerDocData.vb
@@ -24,8 +24,6 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
         Implements IVsPersistDocData2
         Implements OLE.Interop.IObjectWithSite
         Implements IVsTextBufferProvider
-        Implements IVsRunningDocTableEvents
-        Implements IVsRunningDocTableEvents4
 
         'Event support for IVsTextBufferDataEvents
         Public Delegate Sub LoadCompletedDelegate(Reload As Integer)
@@ -49,9 +47,6 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
         ' Dirty and readonly state
         Private _isReadOnly As Boolean
         Private _isDirty As Boolean
-
-        ' Cookie for RunningDocumentTable events
-        Private _rdtCookie As UInteger
 
         ''' <summary>
         ''' Constructor
@@ -189,8 +184,7 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
         End Function
 #End Region
         Public Function Close2() As Integer Implements IVsPersistDocData2.Close
-            'Nothing to do here, no real file to close
-            Return NativeMethods.S_OK
+            Dispose(True)
         End Function
 
         Public Function GetGuidEditorType2(ByRef pClassID As Guid) As Integer Implements IVsPersistDocData2.GetGuidEditorType
@@ -220,9 +214,6 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
         Public Function LoadDocData2(pszMkDocument As String) As Integer Implements IVsPersistDocData2.LoadDocData
             'Nothing to do here, no real file to load
             _mkDocument = pszMkDocument
-            ' Sign up for RDT events so we know when to dispose ourselves - there is no point doing this until we have the document moniker to compare to
-            Dim rdt As New Shell.RunningDocumentTable
-            _rdtCookie = rdt.Advise(DirectCast(Me, IVsRunningDocTableEvents))
             RaiseEvent OnLoadCompleted(0) 'FALSE == 0
         End Function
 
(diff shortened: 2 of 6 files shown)


src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerPanel.vb
```diff
@@ -275,6 +275,13 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
 
                         If DocData IsNot Nothing Then
                             ExistingDocDataPtr = Marshal.GetIUnknownForObject(DocData)
+                        Else
```
davkean: The way I have this pictured in my mind; our editor factory gets called with the existing doc data - why do we have to look it up again? We should already be passed it.

davkean: ie What does `DocData` represent if not the existing doc data?

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerEditorFactory.vb
```diff
@@ -121,6 +122,11 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
                     DesignerLoader.Dispose()
                 End If
 
+                ' If we created the doc data then we should dispose it for a failure
+                If ExistingDocData Is Nothing Then
```
davkean: We always create the DocData, so I'm not sure I understand this comment, or why this line is needed.

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerDocData.vb
```diff
@@ -215,6 +220,9 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
         Public Function LoadDocData2(pszMkDocument As String) As Integer Implements IVsPersistDocData2.LoadDocData
             'Nothing to do here, no real file to load
             _mkDocument = pszMkDocument
+            ' Sign up for RDT events so we know when to dispose ourselves - there is no point doing this until we have the document moniker to compare to
```
davkean: Dispose should be done in response to IVsPersistDocData.Close, that's how folks around the tree do it.

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerEditorFactory.vb
davkean: I don't think we should ever explicitly dispose, and instead make `IVsPersistDocData.Close` do it and it appears to be explicitly called by the RDT when its done with.

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerPanel.vb
davidwengier: Your picture is correct, our editor factory does get called with the existing doc data. This method is the thing that calls it, via `OpenSpecificEditor`, and supplies that doc data.

The `DocData` property is the existing doc data that this instance of the ApplicationDesignerPanel knows about, and would only be a factor if `CreateDesigner()` is more than once on an instance. I didn't see that happen during debugging.

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerEditorFactory.vb
davidwengier: > We always create the DocData, so I'm not sure I understand this comment, or why this line is needed.

We only create the doc data here if there isn't an ExistingDocData passed in. This line is just cleaning up if things fail to initialize under the assumption that in such a failed case, the doc data wouldn't end up in the RDT so could never be cleaned up via its removal.

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerDocData.vb
davidwengier: Thank you! An excellent suggestion and change. I didn't think to look at that because the current Dispose calls Close() on the wrapped text buffer via `IVsPersistDocData` so I thought the relationship went the other way, but you're right the RDT calls Close anyway so there is no need for events!

PR #5773, opened 2020-01-08
Adding OutDir property for msbuild.

ocallesp:
Fixes #4737.

outputRelativeOrFullPath was using the legacy property OutPath.
OutDir is used instead to fix the fast up-to-date check.

PR #5778, opened 2020-01-10
Clean up OutDir property

ocallesp:
Moved to the right position in the alphabetical order.
Removed display name since this doesn't display anywhere.


davkean:
There was still an [outstanding question around tests](https://github.com/dotnet/project-system/pull/5773#issuecomment-572291211), can we have a followup PR for that item too.

PR #5782, opened 2020-01-12
Async methods should have correct suffix

drewnoakes:


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ThreadingExtensions.cs
```diff
@@ -26,7 +26,7 @@ internal static class ThreadingExtensions
         /// <exception cref="OperationCanceledException">
         ///     The result is awaited and <paramref name="cancellationToken"/> is cancelled.
         /// </exception>
-        public static JoinableTaskFactory.MainThreadAwaitable SwitchToUIThread(this IProjectThreadingService threading, CancellationToken cancellationToken)
```
davkean: This was deliberate so that it would overload existing method.

drewnoakes: Ok, I'll back this out and add a comment to that effect.

drewnoakes: #5790

PR #5783, opened 2020-01-13
Property page tidying

drewnoakes:
Part 1 of 2 ahead of remote-debugging support for the property pages. Part 2 is annotations, and is more involved as `DebugPageViewModel` was difficult to annotate.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/ProjectLaunchSettingsUIProvider.cs
```diff
@@ -11,45 +11,29 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
     /// </summary>
     [Export(typeof(ILaunchSettingsUIProvider))]
     [AppliesTo(ProjectCapability.LaunchProfiles)]
-    [Order(0)]              // Lowest priority to allow this to be overridden
+    [Order(Order.Lowest)]              // Lowest priority to allow this to be overridden
```
melytc: Nit: should we remove the white space here too?

drewnoakes: Yes, I'll fix that up too, thanks!

PR #5784, opened 2020-01-13
Basic remote debugging support

### LINKED ISSUE #3756: Support remote debugging project properties and expose in app designer
_From @TFTomSun on July 18, 2018 11:10_

Using the old MSBuild Format it was possible to specify remote debugging settings like this:

```xml
  <PropertyGroup>
    <StartAction>Program</StartAction>
    <StartProgram>C:\MyRemote\Executable.exe</StartProgram>
    <RemoteDebugEnabled>true</RemoteDebugEnabled>
    <RemoteDebugMachine>MyRemoteMachine</RemoteDebugMachine>
  </PropertyGroup>
```
These properties seem to be not be evaluated by the new MSBuild Format, even if the Project is marked as a Net4.6.1 project.

Is that a bug or does the new MSBuild Format don't support remote debugging by simply starting debugging anymore? 

_Copied from original issue: Microsoft/msbuild#3524_

drewnoakes:
Fixes #3756.

Add's a "remote machine hostname" option to the debug property page and passes any entered value to the debugger.

image

If _Use remote machine_ is checked, a non-empty hostname must be specified otherwise the form will not validate, cannot be saved, and will show the validation message in a tool tip:

image

image

Values are saved in `launchSettings.json` alongside other debug page settings:

```json
{
  "profiles": {
    "ConsoleApp152": {
      "commandName": "Project",
      "remoteDebugEnabled": true,
      "remoteDebugMachine": "MyServer"
    }
  }
}
```


davkean:
I think there was a discussion at one point as to whether this was to be persisted in a user location vs launch profile, was that resolved?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfile.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
     /// <summary>
     /// Represents one launch profile read from the launchSettings file.
     /// </summary>
-    internal class LaunchProfile : ILaunchProfile, IPersistOption
+    internal class LaunchProfile : ILaunchProfile, ILaunchProfile2, IPersistOption
```
davkean: Its interesting that when we added native & SQL debuggiong support we didn't add a new version of ILaunchProfile. I don't have a strong opinion on this, just calling out that its different to the last few properties.

davkean:
Does this combine well with LaunchUrl?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.resx
```diff
@@ -252,4 +252,16 @@
   <data name="chkSqlCodeDebuggingText" xml:space="preserve">
     <value>Enable SQL Server debugging</value>
   </data>
+  <data name="chkRemoteDebugEnabledHelpText" xml:space="preserve">
+    <value>Enable remote debugging</value>
+  </data>
+  <data name="RemoteDebug" xml:space="preserve">
+    <value>Use remote m_achine</value>
```
melytc: Is the _ intended or is it a typo?

davkean: Intended, it indicates a _mnemonic_ character in WPF. In Windows Forms, & is the same indicator.

melytc: Thanks for the clarification 👍

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfile.cs
drewnoakes: I'm moving these to the 'other' properties group, removing the new interfaces.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.resx
drewnoakes: I copied the mnemonic from legacy. At the same time I noticed we don't have mnemonics for anything else on this property page. Filed #5797 to track that.

drewnoakes:
> Does this combine well with LaunchUrl?

The _UI_ works, but I don't know whether the feature makes sense for launching URLs in the web providers. The way the debug launch provider system is designed, new options (such as remote debug machine) are automatically included, so the `Microsoft.VisualStudio.ProjectSystem.Web` implementations of `ILaunchSettingsUIProvider` will need to opt-out of remote debugging support if they wish to hide this. I'll reach out and start that discussion.

davkean:
What happens if you put invalid characters in the host name?

drewnoakes:
> What happens if you put invalid characters in the host name?

Good idea, I'll add some validation of that too. Looks like `Uri.CheckHostName` might do the trick:

https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.Private.Uri/src/System/Uri.cs#L1250-L1303

drewnoakes:
> I wonder if this would look better if it was underneath Environment Variables

I considered this. The current placement matches legacy, so I went for consistency. If others think it'd look better beneath the environment variables I'll move it.

drewnoakes:
image

drewnoakes:
> I think there was a discussion at one point as to whether this was to be persisted in a user location vs launch profile, was that resolved?

No resolution about that, no. There are pros and cons to each, boiling down mostly to whether you add `launchSettings.json` to version control. It would be more work to split the config between two locations, which factored in to my decision. It's possible to split it out later if we so choose, and migrate values. For those reasons, the new interface versions and constant for this are `internal` rather than `public`, so we can change our minds later.

Longer term it might be better to store all these values at MSBuild properties and lose `launchSettings.json` altogether. We could default some to the user file (such as remote machine) and users could still have them in the project if they wanted to.

davidwengier:
> Longer term it might be better to store all these values at MSBuild properties and lose launchSettings.json altogether. We could default some to the user file (such as remote machine) and users could still have them in the project if they wanted to.

I like this idea, and we can include a property property to indicate where they should be stored, like the web server debugging configuration in ASP.NET projects.

davkean:
This requires thought, launchSettings.json supports _multiple_ profiles, legacy support only supports 1.

PR #5789, opened 2020-01-13
Display ManifestExplanationLabel as a label

melytc:
The ManifestExplanationLabel is the two-line text below the Icon and manifest radio button. Currently, it's a text box with the label appearance. This affects the accessibility of the UI: there is a tab stop in it (when it shouldn't since it's not interactable) and a screen reader might confuse it as a text input.

**Before:**
before

**After:**
after

By changing the text box to a label, it does not focus nor look intereractive. Also, the position and size remain the same.

Work item tracked in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1049278

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.Designer.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.Designer.vb
@@ -31,7 +31,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         Friend WithEvents TargetFrameworkLabel As System.Windows.Forms.Label
         Friend WithEvents AutoGenerateBindingRedirects As System.Windows.Forms.CheckBox
         Friend WithEvents overarchingLayoutPanel As System.Windows.Forms.TableLayoutPanel
-        Friend WithEvents ManifestExplanationLabel As System.Windows.Forms.TextBox
+        Friend WithEvents ManifestExplanationLabel As System.Windows.Forms.Label
         Friend WithEvents iconTableLayoutPanel As System.Windows.Forms.TableLayoutPanel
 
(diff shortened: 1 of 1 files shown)


davkean:
You might need to adjust the left margin by -3 to compensate for the textbox -> label. I think we do that for labels on other pages.

drewnoakes:
I don't know what this label is used for, but if there's any chance the user would want to copy that text (to paste into a search engine for example) then that will no longer be possible.

Would setting [`TabStop`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.tabstop?view=netframework-4.8) to `false` be a suitable alternative?

melytc:
> Would setting [`TabStop`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.tabstop?view=netframework-4.8) to `false` be a suitable alternative?

It is definitely an easier/straight-forward solution (I tested it and it works), but I think it will raise a semantic issue there: having a text box displayed as a label might mess up a screen reader.

melytc:
> I don't know what this label is used for, but if there's any chance the user would want to copy that text (to paste into a search engine for example) then that will no longer be possible.

However, this might've been why we decided in the past to make it a text box: there is no way to [select a Windows Forms label](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.canselect?view=netframework-4.8#remarks) 👎

melytc:
> You might need to adjust the left margin by -3 to compensate for the textbox -> label. I think we do that for labels on other pages.

I think it didn't change position nor size, should I still do this change? Judging by my sight, the whole block below the Icon and manifest radio button has already a left margin.

davkean:
Ah, I didn't realize it didn't already have a border.

This is what I'm referring to:

image

But I see this PR doesn't change the layout and this page was already broken, so just merge this in.

Issue #102, opened 2016-05-13
Project System: Open containing folder on items


davkean:
_From @davkean on March 24, 2016 17:45_

It would be great be able to right-click on a item in Solution Explorer, and choose Open containing folder.

It should work for items, including linked items, NuGet packages, binaries and project references.

_Copied from original issue: dotnet/roslyn#10071_

- [x] Support Open Folder in File Explorer for multiple items
- [x] Open Containing Folder for items
- [x] Copy Full Path for items
- [x] Add Open Folder in File Explorer for merged context menu (when you select folder and project)
- [x] Open Containing Folder for for merged context menu (when you select folder and file)
- [x] Copy Full Path doesn't work for project
- [x] Copy Full Path doesn't work for node under imports tree (https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/225973?_a=overview)

dsplaisted:
Similarly, it appears that the "Open Folder in Windows Explorer" command doesn't work for a linked folder.  For example, from dotnet/sdk:

```xml
    <Compile Include="..\Common\src\**\*.cs">
      <Link>Common\%(RecursiveDir)\%(FileName)%(Extension)</Link>
    </Compile>
```

Trying to open either the "Common" folder or the "ConflictResolution" folder (which is a subfolder of it) doesn't do anything.

This should also work if expressed via `LinkBase`, ie:

```xml
    <Compile Include="..\Common\src\**\*.cs" LinkBase="Common"/>
```

davkean:
First part of this PR is: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/216780.

davkean:
@dsplaisted Linked folders can be made up of multiple items from multiple locations, not just a single glob. The logic to determine this and enable the command in that case is non-trivial and probably not worth addressing. Would welcome a contribution if you want to fix this. The logic lives in DefaultCommandSet2KHandler in the CPS code base.

davkean:
The final two items are unblocked by:

https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/225296
https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/225297
https://github.com/dotnet/project-system/pull/5792

davkean:
This is now finished.

PR #5792, opened 2020-01-14
Mark ProjectImport's with FileSystemEntity

davkean:
Project imports were being marked with FileOnDisk, but not FileSystemEntity which is the base between FileOnDisk and Folder. This enables the "Copy Full Path" command but there is additional work needed to make the command actually execute which is being tracked by https://github.com/dotnet/project-system/issues/102.


davkean:
No, just past reverse engineering and usage.

davkean:
We can't merge this in until we fix the copy. Right now the behavior is that Copy Full Path just doesn't show up.

davkean:
This https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/225973?_a=overview unblocks this PR.

PR #5796, opened 2020-01-16
Update CPS version

drewnoakes:
Respond to annotation changes.

Needed to unblock #5775.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -76,6 +76,8 @@ protected override void Initialize()
             _broadcastBlock = DataflowBlockSlim.CreateBroadcastBlock<IProjectVersionedValue<DesignTimeInputsDelta>>(nameFormat: nameof(DesignTimeInputsChangeTracker) + "Broadcast {1}");
             _publicBlock = AllowSourceBlockCompletion ? _broadcastBlock : _broadcastBlock.SafePublicize();
 
+            Assumes.Present(_project.Services.ProjectAsynchronousTasks);
+
```
davkean: That these services can be null seems very wrong,, can we file a bug to fix this?

drewnoakes: Are you talking about `ProjectAsynchronousTasks` in particular, or all the props on `UnconfiguredProjectServices`?

davkean: Lots of these services should never be null, if they are - there stuff is severely broken.

drewnoakes: Could move the `Assumes.Present` calls into their getters. I seem to remember code that checked some of these for null though. I don't know enough about which should be non-null to make the call myself. It'll be easy enough to search usages and back these `Assumes` calls out in our code.

drewnoakes: #5799

drewnoakes: Looking at `UnconfiguredProjectServicesImpl` it seems very intentional that they are nullable.

PR #5800, opened 2020-01-16
Debug property page annotations

drewnoakes:
Follows #5783.

Part 2 of 2, in which the debug property page is annotated.

This one was hard to annotate and required a fair bit of refactoring. The PR is only really reviewable if you go commit-by-commit, in which case it should be straightforward,


davkean:
@drewnoakes What's blocking this from being merged in?

drewnoakes:
@davkean the unit tests are completely broken with these changes, due to use of Moq. To fix them and keep these changes requires producing an export provider that returns the right things.

I'm going to abandon this for now as I don't have much bandwidth at the moment.

PR #5802, opened 2020-01-16
Add help text to text boxes in Debug 

melytc:
The NVDA Narrator is not reading the place holder text for _Application arguments_ and _Working directory_ fields. To fix this accessibility bug, I added an `AutomationProperties`'s  HelpText to ensure that users who use screen readers can notice the suggestions made. 

image

**Before:**
image

**After:**
image

Item tracked on AzDO: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1049311


melytc:
@AbhitejJohn gave me a good feedback on this:
When there is content in the text boxes, it might be a good idea to clear the suggestion/help text, and just read whatever the text box contains. Just like what happens visually: the placeholder text disappears.

This is not blocking, but it's definitely a good practice we should follow.

Issue #2916, opened 2017-10-23
Add References Dialog shows referenced Shared Projects as Unselected


natidea:
From [VSFeedback](https://developercommunity.visualstudio.com/content/problem/130875/console-net-core-2-app-still-not-showing-reference.html) | [508624](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/508624)

Repro Steps:
- Create a .NET Core Console app and a Shared Project. 
- Reference the shared project from the .NET Core console app.
- Open Add References dialog again.

Expected
Shared project is checked

Actual 
Shared project is unchecked

![image](https://user-images.githubusercontent.com/7732033/31873724-e861756c-b779-11e7-855f-8d629e633f59.png)

Pilchie:
:link: also reported at https://developercommunity.visualstudio.com/content/problem/328798/cannot-remove-a-shared-project-reference-from-net.html

Pilchie:
Pulling into 16.0 as there have been a couple more VSFeedback reports about this.

etbyrd:
Probably a dupe of: https://github.com/dotnet/project-system/issues/1391

drewnoakes:
Closing as duplicate of #1391. I moved info from here to there.

davkean:
This was fixed in https://github.com/dotnet/project-system/pull/5805.

Issue #1391, opened 2017-01-31
After adding SharedProject to core project and opening Add Reference dialog again, SharedProject is not checked there


abpiskunov:
1. Create a core class library
2. add a SharedProject to solution
3. Add reference to SharedProject for class library
4. Wait for restore 
5. Open Add Reference dialog again 
 
Actual Result: SharedProject is not checked 
 
Expected: Shared project to be checked there in dialog

abpiskunov:
Per Mohit C. commets:

This mail doesn’t have the full description of the RM + Shared project issue; I am guessing you are describing a problem where an added Shared project reference is not showing up as checked in the dialog.

It’s the project systems that inform the Reference Manager about the items that are already referenced, RM does not read the project file directly.
You can set a BP here and that would tell you why the already referenced Shared projects are not being marked for this type of project.

https://devdiv.visualstudio.com/DevDiv/_git/VS?path=%2Fsrc%2Fvsproject%2FReferenceManager%2FProviders%2FProviderImpls%2FSharedProjectProvider%2FSharedProjectReferenceProvider.cs&version=GBlab%2Fd15prerel&_a=contents&line=260&lineStyle=plain&lineEnd=260&lineStartColumn=22&lineEndColumn=44

RaulPerez1:
Is this not related to #954?

abpiskunov:
no, #954 just displaying data and it is unrelated to add reference dialog

RaulPerez1:
It shouldn't have been, when I opened it everything (including this scenario) was busted. I think it's fine to have a separate bug if we have multiple root causes though, I was curious because the repro you have above is exactly what I was doing when I opened that one.

abpiskunov:
yes repros are the same, but code paths are completely different. #954 i am fixing right now and on my machine it is working already :)

abpiskunov:
Same happens when you Browse for assembly reference and add it to project , then re-open Add Reference dialog - check box for assembly would not be checked.

drewnoakes:
From [VSFeedback](https://developercommunity.visualstudio.com/content/problem/130875/console-net-core-2-app-still-not-showing-reference.html) | [508624](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/508624)

Repro Steps:
- Create a .NET Core Console app and a Shared Project. 
- Reference the shared project from the .NET Core console app.
- Open Add References dialog again.

Expected
Shared project is checked

Actual 
Shared project is unchecked

[...]

PR #5805, opened 2020-01-17
Register referenced "Shared Projects"

davkean:
We were not registering "Shared Projects" with IVsSharedMSBuildFilesManagerHierarchy, which is hidden behind the "SharedImports" capability (not to be confused with "SharedProjectReferences"), this led to:

- Context switcher was broken
- Reference Manager didn't think that any Shared Projects were referenced

This fixes:

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/508624
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/792775
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/972385
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/979565
https://developercommunity.visualstudio.com/content/problem/606555/the-code-editor-project-selector-switches-back-to.html
https://developercommunity.visualstudio.com/content/problem/130875/console-net-core-2-app-still-not-showing-reference.html
https://developercommunity.visualstudio.com/content/problem/328798/cannot-remove-a-shared-project-reference-from-net.html
https://developercommunity.visualstudio.com/content/problem/728261/-vshpropid7vshpropid-shareditemcontexthierarchy-is.html
https://github.com/dotnet/project-system/issues/1391
https://github.com/dotnet/project-system/issues/2916

This needs to be a fixed capability because CPS calls the registration once during IVsParentProject.OpenChildren and will never call it again.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs
@@ -29,11 +29,15 @@ namespace Microsoft.VisualStudio.Packaging
         /// <summary>
         ///     Represents C#'s set of capabilities that are always present.
         /// </summary>
-        public const string CSharp = Default + "; " + ProjectCapability.CSharp;
+        public const string CSharp = Default + "; " +
+                                     ProjectCapability.CSharp + "; " + 
+                                     ProjectCapabilities.SharedImports;
 
         /// <summary>
         ///     Represents Visual Basic's set of capabilities that are always present.
         /// </summary>
-        public const string VisualBasic = Default + "; " + ProjectCapability.VisualBasic;
+        public const string VisualBasic = Default + "; " + 
+                                          ProjectCapability.VisualBasic + "; " + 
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs
```diff
@@ -29,11 +29,15 @@ internal class ProjectTypeCapabilities
...
+                                     ProjectCapability.CSharp + "; " + 
+                                     ProjectCapabilities.SharedImports;
 
         /// <summary>
         ///     Represents Visual Basic's set of capabilities that are always present.
         /// </summary>
-        public const string VisualBasic = Default + "; " + ProjectCapability.VisualBasic;
+        public const string VisualBasic = Default + "; " + 
+                                          ProjectCapability.VisualBasic + "; " + 
+                                          ProjectCapabilities.SharedImports;
```
drewnoakes: This doesn't apply to F# projects?

davkean: F# doesn't support shared projects. It requires language support.

PR #5806, opened 2020-01-19
Use MIT license for nullable annotations

drewnoakes:
These definitions are taken from dotnet/runtime (originally from dotnet/corefx) which uses the MIT license.

### DIFF
--- a/THIRD-PARTY-NOTICES.txt
+++ b/THIRD-PARTY-NOTICES.txt
@@ -1,0 +1,39 @@
+.NET Core uses third-party libraries or other resources that may be
+distributed under licenses different than the .NET Core software.
+
+In the event that we accidentally failed to list a required notice, please
+bring it to our attention. Post an issue or email us:
+
+           dotnet@microsoft.com
+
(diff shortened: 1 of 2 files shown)


davkean:
@drewnoakes You'll also need to include a 3rd party notice: https://github.com/dotnet/project-system/blob/dev15.9.x/THIRD-PARTY-NOTICES.md.

drewnoakes:
@davkean I've added `THIRD-PARTY-NOTICES.txt`.

THIRD-PARTY-NOTICES.txt
```diff
@@ -0,0 +1,39 @@
...
+
+The attached notices are provided for information only.
+
+
+License notice for dotnet/runtime
+---------------------------------
+
+https://github.com/dotnet/runtime/blob/master/LICENSE.TXT
+
+he MIT License (MIT)
```
davkean:
```suggestion
The MIT License (MIT)
```

drewnoakes:
We discussed this internally and agreed this was the right approach for now.

PR #5807, opened 2020-01-20
Add 'Nullable' option to the build property page

### LINKED ISSUE #4058: Add support for Nullable build setting

drewnoakes:
Fixes #4058.

This option is only available when the language version is 8.0 or greater.

---

### netcoreapp3.1 (LangVersion 8)

image

The dropdown has the following options:

image

---

### net48 (LangVersion 7.3)

The nullable control is disabled:

image

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/Resources/Microsoft.VisualStudio.Editors.Designer.resx
+++ b/src/Microsoft.VisualStudio.Editors/Resources/Microsoft.VisualStudio.Editors.Designer.resx
@@ -1970,6 +1970,18 @@ Do you want to update the value in the .settings file?</value>
   <data name="PPG_AdvancedBuildSettings_DebugInfo_Portable" xml:space="preserve">
     <value>Portable</value>
   </data>
+  <data name="PPG_BuildSettings_Nullable_Disable" xml:space="preserve">
+    <value>Disable</value>
+  </data>
+  <data name="PPG_BuildSettings_Nullable_Enable" xml:space="preserve">
+    <value>Enable</value>
+  </data>
+  <data name="PPG_BuildSettings_Nullable_Warnings" xml:space="preserve">
+    <value>Warnings</value>
+  </data>
+  <data name="PPG_BuildSettings_Nullable_Annotations" xml:space="preserve">
+    <value>Annotations</value>
+  </data>
   <data name="BlockedResx" xml:space="preserve">
     <value>Unable to load '{0}' because it is not trusted.</value>
     <comment>{0} is resx file name that is blocked</comment>
--- a/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.Designer.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.Designer.vb
@@ -94,6 +94,8 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Me.errorsAndWarningsLabel = New System.Windows.Forms.Label()
             Me.lblSGenOption = New System.Windows.Forms.Label()
             Me.cboSGenOption = New System.Windows.Forms.ComboBox()
+            Me.lblNullable = New System.Windows.Forms.Label()
+            Me.cboNullable = New System.Windows.Forms.ComboBox()
             Me.overarchingTableLayoutPanel.SuspendLayout()
             Me.outputTableLayoutPanel.SuspendLayout()
             Me.generalHeaderTableLayoutPanel.SuspendLayout()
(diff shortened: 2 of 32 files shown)


davkean:
I think we should consider coming up with a way for this property to visible but disabled when changing the TFM will cause it to be enabled.

- What happens if I choose the Application tab, change from NET Core 3.0 and then choose this tab?

davkean:
This is the help for that page: https://docs.microsoft.com/en-us/visualstudio/ide/reference/build-page-project-designer-csharp?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev16.query%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(cs.ProjectPropertiesBuild)%26rd%3Dtrue&view=vs-2019. We should document it there.

drewnoakes:
Changing it to disable rather than hide is straightforward to do, but I felt it would make the UI more confusing for users.

Currently the state of this UI is not updated based on changes to TFM (via either UI or file on disk). The page needs to be reloaded. I looked at `PropertyControlData` for hints on how to respond to such changes but didn't see anything.

We could justify sticking help links on every UI element on the page. Does this one deserve one just because it's new?

Once the UI here is finalised I'll update the docs.

davkean:
We don't hide elements in other cases where changing the TFM would enable it - for example, Prefer 32-bit. Can you see how it reacts when you change from 4.0 -> 4.5 in legacy?

davidwengier:
> Does this one deserve one just because it's new?

I think this one deserves something because "Enable", "Warnings" and "Annotations" don't mean much on their own.

drewnoakes:
Updated to respond to changes to `LangVersion`/`TargetFramework`, either through the UI or project file.

Also changed to disable rather than hide the UI.

drewnoakes:
> I think this one deserves something because "Enable", "Warnings" and "Annotations" don't mean much on their own.

I think you could argue this about any of the options you weren't familiar with. For example, what does the `TRACE` constant mean? What's a constant? What does warning level '4' mean?

The property pages support <kbd>F1</kbd> help, though it's not very discoverable and is currently broken (I filed #5815).

drewnoakes:
Documentation updated in https://github.com/MicrosoftDocs/visualstudio-docs-pr/pull/6024.

src/Microsoft.VisualStudio.Editors/Resources/Microsoft.VisualStudio.Editors.Designer.resx
```diff
@@ -1970,6 +1970,18 @@ Do you want to update the value in the .settings file?</value>
...
+  </data>
+  <data name="PPG_BuildSettings_Nullable_Enable" xml:space="preserve">
+    <value>Enable</value>
+  </data>
+  <data name="PPG_BuildSettings_Nullable_Warnings" xml:space="preserve">
+    <value>Warnings</value>
+  </data>
+  <data name="PPG_BuildSettings_Nullable_Annotations" xml:space="preserve">
+    <value>Annotations</value>
+  </data>
```
tmeschter: By placing these values in .resx file we're choosing to localize them. Is that what you want, or would it make more sense to simply use the literal values understood by the compiler?

In the case of the latter you would hard-code them in the source and use the lower-case forms.

davkean:
The "Advanced Build Options" page is also configuration-dependent. I think you can argue this for lots of options on that page, unsafe code, Prefer 32-bit, etc.

src/Microsoft.VisualStudio.Editors/Resources/Microsoft.VisualStudio.Editors.Designer.resx
davkean: The drop down should be localized, what we persist in the project file shouldn't be.

drewnoakes: I considered using the compiler's strings and ended up following the example set by the advanced build properties page (where I initially added this new UI before moving it to the regular build properties page).

image

image

davkean: Yes, and these were localized in response to a bug.

drewnoakes:
@tmeschter 

> If you want it to be configuration-independent, the best option is the Advanced Build Options dialog.

To be clear, in this PR the setting _behaves_ in a configuration-independent manner due to `HasConfigurationCondition="False"` on its data source.

The question is whether this behaviour is surprising to the user or not.

@davkean @davidwengier and I discussed placement on Monday and decided the setting was not really 'advanced', and that it deserved a more prominent position in the UI. So I moved it from the advanced build prop page to the regular build prop page.

But looking at it now I see that every other property on that page is actually configuration-dependent.

I agree that nullable shouldn't be configuration-dependent. It's a shame to hide the option further down, but on balance I think you're right so will move it back 🤦‍♂

drewnoakes:
_Actually_ it turns out that every setting on the _advanced_ property page is configuration dependent too. So now I'm less sure that moving it does anything useful.

When language version was editable, it would have been the exception on all these pages.

tmeschter:
Sorry, I missed `HasConfigurationCondition="False"` on the data source--I'm so used to thinking about these pages through the lens of the old project system. :-) I agree that with that issue out of the way, the Build page is a much better choice than the Advanced dialog. Approved.

PR #5808, opened 2020-01-20
Update sln changes to shared projects

drewnoakes:
CPS recently changed to register shared projects differently. That seems to have resulted in VS adding these lines to the sln file.


drewnoakes:
@davkean VS master is insisting on this change and I assume it follows your recent changes. Please take a look.

davkean:
This is intentional, merging in.

PR #5810, opened 2020-01-20
Create GitHub Action to post PRs to teams

davidwengier:
Trying this out, will tweak content over time. Should be more reliable than the twitter app we use now.


davidwengier:
Looks like summary only shows in the activity view

PR #5816, opened 2020-01-21
Mark returns as [MaybeNull]

davkean:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
```diff
@@ -36,7 +37,8 @@ public static T GetProperty<T>(this IVsHierarchy hierarchy, VsHierarchyPropID pr
         /// <summary>
         ///     Gets the value of the specified property of the specified item if the hierarchy supports it, or throws an exception if there was an error.
         /// </summary>
-        public static T GetProperty<T>(this IVsHierarchy hierarchy, HierarchyId item, VsHierarchyPropID property, [MaybeNull]T defaultValue = default)
+        [return: MaybeNull]
```
drewnoakes: I think this is `[NotNullIfNotNull(parameterName: "defaultValue")]`. Ditto above.

drewnoakes: (based on the assumption that `Verify.HResult` throws if not `OK`)

davkean: I don't think so, this will return null if the project system says its supports a property but returns null.

drewnoakes: Oh, if resultObject below is null?

davkean: Yes.

sharwell: My interpretation matches @davkean for the information available

PR #5821, opened 2020-01-21
Remove LastFileWriteTime

davkean:
We shouldn't be using local time to determine file write times, as clocks can change while VS is open.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -354,7 +354,7 @@ protected async Task<bool> SettingsFileHasChangedAsync()
         {
             string fileName = await GetLaunchSettingsFilePathAsync();
 
-            return !_fileSystem.FileExists(fileName) || _fileSystem.LastFileWriteTime(fileName) != LastSettingsFileSyncTime;
+            return !_fileSystem.FileExists(fileName) || _fileSystem.LastFileWriteTimeUtc(fileName) != LastSettingsFileSyncTime;
```
drewnoakes: Consider renaming `LastSettingsFileSyncTime` to `LastSettingsFileSyncTimeUtc` for clarity.

PR #5822, opened 2020-01-22
Respect "RestoreProjectStyle"

davkean:
Certain projects in http://github.com/dotnet/runtime "turn off" package restore via <RestoreProjectStyle>Unknown</RestoreProjectStyle>: https://github.com/dotnet/runtime/blob/86bfd8d46b3817e2a027b23d829cd0d035cd8ad6/eng/restore/repoRestore.targets#L4.

This causes us to nominate to NuGet, only for it to fail to produce an assets file. This causes us to never complete operation progress resulting in https://developercommunity.visualstudio.com/content/problem/884350/intellisense-waiting-forever-dialog-in-dotnetrunti.html.

Turn off PackageReferences support if is not PackageReference based.

This works quite well, and can be changed dynamically and does the right thing.


stephentoub:
Thank you for fixing it.

PR #5824, opened 2020-01-22
Handle "quotes", and show more info for issues

davidwengier:
Apparently there are things called "quotes" which can really interrupt your { json } which doesn't help when you're using [json] to pass strings to dotnet console apps via [ { script-type: "yaml" } ].

Or so i've heard.


davidwengier:
Looks like I can get this to work with single quotes, or double quotes, but not both :(

Might have to write my own action

Issue #4782, opened 2019-05-02
Analyzers & AdditionalFiles should be added during evaluation


davkean:
We only add Analyzers and AdditionalFiles during design-time build. For user-perceived performance reasons there are no reasons we can't add them during evaluation. This tracks that.

drewnoakes:
How can evaluation detect analyzers that ship in packages?

PR #5830, opened 2020-01-24
Don't require analyzers to be in evaluation data

### LINKED ISSUE #5418: Analyzer dependencies showing as unresolved full paths
Steps to reproduce:

1. Add a package reference to `xunit.analyzers`
1. Build the solution
1. Close the solution
1. Re-open the solution
1. Expand the project's dependencies node
1. Notice there is no `Analyzers` node
1. Modify whitespace in the project file and save it (triggering a design-time build)
1. Notice an `Analyzers` node appears with unresolved child, which is a full path rather than the analyzer's name
   image
1. Close and reopen the solution and the analyzers appear correctly
1. Modify the project file again (stays resolved)
1. Close and re-open the solution (stays resolved)
1. Build from VS then close/re-open solution and notice `Analyzers` node is missing again

Open questions:

- Why is the `Analyzers` node not initially present? This seems to happen if the build before opening the solution happened within VS.
- Why are analyzers appearing as paths and in unresolved state?

drewnoakes:
Fixes #5418.

The dependency node receives both evaluation (unresolved) data and design-time build (resolved) data. Both sources provide data about the same items and must be correlated and combined.

A while ago the dependency node changed to disallow adding items to the tree that appear _only_ in design-time build results. This broke the Analyzers node as analyzers are not available during evaluation (see #4782).

This change allows specific rules to opt out of the restriction, doing so only for analyzer items.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/AnalyzerRuleHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Subscriptions/RuleHandlers/AnalyzerRuleHandler.cs
@@ -34,6 +34,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscription
 
         public override ImageMoniker ImplicitIcon => ManagedImageMonikers.CodeInformationPrivate;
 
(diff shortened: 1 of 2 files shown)


davkean:
Analyzers can come from evaluation, via the `<Analyzer>` item - how does this treat them?

drewnoakes:
Analyzers that come from `Analyzer` project items in evaluation will still work just fine.

Issue #5847, opened 2020-01-31
Global.json should trigger reevaluation & design-time build


davkean:
Similar to https://github.com/dotnet/project-system/issues/5846, we should trigger reevaluation/design-time build based on a change/add/removal of global.json. 

I think this would be pretty easy if we can mimic the global.json lookup via msbuild-isms, and just have it appear in the `<AdditionalDesignTimeBuildInput>`. CPS will automatically monitor it and reevaluate/design-time build passed on changes to it. This is how we trigger reevaluation on project.assets.json. 

The lookup is here, which is starting at project/solution directory and keep walking until you hit a global.json: https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/installer/corehost/cli/fxr/sdk_resolver.cpp.

drewnoakes:
Where would this fix live? In our DT targets?

davkean:
Yeah in the DT targets - we should hide it behind a indicator that they are using the .NET SDK, what I don't know yet.

PR #5849, opened 2020-01-31
Reevaluate when global.json changes

davkean:
Fixes: https://github.com/dotnet/project-system/issues/5847

CPS monitors and reevaluates when items added to `<AdditionalDesignTimeBuildInput>` are added, removed or deleted. Start monitoring global.json files starting from the solution directory and higher so that we can react to changes to it that affect MSBuild SDK resolution.

Note, this over-eagerly finds global.json files that may not influence the project, ie Given a solution structure like so:

```
C:\
C:\Folder\
C:\Folder\Solution\
C:\Folder\Solution\global.json
```
This will monitor for changes in `C:\`, `C:\Folder` and `C:\Folder\Solution` despite only `C:\Folder\Solution\global.json` ever infuencing SDK resolution. This is a limitation in MSBuild; there's no way to batch or filter the items inside evaluation and outside of a target, which is how `<AdditionalDesignTimeBuildInput>` items need to be produced.


davkean:
tag @nguerrera @lifengl @dsplaisted as an FYI and to revel this magic MSBuild code that @drewnoakes and I came up with.

PR #5851, opened 2020-01-31
Up-to-date check considers AdditionalDependentFileTime items

### LINKED ISSUE #5846: Fast up-to-date check doesn't consider global.json
Changes to `global.json` (such as the SDK version to use) should trigger a build.

Repro steps:

1. `dotnet new globaljson`
1. Build in VS twice. Second build should show 'up to date'.
1. Change the SDK version in `global.json`
1. Build in VS again. Should **not** be considered up-to-date, but it is.

drewnoakes:
Fixes #5846.

Builds on top of #5849.

This change teaches the fast up-to-date check to also consider `IProjectSnapshot.AdditionalDependentFileTimes`. This required a new subscription to `ProjectSource`, and plumbing the values through the state object.

Also, this introduces a class of input that is "not required". Specifically, potential paths to files which are probed for are not required, as those paths are speculative catch-alls. Previously a missing input would cause the project to be considered out-of-date. That behaviour is preserved for all except these new `AdditionalDependentFileTimes` items.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -93,6 +93,24 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             /// </summary>
             public ImmutableHashSet<string> CopyReferenceInputs { get; }
 
+            /// <summary>
+            /// Contains files such as:
+            /// <list type="bullet">
+            ///   <item>Potential <c>.editorconfig</c> paths</item>
+            ///   <item>Potential <c>global.json</c> paths</item>
+            ///   <item>Potential <c>Directory.Build.props</c> paths</item>
+            ///   <item>Potential <c>Directory.Build.targets</c> paths</item>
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -93,6 +93,24 @@ internal sealed class State
             /// </summary>
             public ImmutableHashSet<string> CopyReferenceInputs { get; }
 
+            /// <summary>
+            /// Contains files such as:
+            /// <list type="bullet">
+            ///   <item>Potential <c>.editorconfig</c> paths</item>
+            ///   <item>Potential <c>global.json</c> paths</item>
+            ///   <item>Potential <c>Directory.Build.props</c> paths</item>
```
davkean: Directory.Build.props/targets are not yet in this.

davkean:
Drew the bug associated with this: https://github.com/dotnet/project-system/issues/5846 is currently in 16.6, if you merge this in before the cut off can you move it back to 16.5 to correctly represent the version?

davkean:
@drewnoakes I've rebased this to dev16.5.x, it needs M2 approval.

PR #5852, opened 2020-01-31
Implement the DebugInfo property

davidwengier:
Return a value for DebugInfo to unblock unit test profiling.

Full support for the rest of these is #3622


davkean:
Oh mb, I accidentally merged this in insteaed of approving.

davidwengier:
Same end result :)

PR #5854, opened 2020-01-31
Add an "Edit Project File" command for VS Online scenarios

tmeschter:
This change adds a basic "Edit Project File" command to the context menu when right-clicking on a project node in Solution Explorer in VS Online scenarios. The project file then opens in the XML editor.

The primary purpose of this change is to unblock dogfood scenarios in the immediate future and it is *not* an optimal implementation of this feature. Notably, it has the following deficiencies:

1. Ideally we would re-use the existing "Edit Project File" command and place it on the context menu rather than creating a whole new command that does the exact same thing. However, much still needs to be figured out about the setup authoring for the VS Online client and that will undoubtedly lead to many other components being refactored. When that happens I don't want to run the risk of the existing "Edit Project File" command disappearing from the client and breaking this feature--better we have two copies of this command on the client than none.
2. The logic for identifying the project nodes is sketchy. Ideally we would get the project's GUID from the node and run that through the new Project API to check if it is indeed a managed project _and_ has the appropriate capabilities. However, getting the GUID is not straightforward. For the time being we must be content to check the file extension.
3. This doesn't necessarily play nice with the existing code we have for editing the project file. There may be bugs we haven't run into yet.
4. This change opens the project file, but due to https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1040697 the file will be read-only. There's nothing we can do about that at the moment so I'm not going to hold the change.

### DIFF
--- a/ProjectSystem.sln
+++ b/ProjectSystem.sln
@@ -41,6 +41,10 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "docs", "docs\docs.shproj", 
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Microsoft.VisualStudio.ProjectSystem.Managed.TestServices", "tests\Microsoft.VisualStudio.ProjectSystem.Managed.TestServices\Microsoft.VisualStudio.ProjectSystem.Managed.TestServices.shproj", "{02578366-DFA9-4827-93F7-08E2AE5CE6A4}"
 EndProject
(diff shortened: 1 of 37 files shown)


tmeschter:
Please don't merge this to master yet; I want to wait until we've wrapped up 16.5.

davkean:
EditProjectFile lives in CPS, why isn't a contribution to CPS?

davkean:
Can we file bugs to track making this feature real? I raised some concerns in https://github.com/dotnet/project-system/pull/5701 that weren't addressed in this and this feature does not belong in our repro. We don't implement "Edit Project File".

Issue #3792, opened 2018-07-26
Is "Check for arithmetic overflow/underflow" option name valid?


stjeong:
I think, the option name has to be changed to "Check for arithmetic overflow".

Because, "underflow" means "Arithmetic_underflow", so by the definition as follows,

> http://en.wikipedia.org/wiki/Arithmetic_underflow:
> 
> The condition in a computer program that can occur when the true result of a floating point operation is smaller in magnitude (that is, closer to zero) than the smallest value representable as a normal floating point number in the target datatype.
> 

Actually, MSIL doesn't support any IL code for underflow, see documents.

OpCodes Class (https://msdn.microsoft.com/en-us/library/system.reflection.emit.opcodes(v=vs.110).aspx)

Pilchie:
It's possible that it could be changed, but I'm not sure it's worth it - this string has remained unchanged for many years now.

PR #5855, opened 2020-02-01
Remove "underflow" from /checked description

davkean:
Fixes: https://github.com/dotnet/project-system/issues/3792.

I'll wait for 16.6 before I merge this in.


drewnoakes:
Why remove this? The option does check for both overflow and underflow:

https://docs.microsoft.com/en-us/dotnet/api/vslangproj80.csharpprojectconfigurationproperties3.checkforoverflowunderflow?view=visualstudiosdk-2017

Issue #1419, opened 2017-02-01
Invariant Culture shows up in Assembly Neutral Language list


davkean:
Setting it does nothing. This is basically the same as "(None)".

![image](https://cloud.githubusercontent.com/assets/1103906/22503574/827ab6a0-e8c6-11e6-8d68-a38bfebb481c.png)

wolf99:
Is it the same? I don't know the inner workings but on face value "None" would seem to mean "no setting", whereas "Invariant" would mean set to be insensitive to culture (i.e. a setting to this effect would be added).

I found this issue precisely because I was searching for information on using the Invariant option as my resources are XSD schema files that are only intended to be used with the invariant culture. That is, they do not have no culture *specified* ("none"), but instead should be recognized as being insensitive to culture.

sharwell:
@davkean Can you clarify which of the following is the bug? That way if someone beats us to a PR they can get it right. 😄 

1. The bug is the current behavior is correct and the setting is useless; resolve by not showing the item
2. The bug is the current behavior fails to properly apply the value when a user picks "Invariant Language"; resolve by keeping the item and fixing the way it gets used

PR #5857, opened 2020-02-01
Filter out invariant culture from languages combo

davkean:
Fixes: https://github.com/dotnet/project-system/issues/1419

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoUtils.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoUtils.vb
@@ -165,7 +165,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                                                                    OrderBy(Function(language) language.DisplayName)
                     For Each Culture As CultureInfo In AllCultures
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/AssemblyInfoUtils.vb
```diff
@@ -164,7 +164,12 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                     Dim AllCultures As IEnumerable(Of CultureInfo) = CultureInfo.GetCultures(CultureTypes.NeutralCultures Or CultureTypes.SpecificCultures Or CultureTypes.InstalledWin32Cultures).
                                                                    OrderBy(Function(language) language.DisplayName)
                     For Each Culture As CultureInfo In AllCultures
-                        NeutralLanguageComboBox.Items.Add(Culture.DisplayName)
+
+                        ' Exclude Invariant as its the same as "None" above
```
drewnoakes:
```suggestion
                        ' Exclude Invariant as it's the same as "None" above
```

PR #5867, opened 2020-02-04
Add help text to Executable path watermark.

melytc:
Found another text box without a help text, follows #5802.
Note: no localization needed.

Before:
image

After:
image


tmeschter:
Do we already do the right thing for "Application arguments" and "Working directory"?

melytc:
> Do we already do the right thing for "Application arguments" and "Working directory"?

Yes! That was done in the previous PR #5802.

Issue #5317, opened 2019-08-18
Opt-out of Partial Load Mode for the App Designer


davidwengier:
Currently the App Designer (project properties) is opted in to being deferred for Partial Load Mode, because the Resources editor that it hosts needs to be deferred. If it is opted out now the Resources tab errors with the following:

```
An error occurred trying to load the page.
Member not found. (Exception from HRESULT: 0x80020003 (DISP_E_MEMBERNOTFOUND))


[SHOWN IN DEBUG ONLY] STACK TRACE:
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at Microsoft.VisualStudio.ErrorHandler.ThrowOnFailure(Int32 hr)
   at Microsoft.VisualStudio.Editors.ApplicationDesigner.ApplicationDesignerPanel.CreateDesigner()
   at Microsoft.VisualStudio.Editors.ApplicationDesigner.ApplicationDesignerPanel.ShowDesigner(Boolean Show)
   at Microsoft.VisualStudio.Editors.ApplicationDesigner.ApplicationDesignerView.ShowTab(Int32 Index, Boolean ForceShow, Boolean ForceActivate)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)
```

We should see if we can change the code here so that the Resources tab shows the same waiting UI as is shown if the user opens the Resources file themselves:
![image](https://user-images.githubusercontent.com/754264/63232018-e5d67280-c266-11e9-9c52-8c51b111c14c.png)

davkean:
Triage: This is figuring out what pages don't actually need design-time build results and stop blocking opening the AppDesigner while PLM/operation progress is in progress, and block pages that actually need operation progress.

davidwengier:
Random notes found while debugging other:

* We need to opt both the ApplicationDesignerEditorFactory and PropPageDesignerEditorFactory out of PLM
* We need to have our own flag in the individual designer tabs to flag whether they need to wait for intellisense
* In the AppDesignerPanel.CreateDesigner method (or similar) we would check that flag to go down one of two paths:
  * If the flag is false, then continue as normal
  * If the flag is true, it means wait for intellisense, which means the CreateDesigner method should create the PartialLoadWindowFrame somehow.
[...]

PR #5871, opened 2020-02-05
Opt the PropPageDesignerEditorFactory out of PLM

### LINKED ISSUE #5858: [Regression] Debug properties pages are blank the first time you open them
**Visual Studio Version**:
Version 16.5.0 Preview 3.0 [29802.0.master]

**Summary**:

**Steps to Reproduce**:

1. Create a new .NET Core project
2. Choose Debug -> [Project] Properties

**Expected Behavior**:
Debug property page to show

**Actual Behavior**:
The the page opens but it is blank:
image

I suspect this is related to PLM being enabled.

davidwengier:
Fixes #5858

This is a short term solution to fix the regression for 16.5 where the Debug -> [Project] Properties menu item shows a blank screen. The real fix for this will be #5317.

Whats happening is that the Application tab initializes and then switches to the Debug page, but when the OutputType property is read to initialize the Application table it sets some data flowing through the system including a version-only update of ProjectRuleSource, which gets to the WorkspaceContext, and that triggers operation progress. It happens just in time for the Debug editor factory to be asked to initialize, which sees that operation progress has started, so gets the PLM window frame, which it can't deal with, and bombs out.

The change itself should be safe enough because the only thing that requests the PropPageDesignerEditorFactory is the ApplicationDesignerEditorFactory and that is opted-in so it will wait for the initial Intellisense stage to finish because any PropPageDesignerEditorFactory is requested.

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
+++ b/src/Microsoft.VisualStudio.Editors/package/VBPackage.vb
@@ -36,7 +36,7 @@ Namespace Microsoft.VisualStudio.Editors
     ProvideMenuResource("Menus.ctmenu", 30),
     ProvideEditorFactory(GetType(ApplicationDesigner.ApplicationDesignerEditorFactory), 1300, True, TrustLevel:=__VSEDITORTRUSTLEVEL.ETL_AlwaysTrusted),
     ProvideEditorFactory(GetType(SettingsDesigner.SettingsDesignerEditorFactory), 1200, True, TrustLevel:=__VSEDITORTRUSTLEVEL.ETL_AlwaysTrusted, CommonPhysicalViewAttributes:=3),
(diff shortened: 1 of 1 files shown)


davkean:
I'll buddy test this and see if resolves my problem - I'm not getting the PLM prompt at all, and even then it never exits it.

davidwengier:
I should clarify, when it says "gets the PLM view", I mean "gets the PLM windowframe, which is doesn't know how to deal with, and crashes out", so you don't actually see the PLM frame in the debug pane, just the blank screen.

But yes, please test it. I managed to turn the feature flag on so I could repro it in the end, and validated on my laptop which never had the flags, but who knows what other odd feature flag state I have here.

davkean:
I've tested this and it looks good, we need this for 16.5.

Issue #2745, opened 2017-08-28
Can't remove Any CPU from Project Platforms list


davidmatson:
I'm using VS 2017 15.3.2.

With this csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net47</TargetFramework>
  </PropertyGroup>
</Project>
```

If I add an x64 or x86 configuration (Build, Configuration Manager...), that works fine and edits the proj like this:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net47</TargetFramework>
    <Platforms>AnyCPU;x86;x64</Platforms>
  </PropertyGroup>
</Project>
```

I can remove either the x64 or x86 configuration from the project, but not the Any CPU one. If I try to do so, I get the following error:
```
---------------------------
Microsoft Visual Studio
---------------------------
The operation could not be completed. The parameter is incorrect. 
---------------------------
OK   
---------------------------
```

![image](https://user-images.githubusercontent.com/3360277/29790817-c709e4d6-8bef-11e7-80db-6acf305665bc.png)

amanbha:
Is there a timeline for the fix?

stijnherreman:
As a workaround you can remove it by editing the csproj file by hand.

drewnoakes:
This no longer shows an error dialog, but it also doesn't work as expected. It is removed from the UI, but not the project file. Re-open the UI and the platform is back in the dropdown list.

ocallesp:
PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/228032

PR #5876, opened 2020-02-07
remove extra spaces in project platforms configuration

ocallesp:
this fixes the case when 'Any CPU' instead of 'AnyCPU'

https://github.com/dotnet/project-system/issues/2745

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
@@ -141,7 +141,7 @@ namespace Microsoft.VisualStudio.Build
             bool valueFound = false;
             foreach (string value in GetPropertyValues(evaluatedPropertyValue, delimiter))
             {
-                if (!string.Equals(value, valueToRemove, StringComparisons.PropertyValues))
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs
```diff
@@ -141,7 +141,7 @@ public static void RemovePropertyValue(ProjectRootElement project, string evalua
             bool valueFound = false;
             foreach (string value in GetPropertyValues(evaluatedPropertyValue, delimiter))
             {
-                if (!string.Equals(value, valueToRemove, StringComparisons.PropertyValues))
```
davidwengier: This method looks like a generic helper method for any property. Are you sure its only called in places where spaces aren't important? Or are spaces not supported at all by MSBuild in this scenario?

Is it worth adding a note to the XML comments that says it is space-agnostic?

davkean: This will break configurations `<Configuration>` with spaces in their names "Release (Production)".

I think any fix around this belongs in CPS - it already handles the mapping between "Any CPU" and "AnyCPU", see ProjectConfigProviderBase.ConvertVsConfigPlatformToProjectConfigPlatform usage. It should have already fixed this, if not, then that's the bug.

PR #5877, opened 2020-02-07
Remote debug machine find dialog and authentication option

### LINKED ISSUE #5860: [CPS] Allow "No Auth" property in Project settings
This is the CPS equivalent of https://github.com/dotnet/project-system/issues/3053.

davidwengier:
Fixes #5860

This was going to just be adding the Find button, but the find button selects machine and auth so it didn't make sense to do one without the other, so apologies about the PR, but its actually not too big I hope.

There is a reasonable argument that the auth provider idea is over-designed but I was thinking about WinUI at the time and it seems like a good idea to toggle auth providers by capability. If its too much, or if the speculative WinUI provider should be removed, just let me know.

Reviewing commit-by-commit should hopefully make sense and see the different parts of this (and ignore the strings and xlf files :))

Find dialog looks like this:
image


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
+{
+    /// <summary>
+    /// Implementation of IRemoteAuthenticationProvider for the "Universal" authentication type.
+    /// </summary>
+    [Export(typeof(IRemoteAuthenticationProvider))]
+    [AppliesTo("WinUI")] // TODO: ???
+    internal class UniversalRemoteAuthenticationProvider : IRemoteAuthenticationProvider
```
davidwengier: This can be removed before merge, its speculative, but illustrates why IRemoteAuthenticationProvider exists

davkean: What is it about "universal" that causes us to supplier this port supplier? What does a port supplier actually provide? Is it go to do with the fact that these apps are in an AppContainer? 

(Trying to understand when/when not we would turn this on)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -880,6 +979,16 @@ private async Task OnLaunchSettingsChanged(ILaunchSettings profiles)
             InitializeDebugTargetsCore(profiles);
         }
 
+        private void InitializeAuthenticationProviders()
+        {
+            _authenticationProviders = new OrderPrecedenceImportCollection<IRemoteAuthenticationProvider>(projectCapabilityCheckProvider: Project);
+            IEnumerable<Lazy<IRemoteAuthenticationProvider, IOrderPrecedenceMetadataView>> providers = Project.Services.ExportProvider.GetExports<IRemoteAuthenticationProvider, IOrderPrecedenceMetadataView>();
```
davkean: These are already filtered at this point, there's no point in putting these in a OrderPrecedenceImportCollection as GetExports already does this.

davidwengier: Thanks, I suspected as much from debugging, but followed the patter in the LaunchType list elsewhere in this class. Will update both.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davkean: Oh this is an authentication mode?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
davkean: The won't be sorted though, so you will have to do this. I'd consider pulling these objects out into a single component that does all this for you so just pull on one object and that class has all the collections.

davkean:
Does unchecking remote machine, disable authentication mode? It should.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        [ImportingConstructor]
+        public UniversalRemoteAuthenticationProvider(UnconfiguredProject _) // force MEF scope
```
davkean: These don't need to be scoped to a project - they contain no state, see IItemTypeGuidProvider implementors for an example.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: Yes, this is providing "Universal Auth". It's used by UWP and Desktop Bridge.

The "when" question in my mind would be answered by whatever decides to add the capability, and I was 50/50 on whether the capability for this should be "WinUI" or "SupportsUniversalAuth", but in light of recent news I stopped thinking about it and instead I figure I'll remove it before I merge anyway so it doesn't matter :)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -534,6 +611,27 @@ public ICommand BrowseDirectoryCommand
...
+        {
+            get
+            {
+                return LazyInitializer.EnsureInitialized(ref _findRemoteMachineCommand, () =>
+                new DelegateCommand(state =>
+                {
+                    SVsServiceProvider serviceProvider = Project.Services.ExportProvider.GetExportedValue<SVsServiceProvider>();
+
+                    // HACK: Using VsUIService directly here avoids using a banned UI (serviceProvider.GetService) and can be replaced easily when MEF is hooked up to this page
+                    var vsService = new VsUIService<SVsDebugRemoteDiscoveryUI, IVsDebugRemoteDiscoveryUI>(serviceProvider, _projectThreadingService.JoinableTaskContext.Context);
```
davkean: Can we pull all this logic (getting authentication providers, prompting for machines, etc) into a class and just have this class delegate to it?

davidwengier:
> Does unchecking remote machine, disable authentication mode? It should.

Yes, the checkbox controls the machine name text box, find button and auth combo box. In addition the auth box will be disabled if there aren't any providers for it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -265,6 +270,62 @@ public string RemoteDebugMachine
...
+        }
+
+        private IRemoteAuthenticationProvider GetProviderForPortSupplier(Guid portSupplier)
+        {
+            IRemoteAuthenticationProvider provider = _authenticationProviders.FirstOrDefaultValue(p => p.PortSupplier.Equals(portSupplier));
+
+            // If there is a bad guid in the launch settings we don't want to clobber it by having a default selection, so create a temporary provider
+            if (provider == null)
+            {
+                provider = _unknownProvider = new UnknownRemoteAuthenticationProvider(portSupplier);
```
davkean: If I'm reading this corectly unknownProvider can either represent the thing the user just chose from the dialog if we don't recognize it, or the thing stored in launchSettings.json if we don't recognize it. Is there a way without needing state to store "unknown provider"?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -265,6 +270,62 @@ public string RemoteDebugMachine
...
+            else if (_unknownProvider != null)
+            {
+                _unknownProvider = null;
+                OnPropertyChanged(nameof(RemoteAuthenticationProviders));
+            }
+
+            return provider;
+        }
+
+        public bool HasAuthenticationProviders
```
davkean:
```suggestion
        public bool HasAuthenticationProviders => _authenticationProviders.Count > 0;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
davidwengier: We can do this without the state, it just means the dropdown is blank and the user has to be careful not to change its value or they won't be able to get it back (and the property pages don't ask if you want to save when you close). The state means we can put this unknown provider in the dropdown itself so its easier to use. It does disappear once a valid selection is made though.

Having said that, I guess strictly speaking we don't need the state to provide the nicer functionality, it can just look up the property value and create another new UnknownRemoteAuthenticationProvider when constructing the list of providers if its not found.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -534,6 +611,27 @@ public ICommand BrowseDirectoryCommand
...
+            {
+                return LazyInitializer.EnsureInitialized(ref _findRemoteMachineCommand, () =>
+                new DelegateCommand(state =>
+                {
+                    SVsServiceProvider serviceProvider = Project.Services.ExportProvider.GetExportedValue<SVsServiceProvider>();
+
+                    // HACK: Using VsUIService directly here avoids using a banned UI (serviceProvider.GetService) and can be replaced easily when MEF is hooked up to this page
+                    var vsService = new VsUIService<SVsDebugRemoteDiscoveryUI, IVsDebugRemoteDiscoveryUI>(serviceProvider, _projectThreadingService.JoinableTaskContext.Context);
+
+                    if (ErrorHandler.Succeeded(vsService.Value.SelectRemoteInstanceViaDlg(RemoteDebugMachine, RemoteAuthenticationProvider.PortSupplier, (uint)DEBUG_REMOTE_DISCOVERY_FLAGS.DRD_NONE, out string remoteMachine, out Guid portSupplier)))
```
davkean: This needs a little more thought as "DRD_SUPPORTS_UNIVERSAL_AUTH" needs to be passed if Universal is actually selected.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WindowsRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal class WindowsRemoteAuthenticationProvider : IRemoteAuthenticationProvider
+    {
+        [ImportingConstructor]
+        public WindowsRemoteAuthenticationProvider(UnconfiguredProject _) // force MEF scope
+        {
+        }
+
+        public string Name => Resources.WindowsAuth;
+
+        public Guid PortSupplier => Guid.Empty;
```
davkean: Is this not 708C1ECA-FF48-11D2-904F-00C04FA302A1?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileExtensions.cs
```diff
@@ -67,5 +70,18 @@ public static bool IsRemoteDebugEnabled(this ILaunchProfile profile)
 
             return null;
         }
+
+        public static Guid RemoteAuthenticationPortSupplier(this ILaunchProfile profile)
```
davkean: I would consider that we persist friendly names here, but respect a GUID if specified. Persisting guids in file format isn't the most friendly of things.

davkean:
This is nice Dave!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileExtensions.cs
davidwengier: I thought about that, but I thought the idea that launchSettings.json was valid independent of mapping that VS provides might be useful for other potential consumers, like Code or VS Mac etc.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WindowsRemoteAuthenticationProvider.cs
davidwengier: I got confused by [this line](http://index/?query=UAPDebuggerExtensions&rightProject=Microsoft.VisualStudio.UniversalLibrary&file=Microsoft%5CVisualStudio%5CUniversalLibrary%5CProjectFlavoring%5CDebug%5CUAPDebuggerExtensions.cs&line=20) but based on the code in [this method](http://index/?query=%60GetGuidPortSupplier&rightProject=Microsoft.VisualStudio.UniversalLibrary&file=Microsoft%5CVisualStudio%5CUniversalLibrary%5CProjectFlavoring%5CProjectConfig%5CUAPProjectFlavorCfg.cs&line=159) I think you're right, and in that method Universal uses No_Auth so I'm guessing there are some leaks to this abstraction.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileExtensions.cs
davkean: Help me undersatnd how GUIDs provide that that strings don't? These GUIDs are tied to COM objects that live in VS.

davidwengier: Is that what they are? In that case, I agree, but I couldn't find many references to them (other than passing them to the debugger) so I assumed they meant something important.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: The IItemTypeGuidProvider classes have empty ImportingConstructor constructos, are they important? Or also unnecessary?

And also, yes these contain no state, but they do have capabilities so need to be picked up by different projects. Am I right in assuming based on your statement that it doesn't matter what scope the class is in, the capabilities will still be filter appropriately on the Import side? I've always read the attributes in my head as "Export this if these capabilities are present" but I'm guessing thats not the right mental model.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WindowsRemoteAuthenticationProvider.cs
davidwengier: Okay, dug in a bit more. This should be passing 708C1ECA to the debugger, but if we pass that to the Find dialog it will throw 🤦‍♂️

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davkean: They are filtered on the import side.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WindowsRemoteAuthenticationProvider.cs
davkean: Why does the dialog throw this? Can we fix it, as it just simplifies this whole this.

davidwengier: It's trivial to fix, but I've reached out to confirm the expected behaviour of these guids as the Universal auth mode also uses two different guids at times.

dmonroym:
This'll probably have merge conflicts with #5800, might be worth rebasing on that and adding annotations to your code.
I also see that both of these have similar failing tests, not sure if they're real failures or those tests are flaky - might be worth looking

davidwengier:
Thanks for the heads up, I have a bunch of changes un-pushed for this so I'm likely to rebase any way. Currently waiting for some answers from the debugger team so hopefully the other PR will be merged before I have to worry about it :D

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: I updated the capability on this

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
davidwengier: Removed this whole thing

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IRemoteDebuggerAuthenticationService.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
+{
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private)]
+    internal interface IRemoteDebuggerAuthenticationService
+    {
+        IEnumerable<IRemoteAuthenticationProvider> GetRemoteAuthenticationModes();
+        IRemoteAuthenticationProvider? GetProviderForAuthenticationMode(string remoteAuthenticationMode);
```
davkean: Nit: "FindXXX" is typically the pattern we use for returning null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IRemoteDebuggerAuthenticationService.cs
```diff
@@ -0,0 +1,15 @@
...
+        bool ShowRemoteDiscoveryDialog(ref string remoteDebugMachine, ref IRemoteAuthenticationProvider? remoteAuthenticationProvider);
```
davkean: Should should be marked as not returning null if return is true?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public string DisplayName => "Universal";
+
+        public string Name => "None";
+
+        public Guid PortSupplierGuid => VSConstants.DebugPortSupplierGuids.NoAuth_guid;
+
+        public Guid AuthModeGuid => s_universalPortSupplier;
+
+        public uint AdditionalRemoteDiscoveryDialogFlags => /* DEBUG_REMOTE_DISCOVERY_FLAGS2.DRD_SUPPORTS_UNIVERSAL_AUTH */ 2;
```
davkean: Why a comment instead of a cast?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using Microsoft.VisualStudio.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ZeroOrMore)]
+    public interface IRemoteAuthenticationProvider
```
davkean: Given there are no other implementations of this, can we make this internal for now until we need to add to it?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+        [ImportingConstructor]
+        public UniversalRemoteAuthenticationProvider()
+        {
+        }
+
+        public string DisplayName => "Universal";
+
+        public string Name => "None";
+
+        public Guid PortSupplierGuid => VSConstants.DebugPortSupplierGuids.NoAuth_guid;
```
davkean: Explain the different between AuthMode and PortSupplier?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
+{
+    /// <summary>
+    /// Implementation of IRemoteAuthenticationProvider for the "Universal" authentication type.
+    /// </summary>
+    [Export(typeof(IRemoteAuthenticationProvider))]
+    [AppliesTo("SupportUniversalAuthentication")]
```
davkean: According to Svet, this should be basically always on right?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+        private static readonly Guid s_universalPortSupplier = new Guid("EE56E4E8-E866-4915-A18E-1DE7114BD7BB");
+
+        [ImportingConstructor]
+        public UniversalRemoteAuthenticationProvider()
+        {
+        }
+
+        public string DisplayName => "Universal";
+
+        public string Name => "None";
```
davkean: This looks wrong? It gets persisted the same way as "NoAuthentication"?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davkean: I don't understand how "universal" gets turned on - it just passed NoAuth as the PortSupplier to the debugger the same way that NoAuthentication does.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IRemoteAuthenticationProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// </summary>
+        string Name { get; }
+        /// <summary>
+        /// Guid passed to the debugger
+        /// </summary>
+        Guid PortSupplierGuid { get; }
+        /// <summary>
+        /// Guid used for to specify and read the authentication mode for the Remote Discovery Dialog
+        /// </summary>
+        Guid AuthModeGuid { get; }
```
davkean: "AuthenticationMode".

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileExtensions.cs
```diff
@@ -8,6 +8,7 @@ internal static class LaunchProfileExtensions
         public const string SqlDebuggingProperty = "sqlDebugging";
         public const string RemoteDebugEnabledProperty = "remoteDebugEnabled";
         public const string RemoteDebugMachineProperty = "remoteDebugMachine";
+        public const string RemoteAuthenticationModeProperty = "remoteAuthenticationMode";
```
davkean: UWP and Legacy calls this "AuthenticationMode" can we pick the same name "authenticationMode" so that it makes easier to rationalize when we have to rationalize their settings with ours when we replace legacy?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/RemoteDebuggerAuthenticationService.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
+{
+    [Export(typeof(IRemoteDebuggerAuthenticationService))]
+    [AppliesTo(ProjectCapabilities.AlwaysApplicable)]
+    internal class RemoteDebuggerAuthenticationService : IRemoteDebuggerAuthenticationService
```
davkean: I think these should be applied with [Order] which is respected so that we have a predictable order in which they are displayed in the UI (ie None, Windows, Universal), currently its based on the order that MEF returns them.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/RemoteDebuggerAuthenticationService.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            _remoteDiscoveryUIService = remoteDiscoveryUIService;
+            _threadingService = threadingService;
+            AuthenticationProviders = new OrderPrecedenceImportCollection<IRemoteAuthenticationProvider>(projectCapabilityCheckProvider: unconfiguredProject);
+        }
+
+        [ImportMany]
+        private OrderPrecedenceImportCollection<IRemoteAuthenticationProvider> AuthenticationProviders { get; }
+
+        public IRemoteAuthenticationProvider? GetProviderForAuthenticationMode(string? remoteAuthenticationMode) => AuthenticationProviders.FirstOrDefaultValue(p => p.Name.Equals(remoteAuthenticationMode, StringComparisons.LaunchProfileProperties));
```
davkean: Why is null/empty allowed?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: That type is not in any published Nuget package (that I could find at least) and for one int its not worth doing anything special.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: That's not clear to me, and since he didn't add it to the legacy support in his PR I don't think we should do it here either. Its a simple follow up to add a capability or remove this line if it is determined to be the right thing to do.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IRemoteDebuggerAuthenticationService.cs
davidwengier: No reason the dialog couldn't return something we don't know about.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: AuthMode and PortSupplier are different, and I think its a bug that the parameter name for the Remote Discovery Dialog is called "portSupplier". The term is used for the combination of how msvsmon will be deployed, and how to connect to it. So when using Universal authentication mode, thats the trigger to deploy msvsmon and start it, but no authentication is required for the connection to it.  Obviously the deployment work is not part of this PR which is why this PR doesn't turn on Universal as an option.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/UniversalRemoteAuthenticationProvider.cs
davidwengier: Yep, this was a bad copy/paste, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/RemoteDebuggerAuthenticationService.cs
davidwengier: Assuming you mean for the parameter, empty could be in the json file, or come from a bad json file, and null could come from the ConsoleDebugTargets if the profile isn't full defined - but also because string.Equals handles all of those cases happily.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IRemoteDebuggerAuthenticationService.cs
davkean: So in that case we return the remote machine, but throw away the authentication provider? What situation would we want that?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/RemoteDebuggerAuthenticationService.cs
davkean: Nit: APIs that return "null" to help the caller do a null check make it harder to understand the API itself. It gives the impression that `remoteAuthenticationMode` is optional, when in fact its not if you want anything useful out of it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IRemoteDebuggerAuthenticationService.cs
davidwengier: I would want that in all situations - doing the opposite is quite literally throwing out the baby with the bathwater.

Think about a user who uses this dialog, types in their machine name and picks the "Foo" auth type, which is unsupported. If we throw away both pieces of data, they press OK and nothing happens. With the current code, the machine name changes, and the auth mode changes to a blank entry to indicate that its not in the drop down. Inspecting the drop down also clearly indicates that "Foo" isn't listed.

Neither of those is a perfect scenario, I'll grant you, but one leads to a "the dialog does nothing" bug report, and one leads to "Foo is not in the drop down", the latter being much easier to deal with. We could show a message box if the auth provider isn't found, so they'd know we were throwing away their data, but a message box like that should be shown by the dialog, not by us, as once we've shown it its too late for them to recover, they just have to go back into the dialog and try again, which is also a bad experience. This is where the "UnknownAuthProvider" would have come in in the past, but I removed that as its not something that we can provide when the auth mode guids and port supplier guids are different.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/RemoteDebuggerAuthenticationService.cs
davidwengier: I'm not sure I follow this one. If you're talking about the return value, that allows nulls because this is processing user input from the json file so it could be rubbish data. Would a TryGetXXX form make more sense to you? If you're talking about the parameter, again thats null because it could be null in the JSON file. This could disallow nulls, but that would mean the thing that reads the JSON file needs to know that a null should be mapped to an empty string, or avoid calling this method, which might not be something it can reason about. Maybe its just me but I don't equate nullable with optional, since we have optional parameters. Nullable just means that the method can deal with a null ¯\_(ツ)_/¯

davkean: Providers cannot have a null name, yet this method takes a nullable name. That is my point.

It does this so that the JSON parsing code doesn't need to do a null check before calling this method.

davkean: Why would the JSON code need to map null -> empty? Empty names also do not make sense.

davidwengier:
<Bernie Sanders "I am once again asking for your code review".png>

PR #5883, opened 2020-02-10
Block rename in online scenarios

tmeschter:
We've gotten reports that renaming a file in Solution Explorer in VS Online scenarios causes VS to crash. I haven't been able to reproduce the issue, but it is conceivable that the server is trying to show the rename dialog and things are going sideways at that point. Until we can make this work properly we should block off the scenario and not even try to rename the file.

Fixes [AB#1063020](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1063020).

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
@@ -30,6 +30,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
         private readonly IRoslynServices _roslynServices;
         private readonly IWaitIndicator _waitService;
         private readonly IRefactorNotifyService _refactorNotifyService;
+        private readonly IVsOnlineServices _vsOnlineServices;
 
         [ImportingConstructor]
         internal Renamer(IUnconfiguredProjectVsServices projectVsServices,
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
```diff
@@ -65,6 +65,12 @@ public void HandleRename(string oldFilePath, string newFilePath)
 
         internal async Task HandleRenameAsync(string oldFilePath, string newFilePath)
         {
+            // Do not offer to rename the file in VS Online scenarios.
+            if (Shell.KnownUIContexts.CloudEnvironmentConnectedContext.IsActive)
```
swesonga: Is there an issue tracking enabling the scenario?

davkean:
What happens if you drag a file with the same name over the top of all another file? Does it also crash?  Trying to figure out if this is a general problem of showing dialogs, or specific to rename.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.cs
tmeschter: https://devdiv.visualstudio.com/devdiv/_queries/edit/977305

tmeschter:
@davkean I'm not sure what will happen in that exact case as it depends heavily on how the VS Online Solution Explorer is implemented, and how it interacts with the project system. I'll give it a try and see how it goes. However, what I've been told is that any dialogs that may pop up are potentially problems in VS Online scenarios, so whether or not they cause problems right now we should avoid them.

tmeschter:
I'm going to update this to abstract out `KnownUIContexts` so I can unit test it properly.

davkean:
> I'll give it a try and see how it goes. However, what I've been told is that any dialogs that may pop up are potentially problems in VS Online scenarios, so whether or not they cause problems right now we should avoid them.

Can you figure this out so that we don't piecemeal this for the hundreds of places we open dialogs?

Issue #5888, opened 2020-02-12
Remove Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles and roll it into ProjectSystem.vsix


davidwengier:
At the moment we have two vsixes that are part of a single .vsman file that gets inserted into VS. We don't need two, as they're always inserted together and we need both, but in order to remove Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles we need to work out the right steps with RepoToolset, or the right things to do despite it, to correctly produce a vsman file from a set of SWR files and a vsixmanifest. At the moment the two systems fight with each other.

MiYanni:
During my work on removing the use of RepoToolset, I investigated on why CommonFiles uses `.swixproj` versus the 2 other VSIX packages are .csproj (VS Extension) based. It seems that the insertion mechanics change between them:
- Projects containing a `.csproj` & `.vsixmanifest` are VS extension projects, which create a VSIX package.
  - Allows for F5 experience
  - Doesn't require admin access to install
  - Can only install into 8 specific VS sub-directories
- Projects containing a `.swixproj` & `.swr` are loose files to be put into a VSIX package (NOT AN EXTENSION).
  - No F5 experience
  - Requires admin access to install
  - Can install into any sub-directory in VS
  
So, the actual usage seems to change here. Based on this, I'll be keeping the CommonFiles project and VSIX production mechanism. However, I won't be producing a separate `.vsman` file and then merging it into the ProjectSystemSetup's `.vsman`. Instead, I'll just include the .json manifest that CommonFiles creates via a `<MergeManifest />` node in the `.vsmanproj` for ProjectSystemSetup.

PR #5889, opened 2020-02-12
Add basic dependencies to SWR file

davidwengier:
Our VS setup authoring didn't express any dependencies, so this adds a basic set. In an ideal world this wouldn't be in the CommonFiles project but that one already had an SWR so I logged #5888 rather than hold things up by fighting with repotoolset today.

### DIFF
--- a/setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
+++ b/setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
@@ -8,6 +8,14 @@ vs.localizedResources
                        title="Microsoft VisualStudio Managed Project System Common Files"
                        description="Microsoft VisualStudio ProjectSystem for C#/VB(Managed) Projects"
 
+vs.dependencies
+  vs.dependency id=Microsoft.NetCore.Component.DevelopmentTools
+                version=[$(Version),)
+  vs.dependency id=Microsoft.VisualStudio.ProjectSystem.Full
+                version=[$(Version),)
+  vs.dependency id=Microsoft.VisualStudio.Component.Roslyn.LanguageServices
+                version=[$(Version),)
+
 folder "InstallDir:MSBuild\Microsoft\VisualStudio\Managed"
   file source="$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets"
   file source="$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets"


setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
```diff
@@ -8,6 +8,14 @@ vs.localizedResources
                        title="Microsoft VisualStudio Managed Project System Common Files"
                        description="Microsoft VisualStudio ProjectSystem for C#/VB(Managed) Projects"
 
+vs.dependencies
+  vs.dependency id=Microsoft.NetCore.Component.DevelopmentTools
```
davkean: This dependency doesn't include us?

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
```diff
@@ -8,6 +8,14 @@ vs.localizedResources
...
+                version=[$(Version),)
```
davkean: What is $(Version) here referring to?

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davidwengier: No, with an asterisk that I haven't dug through ever layer of the graph exhaustively. You can see that package here: http://index/?query=%60Microsoft.NetCore.Component.DevelopmentTools&rightProject=VS(Repo)&file=src%5CSetupPackages%5CGroup%5CComponent%5CNetCore.DevelopmentTools%5Ccomponent.swr&line=3

I'm going to take this PR and generate an insertion that will hopefully cause the master package json file in the VS repo to be updated, and then there is a tool I can run that will output the complete closure of the dependency graph for a package, so intend to use that to verify.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davidwengier: Version is the [VsixVersion](https://github.com/dotnet/project-system/blob/master/setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj#L14)

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davkean: I'm looking at that and not understanding why we have a dependency on it? I think we'll get installed in situations that this isn't installed, such as desktop-only./

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davkean: Why does our version determine what versions we have a dependency on?

davidwengier: Don't ascribe any meaning here, this is just convenience because we follow the version of the product, so we don't need to keep updating this file.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davidwengier: To me this is part of what the user would expect when opening a project that uses our project system. eg. the other day you said "we may not depend directly on the debugger, but without it our project system is not very useful", well this package brings in the Microsoft.VisualStudio.PackageGroup.DotNetCoreDebugging package.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davkean: Walk though this:

- We insert a new package from our 16.6 before Roslyn has inserted from 16.6, now we fail to install because our dependencies say we depend on "16.6" of Roslyn which doesn't yet exist.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
davkean: Understood, but I don't this is that dependency, this depends on F# and .NET Core SDK which we do not depend on.

davkean: I would expect the reverse dependency, in fact.

Issue #5834, opened 2020-01-27
PLM: PackageRestore needs to connect jtf link to prevent VS hang when Roslyn blocks UI thread waiting for nuget restore to finish


davkean:
Ported from and contains a dump: https://devdiv.visualstudio.com/DevDiv/_queries/edit/1052518/?triage=true.

[regression]
Every time I ask VS to do a background task, like go to a method definition with F12, or find all references (Shift+F12), or right click and select “Quick Actions and Refactorings” or “Remove and Sort Usings”, I get a popup window with the message:

“Organizing document. Waiting for background work to finish”.

And it never goes away. I have waited up to 10 minutes and it still stays thinking.

Sometimes VS crashes after a while.

I don’t know what was the previous version where it was last working. I have updated 3-4 times, hoping this issue would be fixed with the updates, but that is not true.

My current version is 16.5.0 Preview 2.0 [29708.233.master]

To find diagnostic information, see the original linked feedback ticket.

From: Lifeng Lu <lifengl@microsoft.com>
Sent: Thursday, January 16, 2020 12:21 PM
To: Adrian Vinca <adrianv@microsoft.com>; Carlos Sanchez Lopez <calope@microsoft.com>; David Barbet <David.Barbet@microsoft.com>; Oleg Tkachenko <olegtk@microsoft.com>; Jose Perez Rodriguez <joperezr@microsoft.com>; David Kean <David.Kean@microsoft.com>
Cc: Aditya Kadam <adityk@microsoft.com>; David Pugh <dpugh@microsoft.com>; Hannah Hong <Hannah.Hong@microsoft.com>; Michael Eng <meng@microsoft.com>
Subject: RE: Help with a Visual Studio dialog blocking me

 

The JoinableTask chain actually is broken much earlier than that. Basically, this dead lock can happen for any project with NuGet reference, instead of those which has credential requirements.

 

Basically, the CPS operation progress contract requires all registered IProgressTrackerOutputDataSource also implementing IJoinableProjectValueDataSource.  (The dependency between two types were not added because we didn’t want to introduce a breaking change at that time.) It appeared that the Net Core project didn’t implement the second interface, and the JTF chain is never established between the operation progress feature and NuGet restore logic.  It also means this bug would impact more customers than those who depends on credentials.

 

There might be more JTF work need be done between the Net Core NuGet package restore code and NuGet restoring layer. I will take a look the code. 

 

[...]

PR #5890, opened 2020-02-13
Fix hang if you invoke code fix during restore

davkean:
Fixes: https://github.com/dotnet/project-system/issues/5834.

We were not connecting the chain between NuGet restore operations that needed the UI thread, and Roslyn which was waiting on operation progress to complete. This is because our IProgressTrackerOutputDataSource didn't implement IJoinableProjectValueDataSource and connect it to the same collection that we're using to run/nominate restore.

This connects the dots between the two things, which required:

1) Moving IPackageRestoreService to derive from ProjectValueDataSource<T>
2) We no longer throw it way based on removal of PackageReference, and instead just prevent us from nominating NuGet. It is quite complex to have a data source that shuts itself down and stops producing data after a capability has been removed. We need a pattern for this, but this bug is not the one to do it in.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -0,0 +1,201 @@
...
+{
+    /// <summary>
+    ///     Responsible for pushing ("nominating") project data such as referenced packages and 
+    ///     target frameworks to NuGet so that it can perform a package restore and returns the 
+    ///     result.
+    /// </summary>
+    [Export(typeof(IPackageRestoreDataSource))]
+    [Export(ExportContractNames.Scopes.UnconfiguredProject, typeof(IProjectDynamicLoadComponent))]
+    [AppliesTo(ProjectCapability.PackageReferences)]
+    internal class PackageRestoreDataSource : ChainedProjectValueDataSourceBase<RestoreData>, IPackageRestoreDataSource, IProjectDynamicLoadComponent
```
davkean: Almost all of this code is identical with exception to LinkExternalInput/LoadAsync/UnloadAsync/RestoreAsync.

davkean:
tag @lifengl

davidwengier:
Should this target dev16.5.x?

davkean:
Indeed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
drewnoakes: Is there a pattern to extract here then?

davkean: As in, I deleted PackageRestoreService and replaced it with this class but the diff doesn't show that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -0,0 +1,201 @@
...
+        //                                             ||
+        //                                        RestoreData
+        //                                        //        \\ 
+        //                                       //          \\
+        //  ____________________________________\/___      ___\/____________________________________
+        // |                                         |    |                                         |
+        // |        PackageRestoreProgressTracker    |    |       PackageRestoreProgressTracker     |  Publishes restore progress to operation progress
+        // |            (Debug|AnyCPU|net45)         |    |       (Debug|AnyCPU|netcoreapp30)       |
+        // |_________________________________________|    |_________________________________________|
+        //
```
davidwengier: <3 <3 <3
🎉🎉 🎉  

Yay comments, and yay ascii art!

PR #5894, opened 2020-02-14
Add_DisableCentralPackageVersions

cristinamanum:
Add DisableCentralPackageVersions for NuGet CPVM scenarios.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
@@ -24,7 +24,7 @@
   <StringProperty Name="EnableCentralPackageVersions"
                   ReadOnly="True"
                   Visible="False" />
-                  
+
   <StringProperty Name="MSBuildProjectDirectory"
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
```diff
@@ -185,4 +185,7 @@
                   ReadOnly="True"
                   Visible="False" />
 
+  <StringProperty Name="DisableCentralPackageVersions"
```
davkean: Nit: I would have called this "EnableCentralPackageVersions" and defaulted to true - negative options are hard to parse in the human brain.

davidwengier: Also nit: Once the name is settled, can this be moved to maintain alphabetical order of properties. Makes future diffing easier.

cristinamanum: Updated to address both of the comments.

PR #5896, opened 2020-02-16
Add properties folder icon

### LINKED ISSUE #5803: App Designer Folder: New Icon
We would like a new icon that is folder based.

drewnoakes:
Fixes #5803.

image

image

image

image

image

image


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/AbstractSpecialFolderProjectTreePropertiesProvider.cs
```diff
@@ -92,7 +92,7 @@ private void ApplySpecialFolderProperties(IProjectTreeCustomizablePropertyValues
 
             // Avoid overwriting icon if the image provider didn't provide one
             propertyValues.Icon = icon ?? propertyValues.Icon;
-            propertyValues.ExpandedIcon = expandedIcon ?? propertyValues.ExpandedIcon;
+            propertyValues.ExpandedIcon = expandedIcon ?? icon ?? propertyValues.ExpandedIcon;
```
davkean: This is unneeded that's exactly what CPS will do if unset. Here you are overwriting ExpandIcon if already set.

drewnoakes: Backed the change out.

PR #5906, opened 2020-02-20
Consolidate reference manager capabilities

### LINKED ISSUE #5829: Consolidate reference manager capabilities
Currently we have capabilities such as:

- `ReferenceManagerAssemblies`
- `ReferenceManagerBrowse`
- `ReferenceManagerCOM`
- `ReferenceManagerProjects`
- `ReferenceManagerSharedProjects`
- `ReferenceManagerWinRT`

Additionally we have capabilities such as:

- `AssemblyReferences`
- `COMReferences`
- `PackageReferences`
- `ProjectReferences`
- `SharedProjectReferences`
- `WinRTReferences`

It may be possible to consolidate these.

Note that `ReferenceManagerBrowse` has no analogue, nor does `PackageReferences`.

drewnoakes:
Fixes #5829.

We have existing capabilities for each type of reference. This change removes the Reference Manager specific ones, using the reference type capabilities instead.

It also removes implementations of `IVsReferenceManagerUserAsync`, as  those provided by CPS are equivalent.

This change removes a condition that would hide the COM references tab when targeting .NET Core App versions before 3.0.


davkean:
I think we can take a step further and just delete all of our IVsReferenceManagerUserAsync implementators - they all appear to do nothing over the built-in ones except check capability in IsApplicable instead of at the Export level. I can't see why they were added in the first place? Maybe so we could use different capabilities to drive it?

drewnoakes:
Force-pushed a commit that removes most of this code instead. The implementations provided by CPS are adequate given we don't need these `ReferenceManager*` capabilities.

davidwengier:
> This change removes a condition that would hide the COM references tab when targeting .NET Core App versions before 3.0.

This seems bad. So if I target .NET Standard I'll now be able to add a reference to a COM library even though it might not/won't work?

davkean:
Oh I missed that - that wasn't in the original change. What impact did removing COMReferences capability have?

drewnoakes:
COM reference issue addressed in #5962.

Issue #4797, opened 2019-05-07
Microsoft.VisualStudio.ProjectVS.NuGet.ProjectRestoreInfo uses excessive memory


lifengl:
After loading Roslyn solution, force GC, this structure uses 2.5% of the managed heap.  Almost all of them to represent TargetFrameworks.  I would expect most projects share the same target framework, and so the overall memory usage of this structure to be close to 0%.

![image](https://user-images.githubusercontent.com/11638466/57328338-03236e80-70c6-11e9-8dcd-5d8a567f0a0e.png)

davkean:
It's a little harder than that - they don't share the same "target frameworks" as that object represents NuGet's representation of a configuration.

davkean:
We store this information to compare it with a new version of it to avoid notifying NuGet of unneeded restores - we should hash the thing, and notify if the hash has changed.

PR #5917, opened 2020-02-25
Hash restore inputs to avoid storing them

davkean:
Fixes: https://github.com/dotnet/project-system/issues/4797

This hashes restore inputs so that we can avoid storing them around just so that we can figure out if anything has changed.


davkean:
I think you do in cases of empty values, otherwise this looks the same:

```
Property1: <empty>
Property2: Foo

Property1: Foo
Property2: <empty>
```

PR #5932, opened 2020-03-03
Enable new dependencies tree context menus

### LINKED ISSUE #5922: Provide new menus for dependency tree nodes, and enable/implement new commands
Follows #5845.

We now have menus and their groups defined in the VS repo. Now we need to instruct CPS to select them for these tree items.

We will need command handlers to enable, make visible and handle our new commands too.

Doing both of these tasks at the same time seems to make the most sense to me.

drewnoakes:
Fixes #5922.

- Opts in new menus for dependencies tree items
- Suppresses the old "Add Reference" command 
- Wires up behaviour for new "Add * Reference" commands to launch Reference Manager at the correct page

### Menu for 'Dependencies' node

image

Note that the NuGet team will be modifying package-related items here too.

### Menu for 'Assemblies' group node

image

### Menu for 'COM' group node

image

### Menu for 'Projects' group node

image

### Main "Project" menu

image

### Project | Add context sum-menu

image

### Note

- 'Analyzers', 'Frameworks' and 'Packages' nodes have empty context menus, though they may be added to in future if needed. NuGet will almost certainly be adding the the 'Packages' context menu.
- Target framework nodes (for multi-targeting projects) currently display the same as the 'Dependencies' node itself. We may wish to suppress commands from these menus as they currently apply to all target frameworks. This will require coordination with the NuGet team.
- Requires https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/231736

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/DependenciesContextMenuProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/DependenciesContextMenuProvider.cs
@@ -1,0 +1,157 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Commands
+{
+    /// <summary>
+    /// Provides context menus for nodes in the dependencies tree.
+    /// </summary>
+    [Export(typeof(IProjectItemContextMenuProvider))]
+    [AppliesTo(ProjectCapability.DependenciesTree)]
+    [Order(ProjectSystem.Order.Default)]
+    internal class DependenciesContextMenuProvider : IProjectItemContextMenuProvider
+    {
+        private static class Menus
+        {
+            public const int IDM_VS_CTXT_REFERENCEROOT              = 0x0450;
+            public const int IDM_VS_CTXT_REFERENCE                  = 0x0451;
+            public const int IDM_VS_CTXT_DEPENDENCYTARGET           = 0X04A0; // Target framework group node
+            public const int IDM_VS_CTXT_REFERENCE_GROUP            = 0X04A1; // Assembly reference group
+            public const int IDM_VS_CTXT_PACKAGEREFERENCE_GROUP     = 0x04A2;
+            public const int IDM_VS_CTXT_PACKAGEREFERENCE           = 0x04A3;
+            public const int IDM_VS_CTXT_COMREFERENCE_GROUP         = 0x04A4;
+            public const int IDM_VS_CTXT_COMREFERENCE               = 0x04A5;
+            public const int IDM_VS_CTXT_PROJECTREFERENCE_GROUP     = 0x04A6;
+            public const int IDM_VS_CTXT_PROJECTREFERENCE           = 0x04A7;
+            public const int IDM_VS_CTXT_SHAREDPROJECTREFERENCE     = 0x04A8;
+            public const int IDM_VS_CTXT_FRAMEWORKREFERENCE_GROUP   = 0x04A9;
+            public const int IDM_VS_CTXT_FRAMEWORKREFERENCE         = 0x04AA;
+            public const int IDM_VS_CTXT_ANALYZERREFERENCE_GROUP    = 0x04AB;
+            public const int IDM_VS_CTXT_ANALYZERREFERENCE          = 0x04AC;
+            public const int IDM_VS_CTXT_SDKREFERENCE_GROUP         = 0x04AD;
+            public const int IDM_VS_CTXT_SDKREFERENCE               = 0x04AE;
+        }
+
+        public bool TryGetContextMenu(IProjectTree projectItem, out Guid menuCommandGuid, out int menuCommandId)
+        {
+            Requires.NotNull(projectItem, nameof(projectItem));
+
+            if (projectItem.Flags.Contains(DependencyTreeFlags.DependenciesRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_REFERENCEROOT;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.TargetNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_DEPENDENCYTARGET;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.AssemblySubTreeRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_REFERENCE_GROUP;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.AssemblyDependency))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_REFERENCE;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.NuGetSubTreeRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_PACKAGEREFERENCE_GROUP;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.NuGetDependency))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_PACKAGEREFERENCE;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.ComSubTreeRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_COMREFERENCE_GROUP;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.ComDependency))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_COMREFERENCE;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.ProjectSubTreeRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_PROJECTREFERENCE_GROUP;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.ProjectDependency))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_PROJECTREFERENCE;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.SharedProjectDependency))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_SHAREDPROJECTREFERENCE;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.AnalyzerSubTreeRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_ANALYZERREFERENCE_GROUP;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.AnalyzerDependency))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_ANALYZERREFERENCE;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.FrameworkSubTreeRootNode))
+            {
+                menuCommandId = Menus.IDM_VS_CTXT_FRAMEWORKREFERENCE_GROUP;
+            }
+            else if (projectItem.Flags.Contains(DependencyTreeFlags.FrameworkDependency))
+            {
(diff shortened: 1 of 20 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/DependenciesContextMenuProvider.cs
```diff
@@ -0,0 +1,157 @@
...
+            int count = 0;
+            bool containsProhibited = false;
+
+            foreach (IProjectTree item in projectItems)
+            {
+                count++;
+
+                if (!containsProhibited)
+                {
+                    // TODO when CPS inserts, use ContainsAny here instead
```
davkean: This needs explanation, I don't understand what its trying to prevent.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/ReferenceManagerCommandHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        public ReferenceManagerCommandHandler(UnconfiguredProject unconfiguredProject, IReferencesUI referencesUI)
+        {
+            Requires.NotNull(unconfiguredProject, nameof(unconfiguredProject));
+            Requires.NotNull(referencesUI, nameof(referencesUI));
+
+            _unconfiguredProject = unconfiguredProject;
+            _referencesUI = referencesUI;
+        }
+
+        public CommandStatusResult GetCommandStatus(IImmutableSet<IProjectTree> items, long commandId, bool focused, string? commandText, CommandStatus status)
```
davkean: "progressiveStatus"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/ReferenceManagerCommandHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        public const int CmdidAddSharedProjectReference  = 0x203;
+        public const int CmdidAddWinRTReference          = 0x204;
+
+        private readonly UnconfiguredProject _unconfiguredProject;
+        private readonly IReferencesUI _referencesUI;
+
+        [ImportingConstructor]
+        public ReferenceManagerCommandHandler(UnconfiguredProject unconfiguredProject, IReferencesUI referencesUI)
+        {
+            Requires.NotNull(unconfiguredProject, nameof(unconfiguredProject));
```
davkean: We don't null check MEF constructors.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
```diff
@@ -88,33 +91,44 @@ public static class DependencyTreeFlags
...
-        internal static readonly ProjectTreeFlags SubTreeRootNode = ProjectTreeFlags.Create(ProjectTreeFlags.Common.VirtualFolder) +
-                                                                    ProjectTreeFlags.Create(ProjectTreeFlags.Common.ReferencesFolder);
+        
+        internal static readonly ProjectTreeFlags SubTreeRootNode = ProjectTreeFlags.Create("SubTreeRootNode");
 
+        internal static readonly ProjectTreeFlags SubTreeRootNodeFlags
+                = ProjectTreeFlags.Create(ProjectTreeFlags.Common.VirtualFolder)
+                + SubTreeRootNode;
 
+        internal static readonly ProjectTreeFlags AnalyzerSubTreeRootNode = ProjectTreeFlags.Create("AnalyzerSubTreeRootNode");
```
davkean: Why do these have a different name "SubTreeRootNode" to what you named the context menus? Can we consistently use "GROUP"?

davkean:
It looks like "New Solution View" and "Scope to this" are missing from the new context menus?

davkean:
SDK I think would want to open this page:

image

davkean:
We're missing legacy's Add Analyzer command - do we have a bug on that?

davkean:
- What does the Project -> Add context menu look like?
- What does the Project top level menu look like?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/DependenciesContextMenuProvider.cs
drewnoakes: The current code is fine. It could just be a little cleaner once we have newer CPS bits inserted.

drewnoakes:
> It looks like "New Solution View" and "Scope to this" are missing from the new context menus?

This was intentional, to keep the context menus small and focussed. It's available on the "Dependencies" node, but nothing lower.

> We're missing legacy's Add Analyzer command - do we have a bug on that?

I filed #5934. Honestly that UI doesn't look very useful given analyzers are generally added from packages.

> SDK I think would want to open this page:

What capabilities would that be gated on? `WinRTReferences & SdkReferences`?

> What does *** look like?

These are unchanged.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
drewnoakes: "Root" has a pre-existing meaning in the existing menus/groups, and it's not a good fit for the grouping nodes anyway. I considered updating the flags to reflect this, but the existing flags are public API.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        // TODO once CPS inserts, use this field along with ContainsAny
+//      private static readonly ProjectTreeFlags s_hideAddReferenceForFlags = DependencyTreeFlags.DependenciesRootNode + DependencyTreeFlags.TargetNode;
+
+        protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree node, bool focused, string? commandText, CommandStatus progressiveStatus)
+        {
+//          if (node.Flags.ContainsAny(s_hideAddReferenceForFlags))
+            if (node.Flags.Contains(DependencyTreeFlags.DependenciesRootNode) || node.Flags.Contains(DependencyTreeFlags.TargetNode))
+            {
+                return GetCommandStatusResult.Suppressed;
```
davkean: This should be handled, enabled but hidden and only on the context menu (I'll try and find the flag). Otherwise, this prevents Project -> Add Reference from working if the dependency node is selected, with no replacement.

davkean:
> This was intentional, to keep the context menus small and focussed. It's available on the "Dependencies" node, but nothing lower.

I should be allowed to scope anywhere in the tree. Every single node can be scoped except these ones.

> What capabilities would that be gated on? WinRTReferences & SdkReferences?

Yes, both go through the same provider context.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
davkean: I think the flag to do this OLECMDF_DEFHIDEONCTXTMENU which CPS doesn't yet expose.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
davkean: How are they existing public API? You just just added them.

drewnoakes:
> I should be allowed to scope anywhere in the tree

That is feedback for https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/231736. I'll follow up there.

drewnoakes:
> Every single node can be scoped except these ones.

Graph nodes cannot either. This work doesn't handle menus for graph nodes. I'll fix that as part of #5935.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
drewnoakes: `"SubTreeRootNode"`, from which the naming pattern extends, already existed.

Also, these flags are identical to those recently removed in #5538, before which they were present in many releases with these names.

I prefer the name group and would like to use it consistently, but favoured back compat here. Do you feel it's worth the break here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
drewnoakes: Addressed in d52300585 (will test once VS build is ready).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Input;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Commands
+{
+    /// <summary>
+    /// Suppresses the "Add Reference..." command in specific dependencies tree context menus.
```
davkean: Nit: "in lieu of more specific versions "Add Project Reference", etc"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// which includes specific "Add ___ Reference..." commands, including "Add Assembly Reference...".
+    /// </remarks>
+    [ProjectCommand(VSConstants.CMDSETID.StandardCommandSet2K_string, (long)VSConstants.VSStd2KCmdID.ADDREFERENCE)]
+    [AppliesTo(ProjectCapability.DependenciesTree)]
+    [Order(ProjectSystem.Order.Default)]
+    internal sealed class SuppressAddReferenceCommand : AbstractSingleNodeProjectCommand
+    {
+        protected override Task<CommandStatusResult> GetCommandStatusAsync(IProjectTree node, bool focused, string? commandText, CommandStatus progressiveStatus)
+        {
+            return GetCommandStatusResult.Suppressed;
```
davkean: This again, disables the command which I don't think is correct.

This will break any extension, shortcut, or automatation that uses this command, all we want to do is hide it but still support it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
drewnoakes: The remarks should cover this, no?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Commands/SuppressAddReferenceCommand.cs
drewnoakes: Addressed in 5ccbc12ecaba13a848a37deaf518746fbeefa10f.

drewnoakes:
I'll hold off on merging this until https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/231736?_a=overview is in, as without the VS-side this removes "Add Reference" without a replacement.

drewnoakes:
VS PR now merged.

PR #5945, opened 2020-03-05
Remove unused code related to framework assemblies in pacakges

drewnoakes:
This stems from the fact that `DependencyTreeFlags.FxAssemblyDependency` is not set anywhere, and so testing for it is redundant.

It looks like this code was never used. Searching history shows the flag was added in 19d68652ad469ad602cac25273fe3562926028e8 but no commit since has included it.


drewnoakes:
Filed #5946 to track showing framework assembly references in packages.

PR #5947, opened 2020-03-05
Tidy dependency tree flags

drewnoakes:
- Renaming of dependency 'root' nodes to 'group' nodes.
- Renaming of 'NuGet packages' to 'packages'.
- Removal of public 'BaseReferenceFlags' item, which is unused across index.
- Inline a few aggregate flags.
- Remove general 'NuGetDependency' flag in favour of more specific alternatives.
- Clarify diagnostics as relating to packages.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
@@ -12,38 +12,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
     /// </remarks>
     public static class DependencyTreeFlags
     {
+        /// <summary>
+        /// Identifies the root "Dependencies" node of the dependencies tree. This flag should exist on only one node in a project.
+        /// </summary>
         internal static readonly ProjectTreeFlags DependenciesRootNode = ProjectTreeFlags.Create("DependenciesRootNode");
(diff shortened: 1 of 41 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependencyTreeFlags.cs
```diff
@@ -80,53 +55,61 @@ public static class DependencyTreeFlags
...
         internal static readonly ProjectTreeFlags GenericResolvedDependencyFlags
-                = ResolvedDependencyFlags
-                + ResolvedReferenceFlags
+                = ProjectTreeFlags.Reference
+                + ProjectTreeFlags.ResolvedReference
+                + ResolvedDependencyFlags
                 + GenericDependency;
 
-        internal static readonly ProjectTreeFlags TargetNode = ProjectTreeFlags.Create("TargetNode");
+        internal static readonly ProjectTreeFlags DependencyTargetFrameworkGroup = ProjectTreeFlags.Create("DependencyTargetFrameworkGroup");
```
davkean: This name would change if we chose a different grouping here, I would avoid baking in "TargetFramework".

drewnoakes: Without an alternative I'm not sure what you mean. We currently have specific names for the other kinds of groupings we have here. I agree we may wish to restructure the tree at some point, however if we do that we can also modify the flags we use.

drewnoakes: #5961

PR #5950, opened 2020-03-06
Change nullablility of IVsXXService interfaces

davkean:
This changes the nullability of IVsXXServices interfaces to be determined by the importer of the service. Instead of:

``` C#
IVsUIService<IVsOutputWindow> outputWindow = ...

if (outputWindow.Value != null
{

}

```

and:

``` C#
IVsUIService<IVsOutputWindow> outputWindow = ...
Assumes.Present(outputWindow.Value);
```

We now move to:

``` C#
IVsUIService<IVsOutputWindow?> outputWindow = ...

if (outputWindow.Value != null
{
...
}
```

and:

``` C#
IVsUIService<IVsOutputWindow> outputWindow = ...
outputWindow.Value.ActivatePane(..)
```

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IVsService`1.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IVsService`1.cs
@@ -15,7 +15,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
     ///     The type of the service to retrieve and return from <see cref="GetValueAsync"/>.
     /// </typeparam>
     [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
-    internal interface IVsService<T> where T : class
+    internal interface IVsService<T> 
+        where T : class?
     {
         /// <summary>
         ///     Gets the service object associated with <typeparamref name="T"/>.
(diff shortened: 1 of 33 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IVsService`1.cs
```diff
@@ -15,7 +15,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
     ///     The type of the service to retrieve and return from <see cref="GetValueAsync"/>.
     /// </typeparam>
     [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
-    internal interface IVsService<T> where T : class
+    internal interface IVsService<T> 
+        where T : class?
```
drewnoakes: Is this documented anywhere? I searched around a bit but couldn't find anything.

davkean: https://github.com/dotnet/docs/issues/17355

dmonroym:
Is there a typo on the last example of the description? I don't fully follow the whole PR but from my understanding it should read as: 

``` C#
IVsUIService<IVsOutputWindow?> outputWindow = ...
outputWindow.Value.ActivatePane(..)
```

davkean:
No, not a typo. That usage shows non-optional usage, you use nullable when the service might not be installed.

PR #5965, opened 2020-03-10
move to MIT license

### LINKED ISSUE #5887: Move to MIT license from Apache
http://github.com/dotnet/roslyn and http://github.com/dotnet/runtime have all to MIT license away from Apache. We should follow suit to simplify code moves between these repositories.

dmonroym:
Fixes #5887

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,4 +1,4 @@
-<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information. -->
 <Project>
   <PropertyGroup>
     <IsUnitTestProject Condition="'$(IsUnitTestProject)' == ''">false</IsUnitTestProject>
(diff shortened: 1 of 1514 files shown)


Directory.Build.props
```diff
@@ -1,4 +1,4 @@
-<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information. -->
```
davkean: This says "LICENSE" but the license is called "LICENSE.md" - can we make one of them in sync?

dmonroym: This is equal to what's on roslyn and runtime where they have License.txt instead, should we change our file from md to txt?

davkean: Well they are both wrong too, this is a copy and paste error, it used to be called "LICENSE".

davkean: https://github.com/dotnet/project-system/commit/167b0b83c8108df757989f15b03a8db9a795efd4

dmonroym: I'll change the file name in the comments then and file an issue with the other repos so they do the same. Just to understand the context, why'd we switch to .md?

davkean: Cos it looked better.

davkean: @dmonroym I'm fixing this.

davkean:
I've grabbed your commits and thrown them in this PR, fixed the reference to LICENSE.md and enforced it: https://github.com/dotnet/project-system/pull/5965.

PR #5987, opened 2020-03-16
Persist nullability into PublicAPI.XXX.txt

davkean:
This required upgrading analyzers, which also picked up some bugs resulting in "new" API.


src/Microsoft.VisualStudio.AppDesigner/PublicAPI.Shipped.txt
```diff
@@ -1,3 +1,4 @@
+#nullable enable
```
davidwengier: This makes complete sense, but is also really weird to see 😃

davkean: ¯\_(ツ)_/¯

davkean:
This required a huge song and dance to get it to build on the build server. Needed to upgrade the compilers too.

Issue #5918, opened 2020-02-26
 Fast up-to-date check doesn't trigger build when app.config modified


anlau:
**Visual Studio Version**:

Version 16.4.5

**Summary**:

We recently migrated a .NET Framework WPF application to SDK-style projects and I encountered the following behavior. 

**Steps to Reproduce**:

1. Create a new WPF project: `dotnet new wpf` (`dotnet -v` is 3.1.101)

2. Modify created `csproj` to target `net452` (does not seem to matter, as long it is .NET Framework)

3. Add an `app.config` file with minimal content (like an `appSettings` section with one entry)

4. Build twice, the second time should output "1 up-to-date"

5. Modify added `app.config` (ex. change existing entry key or value)

6. Build

**Expected Behavior**:

Fast up-to-date check triggers a build.

**Actual Behavior**:

```txt
...
1>FastUpToDate: Project is up to date. (MyWpfProject)
1>FastUpToDate: Up to date check completed in 2,9 ms (MyWpfProject)
========== Build: 0 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========
```

**User Impact**:

Not much; something which can be surprising or maybe annoying for developers (the build have to be forced by modifying something else).

---

When looking at the "Up to date checks" output, it seems an intermediate file is used as the source for the generated `.exe.config`:

```txt
1>FastUpToDate: Checking copied output (UpToDateCheckBuilt with Original property) file 'C:\Users\Antoine\Desktop\MyWpfProject\obj\Debug\net452\MyWpfProject.exe.withSupportedRuntime.config': (MyWpfProject)
1>FastUpToDate:     Source 2020-02-25 23:28:32: 'C:\Users\Antoine\Desktop\MyWpfProject\obj\Debug\net452\MyWpfProject.exe.withSupportedRuntime.config'. (MyWpfProject)
1>FastUpToDate:     Destination 2020-02-25 23:28:32: 'C:\Users\Antoine\Desktop\MyWpfProject\bin\Debug\net452\MyWpfProject.exe.config'. (MyWpfProject)
```

If `<GenerateSupportedRuntime>false</GenerateSupportedRuntime>` is added to the csproj, everything works as expected (source is now `C:\Users\Antoine\Desktop\MyWpfProject\App.config`).

Adding the following `UpToDateCheckBuilt` to the csproj seems to work too.

```xml
<ItemGroup>
[...]

PR #5994, opened 2020-03-17
Fix fast up-to-date check to detect app.config

ocallesp:
This fixes https://github.com/dotnet/project-system/issues/5918
App.config was not taking into consideration in FUTD. The issue is that App.config was using the intermediate filename as a source.

This fix enables FUTD to check App.config.
The item AppConfigWithTargetPath now gets the application configuration filename from the property  AppConfig.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -421,7 +421,11 @@
       <UpToDateCheckBuilt Condition="'$(_DebugSymbolsProduced)'=='true' and '$(SkipCopyingSymbolsToOutputDirectory)' != 'true' and '$(CopyOutputSymbolsToOutputDirectory)' != 'false'" Include="@(_DebugSymbolsOutputPath)"/>
 
       <!-- app.config -->
-      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="@(AppConfigWithTargetPath)"/>
+      <!-- The property AppConfig, created in PrepareForBuild, is used instead of AppConfigWithTargetPath because GenerateSupportedRuntime
+           rewrites AppConfigWithTargetPath to point to the intermediate filename. This is needed because Fast up-to-date needs to compare
+           the timestamp of the source filename (AppConfig) with destination filename. 
+           https://github.com/microsoft/msbuild/blob/64c8e4a18e7cf7f064fbad304ea7ed877cdaa0a1/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1109-L1122 -->
+      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="$(AppConfig)"/>
     </ItemGroup>
   </Target>
 


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -421,7 +421,7 @@
       <UpToDateCheckBuilt Condition="'$(_DebugSymbolsProduced)'=='true' and '$(SkipCopyingSymbolsToOutputDirectory)' != 'true' and '$(CopyOutputSymbolsToOutputDirectory)' != 'false'" Include="@(_DebugSymbolsOutputPath)"/>
 
       <!-- app.config -->
-      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="@(AppConfigWithTargetPath)"/>
+      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="%(AppConfigWithTargetPath.OriginalItemSpec)"/>
```
drewnoakes: Is `OriginalItemSpec` always set on these items?

ocallesp: if the configuration files is different than "App.config", the design-time build only will show the correct file name in PrimaryList, and CandidateAssemblyFiles. How to get access to configuration name ? It seems that those variables are not visible for  CollectUpToDateCheckBuiltDesingTime?
The image below is using App66.config 
image

ocallesp: If we add App1.config, App2.config and App.config only the App.config is taking into consideration to generate the Output file. It seems that we should only care about App.config filename

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -421,7 +421,7 @@
       <UpToDateCheckBuilt Condition="'$(_DebugSymbolsProduced)'=='true' and '$(SkipCopyingSymbolsToOutputDirectory)' != 'true' and '$(CopyOutputSymbolsToOutputDirectory)' != 'false'" Include="@(_DebugSymbolsOutputPath)"/>
 
       <!-- app.config -->
-      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="@(AppConfigWithTargetPath)"/>
+      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="@(_ResolveAssemblyReferencesApplicationConfigFileForExes)"/>
     </ItemGroup>
```
davkean: `_ResolveAssemblyReferencesApplicationConfigFileForExes` is a private field (hence the underscore), we can't consume this. We need to introduce a handshake between MSBuild and ourselves if we don't have the right extension. How does legacy project system handle this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -421,7 +421,7 @@
       <UpToDateCheckBuilt Condition="'$(_DebugSymbolsProduced)'=='true' and '$(SkipCopyingSymbolsToOutputDirectory)' != 'true' and '$(CopyOutputSymbolsToOutputDirectory)' != 'false'" Include="@(_DebugSymbolsOutputPath)"/>
 
       <!-- app.config -->
-      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="@(AppConfigWithTargetPath)"/>
+      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="$(AppConfig)"/>
```
rainersigwald: This looks correct to me but I think it deserves a comment and a pointer to both the SDK logic that rewrites `@(AppConfigWithTargetPath)` and the common.targets logic that sets it initially and also sets this property

https://github.com/microsoft/msbuild/blob/3453beee039fb6f5ccc54ac783ebeced31fec472/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1109-L1122

drewnoakes:
LGTM. @rainersigwald does this look right to you?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
igitur: @callesp I'm still experiencing issues with FUTD checking and would like to log a new issue. What tool did you use to generate the [image](https://user-images.githubusercontent.com/510598/76923878-e7662880-6891-11ea-9365-fc6540f32b47.png)? I'd like to do the same when I submit the issue. Thanks.

drewnoakes: @igitur that screenshot is from the msbuild structured log viewer, showing a binlog captured via the Project System Tools extension.

PR #5999, opened 2020-03-19
Veto loading a project with missing capabilities

davkean:
This veto loads a project with missing capabilities to catch situations such as https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1083914/.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectCapabilitiesMissingVetoProjectLoad.ProjectType.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectCapabilitiesMissingVetoProjectLoad.ProjectType.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    internal partial class ProjectCapabilitiesMissingVetoProjectLoad
+    {
+        private class ProjectType
+        {
+            public ProjectType(string extension, IEnumerable<string> capabilities)
+            {
+                Extension = extension;
+                Capabilities = capabilities.ToImmutableArray();
+            }
+
+            public string Extension;
+            public ImmutableArray<string> Capabilities;
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectCapabilitiesMissingVetoProjectLoad.cs
```diff
@@ -0,0 +1,83 @@
...
+        private readonly IProjectCapabilitiesService _projectCapabilitiesService;
+
+        [ImportingConstructor]
+        public ProjectCapabilitiesMissingVetoProjectLoad(UnconfiguredProject project, IProjectCapabilitiesService projectCapabilitiesService)
+        {
+            _project = project;
+            _projectCapabilitiesService = projectCapabilitiesService;
+        }
+
+        public Task<bool> AllowProjectLoadAsync(bool isNewProject, ProjectConfiguration activeConfiguration, CancellationToken cancellationToken = default)
```
drewnoakes: This will never return false.

davkean: Yes.

drewnoakes: Never mind. I see the interface's docs.

drewnoakes: I didn't realise this implemented an interface. I was hinting at changing the return type.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ProjectCapabilitiesMissingVetoProjectLoad.ProjectType.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    internal partial class ProjectCapabilitiesMissingVetoProjectLoad
+    {
+        private class ProjectType
```
drewnoakes: Could be a `readonly struct`.

davkean: Could be but would result in basically unneeded JIT code gen for basically 3 instances on the heap.

PR #6001, opened 2020-03-19
Package diagnostics from assets file

### LINKED ISSUE #3780: Diagnostics not shown under packages in the dependencies node
``` XML

  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="FooBar" Version="1.0.0" />
  </ItemGroup>
</Project>

```

Expected: For the error to be listed under "FooBar" within the dependency node and for the icon to be an error.

Actual:

image

drewnoakes:
NOTE this PR is against a new feature branch `feature/dependencies-tree-performance`. 

This PR will fix #3780 once the feature branch is merged into master.

---

As a first step towards populating transitive references in the dependencies tree directly from the `project.assets.json` file (the assets file), this PR introduces:

- `AssetsFileDependenciesDataSource` &mdash; a dataflow source that monitors the assets file for changes and initiates parsing.
- `AssetsFileDependenciesSnapshot` &mdash; an immutable snapshot of data from the assets file, initially containing only diagnostic messages
- `DependenciesAttachedCollectionSourceProvider` &mdash; attaches children to items in the dependencies tree
- `AttachedCollectionItemBase` &mdash; contains reusable logic/patterns for populating tree items (text/icons/tool tips/etc)
- `BrowseObjectBase`, `BrowseObjectDisplayNameAttribute`, `BrowseObjectDescriptionAttribute` &mdash; make it easy to populate a browse object for an item
- `HierarchyItemFlagsDetector` &mdash; identifies CPS flags on items, without allocating
- `IVsHierarchyItemExtensions` &mdash; determines target/package id/package version from items

---

### Package with invalid version

image

image

### Package with no version

image

image

Note that in this second case, `project.assets.json` contains a diagnostic on transitive dependency `XmpCore` too. This PR won't show it, as those nodes are still being shown via progression. This will be addressed soon, probably in the next PR.

This is a rare case where we want to _add_ yellow triangles to package nodes 😄


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AttachedCollectionItemBase.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public string ToolTipText => Text;
+
+        public bool IsCut => false;
+
+        public virtual object? GetBrowseObject() => null;
+
+        public TPattern? GetPattern<TPattern>() where TPattern : class
+        {
+            if (s_supportedPatterns.Contains(typeof(TPattern)))
```
davidwengier: Is this even worth doing? Just `return this as TPattern` will have the same effect, right? Or do you specifically want to return false for `ITreeDisplayItemWithImages`? If so, perhaps the comment above should be more strongly worded than "we don't have to support"

drewnoakes: Possibly not needed. This PR is a stripped down version of work I have locally in which this method behaves differently when the item is disposed (in which case you want to support a subset of patterns). There's a chance that won't be needed in the end.

Most (all?) [other implementations of this member](http://index/?leftProject=Microsoft.VisualStudio.Shell.Framework&leftSymbol=a2bwcgbamcgc&rightProject=Microsoft.Ide.ProjectSystem&file=Hierarchy%5CHierarchyItem.cs&line=328) call the patterns out explicitly.

davidwengier: You pre-empted the only comment I have on this PR 😛

build/import/Packages.targets
```diff
@@ -126,8 +126,11 @@
     <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="3.0.0-beta2.20059.3+77df2220" />
 
     <!-- NuGet -->
-    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6505" />
-    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6505" />
+    <!-- TODO undo version rollback (was needed as VS ships this lower version) -->
+    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6489" />
+    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6489" />
+    <PackageReference Update="NuGet.ProjectModel"                                                     Version="5.6.0-preview.2.6489" />
+    <PackageReference Update="NuGet.Common"                                                           Version="5.6.0-preview.2.6489" />
```
drewnoakes: @davkean you upgraded this recently. Do you see a problem with this slight downgrade?

davkean: Does it build? If so, this is fine.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/IVsHierarchyItemExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Detects the configuration string associated with a given hierarchy item in the dependencies tree, if
+        /// nested within a target group node. For projects that do not multi-target, this will always return <see langword="false"/>.
+        /// This method searches ancestors until a configuration is found, or the project root is found.
+        /// </summary>
+        /// <param name="item">The item to test for.</param>
+        /// <param name="configuration">The detected configuration, if found.</param>
+        /// <returns><see langword="true"/> if the configuration was found, otherwise <see langword="false"/>.</returns>
+        public static bool TryFindConfiguration(
```
davkean: This looks like it finds a dimension, not a configuration.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractItemFileWatchDataSource.cs
```diff
@@ -39,10 +39,11 @@ protected override IDisposable LinkExternalInput(ITargetBlock<IProjectVersionedV
             IProjectValueDataSource<IProjectSubscriptionUpdate> source = _projectSubscriptionService.ProjectRuleSource;
 
             // Transform the changes from evaluation -> FileWatchData
-            DisposableValue<ISourceBlock<IProjectVersionedValue<FileWatchData>>> transformBlock = source.SourceBlock
-                                                                                                        .TransformWithNoDelta(update => update.Derive(u => CreateFileWatch(u.CurrentState)),
-                                                                                                            suppressVersionOnlyUpdates: true,
-                                                                                                            ruleNames: new[] { ItemSchemaName });
+            DisposableValue<ISourceBlock<IProjectVersionedValue<FileWatchData>>> transformBlock
+                = source.SourceBlock.TransformWithNoDelta(
```
davkean: What changed here?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Assets/AssetsFileDependenciesDataSource.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                AssetsFileDependenciesSnapshot snapshot = lastSnapshot;
+
+                if (subscriptionUpdate.CurrentState.TryGetValue(ConfigurationGeneral.SchemaName, out IProjectRuleSnapshot ruleSnapshot))
+                {
+                    if (ruleSnapshot.Properties.TryGetValue(ConfigurationGeneral.ProjectAssetsFileProperty, out path))
+                    {
+                        if (!string.IsNullOrEmpty(path))
+                        {
+                            if (projectSnapshot.AdditionalDependentFileTimes == null || !projectSnapshot.AdditionalDependentFileTimes.TryGetValue(path, out timestampUtc))
```
davkean: Should never be null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Assets/AssetsFileDependenciesDataSource.cs
```diff
@@ -0,0 +1,112 @@
...
+                                {
+                                    timestampUtc = File.GetLastWriteTimeUtc(path);
+                                }
+                                catch
+                                {
+                                    // ignore
+                                }
+                            }
+
+                            if (path != lastAssetsFilePath || timestampUtc != lastTimestampUtc)
```
davkean: Be aware that this logic can result in a timestamp that regresses (if file is touched 3 times and evaluation skips the first touch, picks up the second, but you manually read the 3rd), however, the logic seems okay in the face of that.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Assets/AssetsFileDependenciesSnapshot.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Produces an updated snapshot by reading the <c>project.assets.json</c> file at <paramref name="path"/>.
+        /// If the file could not be read, or no changes are detected, the current snapshot (this) is returned.
+        /// </summary>
+        public AssetsFileDependenciesSnapshot UpdateFromAssetsFile(string? path)
+        {
+            try
+            {
+                // Parse the file
+                using var fileStream = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 4096 * 10, FileOptions.SequentialScan);
```
davkean: I think you want to combine FileShare.Delete here to avoid blocking folks from deleting the file while you are reading it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Assets/AssetsFileDependenciesSnapshot.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                LockFile lockFile = s_lockFileFormat.Read(fileStream, path);
+
+                return new AssetsFileDependenciesSnapshot(lockFile);
+            }
+            catch (IOException)
+            {
+                return this;
```
davkean: This can throw UnauthorizedAccessException.

davkean:
This is great, very clean and easy to read.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/IVsHierarchyItemExtensions.cs
drewnoakes: It's the target configuration. Would you call this `TryFindTarget` instead?

drewnoakes: @davidwengier about a week ago I mentioned how the `$TFM:Foo` flag was not consumed by the Project System. FYI this PR changes that, and introduces new special flags for package ID/version.

davkean: What makes this a "configuration"? It's returns the value of the "target framework" dimension, which is a part of the configuration.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Assets/AssetsFileDependenciesSnapshot.cs
drewnoakes: Good idea. I never knew this was a `[Flags]` enum!

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/Assets/AssetsFileDependenciesDataSource.cs
drewnoakes: Added some comments around this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/AbstractItemFileWatchDataSource.cs
drewnoakes: Just formatting (f8a93d3ad218bf676c57defb266471b34b9c49da). I meant to pull this out to a different PR.

PR #6002, opened 2020-03-19
Fix annotation warning

drewnoakes:
`TryGetProjectItemPathFromDeployedPath` can output `null` when returning `true`.


davkean:
Bit of a weird API, eh?

drewnoakes:
> Bit of a weird API, eh?

Yep. At least NRT lets us catch incorrect usages.

PR #6005, opened 2020-03-20
WIP: Localization cleanup + validation

melytc:
This PR solves #4944 .

The main objective is to make sure we are localizing only the rules that end up in the UI. To do so, I did two main things:
1. Stopped localizing rules that don't show up (first couple of commits).
2. Created a unit test to validate localization in project file, targets files and rule files*.

The impact this work will have is that we will only have the needed localized rules in our repo, thus ensuring that the work we pass to the localization team is actually needed; and a validation for future new rule files.

** Localization is set in the project file and the corresponding targets file, and then with attributes in the specific rule file.

#### What the unit test does:
1. Adds all the rule file names found in the project file (`XamlPropertyRule`, `XamlPropertyRuleNoCodeBehind`, `XamlPropertyProjectItemsSchema`) with a boolean to know if it's set to be localized ( `XlfInput` value) into a dictionary.
2. Verifies the localization set in the 4 targets files (managed, C#, F# and VB). For each rule file found in the target file:
a. Checks that the rule is also found in the project file.
b. Checks that the rule is set to be localized consistently in the project file and targets file.
c. Checks that rules that contain `PropertyPage`, `ProjectDebugger`, `ProjectItemsSchema` or have a `BrowseObject` context are **localized**; and that the rest, are not.
3. [WIP] Verifies that each rule file has/does not have localization attributes (like `Visible`, `DisplayName`, `Description` and `Category`).

#### Missing work:
- [ ] There was some feedback that `Resource.BrowseObject.xaml` should be deleted from the targets file. Should I also delete it from disk? @davkean 
- [ ] Verify that the rule file has/does not have localization attributes. Figuring out how to get the rule file's path (some are found in /`Rules/`, `/Rules/Items/`, `/Rules/PropertyPages/`).

#### Follow up items:
- Split `Folder.xaml` into BrowseObject and File.
- Renaming several rules based on feedback:
a. `ConfigurationGeneral.xaml` to `Project.xaml`.
b. `General.BrowseObject.xaml` to `Project.BrowseObject.xaml`.
c. `GeneralConfiguredBrowseObject.xaml` to `Configuration.BrowseObject.xaml`.


melytc:
This is how errors would be shown in the Test Explorer:
image

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/MiscellaneousRuleTests.cs
```diff
@@ -2,15 +2,246 @@
...
+        {
+            XElement? root = LoadXamlRule(designTimeTargetFile, out XmlNamespaceManager? namespaceManager);
+            var propertyPageSchemas = root.XPathSelectElements(@"/project:Project/project:ItemGroup/project:PropertyPageSchema", namespaceManager);
+
+            foreach (XElement element in propertyPageSchemas)
+            {
+                string? context = element.XPathSelectElement("./project:Context", namespaceManager)?.Value;
+                string resourceDirectory = element.Attribute("Include").Value;
+
+                Regex rx = new Regex(@"\).*\.xaml", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
drewnoakes: This RegEx pattern looks suspicious. I think it'll match `"\\)foo.xaml"` (i.e. starting with a backslash and a closing parenthesis).

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/MiscellaneousRuleTests.cs
```diff
@@ -2,15 +2,246 @@
...
+            var XamlPropertyRules = root.XPathSelectElements(@"/Project/ItemGroup/XamlPropertyRule");
+            Regex rx = new Regex(@"(\\)(?!.*\\)(.+\.xaml)$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+            foreach (XElement element in XamlPropertyRules)
+            {
+                string rulePath = element.Attribute("Include").Value;
+                string? xlfInput = element.XPathSelectElement("./XlfInput")?.Value;
+                
+                Match match = rx.Match(rulePath);
+                string ruleName = match.Value.Substring(1);
```
drewnoakes: You may like to use [named capture groups](https://docs.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#named_matched_subexpression) here. RegEx supports pulling groups out of matches. You can name the bit of the pattern you want to match, then pull it out by name with something like `match.Groups["mypattern"].Value`.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/MiscellaneousRuleTests.cs
```diff
@@ -2,15 +2,246 @@
...
+                string? context = element.XPathSelectElement("./project:Context", namespaceManager)?.Value;
+                string resourceDirectory = element.Attribute("Include").Value;
+
+                Regex rx = new Regex(@"\).*\.xaml", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+                Match match = rx.Match(resourceDirectory);
+                string ruleName = match.Value.Substring(1);
+                bool alreadySeen = false;
+
+                // To do: How to know the full path. Some are under Rules, Rules/Items, Rules/PropertyPages
+                //string fullPath = Path.Combine(RepoUtil.FindRepoRootPath(), "src", "Microsoft.VisualStudio.ProjectSystem.Managed", "ProjectSystem", "Rules", ruleName);
```
drewnoakes: Can this go the other way, so you start with the path, then look up the file name? The other way around may be harder.

I expect it's an error to have two rules in different folders but with the same name, as they seem to be merged during build. If you loop through the files first, you could also verify there are no duplicate file names. If that'd be a headache to refactor to, I wouldn't worry.

davkean:
@melytc Is this anything blocking this from being worked on?

melytc:
> @melytc Melissa Trevino FTE Is this anything blocking this from being worked on?

Not really, just got other priority work and paused it. I'll break it down and have what's ready merged.

melytc:
Closing this PR; will open a new PR where work done on #6331 is considered, plus the additional code suggestions made in here.

PR #6013, opened 2020-03-23
Add Fast up-to-date the capability to detect project.assets.json

### LINKED ISSUE #6006: Up-to-date check should consider AdditionalDependentFileTimes file creation/deletion
`AdditionalDependentFileTimes` includes files such as `.editorconfig` and `global.json`. Actually it includes all _potential_ locations that one of these files may exist at. Instances which are present on disk have a valid time stamp. Others have some filesystem-specific 'zero' value (maybe 1900-01-01?).

Changes to the set of paths for which we have a valid time stamp should mark the project dirty.

Unfortunately these files do not guarantee that outputs are updated, so we may end up in a case where a project remains out-of-date until an input actually changes.

[AB#1083953](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1083953)

ocallesp:
Fixes #6006
Fixes [AB#1083953](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1083953)

The bug is that Fast up-to-date cannot detect when project.assets.json is deleted from the filesystem.
In case project.assets.json is missing FUTD will consider the project out of date.

This implementation take the output from Evaluation and keep a list (LastAdditionalDependentFileTimes) of the files in ApplicationDependentFileTimes that are present in the filesystem. In case projec.assets.json is deleted, then it compares the LastAdditionalDependentFileTimes with the current existing files.
In case FUTD is enabled, then it will check AdditionalDependentFilesAddedOrRemoved  to decide if the project is out of date or not.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -104,6 +104,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             /// </list>
             /// <para>
             /// We need all <em>potential</em> paths for files which may start affecting build if they were to be added.
+            /// Any files not found on disk have MinValue dates.
             /// </para>
             /// </summary>
             /// <remarks>
@@ -112,12 +113,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             public IImmutableDictionary<string, DateTime> AdditionalDependentFileTimes { get; }
 
             /// <summary>
-            /// Gets the time at which the set of items in additionalDependentFileTimes changed (files added or removed).
+            /// Gets the time at which the set of items with non-<see cref="DateTime.MinValue"/> times
+            /// in <see cref="AdditionalDependentFileTimes"/> changed (files added or removed).
             /// </summary>
-            /// <remarks>
-            /// Additional dependent files are not required for compilation, but if present they might affect the compilation
-            /// output. e.g global.json, project.assets.json.
-            /// </remarks>
             public DateTime LastAdditionalDependentFileTimesChangedAtUtc { get; }
 
             private State()
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -110,7 +110,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             /// The <see cref="DateTime"/> values here do not update dynamically.
             /// </remarks>
             public IImmutableDictionary<string, DateTime> AdditionalDependentFileTimes { get; }
-            private ImmutableHashSet<string> LastAdditionalDependentFileTimes { get; }
             public DateTime LastAdditionalDependentFileTimesChangedAtUtc { get; }
 
             private State()
@@ -131,8 +130,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                 ResolvedCompilationReferencePaths = emptyPathSet;
                 CopyReferenceInputs = emptyPathSet;
                 AdditionalDependentFileTimes = ImmutableDictionary.Create<string, DateTime>(StringComparers.Paths);
-                LastAdditionalDependentFileTimes = ImmutableHashSet.Create<string>();
-                LastAdditionalDependentFileTimesChangedAtUtc = DateTime.MinValue; ;
+                LastAdditionalDependentFileTimesChangedAtUtc = DateTime.MinValue;
             }
 
             private State(
@@ -153,7 +151,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                 ImmutableHashSet<string> resolvedCompilationReferencePaths,
                 ImmutableHashSet<string> copyReferenceInputs,
                 IImmutableDictionary<string, DateTime> additionalDependentFileTimes,
-                ImmutableHashSet<string> lastAdditionalDependentFileTimes,
                 DateTime lastAdditionalDependentFileTimesChangedAtUtc,
                 DateTime lastItemsChangedAtUtc,
                 DateTime lastCheckedAtUtc)
@@ -177,7 +174,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                 LastItemsChangedAtUtc = lastItemsChangedAtUtc;
                 LastCheckedAtUtc = lastCheckedAtUtc;
                 AdditionalDependentFileTimes = additionalDependentFileTimes;
-                LastAdditionalDependentFileTimes = lastAdditionalDependentFileTimes;
                 LastAdditionalDependentFileTimesChangedAtUtc = lastAdditionalDependentFileTimesChangedAtUtc;
 
                 var setNames = new HashSet<string>(s_setNameComparer);
@@ -204,10 +200,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                 string? outputRelativeOrFullPath = jointRuleUpdate.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.OutDirProperty, msBuildProjectOutputPath);
                 string msBuildAllProjects = jointRuleUpdate.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
 
+                var lastExistingAdditionalDependentFiles = AdditionalDependentFileTimes.Where(pair => pair.Value != DateTime.MinValue).Select(pair => pair.Key).ToImmutableHashSet();
                 IImmutableDictionary<string, DateTime> additionalDependentFileTimes = projectSnapshot.AdditionalDependentFileTimes;
+
                 var currentExistingAdditionalDependentFiles = additionalDependentFileTimes.Where(pair => pair.Value != DateTime.MinValue).Select(pair => pair.Key).ToImmutableHashSet();
-                bool AdditionalDependentFilesChanged = !LastAdditionalDependentFileTimes.SetEquals(currentExistingAdditionalDependentFiles);
-                var lastAdditionalDependentFileTimes = AdditionalDependentFilesChanged ? currentExistingAdditionalDependentFiles : LastAdditionalDependentFileTimes;
+                bool AdditionalDependentFilesChanged = !lastExistingAdditionalDependentFiles.SetEquals(currentExistingAdditionalDependentFiles);
                 DateTime lastAdditionalDependentFileTimesChangedAtUtc = AdditionalDependentFilesChanged ? DateTime.UtcNow : LastAdditionalDependentFileTimesChangedAtUtc;
 
                 // The first item in this semicolon-separated list of project files will always be the one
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -110,6 +110,8 @@ internal sealed class State
             /// The <see cref="DateTime"/> values here do not update dynamically.
             /// </remarks>
             public IImmutableDictionary<string, DateTime> AdditionalDependentFileTimes { get; }
+            public ImmutableHashSet<string> LastAdditionalDependentFileTimes { get; set; }
+            public bool AdditionalDependentFilesAddedOrRemoved;
```
drewnoakes: The state object must be immutable. That means the property cannot have a setter, and the field must be a get-only property.

drewnoakes: These two fields are redundant as the bool can be computed in the state update method without requiring memory of the last set.

I think it would be better to remove these new properties and store `LastAdditionalDependentFileTimesChangedAtUtc` instead, matching the pattern used in `LastItemsChangedAtUtc`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -170,6 +170,11 @@ private bool CheckGlobalConditions(Log log, State state)
                 return log.Fail("CopyAlwaysItemExists", "Item '{0}' has CopyToOutputDirectory set to 'Always', not up to date.", _configuredProject.UnconfiguredProject.MakeRooted(copyAlwaysItemPath));
             }
 
+
+            if (state.AdditionalDependentFilesAddedOrRemoved)
```
drewnoakes: The comparison should be against the last output time. See the existing check against `LastItemsChangedAtUtc` in `CheckInputsAndOutputs`.

ocallesp:
Test Explorer from my local repository passes all unit-tests for IsUpToDateAsync_False_AdditionalDependentFileTimeSourceRemoved(), but in GitHub this test fails.
 I execute all failed tests.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -110,6 +110,8 @@ internal sealed class State
             /// The <see cref="DateTime"/> values here do not update dynamically.
             /// </remarks>
             public IImmutableDictionary<string, DateTime> AdditionalDependentFileTimes { get; }
+            private ImmutableHashSet<string> LastAdditionalDependentFileTimes { get; }
```
drewnoakes: I'm pretty sure you don't need this field. Just use `AdditionalDependentFileTimes` instead. When updating the state, the last times are the ones of the old snapshot, the new ones are available. Of they've changed, update `LastAdditionalDependentFileTimesChangedAtUtc`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -129,6 +131,8 @@ private State()
                 ResolvedCompilationReferencePaths = emptyPathSet;
                 CopyReferenceInputs = emptyPathSet;
                 AdditionalDependentFileTimes = ImmutableDictionary.Create<string, DateTime>(StringComparers.Paths);
+                LastAdditionalDependentFileTimes = ImmutableHashSet.Create<string>();
+                LastAdditionalDependentFileTimesChangedAtUtc = DateTime.MinValue; ;
```
drewnoakes:
```suggestion
                LastAdditionalDependentFileTimesChangedAtUtc = DateTime.MinValue;
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1117,9 +1117,6 @@ public async Task IsUpToDateAsync_False_AdditionalDependentFileTimeSourceRemoved
             _buildUpToDateCheck.TestAccess.SetLastCheckedAtUtc(lastCheckTime);
             _buildUpToDateCheck.TestAccess.SetLastItemsChangedAtUtc(itemChangeTime);
 
-            await AssertUpToDateAsync(
```
davkean: Why was this removed?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -170,6 +164,11 @@ private bool CheckGlobalConditions(Log log, State state)
                 return log.Fail("CopyAlwaysItemExists", "Item '{0}' has CopyToOutputDirectory set to 'Always', not up to date.", _configuredProject.UnconfiguredProject.MakeRooted(copyAlwaysItemPath));
             }
 
+
+            if (state.LastAdditionalDependentFileTimesChangedAtUtc > state.LastCheckedAtUtc)
```
drewnoakes: We must avoid checks against `LastCheckedAtUtc` as they are not idempotent. If the user presses <kbd>Ctrl</kbd> + <kbd>F5</kbd> then the up-to-date check is invoked twice, and the second answer will be different to the first.

This comparison should be against the last build output timestamp. Move this next to the existing check against `LastItemsChangedAtUtc`.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1093,7 +1093,40 @@ public async Task IsUpToDateAsync_False_CopyToOutputDirectoryDestinationDoesNotE
...
+            var outputPath = @"C:\Dev\Solution\Project\Output";
+
+            await SetupAsync(
+                projectSnapshot: projectSnapshot,
+                dependentTimeFiles: new[] { (dependentPath, dependentTime) });
+
+            _fileSystem.AddFile(dependentPath, dependentTime);
+            _fileSystem.AddFile(outputPath, outputTime);
+            _buildUpToDateCheck.TestAccess.SetLastCheckedAtUtc(lastCheckTime);
+            _buildUpToDateCheck.TestAccess.SetLastItemsChangedAtUtc(itemChangeTime);
```
drewnoakes: You'll need `TestAccess.SetLastAdditionalDependentFileTimesChangedAtUtc` too.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1093,7 +1093,40 @@ public async Task IsUpToDateAsync_False_CopyToOutputDirectoryDestinationDoesNotE
...
+
+            // Remove dependentPath
+            await SetupAsync(projectSnapshot: projectSnapshot);
+
+            await AssertNotUpToDateAsync(
+                "Some additional dependent files are added or removed, not up to date.",
+                telemetryReason: "AdditionalDependentFiles");
```
drewnoakes: The message should be something like:

> The set of AdditionalDependentFiles was changed more recently ({localChangeTime}) than the earliest output '{outputPath}' ({localOutputTime}), not up to date.

And the telemetry reason should probably be `"Outputs"` to match other instances.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1133,7 +1197,7 @@ public async Task IsUpToDateAsync_True_AdditionalDependentFileTimeSourceIsOlderT
             var itemChangeTime = DateTime.UtcNow.AddMinutes(-4);
             var dependentTime = DateTime.UtcNow.AddMinutes(-3);
             var outputTime = DateTime.UtcNow.AddMinutes(-2);
-            var lastCheckTime = DateTime.UtcNow.AddMinutes(-1);
+            var lastCheckTime = DateTime.UtcNow.AddMinutes(3);
```
drewnoakes: This means the last check is in the future, which is impossible.

I don't think you should be touching the last check time in the implementation though, as commented above.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
ocallesp: I changed the unit-test because it failed on Github, but passed on my local repository.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1093,20 +1093,98 @@ public async Task IsUpToDateAsync_False_CopyToOutputDirectoryDestinationDoesNotE
...
+            var projectSnapshot = new Dictionary<string, IProjectRuleSnapshotModel>
+            {
+                [UpToDateCheckBuilt.SchemaName] = SimpleItems("BuiltOutputPath1")
             };
 
             var itemChangeTime = DateTime.UtcNow.AddMinutes(-4);
             var lastCheckTime = DateTime.UtcNow.AddMinutes(-3);
             var outputTime = DateTime.UtcNow.AddMinutes(-2);
             var dependentTime = DateTime.UtcNow.AddMinutes(-1);
+            var lastAdditionalDependentFileTimesChanged = DateTime.UtcNow.AddMinutes(0);
```
drewnoakes: `DateTime.UtcNow` only updates every 15ms or so. If you add zero minutes, your test will sometimes have `lastAdditionalDependentFileTimesChanged == DateTime.Now` and sometimes not, which can lead to test instability. It's safer to subtract one minute from all these times, so they're all definitely in the past.

drewnoakes: Once this is fixed I think the PR is good to go.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -110,6 +110,7 @@ internal sealed class State
             /// The <see cref="DateTime"/> values here do not update dynamically.
             /// </remarks>
             public IImmutableDictionary<string, DateTime> AdditionalDependentFileTimes { get; }
+            public DateTime LastAdditionalDependentFileTimesChangedAtUtc { get; }
```
drewnoakes: It would be good to add a comment to this property explaining why we need it. For example, this is actually the time at which the set of present items changed (not changing values).

The up to date check has some complex heuristics, so any information added here will likely be helpful in future.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
```diff
@@ -196,8 +200,13 @@ private State()
                 string? outputRelativeOrFullPath = jointRuleUpdate.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.OutDirProperty, msBuildProjectOutputPath);
                 string msBuildAllProjects = jointRuleUpdate.CurrentState.GetPropertyOrDefault(ConfigurationGeneral.SchemaName, ConfigurationGeneral.MSBuildAllProjectsProperty, "");
 
+                var lastExistingAdditionalDependentFiles = AdditionalDependentFileTimes.Where(pair => pair.Value != DateTime.MinValue).Select(pair => pair.Key).ToImmutableHashSet();
```
drewnoakes: Does `AdditionalDependentFileTimes` actually contain `DateTime.MinValue`, or is it `DateTime.FromFileTime(0)` or `DateTime.FromFileTimeUtc(0)`?

PR #6021, opened 2020-03-24
Remove graph nodes / progression

drewnoakes:
Previous graph node (progression) code performed two functions:

- constructing transitive dependencies in the tree as the user expands nodes
- supporting search through the graph

As part of this work, transitive dependencies are being displayed via `IAttachedCollection` APIs (upon which progression is implemented), as this gives us greater control over presentation, lifetime, interaction, etc. It also allows us to reduce the amount of retained memory, and contributes towards reducing the number of DLLs loaded into Visual Studio.

Search will be re-implemented in a future PR using the low-level `ISearchProvider` API. Again, progression implements its search through this interface, and again we will have opportunities for more control and a leaner implementation.


drewnoakes:
Force pushed an update that removes the `Microsoft.VisualStudio.GraphModel` package too.

drewnoakes:
Force pushed removal of more stuff, including another package reference.

PR #6024, opened 2020-03-24
WIP: Build page

melytc:
As I'm trying to add a build property page for VS Online, I keep getting build errors even if I followed Tom's instructions from this PR.

Created this draft PR for @tmeschter , but if anyone find out why those errors are getting triggered, I'd appreciate it.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
@@ -153,6 +153,14 @@
       <SubType>Designer</SubType>
       <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
     </XamlPropertyRule>
+
+    <Compile Update="ProjectSystem\Rules\PropertyPages\BuildPropertyPage.xaml.cs">
+      <DependentUpon>BuildPropertyPage.xaml</DependentUpon>
+    </Compile>
+    <XamlPropertyRule Include="ProjectSystem\Rules\PropertyPages\BuildPropertyPage.xaml">
+      <SubType>Designer</SubType>
+      <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
+    </XamlPropertyRule>
   </ItemGroup>
 
   <!-- Items -->
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
@@ -463,9 +463,6 @@
     <None Update="ProjectSystem\Rules\AdditionalDesignTimeBuildInput.xaml">
       <Generator>MSBuild:Compile</Generator>
     </None>
-    <None Update="ProjectSystem\Rules\PropertyPages\BuildPropertyPage.xaml">
-      <Generator>MSBuild:Compile</Generator>
-    </None>
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -455,6 +463,9 @@
     <None Update="ProjectSystem\Rules\AdditionalDesignTimeBuildInput.xaml">
       <Generator>MSBuild:Compile</Generator>
     </None>
+    <None Update="ProjectSystem\Rules\PropertyPages\BuildPropertyPage.xaml">
+      <Generator>MSBuild:Compile</Generator>
+    </None>
```
tmeschter: I think the project system added this for you automatically, but you shouldn't need it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Rule Name="Build"
+      Description="General"
+      DisplayName="Build"
+      PageTemplate="generic"
+      xmlns="http://schemas.microsoft.com/build/2009/properties" >
+
+  <Rule.Categories>
+    <Category Name="General"
```
davidwengier: Do you need a category if there is just one?

Guess it doesn't hurt to have either way ¯\\\_(ツ)_/¯

src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
davidwengier: We could/should add a `<None Remove />` glob to stop this from happening

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
melytc: Oh this rule file is a small part of the whole page. Just starting with small portions and making sure it works. :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
melytc: I removed it and seems like it doesn't generates it anymore, but will consider it if it comes back.

melytc:
I removed what @tmeschter suggested, and it got rid of one error form the Error List. Don't know why I still see these: 
image

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Rule Name="Build"
```
tmeschter: @mel Ah, here's the problem. When building dotnet/project-system we will generate a BuildPropertyPage.cs file to the obj directory, and it will contain a type with whatever name you provide here.

However, we already have a type named "Build", and that ends up conflicting with the one in the generated file. Rename this to "BuildPropertyPage" and that should take care of at least one of the errors.

drewnoakes: @melytc you could add a unit test to prevent this problem in future.

tmeschter: Well, if you run into this problem you won't be able to build, so the unit test probably doesn't help much. :-)

melytc: This was it, thanks Tom. Closing this PR.

Issue #6012, opened 2020-03-20
Dependencies tree should not collapse on update


drewnoakes:
If the Dependencies tree is expanded and an update occurs, the tree collapses.

The expansion state of tree nodes should only change in response to user requests.

PR #6031, opened 2020-03-25
Don't collapse dependencies tree items on updates

drewnoakes:
Will address #6012 once this feature branch is merged into master.

---

Previously, changed nodes were removed from the tree and re-added, resulting in different hierarchy IDs for the replacements and meaning that any expansion state was not restored.

This logic existed to work around a limitation of the graph node implementation.

We no longer use graph nodes in the dependencies tree so can now fix this.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesTreeViewProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesTreeViewProvider.cs
@@ -275,20 +275,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
                 IProjectTree? dependencyNode = rootNode.FindChildWithCaption(dependency.Caption);
                 bool isNewDependencyNode = dependencyNode == null;
 
-                if (dependencyNode != null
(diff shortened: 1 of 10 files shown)


drewnoakes:
Pushed an update that removes `ProjectTreeFlags.SupportsHierarchy`, which is no longer needed.

This flag indicated a node that could have graph node attached to it. When present, tree construction would force the node to be removed and re-added on resolved-state change, in order to trigger graph node updates.

We no longer use graph nodes, and we no longer remove/add nodes on updates.

This flag is public API. I cannot find usage within Microsoft code or in web search indexes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Tree/Dependencies/DependenciesTreeViewProvider.cs
```diff
@@ -275,20 +275,6 @@ IProjectTree RememberNewNodes(IProjectTree rootNode, IEnumerable<IProjectTree> n
                 IProjectTree? dependencyNode = rootNode.FindChildWithCaption(dependency.Caption);
                 bool isNewDependencyNode = dependencyNode == null;
 
-                if (dependencyNode != null
-                    && dependency.Flags.Contains(DependencyTreeFlags.SupportsHierarchy))
```
drewnoakes: This was the only place we consumed the `SupportsHierarchy` flag.

Issue #5740, opened 2019-12-16
Add package file/contained assembly paths to browse object


vsfeedback:
_This issue has a corresponding [ticket on Developer Community](https://developercommunity.visualstudio.com/idea/781589/cannot-see-the-path-for-a-dll-reference-via-packag.html). Please vote and comment there to make sure your voice is heard._

---
Hi,

After migrating to the `<PackageReference>` syntax, we cannot see the path of dll(s) in **Properties** tool window any more.

I know the default path of these nuget packages is `C:\Users\%UserName%\.nuget\packages`, but it would be more helpful if you can display the path in Properties tool window just as before.

Thank you!
Heku

davkean:
Triage: We provide this functionality in the new project system but we are not investing in this area in the legacy project system.

davkean:
Triage: This is a request to add the full path of the dll to the browse object of the graph nodes:

![image](https://user-images.githubusercontent.com/1103906/70951559-d43b9500-20b7-11ea-8d4c-e5b88a2e7e9a.png)

drewnoakes:
I guess there are two things here:

1. Show the path to the `.nupkg` file when a package is selected
1. Show the path to the `.dll` file when a DLL is selected

heku:
Thank you for consideration.

PR #6034, opened 2020-03-26
Support attached transitive package and project nodes

drewnoakes:
This PR extends the assets file snapshot to include data about packages and projects (collectively referred to as libraries). The snapshot is now split by target to make lookup more efficient.

Assets file snapshot types have been moved into their own "Models" namespace.

Packages have sub-nodes for displaying assembly groups (compile time assemblies, framework assemblies).

image

image

image

image


davkean:
I don't understand the "Compile Time Assemblies" under the project, is that the result of the project?

davkean:
It would be nice to have the "full path" in the browse object for all objects.

drewnoakes:
> I don't understand the "Compile Time Assemblies" under the project, is that the result of the project?

Yeah, it's just the output DLL. Not sure it makes sense either. I've removed it.

> It would be nice to have the "full path" in the browse object for all objects.

Done.

---

image

image

drewnoakes:
Fixed this for assemblies in packages too:

image

drewnoakes:
This is basically #5740.

PR #6037, opened 2020-03-26
Fix disposal of SolutionService

drewnoakes:
This type should be disposed when the solution is unloaded, but did not declare `IDisposable` in its interface list, so its `Dispose` method was never called.


davkean:
This occurs when VS is shut down, not when the solution closes.

Issue #5937, opened 2020-03-03
Implement a "Build" page back end for VS Online


tmeschter:
Now that the "Application" property page has been implemented for VS Online in #5920, the next step is to implement the "Build" property page.

tmeschter:
@melytc @jjmew I've created and assigned this issue to Melissa.

PR #6040, opened 2020-03-27
Add a Build property page for VS Online.

melytc:
PR for #5937.

**Current (offline) view:**
image

**VS Online view:**
image

The `<different options>` text is something that's bugging me. If I remove `SourceOfDefaultValue` from the `DataSource` it will appear as the screenshot below (but I'm not sure we would want to do that):
image


src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -153,6 +153,14 @@
...
       <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
     </XamlPropertyRule>
+    
+    <Compile Update="ProjectSystem\Rules\PropertyPages\BuildPropertyPage.xaml.cs">
+      <DependentUpon>BuildPropertyPage.xaml</DependentUpon>
+    </Compile>
+  <XamlPropertyRule Include="ProjectSystem\Rules\PropertyPages\BuildPropertyPage.xaml">
+    <SubType>Designer</SubType>
+    <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
+  </XamlPropertyRule>
```
drewnoakes: Indentation.

tmeschter:
In the screenshots the Configuration drop-down is set to "All Configurations", so any property that has different values in the Debug and Release configurations would show the "<different options>" text. That is expected.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -0,0 +1,106 @@
...
+
+    <Category Name="Output"
+              Description="Output"
+              DisplayName="Output" />
+  </Rule.Categories>
+
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFile"
+                SourceOfDefaultValue="AfterContext"
+                HasConfigurationCondition="False" />
```
tmeschter: This should be `HasConfigurationCondition="True"` since most of the properties on this page are configuration-specific. For the few that are not configuration-specific, you would tag the individual properties with `HasConfigurationCondition="False"`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -0,0 +1,106 @@
...
+  </Rule.DataSource>
+
+  <StringProperty Name="DefineConstants"
+                  DisplayName="Conditional compilation symbols"
+                  Category="General" />
+
+  <EnumProperty Name="PlatformTarget"
```
tmeschter: I noticed in the screenshots that the Platform and Nullable properties didn't show a value. I suspect this is because the .props/.targets don't provide an explicit default value, and so we can't get the default from MSBuild evaluation.

Can you check the documentation to see if there is a way to specify the default value here in the .xaml file?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -0,0 +1,106 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
```
drewnoakes: We've moved to MIT now. Was this copied from somewhere else that also needs to be changed?

melytc: Oh, that's right. I copied from the Application page, thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
melytc: Good catch, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
melytc: I tried with the EnumProperty's  `Default` property and the EnumValue's `IsDefault` property, but had no luck.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
melytc: Fixed, thanks.

melytc:
I'll go ahead and merge this PR. I'll continue looking for a way to set the default values.

PR #6045, opened 2020-03-30
Small bits

drewnoakes:
A collection of small things from the last few weeks.


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IVsUIService`1.cs
```diff
@@ -16,7 +16,7 @@ internal interface IVsUIService<T>
         where T : class?
     {
         /// <summary>
-        ///     Gets the service object associated with <typeparamref name="T"/>.
+        ///     Gets the service object associated with <typeparamref name="T"/>. Must be called from the UI thread.
```
davkean: You can now put this sort of thing in remarks now and it will show up.

Issue #4665, opened 2019-03-28
Consuming projects are not updated when a CopyToOutputDirectory item is updated in a project with no project references.


PathogenDavid:
**Visual Studio Version**: 2017 - 15.9.10

Also reproduced in 2019 RC.4 - 16.0.0

**Summary:**

Fast up-to-date checks will consider a project to be up-to-date even when a project it references is not up to date due to a stale `CopyToOutputDirectory` item.

(I've found some workarounds which are [described here](https://github.com/PathogenPlayground/TransitiveCopiedConentNotUpdated#workarounds).)

**Steps to Reproduce**:

1. [Clone this repository](https://github.com/PathogenPlayground/TransitiveCopiedConentNotUpdated).
    * The demonstration repository contains details on how the projects are set up. The most important thing to know is that `TextFileA.txt` and `TextFileB.txt` are copied to the output directory with `PreserveNewest` and the program prints their contents when it runs.
    * The repository also contains more details on my findings about the source of the issue.
    * The steps from the repository for reproducing the problem are reproduced below for posterity.
1. Open the solution in Visual Studio.
2. Select `TestProgram` as the startup project.
3. Press F5 to start the program. The output is as follows:
    ```
    Hello World!
    Hello from ClassA!
    Text file A reporting for duty!
    Hello from ClassB!
    Text file B reporting for duty!
    Done.
    ```
4. Close the program and add a message `TextFileA.txt` and press F5 to start the program again. The output includes your message:
    ```
    Hello World!
    Hello from ClassA!
    Text file A reporting for duty! Hello, world!
    Hello from ClassB!
    Text file B reporting for duty!
    Done.
    ```
5. Close the program and add a message to `TextFileB.txt` instead, press F5 to start the program again. **The output will not include your message:**
    ```
    Hello World!
    Hello from ClassA!
    Text file A reporting for duty! Hello, world!
    Hello from ClassB!
    Text file B reporting for duty!
    Done.
    ```
6. Observe that the file is updated in `LibraryB`'s output (`LibraryB\bin\Debug\netstandard2.0\TextFileB.txt`), but not in `TestProgram`'s output (`TestProgram\bin\Debug\netcoreapp2.2\TextFileB.txt`).

**Expected Behavior**:

`TextFileB.txt` is updated in `TestProgram`'s output folder when it is changed.

**Actual Behavior**:

`TextFileB.txt` is not updated.

**User Impact**:

[...]

PR #6046, opened 2020-03-30
[WIP] Fast up-to-date check CopyToOutputDirectory items from project references.

ocallesp:
This PR solves #4665 

Summary:
Fast up-to-date check will consider a project to be out of date when an item from a project reference is marked as CopyToOutputDirectory PreserveNewest changes.

The first approach to fix this was https://github.com/dotnet/project-system/pull/5923, but problematic dependencies of _SplitProjectReferencesByFileExistence.

Work needed:
- [ ] Investigate IDE performance of the fix for the msbuild issue  [4923 ](https://github.com/microsoft/msbuild/issues/4923)
- [ ] Add a new rule for the new target
- [ ] Modify FUTD to use the new rule


src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -515,6 +528,10 @@
     <DesignTimeTargetsFile Include="ProjectSystem\DesignTimeTargets\*.targets" />
   </ItemGroup>
 
+  <ItemGroup>
```
davkean, line 531: Remove

davkean:
@ocallesp I see the SDK change in your fork but not in the SDK, has this been signed off by the SDK/MSBuild? What happens when we're running against SDKs without this property set?

davkean:
Have we figured out the performance impact of this change?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -367,6 +367,9 @@
 
   </Target>
 
+  <!-- This target get all project items that may need to be transferred to the output directory for the up to date check. -->
```
davkean, line 370: Why does this target exist?

There's an existing target that already produces these items: https://github.com/dotnet/msbuild/blob/c10f34c4e00644e06120cb87f323c330f29a1e89/src/Tasks/Microsoft.Common.CurrentVersion.targets#L4732.

ocallesp:
> @ocallesp I see the SDK change in your fork but not in the SDK, has this been signed off by the SDK/MSBuild? What happens when we're running against SDKs without this property set?

The PR to set `MSBuildCopyContentTransitively ` to true was already merged in July https://github.com/dotnet/sdk/pull/12318/commits/86bae87a15f91fa2e97392656431b0d97ae1f4e8

If MSBuildCopyContentTransitively = False we will issues with _SplitProjectReferencesByFileExistence

> @ocallesp I see the SDK change in your fork but not in the SDK, has this been signed off by the SDK/MSBuild? What happens when we're running against SDKs without this property set?

ocallesp:
> Have we figured out the performance impact of this change?






> Have we figured out the performance impact of this change?

Yes. I was using the tool from Paul Vick. There should be no issues with this `MSBuildCopyContentTransitively `

src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
ocallesp: Do you mean that `GetCopyToOutputDirectoryItems `can return items that doesn't exist which will make FUTD to incorrectly validate a project is up-to-date ?

davkean: Let's step back, what is this target producing?

davkean:
> Yes. I was using the tool from Paul Vick. There should be no issues with this MSBuildCopyContentTransitively

Can you describe your methodology to test this? I would expect design-time builds times to be increased.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
ocallesp: this line was removed. 
This line was causing to override CopyToOutputDirectoryItem in design-time that is/was (don't know which executes first) also generated in MSBuild.

ocallesp: removed

ocallesp:
> > Yes. I was using the tool from Paul Vick. There should be no issues with this MSBuildCopyContentTransitively
> 
> Can you describe your methodology to test this? I would expect design-time builds times to be increased.

Compared the output from [Scale-Testing](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/11831/Scale-Testing) when MSBuildCopyContentTransitively = false/true. This tool creates copies of a project and load all of them in a solution. Then run the tests.

Two projects were used with Scale-Testing:
* A project created with 31 references https://github.com/ocallesp/sample_code

* And Roslyn project with Scale-Testing.

When using cache there is no difference, but when Not using cache there was little increase in 3 out of 5 runs.

PR #6047, opened 2020-03-31
Enable AppSettings project capability for WPF/WinForms

JeremyKuhne:
In Core the necessary ConfigurationManager dependency is always present for WinForms and WPF projects.

https://github.com/dotnet/winforms-designer/issues/356 is the WinForms designer issue for this.

I've manually tested this and everything seems fine, with or without a preexisting app.config.


davkean:
Does an App.Config actually work? It used to crash the launch of the project.

JeremyKuhne:
> Does an App.Config actually work? It used to crash the launch of the project.

@davkean It does, it needs the reference to System.Configuration.ConfigurationManager, which is always present for these projects. I spent some time a few weeks ago looking through the System.Configuration code (which I ported to Core) to help validate that the required bits and pieces around this were in place.

davkean:
@JeremyKuhne We're going to do manual testing on this before merging in - there are concerns that will load additional .NET Core types into the VS process.

davidwengier:
This worked fine in my quick test, and since the settings designer already works for SDK projects, I don't think this change introduces any new bugs.

PR #6048, opened 2020-03-31
Use default instead of default(T)

davkean:


drewnoakes:
`DateTime.MinValue` would be clearer and more consistent with the rest of the code/documentation.

PR #6064, opened 2020-04-03
Implement dependencies tree search

drewnoakes:
This PR introduces a functional version of dependencies tree search. Next week I will be working to consolidate the dual tree construction approaches (top-down during user expansion, vs. bottom-up during search) as part of the extensibility API design.

---

Solution Explorer search occurs via the `ISearchProvider` interface. When a search is requested, we pass back tree items that match the search expression for the tree to display. At that point we are unlikely to have materialised items ready to return, so we create just those nodes that match. The tree then uses the "contained by" relationship to populate ancestor nodes, with the expectation that an ancestor will eventually exist in the current tree. For search, these are the `IVsHierarchyItem` objects that back top-level dependencies.

This bottom-up approach to building the tree required some new interfaces and data structures. Items that implement `IContainedByAttachedItems` can report their ancestors to the tree during search, meaning they can appear in search results. The snapshot now has the ability to lazily create a bottom-up dependency map for constant-time queries of a given item's ancestors.

Dependency tree item types now have factories for constructing their instances with a given directionality (relationship) in mind.

`DependenciesSearchContext` acts as a context object for a single search operation and monitors things like cancellation and ensuring maximum result limits are respected.

`ResolvedPackageReference` rule properties `Dependencies` (which listed child dependencies) and `IsTopLevelDependency` (which is computed via different means) have been removed.

A 'legacy mode' is introduced for best-effort support for using older SDKs. Once 16.7 releases with SDK 3.1.4xx, the data exchange between DTB targets and the dependencies node for package references will change considerably. The legacy mode (documented in `PackageRuleHandler`) kicks in when it detects old-style data and makes the tree look as sensible as possible. Specifically, top-level dependencies should appear correctly, though transitive dependencies will be filtered out. There's also some swizzling to IDs necessary in legacy mode which will not be required going forward.

image

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -124,10 +124,10 @@
     <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="3.0.0-beta2.20059.3+77df2220" />
 
     <!-- NuGet -->
-    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6489" />
-    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6489" />
-    <PackageReference Update="NuGet.ProjectModel"                                                     Version="5.6.0-preview.2.6489" />
-    <PackageReference Update="NuGet.Common"                                                           Version="5.6.0-preview.2.6489" />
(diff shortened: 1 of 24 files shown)


drewnoakes:
The SDK-side of these changes is waiting for the 3.1.4xx branch to open. Until then it can be seen at https://github.com/dotnet/sdk/compare/master...drewnoakes:design-time-package-references.

PR #6066, opened 2020-04-06
Show packages' content files in tree

### LINKED ISSUE #603: Show content files from nuget packages as a distinct group in solution explorer
Content files from packages are either source files passed to the compiler or other files treated as content. 

In either case it would be good to show them in the solution explorer. One way to represent them might be similar to a readonly linked file (maybe a different icon to distinguish) along with the rest of the files. 

It may also be nice to show them under the package node itself but maybe unnecessary and maynot even be obvious.

drewnoakes:
Fixes #603.

Adds support for showing and searching a package's `contentFiles` items in the dependencies tree.

Apart from fixing an outstanding feature request, this PR serves as an example of what it would currently take to add support for a new item in the dependencies tree. My next unit of work is to improve the APIs required to do exactly that. My goal is to make adding a feature such as this simpler and more localised.

---

### In tree

image

### In search

image

### Browse object

image

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -124,10 +124,10 @@
     <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="3.0.0-beta2.20059.3+77df2220" />
 
     <!-- NuGet -->
-    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6532" />
-    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6532" />
-    <PackageReference Update="NuGet.ProjectModel"                                                     Version="5.6.0-preview.2.6532" />
-    <PackageReference Update="NuGet.Common"                                                           Version="5.6.0-preview.2.6532" />
+    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6554" />
+    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6554" />
+    <PackageReference Update="NuGet.ProjectModel"                                                     Version="5.6.0-preview.2.6554" />
+    <PackageReference Update="NuGet.Common"                                                           Version="5.6.0-preview.2.6554" />
 
     <!-- MSBuild -->
     <PackageReference Update="Microsoft.Build"                                                        Version="16.4.0"/>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AssetsFileAttachedCollectionSourceBase.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AssetsFileAttachedCollectionSourceBase.cs
@@ -25,6 +25,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedColl
         public event PropertyChangedEventHandler? PropertyChanged;
 
         private readonly IVsHierarchyItem _hierarchyItem;
+        private readonly IFileIconProvider _fileIconProvider;
         private ReplaceableCollection? _items;
 
         public bool IsUpdatingHasItems { get; private set; }
@@ -33,13 +34,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedColl
         public IEnumerable? Items => _items;
         public object SourceItem => _hierarchyItem;
 
-        protected AssetsFileAttachedCollectionSourceBase(IVsHierarchyItem hierarchyItem, IAssetsFileDependenciesDataSource dataSource, JoinableTaskContext joinableTaskContext)
+        protected AssetsFileAttachedCollectionSourceBase(IVsHierarchyItem hierarchyItem, IAssetsFileDependenciesDataSource dataSource, JoinableTaskContext joinableTaskContext, IFileIconProvider fileIconProvider)
         {
             Requires.NotNull(hierarchyItem, nameof(hierarchyItem));
             Requires.NotNull(dataSource, nameof(dataSource));
             Requires.NotNull(joinableTaskContext, nameof(joinableTaskContext));
 
             _hierarchyItem = hierarchyItem;
+            _fileIconProvider = fileIconProvider;
 
             IsUpdatingHasItems = true;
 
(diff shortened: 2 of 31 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/PackageContentFileItem.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            [BrowseObjectDisplayName(nameof(VSResources.PackageContentFileCodeLanguageDisplayName))]
+            [BrowseObjectDescription(nameof(VSResources.PackageContentFileCodeLanguageDescription))]
+            public string? CodeLanguage => _item._contentFile.CodeLanguage;
+
+            [BrowseObjectDisplayName(nameof(VSResources.PackageContentFileBuildActionDisplayName))]
+            [BrowseObjectDescription(nameof(VSResources.PackageContentFileBuildActionDescription))]
+            public string? BuildAction => _item._contentFile.BuildAction;
+
+            [BrowseObjectDisplayName(nameof(VSResources.PackageContentFilePathDisplayName))]
```
davidwengier, line 68: Wrong resource.. should be Copy To Output

davidwengier:
Nice! Not sure whether "code language" is something people will understand though? I say that because I don't understand it, given that your screenshot shows it as `cs` for a `.txt` file :P

drewnoakes:
> Nice! Not sure whether "code language" is something people will understand though? I say that because I don't understand it, given that your screenshot shows it as cs for a .txt file :P

It means that the file should be included for C# projects, not that it's a C# file.

I'll do my best to try and wordsmith this into the browse object's description. Most of the descriptions so far are borderline useless grammatical reformulations of the property name itself. Any improvements here appreciated before these hit the loc team.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/PackageContentFileItem.cs
drewnoakes: Thanks!

davidwengier:
> It means that the file should be included for C# projects, not that it's a C# file.

Really? I wouldn't have guessed that. So if I reference that package from a VB project the file won't be present? I find that really curious because there is no evidence [in the nuspec file](https://github.com/NuGet/Samples/blob/master/ContentFilesExample/authoring/ContentFilesExample.nuspec) that that is the case. Perhaps this is only considered by Nuget in specific cases? But encoding that into the browse object might be tricky.

I wonder if it would almost be better for Nuget to provide this functionality via its VS tooling (obv. once the API is stabilized) as they're the experts.

drewnoakes:
> So if I reference that package from a VB project the file won't be present?

Yep. A `.cs` file wouldn't fare well in a VB project.

The nuspec specifies which files to bring in, but the pack process applies some logic to those files. For that `.nupkg`, the structure of the `contentFiles` directory is:

image

> I wonder if it would almost be better for Nuget to provide this functionality via its VS tooling (obv. once the API is stabilized) as they're the experts.

I'll reach out to the NuGet folks. We have a discussion going about other features at the moment.

davidwengier:
Ahh, I missed that the text file in the screenshot was in the `\cs\` folder. Makes sense now :)

drewnoakes:
Questions for NuGet folks:

1. Do content file paths always begin with `contentFiles` and if so could we remove that prefix from the caption?

2. Is there a reason that, in a C# project, the assets file lists content files in both `cs` and `any` language codes? In the above screenshot, the files in `any` are not present in the project as far as I can tell.

nkolev92:
> Do content file paths always begin with contentFiles and if so could we remove that prefix from the caption?

Yes you can remove that. 

> Is there a reason that, in a C# project, the assets file lists content files in both cs and any language codes? In the above screenshot, the files in any are not present in the project as far as I can tell.

The trick around this is the language. 
The assets file doesn't express the language, so it's taken care of in ResolvePackageAssets in legacy projects, https://github.com/dotnet/NuGet.BuildTasks/blob/5244c490a425353ac12445567d87d674ae118836/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs#L572-L575, but NuGet writes the condition to the props file for SDK based projects.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AssetsFileAttachedCollectionSourceBase.cs
```diff
@@ -142,6 +144,12 @@ protected static void ProcessLibraryContent(ref List<object>? items, AssetsFileT
                 items ??= new List<object>();
                 items.Add(PackageAssemblyGroupItem.CreateWithContainsItems(snapshot, library, PackageAssemblyGroupType.Framework, library.FrameworkAssemblies));
             }
+
+            if (!library.ContentFiles.IsEmpty)
+            {
+                items ??= new List<object>();
+                items.Add(PackageContentFilesGroupItem.CreateWithContainsItems(_fileIconProvider, library.ContentFiles));
```
nkolev92, line 151: Do a language check similar to https://github.com/dotnet/NuGet.BuildTasks/blob/5244c490a425353ac12445567d87d674ae118836/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs#L572-L575

build/import/Packages.targets
```diff
@@ -124,10 +124,10 @@
     <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="3.0.0-beta2.20059.3+77df2220" />
 
     <!-- NuGet -->
-    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6532" />
-    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6532" />
-    <PackageReference Update="NuGet.ProjectModel"                                                     Version="5.6.0-preview.2.6532" />
-    <PackageReference Update="NuGet.Common"                                                           Version="5.6.0-preview.2.6532" />
+    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.6.0-preview.2.6554" />
+    <PackageReference Update="NuGet.VisualStudio"                                                     Version="5.6.0-preview.2.6554" />
+    <PackageReference Update="NuGet.ProjectModel"                                                     Version="5.6.0-preview.2.6554" />
```
nkolev92, line 129: We've avoid having other VS bound components depending on non-interop assemblies because of binding redirects. 
As you mentioned in a thread, I'd expect this to fail often given that there's no mechanism to keep this up to date at this point. 

The "instinctive" solution here is binding redirects, but that'd break many extensions and build tasks that might be depending on NuGet assemblies.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AssetsFileAttachedCollectionSourceBase.cs
drewnoakes: Thanks for confirming the logic. I've added this to the backlog as it's not immediately clear how I plumb the project language through to where it's needed without adding a lot of mess. I'm working on a slightly different design that should make this kind of thing easier (the file icon provider is another example from this PR).

build/import/Packages.targets
drewnoakes: Followed up on this via email.

drewnoakes:
> Yes you can remove that.

Thanks, done.

PR #6067, opened 2020-04-06
Update project selector association

tmeschter:
Update the project selectors associated with C# and VB projects to use
the one added in https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/VS/pullrequest/239926.

Note this cannot be merged until that PR has been merged and made its
way into internal builds, and we cannot remove the old project selector
until this makes its way into internal builds.


davkean:
Wonder if we should just unify F# too at the same time.

tmeschter:
@davkean Their logic is somewhat different from ours, though in general I would support such a move. Let's get this in, check with the F# team, and if they're in favor I'll do that as a separate change.

tmeschter:
I've verified that this change, on top of a recent build from VS `master`, causes us to use the project selector from managed code rather than the one from native code.

I'll merge this in once the tests pass.

PR #6072, opened 2020-04-08
Update analyzers to latest version

davkean:


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
```diff
@@ -238,7 +238,7 @@
   <ItemGroup>
     <None Include="**\*.xlf" />
     <EmbeddedResource Update="ProjectSystem\VS\PropertyPages\PropertyPageResources.resx">
-      <Generator>PublicResXFileCodeGenerator</Generator>
```
davkean, line 241: The API analyzers was complaining about the lack of `~` for these members in PublicAPI.Shipped.txt, but this should class should just not be public.

Issue #6053, opened 2020-04-01
[VS Online] Project factories for unsupported project types should not be registered


davkean:
To avoid loading project types that are not supported, we should make sure that they are not registered so that the project factory load is failed. This was an agreement between Solution and us on how to "light down" project types, for now, we should remove the factory registration for:

C# legacy
VB legacy

When we do this, we _must_ make sure that the IVsProjectSelectors (https://github.com/dotnet/project-system/issues/6052) continue to run. If they do not, then we should come to an agreement on an "opt-in" per project type to get loaded in the solution.

davkean:
C# appears to be registered in: coolpkg.rgs_5483.ddr
VB appears to be registered in /src/SetupPackages/CoreIde/CommunityExtra/core/PkgDef/VB_Project_Reg_VS_26699.3643236F_FC70_11D3_A536_0090278A1BB8.pkgdef or /src/SetupPackages/CoreIde/CommunityVB/core/Neutral/PkgDef/VB_Project_Reg_VS_26699.3643236F_FC70_11D3_A536_0090278A1BB8.pkgdef.

davidwengier:
I'm not quite sure on exactly what "they" is in the description above. Do we just want to avoid registering the ProjectSelector? The legacy package? or the whole `Projects\{FAE04EC0...` key and subkeys?

davidwengier:
I'm guessing the C# registration could equally come here: query=VCSharp_Proj_System_&rightProject=VS(Repo)&file=src%5Cvsproject%5Ccool%5Ccoolpkg%5Cpkg%5CVCSharp_Proj_System_Reg.pkgdef&line=83 ?

Or is there something about `.ddr` files that made you point there?

davkean:
Yeah you are right, that's probably it. DDR was from the old MSI-based setup.

davidwengier:
Internal PR: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/242132?_a=overview

davidwengier:
Internal PR: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/242132?_a=overview

davidwengier:
Internal PR has been completed

PR #6088, opened 2020-04-14
Add some missing values from our project type registration

davidwengier:
We were missing some values that the legacy project system registers. We're still missing some after this, but `PossibleProjectExtensions` unblocks #6053


drewnoakes:
I assume setting `NewProjectRequireNewFolderVsTemplate` disallows instantiating a project templates in an existing folder. Why block that?

davidwengier:
> I assume setting NewProjectRequireNewFolderVsTemplate disallows instantiating a project templates in an existing folder

Yes, this maps to VSDIRFLAG_RequiresNewFolder in [__VSDIRFLAGS2](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.__vsdirflags2?view=visualstudiosdk-2019)

> Why block that?

I don't really have an opinion on that, but I don't want to change behaviour as part of this, so matching legacy is the goal for now.

Issue #6090, opened 2020-04-15
"Collection was modified" in Dependencies node


drewnoakes:
Reported in https://developercommunity.visualstudio.com/content/problem/990379/project-system-data-flow-crosstarget-joint-input-4.html and https://twitter.com/jonsagara/status/1250168538737692672

> I upgraded to VS 2019 16.5.4 this morning (2020-04-14, Pacific Time), and I have started to see this exception in the yellow error bar at the top of the screen when I open my solution.

<details>
<summary>Stack trace</summary>

```
=====================
4/14/2020 2:02:53 PM
Recoverable
System.AggregateException: Project system data flow 'CrossTarget Joint Input: 43073702' closed because of an exception: System.AggregateException: One or more errors occurred. ---> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Immutable.ImmutableDictionary`2.Enumerator.ThrowIfChanged()
   at System.Collections.Immutable.ImmutableDictionary`2.Enumerator.MoveNext()
   at System.Collections.Immutable.ImmutableDictionary`2.Builder.<get_Keys>d__18.MoveNext()
   at System.Linq.Enumerable.<ExceptIterator>d__73`1.MoveNext()
   at Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.DependenciesSnapshot.<FromChanges>g__SyncTargetFrameworks|1_0(<>c__DisplayClass1_0& )
   at Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.DependenciesSnapshot.FromChanges(String projectPath, DependenciesSnapshot previousSnapshot, ITargetFramework changedTargetFramework, IDependenciesChanges changes, IProjectCatalogSnapshot catalogs, ImmutableArray`1 targetFrameworks, ITargetFramework activeTargetFramework, ImmutableArray`1 snapshotFilters, IReadOnlyDictionary`2 subTreeProviderByProviderType, IImmutableSet`1 projectItemSpecs)
   at Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependenciesSnapshotProvider.<>c__DisplayClass34_0.<UpdateDependenciesSnapshot>b__0(DependenciesSnapshot previousSnapshot)
[...]

PR #6094, opened 2020-04-15
[16.6] Don't modify collection while enumerating

drewnoakes:
⚠ This PR is the 16.6.x equivalent of #6091 and should not be merged without QB approval.

---

This bug has existed in the dependencies node for some time. It started surfacing for users in 16.5.4 when faulting dataflow blocks were reported (PR #6073). The bug is tracked in #6090.

The result of `Except` is lazily computed, so attempting to remove from the builder while enumerating that result causes a "Collection was modified" exception. The fix is to materialise the result prior to enumerating the data that backs it.


drewnoakes:
Ask Mode on internal bug: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1105433

PR #6095, opened 2020-04-15
Change warning message to fit and be consistent.

### LINKED ISSUE #4904: PackageLicenseUrl warning message gets clipped in project's package properties
project properties license

image

melytc:
Fixes #4904. 
The warning message was long and couldn't fit in the row space. Changed the text to be consistent to the other warning message.

Here's what it would look like if PackageLicenseUrl and PackageIconUrl were found in the project file:
image

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
@@ -720,6 +720,9 @@
   <data name="ExpressionLabel.Anchor" type="System.Windows.Forms.AnchorStyles, System.Windows.Forms">
     <value>Right</value>
   </data>
+  <data name="ExpressionLabel.AutoSize" type="System.Boolean, mscorlib">
+    <value>True</value>
+  </data>
   <data name="ExpressionLabel.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
     <value>NoControl</value>
   </data>
--- a/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
@@ -700,7 +700,7 @@
     <value>10</value>
   </data>
   <data name="LicenseUrlWarning.Text" xml:space="preserve">
-    <value>PackageLicenseURL property was detected, but it has been deprecated. Please change to either an Expression or File.</value>
+    <value>PackageLicenseURL detected but is deprecated. Please change to either an Expression or File.</value>
   </data>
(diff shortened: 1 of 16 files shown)


davidwengier:
How do we know that one of the translations won't end up being too long to fit in the space? Is there a way to make the text wrap?

melytc:
> How do we know that one of the translations won't end up being too long to fit in the space? Is there a way to make the text wrap?

Good point. The textbox has already `Multiline` and `Wordwrap` set to true, though. Let me see if there's another property to do that.

src/Microsoft.VisualStudio.Editors/PropPages/PackagePropPage.resx
```diff
@@ -700,7 +700,7 @@
     <value>10</value>
   </data>
   <data name="LicenseUrlWarning.Text" xml:space="preserve">
-    <value>PackageLicenseURL property was detected, but it has been deprecated. Please change to either an Expression or File.</value>
+    <value>PackageLicenseURL detected but is deprecated. Please change to either an Expression or File.</value>
```
drewnoakes, line 703: This may be more consistent with the other string:

> PackageLicenseURL detected but is deprecated. Use Expression or File instead.

Or update the other message to be more polite :)

melytc: Changed both to make them consistent and polite 👍

melytc:
After battling with the Designer, I haven't figured out how to expand the visible area to the number of lines the text needs for both the `PackageLicenseURL` and `PackageIconUrl` warning messages.
On the bright side, I fixed another minor bug: the `Expression:` label wasn't completely visible.

Here's what it looks like:
image

melytc:
Notes:
- `AutoSize` only changes the width, not the height of the element.
- Some Winforms scaling APIs have been deprecated. [Docs](https://docs.microsoft.com/en-us/dotnet/framework/winforms/automatic-scaling-in-windows-forms?view=netcore-3.1#current-support-for-automatic-scaling).

An approach that we can take for now is to give those warnings a 2-line size, just in case translations occupy more than 1 line.

If warnings have 1 line:
image

If warnings have 2 lines: 
image

Thoughts? @davidwengier @drewnoakes

davidwengier:
I once wrote a `WordWrappingLabel` control for a WinForms app specifically because the `AutoSize` property on the built in one only changed width. It's been a while but I would happily try to write one again, or pair with you in doing so.

davkean:
Doesn't a text block do that?

davidwengier:
image

davkean:
Oh this was a new page so assumed it was WPF.

melytc:
Thanks @davidwengier for pair programming this with me!

PR #6106, opened 2020-04-20
LOC CHECKIN | dotnet/project-system dev16.6.x | 20200420

v-chmart:
Localization update

PR #6107, opened 2020-04-20
Fixed a typo in Resources.resx

melytc:


src/Microsoft.VisualStudio.ProjectSystem.Managed/Resources.resx
```diff
@@ -181,9 +181,9 @@ If this file or folder does not appear in Solution Explorer, then it is not curr
   <data name="ProjectLoadedWithWrongProjectType" xml:space="preserve">
     <value>The project file '{0}' cannot be opened. 
 
-This project was loaded using the wrong project type, likely as a result of renaming the project extension outside of Visual Studio. Remove and readd the project to have the solution pick up the right project type.</value>
+This project was loaded using the wrong project type, likely as a result of renaming the project extension outside of Visual Studio. Remove and read the project to have the solution pick up the right project type.</value>
```
davkean, line 184: This should be "re-add", not "read".

davidwengier, line 184: This is talking about removing and then adding back the project, not reading the project :P

```suggestion
This project was loaded using the wrong project type, likely as a result of renaming the project extension outside of Visual Studio. Remove and re-add the project to have the solution pick up the right project type.</value>
```

melytc: That makes sense, thanks for clarifying it.

melytc:
It got localized as `read`, at least in the Spanish and French ones. #6106

davkean:
You'll need to regenerate the strings by building at the command-line.

PR #6117, opened 2020-04-22
Replace usages of "flavor" with "favor"

terrajobst:
Presumably the project system concept of "flavoring" has made its way in @davkean's subconsciousness 😄


davkean:
lol

Issue #5904, opened 2020-02-19
PLM: Intellisense stage may get stuck after switching configuration in PackageRestore


adrianvmsft:
**Visual Studio Version**: 16.5

**Summary**:
No repro, found by investigating a dump where intellisense stage was stuck.
This happened because some data source versions regressed, which triggered exceptions. These exceptions were not reported (filed #5903) and are causing blocks to be faulted.
The versions associated with the exception indicate a configuration change. Additional filtering should address this problem by using the following CPS method (currently internal):
ProjectDataSources.CreateDataSourceVersionForwardOnlyFilteringBlock

https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs

before creating mergeBlock

```
Exception object: 8b9e5434
Exception type:   Microsoft.VisualStudio.ProjectSystem.ProjectDataSources+DataSourceRegressionException
Message:          Some data sources ($AdditionalRuleDefinitionsVersion) have regressed in their version number for dataflow '$SyncLink (Microsoft.VisualStudio.ProjectSystem.ProjectDataSources+SourceBlockAndLink`1[Microsoft.VisualStudio.ProjectSystem.IProjectValueVersions][])'.
InnerException:   <none>
StackTrace (generated):
    SP       IP       Function
    4178E97C 5D6D2C65 Microsoft_VisualStudio_ProjectSystem_ni!Microsoft.VisualStudio.ProjectSystem.ProjectDataSources.CheckValidVersionProgression(System.Func`1<System.String>, System.Collections.Immutable.IImmutableDictionary`2<Microsoft.VisualStudio.ProjectSystem.NamedIdentity,System.IComparable>, System.Collections.Immutable.IImmutableDictionary`2<Microsoft.VisualStudio.ProjectSystem.NamedIdentity,System.IComparable>, Boolean)+0x25c425
    4178E9D4 5D47516E Microsoft_VisualStudio_ProjectSystem_ni!Microsoft.VisualStudio.ProjectSystem.SyncLinkToLink`1[[System.__Canon, mscorlib]].ProcessData(Boolean)+0x37e

0:155> !do 8b9e5434
Name:        Microsoft.VisualStudio.ProjectSystem.ProjectDataSources+DataSourceRegressionException
MethodTable: 5d66cf34
EEClass:     5d398b24
[...]

PR #6118, opened 2020-04-22
[16.6] Prevent restore versions going back in time

davkean:
Fixes: https://github.com/dotnet/project-system/issues/5904 for 16.6.

We are merging data from configurations into a single value. This can cause data source versions to go back in time when switching from Debug -> Release -> Make some changes -> Debug.

We want to prevent that, so introduce a forward only block that will only publish new versions of a value when it has the same or higher version than the previous value. The expectation is that another value with a more recent data source version will appear shortly.


jjmew:
@davkean Do we really need this for 16.6?

davkean:
This represents 8% of yellow bars in 16.6 pre3 and 16.5.

Issue #3101, opened 2018-01-04
No thread-safe way to iterate VisualBasicNamespaceImportsList


davkean:
There appears to be no thread-safe way to iterate VisualBasicNamespaceImportsList:

https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.VisualBasic.VS/ProjectSystem/VS/Automation/VisualBasicNamespaceImportsList.cs.

While we lock when you access the indexer/count, etc, that lock is not held in the following situations:

``` C#

foreach (string import in vsProject.Imports)    // No lock taken at all
{

}
```

-or-

``` C#
for (int i = 0; i < vsProject.Imports.Count; i++)   // Lock is only held while you access `Count`
{
   var import = vsProject.Imports[i]; // Lock is only held while you read the indexer
}
```

At any time while iterating those a background thread could update the imports.

Pilchie:
Seems like this could be fixed by changing so that:

1. Changes to the list are serialized through the UI thread.
2. Mutating operations like `VisualBasicVSImports.Remove` wait for the dataflow changes to be processed before returning.

I guess that's problematic if there was an unrelated change that also caused a dataflow change?

Pilchie:
Presumably this issue applies to many automation objects, not just this one too?

davkean:
No, we wrote this one - CPS wrote the others. I believe they are thread-safe.

Pilchie:
Are there pointers to the CPS ones?  How do they implement their thread safety?

davkean:
OAVSReferences and ReferencesHostBridge are examples.

davkean:
> Changes to the list are serialized through the UI thread.
That's exactly what it does via the `ProjectHostBridge<TInput, TOutput, TApplied>` base class.

Pilchie:
That still suffers from the problem that mutations that happen through that interface on the UI thread are not immediately visible, doesn't it?

davkean:
That's what this line does:

``` C#
await this.ProjectNode.Value.ReferencesHostBridge.ReceiveLatestSnapshotAsync();
```

See OAVSReferences.AddProject.

Pilchie:
Got it.  Seems like that's the pattern to follow here then.

davkean:
Yeah exactly, either we copy the base class or just get CPS to make it public.

Pilchie:
[...]

PR #6119, opened 2020-04-23
Update VisualBasicImports to use ProjectHostBridge and fire events

davidwengier:
Fixes https://github.com/dotnet/project-system/issues/3101

This updates VisualBasicNamespaceImports to use ProjectHostBridge to guarantee consistency of the collection when on the UI thread. It also moves DTE events to be triggered after the list of imports is actually updated which fixes a UI bug in the app designer, where clicking an import adds it to the project file but doesn't make the list item appear checked.

~This _doesn't_ fix the (possible?) bug that after calling `VSProject.Imports.Add` the new import is not visible in the collection upon that method returning, however since that is already a bug we have now, and the ideal solution for that involves a change to CPS's public API, and nobody answered me when I asked about it on Teams (😛), it's not worth blocking this change on.~ Changed my mind! Trying to make the API public in CPS was ugly.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicNamespaceImportsList.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicNamespaceImportsList.cs
@@ -82,11 +82,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation.VisualBasic
             return Task.CompletedTask;
         }
 
-        protected override Task ApplyAsync(IProjectVersionedValue<ImmutableList<string>> value)
+        protected override async Task ApplyAsync(IProjectVersionedValue<ImmutableList<string>> value)
         {
+            await JoinableFactory.SwitchToMainThreadAsync();
+
             ImmutableList<string> current = AppliedValue?.Value ?? ImmutableList<string>.Empty;
             ImmutableList<string> input = value.Value;
-            
+
             IEnumerable<string> removed = current.Except(input);
             IEnumerable<string> added = input.Except(current);
 
@@ -106,8 +108,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation.VisualBasic
                     imports.OnImportAdded(import);
                 }
             }
-
-            return Task.CompletedTask;
         }
 
         protected override Task<IProjectVersionedValue<ImmutableList<string>>> PreprocessAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> input, IProjectVersionedValue<ImmutableList<string>>? previousOutput)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicNamespaceImportsList.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicNamespaceImportsList.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using Microsoft.VisualStudio.ProjectSystem.Build;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicNamespaceImportsList.cs
```diff
@@ -70,53 +77,61 @@ public IEnumerator<string> GetEnumerator()
...
         internal Task OnProjectFactoryCompletedAsync()
         {
-            EnsureInitialized();
+            TryInitialize();
+
             return Task.CompletedTask;
         }
 
-        internal void OnNamespaceImportChanged(IProjectVersionedValue<IProjectSubscriptionUpdate> e)
+        protected override Task ApplyAsync(IProjectVersionedValue<ImmutableList<string>> value)
```
davkean, line 85: Verify that we're on the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicVSImports.cs
```diff
@@ -68,8 +69,6 @@ public void Add(string bstrImport)
                         project.AddItem(ImportItemTypeName, bstrImport);
```
davkean, line 69: I think here you want to wait for latest snapshot, see ReferencesHostBridge.ReceiveLatestSnapshotAsync as an example.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicVSImports.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation.VisualBasic
 {
     [Export(typeof(Imports))]
     [Export(typeof(ImportsEvents))]
+    [Export(typeof(VisualBasicVSImports))]
     [AppliesTo(ProjectCapability.VisualBasic)]
     [Order(Order.Default)]
     internal class VisualBasicVSImports : ConnectionPointContainer,
```
davkean, line 18: I think we should have a similar relationship between this and VisualBasicNamespaceImportsList like we do with OSVSReferences and ReferenceHostBridge. It doesn't feel like VisualBasicNamespaceImportsList should know anything about firing VS events, and VisualBasicVSImports waits for results from the HostBridge.

davkean: Or maybe jsut combine the two like VsPersistConnectedServices? Their relationship is a circular reference.

davkean: Oh my bad, that's exactly what ReferenceHostBridge does. Ugh - circular dependencies are the pits.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/VisualBasic/VisualBasicVSImports.cs
davidwengier: As per the description, I'm going to leave that till later since this behaviour is the same as the current implementation, and I think its going to be better to move ReceiveLatestSnapshotAsync down to ProjectHostBridge, rather than copying it into here.

davkean: Can we file a bug for this so that we don't lose track of it?

davidwengier: Tried this on Friday and there wasn't a nice way to do make it public, so just going to do this in here after all :)

davidwengier:
Merging this without approvals. Let me know if there is any follow up required.

Issue #6124, opened 2020-04-23
NFW: Microsoft.VisualStudio.ProjectSystem.VS.TempPE.DesignTimeInputsChangeTracker.ProcessDataflowChanges(IProjectVersionedValue`1)


davkean:
**Visual Studio Version**:

16.5

```
The given key was not present in the dictionary.
["System.Collections.Immutable.ImmutableDictionary`2.get_Item(TKey)
>>>Microsoft.VisualStudio.ProjectSystem.VS.TempPE.DesignTimeInputsChangeTracker.ProcessDataflowChanges(IProjectVersionedValue`1)
>>>Microsoft.VisualStudio.ProjectSystem.ActionBlockSlimSync`1.ProcessInputAsync(TInput)
>>>Microsoft.VisualStudio.ProjectSystem.DataReceivingBlockSlim`1.<ProcessInputQueueAsync>d__5.MoveNext()
>>>System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>>>Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.Rethrow(Exception)
>>>Microsoft.VisualStudio.ProjectSystem.ProjectErrorReporting.<>c__DisplayClass6_0.<SubmitErrorReport>b__0()
>>>Microsoft.VisualStudio.ProjectSystem.ExceptionFilter.<>c__DisplayClass2_0.<Guard>g__action|0()
>>>GuardMethodClass.GuardMethod(Func`1,Func`2,Func`2)"]
```

**Expected Behavior**:

**Actual Behavior**:

**User Impact**:

PR #6125, opened 2020-04-23
Protect against values missing from dictionaries in TempPE compilation

davidwengier:
Fixes https://github.com/dotnet/project-system/issues/6124

How common is this NFW @davkean ? Do we need this in 16.6?

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
@@ -152,6 +152,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
             if (!input.Value.Item2.ProjectChanges.TryGetValue(ConfigurationGeneral.SchemaName, out IProjectChangeDescription configChanges))
             {
                 // If this isn't an update we can deal with, just ignore it
+                _currentState = null;
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


davkean:
This doesn't need to be in 16.5.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -148,7 +148,12 @@ internal void ProcessFileChangeNotification(IProjectVersionedValue<string[]> arg
         internal void ProcessDataflowChanges(IProjectVersionedValue<ValueTuple<DesignTimeInputs, IProjectSubscriptionUpdate>> input)
         {
             DesignTimeInputs inputs = input.Value.Item1;
-            IProjectChangeDescription configChanges = input.Value.Item2.ProjectChanges[ConfigurationGeneral.SchemaName];
+
+            if (!input.Value.Item2.ProjectChanges.TryGetValue(ConfigurationGeneral.SchemaName, out IProjectChangeDescription configChanges))
```
davkean, line 152: This could be the project going from a correct state -> bad state, in which case your state has changed and you should update your outputs. Otherwise, when the project is fixed the next "diff" is going to be out of sync with what you think the current "state" is.

Use the GetSnapshotOrEmpty to get an empty snapshot.

davidwengier:
> This doesn't need to be in 16.5.

Sure, but what about 16.6? This is currently targeting master.

davkean:
Sorry, the hits are pretty low - master is good.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
davidwengier: This block works on ProjectChanges, not a snapshot, so GetSnapshotOrEmpty doesn't apply, but you're right this should have been clearing the state.

davidwengier:
Merging this without approvals. Let me know if there is any follow up required.

PR #6131, opened 2020-04-27
Add extension API for transitive dependencies

drewnoakes:
Introduces the new public API for populating transitive references in the Dependencies tree and reimplements previous code in terms of it.

The new API surface area is visible in `PublicAPI.Unshipped.txt`.

New concepts are described in `dependencies-node-roadmap.md` and relevant public API docs. I highly recommend starting a review by reading this file.

The provisioning of package and project reference transitive dependencies will be moved to NuGet. For now it has been consolidated within the folder `src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\Tree\Dependencies\AttachedCollections\NuGet\`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6131)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateContainsRelationCollection.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        bool IList.Contains(object value) => value is IRelatableItem item && _spans.Any(span => span.Items?.Contains(item) == true);
+
+        void ICollection.CopyTo(Array array, int index) => throw new InvalidOperationException();
+        object ICollection.SyncRoot => throw new InvalidOperationException();
+        bool ICollection.IsSynchronized => false;
+        int IList.Add(object value) => throw new InvalidOperationException();
+        void IList.Clear() => throw new InvalidOperationException();
+        void IList.Insert(int index, object value) => throw new InvalidOperationException();
```
davkean, line 184: NotSupportedException.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateContainsRelationCollection.cs
```diff
@@ -0,0 +1,190 @@
...
+                    baseIndex += spanCount;
+                }
+            }
+
+            return -1;
+        }
+
+        bool IList.Contains(object value) => value is IRelatableItem item && _spans.Any(span => span.Items?.Contains(item) == true);
+
+        void ICollection.CopyTo(Array array, int index) => throw new InvalidOperationException();
```
davkean, line 179: Why?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/BrowseObjectDescriptionAttributeBase.cs
```diff
@@ -3,18 +3,19 @@
...
+using System.Resources;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections
 {
     /// <summary>
     /// Specifies a localized description for a property or event.
     /// </summary>
     [AttributeUsage(AttributeTargets.Property)]
-    internal sealed class BrowseObjectDescriptionAttribute : DescriptionAttribute
+    public abstract class BrowseObjectDescriptionAttributeBase : DescriptionAttribute
```
davkean, line 14: I would not make this public.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/IDependenciesTreeProjectSearchContext.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Gets the unconfigured project being searched.
+        /// </summary>
+        UnconfiguredProject UnconfiguredProject { get; }
+
+        /// <summary>
+        /// The set of target frameworks exposed by this project.
+        /// </summary>
+        ImmutableArray<string> TargetFrameworks { get; }
```
davkean, line 26: We should avoid exposing TargetFrameworks, and instead be using CPS concepts; Configurations, Dimensions Names and Values, ConfigurationGroups, etc.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/IDependenciesTreeProjectSearchContext.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Gets a sub-context, specific to a given target framework of the project.
+        /// </summary>
+        /// <param name="targetFramework">The target framework being searched.</param>
+        /// <returns>The sub-context for the target framework, or <see langword="null"/> if not found.</returns>
+        IDependenciesTreeProjectTargetSearchContext? ForTarget(string targetFramework);
```
davkean, line 33: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/BrowseObjectBase.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedColl
     /// right from it)
     /// </summary>
     [ComVisible(true)]
-    internal abstract class BrowseObjectBase : ICustomTypeDescriptor
+    public abstract class BrowseObjectBase : ICustomTypeDescriptor
```
davkean, line 19: Does this actually need to be public and COMVisible? This is typically only needed for IDispatch to work, but ICustomTypeDescriptor is skipping CLR's IDispatch.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/DependenciesTreeSearchProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+                IProjectTree? dependenciesNode = projectVsServices?.ProjectTree.CurrentTree?.FindChildWithFlags(DependencyTreeFlags.DependenciesRootNode);
+
+                if (projectVsServices == null || dependenciesNode == null)
+                {
+                    return Task.CompletedTask;
+                }
+
+                var targets = new List<string>();
+
+                foreach (ConfiguredProject configuredProject in projectVsServices.Project.LoadedConfiguredProjects)
```
davkean, line 109: Why loaded projects? This property will return all projects that have ever been loaded. I think you want "implicitly active configs" via IActiveConfigurationGroupService2.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/DependenciesTreeSearchProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+                {
+                    if (configuredProject.ProjectConfiguration.Dimensions.TryGetValue(ConfigurationGeneral.TargetFrameworkProperty, out string target))
```
davkean, line 111: Should be based on `IConfigurationGroup<T>`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/AssetsFileTopLevelDependenciesCollectionSourceProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+        protected abstract bool TryUpdateItem(TItem item, AssetsFileTarget targetData, AssetsFileTargetLibrary library);
+
+        protected override bool TryCreateCollectionSource(
+            IVsHierarchyItem hierarchyItem,
+            string flagsString,
+            string? target,
+            IRelationProvider relationProvider,
+            [NotNullWhen(returnValue: true)] out AggregateRelationCollectionSource? containsCollectionSource)
+        {
+            if (!TryGetIdentity(flagsString, out TIdentity identity))
```
davkean, line 47: Can we add an assert for UI thread?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/AssetsFileTopLevelDependenciesCollectionSourceProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+                                item = CreateItem(targetData, library);
+                                if (AggregateContainsRelationCollection.TryCreate(item, relationProvider, out collection))
+                                {
+                                    await _joinableTaskContext.Factory.SwitchToMainThreadAsync();
+                                    collectionSource.SetCollection(collection);
+                                }
+                            }
+                            else if (TryUpdateItem(item, targetData, library) && collection != null)
+                            {
+                                await _joinableTaskContext.Factory.SwitchToMainThreadAsync();
```
davkean, line 92: I think you want to use IProjectThreadingService from IProjectService here. This is effectively part of the project, and you want to make sure you end up going through CPS's JTF collection.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/Relations/PackageToDiagnosticRelation.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        protected override bool HasContainedItems(PackageReferenceItem parent)
+        {
+            return parent.Target.Logs.Any(log => log.LibraryId == parent.Library.Name);
+        }
+
+        protected override void UpdateContainsCollection(PackageReferenceItem parent, AggregateContainsRelationCollectionSpan span)
+        {
+            span.UpdateContainsItems(
+                parent.Target.Logs.Where(log => log.LibraryId == parent.Library.Name).OrderBy(log => log.LibraryId).ThenBy(log => log.Message),
```
davkean, line 22: This is ordered by Culture by default, is that deliberate?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/Models/AssetsFileLogMessage.cs
```diff
@@ -20,7 +20,7 @@ public AssetsFileLogMessage(IAssetsLogMessage logMessage)
         }
 
         public NuGetLogCode Code { get; }
-        public NuGet.Common.LogLevel Level { get; }
+        public LogLevel Level { get; }
         public WarningLevel WarningLevel { get; }
         public string Message { get; }
         public string LibraryId { get; }
```
davkean, line 26: I would consider calling this LibraryName to make the original data.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/Models/AssetsFileTarget.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        /// <summary>
+        /// Gets the set of dependents (parents) of <paramref name="libraryName"/>.
+        /// </summary>
+        /// <returns><see langword="true"/> if dependents were found, otherwise <see langword="false"/>.</returns>
+        public bool TryGetDependents(string libraryName, out ImmutableArray<AssetsFileTargetLibrary> dependents)
+        {
+            // Defer construction of dependents collection until needed. It's only needed for Solution Explorer search.
+            if (_dependentsByLibrary == null)
```
davkean, line 67: Can this overlap?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateContainsRelationCollection.cs
drewnoakes: Because it's not needed by `ListCollectionView` and implementation is non-trivial.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/BrowseObjectDescriptionAttributeBase.cs
drewnoakes: There are three of these base classes which are provided to make implementation easier for users of this system:

- `BrowseObjectBase`
- `BrowseObjectDescriptionAttributeBase`
- `BrowseObjectDisplayNameAttributeBase`

You could argue that `AttachedCollectionItemBase` falls into this category, or even `RelatableItemBase` and `RelationBase<TParent, TChild>`.

Collectively they are a decent amount of code for folks to copy/paste into their projects. Providing them makes it easier to implement the required protocols correctly.

If your concern is that we may wish to change these implementations in future, then one of two things would happen:

- We modify the implementation &mdash; these are abstract base classes which are easier to modify than interfaces.
- We want to use an alternative implementation &mdash; we create new implementations and leave these types in the product.

This PR has 92 new public API members. The attributes add 8. `BrowseObjectBase` adds 5. I would leave them in.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/Relations/PackageToDiagnosticRelation.cs
drewnoakes: So long as it's stable, it's fine. There order here is just used for merging data structures. The UI applies its own sorting later.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/BrowseObjectBase.cs
drewnoakes: `ComVisible` is not needed. Removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/NuGet/Models/AssetsFileTarget.cs
drewnoakes: Only if operations performed within `Enumerable.ToDictionary` are reordered after the field write, which seems unlikely given the stack pop. Using `Volatile.Write` to the field just in case.

drewnoakes:
@davkean I've updated the search feature to use `ConfiguredProject` rather than TFMs. Please take another look, especially at 9e0e6d7b7cd0cf65aa501dec2cda4ebc1b832dc6.

Issue #6080, opened 2020-04-10
Trigger In-Product Acquisition for optional SDK workloads


dsplaisted:
When opening a project that requires an optional SDK workload (such as iOS or Android) that isn't installed, the project system should trigger an in-product acquisition experience to allow the user to click through to install the required workload.  The design for this (from the MSBuild / SDK side) is here: https://github.com/dotnet/designs/pull/104

dsplaisted:
FYI @jjmew, I've filed this issue to track the in-product acquisition experience we've discussed a few times.

jjmew:
Triage: @davkean do you have a recommendation who should work on this?

jjmew:
@cartermp We need your help designing this. This is high priority. Please work with @davkean

cartermp:
Yep, happy to do it. I'm not quite clear on if we get access to evaluation results at "in product acquisition time", since my understanding is that the VS system triggers this when it _attempts_ to load a project. Can IPA also be triggered after evaluation? @davkean or @jmarolf would know more.

@dsplaisted The question in your spec is a good one:

> QUESTION: Will it be OK to put this mapping of .NET workload IDs to VS workload/component IDs in the workload manifest packages?  Or do we need to supply it some other way

I'd defer to @jmarolf since he authored most of the .NET manifests and probably has the best idea for where that mapping should live. What VS tells the installer to acquire ultimately depends on how things are factored and what level of granularity (VS Workload, VS Component Group, VS Component) we're expecting with .NET SDK workloads and what state someone's VS might be in w.r.t what they have installed. @dsplaisted are you the owner of that?

Side Note: the terminology here is going to get confusing for everyone involved 😿

davkean:
@cartermp Let's sync up, I have a good idea how to approach based on existing infrastructure we already have in the product.

davidwengier:
Plan:

* Solution-level IProjectRetargetingManager service that waits to be told about problems
* Project-level (configured? unconfigured?) service that hooks up to evaluation on project load
* When evaluation data comes in, above service calls to its providers, and each one checks for a specific problem
* If problems are found, they are reported to the IProjectRetargetingManager
* Once the solution is loaded (???) the IProjectRetargetingManager calls `IVsTrackProjectRetargeting2.CheckSolutionForRetarget` to kick off the retargeting
* Retargeting is handled by a `IProjectRetargetingProvider` that reports the problems found above

davidwengier:
[...]

PR #6134, opened 2020-04-27
In-product acquisition proposal

cartermp:
https://github.com/dotnet/project-system/issues/6080

This still needs a lot of work, mostly in getting some mockups in there. But the set of applicable scenarios should mostly be listed.

Please consider this more as a proposal than a spec. Almost everything here is _proposed_ - I'm not tied to a particular scenario or way to do things (e.g., preferring to suggestion workloads over components if possible). It could also be that some of what I'm proposing here isn't feasible. We'll need to suss that out over time as the work actually happens.

The main thing here should be ensuring that we have adequate acquisition scenario coverage. If the rest of the doc should just get chopped to just that I'm happy to do that too.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6134)


davkean:
tag @davidwengier

davidwengier:
Just to confirm the scope of this, not all of the project types mentioned use the .NET Project System. eg, pre-net5 Xamarin and UWP are both legacy projects, and C++/CLI is the C++ project system.

cartermp:
Re: Xamarin - not necessarily. You can create .NET Standard projects that reference Xamarin.Forms and act as core UI libraries today: https://github.com/xamarin/Essentials/blob/develop/Samples/Samples/Samples.csproj - while it's unlikely that someone has created/opened one of these things _without_ the head projects that use legacy projects (which then kick off current IPA), it's still possible. I can clarify that.

Re: C++/CLI - since they use CPS it could conceivably work, yes? Though scoping that out is something I'm certainly fine with

dmonroym:
@cartermp can we merge this PR and maintain the discussion on https://github.com/dotnet/project-system/issues/6080?

cartermp:
Yeah, I think this is merge-able for now. The design might change for .NET 6 but for now it's accurate in terms of what we _need_ to support.

PR #6135, opened 2020-04-28
Clean up of the Aliases property

davkean:
- Description that is copied from the one on AssemblyReferences
- Applied to the resolved package reference so that it gets persisted correctly

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6135)


davkean:
tag @nkolev92

davidwengier:
We should add unit test coverage for the comments you made on the other PR.

PR #6138, opened 2020-04-28
Fix build break

davkean:
This string was changed in master and Dev16.6.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6138)


melytc:
Good catch, thanks David!

PR #6146, opened 2020-04-30
Show diagnostics on project reference nodes

### LINKED ISSUE #2436: Diagnostics directly on P2P References do not display in the Dependencies node
**Steps**
Create a P2P reference from a netstandard1.3 project to a netstandard1.6 project

**Expected**
Nuget warning NU1201 in the error list, and a diagnostic message on the P2P reference in the dependencies node, with a warning icon that bubbles up.

**Actual**
NuGet warning NU1201 does appear, but the P2P node in the dependencies tree has no diagnostic or warning icon

Fixes are needed in:
- NuGet: generated diagnostic does not have a LibraryId or Target Graph (filed https://github.com/NuGet/Home/issues/5405)
- SDK: diagnostic on projects are not added to dependency node graph
- Project System: The P2P node in dependencies tree is currently only capable of showing information from the _referenced_ project, but the NuGet diagnostic comes from the _referring_ project. Previously, Project System tried to bypass NuGet/SDK and determine incompatibilities directly, but this leads to .NET Frameworks being treated as incompatible. This is fixed in https://github.com/dotnet/project-system/pull/2435

drewnoakes:
Fixes #2436.

Adds the relation required to show NuGet diagnostics as children of project reference nodes.

image

In future I'd like to distinguish between warnings and errors on the ancestors too. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6146)


davkean:
Nice!

PR #6152, opened 2020-05-04
Quick test to validate the an export can actually be constructed

davidwengier:
It probably won't help me falling into this trap, but it will give me something to feel shameful about when I finally track down the issue and realise I could have just run the tests to tell me the problem.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6152)

### DIFF
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentComposition.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentComposition.cs
@@ -15,7 +15,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         /// <summary>
         /// The list of assemblies that may contain <see cref="ProjectSystemContractProvider.System"/> exports.
         /// </summary>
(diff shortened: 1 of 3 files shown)


davidwengier:
I wasn't actually sure if multiple ImportingConstructors are allowed or not. It wouldn't surprise me if MEF had a convoluted overload resolution style system based on which parts where available or not ¯\\\_(ツ)_/¯

drewnoakes:
https://docs.microsoft.com/en-us/dotnet/framework/mef/attributed-programming-model-overview-mef#prerequisite-imports

> Each part may have only one constructor for use by the composition engine. Providing no parameterless constructor and no `ImportingConstructor` attribute, or providing more than one `ImportingConstructor` attribute, will produce an error.

davidwengier:
Thank you :D

davkean:
Bear in mind that you are allowed to have a parameter less constructor and one marked with ImportingConstructor.

davidwengier:
Thanks for that, had to update to not use MEF at all for this test as two ImportingConstructors meant the part was invalid so it wasn't part of the theory data.

Issue #5935, opened 2020-03-03
Review context menus for transitive dependencies


drewnoakes:
#5932 updated context menus within the Dependencies tree for the root, target frameworks, group nodes and individual top-level dependencies.

Some top-level dependencies may be expanded to drill into further transitive dependencies. Context menus on those transitive dependencies are currently showing `IDM_VS_CTXT_PROJWIN_FILECONTENTS` but more specific menu types should be used.

- [x] Create menus/groups for "Fakes" menu items to be added on relevant assembly nodes
- [x] Ensure "Scope to this" is available on all tree items (https://github.com/dotnet/project-system/pull/5932#issuecomment-593827721)
- [ ] "View in Object Browser" for various items

**Performance note:** By using dedicated menus per item type, it's possible to add items which are always visible and therefore avoid loading assemblies to call `QueryStatus` on items which appear in menus but are not invoked.

drewnoakes:
PR to define these menus is up: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/250942

drewnoakes:
https://github.com/NuGet/NuGet.Client/pull/3392/commits/a462d871eebed985c3a3fe7fd145d75f7d30267e uses these new menus for various tree items.

drewnoakes:
Closing this. We have other issues tracking "View in Object Browser". See #6253 for details.

PR #6153, opened 2020-05-04
Support context menus on transitive dependency tree items

drewnoakes:
Contributes to #5935.

Adds support for showing item-specific context menus and adds a default menu.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6153)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AttachedCollectionItemBase.cs
```diff
@@ -42,7 +42,8 @@ public abstract class AttachedCollectionItemBase
         private static readonly HashSet<Type> s_supportedPatterns = new HashSet<Type>
         {
             typeof(ITreeDisplayItem),
-            typeof(IBrowsablePattern)
+            typeof(IBrowsablePattern),
+            typeof(IContextMenuPattern)
```
davkean, line 46: Should it return null for this pattern if ContextMenuController is null?

drewnoakes: It works out the same in the end: http://index/?leftProject=Microsoft.VisualStudio.DesignTools.Diagnostics&leftSymbol=kx9opq4zj20s&file=Tool%5CUI%5CControls%5CTreeView%5CInteractionPatternProvider.cs&line=30

davkean:
Do you know why the other commands don't show?

drewnoakes:
> Do you know why the other commands don't show?

Pushed an update that brings in "Scope to This".

image

Issue #5703, opened 2019-12-05
Investigate providing browse objects lazily


drewnoakes:
Currently in CPS, all tree items are constructed with a populated browse object. This means each tree item retains its `IRule` instance even though the vast majority of these will never be displayed.

CPS provides these objects to VS in response to an event.

We should investigate whether there would be benefit to a CPS extension point that would allow producing these objects lazily.

These objects could use standard WinForms-era component model attributes for the properties grid rather than requiring XAML-backed rules.

---

Look at `GetPropertySlow` in CPS's `ProjectNode.Hierarchy.cs` and it's call through to `SimpleHierarchyNode.GetNodeProperties`.

Also look at CPS PR for folder items https://dev.azure.com/mseng/VSIDEProj/_git/VSIDEProj.CPS/pullrequest/141315?_a=files

PR #6155, opened 2020-05-04
Populate Dependency tree transitive items via the assets file

### LINKED ISSUE #603: Show content files from nuget packages as a distinct group in solution explorer

### LINKED ISSUE #927: Show path to package for graph nodes
That's the one piece of information I want when I use the property grid:

image

### LINKED ISSUE #1255: Group package/project dependencies by provider type
The dependency node shows the contents (Newtonsoft.Json) along side it's dependencies (the other things). These aren't siblings:

image

### LINKED ISSUE #1942: Improve presentation of dependencies tree search results
Currently when search happens in Solution explorer and user wants to find where dependency X or dependency with version Y comes from , we show only linear results, meaning: for each top level dependency, we show a flat list of unique dependencies from any level of hierarchy below it.  That means we would show that top level dependency depends on X or Y, but at what level? What exactly brings it there? 

We used to have search results with full hierarchy but:
- progression is super slow when dealing with deep hierarchies in search - VS hungs very often in this case
- results have multiple pages, so it becomes pretty unusable if something very common was searched.

Proposal: 
- we could request improvements in progression  to fix perf, but it is dangerous - code is very old there
- Or there were talks in CPS that they could redesign their tree logic, which would eliminate need for progression and graph nodes , and could help with search results too
- In any case we need to come up with representation that is usable when there are many results and deep hierarchies, may be create a completely separate tool window or extension for Ctrl+g search dialog.

See if there will be customer need for that when netstandard2.0 comes , maybe add telemetry - how often users search for dependenices or some polls etc ...

### LINKED ISSUE #2436: Diagnostics directly on P2P References do not display in the Dependencies node

### LINKED ISSUE #2509: Expanded unresolved Graph Nodes remain unresolved after problem is fixed
**Repro Steps**

Create a .NET Standard class library and add the following mismatched package references:

```xml
  <ItemGroup>
    <PackageReference Include="NuGet.Commands" Version="4.0.0" />
    <PackageReference Include="NuGet.Packaging" Version="3.5.0" />
  </ItemGroup>
```

This generates expected NU1605 package downgrade warnings for `NuGet.Packaging`. There is also a nested warning under `NuGet.Commands` which bubbles up as expected. Expand the Dependencies node to show nested warning under `NuGet.Commands`:

image

Resolve the problem by changing the version for NuGet.Packaging to 4.0.0

**Expected**
All unresolved nodes become resolved

**Actual**
While the top 3 levels of nodes do become resolved, the lower levels which contain Graph Nodes remain unresolved. This does not occur if the nodes were not expanded prior to fixing the package downgrade issue

### LINKED ISSUE #2582: Dependencies Node does not display Error icon where expected
**Repro**

Given the following Console App project

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="NuGet.Commands" Version="4.0.0" />
    <PackageReference Include="NuGet.Packaging" Version="3.5.0" />
  </ItemGroup>

</Project>
```

The resulting package downgrade diagnostics show up as errors in the Error List, but as Warnings in the Dependencies node:

image

This is despite the fact that the model attempts to distinguish between Errors and Warnings: [DiagnosticDependencyModel.cs#L38-L49](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Models/DiagnosticDependencyModel.cs#L38-L49)

### LINKED ISSUE #2705: Dependencies tree should show package's <frameworkAssemblies>
When referencing a package that includes framework assembly references, it's currently impossible to tell that the framework assembly references are being added to the project. They don't show up in the Assemblies node for the project, and they aren't listed under the package's node either.

The only way I've found to be able to see these references right now is by looking at the project.assets.json file

I'd really like to see these displayed in the Dependencies node, ideally as an item under the package's node.

### LINKED ISSUE #3335: Dependencies node should share identical subtrees between projects
(Split from #3309).

Putting in unknown, as this will be a significant amount of work, and we have a bunch of other things to do in the dependencies node code first.

### LINKED ISSUE #3435: Design-time build only provides top-level package/project dependencies, sourcing transitive references from project.assets.json
Project system half of dotnet/sdk#2100:

Notes from @abpiskunov from our meeting:

>Here is what we got at the meeting today:
>
> - Currently there is too much unnecessary data being sent through DT builds for nuget packages tree and assets file is being read multiple times for each configuration (TFM) and then again when new DT build is triggered. 
>
> - We agreed to redesign it in following way:
>
>    - DT build targets would send us only top level resolved packages (with the same metadata they had before, to avoid compatibility issues) and a timestamp for the assets file those packages are received from. SDK can cache this info and avoid reading unchanged assets file.
>    - In VS we will read top level packages for each configuration (TFM) in the similar way we do now – through DT builds. In addition to that we would have new logic that would be able to parse assets file (we will schedule parsing the assets file and stop/avoid parsing if file have old/mismatching timestamp than what we received form last DT build) and then map to top level packages. This would allow to show top level package quickly and then produce secondary lower level tree data.
>
>By having this new logic we could also try to optimize data structure in VS and keep unique package objects across solution, instead duplicating it per each project (this is orthogonal problem, but we might think about this while doing matching DT changes)

cc @Pilchie @davkean @abpiskunov

### LINKED ISSUE #3780: Diagnostics not shown under packages in the dependencies node

### LINKED ISSUE #4444: Show assemblies within shared framework targeting packs

### LINKED ISSUE #4474: Project path is used as an identifier, but can change over time causing breakages
The dependencies node uses a project's full path as an identifier for several purposes:

- Mapping project to `IDependenciesSnapshotProvider` (in `AggregateDependenciesSnapshotProvider`)
- Retrieving a snapshot from an `IDependency` for various purposes
- Retrieving a snapshot from a `GraphNode` for various purposes
- Identifying whether a P2P reference should be updated due to another project's changed snapshot
- Identifying whether a P2P reference's target has unresolved dependencies in order to show resolved/unresolved icon

Projects can be renamed, which changes that path. Using a value that changes over time as an identifier is asking for trouble.

Trying to update all usages of the project path in response to renames would involve a lot of code and still be open to breakages down the road.

We need a more robust identifier for projects, at least in the dependencies node subsystem.

The project path is treated as an opaque identifier. Using an autoincrementing integral ID instead seems the easiest robust alternative. This value would only be used inside the dependencies node code. It wouldn't have to be stable across VS restarts, though should be stable across project unload/reload.

Likely responsible for [755387](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/755387). I also observe deeper issues during testing.

---

Note, a month back a fix (e89b1f95, part of #4386) went in that stopped `DependenciesSnapshotProvider` from caching `ProjectPath` so it would update after a rename. That bug was patching over other project rename bugs that are now revealing themselves.

### LINKED ISSUE #5355: Package restore warning not appearing on package node
**Visual Studio Version**: 16.2.2

**Summary**:

Target framework warning on package reference should show error on package's node.

**Steps to Reproduce**:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.2</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="EntityFramework" Version="6.2.0" />
  </ItemGroup>

</Project>
```

image

This is a legitimate warning and the error list shows it:

> NU1701 Package 'EntityFramework 6.2.0' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.2'. This package may not be fully compatible with your project.

Expanding the node shows the warning:

image

In this example there's only one package reference, but if there were many it's not clear which has/have a problem.

**Expected Behavior**:

The "EntityFramework" node should have a yellow triangle.

**Actual Behavior**:

No yellow triangle.

**User Impact**:

Minor confusion.

### LINKED ISSUE #5697: Replace progression APIs with IAttachedCollection
We currently use progression APIs (graph nodes) for populating transitive dependencies in the Dependencies node.

There's an initiative by the IDE team to avoid loading the seven DLLs needed for progression to reduce startup time, working set and also make Solution Explorer more responsive.

To deliver that benefit to Visual Studio, the Dependencies node will need to move away from progression, using `IAttachedCollection` and related APIs instead. These are actually much nicer APIs from the looks of it.

### LINKED ISSUE #5740: Add package file/contained assembly paths to browse object

### LINKED ISSUE #5973: Dependencies node should respect PrivateAssets on ProjectReference
When a `ProjectReference` has `PrivateAssets` set to `true`, the dependencies node should not display that project's child project/package references.

#### App.csproj

```xml
<PropertyGroup>
  <TargetFramework>netcoreapp3.1</TargetFramework>
</PropertyGroup>
<ItemGroup>
  <ProjectReference Include="..\ClassLibrary1\ClassLibrary1.csproj" />
</ItemGroup>
```

#### ClassLibrary1.csproj

```xml
<ItemGroup>
  <ProjectReference Include="..\ClassLibrary2\ClassLibrary2.csproj" PrivateAssets="all" />
  <PackageReference Include="MetadataExtractor" Version="2.3.0" PrivateAssets="all" />
</ItemGroup>
```

#### App's Dependencies Tree

image

## Expected

`ClassLibrary1` in the `App` project does not display any child nodes.

---

Note that `App`'s `project.assets.json` file does not list these dependencies:

```json
{
  "version": 3,
  "targets": {
    ".NETCoreApp,Version=v3.1": {
      "ClassLibrary1/1.0.0": {
        "type": "project",
        "framework": ".NETStandard,Version=v2.0",
        "compile": {
          "bin/placeholder/ClassLibrary1.dll": {}
        },
        "runtime": {
          "bin/placeholder/ClassLibrary1.dll": {}
        }
      }
    }
  },
  ...
}
```

### LINKED ISSUE #6120: NFW: Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.<.ctor>g__Scan|2_0(ImmutableDictionary`2,ITargetFramework)
**Visual Studio Version**:
16.5

**Summary**:

This has ~411 hits and is being incorrectly bucketed with https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1108320/.

System.ArgumentException - System.ThrowHelper.ThrowArgumentException

```
["System.ThrowHelper.ThrowArgumentException(ExceptionResource)
>>>System.Collections.Generic.Dictionary`2.Insert(TKey,TValue,Boolean)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.<.ctor>g__Scan|2_0(ImmutableDictionary`2,ITargetFramework)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot..ctor(String,ITargetFramework,IProjectCatalogSnapshot,ImmutableDictionary`2)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.FromChanges(String,TargetedDependenciesSnapshot,IDependenciesChanges,IProjectCatalogSnapshot,ImmutableArray`1,IReadOnlyDictionary`2,IImmutableSet`1)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.DependenciesSnapshot.FromChanges(String,DependenciesSnapshot,ITargetFramework,IDependenciesChanges,IProjectCatalogSnapshot,ImmutableArray`1,ITargetFramework,ImmutableArray`1,IReadOnlyDictionary`2,IImmutableSet`1)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependenciesSnapshotProvider.<>c__DisplayClass34_0.<UpdateDependenciesSnapshot>b__0(DependenciesSnapshot)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependenciesSnapshotProvider.SnapshotUpdater.TryUpdate(Func`2,CancellationToken)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependenciesSnapshotProvider.UpdateDependenciesSnapshot(ITargetFramework,IDependenciesChanges,IProjectCatalogSnapshot,ImmutableArray`1,ITargetFramework,CancellationToken)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependenciesSnapshotProvider.<get_Subscribers>g__OnSubscriberDependenciesChanged|29_0(Object,DependencySubscriptionChangedEventArgs)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependencyRulesSubscriberBase`1.RaiseDependenciesChanged(ITargetFramework,IDependenciesChanges,AggregateCrossTargetProjectContext,IProjectCatalogSnapshot)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependencyRulesSubscriber.Handle(AggregateCrossTargetProjectContext,ITargetFramework,Tuple`3)
>>>Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Subscriptions.DependencyRulesSubscriberBase`1.<>c__DisplayClass12_0.<<OnProjectChangedAsync>b__1>d.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.<>c__DisplayClass13_0.<<LoadedProjectAsync>b__0>d.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>Microsoft.VisualStudio.Threading.ReentrantSemaphore.StackSemaphore.<>c__DisplayClass3_0.<<ExecuteAsync>b__0>d.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>Microsoft.VisualStudio.Threading.AwaitExtensions.ExecuteContinuationSynchronouslyAwaiter.GetResult()
>>>Microsoft.VisualStudio.Threading.ReentrantSemaphore.StackSemaphore.<ExecuteAsync>d__3.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>Microsoft.VisualStudio.ProjectSystem.OnceInitializedOnceDisposedUnderLockAsync.<ExecuteUnderLockCoreAsync>d__7.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>Microsoft.VisualStudio.ProjectSystem.DataReceivingBlockSlim`1.<ProcessInputQueueAsync>d__5.MoveNext()
>>>System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>>>Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.Rethrow(Exception)
>>>Microsoft.VisualStudio.ProjectSystem.ProjectErrorReporting.<>c__DisplayClass6_0.<SubmitErrorReport>b__0()
>>>Microsoft.VisualStudio.ProjectSystem.ExceptionFilter.<>c__DisplayClass2_0.<Guard>g__action|0()
>>>GuardMethodClass.GuardMethod(Func`1,Func`2,Func`2)"]
```

**Expected Behavior**:

**Actual Behavior**:

**User Impact**:

drewnoakes:
This PR merges the `feature/dependencies-tree-performance` branch.

Previously, the data model for a project's complete dependency graph was obtained via design-time builds. This data is only used to populate the "Dependencies" tree in Solution Explorer.

Design-time builds run when the project (and therefore also solution) is loaded, and again on project updates. The data such builds produce is retained in memory for the lifetime of the project.

In any given VS session, the majority of project's will never have their "Dependencies" node expanded. Loading the entire dependency graph up-front just in case it's needed slows down solution load, and more importantly increases the working set in VS.

This PR divides dependencies into two types:

1. Top-level dependencies, which are those the project references directly
2. Transitive dependencies, which are the closure of dependencies brought in by top-level dependencies

Top-level dependencies are handled in much the same way as before.

Transitive dependencies are handled differently:

- Data about transitive dependencies is loaded only when the user expands the "Dependencies" node enough to show a top-level dependency, or when searching in Solution Explorer
- Tree items are created lazily, only when needed
- For packages and projects, data about their transitive dependencies is obtained from the `project.assets.json` file
- The SDK has been changed to only return top-level items (https://github.com/dotnet/sdk/pull/11358)

This PR converts to `IAttachedCollection*` APIs instead of progression/graph nodes. We also provide an `ISearchProvider` that produces search results in the same structure as they're presented in the tree usually.

New concepts are introduced to define the tree of dependencies:

- `IRelatableItem` &mdash; each kind of item in the tree implements this (usually via `RelatableItemBase`)
- `IRelation` &mdash; describes a parent/child relationship between `IRelatableItem` types (usually implemented via `RelatableItemBase`)
- `DependenciesAttachedCollectionSourceProviderBase` subclasses identify top-level dependency tree items (of type `IVsHierarchyItem`) and specify their children, bridging between eagerly loaded hierarchy items and lazily constructed attached items

NuGet has APIs for parsing the assets file, however it's not packaged for external consumption. Therefore code for working with the assets file and producing relevant tree items in this PR is removed in towards the end, and will be migrated to a NuGet VSIX.

Comparing steady-state memory consumption for `Roslyn.sln` (dotnet/roslyn@b12baf7e57822a85e662275663113b68d0de52fc) with ~200 projects, approximately two minutes after load (averaged over three runs) gives:

| | Private bytes | All .NET heaps | `.dtbcache.v2` size |
|-|-|-|-|
| Before | 952 MB | 722 MB | 5,956 kB |
| After | 901 MB | 665 MB | 4,030 kB |
| Delta | -51 MB (5%) | -57 MB (8%) | -1,926 kB (32%) |

For 25 copies of `ProjectSystem.sln` (400 projects), approximately two minutes after load (averaged over three runs) gives:

| | Private bytes | `.dtbcache.v2` size |
|-|-|-|
| Before | 1.7 GB | 8,447 kB |
| After | 1.5 GB | 6,072 kB |
| Delta | 0.2 GB (12%) | -2,375 kB (28%) |

As well as improving performance, this work adds some new features and addresses several known bugs.

Fixes #3435 Design-time build only provides top-level package/project dependencies, sourcing transitive references from `project.assets.json`
Fixes #3780 Diagnostics not shown under packages in the dependencies node
Fixes #2436 Diagnostics directly on P2P References do not display in the Dependencies node
Fixes #2705 Dependencies tree should show package's `<frameworkAssemblies>`
Fixes #5740 Show resolved path for package and its assemblies in browse object
Fixes #603 Show content files from nuget packages as a distinct group in solution explorer
Fixes #1255 Group package/project dependencies by provider type
Fixes #5697 Replace progression APIs with IAttachedCollection
Fixes #927 Show path to package for graph nodes
Fixes #4444 Show assemblies within shared framework targeting packs
Fixes #1942 Improve presentation of dependencies tree search results
Fixes #2509 Expanded unresolved Graph Nodes remain unresolved after problem is fixed
Fixes #2582 Dependencies Node does not display Error icon where expected
Fixes #4474 Project path is used as an identifier, but can change over time causing breakages
Fixes #5355 Package restore warning not appearing on package node
Fixes #5973 Dependencies node should respect PrivateAssets on ProjectReference
Fixes #6120 NFW: Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.Snapshot.TargetedDependenciesSnapshot.<.ctor>g__Scan|2_0(ImmutableDictionary`2,ITargetFramework)
Closes #3335 Dependencies node should share identical subtrees between projects
Contributes to #6012 Dependencies tree should not collapse on update
Contributes to #5051 DependencyIDs create a lot of long, duplicate strings
Contributes to #5703 Investigate providing browse objects lazily
Contributes to #5935 Review context menus for transitive dependencies
Contributes to [AB#723450](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/723450) Reduce usage of progression search in solution explorer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6155)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IUnconfiguredProjectVsServices.cs
```diff
@@ -20,5 +20,10 @@ internal interface IUnconfiguredProjectVsServices : IUnconfiguredProjectCommonSe
         ///     Gets <see cref="IVsProject4"/> provided by the <see cref="UnconfiguredProject"/>.
         /// </summary>
         IVsProject4 VsProject { get; }
+
+        /// <summary>
+        ///     Gets <see cref="IPhysicalProjectTree"/> provided by the <see cref="UnconfiguredProject"/>.
+        /// </summary>
+        IPhysicalProjectTree ProjectTree { get; }
```
davkean, line 27: Why in the VS layer?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateContainedByRelationCollection.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public int IndexOf(object value) => _parentItems.IndexOf(value);
+
+        object IList.this[int index]
+        {
+            get => _parentItems[index];
+            set => throw new NotSupportedException();
+        }
+
+        void ICollection.CopyTo(Array array, int index) => throw new NotSupportedException();
```
davkean, line 48: Why isn't this implemented?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
```diff
@@ -4,3 +4,98 @@ Microsoft.VisualStudio.ProjectSystem.VS.Properties.CSharp.CSharpProjectConfigura
...
+virtual Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.FontStyle.get -> System.Windows.FontStyle
+virtual Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.FontWeight.get -> System.Windows.FontWeight
+virtual Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.GetBrowseObject() -> object?
+virtual Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.OverlayIconMoniker.get -> Microsoft.VisualStudio.Imaging.Interop.ImageMoniker
+virtual Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.StateIconMoniker.get -> Microsoft.VisualStudio.Imaging.Interop.ImageMoniker
+virtual Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.StateToolTipText.get -> string?
+Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.BrowseObjectBase
```
davkean, line 63: Why do we expose this type?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -221,4 +218,31 @@ In order to debug this project, add an executable project to this solution which
...
     <value>Renaming Type from '{0}' to '{1}'</value>
   </data>
+  <data name="FrameworkAssemblyBrowseObjectClassName" xml:space="preserve">
+    <value>Framework Assembly</value>
+  </data>
+  <data name="FrameworkAssemblyAssemblyNameDisplayName" xml:space="preserve">
+    <value>Assembly Name</value>
+  </data>
+  <data name="FrameworkAssemblyAssemblyNameDescription" xml:space="preserve">
+    <value>The framework assembly's assembly name.</value>
```
davkean, line 228: "The name of the framework assembly".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -221,4 +218,31 @@ In order to debug this project, add an executable project to this solution which
...
+  </data>
+  <data name="FrameworkAssemblyPathDisplayName" xml:space="preserve">
+    <value>Path</value>
```
davkean, line 231: "Full Path" to be consistent.

src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
```diff
@@ -102,5 +103,17 @@ public static bool Any<T>(this T[] array, Func<T, bool> predicate)
 
             return false;
         }
+
+        public static ImmutableArray<TValue> ToImmutableValueArray<TKey, TValue>(this Dictionary<TKey, TValue> source)
```
davkean, line 107: Why "ImmutableValueArray" instead of just "ImmutableArray"?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/BrowseObjectBase.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.ComponentModel;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections
+{
+    /// <summary>
+    /// This is a slightly modified copy of Microsoft.VisualStudio.Shell.LocalizableProperties.
+    /// http://index/#Microsoft.VisualStudio.Shell.12.0/LocalizableProperties.cs.html
+    /// Unfortunately we can't reuse that class because the GetComponentName method on
+    /// it is not virtual, so we can't provide a name string for the VS Property Grid's
```
davkean, line 13: Why don't we just make it virtual? That would be preferable versus adding public API to our assembly for something that doesn't belong in this assembly.

drewnoakes: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/248337

src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
drewnoakes: I would expect `ToImmutableArray` to return `ImmutableArray<KeyValuePair<TKey, TValue>>`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
drewnoakes: Regular assembly references use "Path":

image

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
drewnoakes: Agree this reads better, but it's inconsistent with the other name descriptions. Will change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateContainedByRelationCollection.cs
drewnoakes: Discussed in earlier review. This method is not called. The implementation for "contains" is non-trivial so I opted against adding a chunk of code that wouldn't be used.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IUnconfiguredProjectVsServices.cs
drewnoakes: I don't understand. This member existed on the implementation. This just pulls it up to the interface. It's needed in `DependenciesTreeSearchProvider`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
drewnoakes: Will do.

src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
davkean: Oh, why isn't this just source.Values.ToImmutableArray()?

drewnoakes: Because that allocates an enumerator.

PR #6157, opened 2020-05-04
Mark command as client compliant

melytc:
This marks the 'Edit project file' command as client compliant.

Testing is **blocked**, not merging this to master yet.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6157)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -38,7 +38,7 @@
     <PackageReference Update="Microsoft.DevDiv.Validation.MediaRecorder"                              Version="15.0.199" />
 
     <!-- VS SDK -->
-    <PackageReference Update="Microsoft.VSSDK.BuildTools"                                             Version="16.5.2044" />
+    <PackageReference Update="Microsoft.VSSDK.BuildTools"                                             Version="16.7.2" />
     <PackageReference Update="Microsoft.VisualStudio.ComponentModelHost"                              Version="16.0.198-g52de9c2988"/>
     <PackageReference Update="Microsoft.VisualStudio.Composition"                                     Version="16.1.8"/>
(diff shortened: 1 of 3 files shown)


davkean:
Now that we know when and how commands will stop up, @tmeschter do we have a bug to change this code to use the real EditProjectFile command?

tmeschter:
@melytc Do we need to add the .clientenabledpkg files that the CPS change has? E.g. in https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/240597.

melytc:
@tmeschter I don't think so, looking at the [wiki](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/8735/How-to-mark-a-component-as-compliant-for-Nexus?anchor=mark-extensions-and-packages-as-compliant), I followed the `Extensions` steps rather than the `Legacy packages` ones. Let me confirm this.

melytc:
@tmeschter, you're right, thanks!

davidwengier:
So we're a legacy package _and_ and extension?

melytc:
Talked with Ankit from the team leading this. We both misunderstood and we don't need that clientenabledpkg file, since we create an extension (vsix file) that includes the `extension.vsixmanifest` file, `.pkgdef` files and already a `.clientenabledpkg` file. 

Sorry for the confusion there, deleting that last commit :)

dmonroym:
Tests have passed twice, they're not getting reported back correctly though: https://dev.azure.com/dnceng/public/_build/results?buildId=640251&view=results

PR #6167, opened 2020-05-06
Remove NuGet package references

drewnoakes:
These packages are the reason we have to move a bunch of code from here to NuGet. They contain assemblies for which there are no binding redirects.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6167)


swesonga:
Just curious about these NuGet changes: what was the downside of was adding binding directs for those assemblies?

PR #6172, opened 2020-05-06
Force severity to be specified

davkean:
Force severity of specified for faults, and change PackageRestoreProgressTrackerInstance to up to it LimitedFunctionality so that it gets unsubscribed from operation progress when it faults.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6172)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/AddItemDialogService.cs
```diff
@@ -81,7 +81,7 @@ private async Task<bool> ShowDialogAsync(IProjectTree node, __VSADDITEMFLAGS fla
                 ref filter,
                 out _);
 
-            if (result == HResult.Ole.PromptSaveCancelled)
+            if (result == HResult.Ole.PromptSaveCancelled || result == HResult.Abort)
```
drewnoakes, line 84: Is this intentional? It seems unrelated.

PR #6175, opened 2020-05-07
Make IVsHierarchyItemExtensions public

drewnoakes:
This type is needed by the NuGet implementation. It serves as a way for others to extract information from tree items for other purposes too, such as command handlers.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6175)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
@@ -7,6 +7,7 @@ Microsoft.VisualStudio.ProjectSystem.VS.Properties.VisualBasic.VisualBasicProjec
 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AggregateContainsRelationCollection
 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AggregateContainsRelationCollection.OnStateUpdated() -> void
 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.IAggregateRelationCollection.HasItemsChanged -> System.EventHandler!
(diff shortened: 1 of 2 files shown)


davkean:
Why does this belong in our code base? We don't own this type.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Unshipped.txt
```diff
@@ -51,6 +52,12 @@ Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.At
 abstract Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.IconMoniker.get -> Microsoft.VisualStudio.Imaging.Interop.ImageMoniker
 abstract Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.Priority.get -> int
 override Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.AttachedCollectionItemBase.ToString() -> string!
+static Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.IVsHierarchyItemExtensions.TryFindTarget(this Microsoft.VisualStudio.Shell.IVsHierarchyItem! item, out string? target) -> bool
```
davkean, line 55: This returns a single value, when it could be one of many in the future.

davkean:
What exactly are these returning data for? Top-level packages in a tree? Is there a reason why this data is being persisted in flags? This is just data that should be provided by ProjectItem.Properties and hence these helpers don't exist.

drewnoakes:
> We don't own this type, IVsHierarchyItem, so calling something IVsHierarchyItemExtensions is bit presumptuous.

I don't follow. We have a bunch of extension classes for types we don't own.

> This is just data that should be provided by ProjectItem.Properties/BrowseObject and hence these helpers don't exist.

Will investigate.

PR #6178, opened 2020-05-11
Make some images public

drewnoakes:
These images are the unadorned versions of icons we ship for top-level dependencies. Scenarios exist where transitive dependency providers will want to create tree items that use the same icon as top-level dependencies. Making this public enables this scenario. For example, the NuGet provider shows nested project references.

These names match those above (with "private" and "warning" suffixes).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6178)


davidwengier:
Is this just allowing for us to change what, for example, the Application icon looks like at a top level, and have the sub level usages match? Because I can't imagine when we would do that and not also change the name of the image. If the images came from our repo I'd understand I think, but I'm guessing that the images come from the platform, right?

drewnoakes:
Some images from our repo, some from platform. Names come from platform. The suffixed versions are already public API.

PR #6184, opened 2020-05-13
Turn on CPS diagnostics for debugging

davkean:
- Turn on tree verification and RDT validation
- Turns on item ids in caption for display nodes under DEBUG
- Turns on capabilities tree under DEBUG
- Removed unneeded variables for client debugging as our bits aren't present

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6184)

### DIFF
--- a/setup/ProjectSystemSetup/Properties/launchSettings.json
+++ b/setup/ProjectSystemSetup/Properties/launchSettings.json
@@ -8,7 +8,9 @@
         "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
         "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
         "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
-        "PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED": "1"
+        "PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED": "1",
+        "CPS_DiagnosticRuntime": "1",
+        "CPS_MetricsCollection": "1"
       }
     },
     "Start (with native debugging)": {
@@ -19,7 +21,9 @@
         "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
         "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
         "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
-        "PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED": "1"
+        "PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED": "1",
+        "CPS_DiagnosticRuntime": "1",
+        "CPS_MetricsCollection": "1"
       },
       "nativeDebugging": true
     },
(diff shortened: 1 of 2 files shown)


drewnoakes:
> - Turns on item ids in caption for display nodes

Does this change how the tree looks even when using Ctrl+F5? Given I don't think we'd want this so often, it feels like a better candidate for a `ProjectCapabilities` item, similar to `DiagnoseCapabilities`.

davkean:
It's a service capability so it doesn't actually work like that now, so we either turn it on always, or alternatively we have different launch profile.

davidwengier:
A different launch profile would be ideal I think, or a command line argument on `launch.cmd`. I don't like the idea that we will be using a view of Visual Studio that users won't be using, even if the difference is slight.

drewnoakes:
I often paste screenshots to GitHub for customers or in PR descriptions.

Seeing item IDs is useful only some of the time. A different launch profile would be better.

davidwengier:
I've approved this, its easy enough to change if it is too obnoxious, its not shipping code.

davkean:
I wanted to play around with this a bit before merging to see how obnoxious it was.

setup/ProjectSystemSetup/Properties/launchSettings.json
```diff
@@ -44,19 +50,15 @@
...
     "Start VS Online Client": {
       "commandName": "Executable",
       "executablePath": "$(DevEnvDir)devenv.exe",
-      "commandLineArgs": "/rootsuffix $(VSSDKTargetPlatformRegRootSuffix) /client \"vsls:?workspaceId=<workspace_ID>&remoteJoin=true\"",
-      "environmentVariables": {
-        "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
-        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
-        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
-        "PROJECTSYSTEM_PROJECTOUTPUTPANEENABLED": "1"
-      }
```
drewnoakes, line 59: Has this been tested?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/DisplayItemIdsOnItemsCapabilityProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.ComponentModel.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Microsoft.VisualStudio.Packaging
+{
+    [Export(ExportContractNames.Scopes.ProjectService, typeof(IProjectCapabilitiesProvider))]    
+    [AppliesTo(ProjectCapabilities.AlwaysApplicable)]
+    internal class DisplayItemIdsOnItemsCapabilityProvider : ProjectCapabilitiesProviderBase
```
drewnoakes, line 15: Does this type add `CPS_DisplayItemIdOnProjectNodes` and `DiagnoseCapabilities` to all projects unconditionally?

setup/ProjectSystemSetup/Properties/launchSettings.json
davkean: This line doesn't even F5. This has not been tested - but these environment variables affect server components, not client providers.

davkean:
This feature is replacing turning the item ids and capabilities tree: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/260022. As soon as we enable the Diagnostic runtime, this button will show up.

PR #6189, opened 2020-05-14
Use Module instead of class with private ctor

Youssef1313:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6189)

### DIFF
--- a/src/Common/ManagedCodeMarkers.vb
+++ b/src/Common/ManagedCodeMarkers.vb
@@ -11,63 +11,59 @@ Namespace Microsoft.Internal.Performance
         ' Singleton access
         Public Shared ReadOnly Instance As CodeMarkers = New CodeMarkers()
 
-        Private Class NativeMethods
-
-            ' Add a private constructor to prevent compiler from generating a default constructor (FxCop warning CA1812)
-            Private Sub New()
-            End Sub
+        Private Module NativeMethods
 
             ' ********************* Imported Win32 functions *********************
             ' Code markers test function imports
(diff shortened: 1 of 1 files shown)


davkean:
Thanks for the contribution, the build failed:

```
    F:\workspace\_work\1\s\src\Common\ManagedCodeMarkers.vb(14,9): error BC30617: 'Module' statements can occur only at file or namespace level. [F:\workspace\_work\1\s\src\Microsoft.VisualStudio.AppDesigner\Microsoft.VisualStudio.AppDesigner.vbproj]


```

dmonroym:
Thanks again for the contribution.

The build now has the following errors:
F:\workspace\_work\1\s\src\Common\ManagedCodeMarkers.vb(9,20): error BC31089: Types declared 'Private' must be inside another type. 
F:\workspace\_work\1\s\src\Common\ManagedCodeMarkers.vb(23,20): error BC30451: 'TestDllName' is not declared. It may be inaccessible due to its protection level.
F:\workspace\_work\1\s\src\Common\ManagedCodeMarkers.vb(38,20): error BC30451: 'ProductDllName' is not declared. It may be inaccessible due to its protection level.

Youssef1313:
Well, Although I think `Module` is more appropriate than a class with private ctor, but it seems things may break. Changing the `Module` to public doesn't seem like a good idea to fix the first error (`Types declared 'Private' must be inside another type.`). So, I'm going to close.

PR #6191, opened 2020-05-15
Fix Add XXX Reference items missing

davkean:
Fixed:  https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121993/

We were overaggressive in hiding these, when we only want them hidden for Folders.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6191)


davkean:
@drewnoakes  As an FYI.

PR #6195, opened 2020-05-15
Ban Debugger.Launch

tmeschter:
Ban the use of `System.Diagnostics.Debugger.Launch()`. We've seen other teams accidentally commit code that used this method, leading to a completely unexpected dialog appearing in the middle of a scenario.

It's fine to use this in local development; in that case you can suppress the error with  #pragmas. Just please remember to remove them. 😊

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6195)


tmeschter:
Maybe... I can see Break() being used (more) extensively for local testing, but it does have the same fundamental problem...

PR #6219, opened 2020-05-21
Opt-in for VS Codespaces

### LINKED ISSUE #6217: Opt-in flag for VS Codespaces for c#
**Visual Studio Version**:

**Summary**:
Potentially 3rd party project systems will get installed on the server in the future. This work is  to require all project systems to opt-in to VS Codespaces. This is an opt-in model for the .NET Core project systems that we know support VS Codespaces. 

**Steps to Reproduce**:

Not applicable

**Expected Behavior**:

**Actual Behavior**:

**User Impact**:

arkalyanms:
Fixes #6217.

The build will fail for a day until the CPS packaging changes get ingested.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6219)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
@@ -45,6 +45,7 @@ using Microsoft.VisualStudio.ProjectSystem.VS;
     Capabilities = ProjectTypeCapabilities.CSharp,
     DisableAsynchronousProjectTreeLoad = true,
     PossibleProjectExtensions = "csproj",
-    NewProjectRequireNewFolderVsTemplate = true)]
+    NewProjectRequireNewFolderVsTemplate = true,
+    SupportsCodespaces = true)]
 [assembly: ProvideDiffSupportedContentType(".csproj", "")]   // Empty string because content type is not important, we just want to tell the diff that the file type is supported
 [assembly: ProvideEditorFactoryMapping("{f6819a78-a205-47b5-be1c-675b3c7f0b8e}", ".csproj")] // Use the XML editor
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
@@ -15,7 +15,8 @@ using Microsoft.VisualStudio.ProjectSystem.VS;
     Capabilities = ProjectTypeCapabilities.VisualBasic,
     DisableAsynchronousProjectTreeLoad = true,
     PossibleProjectExtensions = "vbproj",
-    NewProjectRequireNewFolderVsTemplate = true)]
+    NewProjectRequireNewFolderVsTemplate = true,
+    SupportsCodespaces = true)]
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
```diff
@@ -45,6 +45,7 @@
     Capabilities = ProjectTypeCapabilities.CSharp,
     DisableAsynchronousProjectTreeLoad = true,
     PossibleProjectExtensions = "csproj",
-    NewProjectRequireNewFolderVsTemplate = true)]
+    NewProjectRequireNewFolderVsTemplate = true,
+    SupportsCodespaces = true)]
```
GearedToWar, line 49: I thought this was for all .NET core projects. Is there a reason VB and F# are not included currently?

davkean:
This has a build error because this attribute doesn't have this property:

> Packaging\ProjectTypeRegistration.cs(49,5): error CS0246: The type or namespace name 'SupportsCodespaces' could not be found (are you missing a using directive or an assembly reference?) [F:\workspace\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\Microsoft.VisualStudio.ProjectSystem.Managed.VS_dy1whlio_wpftmp.csproj]

I don't see an SDK update as part of this change, which package version includes this new API?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
davkean: Agreed; both C# and VB need to be opt'd in.

How does this interplay with selectors? The GUID persisted in the solution is the legacy project system, which the selector redirects to this one. Does this new property get handled by that?  @davidwengier

davkean:
>  CPS packaging changes get ingested.

Oh missed this, which explains the build error.

GearedToWar:
Arun is working on this.

Mohit

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
davkean: What about Shared Projects?

davidwengier: > How does this interplay with selectors? The GUID persisted in the solution is the legacy project system, which the selector redirects to this one. Does this new property get handled by that? @davidwengier

We discussed this, and I don't know. It needs to be tested. It depends exactly where the code is that reads this (or perhaps should I say "when" the code is) as to whether its after the pre-fetch or not.

GearedToWar: > What about Shared Projects? That registration is in the VS repository.

t's in the CPS repo - codesharingproj.pkgdef. And yes, that should also be opted in. @arkalyanms

arkalyanms: > Agreed; both C# and VB need to be opt'd in.
@davkean Made the change assuming you meant F# and VB.

Shared projects PR - https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/250253

PR #6228, opened 2020-05-25
Opt in for codespaces

### LINKED ISSUE #6217: Opt-in flag for VS Codespaces for c#

davidwengier:
Fixes #6217.
Replaces #6219.

Upgrading the CPS dependency, to be able to merge #6219, brought in a bunch of other packages that needed updating, which needed changes for `Obsolete` attributes and new code analysis warnings etc. so thought I'd do this as a new PR to make sure the changes were clear to all.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6228)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
```diff
@@ -43,7 +43,7 @@ public int Exec(List<WorkspaceVisualNodeBase> selection, Guid pguidCmdGroup, uin
         {
             if (pguidCmdGroup == CommandGroup.ManagedProjectSystemClientProjectCommandSetGuid)
             {
-                var nCmdIDInt = (int)nCmdID;
+                int nCmdIDInt = (int)nCmdID;
```
davkean, line 46: This looks like a bug?

davidwengier:
This builds on my machine fine, but fails in CI with "Microsoft.VisualStudio.Utilities 16.6.30008.236 depends on Microsoft.ServiceHub.Client (>= 2.4.224) but Microsoft.ServiceHub.Client 2.4.224 was not found. An approximate best match of Microsoft.ServiceHub.Client 2.4.227 was resolved". I'll see if I can find better versions or reference it directly or something, but I wonder if perhaps the build images need updating too.

davkean:
How would updating the build images resolve the problem? The build images don't come with packages.

davidwengier:
Don't the images have Visual Studio installed, hence the "VS2019" in the name? If the build works on my machine with latest Visual Studio installed, and fails in CI with a presumably older Visual Studio installed, then it seems reasonable that updating the image could fix the problem. I don't know if that _is_ the solution, but it seems like a reasonable conclusion to me ¯\\\_(ツ)_/¯

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
davidwengier: It's consistent with 16.6 and 15.9 ¯\\\_(ツ)_/¯

PR #6261, opened 2020-06-01
Remove full path from items

### LINKED ISSUE #5837: Remove FullPath from source item evaluation rules for C#/VB projects
It is the single largest string in the item properties collection in project evaluation rule snapshots, and also prevents the immutable properties collection to be shared cross different items.

drewnoakes:
Fixes #5837.

Removing `FullPath` from evaluated item metadata helps perf in two ways:

1. These can be lengthy paths, and there can be a lot of them. They no longer need to be retained, but can be computed as needed via `MakeRooted`.
2. Having `FullPath` in an item's metadata mostly guaranteed the metadata to be unique to that item, meaning CPS couldn't share the metadata data structure between multiple items. With this change, such sharing becomes possible.

Removes dependency from `DesignTimeInputsDataSource` on item `FullPath`.

Removes `FullPath` from the default item rules.

F# projects require `FullPath` in `TreeItemOrderPropertyProvider`, so this PR creates new FSharp overrides for those rules that preserve `FullName` only in F# projects. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6261)


swesonga:
> F# projects require FullPath in TreeItemOrderPropertyProvider, so this PR creates new FSharp overrides for those rules that preserve FullName only in F# projects.

What leads to the conclusion that this property is only used by F#?

drewnoakes:
> What leads to the conclusion that this property is only used by F#?

The `TreeItemOrderPropertyProvider` is only used in F# projects, as the order of files in an F# project is significant (unlike C#/VB). It could be used in other project types too if they add `SortByDisplayOrder` and `EditableDisplayOrder` capabilities, in which case `FullPath` would not be present and the feature would not work.

I looked into working around the absence of `FullPath` but there just isn't enough information at the place its used to replace it (https://github.com/dotnet/project-system/issues/5837#issuecomment-636636672) and @davkean had concerns about that direction. There may be another way but I can't see it. If we find one, we can remove these F# rules.

Issue #6210, opened 2020-05-19
Add missing properties to debug property pages


tmeschter:
ExecutableDebugPropertyPage.xaml and ProjectDebugPropertyPage.xaml are missing the following properties:

- [ ] "Use remote machine" and corresponding machine name
- [ ] "Authentication mode"
- [ ] ~~"Environment variables"~~ This is being tracked by #6268 
- [ ] "Enable native code debugging"
- [ ] "Enable SQL Server debugger"

PR #6262, opened 2020-06-03
Collapse related types into one

tmeschter:
Now that a single type can intercept multiple different properties collapse all of the  interceptors that handle the common properties of the active launch profile into a single type. This eliminates a great deal of redundant code.

Related to #6210.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6262)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptedProjectProperties/ActiveLaunchProfileValueProvidersTests.cs
```diff
@@ -56,9 +56,9 @@ public async Task ExecutablePath_OnGetEvaluatedPropertyValueAsync_GetsExecutable
 
             var project = UnconfiguredProjectFactory.Create();
             var threadingService = IProjectThreadingServiceFactory.Create();
-            var launchProfileProvider = new ExecutablePathValueProvider(project, settingsProvider, threadingService);
+            var launchProfileProvider = new ActiveLaunchProfileCommonValueProvider(project, settingsProvider, threadingService);
 
-            var actualValue = await launchProfileProvider.OnGetEvaluatedPropertyValueAsync(string.Empty, string.Empty, Mock.Of<IProjectProperties>());
+            var actualValue = await launchProfileProvider.OnGetEvaluatedPropertyValueAsync("ExecutablePath", string.Empty, Mock.Of<IProjectProperties>());
```
swesonga, line 61: >"ExecutablePath" [](start = 91, length = 16)

Would it be better to use the constants from ActiveLaunchProfileCommonValueProvider instead of all the inlined strings?

tmeschter: Done.

PR #6269, opened 2020-06-08
Add further debug properties

### LINKED ISSUE #6210: Add missing properties to debug property pages

tmeschter:
Fixes #6210.

This updates the ExecutableDebugPropertyPage and ProjectDebugPropertyPage to include the following properties:

- "Use remote machine" and the corresponding machine name
- "Authentication mode"
- "Enable native code debugging"
- "Enable SQL Server debugging"

To support this a new `ActiveLaunchProfileExtensionValueProvider` type has been added which knows how to read and write properties stored in `ILaunchProfile.OtherSettings`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6269)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileExtensionValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileExtensionValueProvider.cs
@@ -28,6 +28,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         internal const string RemoteDebugMachinePropertyName = "RemoteDebugMachine";
         internal const string SqlDebuggingPropertyName = "SqlDebugging";
 
+        // The CPS property system will map "true" and "false" to the localized versions of
+        // "Yes" and "No" for display purposes, but not other casings like "True" and
+        // "False". To ensure consistency we need to map booleans to these constants.
+        private const string True = "true";
+        private const string False = "false";
+
         private readonly UnconfiguredProject _project;
         private readonly ILaunchSettingsProvider _launchSettingsProvider;
         private readonly IProjectThreadingService _projectThreadingService;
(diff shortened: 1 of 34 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileExtensionValueProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        public override Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+        {
+            _projectThreadingService.RunAndForget(async () =>
+            {
+                ILaunchSettings launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
+
+                var writableLaunchSettings = launchSettings.ToWritableLaunchSettings();
+                var activeProfile = writableLaunchSettings.ActiveProfile;
```
davidwengier, line 85: Is it possible for "Active" to change before we get to this line? Perhaps it would be better to capture the name of the project in question outside of the `RunAndForget` and update the specific profile?

tmeschter: There may we ways to reduce the window of time where that could potentially occur, but to guarantee that we never hit that situation we would need to ensure all paths that read or write the active profile gain exclusive access to it. We should keep that in mind, but right now I think it's unlikely to happen and we can leave things as they are.

PR #6276, opened 2020-06-10
Remove some target framework usages

drewnoakes:
Relates to #5051.

`IDependency` instances are always accessed within the context of a `TargetedDependenciesSnapshot` which determines their `TargetFramework`.

Dependency IDs previously also included the TFM as a prefix. This change removes that prefix. This will further reduce memory consumption by the dependencies node.

The dependency ID (now of form `ProviderType/ModelId`) is no longer used as the `FilePath` for created `IProjectTree` items. This means `DependenciesProjectTreeProvider` no longer overrides `GetPath` and `FindByPath`, resulting in these tree items having a pseudo path of form `>123`, where the number is the project tree node's identity.

Collectively these changes allow removal of target framework parameters in several places, as well as the simplification of `IDependencyViewModel`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6276)


davidwengier:
Wow, I didn't realise Codeflow added the squirrel automatically!

Issue #4196, opened 2018-10-30
Why escape model ID in dependency ID?


drewnoakes:
A dependency's ID is (usually) composed of:

1. the target framework name,
2. the dependency's provider type, and
3. a "model ID", which is usually a path.

These three strings are concatenated with `/` between.

In the process, the model ID is escaped:

- `/` becomes `\`
- `..` becomes `__`

Why does this escaping happen? As far as I can tell this is just an opaque identifier. I'm hoping we can remove this escaping as part of some work to reduce string allocations.

https://github.com/dotnet/project-system/blob/470e13362e2a8898845af5df177afd17d10d1595/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Snapshot/Dependency.cs#L393-L394

//cc: @abpiskunov

abpiskunov:
It is needed because so weirdness in between Graphnodes communication and CPS, i don't remember specifics, but without normalization our nodes were not processed correctly, since we put id into node's filePath property which somewhere along the way in Graph logic (or CPS) was treated as path and was trimmed/modified and we could not see actual ID in graph nodes logic.

drewnoakes:
Thanks for the explanation.

Pilchie:
Similar to https://github.com/dotnet/project-system/issues/2559 and https://github.com/dotnet/project-system/issues/2558?

drewnoakes:
@Pilchie fantastic, thanks.

PR #6282, opened 2020-06-12
Simplify dependency node model data

### LINKED ISSUE #5051: DependencyIDs create a lot of long, duplicate strings

### LINKED ISSUE #6267: "Remove" item missing from package reference
**Visual Studio Version**: master & 16.7p2

Does not repro in 16.6.1. This is a regression.

image

This reference is explicit in the project and should be removable. There should be a "Remove" menu item on this reference.

drewnoakes:
Fixes #6267.
Fixes #5051.

Some simplification of data used in the modelling of dependencies.

1. Remove the composite dependency ID. #6276 removed target framework from the ID. This PR splits the remaining parts (provider type and model ID) out into their own fields so no string concatenation occurs, and all escaping (#4196) is removed. Methods `IdEquals`, `GetID` and `GetTopLevelId` are all removed.

2. Remove `IDependencyModel.Name` throughout (leave on public API as obsolete).

3. #6276 removed `FilePath` from several places, but that was too aggressive. The path is needed for some features such as Roslyn's analyzer nodes (see sibling PR dotnet/roslyn#45101). This change also brings back a minimal implementation of `DependenciesProjectTreeProvider.GetPath`/`FindByPath`.

4. Derive `IDependency` from `IDependencyViewModel`, avoiding some allocations when producing `IProjectTree` objects.

5. Fix dependency removal (#6267) and simplify implementation of `CanRemove` (which now only checks `SupportsRemove` flag, avoiding snapshot lookup) and `RemoveAsync` (which can now uses the `IProjectTree.FilePath` to identify shared projects, avoiding snapshot lookup).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6282)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/DependencyModel.cs
```diff
@@ -19,21 +19,23 @@ private enum DependencyFlags : byte
...
             string originalItemSpec,
             ProjectTreeFlags flags,
             bool isResolved,
             bool isImplicit,
             IImmutableDictionary<string, string>? properties,
             bool isVisible = true)
         {
-            Requires.NotNullOrEmpty(path, nameof(path));
+            // IDependencyModel allows original item spec to be null, but we can satisfy a more strict
+            // requirement on this values for the dependency types produced internally.
```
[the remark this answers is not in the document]
drewnoakes: Thanks. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -100,75 +100,50 @@ Task OnUnconfiguredProjectUnloading(object? sender, EventArgs args)
...
+        }
 
-            DependenciesSnapshot snapshot = _dependenciesSnapshotProvider.CurrentSnapshot;
-            if (snapshot == null)
-            {
-                return false;
-            }
+        public override IProjectTree? FindByPath(IProjectTree root, string path)
+        {
+            Assumes.True(root.Flags.Contains(ProjectTreeFlags.ProjectRoot), "Expected search to operate from project root");
```
davkean, line 117: This is not a valid assumption - why was this added?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/DependencyModel.cs
```diff
@@ -19,21 +19,23 @@ private enum DependencyFlags : byte
...
+            // IDependencyModel allows original item spec to be null, but we can satisfy a
+            // more strict requirement for the dependency types produced internally.
+            // External providers may not have a meaningful value, but do not use this type.
             Requires.NotNullOrEmpty(originalItemSpec, nameof(originalItemSpec));
 
             Path = path;
             OriginalItemSpec = originalItemSpec;
             Properties = properties ?? ImmutableStringDictionary<string>.EmptyOrdinal;
-            Caption = path;
+            Caption = originalItemSpec;
```
davkean, line 38: This looks strange that we would use original item spec without any manipulation for items, what items does this affect?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -100,75 +100,50 @@ Task OnUnconfiguredProjectUnloading(object? sender, EventArgs args)
...
-                return false;
-            }
+            // Returning null from a root graft causes CPS to use the "pseudo path" for the item, which has
+            // form ">123" where the number is the item's identity. This is a short string (low memory overhead)
+            // and allows fast lookup. So in general we want to return null here unless there is a compelling
+            // requirement to use the path.
+            //
+            // Analyzer diagnostics (and possibly other consumers) require a file path on items.
+
+            return node.FilePath;
```
davkean, line 112: I think this needed a comment that said "node.FilePath" can be null for items other than analyzers for me to understand the blurb above it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/DependencyModel.cs
drewnoakes: This is the base class ctor. Most subtypes reassign `Caption` with type-specific logic that's often too involved to do inline in the `base(...)` statement.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: Would this be clearer?

```diff
diff --git a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
index c0ddd2351..1c0d2551d 100644
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -102,12 +102,13 @@ Task OnUnconfiguredProjectUnloading(object? sender, EventArgs args)

         public override string? GetPath(IProjectTree node)
         {
+            // node.FilePath can be null. Some dependency types (e.g. packages) do not require a file path,
+            // while other dependency types (e.g. analyzers) do.
+            //
             // Returning null from a root graft causes CPS to use the "pseudo path" for the item, which has
             // form ">123" where the number is the item's identity. This is a short string (low memory overhead)
             // and allows fast lookup. So in general we want to return null here unless there is a compelling
             // requirement to use the path.
-            //
-            // Analyzer diagnostics (and possibly other consumers) require a file path on items.

             return node.FilePath;
         }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
drewnoakes: As a root graft, we're only called via `PhysicalProjectTreeProvider.FindByPath` and tracing back usages of that shows lots of callers that pass `tree.Root` however I cannot definitely prove we'd never receive a node other than the root. This was added during development to help me understand when that might not be the case, because if we're given a root that's within the dependencies tree itself, we would not (currently) return results.

I'll file a follow up PR that removes the `Assumes.True` line and improves the docs as below.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
davkean: Yes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
davkean: I think in most cases that would be true that you are only ever passed the root, however, I've definitely written code against providers that pass in more than the root. It's very odd in fact that CPS implementation doesn't give you the root that you provided to start with.

drewnoakes: Yes I wondered the same, whether CPS should filter which graft providers to call based upon the root. I think the assumption is that the implementation can interpret the path in some way that gives better complexity than O(N), as this implementation does.

Issue #3664, opened 2018-06-18
SDK style F# projects return broken FileCodeModel from ProjectItem


jcansdale:
Old style F# projects return `null` from `ProjectItem.FileCodeModel`.

I noticed that SDK style F# projects do return a `FileCodeModel`, but when I call `FileCodeModel.CodeElementFromPoint` it throws a `NullReferenceException` from:
```
Microsoft.VisualStudio.LanguageServices.Implementation.CodeModel.FileCodeModel.CodeElementFromPosition(Int32 position, vsCMElement scope)
at Microsoft.VisualStudio.LanguageServices.Implementation.CodeModel.FileCodeModel.CodeElementFromPoint(TextPoint point, vsCMElement scope)
at EnvDTE.FileCodeModel.CodeElementFromPoint(TextPoint Point, vsCMElement Scope)
```

I noticed the problem when working on TestDriven.Net.

I noticed that @StefanKert's [BuildVision](https://github.com/StefanKert/BuildVision) is also have problem with this project type:

```
---------------------------
BuildVision error
---------------------------
Unknown exception

Exception:
Object reference not set to an instance of an object.
StackTrace:
   at Microsoft.VisualStudio.LanguageServices.Implementation.CodeModel.AbstractCodeModelObject.get_Language()
   at EnvDTE.CodeModel.get_Language()
   at BuildVision.Helpers.ProjectExtensions.GetLanguageName(Project project)

Source:
EnvDTE

TargetSite:
System.String get_Language()

BaseException:
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.LanguageServices.Implementation.CodeModel.AbstractCodeModelObject.get_Language()
   at EnvDTE.CodeModel.get_Language()
   at BuildVision.Helpers.ProjectExtensions.GetLanguageName(Project project)
---------------------------
OK   
---------------------------
```

davkean:
We should return null here to be consistent with legacy. The real change should be in Roslyn - project system is simply delegating onto Roslyn - it should be aware that F# doesn't implement symbol model.

jcansdale:
@davkean Do you know if there's a built in alternative to `FileCodeModel` that I could use in F# projects? At the moment I'm using a horrible regex to scrape just enough symbol information when targeting an F# method. Unfortunately it's limited and there are plenty of times when it's not enough.

I basically need to find a `cref` token for the method/type the caret is inside. 🤞 there's a better solution!

[...]

PR #6293, opened 2020-06-16
Changed to only be turned on for C# and VB

ocallesp:
F# doesn't implement symbol model #3664 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6293)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ProjectContextCodeModelProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ProjectContextCodeModelProvider.cs
@@ -14,7 +14,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
     /// </summary>
     [Export(typeof(ICodeModelProvider))]
     [Export(typeof(IProjectCodeModelProvider))]
-    [AppliesTo(ProjectCapability.DotNetLanguageService)]
+    [AppliesTo(ProjectCapability.CSharpAndVisualBasic)]
     internal class ProjectContextCodeModelProvider : ICodeModelProvider, IProjectCodeModelProvider
     {
         private readonly IProjectThreadingService _threadingService;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/ProjectContextCodeModelProvider.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
     /// </summary>
     [Export(typeof(ICodeModelProvider))]
     [Export(typeof(IProjectCodeModelProvider))]
-    [AppliesTo(ProjectCapability.DotNetLanguageService)]
+    [AppliesTo(ProjectCapability.CSharpAndVisualBasic)]
```
davidwengier, line 17: This should be CSharp _or_ Visual Basic, right? I don't think a single project will ever have both

davidwengier: Also, perhaps `CSharpOrVisualBasicLanguageService` would be a good fit, since previously `LanguageService` was checked. Unless that is part of the problem?

ocallesp: @davkean 
your comment in here says :  turned on for C#  and  VB
https://github.com/dotnet/project-system/issues/3664#issuecomment-579005448

Do you wanted to say OR ?

davkean: This must include LanguageService, to avoid being called for shared projects. Shared Projects get renamed by the projects that consume them.

davkean: Sorry, I misread the comment/change - I thought it was rename.

@davidwengier is right it must be CSharpOrVisualBasicLanguageService, there's never a case that a project is both C# and VB at the same time.

PR #6305, opened 2020-06-24
Only call OpenDocumentAsync when openDocumentService is not null.

melytc:
This is the first step to ensure that when we call GetProxyAsync for CodeSpaces, we handle the case we get a null value. I'll create an issue to track how we want to handle when openDocumentService returns null.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6305)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
@@ -100,7 +100,8 @@ namespace Microsoft.VisualStudio.SolutionExplorer
                 {
                     Assumes.NotNull(openDocumentService);
 
-                    await openDocumentService.OpenDocumentAsync(node.NodeMoniker, cancellationToken: default);
+                    if (openDocumentService != null)
+                        await openDocumentService.OpenDocumentAsync(node.NodeMoniker, cancellationToken: default);
                 }
                 finally
                 {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
@@ -102,6 +102,7 @@ namespace Microsoft.VisualStudio.SolutionExplorer
                         await openDocumentService.OpenDocumentAsync(node.NodeMoniker, cancellationToken: default);
                     //else
                         // TODO: figure out what to tell the user if we can't get the service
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
```diff
@@ -100,7 +100,8 @@ private void OpenFile(WorkspaceVisualNodeBase node)
                 {
                     Assumes.NotNull(openDocumentService);
```
tmeschter, line 101: Remove this, as we're no longer assuming `openDocumentService` is not null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
```diff
@@ -100,7 +100,8 @@ private void OpenFile(WorkspaceVisualNodeBase node)
...
 
-                    await openDocumentService.OpenDocumentAsync(node.NodeMoniker, cancellationToken: default);
+                    if (openDocumentService != null)
+                        await openDocumentService.OpenDocumentAsync(node.NodeMoniker, cancellationToken: default);
```
tmeschter, line 104: Maybe add an empty `else` clause with a `// TODO: figure out what to tell the user if we can't get the service.`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
melytc: Removed, thanks to both! @drewnoakes

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs
melytc: Added it. Also created a tracking issue: https://github.com/dotnet/project-system/issues/6306

jjmew: Can you add the a link to the github issue in the comments? 

---
In reply to: [445717302](https://github.com/dotnet/project-system/pull/6305#discussion_r445717302) [](ancestors = 445717302)

melytc: Sure, will add it and merge the pr

PR #6309, opened 2020-06-26
Expose LogicalName for embedded resources's browse object

davkean:
This is one half of https://github.com/dotnet/project-system/issues/1058, and exposes it so that it can be consumed by the generation code.


tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/ItemRuleTests.cs
```diff
@@ -49,16 +49,22 @@ public void FileRulesShouldMatchNone(string ruleName, string fullPath)
...
                 return;
             }
             // No need to check None against None
-            if (ruleName.Equals("None", StringComparison.Ordinal))
+            if (ruleName.Equals("None.BrowseObject", StringComparison.Ordinal))
             {
                 return;
             }
 
+            string[]? excludedElements = Array.Empty<string>();
```
drewnoakes, line 62:
```suggestion
            string[] excludedElements = Array.Empty<string>();
```

davkean:
@dotnet/project-system Please take another look, this is slightly different from before; I've replaced exposing LogicalName with ManifestResourceName, which is the friendlier property (as you don't include .resources). We provide both properties, but only the latter is visible.

I've confirmed this PR and this PR: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/262111, causes us to correctly generate the correct code.

PR #6314, opened 2020-06-29
Update LaunchTargetPropertyPageEnumProvider

tmeschter:
Update `LaunchTargetPropertyPageEnumProvider` to look for pages that use the "CommandNamedBasedDebugger" template instead of the "Debugger" template; the former is what we actually use for managed projects.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6314)


tmeschter:
Basically the same change as #6308 but for the `LaunchTargetPropertyPageEnumProvider` rather than the `LaunchTargetPropertyPageValueProvider`.

PR #6315, opened 2020-06-29
Use short name for target nodes in dependencies tree

drewnoakes:
This change was motivated by two factors:

1. For .NET 5, the target node showed ".NETCoreApp 5.0", but ".NETCoreApp" is not a concept in .NET 5.
2. We would one day like to support arbitrary user-defined names for TargetFrameworks (NuGet/Home#5154)

### Before

image

### After

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6315)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFramework.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFramework.cs
@@ -16,10 +16,5 @@ namespace Microsoft.VisualStudio.ProjectSystem
         /// Gets the short name.
         /// </summary>
         string ShortName { get; }
-
-        /// <summary>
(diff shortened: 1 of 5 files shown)


tmeschter:
Couldn't we use the display names provided by the SDK? E.g., from https://github.com/dotnet/sdk/blob/e2faebad758a7d38b5965cda755a17e9e9881599/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedTargetFrameworks.props.

drewnoakes:
@tmeschter that does look interesting. Currently we need these values in unconfigured project scope, and they're available via evaluation which requires configured project scope. I'm not sure what's the best way to push that data upstream. There was some support for showing the raw string directly, but I've asked on Teams and if we decide to go this route I'll dig into what's required to make that happen.

PR #6318, opened 2020-06-30
Remove unused types

drewnoakes:
I cannot find any usage of this code.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6318)


tmeschter:
Many of these types/members are `internal`. Are the corresponding assemblies making their `internal` members visible to any other assemblies?

drewnoakes:
@tmeschter yes I went through the assembly IVTs and scanned //index for string matches against these type names, concluding they're safe to remove.

PR #6323, opened 2020-07-02
Restore case insensitive dependency ID comparison

drewnoakes:
#6282 changed how dependencies were identified to avoid using a composite string. However, that string was forced to lower-case, which meant that comparisons on dependencies' IDs (item specs) was case insensitive.

Since that change the following project would result in two package references being shown in the tree:

```xml

    <PropertyGroup>
        <TargetFramework>net5</TargetFramework>
    </PropertyGroup>

    <ItemGroup>
        <!-- NOTE invalid name casing, and invalid version number -->
        <PackageReference Include="Newtonsoft.JSON" Version="11.0.10" />
    </ItemGroup>

</Project>
```

This PR introduces a new `DependencyId` struct that provides proper hashing and equality.

### Before

image

### After

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6323)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/DependencyId.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/DependencyId.cs
@@ -1,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies.Models
+{
+    internal readonly struct DependencyId
+    {
+        public string ProviderId { get; }
+        public string ModelId { get; }
+
+        public DependencyId(string providerId, string modelId)
+        {
+            Requires.NotNull(providerId, nameof(providerId));
+            Requires.NotNull(modelId, nameof(modelId));
+
+            ProviderId = providerId;
+            ModelId = modelId;
+        }
+
+        public bool Equals(DependencyId other)
+        {
+            return StringComparers.DependencyProviderTypes.Equals(ProviderId, other.ProviderId) &&
+                   StringComparers.DependencyTreeIds.Equals(ModelId, other.ModelId);
+        }
+
+        public override bool Equals(object? obj) => obj is DependencyId other && Equals(other);
+
+        public override int GetHashCode()
+        {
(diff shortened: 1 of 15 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/DependencyId.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            return StringComparers.DependencyProviderTypes.Equals(ProviderId, other.ProviderId) &&
+                   StringComparers.DependencyTreeIds.Equals(ModelId, other.ModelId);
+        }
+
+        public override bool Equals(object? obj) => obj is DependencyId other && Equals(other);
+
+        public override int GetHashCode()
+        {
+            return unchecked(StringComparers.DependencyProviderTypes.GetHashCode(ProviderId) * 397) ^ StringComparers.DependencyTreeIds.GetHashCode(ModelId);
```
swesonga, line 28: >397 [](start = 95, length = 3)

Where does this particular constant come from?

drewnoakes: R# generated this code. It's a prime that's large enough to overflow the first value a bit and distribute bits with the next value.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Snapshot/TargetedDependenciesSnapshot.cs
```diff
@@ -110,7 +111,7 @@ void Remove(RemoveDependencyContext context, IDependencyModel dependencyModel)
                     }
                 }
 
-                dependencyById.Remove((dependencyModel.ProviderType, dependencyModel.Id));
+                dependencyById.Remove(dependencyModel.GetDependencyId());
```
swesonga, line 114: Would saving this value into a local remove an unnecessary allocation from this 2nd call?

drewnoakes: `DependencyId` is a value type so there's no extra allocation in this case. Any difference in the generated code will be minimal.

PR #6328, opened 2020-07-06
Fix potentially invalid expression in props file

drewnoakes:
After upgrading to 16.6.3 and with a `global.json` pinning to 3.1.4xx (which requires MSBuild 16.7), opening this repo caused all projects to fail to load.

The problem is that when `IsTestProject` is undefined, we see the error:

> Specified condition "$(IsTestProject)" evaluates to "" instead of a boolean

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6328)

### DIFF
--- a/build/import/HostAgnostic.props
+++ b/build/import/HostAgnostic.props
@@ -1,7 +1,7 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project>
   <PropertyGroup>
-    <BannedSymbolsOptOut Condition="$(IsTestProject)">true</BannedSymbolsOptOut>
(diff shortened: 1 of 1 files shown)


davkean:
Which means Directory.Build.props wasn't picked up?

drewnoakes:
I assume that's a consequence of not meeting the minimum MSBuild version somehow. The in-product experience wasn't great here though.

Also this condition checks `IsTestProject`, and further down the file another condition checks `IsUnitTestProject`.

davkean:
What was the experience? I'd expect the project fail to load?

drewnoakes:
That's what happened.

The info I needed to fix it was in the "Solution" output page, however it didn't explain about the minimum MSBuild version requirement until after I fixed the issue and attempted to reload the project (whereupon it still refused to load).

PR #6338, opened 2020-07-07
Fix lingering warning icon on SDK references

drewnoakes:
In `netstandard2.0`, the framework is shown as an SDK reference. This is delivered via a package which is not shown in the tree. The project system has a filter which applies the resolved state of the package to the SDK item.

#6330 made a change where the diagnostic level determined whether a reference's icon showed a warning or not, rather than the resolved state. That PR did not update this SDK/package filter to correctly clear/apply the diagnostic level on the SDK node. This meant that a resolved SDK would appear with a warning icon.

### Before

image

### After

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6338)


davkean:
Was this regressed in 16.7? Should it be fixed there?

drewnoakes:
Yes, if we accept this to master I'll cherry pick it into the corresponding open 16.7 PR that'll be put to the QB.

PR #6340, opened 2020-07-07
Delete reference to unused/unshipped browse object

davkean:
Note: That this and Page item types when chosen in a File -> New -> Console App cause the item to be hidden in the tree, which is unrelated to below. I'm still investigating this.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6340)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -181,10 +181,6 @@
       <Context>File;ProjectSubscriptionService</Context>
     </PropertyPageSchema>
 
-    <PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)Resource.BrowseObject.xaml">
-      <Context>BrowseObject</Context>
-    </PropertyPageSchema>
-
     <!-- Project related settings -->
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -181,10 +181,6 @@
       <Context>File;ProjectSubscriptionService</Context>
     </PropertyPageSchema>
 
-    <PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)Resource.BrowseObject.xaml">
```
swesonga, line 184: >Resource.BrowseObject.xaml [](start = 65, length = 26)

Out of curiosity, which item is hidden in the tree (mentioned in your PR description)? I'm not sure what this BrowseObject is or how it relates to your PR description.

davkean: Resource.BrowseObject describes the properties that will show in the Property window of the `<Resource  />` item. Currently, `<Resource/>` and `<Page/>` items don't show in the tree (can be caused by missing "BrowseObject") but its unrelated this exact change. These browse objects were deliberately removed in lieu of other code that was supposed to create them on the fly - this isn't working.

PR #6346, opened 2020-07-07
Fix Startup Project to show only debuggable projects.

### LINKED ISSUE #1308: Class Libraries show up in debug startup project list
This list is supposed to show projects that can be launched (so the user can quickly switch start up projects), neither of these projects can be launched:

image

ocallesp:
This pr fixes #1308 

The Startup project was showing NetCore Libraries which overwhelm users when using VS with large projects. This bug was not about whether Libraries should be in Startup or not, but to decide which projects are non-debuggable and only display debuggables.

- Added the project capability `DisableBuiltInDebuggerServices` to disable CPS debuggers services for managed projects.

- Implemented the missing functionality in `CanLaunch()` to decide which projects should appear in the Startup list.
  The logic to decide if a project is non-debuggable is in `LaunchProfilesDebugLaunchProvider`

-  Fixed unit tests

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6346)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IStartupProjectProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IStartupProjectProvider.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
+{
+    /// <summary>
+    /// Interface definition used by StartupProjectRegistrar to display only debuggable projects
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.ConfiguredProject, ProjectSystemContractProvider.Extension)]
+    internal interface IStartupProjectProvider
+    {
+        /// <summary>
+        /// Returns true if this is a project is debuggable and should appear in the startup list.
+        /// </summary>
+        Task<bool> IsProjectDebuggableAsync(DebugLaunchOptions launchOptions);
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IStartupProjectProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IStartupProjectProvider.cs
@@ -9,7 +9,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
     /// Interface definition used by StartupProjectRegistrar to display only debuggable projects
     /// </summary>
     [ProjectSystemContract(ProjectSystemContractScope.ConfiguredProject, ProjectSystemContractProvider.Extension)]
-    internal interface IStartupProjectProvider
+    public interface IStartupProjectProvider
     {
         /// <summary>
         /// Returns true if this is a project is debuggable and should appear in the startup list.
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
@@ -9,7 +9,6 @@ using System.Threading.Tasks;
 using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.ProjectSystem.Debug;
 using Microsoft.VisualStudio.Shell.Interop;
-using Microsoft.VisualStudio.Threading;
 using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
@@ -56,17 +55,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             try
             {
                 // Launch providers to enforce requirements for debuggable projects
-                await QueryDebugTargetsInternalAsync(launchOptions, true);
+                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: true);
             }
-            catch(Exception e)
+            catch (ProjectNotRunnableDirectlyException)
             {
-                if (string.Compare(e.Message, VSResources.ProjectNotRunnableDirectly, StringComparison.OrdinalIgnoreCase) == 0)
-                {
-                    return await TplExtensions.FalseTask;
-                }
+                return false;
+            }
+            catch (Exception)
+            {
             }
 
-            return await TplExtensions.TrueTask;
+            return true;
         }
 
         /// <summary>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
@@ -25,14 +25,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         private readonly IVsService<IVsDebugger4> _vsDebuggerService;
         private readonly ILaunchSettingsProvider _launchSettingsProvider;
         private IDebugProfileLaunchTargetsProvider? _lastLaunchProvider;
+        private readonly IProjectThreadingService _threadingService;
 
         [ImportingConstructor]
         public LaunchProfilesDebugLaunchProvider(
+            IProjectThreadingService threadingService,
             ConfiguredProject configuredProject, 
             ILaunchSettingsProvider launchSettingsProvider, 
             IVsService<SVsShellDebugger, IVsDebugger4> vsDebuggerService)
             : base(configuredProject)
         {
+            _threadingService = threadingService;
             _launchSettingsProvider = launchSettingsProvider;
             _vsDebuggerService = vsDebuggerService;
 
@@ -48,9 +51,22 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         /// <summary>
         /// Called by CPS to determine whether we can launch
         /// </summary>
-        public override Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions)
+        public override async Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions)
         {
-            return TplExtensions.TrueTask;
+            try
+            {
+                // Launch providers to enforce requirements for debuggable projects
+                await QueryDebugTargetsInternalAsync(launchOptions, true);
+            }
+            catch(Exception e)
+            {
+                if (string.Compare(e.Message, VSResources.ProjectNotRunnableDirectly, StringComparison.OrdinalIgnoreCase) == 0)
+                {
+                    return await TplExtensions.FalseTask;
+                }
+            }
+
+            return await TplExtensions.TrueTask;
         }
 
         /// <summary>
@@ -79,7 +95,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             }
 
             // Now find the DebugTargets provider for this profile
-            IDebugProfileLaunchTargetsProvider launchProvider = GetLaunchTargetsProvider(activeProfile) ??
+            IDebugProfileLaunchTargetsProvider launchProvider = await GetLaunchTargetsProviderAsync(activeProfile) ??
                 throw new Exception(string.Format(VSResources.DontKnowHowToRunProfile, activeProfile.Name));
 
             IReadOnlyList<IDebugLaunchSettings> launchSettings;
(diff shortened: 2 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -48,9 +51,22 @@ internal class LaunchProfilesDebugLaunchProvider : DebugLaunchProviderBase, IDep
...
+            try
+            {
+                // Launch providers to enforce requirements for debuggable projects
+                await QueryDebugTargetsInternalAsync(launchOptions, true);
+            }
+            catch(Exception e)
+            {
+                if (string.Compare(e.Message, VSResources.ProjectNotRunnableDirectly, StringComparison.OrdinalIgnoreCase) == 0)
+                {
+                    return await TplExtensions.FalseTask;
```
swesonga, line 65: >return await TplExtensions.FalseTask; [](start = 20, length = 37)

You should be able to simply return false; (or return true;) in this method

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -48,9 +51,22 @@ internal class LaunchProfilesDebugLaunchProvider : DebugLaunchProviderBase, IDep
...
         /// Called by CPS to determine whether we can launch
         /// </summary>
-        public override Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions)
+        public override async Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions)
         {
-            return TplExtensions.TrueTask;
+            try
+            {
+                // Launch providers to enforce requirements for debuggable projects
+                await QueryDebugTargetsInternalAsync(launchOptions, true);
```
swesonga, line 59: >true [](start = 68, length = 4)

Consider naming this parameter (like on line 77) to make it immediate what "true" is

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -48,9 +51,22 @@ internal class LaunchProfilesDebugLaunchProvider : DebugLaunchProviderBase, IDep
...
+            }
+            catch(Exception e)
+            {
+                if (string.Compare(e.Message, VSResources.ProjectNotRunnableDirectly, StringComparison.OrdinalIgnoreCase) == 0)
```
swesonga, line 63: >Compare [](start = 27, length = 7)

Should we have a StringComparer (defined in the StringComparers class) for error messages?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -48,9 +51,22 @@ internal class LaunchProfilesDebugLaunchProvider : DebugLaunchProviderBase, IDep
...
-        public override Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions)
+        public override async Task<bool> CanLaunchAsync(DebugLaunchOptions launchOptions)
         {
-            return TplExtensions.TrueTask;
+            try
+            {
+                // Launch providers to enforce requirements for debuggable projects
+                await QueryDebugTargetsInternalAsync(launchOptions, true);
+            }
+            catch(Exception e)
```
swesonga, line 61: >Exception [](start = 18, length = 9)

Just curious about this: should we throw a more specific exception type to avoid comparing exception messages? That way we don't return true if other unrelated exceptions are thrown (although that may be the desirable behavior in this case)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
ocallesp: I agree. I already made the change. Will be fixed in the next commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
ocallesp: I agree. Will be fixed in the next commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
ocallesp: Which StringComparer property is the one that applies to e.Message ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
ocallesp: While finding the fix I was checking all the exceptions. I agree. Changing this so I can get rid of the string comparer.

ocallesp: The exception I am interested is a general exception with that specific message. Maybe I should create their own exception and catch that new exception.

drewnoakes: A new type of exception would be clearer. Comparing the exception message is not commonly done, though I can imagine cases where it's the only option. If you control the throw too, then a new exception would be good.

If we keep the string comparison, then if the comparison fails the exception should be rethrown (i.e. `else throw;`).

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IProjectThreadingServiceFactory.ProjectThreadingService.cs
```diff
@@ -67,6 +67,11 @@ public IDisposable SuppressProjectExecutionContext()
             {
                 JoinableTaskFactory.Run(asyncAction);
             }
+
+            public Task SwitchToUIThread()
+            {
+                return Task.CompletedTask;
+            }
```
drewnoakes, line 74: Is this called? Don't consumers of the interface call an extension method for this?

ocallesp: I found this class was already implementing a mock. xUnit  cannot override extension methods. This approach was much easier and simpler.

drewnoakes: But `SwitchToUIThread` is not on `IProjectThreadingService`, it is an extension method. So the caller likely doesn't call this method. If you put a breakpoint in here, I wouldn't expect it to be hit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -630,4 +630,24 @@ private enum StringState
...
+        {
+        }
+
+        public ProjectNotRunnableDirectlyException(string message, Exception inner)
+            : base(message, inner)
+        {
+        }
+    }
+
+
```
ocallesp, line 652: umm....delete this extra line

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -630,4 +630,24 @@ private enum StringState
             NormalCharacter, EscapedCharacter, QuotedString, QuotedStringEscapedCharacter
         }
     }
+
+    [Serializable]
+    internal class ProjectNotRunnableDirectlyException : Exception
+    {
+        public ProjectNotRunnableDirectlyException()
+        {
+        }
```
drewnoakes, line 639:
```suggestion
        public ProjectNotRunnableDirectlyException()
            : this(VSResources.ProjectNotRunnableDirectly)
        {
        }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -428,7 +428,7 @@ private static bool UseCmdShellForConsoleLaunch(ILaunchProfile profile, DebugLau
                 // If we're launching for debug purposes, prevent someone F5'ing a class library
                 if (validateSettings && await IsClassLibraryAsync())
                 {
-                    throw new Exception(VSResources.ProjectNotRunnableDirectly);
+                    throw new ProjectNotRunnableDirectlyException(VSResources.ProjectNotRunnableDirectly);
```
drewnoakes, line 431:
```suggestion
                    throw new ProjectNotRunnableDirectlyException();
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -630,4 +630,24 @@ private enum StringState
             NormalCharacter, EscapedCharacter, QuotedString, QuotedStringEscapedCharacter
         }
     }
+
+    [Serializable]
```
drewnoakes, line 634: Does this need to be serializable? If so, it should have the serialization constructor too. I suspect it doesn't though.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IProjectThreadingServiceFactory.ProjectThreadingService.cs
ocallesp: Yeah. You are right. I deleted it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
ocallesp: done in next commit

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
ocallesp: done in next commit

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
ocallesp: I removed, it was added by codefix.

davkean:
I played around with this, this has regressed the error experience of class libraries. 

Given the following steps:

1. Create new Class Library (.NET Core)
2. F5

Expected:

```
---------------------------
Microsoft Visual Studio
---------------------------
Unable to start debugging. The startup project cannot be launched. Ensure that the correct project is set as the startup project. The startup project can be changed by selecting the 'Set as Startup Project' command from the right click menu in Solution Explorer.

Additionally make sure its debug settings are correctly configured in project properties.
---------------------------
OK   
---------------------------
```

Actual: Nothing
Debug commands are also disabled:
image

This is because CanLaunchAsync is used to determine whether debugging is enabled or not, we've already returned true in legacy and CPS so that the error experience is much better.

davkean:
I've gone through the API we have between the registrar and the debug launch providers and the API provided doesn't have a clean hand shake to determine if the current project is "really" debuggable, without regressing the F5 experience. I'm thinking we change the StartupProjectRegistrar to be tied to LaunchProfiles, and have a private handshake that provides that information and stop going thought the public CPS APIs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/IStartupProjectProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
+{
+    /// <summary>
+    /// Interface definition used by StartupProjectRegistrar to display only debuggable projects
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.ConfiguredProject, ProjectSystemContractProvider.Extension)]
+    internal interface IStartupProjectProvider
```
ocallesp, line 12: Adding new interface instead of adding more methods to IDebugLaunchProvider.

This is to avoid implementing unused method in CPS debuggers. It seems that the responsibility is more clear here perhaps not the names.

ocallesp:
@davkean  Please check.
A new [interface ](https://github.com/dotnet/project-system/pull/6346/commits/0fb07deca86ac5d35d840c1c0dac7a673c3c26a5#diff-7708528ea8443f324a7eb13f00cc406fR12)was added in the last commit. `IStartupProjectProvider `has the method `IsProjectDebuggableAsync`() can be used as a handshake with StartupProjectRegistrar.

Issue #5797, opened 2020-01-16
Debug page mnemonics


drewnoakes:
The 'Debug' property page doesn't define any keyboard accelerators/mnemonics for text boxes, drop downs or buttons. This makes the page less accessible than it should be.

These are usually visible when holding down the <kbd>Alt</kbd> key:

![image](https://user-images.githubusercontent.com/350947/72491013-e05c7300-386c-11ea-83b4-a3444b12c59a.png)

Notice how the main menu items have underlined letters, but none of the property page labels or buttons do.

Compare this with the 'Application' property page:

![image](https://user-images.githubusercontent.com/350947/72490944-b014d480-386c-11ea-8323-705db91ed220.png)

Adding these is a matter of prepending the desired character with `_` in the resource string. We should take care to not override other mnemonics such as those used in the main menu.

drewnoakes:
We should copy these mnemonics from legacy where that makes sense, to help people with muscle memory.

jjmew:
Moving to 16.8 as this specific item will not meet the QB bar for for 16.7

PR #6350, opened 2020-07-08
Add mnemonics to debug page.

melytc:
As noted in #5797, Debug page was missing the key accelerators, or mnemonics.

Visual Studio has already defined some menu mnemonics:
screenshot_vs_menubar

To avoid overlaps, I eliminated the letters _B, D, E, F, H, N, P, S, T, V, W, X_ from the set. I was able to assign a key accelerator to the majority of the properties, as shown below.

**Before:**
screenshot_before

**After:**
screenshot_after

You'll notice that buttons have no key accelerator assigned: I was left with 4 unique letters _(J, K, M, O)_ for the 6 different buttons. Any suggestion?

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6350)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.Designer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/PropertyPageResources.Designer.cs
@@ -61,7 +61,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Add.
+        ///   Looks up a localized string similar to _Add.
         /// </summary>
         public static string AddBtn {
             get {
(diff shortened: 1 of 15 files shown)


tmeschter:
What happens if two controls have the same accelerator? Does pressing that accelerator toggle between them? If so that _might_ be an acceptable solution, though you probably want to run it by the Accessibility champ.

melytc:
> What happens if two controls have the same accelerator? Does pressing that accelerator toggle between them? If so that _might_ be an acceptable solution, though you probably want to run it by the Accessibility champ.

Just found it does work, thanks! (But only within the properties page, not with the VS menu) 
I'll ask an a11y champ to confirm that experience is okay and add it to the PR.

drewnoakes:
Looks like "Executable" needs a mnemonic as well.

image

melytc:
> Looks like "Executable" needs a mnemonic as well.
> 
> image

Good catch, I'll add it in my next commit.

Small update: I talked with Dante, an accessibility champ and he gave me great guidelines; like it's better to have buttons with unique accelerators, and accelerators are scoped (so it's okay to reuse one from the VS top level menu). I'll make some changes to better follow these.

melytc:
Latest change:
image

PR #6352, opened 2020-07-08
Move URI to secret and remove if statement

dmonroym:
We used this if statement in order to avoid getting updates in our Teams only when PRs were created in the main repo (we didn't want them coming from forks).
Since we moved to using a different URI and it's now stored as a secret, secrets don't get copied to forks so we won't see this problem anymore.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6352)

### DIFF
--- a/.github/workflows/teams-notifications.yml
+++ b/.github/workflows/teams-notifications.yml
@@ -13,6 +13,5 @@ jobs:
     steps:
       - name: Notify
         uses: davidwengier/PostAdaptiveCard@v1.0.0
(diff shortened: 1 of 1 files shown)


davidwengier:
I don't see a notification for this PR :P

dmonroym:
@davidwengier weirdly enough the PR for project-system-tools doing this action didn't appear on teams but once it got merged issues and PRs still appear correctly

PR #6361, opened 2020-07-10
[DO NOT INSERT] Experimental Porting project-system-tools into project-system

ericdyoon:
continuation of: #6355
Using new directory structure, to avoid build warnings in visual studio editor 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6361)


project-system-tools/.github/workflows/teams-notifications.yml
```diff
@@ -0,0 +1,18 @@
+name: Post PRs and issues to Teams
```
davkean, line 1: We don't want two of these; we already have a notifcation that notifies teams. Just remove this.

PR #6365, opened 2020-07-14
[16.7] Protect against exception due to empty string

drewnoakes:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1151765

Applying this fix to `dev16.7.x`. It must merge forward to `master` (automation should do this).

---

We have received Watson data showing the value of `TargetingPackPath` or `OriginalItemSpec` can be empty, based on stack traces where these value are validated during the construction of `FrameworkReferenceIdentity`. This change prevents items from being attached in such a case.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6365)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
@@ -38,7 +38,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedColl
                 EnvDTE.Properties? props = projectItem?.Properties;
 
                 if (props?.Item("TargetingPackPath")?.Value is string path &&
-                    props?.Item("OriginalItemSpec")?.Value is string name)
+                    props?.Item("OriginalItemSpec")?.Value is string name &&
+                    !string.IsNullOrWhiteSpace(path) &&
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
```diff
@@ -38,7 +38,9 @@ public FrameworkReferenceAssemblyAttachedCollectionSourceProvider(IRelationProvi
                 EnvDTE.Properties? props = projectItem?.Properties;
 
                 if (props?.Item("TargetingPackPath")?.Value is string path &&
-                    props?.Item("OriginalItemSpec")?.Value is string name)
+                    props?.Item("OriginalItemSpec")?.Value is string name &&
+                    !string.IsNullOrWhiteSpace(path) &&
```
davkean, line 42: Nit: This can just be name.Length check. While this whitespace check will probably never be hit, what NTFS considers whitepace and what string.IsNullOrWhitespace considers whitespace are not the same.

drewnoakes: I used `IsNullOrWhiteSpace` here as there's another test deeper down in the call that does the same. We would use `IsNullOrEmpty` throughout if you think it's likely that a .NET white space string would be a valid path in NTFS.

PR #6383, opened 2020-07-16
Remove unneeded suppressions

davkean:
These suppressions no longer apply and are appearing as messages as unused.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6383)


drewnoakes:
I noticed these too but thought they were errors in the analyzers. Some of these still look like they're needed, especially the banned API suppressions. Have you verified each of these?

davkean:
The banned API was hidden by another (global) suppression which I’ve promoted to an inline pragma, it confused me too. I’ve verified the rest.

PR #6385, opened 2020-07-17
Fix priority ordering of attached tree children

drewnoakes:
Works around a limitation of Solution Explorer's tree view where the `IPrioritizedComparable` pattern is not used uniformly, and children can appear out of order.

---

Given:

```c#
    /// <summary>
    /// Specifies the order of attached items in the dependencies tree.
    /// </summary>
    /// <remarks>
    /// Used in conjunction with <see cref="IPrioritizedComparable"/>.
    /// </remarks>
    internal static class AttachedItemPriority
    {
        // Not all of these can be siblings.

        public const int Diagnostic               = 100;
        public const int Package                  = 200;
        public const int Project                  = 300;
        public const int CompileTimeAssemblyGroup = 400;
        public const int FrameworkAssemblyGroup   = 500;
        public const int ContentFilesGroup        = 600;
    }
```

Diagnostics should appear before any other kind of item.

## Before

Items appear alphabetically:

image

## After

Priority is respected:

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6385)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateRelationCollectionSource.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateRelationCollectionSource.cs
@@ -3,6 +3,8 @@
 using System;
 using System.Collections;
 using System.ComponentModel;
+using System.Windows.Data;
+using Microsoft.Internal.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.ProjectSystem.VS.Utilities;
 using Microsoft.VisualStudio.Shell;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/AggregateRelationCollectionSource.cs
```diff
@@ -73,8 +75,52 @@ public void SetCollection(IAggregateRelationCollection collection)
                 PropertyChanged?.Invoke(this, KnownEventArgs.HasItemsPropertyChanged);
             };
 
+            // Work around inconsistent use of IPrioritizedComparable in the tree view control
+            var view = (ListCollectionView)CollectionViewSource.GetDefaultView(_collection);
+            view.CustomSort = PrioritizedComparableComparer.Instance;
```
davkean, line 80: This feels like an implementation detail, can we not fix the source? Or can we file a bug to fix this?

drewnoakes: Agree. Filed https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1158136.

PR #6392, opened 2020-07-19
Annotate code

drewnoakes:
Reduces the number of un-annotated files from 14 to 5.

Includes annotations for `LaunchSettingsProvider`, which would have prevented bug [AB#1042733](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1042733) (fixed in #6389).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6392)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
@@ -14,8 +14,6 @@ using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
 using Task = System.Threading.Tasks.Task;
 
-#nullable disable
-
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
 {
     /// <summary>
@@ -65,7 +63,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             _remoteDebuggerAuthenticationService = remoteDebuggerAuthenticationService;
         }
 
-        private Task<ConfiguredProject> GetConfiguredProjectForDebugAsync()
+        private Task<ConfiguredProject?> GetConfiguredProjectForDebugAsync()
         {
             return _activeDebugFramework.GetConfiguredProjectForActiveFrameworkAsync();
         }
@@ -110,7 +108,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             // pause to end of window when CTRL+F5'ing a console application
             ConfigurationGeneral configuration = await _properties.GetConfigurationGeneralPropertiesAsync();
 
-            var actualOutputType = (IEnumValue)await configuration.OutputType.GetValueAsync();
+            var actualOutputType = (IEnumValue?)await configuration.OutputType.GetValueAsync();
+
+            Assumes.NotNull(actualOutputType);
 
             return StringComparers.PropertyLiteralValues.Equals(actualOutputType.Name, outputType);
         }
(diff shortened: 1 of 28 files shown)


drewnoakes:
Looks like the "Spanish" queue isn't picking up jobs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -289,7 +292,7 @@ private async Task<DebugLaunchSettings> GetConsoleTargetForProfile(ILaunchProfil
                     {
                         // Try to resolve against the current working directory (for compat) and failing that, the environment path.
                         string exeName = executable.EndsWith(".exe", StringComparisons.Paths) ? executable : executable + ".exe";
-                        string fullPath = _fileSystem.GetFullPath(exeName);
+                        string? fullPath = _fileSystem.GetFullPath(exeName);
```
davkean, line 295: This will throw on null, not return null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchProfile.cs
```diff
@@ -9,14 +9,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
...
+        string? ExecutablePath { get; }
+        string? CommandLineArgs { get; }
+        string? WorkingDirectory { get; }
         bool LaunchBrowser { get; }
-        string LaunchUrl { get; }
-        ImmutableDictionary<string, string> EnvironmentVariables { get; } // TODO-NULLABLE in some contexts this is null
-        ImmutableDictionary<string, object> OtherSettings { get; }        // TODO-NULLABLE in some contexts this is null
+        string? LaunchUrl { get; }
+        ImmutableDictionary<string, string>? EnvironmentVariables { get; }
+        ImmutableDictionary<string, object>? OtherSettings { get; }
```
davkean, line 20: Ugh, very unexpected.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -44,28 +42,28 @@ internal class LaunchSettingsProvider : OnceInitializedOnceDisposed, ILaunchSett
...
 
         [ImportingConstructor]
         public LaunchSettingsProvider(
             UnconfiguredProject project,
             IUnconfiguredProjectServices projectServices,
             IFileSystem fileSystem,
             IUnconfiguredProjectCommonServices commonProjectServices,
-            IActiveConfiguredProjectSubscriptionService projectSubscriptionService,
-            IActiveConfiguredValue<IAppDesignerFolderSpecialFileProvider> appDesignerSpecialFileProvider,
+            IActiveConfiguredProjectSubscriptionService? projectSubscriptionService,
```
davkean, line 65: This will never be null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
drewnoakes: The variable is reused on line 302 to store a nullable reference.

I'll introduce another nullable variable for that usage and keep this non-null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchProfile.cs
drewnoakes: `LaunchProfile(IWritableLaunchProfile)` explicitly sets these dictionaries to `null`. They're tested for `null` in several places. It may be that `null` can be replaced with `Empty`, but I haven't checked.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
drewnoakes: `null` is being passed in tests, and the implementation explicitly tests for `null` to avoid setting up dataflow subscriptions.

PR #6394, opened 2020-07-20
Register project properties editor with placeholder UI

drewnoakes:
Creates an editor factory that shows a placeholder UI when created.

This editor factory uses the well-known GUID for the project properties editor. The 'legacy' editor is given a new GUID. The intention is to introduce a feature flag or other mechanism to conditionally delegate to the legacy implementation.

editor-factory

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6394)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
@@ -25,6 +25,8 @@ namespace Microsoft.VisualStudio.Packaging
         {
             await base.InitializeAsync(cancellationToken, progress);
 
+            await JoinableTaskFactory.SwitchToMainThreadAsync();
+
             RegisterEditorFactory(new ProjectPropertiesEditorFactory());
         }
     }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
@@ -2,15 +2,30 @@
 
 using System;
 using System.Runtime.InteropServices;
+using System.Threading;
 using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.VisualStudio.Packaging
 {
     [Guid(PackageGuid)]
     [PackageRegistration(AllowsBackgroundLoading = true, RegisterUsing = RegistrationMethod.CodeBase, UseManagedResourcesOnly = true)]
+    [ProvideEditorFactory(
+        typeof(ProjectPropertiesEditorFactory),
+        nameResourceID: 1100,
+        deferUntilIntellisenseIsReady: false,
+        TrustLevel = __VSEDITORTRUSTLEVEL.ETL_AlwaysTrusted)]
     [ProvideMenuResource("Menus.ctmenu", 1)]
     internal sealed class ManagedProjectSystemClientPackage : AsyncPackage
     {
         public const string PackageGuid = "AE74FDFC-B9CE-4948-9E2F-F443B5BE8D37";
+
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            await base.InitializeAsync(cancellationToken, progress);
+
+            RegisterEditorFactory(new ProjectPropertiesEditorFactory());
+        }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
@@ -29,6 +29,7 @@ namespace Microsoft.VisualStudio.Packaging
         /// <remarks>
         /// Triggered via menu item "Debug | [Project Name] Debug Properties".
         /// </remarks>
+        // TODO if we need this value, try and source it from VSHPROPID_ProjectPropertiesDebugPageArg
         internal static readonly Guid DebugPageLogicalViewGuid = new Guid("0273C280-1882-4ED0-9308-52914672E3AA");
 
         private ProjectPropertiesWindowPaneData? _paneData;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
@@ -1,0 +1,168 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Windows;
+using System.Windows.Controls;
+using Microsoft.VisualStudio.ProjectSystem.VS;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using IServiceProvider = Microsoft.VisualStudio.OLE.Interop.IServiceProvider;
+
+namespace Microsoft.VisualStudio.Packaging
+{
+    [Guid(EditorFactoryGuidString)]
+    [ProvideView(LogicalView.Designer, "Design")]
+    internal sealed class ProjectPropertiesEditorFactory : IVsEditorFactory
+    {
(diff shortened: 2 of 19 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
```diff
@@ -2,15 +2,30 @@
...
     [ProvideMenuResource("Menus.ctmenu", 1)]
     internal sealed class ManagedProjectSystemClientPackage : AsyncPackage
     {
         public const string PackageGuid = "AE74FDFC-B9CE-4948-9E2F-F443B5BE8D37";
+
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            await base.InitializeAsync(cancellationToken, progress);
+
+            RegisterEditorFactory(new ProjectPropertiesEditorFactory());
```
davkean, line 28: This should switch to the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                return HResult.OK;
+            }
+
+            if (rguidLogicalView == VSConstants.LOGVIEWID.Designer_guid)
+            {
+                return HResult.OK;
+            }
+
+            // We may also be called with other GUIDs for specific pages in the UI. Rather than list them all here,
```
davkean, line 107: NOTE: This is actually used for Debug -> [Project] Debug Properties command.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private sealed class ProjectPropertiesWindowPaneData
+        {
+            public ProjectPropertiesWindowPaneData(IVsHierarchy vsHierarchy, System.IServiceProvider serviceProvider)
+            {
+                
+            }
+        }
+
+        private sealed class ProjectPropertiesWindowPane : WindowPane
```
davkean, line 126: There's async support for window panes, should we be using it? I don't know much about it.

drewnoakes: As far as I can tell there's async support for tool windows, but not editor/designer windows.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Packaging/ManagedProjectSystemClientPackage.cs
drewnoakes: Thanks, fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
drewnoakes: Thanks. When invoked that way we see GUID [`0273c280-1882-4ed0-9308-52914672e3aa`](https://github.com/dotnet/project-system/search?q=0273c280-1882-4ed0-9308-52914672e3aa&unscoped_q=0273c280-1882-4ed0-9308-52914672e3aa). I'll capture this info in the code.

We are considering pulling the debug launch settings out into its own UI in future. We'll have to handle this code path at that point I expect.

We will verify whether menu item/command is marked as Codespaces-compliant.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
```diff
@@ -104,8 +113,15 @@ public int MapLogicalView(ref Guid rguidLogicalView, out string? pbstrPhysicalVi
                 return HResult.OK;
             }
 
+            if (rguidLogicalView == DebugPageLogicalViewGuid)
```
davkean, line 116: This shouldn't be hard-coded, it should come from the hierarchy property; VSHPROPID_ProjectPropertiesDebugPageArg

drewnoakes: I don't think that will be possible here. Seems like there's no hierarchy to query at this point. Am I missing something?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
```diff
@@ -0,0 +1,184 @@
...
+        private const string LegacyEditorFactoryGuidString = "990036EB-F67A-4B8A-93D4-4663DB2A1033";
+
+        /// <summary>
+        /// Logical view identifier (passed to <see cref="MapLogicalView"/>) when the properties
+        /// pages are to be launched displaying the "debug" page.
+        /// </summary>
+        /// <remarks>
+        /// Triggered via menu item "Debug | [Project Name] Debug Properties".
+        /// </remarks>
+        private static readonly Guid DebugPageLogicalViewGuid = new Guid("0273C280-1882-4ED0-9308-52914672E3AA");
```
swesonga: >DebugPageLogicalViewGuid [](start = 37, length = 24)

Please remind me, what makes the debug page different?

jjmew: The debug pages has behavior that doesn't really belong to the properties pages. It has the ability to create profiles for debugging which get saved to launchsettings.json

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
tmeschter: The hierarchy was passed to `CreatedEditorInstance`--is that called before this method?

Also note that on the Codespace client the hierarchy will have limited functionality, and it is possible that `VSHPROPID_ProjectPropertiesDebugPageArg` won't work.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
tmeschter: The idea here is that you may want to open the project property pages directly to the Debug page. I can't recall if we ever supported opening directly to the 

Whether the debug settings should even live in the project property pages is a debatable point. :-) They are logically project-level settings; they just happen to be persisted in launchSettings.json rather than an MSBuild file.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
drewnoakes: `MapLogicalView` is called first. Only if it indicates that the view can be mapped will `CreateEditorInstance` be invoked.

The legacy property pages simply accept all logical GUIDs. The first version of this PR did the same. I'll convert this code to a comment explaining the situation.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/ProjectProperties/ProjectPropertiesEditorFactory.cs
drewnoakes: In this case, the debug page is treated separately because there's a feature that navigates directly to it.

> Whether the debug settings should even live in the project property pages is a debatable point. :-) They are logically project-level settings; they just happen to be persisted in launchSettings.json rather than an MSBuild file.

To expand upon this a bit, the debug page has a few differences to the other pages:

- There are logically multiple debug pages, one per launch profile, although as the user it's hard to see this (a dropdown controls this, which is unique to this page):
   image
   This adds a lot of complexity to the code today, as it doesn't really fit the model of the other pages.
- Configurations (debug/release, platform) do not apply to them.
- It makes sense for a user to want to not only navigate directly to the launch profile editor UI, but to a specific launch profile (which is currently not supported). For example via this toolbar button:
   image

- They are stored in a separate file (though that could change).
- They are usually not added to source control (though we do, and there is a use case for doing so).

Breaking them out to a new UI would allow us create an experience that better matches the nature of these launch profiles, and makes it easier to manage them. It'd also likely have a different extensibility model than the other property pages, because the options available depend upon the specific provider.

It's worth noting that we would be able to re-use much of the code between project properties and launch profiles.

drewnoakes:
Will close/open to hopefully kick the WIP bot.

PR #6395, opened 2020-07-20
Remove extra directory separator

drewnoakes:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6395)

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -9,7 +9,7 @@
   <Import Project="build\import\Workarounds.targets"/>
 
   <ItemGroup>
-    <AdditionalFiles Include="$(MSBuildThisFileDirectory)\src\Common\BannedSymbols.txt" Condition="'$(BannedSymbolsOptOut)' != 'true'" />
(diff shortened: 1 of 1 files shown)


Directory.Build.targets
```diff
@@ -9,7 +9,7 @@
   <Import Project="build\import\Workarounds.targets"/>
 
   <ItemGroup>
-    <AdditionalFiles Include="$(MSBuildThisFileDirectory)\src\Common\BannedSymbols.txt" Condition="'$(BannedSymbolsOptOut)' != 'true'" />
```
swesonga, line 12: >\ [](start = 57, length = 1)

Just to confirm: this doesn't affect the build itself, right?

drewnoakes: It just avoids a `\\` in the item spec. It's not a big deal, but potentially results in the string allocations due when getting the canonical path and normalizing separators.

PR #6396, opened 2020-07-20
Remove unused UnitTestHelper class

drewnoakes:
`UnitTestHelper ` was intended to bypass certain checks in unit tests, however the behaviour is not used anywhere any more.

If we need it back in future we can revert 37a92e09d49a47eb2115aa438a6866c8b07a151e.

Also fixes a typo and avoids redundant try/catch in non-`DEBUG` builds.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6396)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/UIThreadHelper.cs
```diff
@@ -11,26 +11,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
...
-                    throw;
-                }
             }
+            catch
+            {
+                System.Diagnostics.Debug.Fail("Call made on the Non-UI thread by " + memberName);
+                throw;
+            }
+#else
+            ThreadHelper.ThrowIfNotOnUIThread(memberName);
```
swesonga, line 33: >ThrowIfNotOnUIThread [](start = 25, length = 20)

Will RS0030 show up in release builds?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/UIThreadHelper.cs
```diff
@@ -11,26 +11,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
     internal static class UIThreadHelper
     {
         /// <summary>
-        /// Helper utility to ensure that we are on UI thread. Needs to be called 
-        /// in every method/propetrty needed UI thread for our protection (to avoid hangs 
+        /// Helper utility to ensure that we are on UI thread. Needs to be called
+        /// in every method/property with UI thread affinity (to avoid hangs
         /// which are hard to repro and investigate).
         /// </summary>
         public static void VerifyOnUIThread([CallerMemberName] string memberName = "")
```
ocallesp, line 18: If `VerifyOnUIThread` is only used in Debug configuration, but it still used in `GetUnconfiguredProject()` and `GetDTEProject()`, should `UIThreadHelper` be renamed to something more specific like DebugOnlyUIThreadHelper ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/UIThreadHelper.cs
drewnoakes: Yes, thank you. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/UIThreadHelper.cs
drewnoakes: > If VerifyOnUIThread is only used in Debug configuration

It's used in all configurations. It just doesn't assert in release builds. The `#if` here is just a minor perf thing that avoids the overhead of having `try`/`catch` in the method for release builds.

PR #6400, opened 2020-07-20
Don't launch the debug target

tmeschter:
PR #6346 added checks to see if a project was really "debuggable" with the goal of eliminating undebuggable class libraries from the startup projects list. However, it currently does this by actually trying to run the debug target. So when we check if a project is debuggable we may end up launching a console app, or IIS Express, or something else.

Clearly this is undesirable, and is causing problems for many people while dogfooding as well as test failures.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6400)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
@@ -63,8 +63,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         {
             try
             {
-                // Launch providers to enforce requirements for debuggable projects
-                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: true);
+                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: false);
             }
             catch (ProjectNotRunnableDirectlyException)
             {
@@ -90,6 +89,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         /// This is called on F5 to return the list of debug targets. What is returned depends on the debug provider extensions
         /// which understands how to launch the currently active profile type. 
         /// </summary>
+        /// <param name="launchOptions">
+        /// Options to pass to the debugger.
+        /// </param>
+        /// <param name="fromDebugLaunch">
+        /// <c>true</c> to query the information by actually running the launch target; <c>false</c>
+        /// otherwise.
+        /// </param>
         private async Task<IReadOnlyList<IDebugLaunchSettings>> QueryDebugTargetsInternalAsync(DebugLaunchOptions launchOptions, bool fromDebugLaunch)
         {
             // Get the active debug profile (timeout of 5s, though in reality is should never take this long as even in error conditions


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -64,7 +64,7 @@ public async Task<bool> IsProjectDebuggableAsync(DebugLaunchOptions launchOption
             try
             {
                 // Launch providers to enforce requirements for debuggable projects
```
swesonga, line 66: >Launch providers to enforce requirements for debuggable projects [](start = 19, length = 64)

This comment is no longer true then?

tmeschter: @ocallesp Will need to comment on that. I wasn't sure if launching the debug target was intended or not.

ocallesp: Yes. It was intended.  The idea is to hit ProjectNotRunnableDirectlyException in one of the debug providers.

ocallesp: Passing argument False will display class libraries in the startup. I need to reopen the issue.

BillHiebert: @ocallesp You can use it this way. Passing true means that the project is being told to debug the app. For webs this means starting web servers, launching browsers etc. For Functions projects, it downloads runtimes, starts the Azure storage emulator and runs the project.

ocallesp: `IsProjectDebuggableAsync`() is only used in `StartupProjectRegistrar.cs`. My assumption was that `QueryDebugTargetsInternalAsync`() was only about to query, not execute.

tmeschter: Comment removed.

BillHiebert: For simple projects like console applications it amounts to the same thing. But for more complication apps it doesn't. From a VS perspective the flag setting represents whether [IVsDebuggableProjectCfg](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsdebuggableprojectcfg?view=visualstudiosdk-2019) QueryDebugLaunch() was called, DebugLaunch() was called.

davkean:
Does this regress what https://github.com/dotnet/project-system/pull/6346 tried to fix? Can we just revert that change and have another go at fixing this?

tmeschter:
@davkean Based on @ocallesp's comments yes, I believe it will regress what #6346 tried to fix.

tmeschter:
Closing in favor of #6402.

PR #6401, opened 2020-07-20
Fix debug target query in non-launch scenario

drewnoakes:
Fixes [AB#1158301](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1158301).

Commit 5f85e2c5e5a9c8a3d00f64665b717a296092f803 introduced a bug where the querying of debug targets by `IsProjectDebuggableAsync` was passing a flag that indicated the query occurred as part of a debug launch operation.

This flag caused WebTools to launch a debugger on many actions which are unrelated to debugging:

d0dec3c0-41b1-4557-b6f2-acb176722baa

Thanks @BillHiebert for finding the root cause here.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6401)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
@@ -64,7 +64,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             try
             {
                 // Launch providers to enforce requirements for debuggable projects
-                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: true);
+                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: false);
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -64,7 +64,7 @@ public async Task<bool> IsProjectDebuggableAsync(DebugLaunchOptions launchOption
             try
             {
                 // Launch providers to enforce requirements for debuggable projects
-                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: true);
+                await QueryDebugTargetsInternalAsync(launchOptions, fromDebugLaunch: false);
```
swesonga, line 67: >QueryDebugTargetsInternalAsync [](start = 22, length = 30)

This fix looks like a dup of #6400

PR #6413, opened 2020-07-22
Fire launch completion when there are no debug targets.

auott:
If there are no debug targets, manually invoke the completion routine
so that the after launch hooks are called.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6413)


dmonroym:
Closing, this is already in master via #6419

PR #6415, opened 2020-07-22
Set initial AppliedValue to avoid hangs waiting on this

ocallesp:
This pr adds a default initial value similar to [InitializeInnerCoreAsync](http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.VS&leftSymbol=faheqpyvnq9v&rightProject=Microsoft.VisualStudio.ProjectSystem.VS.Implementation&file=Package%5CReferencesHostBridge.cs&line=362)() in `ReferencesHostBridge`

Avoid [waiting](http://index/?rightProject=Microsoft.VisualStudio.ProjectSystem.VS&file=Package%5cProjectHostBridge.cs&line=271) in   `ReferencesHostBridge.InitializeInnerCoreAsync()` 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6415)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsBuildManagerBridge.cs
@@ -106,6 +106,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
         /// </summary>
         protected override Task InitializeInnerCoreAsync(CancellationToken cancellationToken)
         {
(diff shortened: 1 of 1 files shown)


davkean:
@ocallesp What prompted this? This will result in this public API to return the wrong value if it is queried before the results arrive: http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Managed.VS&leftSymbol=vw2wviyotcal&file=ProjectSystem%5CVS%5CAutomation%5CVsBuildManager.cs&line=73.

ocallesp:
> @ocallesp Can you point to the hang? This will result in this public API returning the wrong value the first time it is queried: http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Managed.VS&leftSymbol=vw2wviyotcal&file=ProjectSystem%5CVS%5CAutomation%5CVsBuildManager.cs&line=73.
> 
> ie It's deliberate that this is blocking on the value, because we're returning true for this value: http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.VS&leftSymbol=kxrnhr7g03l3&file=Package%5CProjectHostBridge.cs&line=127.

Yes. The hang happens in DesignTimeMoniker. In the initialization it waits in Task.WhenAny( tasks )

It seems that there was another issue reported in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1085702/
image

PR #6424, opened 2020-07-23
Fix root graft tree provider GetPath logic

### LINKED ISSUE #6420: Bugs with imports/dependency root graft providers and GetPath
We have two root graft tree providers, one for the "Dependencies" tree, another for the "Imports" tree.

Each has a subtle bug around `GetPath`, which manifests via CPS's `TryGetPathFromRootGrafts`.

- The dependencies tree should return `null` for any items it does not contain (test ancestry)
- The imports tree should provide an implementation that returns `FilePath` for its items

The bug can be reproduced by disabling the dependencies tree via capabilities, showing the imports tree and invoking the "Copy Path" command. You'll get the psuedo-path of form `>123` (where 123 is the item's Id).

drewnoakes:
Fixes #6420.

Root graft project tree providers should only return a value for `GetPath` for tree nodes that they own. Previously the assumption was that this method was only passed its own descendants, but this is not the case.

This fixes a problem where, when the Dependencies tree was not enabled, the "Copy Full Path" command on import tree items would result in the pseudo-path (">123") being copied rather than the actual file path.

---

In a nanoFramework project:

## Before

image

Note item appears expanded (by arrow direction) but no child is visible.

"Copy Full Path" produces: `D:\repos\NFApp1\NFApp1\>34`, where `>34` is the item ID

## After

image

"Copy Full Path" produces: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview-master\MSBuild\nanoFramework\v1.0\NFProjectSystem.Default.props`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6424)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
@@ -103,15 +103,34 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
 
         public override string? GetPath(IProjectTree node)
         {
-            // node.FilePath can be null. Some dependency types (e.g. packages) do not require a file path,
-            // while other dependency types (e.g. analyzers) do.
-            //
-            // Returning null from a root graft causes CPS to use the "pseudo path" for the item, which has
-            // form ">123" where the number is the item's identity. This is a short string (low memory overhead)
-            // and allows fast lookup. So in general we want to return null here unless there is a compelling
-            // requirement to use the path.
+            // If the node's FilePath is null, we are going to return null regardless of whether
+            // this node belongs to the dependencies tree or not, so avoid extra work by retuning
+            // immediately here.
+            if (node.FilePath == null)
+            {
+                return null;
+            }
 
-            return node.FilePath;
+            // Walk up from node through all its ancestors.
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependenciesProjectTreeProvider.cs
```diff
@@ -103,15 +103,34 @@ Task OnUnconfiguredProjectUnloading(object? sender, EventArgs args)
...
+            // If the node's FilePath is null, we are going to return null regardless of whether
+            // this node belongs to the dependencies tree or not, so avoid extra work by retuning
+            // immediately here.
+            if (node.FilePath == null)
+            {
+                return null;
+            }
+
+            // Walk up from node through all its ancestors.
+            for (IProjectTree? step = node; step != null; step = step.Parent)
```
davkean, line 115: You can just call node.GetAncestry() and walk that.

drewnoakes: `GetAncestry` excludes the node itself.

davkean: Dependency node would never have a filepath, right?

PR #6438, opened 2020-07-24
First pass at embedding rules

davkean:
This is the first pass at embedding non-localized rules, which will let us use rules even in cases where our design-time targets are not imported. This is important for enabling PackageReference support for C++ and other project types.

This is first pass and picked some of the easy rules that lived in a single context. Localization support needs CPS changes.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6438)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/ExportedRuleTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            using Stream stream = assembly.GetManifestResourceStream(attribute.XamlResourceStreamName);
+
+            Assert.True(stream != null, $"The rule '{attribute.XamlResourceStreamName}' indicated by '{name}' does not exist in assembly '{attribute.XamlResourceAssemblyName}'.");
+        }
+
+        public static IEnumerable<object[]> GetAllExportedMembers()
+        {
+            foreach (var member in typeof(RuleExporter).GetMembers())
+            {
+                if (member.DeclaringType == typeof(RuleExporter))
```
drewnoakes, line 48: Could pass `BindingFlags.DeclaredOnly` to `GetMembers` instead of this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/RuleExporter.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// <summary>
+    ///     Responsible for exporting our embedded rules so that CPS can pick them.
+    /// </summary>
+    internal static class RuleExporter
+    {
+        /// <summary>
+        ///     Represents the evaluation properties that is used for AppDesigner folder services.
+        /// </summary>
+        [ExportRule(nameof(AppDesigner), PropertyPageContexts.ProjectSubscriptionService)]
+        [AppliesTo(ProjectCapability.AppDesigner)]
```
drewnoakes, line 16: What's the implication of `AppliesTo` here, compared to what we had before? CPS won't make these rules available unless the capability expression matches the project's capabilities, I assume. Is there a perf buff, or is this more about correctness?

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Rules/ExportedRuleTests.cs
davkean: And `BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public`, this ended up being much simpiler.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/RuleExporter.cs
davkean: Performance; our assembly & satellite assemblies will be loaded in scenarios it shouldn't.

davkean:
@drewnoakes  Take another look at this, this has changed a bit.

build/import/VisualStudio.XamlRules.targets
```diff
@@ -5,13 +5,30 @@
     <XamlPropertyRule>
       <Namespace>Microsoft.VisualStudio.ProjectSystem</Namespace>
       <RuleInjectionClassName>$(XamlPropertyRuleInjectionClassName)</RuleInjectionClassName>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </XamlPropertyRule>
+    <XamlPropertyRuleNoCodeBehind>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
davkean, line 11: These are needed so that I can refer to the "CopyToOutputDirectory" metadata item without qualification (ie no need for DesignTimeTargetsFile.CopyToOutputDirectory);

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ExportRuleAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Rules
+{
+    /// <summary>
+    ///     Exports a XAML-based embedded rule.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Field)]
+    internal class ExportRuleAttribute : ExportPropertyXamlRuleDefinitionAttribute
```
drewnoakes, line 12: It used to be the case that sealed attributes were considerably faster to reflect upon ([CA1813](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca1813)). Not sure if that's still the case.

davkean: In this case, its only affecting the tests. No code other than tests are reading these attributes via Reflection. MEF uses a metadata reader.

PR #6442, opened 2020-07-27
Fix hang in GetDesignTimeMoniker

ocallesp:
Fixes AzDO 1085702
DesignTimeOutputMonikers calls initializes DesignTimeInputsBuildManagerBridge
that waits (hangs) for AppliedValue to be generated/created.

AppliedValue was not generated when bad input data block arrives and
in here we are producing an empty value that will get assigned
to AppliedValue.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6442)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
@@ -9,6 +9,8 @@ using System.Linq;
 using System.Threading.Tasks.Dataflow;
 using Microsoft.VisualStudio.ProjectSystem.Utilities;
 
+using EmptyCollections = Microsoft.VisualStudio.ProjectSystem.Empty;
+
 namespace Microsoft.VisualStudio.ProjectSystem.VS.TempPE
 {
     [Export(typeof(IDesignTimeInputsChangeTracker))]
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -146,19 +148,32 @@ internal void ProcessFileChangeNotification(IProjectVersionedValue<string[]> arg
...
+            {
+                // Lets return a default value to avoid hangs waiting on this.
+                _currentState = new DesignTimeInputsDelta(EmptyCollections.OrdinalStringSet,
+                    EmptyCollections.OrdinalStringSet, Enumerable.Empty<DesignTimeInputFileChange>(), string.Empty);
+            }
+
+            PublishDelta(_currentState);
+        }
+
+        private static DesignTimeInputsDelta GenerateOutputData(
```
drewnoakes, line 165:
```suggestion
        private static DesignTimeInputsDelta? GenerateOutputData(
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsChangeTracker.cs
```diff
@@ -146,19 +148,32 @@ internal void ProcessFileChangeNotification(IProjectVersionedValue<string[]> arg
...
         internal void ProcessDataflowChanges(IProjectVersionedValue<ValueTuple<DesignTimeInputs, IProjectSubscriptionUpdate>> input)
+        {
+            _currentState = GenerateOutputData(_currentState, input);
+
+            // Processed bad input data
+            if (_currentState is null)
+            {
+                // Lets return a default value to avoid hangs waiting on this.
+                _currentState = new DesignTimeInputsDelta(EmptyCollections.OrdinalStringSet,
+                    EmptyCollections.OrdinalStringSet, Enumerable.Empty<DesignTimeInputFileChange>(), string.Empty);
```
drewnoakes, line 159: Are we sure that downstream consumers of this data are able to handle empty state?

ocallesp: Not sure If BuildDesignTimeOutput() should return "" when AppliedValue.Value.TempPEOutputPath. 

`BuildDesignTimeOutput()` -> `GetDesignTimeInputXmlAsync()` -> `GetDesignTimeInputXmlAsync()` -> [GetOutputFileNameFromRelativePath](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/TempPE/DesignTimeInputsCompiler.cs#L277)( "" ) => return Path.Combine(**tempPEOutputPath**, relativePath.Replace('\\', '.') + ".dll");

PR #6447, opened 2020-07-29
adding TargetPlatformMinVersion to configuration

zkat:
See https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#lighting-up-on-later-os-versions

I believe this is all that's needed to have the project system thread this value through? I think this covers everything we need, now. /cc @nkolev92 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6447)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
@@ -138,6 +138,14 @@
                   Visible="False" />
 
 
+  <StringProperty Name="TargetPlatformMoniker"
+                  ReadOnly="True"
+                  Visible="False" />
+
+  <StringProperty Name="TargetPlatformMinVersion"
+                  ReadOnly="True"
+                  Visible="False" />
+
   <StringProperty Name="TreatWarningsAsErrors"
                   ReadOnly="True"
                   Visible="False" />


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -302,6 +302,9 @@
   <StringProperty Name="TargetPlatformVersion"
                   Visible="false" />
 
+  <StringProperty Name="TargetPlatformMinVersion"
```
davkean, line 305: This is unneeded. 

We might want to consider exposing this property inside VS to other components, but this is the correct place to do it. I'll hold off exposing until there's an ask.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
```diff
@@ -169,6 +169,26 @@
     </StringProperty.DataSource>
   </StringProperty>
 
+  <StringProperty Name="TargetPlatformIdentifier"
+                  Visible="False">
+    <StringProperty.DataSource>
```
davkean, line 174: You can remove the DataSource, and just mark the property as `ReadOnly="true"`. I know the other properties did this, but they are wrong and I'll fix them.

We don't need the data source; it was copied from ConfigurationGeneral and was only used when writing to the property which we don't ever do from this rule.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
```diff
@@ -169,6 +169,26 @@
...
+      <DataSource HasConfigurationCondition="False"
+                  PersistedName="TargetPlatformIdentifier"
+                  Persistence="ProjectFile"
+                  SourceOfDefaultValue="AfterContext" />
+    </StringProperty.DataSource>
+  </StringProperty>
+
+  <StringProperty Name="TargetPlatformVersion"
+                  Visible="False">
+    <StringProperty.DataSource>
```
davkean, line 184: Ditto as above.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
davkean: I fixed the data sources here: https://github.com/dotnet/project-system/pull/6450, so just pull that change and copy the other properties.

zkat:
Fix't :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
```diff
@@ -138,6 +138,14 @@
                   Visible="False" />
```
davkean, line 140: @dsplaisted @nkolev92 Should this now be `TargetPlatformMoniker` given our discussions internally?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
```diff
@@ -138,6 +138,14 @@
...
 
 
+  <StringProperty Name="TargetPlatformIdentifier"
+                  ReadOnly="True"
+                  Visible="False" />
+
+  <StringProperty Name="TargetPlatformVersion"
+                  ReadOnly="True"
+                  Visible="False" />
+
```
davkean, line 148: Are you missing TargetPlatformMinVersion?

zkat: It's there now :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/NuGetRestore.xaml
zkat: Added!

nkolev92:
@davkean 

Can we get this merged soon?

nkolev92:
Actually if we want a fully compatible experience for P2, we might want the project-system changes in as well...trick is P2 is  reading TargetPlatformIdentifier,TargetPlatformVersion, while P3 will read TargetPlatformMoniker. 

PR for reading TargetPlatformMoniker is here: https://github.com/NuGet/NuGet.Client/pull/3578

davkean:
@nkolev92 Do you want TargetPlatformIdentifier/TargetPlatformVersion in this?

davkean:
I've merged this in anyway, if want a P2 compatible solution then submit another PR that we can file a bug to revert.

nkolev92:
@davkean 

I think we want a P2 solution too.
Created a PR: https://github.com/dotnet/project-system/pull/6495

I will loop you in the e-mail thread.

PR #6448, opened 2020-07-29
Merge Dev16.7.x Accessibility fix into master

jmarolf:
These are accessibility fixes that we want to be in 16.8 when it ships but it is not critical that they be in 16.8P1 (and go through QB approval). 

CC: @dmonroym 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6448)


davkean:
@dmonroym Why didn't auto merge pick this up?

dmonroym:
> @dmonroym Daniel Monroy FTE Why didn't auto merge pick this up?

This was inserted to 16.7 via https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/263151 just yesterday so the 16.7.x to 16.8.x merger had been blocked. Since we don't want this in 16.8P1 we're going directly to master.

Here's @jmarolf's justification
"we are ok having this fix be in P2 but not P1 for 16.8 as its an accessibility fix and not something we need to have till 16.8 releases"

PR #6374, opened 2020-07-15
Upgrade Code Analysis analyzers and fix results

davkean:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6374)

PR #6464, opened 2020-08-04
Suppress build warnings

dmonroym:
I believe this was introduced when updating the analyzers in #6374 

This also removes the suppression for warning CS8653 as it appears to no longer be necessary

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6464)


dmonroym:
@drewnoakes does this mean that https://github.com/dotnet/roslyn/issues/37069 is no longer an issue?

drewnoakes:
@dmonroym I updated that issue to use the newer diagnostic ID.

davkean:
As FYI, this is a compiler error; so it will be either because:

1) VS is updated to a newer version with a newer compiler and so you see different analysis inside VS

-or-

2) We've updated Microsoft.Net.Compilers.Toolset and it produces more errors.

It's not because of the analyzers.

It's almost always 1) because our CI will catch 2) and treat it as an error.

dmonroym:
Merging as a temporary workaround for https://github.com/dotnet/roslyn/issues/46555

PR #6465, opened 2020-08-05
Avoid overflowing build number

davkean:
This avoids overflowing the build number for the next 10 years.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6465)

### DIFF
--- a/build/import/Versions.props
+++ b/build/import/Versions.props
@@ -1,7 +1,7 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ProjectSystemVersion>16.8.0</ProjectSystemVersion>
+    <ProjectSystemVersion>16.8.1</ProjectSystemVersion>
     <VersionBase>$(ProjectSystemVersion)</VersionBase>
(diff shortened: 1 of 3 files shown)


dmonroym:
Anyone else using RepoToolset should also be experiencing the same breaks right? Perhaps it’s worth filling an issue to point them to a potential workaround

davkean:
We're the only ones using it. We just need to fork it and bring into our repo.

build/import/Versions.props
```diff
@@ -1,7 +1,7 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ProjectSystemVersion>16.8.0</ProjectSystemVersion>
+    <ProjectSystemVersion>16.8.1</ProjectSystemVersion>
```
drewnoakes, line 4: Why bump this?

davkean: Without this we regress in version.

davidwengier:
> Anyone else using RepoToolset should also be experiencing the same breaks right?

Strictly speaking they would also need their Azure DevOps Pipeline to have a build number of 20208001 or above.

davkean:
Which is what will happen (if there were other consumers) given that its based on a BUILD_NUMBER that comes from MicroBuild.

PR #6469, opened 2020-08-06
Workaround WPF temp project bug

davkean:
This fixes the reason these warnings were seen at command-line build time: https://github.com/dotnet/project-system/pull/6464.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6469)


davkean:
When we update compiler versions will hit the same thing.

PR #6474, opened 2020-08-06
Support environment variables in debugger property pages

### LINKED ISSUE #6268: Add "Environment variables" to debug property pages
ExecutableDebugPropertyPage.xaml and ProjectDebugPropertyPage.xaml are missing the "Environment variables" property.

This was split off from #6210; supporting "Environment variables" is a bit more difficult than the other properties because we need to figure out a string representation of a set of key-value pairs to communicate between the client and server.

tmeschter:
Fixes #6268.

Add basic support for the "Environment variables" property as used by various debugger property pages. This changes adds 
 the `ActiveLaunchProfileEnvironmentVariableValueProvider` type to serialize the `ILaunchProfile.EnvironmentVariables` dictionary to a string when the value is retrieved, and deserialize it from a string back to a dictionary when it is set.

In serializing to a string we need some way to separate the environment variable name from its value, and also some way to separate one environment variable from the next. Here we use an equal sign `=` for the former, and a comma `,` for the latter. These were chosen as they were unlikely to appear in variable names or values and wouldn't need to be escaped. When we do need to escape them, we do so using a forward slash `/`.

The "Executable" and "Project" debug pages have also been updated to expose the serialized string as-is. This is functional but not a great user experience. In the future we will need to put a better editor in front of this data, but the specifics of that depend on the new property pages UI.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6474)

### DIFF
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptedProjectProperties/ActiveLaunchProfileValueProvidersTests.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptedProjectProperties/ActiveLaunchProfileValueProvidersTests.cs
@@ -581,6 +581,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         }
 
         [Fact]
+        public async Task EnvironmentVariables_OnGetPropertyValueAsync_VariablesAreSorted()
+        {
+            var activeProfileEnvironmentVariables = new Dictionary<string, string>
+            {
+                { "var3", "value3" },
+                { "var2", "value2" },
+                { "var1", "value1" }
+            };
+
+            var settingsProvider = SetupLaunchSettingsProvider(activeProfileName: "One", activeProfileEnvironmentVariables: activeProfileEnvironmentVariables);
+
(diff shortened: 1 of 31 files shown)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptedProjectProperties/ActiveLaunchProfileValueProvidersTests.cs
```diff
@@ -559,6 +560,77 @@ public async Task SqlDebugEnabled_OnSetPropertyValueAsync_SetsDirectoryInActiveP
             Assert.True(activeProfileSqlDebugEnabled);
         }
 
+        [Fact]
+        public async Task EnvironmentVariables_OnGetPropertyValueAsync_GetsEscapedValuesFromActiveProfile()
+        {
+            var activeProfileEnvironmentVariables = new Dictionary<string, string>
+            {
+                { "Alpha", "Comma: , Equals: =" },
```
dmonroym, line 568: Also worth adding "Slash: /" ?

PR #6476, opened 2020-08-07
Upgrade to the latest version of VS SDK packages

davkean:
This updates all our package versions to latest available in master, and reacts to associated fallout:

- Needed to move interop embed down to HostAgnostic layer as there are dependencies on interop assemblies at that level.
- Needed to add a reference to Microsoft.IO.Redist which CPS runtime now has a dependency.
- Need to fix nullable warnings due to update references
- Need to fix VB's reference to namespace "IO" as it conflicts with "Microsoft.IO"

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6476)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/IProjectConfigurationDimensionsProvider5.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/IProjectConfigurationDimensionsProvider5.cs
@@ -6,7 +6,7 @@ using Microsoft.Build.Construction;
 
 namespace Microsoft.VisualStudio.ProjectSystem
 {
-    internal interface IProjectConfigurationDimensionsProvider4 : IProjectConfigurationDimensionsProvider3
+    internal interface IProjectConfigurationDimensionsProvider5 : IProjectConfigurationDimensionsProvider3
     {
         IEnumerable<string> GetBestGuessDimensionNames(ImmutableArray<ProjectPropertyElement> properties);
     }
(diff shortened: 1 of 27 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/IProjectConfigurationDimensionsProvider5.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.VisualStudio.ProjectSystem
 {
-    internal interface IProjectConfigurationDimensionsProvider4 : IProjectConfigurationDimensionsProvider3
+    internal interface IProjectConfigurationDimensionsProvider5 : IProjectConfigurationDimensionsProvider3
```
tmeschter, line 9: Why the rename?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -439,7 +439,7 @@ private bool CheckMarkers(Log log, in TimestampCache timestampCache, State state
             // here if the project actually produced a marker and we only check it against references that
             // actually produced a marker.
 
-            if (string.IsNullOrWhiteSpace(state.CopyUpToDateMarkerItem) || !state.CopyReferenceInputs.Any())
+            if (Strings.IsNullOrWhiteSpace(state.CopyUpToDateMarkerItem) || !state.CopyReferenceInputs.Any())
```
tmeschter, line 442: What is this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/IProjectConfigurationDimensionsProvider5.cs
davkean: Because it clashed with the existing one from CPS.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
davkean: It's a version of String that is marked with nullable.

Issue #2141, opened 2017-05-08
Disallow editing contentfiles from packages


srivatsn:
Content files from nuget package simply show up in the tree as editable items. https://github.com/dotnet/project-system/issues/604 tracks coming up with a proper editing experience by cloning the file but until then we should just mark them as readonly so that users don't accidentally change the packages.

drewnoakes:
When addressing this, see if we can get invocation to open content file items in the dependencies tree to open the file, also non-editable.

Barsonax:
Any chance this is gonna get fixed soon? This seems to be a pretty serious bug as it makes it possible to edit a shared package.

According to the documentation contenfiles should be readonly so tooling should respect that!

jjmew:
@ocallesp This is came up because #6290 we believe this is important as we shouldn't allow those files to be edited.

ocallesp:
@drewnoakes I created a simple Console NetCore project and added the nuget package AdamsLair.Duality.Samples.Tilemaps. There are some files under Content Files, but cannot open or edit them. 
Was editing already disabled ? It seems that this issue was fixed, but should be able to open the res files ? 

![image](https://user-images.githubusercontent.com/510598/89735650-028c2d00-da19-11ea-84ea-62e4359c848f.png)

drewnoakes:
@ocallesp these nodes were never wired up for interaction beyond the browse object. We may do that in future.

This bug relates to the equivalent items within the "Data" folder at the bottom of your screenshot.

davkean:
PR: https://github.com/dotnet/project-system/pull/6480 and https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/266425 (Microsofties only).

PR #6480, opened 2020-08-10
Mark package content items as "user read-only"

davkean:
Fixes: https://github.com/dotnet/project-system/issues/2141

This marks package content items as user read-only, which causes them to be opened read-only in the editor.

This depends on https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/266425.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6480)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/Compile.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/Compile.xaml
@@ -67,6 +67,14 @@
     </StringProperty.Metadata>
   </StringProperty>
 
+  <StringProperty Name="NuGetPackageId"
+                  Visible="false">
+    <StringProperty.Metadata>
+      <NameValuePair Name="DoNotCopyAcrossProjects"
+                     Value="true" />
+    </StringProperty.Metadata>
+  </StringProperty>
+
   <StringProperty Name="SubType"
                   Visible="false">
     <StringProperty.DataSource>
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/EditorConfigFiles.xaml
```diff
@@ -67,6 +67,14 @@
...
   </StringProperty>
 
+  <StringProperty Name="NuGetPackageId"
+                  Visible="false">
+    <StringProperty.Metadata>
+      <NameValuePair Name="DoNotCopyAcrossProjects"
+                     Value="true" />
+    </StringProperty.Metadata>
+  </StringProperty>
+
```
drewnoakes, line 77: A little surprised to see this for editor config files.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/Compile.xaml
```diff
@@ -67,6 +67,14 @@
     </StringProperty.Metadata>
   </StringProperty>
 
+  <StringProperty Name="NuGetPackageId"
+                  Visible="false">
+    <StringProperty.Metadata>
+      <NameValuePair Name="DoNotCopyAcrossProjects"
+                     Value="true" />
```
drewnoakes, line 74: TIL about DoNotCopyAcrossProjects.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/EditorConfigFiles.xaml
davkean: NuGet can use any item, we have a specific item for EditorConfigFiles for reasons I don't know, so we need to include it. All other items not listed are going to use the None rule.

PR #6487, opened 2020-08-11
[In progress] Shim/resource editor nexus client

ocallesp:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6487)

### DIFF
--- a/setup/ProjectSystemSetup/Properties/launchSettings.json
+++ b/setup/ProjectSystemSetup/Properties/launchSettings.json
@@ -6,10 +6,10 @@
       "commandLineArgs": "/rootsuffix $(VSSDKTargetPlatformRegRootSuffix) /log",
       "environmentVariables": {
         "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
-        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
-        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
         "CPS_DiagnosticRuntime": "1",
-        "CPS_MetricsCollection": "1"
+        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
+        "CPS_MetricsCollection": "1",
+        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets"
       }
     },
     "Start (with native debugging)": {
@@ -18,10 +18,10 @@
       "commandLineArgs": "/rootsuffix $(VSSDKTargetPlatformRegRootSuffix) /log",
       "environmentVariables": {
         "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
-        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
-        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
         "CPS_DiagnosticRuntime": "1",
-        "CPS_MetricsCollection": "1"
+        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
+        "CPS_MetricsCollection": "1",
+        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets"
       },
       "nativeDebugging": true
     },
@@ -31,10 +31,10 @@
       "commandLineArgs": "/rootsuffix $(VSSDKTargetPlatformRegRootSuffix) /log",
       "environmentVariables": {
         "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
-        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
-        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
         "CPS_DiagnosticRuntime": "1",
+        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
         "CPS_MetricsCollection": "1",
+        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
         "COMPlus_ZapDisable": "1"
       },
       "nativeDebugging": true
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
```diff
@@ -0,0 +1,146 @@
...
+        Private _nexusClientHierarchy As IVsHierarchy
+
+        Public Sub New(resxFilename As String, serviceProvider As System.IServiceProvider)
+
+            ThreadHelper.JoinableTaskFactory.Run(
+                Async Function() As Task(Of Task)
+                    Dim systemQueryService As IProjectSystemQueryService = CType(serviceProvider.GetService(GetType(IProjectSystemQueryService)), IProjectSystemQueryService)
+                    Dim queryableSpace = Await systemQueryService.GetProjectModelQueryableSpaceAsync()
+                    ' TODO : Exception. this operation is not valid
+                    Dim projectQuery = queryableSpace.Projects.With(Function(project) project.SourceFiles.With(Function(f) f.FileName = resxFilename))
```
tmeschter, line 62: I think you want `Where` not `With`.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
```diff
@@ -0,0 +1,146 @@
...
+Namespace Microsoft.VisualStudio.Editors.Common
+
+    ''' <summary>
+    ''' Shim layer to handle nexus and non-nexus APIs based on project
+    ''' </summary>
+    NotInheritable Class ClientShimUtils
+
+        Public Shared Function VSHierarchyShim(itemId As UInteger, VsHierarchy As IVsHierarchy, ServiceProvider As System.IServiceProvider) As IVsHierarchy
+           
+            'TODO - do we need to check if on nexus client ? If yes, how to check ?
```
drewnoakes, line 21: Today @tmeschter pointed me to `KnownUIContexts.CloudEnvironmentConnectedContext.IsActive` which may be helpful. I don't think it differentiates between client/server, but should differentiate between Codespaces/local.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
tmeschter: To expand on that: `With` is used to select the data that will be returned by the query. `Where` is used to filter the query. So I think you want something like this instead:

``` VB
queryableSpace.Projects.With(Function(project) project.Name).With(Function(project) project.SourceFiles.Where(Function(source) source.FileName = resxFilename), isRequiredNotEmpty := True)
```

The first `With` says that you want the project name included in the results.

The second `With` says you want the source files in the project, but only the ones that match `resxFilename` (that's what `Where` does), and you only want projects where the filtered set of source files is not empty (`isRequiredNotEmpty := True`).

At least I think that's what you want. I'm basing this on the `TestQueryProjectWithSourceItemMatchingCondition` unit test in the CPS repo.

ocallesp: This query also returns returns the same exception. This operation is not supported.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
ocallesp: Thanks. It works.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
ocallesp: I just confirmed with Lifeng, this is an issue that is reproducible in VB.

setup/ProjectSystemSetup/Properties/launchSettings.json
```diff
@@ -31,10 +31,10 @@
       "commandLineArgs": "/rootsuffix $(VSSDKTargetPlatformRegRootSuffix) /log",
       "environmentVariables": {
         "VisualBasicDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.VisualBasic.DesignTime.targets",
-        "FSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.FSharp.DesignTime.targets",
-        "CSharpDesignTimeTargetsPath": "$(VisualStudioXamlRulesDir)Microsoft.CSharp.DesignTime.targets",
         "CPS_DiagnosticRuntime": "1",
```
davkean, line 34: Why were these moved around? Can you do that in a separate PR?

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
```diff
@@ -0,0 +1,210 @@
...
+                Return New ClientVsHierarchy(vsHierarchy, path, serviceProvider)
+            End If
+
+            Return vsHierarchy
+        End Function
+
+        Private Shared Function IsMiscProjectFile(itemId As UInteger, vsHierarchy As IVsHierarchy) As Boolean
+
+            Dim pguid As Guid
+            vsHierarchy.GetGuidProperty(itemId, __VSHPROPID.VSHPROPID_ProjectIDGuid, pguid)
```
davkean, line 38: ProjectID is the unique identifier of a project within a solution, you want to get the type of the project, which is provided by TypeGuid.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
```diff
@@ -0,0 +1,210 @@
...
+Imports System.IO
+
+Namespace Microsoft.VisualStudio.Editors.Common
+
+    ''' <summary>
+    ''' Shim layer to handle nexus and non-nexus APIs based on project
+    ''' </summary>
+    Friend NotInheritable Class ClientShimUtils
+
+        Public Shared Function VSHierarchyShim(itemId As UInteger, vsHierarchy As IVsHierarchy, serviceProvider As System.IServiceProvider) As IVsHierarchy
```
davkean, line 19: Methods start with verbs, not nouns; CreateCloudEnvironmentShim for example.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
```diff
@@ -0,0 +1,210 @@
...
+            If pguid = VSConstants.CLSID.MiscellaneousFilesProject_guid Then
+                Return True
+            End If
+
+            Return False
+        End Function
+
+    End Class
+
+    Friend Class ClientVsHierarchy
```
davkean, line 49: I think this is going be problematic; IVsHierarchy is an exchange type and is used commonly in reference quality and passed to APIs such as IVsSolution, etc. I think it will be harder maintain to track down and make changes in the future if we try to "fake" ourselves a hierarchy. I would consider introduce a new API that you pass around that provides the same functionality, but one wraps a real hierarchy; one wraps the Query API. Or alternatively, more entirely across to the Query API everywhere.

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
```diff
@@ -0,0 +1,210 @@
...
+        Implements IVsHierarchy, IVsProject
+
+        Private ReadOnly _nexusClientHierarchy As IVsHierarchy
+        Private ReadOnly _vsProject As IVsProject
+        Private _file As ISourceFile
+        Private _project As IProject
+
+        Public Sub New(vsHierarchy As IVsHierarchy, resxFilename As String, serviceProvider As System.IServiceProvider)
+            _nexusClientHierarchy = vsHierarchy
```
davkean, line 58: We avoid code names in type names. "Nexus" is a code name.

This is actually the misc project hierarchy, and should probably be named that.

setup/ProjectSystemSetup/Properties/launchSettings.json
ocallesp: this settings will be ignored in the end

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
ocallesp: done

src/Microsoft.VisualStudio.Editors/Common/ClientShimUtils.vb
ocallesp: I think I should wait for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1178677

PR #6490, opened 2020-08-12
[DO NOT MERGE] Separate the RPS fix

dmonroym:
[DO NOT MERGE] 
Trying to isolate any other merges that might've happened in between the regression and the fix. This deletes many work in between so DO NOT MERGE - for testing purposes only to make sure DDRITs and RPS have been fixed

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6490)


dmonroym:
There's way too many changes that make this unusable

PR #6493, opened 2020-08-13
Implement Signing property page

### LINKED ISSUE #5954: Create a "Signing" project properties page back-end for VS Online
For the short-term we should concentrate on the strong-name signing aspects of the page. That is, the parts under "Sign the assembly" in the existing page are the priority here. The ClickOnce-specific parts can wait until later.

tmeschter:
Fixes #5954.

Implement a basic Signing property page. The existing Signing property page is implemented in the VS repo and also includes settings specific to ClickOnce scenarios. Here we just implement the parts specific to assembly signing as those are what we need for Codespaces scenarios.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6493)


dmonroym:
Is this a workaround with the final intention to enable ClickOnce for Codespaces?

tmeschter:
@dmonroym I'm not sure. Right now Codespaces is focused on ASP.NET Core development as the motivating scenario (at least in relation to .NET) and that doesn't overlap with ClickOnce.

PR #6494, opened 2020-08-13
Publish packages to an Azure DevOps VS IDE (vs-impl) public feed instead of myget

### LINKED ISSUE #6079: Publish packages to an Azure DevOps VS IDE public feed instead of myget
Currently the repo publishes packages to Roslyn MyGet feed.

It should instead publish to one of the following Azure DevOps feeds:

- https://dev.azure.com/azure-public/vside/_packaging?_a=feed&feed=vssdk
- https://dev.azure.com/azure-public/vside/_packaging?_a=feed&feed=vs-impl

The former feed is for packages that third party customers might need to reference, the latter are to be referenced only by VS own components.

The goal is to remove dependencies on MyGet servers.

Related issue: https://github.com/dotnet/roslyn/issues/39279

### LINKED ISSUE #6602: The xml files in the nupkg fail the codesign verify despite being signed
The xml files in the nupkg fail the codesign verify despite being signed in https://github.com/dmonroym/project-system/blob/9e3cf5fe859c0f4d188412e32456f7611782cb2f/build/ci/SignToolData.json#L37

This task should be added to https://github.com/dotnet/project-system/blob/HEAD/build/ci/official.yml

```yml
- task: ms-vseng.MicroBuildShipTasks.7c429315-71ba-4cb3-94bb-f829c95f7915.MicroBuildCodesignVerify@2
  displayName: Verify Signed Files
  inputs:
    TargetFolders: |
      $(Build.SourcesDirectory)/artifacts/$(BuildConfiguration)/packages
```

dmonroym:
Fixes #6079
Fixes #6602

We chose to publish all packages to vs-impl (as opposed to vssdk) as they're not packages that third party customers might need to reference, but rather are to be referenced only by VS own components.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6494)


build/ci/official.yml
```diff
@@ -129,11 +129,21 @@ steps:
...
   condition: succeededOrFailed()
 
+- task: NuGetToolInstaller@0
+  inputs:
+  versionSpec: '4.9.2'
+
+# Authenticate with service connections to be able to publish packages to external nuget feeds.
+- task: NuGetAuthenticate@0
+  inputs:
+  nuGetServiceConnections: azure-public/vs-impl, azure-public/vssdk
```
davkean, line 139: Why do we need VSSDK as a connection if we're publishing to vs-impl?

dmonroym: This one is still a draft, just copied over the same thing Roslyn is doing - don't actually think we even need this step if the NuGet Publisher task succeeds. They need to Authenticate because they're using a ps1

build/ci/official.yml
```diff
@@ -129,12 +129,29 @@ steps:
     ArtifactType: Container
   condition: succeededOrFailed()
 
-- task: NuGetPublisher@0
-  displayName: NuGet Publisher - Roslyn Myget
+- task: ms-vseng.MicroBuildShipTasks.7c429315-71ba-4cb3-94bb-f829c95f7915.MicroBuildCodesignVerify@2
+  displayName: Verify Signed Files
```
davkean, line 133: Do we currently sign our packages? Have you tried this on a signed build?

build/ci/official.yml
```diff
@@ -129,12 +129,29 @@ steps:
...
+- task: NuGetAuthenticate@0
+  inputs:
+    nuGetServiceConnections: azure-public/vs-impl
+
+# TODO - this'll most likely not work so use something like eng\publish-assets.ps1 which is used by Roslyn
+- task: NuGetCommand@2
+  displayName: NuGet push - vs-impl
+  inputs:
+    command: push
+    packagesToPush: 'artifacts\$(BuildConfiguration)\packages\*.nupkg'
```
davkean, line 152: I know you didn't change this, but more for education, what happens with builds against private branches? How do we avoid overlap with package version numbers?

build/ci/official.yml
rrelyea: when i look at this package with "nuget verify -all microsoft.visualstudio.projectsystem.managed.16.7.0-beta1-65318-02.nupkg"
it looks like that package is author signed by microsoft.

dmonroym: Yes, it is signed. This verification should be added and it's one of the requirements when pushing to vs-impl. When talking with James Williams he suggested this be done in the release pipeline instead though

build/ci/official.yml
dmonroym: Moving this to the release pipeline will help avoid that

build/ci/official.yml
dmonroym: Opened https://github.com/dotnet/project-system/issues/6602 to track this

dmonroym:
Verified the build worked with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4053419&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9
and the release works with https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=808269

Additionally you can now find the packages in https://dev.azure.com/azure-public/vside/_packaging?_a=feed&feed=vs-impl

davkean (approved):

PR #6517, opened 2020-08-20
Update WarningLevel values to include 5

Youssef1313:
See https://github.com/dotnet/roslyn/blob/master/docs/compilers/CSharp/Warnversion%20Warning%20Waves.md

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6517)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
@@ -1431,9 +1431,9 @@
   <data name="&gt;&gt;chkDefineDebug.ZOrder" xml:space="preserve">
     <value>23</value>
   </data>
-  <metadata name="$this.Localizable" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
+  <data name="$this.Localizable" type="System.Boolean, mscorlib">
     <value>True</value>
-  </metadata>
+  </data>
   <data name="$this.AutoScaleDimensions" type="System.Drawing.SizeF, System.Drawing">
     <value>6, 13</value>
   </data>
@@ -1448,5 +1448,8 @@
   </data>
   <data name="&gt;&gt;$this.Type" xml:space="preserve">
     <value>Microsoft.VisualStudio.Editors.PropertyPages.PropPageUserControlBase, Microsoft.VisualStudio.AppDesigner, Version=42.42.42.42, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
+  </data>
+  <data name="cboWarningLevel.Items5" xml:space="preserve">
+    <value>5</value>
   </data>
 </root>
--- a/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.Designer.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.Designer.vb
@@ -165,7 +165,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             resources.ApplyResources(Me.cboWarningLevel, "cboWarningLevel")
             Me.cboWarningLevel.DropDownStyle = System.Windows.Forms.ComboBoxStyle.DropDownList
             Me.cboWarningLevel.FormattingEnabled = True
-            Me.cboWarningLevel.Items.AddRange(New Object() {resources.GetString("cboWarningLevel.Items"), resources.GetString("cboWarningLevel.Items1"), resources.GetString("cboWarningLevel.Items2"), resources.GetString("cboWarningLevel.Items3"), resources.GetString("cboWarningLevel.Items4")})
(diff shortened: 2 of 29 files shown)


drewnoakes:
@Youssef1313 the rule file is localised and needs to have its corresponding `.xlf` files updated. Building the solution will do this. That updated output needs to be included in this PR.

davkean:
This isn't enough to plumb through this warning level, this will only enable it for Codespaces. It's likely we'll need to plumb this through the AppDesigner. Opt'ing into this warning, completely breaks all options inside VS (its as if the options are being thrown away) - so I think there's more work to enable this on the Roslyn side. Plus, we need to think about legacy project system.

Youssef1313:
> @Youssef1313 the rule file is localised and needs to have its corresponding .xlf files updated. Building the solution will do this. That updated output needs to be included in this PR.

@drewnoakes I'm trying to do that locally now.


> 
> 
> This isn't enough to plumb through this warning level, this will only enable it for Codespaces. It's likely we'll need to plumb this through the AppDesigner. Opt'ing into this warning, completely breaks all options inside VS (its as if the options are being thrown away) - so I think there's more work to enable this on the Roslyn side. Plus, we need to think about legacy project system.

I honestly don't understand most of this :-(
I'm not that too experiened. Just was updating dotnet/docs to include this warning level and thought to search through the whole dotnet organization for other places that might need this.

Could you please clarify more what should be done whether in project-system or roslyn?

davkean (changes requested):
I've started a thread internally to figure this out, TBH, I'm little confused with warn levels iteraction with warning waves: https://github.com/dotnet/roslyn/pull/45750/files.

davkean (approved):

davkean:
Okay, I understand how this works. Instead of the original warning waves proposal, C# will just add another warning level and will continue to add warning levels in the future. This isn't sustainable for us to continually updated this dialog, however, we're working on a rewrite of these property pages, and we might might want to think about how we pull "available" warning levels from Roslyn. // @melytc @drewnoakes 

For now I've updated the PR with all the rest of the changes, and made sure localization files got updated. Thanks for the contribution!

davkean:
New project system: Correctly handles and passes it through correctly.
Old project system: It rejects the warning level as you try to set it. As we're not making changes to it, this behavior is fine.

src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
```diff
@@ -1431,9 +1431,9 @@
   <data name="&gt;&gt;chkDefineDebug.ZOrder" xml:space="preserve">
     <value>23</value>
   </data>
-  <metadata name="$this.Localizable" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
```
davkean, line 1434: This is moving from old style of representing this property, to the new one. It was automatically done by the designer.

src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
```diff
@@ -1449,4 +1449,7 @@
   <data name="&gt;&gt;$this.Type" xml:space="preserve">
     <value>Microsoft.VisualStudio.Editors.PropertyPages.PropPageUserControlBase, Microsoft.VisualStudio.AppDesigner, Version=42.42.42.42, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
   </data>
+  <data name="cboWarningLevel.Items5" xml:space="preserve">
+    <value>5</value>
+  </data>
```
drewnoakes, line 1454: Is there a way to mark a specific resource string as non-localisable?

drewnoakes (approved):

src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
swesonga: Did we recently update the designer to do this?

---
In reply to: [481792589](https://github.com/dotnet/project-system/pull/6517#discussion_r481792589) [](ancestors = 481792589)

melytc (approved):

src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
davkean: No this was either the transition from 1.0 -> 2.0 or from 3.5 -> 4.0. We just haven't opened the file in VS since then.

src/Microsoft.VisualStudio.Editors/PropPages/BuildPropPage.resx
davkean: Possibly but I couldn't figure it out. @tmeschter Do you know? I'll file a followup to fix if there is, this just following the existing pattern.

tmeschter: The best option is to just keep the value in code rather than in a .resx file.

Second best option is to add a `<comment>` element (or maybe it is `<note>`; I forget) to the `<data>` item in the .resx file and state that the text should be left as-is across all languages. Those notes are copied to the .xlf files and are visible to the translators.

PR #6523, opened 2020-08-20
RCS1239: Use 'for' statement instead of 'while' statement

elachlan:
For statement makes more sense.
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1239.md

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6523)


davkean (changes requested):
We don't believe this improves the code base.

PR #6524, opened 2020-08-20
RCS1118: Mark local variable as const

elachlan:
Marking local variables as const for explicitness and performance.
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1118.md

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6524)


drewnoakes:
There is no performance impact here that I can see.

While it is more explicit, I don't see the value here personally. I hope we one day have something like `let` for readonly locals.

davkean (changes requested):
We've don't think this adds much value to the code base.

Issue #6528, opened 2020-08-20
Add AnalysisLevel to code analysis  property pages


jmarolf:
Related: https://github.com/dotnet/roslyn-analyzers/issues/3823

Not that the msbuild property has been finalized and merged into the SDK we should add support for it in the.

UX is envisioned to be:

![image](https://user-images.githubusercontent.com/9797472/90820341-191e6800-e2e6-11ea-8786-2baaf3e51b26.png)

davkean:
How does the Version drop down associated with AnalysisLevel?

PR #6529, opened 2020-08-21
Add support for AnalysisLevel and EnableNETAnalyzers properties

jmarolf:
fixes https://github.com/dotnet/project-system/issues/6528

This is what the code analysis page looks like with this change:

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6529)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
@@ -1,5 +1,6 @@
 ﻿' Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+Imports System.ComponentModel
 Imports System.Windows.Forms
 
 Namespace Microsoft.VisualStudio.Editors.PropertyPages
@@ -8,6 +9,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         Inherits PropPageUserControlBase
 
         Private Const RoslynAnalyzersDocumentationLink As String = "https://docs.microsoft.com/visualstudio/code-quality/roslyn-analyzers-overview"
+        Private Const NETAnalyzersDocumentationLink As String = "https://aka.ms/dotnetanalyzers"
 
         Public Sub New()
             MyBase.New()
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.Editors/PropPages/CodeAnalysisPropPage.vb
```diff
@@ -19,20 +21,45 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
 
         Protected Overrides ReadOnly Property ControlData As PropertyControlData()
             Get
                 If m_ControlData Is Nothing Then
                     m_ControlData = New PropertyControlData() {
                         New PropertyControlData(1, "RunAnalyzersDuringBuild", RunAnalyzersDuringBuild, ControlDataFlags.None),
-                        New PropertyControlData(2, "RunAnalyzersDuringLiveAnalysis", RunAnalyzersDuringLiveAnalysis, ControlDataFlags.PersistedInProjectUserFile)
+                        New PropertyControlData(2, "RunAnalyzersDuringLiveAnalysis", RunAnalyzersDuringLiveAnalysis, ControlDataFlags.PersistedInProjectUserFile),
+                        New PropertyControlData(3, "EnableNETAnalyzers", EnableNETAnalyzersCheckBox, ControlDataFlags.None),
+                        New PropertyControlData(4, "AnalysisLevel", AnalysisLevelComboBox, AddressOf AnalysisLevelSet, AddressOf AnalysisLevelGet, ControlDataFlags.None)
```
drewnoakes, line 35: Should this control only be enabled when "EnabledNETAnalyzers" is true?

jmarolf: I thought about it and decided to not disable the control (and set its value to `none`) since there is a case where you would want to configure both.

So `AnalysisLevel` set to "none"  implicitly disables analyzers (and therefore automatically unchecks the `EnableNETAnalyzersCheckBox`)

However it is valid to have `<EnableNETAnalyzers>false</EnableNETAnalyzers>` with `<AnalysisLevel>5.0</AnalysisLevel>`. This indicates that you do not want any analyzers to give you warnings about mis-using apis but you still want new language warnings from the compiler.

Maybe this is too much of a corner case and we should just tie the two together?

drewnoakes: Unclear to me at this point. My question was motivated by thinking how this might work in the new property pages. If you feel there's a better experience here, let's work through a design.

The UI rarely offers the full range of configuration options that exist by directly editing the project file. If we can simplify and improve the experience for 99% of users, I'm for it.

PR #6486, opened 2020-08-11
Enable MyApplication properties for WinForms apps targeting .NET5 or higher.

melytc:
We want to support these properties found in the Application page:
windows-application-framework-porperties
(`UseApplicationFrameworkCheckBox` and `WindowsAppGroupBox`)

These properties are enabled by `MyApplicationProperties`, where we validate if the capability `EnableMyApplication` is present.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6486)

PR #6548, opened 2020-08-26
Fix exception for empty TargetFrameworkVersion string

melytc:
Separated the conditionality for .NET Framework and .NET Core, and for the latter added the case for an empty TargetFrameworkVersion string.

Fix for bug introduced in #6486. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6548)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.VisualBasic.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.VisualBasic.DesignTime.targets
@@ -40,7 +40,9 @@
                           ClassDesigner;
                           SharedProjectReferences;" />
 
-      <ProjectCapability Include="EnableMyApplication" Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)','5.0')) And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(UseWindowsForms)' == 'true'" />
+      <ProjectCapability Include="EnableMyApplication" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'"/>
+
+      <ProjectCapability Include="EnableMyApplication" Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(UseWindowsForms)' == 'true' And '$(TargetFrameworkVersion)' != '' And $([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)','5.0'))"/>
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.VisualBasic.DesignTime.targets
```diff
@@ -40,7 +40,9 @@
                           ClassDesigner;
                           SharedProjectReferences;" />
 
-      <ProjectCapability Include="EnableMyApplication" Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)','5.0')) And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(UseWindowsForms)' == 'true'" />
+      <ProjectCapability Include="EnableMyApplication" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'"/>
```
davidwengier, line 43: Does this work? MyApplication wasn't enabled for SDK-style projects targeting .NET Framework before.

drewnoakes: The `EnableMyApplication` capability has been added to legacy as well.

davkean: There's a couple of combos:

1) Legacy with WPF
2) Legacy with WinForms
3) CPS (.NET Framework) with WPF
4) CPS (.NET Framework) with WinForms
5) CPS (.NET Core) with WPF
6) CPS (.NET Core) with WinForms

What exactly is the state for all these? And what hit the path guarded by EnableMyApplication?

davidwengier: Previous to #6486 only 1 and 2 in that list would use MyApplication features.

drewnoakes: This PR doesn't take WPF into account.

I suggest merging this as-is to unblock dogfooders. If there's any outstanding action required, we can file an issue. It should be easy to create a follow up PR.

drewnoakes (approved):

PR #6549, opened 2020-08-26
Simplify IWaitIndicator 

davkean:
IWaitIndicator had a significant amount of code handling situations that were never hit, just simplify into a single method that returns a result.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6549)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Waiting/WaitIndicatorResult.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Waiting/WaitIndicatorResult.cs
@@ -1,15 +1,50 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
+
 namespace Microsoft.VisualStudio.ProjectSystem.Waiting
 {
-    internal enum WaitIndicatorResult
+    /// <summary>
+    ///     Represents the result of <see cref="IWaitIndicator.Run{T}(string, string, bool, Func{System.Threading.CancellationToken, System.Threading.Tasks.Task{T}})"/>.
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Waiting/WaitIndicatorResult.cs
```diff
@@ -1,15 +1,50 @@
...
+            get
+            {
+                Verify.Operation(!_isCancelled, "Cannot get the result of a cancelled operation.");
+
+                return _result!;
+            }
+        }
+
+#pragma warning disable RS0038 // Prefer null literal
+        public static WaitIndicatorResult<T> Cancelled = new WaitIndicatorResult<T>(isCancelled: true, result: default!);
```
drewnoakes, line 45: `readonly` field or a `get`-only property.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Waiting/WaitIndicatorResult.cs
```diff
@@ -1,15 +1,50 @@
...
 
+using System;
+
 namespace Microsoft.VisualStudio.ProjectSystem.Waiting
 {
-    internal enum WaitIndicatorResult
+    /// <summary>
+    ///     Represents the result of <see cref="IWaitIndicator.Run{T}(string, string, bool, Func{System.Threading.CancellationToken, System.Threading.Tasks.Task{T}})"/>.
+    /// </summary>
+    internal class WaitIndicatorResult<T>
```
drewnoakes, line 10: Could be a `readonly struct`.

drewnoakes (approved):

Issue #402, opened 2016-08-27
Confusing error when you attempt to delete a reference


davkean:
In legacy csproj this doesn't prompt, in CPS it does:

![image](https://cloud.githubusercontent.com/assets/1103906/18023838/270ab3e6-6bb3-11e6-89c4-c7571e83898d.png)

This is confusing - what does it mean that is contents will be removed?

basoundr:
Blocked by #708

abpiskunov:
yo are not blocked - open Add Reference dialog and uncheck your reference there

basoundr:
This scenario happens only when going through the Remove Button and not through the Add Reference dialog.

davkean:
Also get the same error when removing a Package reference:

![image](https://cloud.githubusercontent.com/assets/1103906/22102713/84959200-de8c-11e6-8a91-ee697d3e6b12.png)

panopticoncentral:
Removing the Parity-Legacy tag from this one because it's not really a parity issue in terms of functionality. It's just a strange behavior of the project system that it would be nice to fix in general.

Issue #6183, opened 2020-05-12
Use Dataflow and snapshots for dependencies node data processing


drewnoakes:
The processing of data through the Dependencies node code has a few issues that I'd like to address in a batch, potentially in 16.8:

1. We have bugs around project configurations (e.g. multi-targeting bugs)
   - #3642
   - #2618
   - #2525
   - #2524
   - #2795
   - #2617
   - #6203
   - Historically: #6204
   - Historically: #6319
   - Historically: #2713
   - Historically: #2388
1. Data providers currently notify of updates provide via .NET events
   - This is in contrast to most everywhere else where we use Dataflow
   - Makes it hard to implement features such as:
      - #3426
1. Data providers are stateless, and publish deltas
   - This is in contrast to most everywhere else where we send snapshots
   - Makes it difficult to reconcile differences between evaluation and design-time build data
      - #2510
         - Seen a fair few dupes of this (#5335, #4550, possibly #5356)
      - #5044
   - Makes it hard to implement features such as:
      - #6139
      - #76
      - @ocallesp's work to show unresolved references in the tree
      - NuGet potentially showing showing upgradeable packages in the tree
   - A per-provider snapshot should make it easier to reduce memory wasted on long string identifiers
     - #5051
1. There are two ways to submit tree updates (public `IProjectDependenciesSubTreeProvider` and internal `IDependencyCrossTargetSubscriber`) where there should only be one
1. ~The concept of 'filters' should be removed~
   - They attempt to work around the statelessness of dependency providers but distribute logic in multiple places, which makes design and debugging difficult
   - They require passing forward lots of state for later steps to use, which could be avoided if the providers were more self-contained and did that work themselves
   - Fixed in #7992
1. Debugging Dependencies node issues is harder than it should be
   - If providers produced snapshots it'd be much easier to inspect their behaviour
   - Easier to know why we have a yellow triangle
      - #4885
1. By using Dataflow we could have fewer concepts and less code
   - Remove e.g. `IDependencyCrossTargetSubscriber`, `AggregateCrossTargetProjectContext`, `AggregateCrossTargetProjectContextProvider`
   - Use CPS's infrastructure e.g. `UnwrapCollectionChainedProjectValueDataSource`
1. The "Dependencies" tree does not populate with the first version of the visible tree. Its children appear later.

The work is mostly self-contained within this repo, however it will also require synchronising changes with the WebTools team for NPM dependencies.

PR #6552, opened 2020-08-26
Review use of ProjectTreeFlags in the dependencies tree

drewnoakes:
Tidies up a bunch of the tree flags used by the dependencies tree.
Fixes: https://github.com/dotnet/project-system/issues/402

Review commit by commit.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6552)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeFlags.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeFlags.cs
@@ -60,8 +60,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
         /// </summary>
         public static readonly ProjectTreeFlags DependencyFlags
                 = Dependency
-                + ProjectTreeFlags.VirtualFolder
-                + ProjectTreeFlags.BubbleUp
                 + SupportsRuleProperties
                 + SupportsRemove;
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeFlags.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeFlags.cs
@@ -19,7 +19,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies
         /// </summary>
         internal static readonly ProjectTreeFlags DependenciesRootNode = ProjectTreeFlags.Create("DependenciesRootNode");
 
-        internal static readonly ProjectTreeFlags GenericDependency = ProjectTreeFlags.Create("GenericDependency");
+        /// <summary>
+        /// Applied to all top-level dependency items under the dependencies tree.
+        /// </summary>
+        internal static readonly ProjectTreeFlags Dependency = ProjectTreeFlags.Create("Dependency");
 
         /// <summary>
         /// Added to dependency tree items which can be removed from the project.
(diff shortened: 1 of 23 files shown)


davkean:
`BrokenReference` is meant to indicate references that failed to resolve, not has not yet been resolved (which would be Dependency but no ResolveReference flag). I know (unlike legacy) we don't have a way to separate these states but we may want to consider if this is the right approach.

Does this impact web tools?

drewnoakes:
> BrokenReference is meant to indicate references that failed to resolve, not has not yet been resolved (which would be Dependency but no ResolveReference flag). I know (unlike legacy) we don't have a way to separate these states but we may want to consider if this is the right approach.

Makes sense. #6183 will give us a way to model that.

> Does this impact web tools?

There's no public API change. Scanning index shows no use of the GenericDependency flag.

davkean:
Based on my reading the code, this will fix https://github.com/dotnet/project-system/issues/402. "Removes" (not deletes) of single items do not prompt. It's prompting  because it's marked with virtual folder.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeFlags.cs
```diff
@@ -56,35 +59,20 @@ public static class DependencyTreeFlags
         /// implementations. This is to have a way to distinguish dependency nodes in general.
         /// </summary>
         public static readonly ProjectTreeFlags DependencyFlags
-                = ProjectTreeFlags.Create("Dependency")
-                + ProjectTreeFlags.VirtualFolder
-                + ProjectTreeFlags.BubbleUp
+                = Dependency
+                + ProjectTreeFlags.Reference
```
swesonga, line 63: >Reference [](start = 35, length = 9)

Doesn't this contradict the comment above about this flag being used with all nodes?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/TargetDependencyViewModel.cs
```diff
@@ -26,7 +26,9 @@ static ProjectTreeFlags GetCachedFlags(TargetFramework targetFramework)
                 return ImmutableInterlocked.GetOrAdd(
                     ref s_configurationFlags,
                     targetFramework.FullName,
-                    fullName => DependencyTreeFlags.TargetNode.Add($"$TFM:{fullName}"));
+                    fullName => DependencyTreeFlags.TargetNode
+                        .Add($"$TFM:{fullName}")
+                        .Add(ProjectTreeFlags.Common.VirtualFolder));
```
swesonga, line 31: >VirtualFolder [](start = 53, length = 13)

What does adding this flag change in the behavior of the node?

davkean (approved):

davkean:
I had a look, we don't use BrokenReference as any indication right now, but I hope in the future we use it as indicator to show the warning icon.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeFlags.cs
drewnoakes: I don't see the contradiction. Can you elaborate?

Note that this flag was being added via `GenericUnresolvedDependencyFlags` and `GenericResolvedDependencyFlags` below. This just moves it higher up.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/TargetDependencyViewModel.cs
drewnoakes: This PR actually _removes_ `VirtualFolder` from most nodes (via `DependencyTreeFlags.DependencyFlags`). This addition here keeps it present on a node for which it makes sense.

The flag itself isn't used very widely in CPS. It's used to short circuit some tree scanning. The docs on the flag itself aren't very helpful.

PR #6560, opened 2020-08-27
allow the user to opt-in to enforcing code analysis rules on build.

jmarolf:


jmarolf:
@dotnet/project-system ping?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
```diff
@@ -448,4 +448,14 @@
                   SourceOfDefaultValue="AfterContext" />
     </StringProperty.DataSource>
   </StringProperty>
+
+  <BoolProperty Name="EnforceCodeStyleInBuild"
```
swesonga, line 452: >In [](start = 38, length = 2)

this is super nit-picky but here goes... this is referred to as on build everywhere but in the actually property name

swesonga (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralBrowseObject.xaml
jmarolf: It looked to me that the naming pattern is to use whatever the property is called in MSBuild. I named the property this in the sdk targets so it would appear like this in the project file:

```xml
<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```

What is the general naming scheme that we want to use for xaml rules files? Happy to fix this in a follow up PR :)

dmonroym (approved):

PR #6562, opened 2020-08-28
Bind F12 to "Navigate to Project"

davkean:
All the navigate commands, "Go To Test", "Go To Definition" are all bound to F12. Do the same here, but make sure its only in Solution Explorer which does not use this shortcut.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6562)


davkean:
https://twitter.com/DamianEdwards/status/1299223626659450880

Issue #81, opened 2016-05-13
Renaming Properties folder should work or be disabled


davkean:
_From @davkean on March 22, 2016 23:45_

CPS allows you to rename Properties/My Project folder. Either we should disallow like csproj/msvbproj, or make sure we do the right things to update the associated project property; `<AppDesignerFolder>`

_Copied from original issue: dotnet/roslyn#9993_

jjmew:
Look at ProjectTreeActionHandler as way to fixing this

PR #6565, opened 2020-08-29
Fix AppDesignerFolder when renamed

davkean:
Fixes: https://github.com/dotnet/project-system/issues/81

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6565)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AppDesigner.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AppDesigner.xaml
@@ -15,7 +15,6 @@
                 Visible="false">
     <BoolProperty.DataSource>
       <DataSource HasConfigurationCondition="false"
-                  Label="Configuration"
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/AppDesigner.xaml
```diff
@@ -27,7 +26,6 @@
                   Visible="false">
     <StringProperty.DataSource>
       <DataSource HasConfigurationCondition="false"
-                  Label="Configuration"
```
swesonga, line 30: >Label [](start = 18, length = 5)

I'm guessing this is cleanup/orthogonal to the fix?

davkean: It's related, as we started writing this property, we would write out a special ItemGroup with a Label="Confioguration". We never hit it because we weren't writing it out.

PR #6567, opened 2020-09-01
Avoid SubType fighting in multi-target projects

davkean:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/961596

Roslyn is responsible for applying SubType to Forms, Components and other types. In multi-targeted projects, the same file can have conflicting "sub types":

```
#if NET45
    class MyComponent : Form
#else
    class MyComponent : Component
#endif
```

This results Roslyn pushing "Form" when it encounters scans the first definition and "Component" when it encounters the second definition.
There is no guaranteed order to which one of these get scanned first, so this can result in the same source file having a different sub type every time Roslyn does analysis. Sub Type affects the designer and icon that a type gets, so this results in tree flickering between two different icons.

This was worse before https://github.com/dotnet/project-system/pull/5880 as we'd see the sub type change and go back to Roslyn to add/remove the file, resulting in a constant churn. This longer occurs, but we still have the fighting sub type dance whenver the file content changes.

To fix this, Roslyn will now only scan the the primary context: https://github.com/dotnet/roslyn/pull/44270.

This change plumbs through this property by finding the context that lives in the "active configuration" and passing this through to Roslyn.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6567)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectPropertiesItemHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectPropertiesItemHandler.cs
@@ -22,7 +22,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
             get { return LanguageService.SchemaName; }
         }
 
-        public void Handle(IComparable version, IProjectChangeDescription projectChange, bool isActiveContext, IProjectLogger logger)
+        public void Handle(IComparable version, IProjectChangeDescription projectChange, ContextState state, IProjectLogger logger)
         {
             Requires.NotNull(version, nameof(version));
             Requires.NotNull(projectChange, nameof(projectChange));
@@ -42,6 +42,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
                 logger.WriteLine("{0}: {1}", name, value);
                 Context.SetProperty(name, value);
             }
+
+            // NOTE: Roslyn treats "unset" as true, so always set it.
+            Context.IsPrimary = state.IsActiveConfiguration;
         }
 
         private bool TryHandleSpecificProperties(string name, string value, IProjectLogger logger)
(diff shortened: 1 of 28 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectPropertiesItemHandler.cs
```diff
@@ -42,6 +42,9 @@ public void Handle(IComparable version, IProjectChangeDescription projectChange,
                 logger.WriteLine("{0}: {1}", name, value);
                 Context.SetProperty(name, value);
             }
+
+            // NOTE: Roslyn treats "unset" as true, so always set it.
+            Context.IsPrimary = state.IsActiveConfiguration;
         }
```
davkean, line 48: Fix is here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -78,24 +82,39 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
...
-                        suppressVersionOnlyUpdates: false,
-                        ruleNames: _applyChangesToWorkspaceContext.Value.GetProjectEvaluationRules()),
-
-                    _projectSubscriptionService.ProjectBuildRuleSource.SourceBlock.LinkToAsyncAction(
-                        target: e => OnProjectChangedAsync(e, evaluation: false),
-                        _project.UnconfiguredProject,
-                        ProjectFaultSeverity.LimitedFunctionality,
-                        suppressVersionOnlyUpdates: false,
-                        ruleNames: _applyChangesToWorkspaceContext.Value.GetProjectBuildRules())
+                    ProjectDataSources.SyncLinkTo(
```
davkean, line 85: We now also hook onto data flow changes of the active configuration.

drewnoakes (approved):

Issue #6444, opened 2020-07-29
Consume icons from image service


davkean:
For ownership/updating reasons, the images are being taken from this repro and moved to the image service, which they will be updated with the new style.

This bug represents our reaction to this.

- [x] Consume new Microsoft.VisualStudio.ImageCatalog package and consume icons previously consumed via `ManagedImageMonikers` to `KnownMonikers`
- [ ] Delete 
  - [x] Microsoft.VisualStudio.ProjectSystem.Managed.VS\ManagedImages.imagemanifest
  - [ ] ManagedImageMonikers
  - [x] src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\Resources\Icons
  - [x] resources\*
- [x] Remove `<qualifyAssembly/>` for `Microsoft.VisualStudio.ProjectSystem.Managed.VS` in VS repro under src/appid/devenv/stub/devenv.urt.config.tt. [271918](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/271918)

davkean:
Majority of monikers were fixed in https://github.com/dotnet/project-system/pull/6570. The icon folks are working on the others.

PR #6570, opened 2020-09-01
Consume monikers from ImageCatalog vs our own

davkean:
This fixes majority of: https://github.com/dotnet/project-system/issues/6444.

This still remains a few images that have yet to be ported.

This is easier to review you if you go a commit at a time.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6570)


davkean:
First issue:

Dependency node + Dependency warning are not pixel perfect, so when the warning is overlayed the icon appears to shift upwards.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Imaging/ImageMonikerDebuggerDisplay.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    ///     Provides a friendly display for <see cref="ProjectImageMoniker"/> and <see cref="ImageMoniker"/> instances
+    ///     based on the well known values from <see cref="KnownImageIds"/>.
+    /// </summary>
+    internal static class ImageMonikerDebuggerDisplay
+    {
+        private static readonly Dictionary<(Guid, int), string> s_displayNames = CalculateDisplayNames();
+
+        internal static string FromImageMoniker(ImageMoniker moniker) => DebugDisplay(moniker.Guid, moniker.Id);
```
drewnoakes, line 25: This should move to the package containing `Microsoft.VisualStudio.Imaging` and be made public for use in test code and for other types such as `ProjectImageMoniker`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Imaging/ImageMonikerDebuggerDisplay.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        private static Dictionary<(Guid, int), string> CalculateDisplayNames()
+        {
+            Type type = typeof(KnownImageIds);
+
+            return type.GetFields()
+                       .Where(field => field.IsLiteral && field.FieldType == typeof(int))
+                       .ToDictionary(field =>
+                            (KnownImageIds.ImageCatalogGuid, (int)field.GetRawConstantValue()),
+                            field => $"{type.Name}.{field.Name}");
```
drewnoakes, line 47: Does this only run when debugging? We don't want the startup and allocation overhead unless it's used. I'm unclear when a type used with `DebuggerDisplayAttribute` is initialized.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Imaging/ImageMonikerDebuggerDisplay.cs
drewnoakes: I wonder if the code generation in `ManifestToCode` could be extended to help here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Imaging/ImageMonikerDebuggerDisplay.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.Imaging
+{
+    /// <summary>
+    ///     Provides a friendly display for <see cref="ProjectImageMoniker"/> and <see cref="ImageMoniker"/> instances
+    ///     based on the well known values from <see cref="KnownImageIds"/>.
+    /// </summary>
+    internal static class ImageMonikerDebuggerDisplay
+    {
+        private static readonly Dictionary<(Guid, int), string> s_displayNames = CalculateDisplayNames();
```
drewnoakes, line 23: This will result in ~3800 copies of an identical GUID being allocated (~60kB).

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Imaging/ImageMonikerDebuggerDisplay.cs
drewnoakes: Spoke with @davkean offline about this and the type initializer won't run until needed by the debugger, so there's no risk of initializing this type and running this code in non-debugger situations.

davkean:
Filed https://github.com/NuGet/Home/issues/9977 against NuGet.

PR #6593, opened 2020-09-08
Produce telemetry on yellow triangles in the dependency tree

### LINKED ISSUE #6575: Produce telemetry on yellow triangles in the dependency tree
We would like to obtain data about the prevalence of warning icons in the dependencies tree in order to monitor progress in their reduction over time.

It's not currently possible to know when a yellow triangle is displayed due to the dependency being incorrectly specified and therefore unresolvable vs. there being a product bug that results in this error state, being that the tree is the last line of defence.

However we do believe that as we reduce the number of product bugs, we should see a reduction in yellow triangles in telemetry.

---

A pragmatic approach here may be to report the number of resolve/unresolved dependencies of each type when the project is unloaded.

drewnoakes:
Fixes #6575.

On project unload, publish a telemetry event with data about a project's dependencies.

A sample event:

| Field | Value |
|---|---|
| vs.projectsystem.managed.projectunload.dependencies.dependencybreakdown | `{"netcoreapp3.1":{"Framework":{"TotalCount":1,"UnresolvedCount":0},"Package":{"TotalCount":1,"UnresolvedCount":0}},"net48":{"Assembly":{"TotalCount":9,"UnresolvedCount":0},"Package":{"TotalCount":1,"UnresolvedCount":0}}}` |
| vs.projectsystem.managed.projectunload.dependencies.project | 0c63db02-c6b6-465e-a104-4af7bd775739 |
| vs.projectsystem.managed.projectunload.dependencies.projectagemillis | 35636 |
| vs.projectsystem.managed.projectunload.dependencies.targetframeworkcount | 2 |
| vs.projectsystem.managed.projectunload.dependencies.totaldependencycount | 12 |
| vs.projectsystem.managed.projectunload.dependencies.unresolveddependencycount | 0 |

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6593)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
@@ -56,8 +56,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
             }
         }
 
-        public bool IsActive => _stateByFramework != null;
-
         public void InitializeTargetFrameworkRules(ImmutableArray<TargetFramework> targetFrameworks, IReadOnlyCollection<string> rules)
         {
             if (_stateByFramework == null)
@@ -103,7 +101,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
             }
         }
 
-        public async Task ObserveTreeUpdateCompletedAsync(bool hasUnresolvedDependency)
+        public async ValueTask ObserveTreeUpdateCompletedAsync(bool hasUnresolvedDependency)
         {
             if (_stateByFramework == null)
                 return;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
@@ -4,8 +4,10 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies.Snapshot;
 using Microsoft.VisualStudio.Telemetry;
 
 namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
@@ -22,13 +24,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
     /// </remarks>
     [Export(typeof(IDependencyTreeTelemetryService))]
     [AppliesTo(ProjectCapability.DependenciesTree)]
-    internal sealed class DependencyTreeTelemetryService : IDependencyTreeTelemetryService
+    internal sealed class DependencyTreeTelemetryService : IDependencyTreeTelemetryService, IDisposable
     {
         private const int MaxEventCount = 10;
 
         private readonly UnconfiguredProject _project;
         private readonly ITelemetryService? _telemetryService;
         private readonly ISafeProjectGuidService _safeProjectGuidService;
+        private readonly Stopwatch _projectLoadTime = Stopwatch.StartNew();
         private readonly object _stateUpdateLock = new object();
 
         /// <summary>
@@ -39,6 +42,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
 
         private string? _projectId;
         private int _eventCount;
+        private DependenciesSnapshot? _dependenciesSnapshot;
 
         [ImportingConstructor]
         public DependencyTreeTelemetryService(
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -153,11 +157,83 @@ async Task<string> GetProjectIdAsync()
...
+                        {
+                            unresolvedDependencyCount++;
+                        }
+                    }
+                }
+
+                _telemetryService.PostProperties(TelemetryEventName.ProjectUnloadDependencies, new (string, object)[]
+                {
+                    (TelemetryPropertyName.ProjectUnloadProject, _projectId),
+                    (TelemetryPropertyName.ProjectUnloadProjectAgeMillis, _projectLoadTime.ElapsedMilliseconds),
```
davkean, line 212: What are you trying to gather from this? it has no real meaning to anything useful, other than the time that the dependency telemetry service was created to the time it disposed. We have existing events for determine how long the solution was open for, or project load times.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryPropertyName.cs
```diff
@@ -27,6 +27,42 @@ internal static class TelemetryPropertyName
         /// </summary>
         public static readonly string TreeUpdatedUnresolvedProject = BuildPropertyName(TelemetryEventName.TreeUpdatedUnresolved, "Project");
 
+        /// <summary>
+        ///     Identifies the project to which data in the telemetry event applies.
+        /// </summary>
+        public static readonly string ProjectUnloadProject = BuildPropertyName(TelemetryEventName.ProjectUnloadDependencies, "Project");
```
davkean, line 33: "ProjectId" is the common used name of this property, such as "VS.Solution.Perf.ProjectLoad"

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -289,6 +292,11 @@ private void DisposeCore()
                     projectItemSpecs),
                 token);
 
+            if (updatedSnapshot != null)
```
davkean, line 295: This could result in us sending an incomplete snapshot right? ie You do a big branch switch, close VS before the tree has finished we might send unresolved states that are misleading (ie its just that we never received design-time build data).

Perhaps we should only update the snapshot state when we're up-to-date with the latest design-time build data?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryPropertyName.cs
davkean: Nit: These properties are named after the event; so ProjectUnloadDependenciesProject.

drewnoakes: This name was chosen for consistency with `TreeUpdatedResolvedProject` and `TreeUpdatedUnresolvedProject` above.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
drewnoakes: The intention is to make it easy to filter out projects which were not alive for very long and for which all projects were unresolved, as they were likely to not have had long enough to complete and process the results of a design-time build. Unless it's trivial to join this event to other events from the same session, I'd rather keep this here for that reason.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
drewnoakes: No, the behaviour is the same as before. `updatedSnapshot` is only non-`null` if the `TryUpdate` method actually made a change in response to the update.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryPropertyName.cs
drewnoakes: Your second comment wasn't visible when I replied. I can address that. This was intentional based on the idea that we may have other `.../ProjectUnload/*` events which could re-use this property.

davkean: Properties can't be reused, they have the event name embedded in them.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
davkean: I think its more than when we get a JointRule; we can get inbetween states that might fixed themselves once we've caught up. For example, go change the version number a commonly used package in Package.targets to see this - you'll a bunch of errors that will eventually go away, you don't ever want to report a telemetry event during that because we'll get misleading data (if had let it finish, they would have been resolved).

I think we'll want to make sure that we're up-to-date with the latest version of JointRule or perhaps operation progress. @lifengl Thoughts?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
davkean: If you only report the data when we're up-to-date with builds as per below, that wouldn't happen. Using a time span as an indicator of whether the design-time builds results has been processed is not a good fit for telemetry, IMO.

drewnoakes: Agree, but it seemed like a pragmatic way to get indicative data without a clear way to get correct data. This approach would slightly overreport warnings, which is better than underreporting them. Let's discuss other approaches for more correct timing below.

The changes in #6183 would make it clearer when to publish this data, but we want a baseline before starting that work to see how much impact it will have.

jjmew (approved):
:shipit:

PR #6605, opened 2020-09-11
Minor cleanup in DependenciesRuleHandler

davkean:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6605)

PR #6606, opened 2020-09-14
Fix potential ProjectInstance leak

### LINKED ISSUE #5836: Old projectInstance & catalog snapshots can be leaked in TargetedDependenciesSnapshot code
TargetedDependenciesSnapshot can retain old copy of IProjectCatalogSnapshot, when it doesn't need update.

The code will return the previousSnapshot, when no change need to be applied. However, the previousSnapshot will reference a copy of the old project catalog snapshot, which holds a copy of old project instance.

When the project is changed, a new copy of project instance will be created. Because the dependency tree is less likely to be updated, it ends up to retain two copies of the project evaluation results.

drewnoakes:
Fixes #5836.

Removes a guard that attempts to return early from processing on data changes. The condition here only considered one kind of change, which may result it downstream code retaining objects from earlier updates, such as a `ProjectInstance` via `IProjectCatalogSnapshot`.

Downstream code will handle the no-op case correctly anyway, so removing this guard here does not introduce considerably more work for the tree.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6606)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
@@ -109,6 +109,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies.Subscriptions
             IProjectSubscriptionUpdate projectUpdate = e.Item1;
             IProjectCatalogSnapshot catalogSnapshot = e.Item2;
 
+            // Broken design-time builds can produce updates containing no rule data.
+            // Later code assumes that the requested rules are available.
+            // If we see no rule data, return now.
+            if (projectUpdate.ProjectChanges.Count == 0)
+            {
+                return;
+            }
+
             // Create an object to track dependency changes.
             var changesBuilder = new DependenciesChangesBuilder();
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
@@ -109,19 +109,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies.Subscriptions
             IProjectSubscriptionUpdate projectUpdate = e.Item1;
             IProjectCatalogSnapshot catalogSnapshot = e.Item2;
 
-            // Broken design time builds sometimes cause updates with no project changes and sometimes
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
```diff
@@ -109,19 +109,6 @@ IDisposable SyncLink((BufferBlock<IProjectVersionedValue<IProjectSubscriptionUpd
             IProjectSubscriptionUpdate projectUpdate = e.Item1;
             IProjectCatalogSnapshot catalogSnapshot = e.Item2;
 
-            // Broken design time builds sometimes cause updates with no project changes and sometimes
-            // cause updates with a project change that has no difference.
-            // We handle the former case here, and the latter case is handled in the CommandLineItemHandler.
```
drewnoakes, line 114: This comment regarding `CommandLineItemHandler` is a hint that this condition was copied over from an earlier version of the language service (if I remember the origins of the dependencies node code correctly) which likely had different constraints.

davkean: Have you actually verified this works? I'm pretty sure that we have guards later than this that will prevent the new changes from being picked up. See the ones inside: https://github.com/dotnet/project-system/pull/6605.

davkean: There's also assumptions later that evaluation rules always exist which will fail after this empty check,

drewnoakes: The checks you're referring to apply only to one kind of change. The production of a new snapshot (which prevents the leak) is also conditional on changes to the catalog. So long as we allow the change to continue here, a new catalog will be produced which addresses the leak.

> There's also assumptions later that evaluation rules always exist which will fail after this empty check

I'll investigate. It may be that only the second `if` here should be removed. In my testing I didn't see this case trigger, but then I didn't have a broken dtb.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriber.cs
davkean: The comment is wrong, this will not trigger on broken design-time build (there will always be evaluation results in that case), it will trigger when the rules are missing due to bad import, for example.

Issue #6601, opened 2020-09-09
Handle retargeting to/from .NET5.0


melytc:
As there is an effort going on for .NET5 to explicitly specify the platform and platform version in the target framework moniker in the project file, we should handle both cases of upgrading and downgrading.

For WPF and WinForms projects, when trying to retarget to .NET5.0, we should add `-windows` to the target framework moniker in the project file; i.e. from `netcoreapp3.1` to `net5.0-windows`.
- The user will get a warning from the SDK that they are using the WindowsDesktop SDK.

jjmew:
We need a clear spec here of what we want to build. I am keeping this untriaged until we have a spec.

jjmew:
Assigning to @cartermp for the spec

davkean:
@cartermp I think we should ignore the UseXXX properties on upgrade and just always remove it. If we do downgrade, we need to think about if should add it back given they now support building downlevel WPF/WinForms without it on the latest SDK.

cartermp:
Some notes from a meeting with SDK folks:

* The .NET 5 TFM that should be shown in the property pages for WPF and WinForms apps on .NET Core should be `net5.0-windows`, @melytc and @sfoslund to take a look and ensure this gets
* .NET SDK now has logic such that it will import the `WindowsDesktop` targets even if you don't specify it in the SDK attribute in your project file if it is a WinForms or WPF app. So the "downtarget from a new 5.0 winforms app to 3.1 winforms app" scenario should no longer give a full page of errors in the error list anymore.
* The SDK will requires specifying `UseWPF` or `UseWinforms` if you want access to WPF or WinForms APIs. We should still keep this when uptargeting and we should retain them when downtargeting.

jjmew:
@melytc is this completed?

drewnoakes:
One but we haven't addressed is;

> The .NET 5 TFM that should be shown in the property pages for WPF and WinForms apps on .NET Core should be net5.0-windows

I'm not sure whether we need this, or how we would implement it.

melytc:
> I'm not sure whether we need this, or how we would implement it.

I think this was our first way of handling the new platform properties, but we are now explicitly declaring it. 
When creating a new WPF app, we set the TFM to include the `-windows`. For a WinForms app, `-windows7.0` is set.

cc: @KathleenDollard, @sfoslund

sfoslund:
[...]

PR #6607, opened 2020-09-15
Read <SupportedTargetFramework> values from design-time build

davkean:
This is one half of https://github.com/dotnet/project-system/issues/6601 and consumes https://github.com/dotnet/sdk/pull/13545.

This is a replacement of SupportedTargetFrameworksEnumProvider that reads from SupportedTargetFramework from a target instead of evaluation, todo:

- [ ] ~Need two versions of this; one for TargetFramework values (when supporting new format), and one for TargetFrameworkMoniker values (when supporting old format).~
- [x] Need to handle when this target is not present in the SDK (needs to fall back to returning the old SupportedTargetFramework items)
- [x] Perf: We need to initialize in time to catch first design-time build, otherwise, we'll introduce a UI delay showing the page.
- [x] We need a combo on the App pages that is bound to TargetFramework. We can no longer bind to TargetFrameworkMoniker as it cannot represent the OS portion of the TargetFramework, ie `windows` in `net5.0-windows`: https://github.com/dotnet/project-system/pull/6625

//cc @melytc @sfoslund

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6607)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <summary>
+    ///     Responsible for producing valid values for the TargetFramework property from a design-time build.
+    /// </summary>
+    [ExportDynamicEnumValuesProvider("SupportedTargetFrameworksEnumProvider")]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class SupportedTargetFrameworksProvider : ChainedProjectValueDataSourceBase<ICollection<IEnumValue>>, IDynamicEnumValuesProvider, IDynamicEnumValuesGenerator
+    {
+        private readonly IProjectSubscriptionService _subscriptionService;
+
+        public SupportedTargetFrameworksProvider(
```
tmeschter, line 24: Does this need to be tagged with `[ImportingConstructor]`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+            ConfiguredProject project,
+            IProjectSubscriptionService subscriptionService)
+            : base(project, synchronousDisposal: true, registerDataSource: false)
+        {
+            _subscriptionService = subscriptionService;
+        }
+
+        protected override IDisposable? LinkExternalInput(ITargetBlock<TOutput> targetBlock)
```
tmeschter, line 32: Who is the "external input" and who is the "target" here?

I can follow that we're taking data from the `ProjectBuildRuleSource`, extracting the `SupportedTargetFrameworkAlias` items, and transforming them into a collection of `IEnumValue` but I don't have a sense of where that operation fits into the bigger picture.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+        public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>? options)
+        {
+            return Task.FromResult<IDynamicEnumValuesGenerator>(this);
+        }
+
+        public async Task<ICollection<IEnumValue>> GetListedValuesAsync()
+        {
+            using (JoinableCollection.Join())
+            {
+                TOutput snapshot = await SourceBlock.ReceiveAsync();
```
tmeschter, line 82: Does `SourceBlock` here refer to the same thing as the `source` local in `LinkExternalInput`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+using Microsoft.Build.Framework.XamlTypes;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using TOutput = Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<System.Collections.Generic.ICollection<Microsoft.VisualStudio.ProjectSystem.Properties.IEnumValue>>;
```
tmeschter, line 11: Could we call this `EnumValueCollection` instead of `TOutput`? I was really confused the first time I looked at `LinkExternalInput` because I missed this `using` and thought that `TOutput` was an open type parameter rather than a specific type.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
davkean: No, it's "target block" passed to LinkExternalInput. We're grabbing subscription data, transforming into a specific format, and pushing that onto the target block. That block is returned by SourceBlock.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
davkean: If you haven't done much data flow, then this entire chain won't make a lot of sense. We are basically just transforming subscription into IEnumValue, which we end up returning in GetListedValuesAsync.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,106 @@
...
+        private static IEnumValue ToEnumValue(KeyValuePair<string, IImmutableDictionary<string, string>> item)
+        {
+            return new PageEnumValue(new EnumValue()
+            {
+
+                // Example: <SupportedTargetFrameworkAlias  Include=".NETCoreApp, Version=v5.0"
+                //                                          TargetFramework="net5.0-windows"
+                //                                          DisplayName=".NET 5.0" />
+
+                Name = item.Key,
```
davkean, line 81: You would get the wrong value here.

davkean: item.Value[SupportedTargetFrameworkAlias.TargetFrameworkProperty]

drewnoakes (approved):

dmonroym (approved):

melytc (approved):

PR #6618, opened 2020-09-17
Add metadata to filter MEF exports until needed

### LINKED ISSUE #6617: Use [AppliesToProject] on IAttachedCollectionSourceProvider implementations
Visual Studio has recently [added `[AppliesToProject]`](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/271500) support for `IAttachedCollectionSourceProvider` providers, which can be used to prevent unneeded dlls. Project System should use this for all their implementations to avoid unneeded loads of this dll in projects/solutions where it is not applicable, for example:

``` C#
    [Export(typeof(IAttachedCollectionSourceProvider))]
    [Name(nameof(PackageReferenceAttachedCollectionSourceProvider))]
    [Order(Before = HierarchyItemsProviderNames.Contains)]
    [AppliesToProject("DependenciesTree")] 
```

This is required to fix this [performance gate that was broken](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1125457) that got an exception for 16.7.

**Note:** Similar to UI contexts for commands/packages this only controls whether your implementation is loaded for a given solution. You will still be called back for projects that do not contain these capabilities and you will need to continue to handle this.

drewnoakes:
Fixes #6617

Previously the import of `IAttachedCollectionSourceProvider` parts did not filter exports based on metadata. This meant that all exports were loaded, which resulted in DLLs being loaded even when the dependencies tree was unused.

This filtering is now supported by Solution Explorer, and this commit adds the relevant condition to ensure DLLs are not loaded unnecessarily.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6618)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
@@ -10,6 +10,7 @@ using Microsoft.VisualStudio.Utilities;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.Implementation
 {
+    [AppliesTo(ProjectCapability.DependenciesTree)]
     [Export(typeof(IAttachedCollectionSourceProvider))]
     [Name(nameof(FrameworkReferenceAssemblyAttachedCollectionSourceProvider))]
     [VisualStudio.Utilities.Order(Before = HierarchyItemsProviderNames.Contains)]
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
@@ -10,7 +10,7 @@ using Microsoft.VisualStudio.Utilities;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.Implementation
 {
-    [AppliesTo(ProjectCapability.DependenciesTree)]
+    [AppliesToProject(ProjectCapability.DependenciesTree)]
     [Export(typeof(IAttachedCollectionSourceProvider))]
     [Name(nameof(FrameworkReferenceAssemblyAttachedCollectionSourceProvider))]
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
```diff
@@ -10,6 +10,7 @@
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.Implementation
 {
+    [AppliesTo(ProjectCapability.DependenciesTree)]
```
davkean, line 13: We should consistently apply these attributes, we're mixing some from Utilities and some from CPS.

davkean:
@dmonroym Can we watch this insertion? This is supposed to fix a [performance issue](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1125457) so we want to make sure it actually fixes it before merge.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/AttachedCollections/Implementation/FrameworkReferenceAssemblyAttachedCollectionSourceProvider.cs
davkean: They might end up with the same effective metadata, but sometimes they have different metadata for the same name, such as Order.

drewnoakes: What would you like to change specifically? `Microsoft.VisualStudio.Utilities.OrderAttribute` has different metadata than `Microsoft.VisualStudio.ProjectSystem.OrderAttribute`. We use ``Microsoft.VisualStudio.ProjectSystem.AppliesToAttribute` everywhere else, so it would be the odd one out. I can't see how it matters either way.

davkean: It is a concidence that AppliesToAttribute produces to the same metadata as AppliesToProjectAttribute. If AppliesToProjectAttribute was changed to include new metadata, this usage would not be found.

There is no actual impact to the product at this very moment, but it's like mixing attributes between System.Composition and System.ComponentModel.Composition and just getting lucky that both work.

davkean: NameAttribute is the same bucket.

tmeschter (approved):

PR #6624, opened 2020-09-22
Refactor logic to read supported frameworks

davkean:
This refactors the logic to read supported frameworks, and removes the CPS implementation from parsing/understanding the "moniker" being based through from SupportedTargetFramworksEnumProvider.

This is a precursor to enabling us to bind to the "TargetFramework" property in https://github.com/dotnet/project-system/pull/6607.

Before this change;

1) The moniker returned from SupportedTargetFrameworksEnumProvider was being parsed and filtered. This breaks when we change this moniker to a "friendly identifier" in #6607 such as "net5.0-windows".

2) Web/Legacy/CPS filtering was all intertwined. This separates the logic into separate types.

3) This fixes a slight bug where the display name for .NET Framework-only targets in SDK-based projects would show as ".NETFramework,v4.5" if the TFM wasn't installed.

There is not behavior change in all scenarios except for 3).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6624)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/TargetFrameworkMoniker.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/TargetFrameworkMoniker.vb
@@ -50,105 +50,58 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
             Return _displayName
         End Function
 
-        'TODO: Remove this hardcoded list (Refer Bug: #795)
-        Private Shared Function AddDotNetCoreFramework(prgSupportedFrameworks As Array, supportedTargetFrameworksDescriptor As PropertyDescriptor) As Array
-            Dim _TypeConverter As TypeConverter = supportedTargetFrameworksDescriptor.Converter
-            If _TypeConverter IsNot Nothing Then
-                Dim supportedFrameworksList As List(Of String) = New List(Of String)
-                For Each moniker As String In prgSupportedFrameworks
-                    supportedFrameworksList.Add(moniker)
-                Next
-
-                For Each frameworkValue In _TypeConverter.GetStandardValues()
-                    Dim framework = CStr(frameworkValue)
-                    If framework IsNot Nothing Then
-                        supportedFrameworksList.Add(framework)
-                    End If
-                Next
-
-                Return supportedFrameworksList.ToArray
-            End If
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/SupportedTargetFrameworksEnumProvider.cs
```diff
@@ -52,7 +53,10 @@ public Task<ICollection<IEnumValue>> GetListedValuesAsync()
             {
```
sfoslund, line 53: Note that the item group containing the target frameworks will have a different name. `SupportedTargetFramework` will still contain the `TargetFrameworkMoniker`s whereas `SupportedTargetFrameworkAlias` is the corresponding item group with the `TargetFramework`s.

davkean: I think you meant this comment on https://github.com/dotnet/project-system/pull/6607? Yep aware of this of this difference. This PR unblocks being able to pass TargetFramework through.

src/Microsoft.VisualStudio.Editors/PropPages/TargetFrameworkMoniker.vb
```diff
@@ -50,105 +50,57 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+            Return provider.GetSupportedTargetFrameworks(framework)
+
+        End Function
+
+        Private Shared Function GetSupportedTargetFrameworksProvider(
+                frameworkMultiTargeting As IVsFrameworkMultiTargeting,
+                project As Project,
+                converter As TypeConverter) As ISupportedTargetFrameworksProvider
+
+            ' Is this a CPS project?
```
drewnoakes, line 76: Would be useful to explain why only CPS provides a non-null converter, and what it's for.

drewnoakes (approved):

PR #6626, opened 2020-09-23
Version two of project unload dependency telemetry

drewnoakes:
- Change the format to be easier to query
- Include totals at the TFM level
- Include a version number so we can exclude earlier representations

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6626)


dmonroym (approved):

jjmew (approved):
:shipit:

Issue #5139, opened 2019-07-23
Undoing a Roslyn rename prompts a rename


davkean:
**Steps to Reproduce**:

1. File -> New -> Console App (.NET Core)
2. Press F2 to rename Program -> NotProgram

-- observe that Program.cs has been renamed to NotProgram.cs --

3.  In same editor, CTRL+Z to undo rename

**Expected Behavior**:
No prompt to rename

**Actual Behavior**:
![image](https://user-images.githubusercontent.com/1103906/61709604-f217fe80-ad92-11e9-9be4-5dd2ff255a88.png)

**User Impact**:

davkean:
/cc @jmarolf @ryzngard 

Looks like both your bits arrived in master, and the above is the result.

jmarolf:
@davkean 	I cannot reproduce this with the repro steps provided.  Is there anything else that you did other than what you've suggested?

davkean:
Here's a video:

![cfbce04d-3bcf-41e8-a504-084f6f04949d](https://user-images.githubusercontent.com/475144/66791109-a1063b00-eea7-11e9-974e-7b27cbb06561.gif)

davkean:
This is going to because we conflating "renames" via dataflow with user actions.

jmarolf:
@davkean 	I cannot reliably repo this today, can you provide consistent repro steps on a build we've release?

davkean:
I mentioned early on in the rewrite that we shouldn't be using dataflow renames as an indication that the user renamed the type, any use of ProjectItem.Rename will result in this.

jmarolf:
I think I see where I mis-understood @davkean 	when you saith this:

> Press F2 to rename Program -> NotProgram

I thought you mean't F2 in the solution explorer, I assume you actually meant F2 on the class definition, which I can repro

jmarolf:
yeah, well need to re-write everything again...

ocallesp:
@davkean  I will investigate this issue.
This issue is related to https://youtrack.jetbrains.com/issue/RSRP-476479 
e-mail thread with subject [EXTERNAL] Bogus 'You are renaming a file...`

ocallesp:
@jmarolf  @davkean I cannot repro this issue using the steps in the description. Is there any missing step ?

davkean:
@ocallesp The "Rename File" code fix isn't showing up for me anymore, can you see it? Regardless, this will still occur when any component (such as R#) uses APIs to rename files.

[...]

PR #6628, opened 2020-09-23
Enable feature to update the namespace.

ocallesp:
Use IProjectChangeHintReceiver to get event notifications of file renames.

HintingAsync() and HintedAsync() are called before and after the file rename.

HintingAsync() will get the list of actions to update namespace and
HintedAsync() will apply the changes to the corresponding files.

If case CPS fails to rename a file, HintedAsync() will not execute and hence
the actions will be not applied.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6628)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.HandleFileMove.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.HandleFileMove.cs
@@ -1,0 +1,108 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Rename;
+using Microsoft.VisualStudio.ProjectSystem.Waiting;
+using Solution = Microsoft.CodeAnalysis.Solution;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
+{
+    internal partial class RenamerProjectTreeActionHandler
+    {
+        public async Task HintingAsync(IProjectChangeHint hint)
+        {
+            if (!(hint is IProjectChangeFileRenameHint renameHint))
+            {
+                return;
+            }
+
+            // Get the list of files that changed folder
+            Dictionary<string, string> filesMoved = new Dictionary<string, string>();
+
+            foreach (var renamedFile in renameHint.RenamedFiles)
+            {
+                string sourceDirectory = Path.GetDirectoryName(renamedFile.Key);
+                string destinationDirectory = Path.GetDirectoryName(renamedFile.Value);
+                if (sourceDirectory.Equals(destinationDirectory))
+                {
+                    continue;
+                }
+
+                filesMoved[renamedFile.Key] = Path.GetFileName(destinationDirectory);
+            }
+
+            CodeAnalysis.Project? project = GetCurrentProject();
+
+            if (project is null)
+            {
+                return;
+            }
+
+            // Get the list of actions that will update the namespace
+            _results.Clear();
+            foreach (var fileMoved in filesMoved)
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.HandleFileMove.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Rename;
+using Microsoft.VisualStudio.ProjectSystem.Waiting;
+using Solution = Microsoft.CodeAnalysis.Solution;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
+{
+    internal partial class RenamerProjectTreeActionHandler
```
swesonga, line 16: >RenamerProjectTreeActionHandler [](start = 27, length = 31)

Based on this class name, I think the convention would be for this file to be called RenamerProjectTreeActionHandler.HandleFileMove (the prefix matching the class name)

ocallesp: I agree with you, but Undo is also part of this partial class and I just followed that same convention.
Maybe we should rename both to follow RenamerProjectTreeActionHandler.HandlerFileMove

image

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.HandleFileMove.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        public async Task HintingAsync(IProjectChangeHint hint)
+        {
+            if (!(hint is IProjectChangeFileRenameHint renameHint))
+            {
+                return;
+            }
+
+            // Get the list of files that changed folder
+            Dictionary<string, string> filesMoved = new Dictionary<string, string>();
```
swesonga, line 26: >Dictionary<string, string> [](start = 12, length = 26)

nit: using var would be succinct since the type in the assignment is obvious

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.HandleFileMove.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            foreach (var renamedFile in renameHint.RenamedFiles)
+            {
+                string sourceDirectory = Path.GetDirectoryName(renamedFile.Key);
+                string destinationDirectory = Path.GetDirectoryName(renamedFile.Value);
+                if (sourceDirectory.Equals(destinationDirectory))
```
swesonga, line 32: >Equals [](start = 36, length = 6)

I think there should be a string comparison specified here

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/Renamer.HandleFileMove.cs
swesonga: Good point, existing convention is file to use

---
In reply to: [493901852](https://github.com/dotnet/project-system/pull/6628#discussion_r493901852) [](ancestors = 493901852)

dmonroym (approved):

davkean:
As mentioned offline, IProjectChangeHintReceiver & other data flow appraoches we already taken in two previous versions of the renamer. We replaced them with the action handler because it directly related to the user action of renaming the file. Reintroducing this, brings back bugs such as https://github.com/dotnet/project-system/issues/5139.

We should polish off your approach for introducing action handlers for Copy/Move: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/276244.

ocallesp:
Abandon this pr since IProjectChangeHintReceiver is an old deprecated interface that might cause issues with Resharper

PR #6629, opened 2020-09-24
Load implicit services on PackageReference presence

davkean:
We have some implicit PackageReference services that don't trigger initialization for projects without ".NET". After this change, we get loaded for a C++ project with PackageReferences defined.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6629)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
     ///     Provides an implementation of <see cref="IConfiguredProjectActivationTracking"/> that is based on the results of <see cref="IActiveConfigurationGroupService"/>.
     /// </summary>
     [Export(typeof(IConfiguredProjectImplicitActivationTracking))]
-    [AppliesTo(ProjectCapability.DotNet)]
+    [AppliesTo(ProjectCapability.DotNet + " | " + ProjectCapability.PackageReferences)]
```
swesonga, line 18: >ProjectCapability.DotNet + " | " + ProjectCapability.PackageReferences [](start = 15, length = 70)

can this be a string constant in the class to reduce repetition?

swesonga (approved):
:shipit:

tmeschter (approved):

melytc (approved):

melytc:
ComponentVerificationTests.cs fails at the `CertainExportsMustNotBeMarkedWithDynamicCapabilities` test, with: 

**Microsoft.VisualStudio.ProjectSystem.ConfiguredProjectImplicitActivationTracking** exports **Microsoft.VisualStudio.ProjectSystem.ConfiguredProject.AutoLoad** based on a dynamic capabilities 'PackageReferences'. This contract is used during project initialization and must use a capability that doesn't change over the lifetime of a project. These capabilities are specified in src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\Packaging\ProjectTypeRegistration.cs

davkean:
Yep I need a different approach.

davkean:
@dotnet/project-system Please take another look, this looks very different.

I time boxed writing unit tests for ~3 hours and gave up. It's close to impossible to mock out the right dependencies here given the capabilities usage, these sorts of components we need to end up testing with a real live project, which needs CPS implementation.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -1,147 +1,52 @@
...
-using System.Threading;
-using System.Threading.Tasks;
-using System.Threading.Tasks.Dataflow;
-using Microsoft.VisualStudio.Threading.Tasks;
 
 namespace Microsoft.VisualStudio.ProjectSystem
 {
     /// <summary>
-    ///     Provides an implementation of <see cref="IConfiguredProjectActivationTracking"/> that is based on the results of <see cref="IActiveConfigurationGroupService"/>.
+    ///     Responsible for activating and deactiving <see cref="IImplicitlyActiveService"/> instances.
```
swesonga, line 8: >deactiving [](start = 43, length = 10)

nit: deactivating

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -1,147 +1,52 @@
...
-    [AppliesTo(ProjectCapability.DotNet)]
-    internal class ConfiguredProjectImplicitActivationTracking : OnceInitializedOnceDisposed, IConfiguredProjectImplicitActivationTracking
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
+    [AppliesTo(LoadCapabilities)]
+    internal partial class ConfiguredProjectImplicitActivationTracking : AbstractMultiLifetimeComponent<ConfiguredProjectImplicitActivationTracking.ConfiguredProjectImplicitActivationTrackingInstance>, IProjectDynamicLoadComponent
     {
+        // NOTE: Ideally this component would be marked with 'AlwaysApplicable' so that we always load
+        // IImplicitlyActiveService instances in all project types regardless of exported capabilities,
+        // but doing so would cause us to be loaded in lots of situations even when not needed. Instead,
+        // we explicitly we current hardcode the set of capabilities of all our IImplicitlyActiveService
```
swesonga, line 17: >explicitly we current hardcode [](start = 14, length = 30)

nit: phrasing

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -1,147 +1,52 @@
...
     /// </summary>
-    [Export(typeof(IConfiguredProjectImplicitActivationTracking))]
-    [AppliesTo(ProjectCapability.DotNet)]
-    internal class ConfiguredProjectImplicitActivationTracking : OnceInitializedOnceDisposed, IConfiguredProjectImplicitActivationTracking
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
+    [AppliesTo(LoadCapabilities)]
+    internal partial class ConfiguredProjectImplicitActivationTracking : AbstractMultiLifetimeComponent<ConfiguredProjectImplicitActivationTracking.ConfiguredProjectImplicitActivationTrackingInstance>, IProjectDynamicLoadComponent
     {
+        // NOTE: Ideally this component would be marked with 'AlwaysApplicable' so that we always load
+        // IImplicitlyActiveService instances in all project types regardless of exported capabilities,
```
swesonga, line 15: It's not clear to me how this would be desirable since it's what we are trying to avoid doing. Is the concern about loading IImplicitlyActiveServices that we do not own?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -55,45 +57,25 @@ protected override void Dispose(bool disposing)
...
 
-            if (nowActive)
-            {
-                return OnImplicitlyActivated();
-            }
-            else if (_isActive)
-            {
-                return OnImplicitlyDeactivated();
-            }
+            // Deactive currently "active" services that no longer applicable
```
drewnoakes:
```suggestion
            // Deactive currently "active" services that are no longer applicable
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -55,45 +57,25 @@ protected override void Dispose(bool disposing)
...
-        }
-
-        private Task OnImplicitlyDeactivated()
-        {
-            _isActive = false;
-
-            IEnumerable<Task> tasks = ImplicitlyActiveServices.Select(c => c.Value.DeactivateAsync());
+            // Activate "non-active" services that are now applicable
+            IEnumerable<IImplicitlyActiveService> servicesToActivate = futureServices.Except(_activeServices);
+            await Task.WhenAll(servicesToActivate.Select(c => c.ActivateAsync()));
```
drewnoakes: Can the activations here be made concurrent with the deactivations above? If it's safe to do so, it may complete more quickly.

drewnoakes: The [`SetDiff`](https://github.com/dotnet/project-system/blob/3b9c57cbb15a84077b6c51c1e2df2043cd9218ee/src/Microsoft.VisualStudio.ProjectSystem.Managed/Utilities/SetDiff.cs) class computes the symmetric diff in one pass.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
```diff
@@ -1,81 +1,52 @@
...
     ///     Responsible for activating and deactiving <see cref="IImplicitlyActiveService"/> instances.
     /// </summary>
-    internal class ConfiguredProjectImplicitActivationTracking : OnceInitializedOnceDisposed
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
+    [AppliesTo(LoadCapabilities)]
+    internal partial class ConfiguredProjectImplicitActivationTracking : AbstractMultiLifetimeComponent<ConfiguredProjectImplicitActivationTracking.ConfiguredProjectImplicitActivationTrackingInstance>, IProjectDynamicLoadComponent
     {
+        // NOTE: Ideally this component would be marked with 'AlwaysApplicable' so that we always load
+        // IImplicitlyActiveService instances in all project types regardless of exported capabilities,
+        // but doing so would cause us to be loaded in lots of situations even when not needed. Instead,
```
drewnoakes, line 16:
```suggestion
        // but doing so would cause the .NET Project System's assemblies to be loaded in lots of situations even when not needed. Instead,
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.Instance.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                _subscription?.Dispose();
+                _targetBlock.Complete();
+
+                return DeactivateAsync(_activeServices);
+            }
+
+            internal async Task OnActiveConfigurationsChanged(IProjectVersionedValue<ValueTuple<IProjectCapabilitiesSnapshot, IConfigurationGroup<ProjectConfiguration>>> e)
+            {
+                // We'll get called back in main two situations (not withstanding version-only updates):
```
drewnoakes, line 66:
```suggestion
                // We'll get called back in main two situations (notwithstanding version-only updates):
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.Instance.cs
```diff
@@ -0,0 +1,107 @@
...
+                IReadOnlyCollection<IImplicitlyActiveService> futureServices = isActive ? _implicitlyActiveServices.Select(s => s.Value).ToList()
+                                                                                        : Array.Empty<IImplicitlyActiveService>();
+
+                // Deactive currently "active" services that no longer applicable
+                IEnumerable<IImplicitlyActiveService> servicesToDeactivate = _activeServices.Except(futureServices);
+                await DeactivateAsync(servicesToDeactivate);
+
+                // Activate "non-active" services that are now applicable
+                IEnumerable<IImplicitlyActiveService> servicesToActivate = futureServices.Except(_activeServices);
+                await ActivateAsync(servicesToActivate);
```
drewnoakes, line 91: Ditto concurrency and `SetDiff`.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentComposition.cs
```diff
@@ -17,7 +17,7 @@ internal partial class ComponentComposition
         /// </summary>
         internal static readonly IReadOnlyList<Assembly> BuiltInAssemblies = new Assembly[]
         {
-            typeof(IConfiguredProjectImplicitActivationTracking).Assembly,  // Microsoft.VisualStudio.ProjectSystem.Managed
+            typeof(ConfiguredProjectImplicitActivationTracking).Assembly,  // Microsoft.VisualStudio.ProjectSystem.Managed
```
drewnoakes, line 20:
```suggestion
            typeof(ConfiguredProjectImplicitActivationTracking).Assembly,   // Microsoft.VisualStudio.ProjectSystem.Managed
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/ComponentComposition.cs
```diff
@@ -28,7 +28,7 @@ internal partial class ComponentComposition
         {
             typeof(IProjectService).Assembly,                               // Microsoft.VisualStudio.ProjectSystem
             typeof(IVsProjectServices).Assembly,                            // Microsoft.VisualStudio.ProjectSystem.VS
-            typeof(IConfiguredProjectImplicitActivationTracking).Assembly,  // Microsoft.VisualStudio.ProjectSystem.Managed
+            typeof(ConfiguredProjectImplicitActivationTracking).Assembly,  // Microsoft.VisualStudio.ProjectSystem.Managed
```
drewnoakes, line 31:
```suggestion
            typeof(ConfiguredProjectImplicitActivationTracking).Assembly,   // Microsoft.VisualStudio.ProjectSystem.Managed
```

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
davkean: It would prevent the need to have LoadCapabilities match every single capability that every implementation of IImplicitlyActiveServices uses.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.cs
davkean: In theory, yes, however, this will result in an overlap of Roslyn projects for language service integration for the same project, which is untested and might result in more OOMs - especially in multi-targeted projects. I tried to maintain as much behavior as I could at this moment, and will hold off on that.

PR #6631, opened 2020-09-28
Allow the user to disable the feature symbolic renaming on File-rename

ocallesp:
This fixes [AB#1144262](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1144262)

It seems that this option is misleading since we are enabling the feature unconditionally. People might get confused when trying to disable the prompt and still seeing changes in your code.

This change replaces the dialog box to allow the option "Don't Ask Again" so users can disable the prompt message from here.
CheckUserConfirmation() now returns false and that means the feature is disabled by default unless there is a prompt and the user selects Yes.

This change will allow the user to disable this feature when the message prompt is disabled.

Perhaps a future enhancement is to add another checkbox in Tools |Options | Project and Solutions | General | "Always apply symbol rename" so this will allow the user to apply symbol rename without the prompt.

image

https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/278294
image

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DteEnvironmentOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DteEnvironmentOptions.cs
@@ -32,5 +32,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
             return defaultValue;
         }
+
+        public void SetOption<T>(string category, string page, string option, T newValue)
+        {
+            EnvDTE.Properties? properties = _dte.Value.Properties[category, page];
+
+            if (properties != null)
+            {
+                properties.Item(option).Value = newValue;
+            }
+        }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
@@ -28,6 +28,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         /// </param>
         T GetOption<T>(string category, string page, string option, T defaultValue);
 
+        /// <summary>
+        ///     Sets a value associated with the specified category, page and option, if it exists,
+        /// </summary>
+        /// <param name="category">
+        ///     A <see cref="string"/> containing the category of the option to return.
+        /// </param>
+        /// <param name="page">
+        ///     A <see cref="string"/> containing the page of the option to return.
+        /// </param>
+        /// <param name="option">
+        ///     A <see cref="string"/> containing the name of the option to return.
+        /// </param>
+        /// <param name="newValue">
+        ///     The value to set the property.
+        /// </param>
         void SetOption<T>(string category, string page, string option, T newValue);
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
@@ -27,5 +27,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         ///     The value to return if the value does not exist.
         /// </param>
         T GetOption<T>(string category, string page, string option, T defaultValue);
+
+        void SetOption<T>(string category, string page, string option, T newValue);
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
@@ -32,13 +32,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         ///     Sets a value associated with the specified category, page and option, if it exists,
         /// </summary>
         /// <param name="category">
-        ///     A <see cref="string"/> containing the category of the option to return.
+        ///     A <see cref="string"/> containing the category of the option to set.
         /// </param>
         /// <param name="page">
-        ///     A <see cref="string"/> containing the page of the option to return.
+        ///     A <see cref="string"/> containing the page of the option to set.
         /// </param>
         /// <param name="option">
-        ///     A <see cref="string"/> containing the name of the option to return.
+        ///     A <see cref="string"/> containing the name of the option to set.
         /// </param>
         /// <param name="newValue">
         ///     The value to set the property.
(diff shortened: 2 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/DteEnvironmentOptions.cs
```diff
@@ -32,5 +32,15 @@ public T GetOption<T>(string category, string page, string option, T defaultValu
 
             return defaultValue;
         }
+
+        public void SetOption<T>(string category, string page, string option, T newValue)
```
ocallesp, line 36: Comments needed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
```diff
@@ -27,5 +27,7 @@ internal interface IEnvironmentOptions
         ///     The value to return if the value does not exist.
         /// </param>
         T GetOption<T>(string category, string page, string option, T defaultValue);
+
+        void SetOption<T>(string category, string page, string option, T newValue);
```
ocallesp, line 31: Comments needed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
```diff
@@ -230,15 +231,23 @@ protected virtual async Task<bool> IsAutomationFunctionAsync()
...
             await _projectVsServices.ThreadingService.SwitchToUIThread();
+
+            bool disablePromptMessage = false;
             bool userNeedPrompt = _environmentOptions.GetOption("Environment", "ProjectsAndSolution", "PromptForRenameSymbol", false);
             if (userNeedPrompt)
             {
                 string renamePromptMessage = string.Format(CultureInfo.CurrentCulture, VSResources.RenameSymbolPrompt, oldFileName);
+                string dontShowAgainMessage = string.Format(CultureInfo.CurrentCulture, VSResources.DontShowAgain);
+
+                var result = MessageDialog.Show("Microsoft Visual Studio", renamePromptMessage, MessageDialogCommandSet.YesNo, dontShowAgainMessage,
```
tmeschter, line 242: I would prefer you add a new method on `IUserNotificationServices` to handle this, and move the call to `MessageDialog.Show` to the implementation of that new method.

tmeschter (changes requested):
I have two high-level comments on this:

1. I don't expect `MessageDialog.Show` to work properly in Codespaces. However, we haven't replaced our other dialogs with Codespaces-friendly equivalents yet, so I don't want to hold things up for that. We do, however, need to make sure that with this change Visual Studio doesn't crash or hang when a file rename happens in a Codespaces scenario.
2. The big problem I see is that this changes the meaning of the setting. Previously, if the setting were turned off then the user wouldn't see a prompt, and the rename would just happen automatically. Now, if the setting is turned off the user still won't see a prompt--but the rename won't happen either. Also, if the dialog comes up and they check the "Don't show again" box and click "Yes" to allow the rename then they probably expect that future renames will occur automatically without a prompt--but that isn't what will happen.

I think what we really need here is to give the user three options for automatic renames: "Yes", "No", and "Prompt".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
ocallesp: comments added

davkean:
Rename is already turned off for codespaces.

I'm in agreement here with Tom, the new dialog implies an option with 3 states.

cartermp:
Considering the options page:

This feels very clearly like a nested option scenario.

* Enable symbolic renaming when renaming files
   * Show prompt before symbolic rename when renaming files

Both options are selected by default.

If the top-level option is unchecked, we do not do symbolic rename.

If the top-level option is checked but the nested one is unchecked, we do not prompt.

If a user de-selected the current option, then it is de-selected for them here (so as not to wipe out the previous settings choice).

Considering the dialog:

If they check "don't show this again" and press "No", then the top-level option is deselected. Symbolic rename doesn't happen anymore.

If they check "don't show this again" and press "Yes", then the top-lvel option remains selected, but the nested option is deselected.

Does that make sense?

ocallesp:
> **Considering the options page:**
> 
> This feels very clearly like a nested option scenario.
> 
> * Enable symbolic renaming when renaming files
>   
>   * Show prompt before symbolic rename when renaming files
> 
> Both options are selected by default.
> 
> If the top-level option is unchecked, we do not do symbolic rename.
> 
> If the top-level option is checked but the nested one is unchecked, we do not prompt.
> 
> If a user de-selected the current option, then it is de-selected for them here (so as not to wipe out the previous settings choice).
> 
> **Considering the dialog:**
> 
> If they check "don't show this again" and press "No", then the top-level option is deselected. Symbolic rename doesn't happen anymore.
> 
> If they check "don't show this again" and press "Yes", then the top-lvel option remains selected, but the nested option is deselected.
> 
> Does that make sense? Does that impact legacy?

I changed the UX as suggested.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IEnvironmentOptions.cs
```diff
@@ -27,5 +27,22 @@ internal interface IEnvironmentOptions
         ///     The value to return if the value does not exist.
         /// </param>
         T GetOption<T>(string category, string page, string option, T defaultValue);
+
+        /// <summary>
+        ///     Sets a value associated with the specified category, page and option, if it exists,
+        /// </summary>
+        /// <param name="category">
+        ///     A <see cref="string"/> containing the category of the option to return.
```
tmeschter, line 35: I think you mean "the option to set" rather than "the option to return". Also affects the next two parameters.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IUserNotificationServices.cs
```diff
@@ -35,5 +35,16 @@ internal interface IUserNotificationServices
...
+        /// <summary>
+        ///     Shows a Yes/No confirmation with Don't Show Again checkbox
+        /// </summary>
+        /// <param name="message">The message related to the Yes/No question</param>
+        /// <param name="disablePromptMessage">checkbox selection</param>
+        /// <returns>
+        ///     <see langword="true"/> if the user clicked the Yes button, otherwise;
+        ///     <see langword="false"/> if the user clicked the No button.
+        /// </returns>
+        bool Show(string message, out bool disablePromptMessage);
```
tmeschter, line 48: I prefer "Confirm" over "Show" for consistency with the first method in this type.

tmeschter (approved):
Approved with some minor feedback items.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IUserNotificationServices.cs
ocallesp: done

softworkz:
Well done, great job! Thanks for getting this right.

PR #6625, opened 2020-09-22
Populate TFM dropdown with new supported list.

melytc:
This PR is part of a collective work to show the supported target frameworks in the Properties Pages, where in the case of WPF or WinForms, `.NET 5` will be saved as `net5.0-windows`.

In the Properties pages, we are now binding the `TargetFramework` property name during its class initialization to be able to know if it's a legacy project or not.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6625)

PR #6643, opened 2020-10-01
Read <SupportedTargetFramework> values from design-time build

melytc:
Re-opening this pr from @davkean, since it was merged by accident into master. This one targets 16.8 and should be merged with #6625. 

-----
This is one half of https://github.com/dotnet/project-system/issues/6601 and consumes https://github.com/dotnet/sdk/pull/13545.

This is a replacement of SupportedTargetFrameworksEnumProvider that reads from SupportedTargetFramework from a target instead of evaluation, todo:

- [ ] ~Need two versions of this; one for TargetFramework values (when supporting new format), and one for TargetFrameworkMoniker values (when supporting old format).~
- [x] Need to handle when this target is not present in the SDK (needs to fall back to returning the old SupportedTargetFramework items)
- [x] Perf: We need to initialize in time to catch first design-time build, otherwise, we'll introduce a UI delay showing the page.
- [x] We need a combo on the App pages that is bound to TargetFramework. We can no longer bind to TargetFrameworkMoniker as it cannot represent the OS portion of the TargetFramework, ie `windows` in `net5.0-windows`: https://github.com/dotnet/project-system/pull/6625

//cc @melytc @sfoslund 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6643)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>? options)
+        {
+            return Task.FromResult<IDynamicEnumValuesGenerator>(this);
+        }
+
+        public async Task<EnumCollection> GetListedValuesAsync()
+        {
+            using (JoinableCollection.Join())
+            {
+                EnumCollectionProjectValue snapshot = await SourceBlock.ReceiveAsync();
```
tmeschter, line 94: When this change was briefly merged into master, I saw tasks getting stuck here with `await SourceBlock.ReceiveAsync()` never completing. Do you have any idea why this might not complete, and do we need to do anything further to guarantee that it does?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                return snapshot.Value;
+            }
+        }
+
+        bool IDynamicEnumValuesGenerator.AllowCustomValues => false;
+
+        Task<IEnumValue?> IDynamicEnumValuesGenerator.TryCreateEnumValueAsync(string userSuppliedValue) => throw new NotSupportedException();
```
tmeschter, line 102: This needs to throw a `NotImplementedException`--that is the one (and only) exception type CPS understands when creating new enum values is not supported.

tmeschter (changes requested):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
davkean: This will block if the configuration is not active or one of the implicit active configurations because they are not enabled to build. It's impossible to produce data for that, either we block, throw, or return nothing.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        protected override IDisposable? LinkExternalInput(ITargetBlock<EnumCollectionProjectValue> targetBlock)
+        {
+            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _subscriptionService.ProjectBuildRuleSource;
+
+            // Transform the changes from design-time build -> Supported target frameworks
+            DisposableValue<ISourceBlock<EnumCollectionProjectValue>> transformBlock = source.SourceBlock.TransformWithNoDelta(
+                update => update.Derive(Transform),
+                suppressVersionOnlyUpdates: true,
```
lifengl, line 53: setting suppressVersionOnlyUpdates to true is almost always wrong other than linking to an end action block. It is usually will lead into dead locks.

davkean: This is a self-contained block (no one else waits on it other than itself) - are saying if this gets sync'linked with something else?

lifengl: Because this class implements IProjectValueDataSource<EnumCollection>, and it is hard to say that no other code chain to it in the future by reading the code alone.  Yeah, if it is not chained to SyncLink,  no code call GetLatestAsync/GetSpecificVersionAsync to the endpoint, or need to report to the operation progress feature, suppressVersionOnlyUpdates is ok to use.

jjmew: @davkean and @lifengl Can we get a resolution on this today on the CPS meeting? Right now 16.8 is broken and this is last comment we need resolved to fix the experience. We need this ASAP.

davkean:
@melytc This is ready for merge from my side.

jjmew (approved):
:shipit:

jjmew:
@melytc can we go ahead and merge this? hopefully we can get an insertion ready for QB approval

melytc:
Pulled David's changes to the combined PR (#6625), will close this one and merge it over there.

Issue #6647, opened 2020-10-02
UpToDate "set of project items was changed more recently" infinite loop (DLL stays the same, LastItemsChangedAtUtc stays the same)


tompazourek:
**Visual Studio Version**: 16.7.3

**Summary**: I have a very large solution and some projects can somehow get into this weird state where they get built over and over even when they don't have any changes. For example, let's say `MyProject` is stuck in this state. I build it and get this log message:

```
FastUpToDate: The set of project items was changed more recently (2020-10-02 13:01:00) than the earliest output 'xxx\obj\Debug\MyProject.dll' (2020-10-02 13:00:58), not up to date. (MyProject)
```

Then Visual Studio builds the project, but the build doesn't actually end up doing anything, it only copies the already existing DLL that has the same `2020-10-02 13:00:58` timestamp. So next time I build it, the DLL timestamp is still the same, it'll still be older than the timestamp `2020-10-02 13:01:00` that's remembered by Visual Studio.

The log message comes from:

https://github.com/dotnet/project-system/blob/519340c5d37ab96aaddadb5eca93a5955831a869/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs#L242

And by reading the code, I just think that this process is faulty. The `LastItemsChangedAtUtc` never gets updated, the project gets built, but the DLL is only copied, so the DLL timestamp never gets updated, and I get stuck in a loop. I don't know exactly how it is that the `LastItemsChangedAtUtc` timestamp is 2 seconds after the timestamp of the DLL, but I guess it can happen in some way (maybe the items are loaded with some delay if the solution is big? maybe something caused some "reload"?). Anyway, there needs to be some circuit breaker to get out of this endless loop. Both timestamps stay the same forever, and the build just happens over and over. **Maybe a solution would be that the `LastItemsChangedAtUtc` is somehow reset when the project starts building?**

**Steps to Reproduce**: Unfortunately, I don't have a reliable way of reproducing this, it doesn't happen every time. On my big solution, it happens often, but for different projects each time, and I don't exactly know what determines that. But it's still a situation with those 2 timestamps that can happen, and there should be a way how to break out of the loop.

Some notes regarding reproducing this: I didn't modify any code files, only opened a solution, clicked Build, and then watched how some projects get stuck with those timestamps. I have automated restore of NuGet packages enabled (maybe it's somehow related to the strange `LastItemsChangedAtUtc` value).

[...]

PR #6660, opened 2020-10-08
Up to date check logs changed item info

drewnoakes:
This additional logging info will help debug certain situations such as that described in #6647.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6660)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.State.cs
@@ -28,6 +28,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             public string? NewestImportInput { get; }
 
             public IComparable? LastVersionSeen { get; }
+
             public bool IsDisabled { get; }
 
             /// <summary>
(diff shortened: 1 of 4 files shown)


davkean:
Is this persisted between calls or just thrown away? ie Does this increase the footprint of this class?

drewnoakes:
Thrown away and recreated on project load.

And yes there is overhead in keeping all these items in memory. However the strings are unchanged, so we're only paying for a densely packed array of:

```c#
(bool isAdd, string ItemType, string Path, string? Link, CopyType)
```

Rearranging this value tuple or making `CopyType : byte` may make it pack more efficiently.

davkean:
Don't we keep the removed items? ie We're holding onto strings that would have otherwise been collected? Can we only capture this data if logging is on?

drewnoakes:
> Don't we keep the removed items? ie We're holding onto strings that would have otherwise been collected?

How much of an issue is this in reality? It's rare to remove a large number of items, and those strings would be released upon the next change to the project's set of items anyway.

> Can we only capture this data if logging is on?

The logging setting can be changed dynamically. If we don't retain this data until logging is enabled, then when it's turned on we won't have the "before" data to diff against, which would appear as a bug.

drewnoakes:
Ping @davkean

davkean (approved):

PR #6662, opened 2020-10-08
Return null rather than throw

drewnoakes:
A null return value is equivalent to throwing NotImplementedException but is less disruptive during debugging.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6662)


davkean (approved):

tmeschter (approved):

drewnoakes:
Spanish leg seems to have an issue unrelated to this PR. Merging.

PR #6670, opened 2020-10-12
Add search terms support

tmeschter:
Update our Project Query API implementation to return search terms from .xaml `Rule` files.

The new property page UI will support searching across the properties, and we expect that the terms searched for will generally appear in the name, description, or other displayable text associated with the property. However, there may be times we want to associate a search term with a property even when it does _not_ appear in the displayable text. To support this, the `IUIProperty` type in the Project Query API includes a `SearchTerms` member that returns a list of strings.

This change populates the `SeachTerms` by examining the property's metadata for an item named "SearchTerms". If found the value is split on semi-colons into individual terms and returned.

Note that this is something of a temporary approach. This kind of metadata does not appear to be localizable, and so we will not be able to provide localizable search terms. This is acceptable for the time being as the display text is localizable, and we don't expect to make significant use of the search terms. I see two options we can pursue to support localizable search terms in the future:

1. Update our localization tooling to understand that this particular piece of metadata is localizable.
2. Make search terms a first-class property of the `BaseProperty`, and then update the localization tooling to understand that it is localizable.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6670)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
@@ -19,8 +19,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
     /// </summary>
     internal static class UIPropertyDataProducer
     {
-        private static readonly char[] s_searchTermsSeparators = new[] { ';' };
-
         public static IEntityValue CreateUIPropertyValue(IEntityValue parent, IPropertyPageQueryCache cache, BaseProperty property, int order, IUIPropertyPropertiesAvailableStatus requestedProperties)
         {
             Requires.NotNull(parent, nameof(parent));
@@ -92,14 +90,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 
             if (requestedProperties.SearchTerms)
             {
-                string? searchTermsString = property.GetMetadataValue("SearchTerms");
+                string? searchTermsString = property.GetMetadataValueOrNull("SearchTerms");
                 if (searchTermsString is null)
                 {
                     newUIProperty.SearchTerms = ImmutableList<string>.Empty;
                 }
                 else
                 {
-                    newUIProperty.SearchTerms = searchTermsString.Split(s_searchTermsSeparators, StringSplitOptions.RemoveEmptyEntries).ToImmutableList();
+                    newUIProperty.SearchTerms = searchTermsString.Split(Delimiter.Semicolon, StringSplitOptions.RemoveEmptyEntries).ToImmutableList();
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
```diff
@@ -18,6 +19,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
     /// </summary>
     internal static class UIPropertyDataProducer
     {
+        private static readonly char[] s_searchTermsSeparators = new[] { ';' };
```
davkean, line 22: Delimiter.Semicon

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/BasePropertyExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        ///     The value of the corresponding metadata item, or <see langword="null"/> if it is not
+        ///     found.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        ///     <paramref name="property"/> is <see langword="null"/>.
+        /// </exception>
+        /// <exception cref="ArgumentException">
+        ///     <paramref name="property"/> is <see langword="null"/> or empty.
+        /// </exception>
+        public static string? GetMetadataValue(this BaseProperty property, string metadataName)
```
davkean, line 29: GetMetadataValueOrNull. Metadata methods almost everywhere else return empty string to match MSBuild.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
```diff
@@ -89,8 +92,15 @@ public static IEntityValue CreateUIPropertyValue(IEntityRuntimeModel runtimeMode
...
-                newUIProperty.SearchTerms = ImmutableList<string>.Empty;
+                string? searchTermsString = property.GetMetadataValue("SearchTerms");
+                if (searchTermsString is null)
+                {
+                    newUIProperty.SearchTerms = ImmutableList<string>.Empty;
+                }
+                else
+                {
+                    newUIProperty.SearchTerms = searchTermsString.Split(s_searchTermsSeparators, StringSplitOptions.RemoveEmptyEntries).ToImmutableList();
+                }
```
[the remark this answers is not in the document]
tmeschter: I don't have a strong opinion here. If a single string is a better representation then I'll put out a PR to change it in the Project Query API and we'll see how the CPS team feels about it.

As for `ImmutableArray<T>` vs `ImmutableList<string>`, we're limited by what the Project Query API generator produces for us.

jjmew (approved):
:shipit:

PR #6671, opened 2020-10-13
Map from flavors -> capabilities

davkean:
This maps from flavor project type GUIDs -> capabilities, so that we can tie specific components that light up ASP.NET full framework support behind a capability. Also adds an incomplete implementation of IVsAggregatableProject for components that sniff support based on the GUIDs it returns.

This currently maps:

Web Application Projects (not a complete list of them) -> "AspNet"
Test Projects (Legacy) -> "TestContainer"

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6671)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeCapabilitiesProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeCapabilitiesProvider.cs
@@ -1,0 +1,63 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Linq;
+using System.Threading.Tasks.Dataflow;
+
+using ProjectCapabilitiesProjectValue = Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<Microsoft.VisualStudio.ProjectSystem.IProjectCapabilitiesSnapshot>;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Web
+{
+    /// <summary>
+    ///     Produces capabilities from project type GUIDs.
+    /// </summary>
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectCapabilitiesProvider))]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class ProjectTypeCapabilitiesProvider : ChainedProjectValueDataSourceBase<IProjectCapabilitiesSnapshot>, IProjectCapabilitiesProvider
+    {
+        private readonly ProjectTypeGuidsDataSource _dataSource;
+
+        [ImportingConstructor]
+        protected ProjectTypeCapabilitiesProvider(ConfiguredProject configuredProject, ProjectTypeGuidsDataSource dataSource)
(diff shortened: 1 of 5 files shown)


davkean:
tag @BillHiebert

davkean:
Here's an example of how it worked. 

We map Test Flavor -> "TestContainer" capability which causes a component in Unit Test tree to set the icon:

Before:
image

After:

image

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeGuidsDataSource.cs
```diff
@@ -0,0 +1,108 @@
...
+            block.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+
+            return block;
+        }
+
+        private IEnumerable<GuidCollectionProjectValue> CreateSnapshot(IProjectVersionedValue<IProjectSnapshot> snapshot)
+        {
+            string value = snapshot.Value.ProjectInstance.GetPropertyValue(ConfigurationGeneral.ProjectTypeGuidsProperty);
+
+            if (StringComparers.PropertyValues.Equals(value, _lastValue))
```
BillHiebert, line 64: I guess this check is here, because this function creates a difference between the current and the previous snapshots?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeCapabilitiesProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        private IProjectCapabilitiesSnapshot CreateCapabilitiesSnapshot(IImmutableList<Guid> projectTypes)
+        {
+            ImmutableHashSet<string> capabilities = Empty.CapabilitiesSet;
+
+            // TODO: Maybe pull this from a specific item type, or metadata on a capability?
+            if (projectTypes.Contains(ProjectType.LegacyWebApplicationProjectGuid))
+            {
+                capabilities = capabilities.Add("AspNet");
```
BillHiebert, line 50: Just a general question. where in the load sequence are capability providers loaded? Do all components which are associated with these capabililties need to be DyanamicLoad components because the capability can come and go? Or are they considered stable? I would really prefer the latter.

davkean: These capabilities are not fixed, they are dynamic. This is very deliberate - this will mean that a project behaves identically between switching to/from a branch with `<ProjectTypeGuids/>` property and loading a project for the first time with a `<ProjectTypeGuids/>`. Otherwise, they will behave differently. If components/feature have too much problems handling this downstream, then we can cause a reload of the project based on `ProjectTypeGuids` changing, but I'd like that as a last resort.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeGuidsDataSource.cs
davkean: This was to avoid unnecessarily posting an update on no changes, but this ended up causing the hang because downstream a component was sync-linking the result of ProjectTypeCapabilitiesProvider and wasn't seeing version-only changes.

PR #6676, opened 2020-10-14
Enable high contrast for Debug page

melytc:
The Debug page was not being responsive to high contrast changes. This resulted in an accessibility bug reported here [AzDO#1172163](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1172163).

The Debug page is a special case in the Properties pages: it's a WPF control inside the WinForms window. The approach taken was to use the `SetResourceReference` to set the `BackgroundProperty` to `SystemColors.ControlBrushKey` value. In that way, we pass down the [ControlBrushKey](https://docs.microsoft.com/en-us/dotnet/api/system.windows.systemcolors.controlbrushkey?view=netcore-3.1#System_Windows_SystemColors_ControlBrushKey) dynamically.

**Before:**
image

**After:**
image

There are rough edges when toggling back and forth the high contrast setting, but I think this is a good start.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6676)


drewnoakes (approved):

jjmew (approved):
:shipit:

jjmew:
@melytc we need this in 16.8

melytc:
Closing and opening #6678, targeting 16.8.

PR #6681, opened 2020-10-14
Add support for updating properties via the Project Query API

tmeschter:
In addition to supporting querying for data, the Project Query API also specifies a set of actions you can take on various entities that are part of the API. In [PR 277099](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/277099) we defined two new actions that work on projects: setting the "evaluated" value of one of the project's properties, and setting the "unevaluated" value. The differences between the two are relatively small and not particularly interesting, so I'm not going to go into detail on those. Rather, I will focus on the code shared by both.

The `ProjectActionProvider` type plugs our implementation of these actions into the larger Project Query API system. The attributes on this type specify two critical pieces of information: which actions are supported (in this case `SetEvaluatedUIPropertyValue` and `SetUnevaluatedUIPropertyValue`) and which action _targets_ are supported (in this case, the "project" entity indicated by `ProjectType.TypeName`). Most target entities will support multiple actions, and certain actions may be able to target multiple entities. For example, a "delete" action could apply to a file, a reference, or an entire project. The attributes allow us to specify which combination of targets and actions are handled by a particular provider.

An `IQueryActionProvider` such as `ProjectActionProvider` is responsible for returning an `IQueryActionExecutor` for a particular step of a query via the `CreateQueryActionDataTransformer` method. Here we create either a `ProjectSetEvaluatedUIPropertyValueAction` or a `ProjectSetUnevaluatedUIPropertyValueAction` instance. All of the interesting logic is in their shared base type, `ProjectSetUIPropertyValueActionBase<T>`. The entire purpose of this type is to extract interesting information from the Project Query API objects passed in, and the processing of this data is delegated to `ProjectSetUIPropertyValueActionCore`. This division means each type has fewer responsibilities, but the main purpose of the separation is to make it easier to test the core logic of updating property values without the tests having to deal with the Project Query API types.

Executing actions within a query takes place in three stages, corresponding to the methods in `ProjectSetUIPropertyValueActionCore`. The `OnBeforeExecutingBatchAsync` method is called once before the "body" of the query executes, and provides us a chance to do some setup. In this case, we retrieve and cache all of the `IRule`s representing the property pages we are updating across all relevant configurations of all targets projects. Once the body of the query starts the execute, the Project Query API will be holding a lock on the project and we will no longer be able to retrieve these.

The `ExecuteAsync` method is called once for each targeted project. We pull out all of the relevant `IRule`s from the cache, and apply the property update to each in turn.

The `OnAfterExecutingBatch` method is called once at the end of the query execution and provides an opportunity for clean-up. Here we just clean out the cache. This is probably unnecessary as we don't expect this instance to be re-used for other query executions, but this ensures we don't hold on to configuration-specific data longer than necessary.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6681)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/IPropertyPageQueryCacheProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// Creates instances of <see cref="IPropertyPageQueryCache"/>.
+    /// </summary>
+    /// <remarks>
+    /// This type exists largely to enable injected mock implementations of
+    /// <see cref="IPropertyPageQueryCache" /> for unit testing purposes.
+    /// </remarks>
+    [ProjectSystemContract(ProjectSystemContractScope.ProjectService, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    internal interface IPropertyPageQueryCacheProvider
+    {
+        public IPropertyPageQueryCache CreateCache(UnconfiguredProject project);
```
drewnoakes, line 17:
```suggestion
        IPropertyPageQueryCache CreateCache(UnconfiguredProject project);
```

drewnoakes (approved):

drewnoakes:
Testing this via the UI and found that (in the local scenario) changes are not persisted to disk until I hit <kbd>Ctrl</kbd>+<kbd>s</kbd>. Not sure why this is exactly. It's late here so I'll investigate more tomorrow.

tmeschter:
> Testing this via the UI and found that (in the local scenario) changes are not persisted to disk until I hit Ctrl+s. Not sure why this is exactly. It's late here so I'll investigate more tomorrow.

I don't think that's necessarily a problem--CPS sees the changes even if they haven't been persisted to disk.

drewnoakes:
The legacy property pages update the file on disk immediately. I'm unsure how saving the new editor actually causes the file on disk to change. There must be a link between the project file and the editor via the RDT. I've noticed that while the new editor is open, that attempts to edit the project file's XML result in the UI editor being focussed.

PR #6685, opened 2020-10-15
Set editor for StringProperty with subtype "file" or "path"

drewnoakes:
Properties may specify `ValueEditor`s to control how they are edited in the UI. However for `StringProperty` it's more common to specify the `Subtype` property with a value such as `"file"` to indicate that the property expects a file path.

We don't currently plumb `Subtype` through the Project Query API project model. This value only applies to `StringProperty` and we want the shape to be identical across property types.

The solution is to maps common `StringProperty.Subtype` values to `ValueEditor` values.

- Subtype `file` becomes editor `FileBrowse`
- Subtype `path` becomes editor `PathBrowse`

This causes, for example, the icon path property to show a "browse" button:

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6685)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
@@ -64,6 +64,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                     IEntityValue editorValue = CreateEditorValue(parent, editor, properties);
                     yield return editorValue;
                 }
+
+                if (property is StringProperty stringProperty)
+                {
+                    if (stringProperty.Subtype == "file")
+                    {
+                        yield return CreateEditorValue(parent, new ValueEditor { EditorType = "FilePath" }, properties);
+                    }
+                    else if (stringProperty.Subtype == "directory")
+                    {
+                        yield return CreateEditorValue(parent, new ValueEditor { EditorType = "DirectoryPath" }, properties);
+                    }
+                }
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -121,12 +121,12 @@
                   DisplayName="Output path"
                   Description="Specifies the location of the output files for this project's configuration."
                   Category="Output"
-                  Subtype="file"/>
+                  Subtype="directory" />
 
   <StringProperty Name="DocumentationFile"
                   DisplayName="XML documentation file path"
                   Description="Specifies the name of a file into which documentation comments will be processed."
                   HelpUrl="https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/doc-compiler-option"
                   Category="Output"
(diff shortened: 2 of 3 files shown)


jjmew (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
```diff
@@ -64,6 +64,18 @@ IEnumerable<IEntityValue> createEditorValues()
                     IEntityValue editorValue = CreateEditorValue(parent, editor, properties);
                     yield return editorValue;
                 }
+
+                if (property is StringProperty stringProperty)
+                {
+                    if (stringProperty.Subtype == "file")
+                    {
+                        yield return CreateEditorValue(parent, new ValueEditor { EditorType = "FileBrowse" }, properties);
```
tmeschter, line 72: Can we say "FileBrowse**r**" instead of "FileBrowse", as in the editor we want is a file browser?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
```diff
@@ -64,6 +64,18 @@ IEnumerable<IEntityValue> createEditorValues()
...
+                    }
+                    else if (stringProperty.Subtype == "path")
+                    {
+                        yield return CreateEditorValue(parent, new ValueEditor { EditorType = "PathBrowse" }, properties);
```
tmeschter, line 76: How about "DirectoryBrowser"? A path could be either a file or a directory.

tmeschter (approved):

tmeschter:
Approved, with a couple of naming suggestions.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
drewnoakes: The other editor types are things like "String", "Bool", "Enum". Does that change anything for you?

drewnoakes: Maybe "Path" makes more sense.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
drewnoakes: The API docs on `Subtype` suggested "path", but //index suggests no one is actually using that, so I'll add support for `Subtype="directory"` instead and use "Directory" or "DirectoryBrowse" for the editor type.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
tmeschter: Makes sense to me.

PR #6691, opened 2020-10-20
Support special ASP.NET folders

davkean:
- Marks ASP.NET folders with a specific flag based on its name.
- Gives them an icon to visiblity distinguish from normal folders. Note we need to request icons for folders as they are missing.
- Code folders support isn't completely hooked as we're missing the right hooks in CPS to determine where in the tree we are in a property provider.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6691)


davkean:
tag @BillHiebert

As I was filling out IWebProjectSupport, I started implementing this and just pulled it out to its own PR.

PR #6692, opened 2020-10-20
Feature remove unused references

ocallesp:
Fixes [AB#1103891](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1103891)

Implements `IProjectSystemReferenceCleanupService` for the feature _Remove Unused References_.

This is a request from the partner team Roslyn.

What this `ReferenceCleanupService `does:
Remove unused references for the given Project show the unused project or package references and allow users to selectively remove them.

Treat reference as used – Provide the ability to users to mark unused references as “TreatAsUsed” so that they will not be considered unused. This will be stored in the project file as an attribute on the Reference element. 

API definition is [here](https://microsoft-my.sharepoint.com/:w:/r/personal/jorobich_microsoft_com/_layouts/15/guestaccess.aspx?e=fhvJnB&CID=D2E3E0BA-D812-4F82-942B-8EE80470CD21&wdLOR=c0A841ED8-D99B-472A-9BD5-7ACEEFA8AFFA&share=EZUhVyqq8HBLo-33B7i1ADsBur5UWxENCydW53hdQT59kw).
Design document is [here](https://microsoft-my.sharepoint.com/:w:/r/personal/ivanbaso_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8EA5BB2B-E10C-4912-9536-4A6E3D84C7AA%7D&file=Remove%20Unused%20References%20-%20Speficiations.docx&wdLOR=c4418BB04-1714-4CC0-9F33-DA1D35099BB4&action=default&mobileredirect=true).

Future work:
Add overlay icon for dependencies nodes.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6692)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.VS.References.Roslyn;
```
ocallesp, line 11: This line will be removed. This was used only for testing purposes.

davkean:
Please build this on top of the IXXXReferences services (IBuildDependencyProjectReferencesService, IAssemblyReferencesService, IPackageReferencesService) and avoid manipulating the project file itself. Snapshot usage should also go through these services.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            {
+                return projectHierarchy;
+            }
+
+            return null;
+        }
+
+        private Project? TryGetProjectFromPath(string projectPath)
+        {
+            foreach (Project project in _dte.Value.Solution.Projects)
```
davkean, line 230: Avoid DTE usage and use IProjectService2.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                Assumes.Present(serviceSubscription);
+
+                serviceSubscription.ProjectRuleSource.SourceBlock.TryReceive(filter, out IProjectVersionedValue<IProjectSubscriptionUpdate> item);
+
+                _snapshotReferences = item.Value.CurrentState[_schema];
+
+                return _snapshotReferences;
+            }
+
+            private static bool filter(IProjectVersionedValue<IProjectSubscriptionUpdate> obj)
```
drewnoakes, line 48: Methods should start with an uppercase letter. You might also consider making this a lambda or local function.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            private string GetAttributeTreatAsUsed(KeyValuePair<string, IImmutableDictionary<string, string>> item)
+            {
+                item.Value.TryGetValue("TreatAsUsed", out string treatAsUsed);
+                treatAsUsed = string.IsNullOrEmpty(treatAsUsed) ? "False" : treatAsUsed;
+
+                return treatAsUsed;
+            }
+
+            private string GetAttributeItemSpecification(KeyValuePair<string, IImmutableDictionary<string, string>> item)
```
drewnoakes, line 76:
```suggestion
            private string? GetAttributeItemSpecification(KeyValuePair<string, IImmutableDictionary<string, string>> item)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                    references.Add(new ProjectSystemReferenceInfo(_referenceType, itemSpecification, treatAsUsed == "True"));
+                }
+
+                return references;
+            }
+
+            private string GetAttributeTreatAsUsed(KeyValuePair<string, IImmutableDictionary<string, string>> item)
+            {
+                item.Value.TryGetValue("TreatAsUsed", out string treatAsUsed);
```
drewnoakes, line 70:
```suggestion
                item.Value.TryGetValue(ProjectReference.TreatAsUsedProperty, out string treatAsUsed);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Exceptions;
+using Microsoft.VisualStudio.ProjectSystem.VS.References.Roslyn;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    [Export(typeof(IReferenceCleanupService))]
```
davkean, line 18: How does Roslyn get their hands on this service?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                            await access.CheckoutAsync(projectPath);
+
+                            item.SetMetadataValue("TreatAsUsed", newValue);
+
+                            wasUpdated = true;
+                        }
+                    }
+                    catch
+                    {
+                        wasUpdated = false;
```
drewnoakes, line 112: This line is redundant.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        private class ProjectReferenceHandler : ReferenceHandler
+        {
+            internal ProjectReferenceHandler() : base(ProjectSystemReferenceType.Project, ProjectReference.SchemaName, ProjectReference.SchemaName, "Identity")
+            { }
+
+            public override async Task<bool> RemoveReferenceAsync(ConfiguredProject configuredProject, ProjectSystemReferenceInfo reference)
+            {
+                Assumes.Present(configuredProject);
```
drewnoakes, line 127:
```suggestion
                Requires.NotNull(configuredProject, nameof(configuredProject));
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
davkean: This service is per "configuration" but the API seems to indicate there's a single version of the interface for the entire solution (because the APIs it provides take projectPaths). Which is it?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.VS.References.Roslyn;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    internal partial class ReferenceCleanupService
+    {
+        private abstract class ReferenceHandler
```
drewnoakes, line 17: It's not uncommon to name abstract base classes with a `Base` suffix.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                string newValue = referenceUpdate.Action == ProjectSystemUpdateAction.TreatAsUsed ? "True" : "False";
+
+                await activeConfiguredProject.Services.ProjectLockService.WriteLockAsync(async access =>
+                {
+                    var projectOther = await access.GetProjectAsync(activeConfiguredProject, cancellationToken);
+
+                    try
+                    {
+                        var item = projectOther.AllEvaluatedItems.Where(i =>
+                            string.Compare(i.ItemType, _referenceName, StringComparison.OrdinalIgnoreCase) == 0 &&
```
drewnoakes, line 98:
```suggestion
                            StringComparer.OrdinalIgnoreCase.Equals(i.ItemType, _referenceName) &&
```

Better yet, use the appropriate member from our `StringComparers` class.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                    }
+                }, cancellationToken);
+
+                return wasUpdated;
+            }
+        }
+
+        private class ProjectReferenceHandler : ReferenceHandler
+        {
+            internal ProjectReferenceHandler() : base(ProjectSystemReferenceType.Project, ProjectReference.SchemaName, ProjectReference.SchemaName, "Identity")
```
drewnoakes, line 122:
```suggestion
            internal ProjectReferenceHandler() : base(ProjectSystemReferenceType.Project, ProjectReference.SchemaName, ProjectReference.SchemaName, ProjectReference.IdentityProperty)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                await configuredProject.Services.ProjectReferences.RemoveAsync(reference.ItemSpecification);
+
+                return true;
+            }
+        }
+
+        private class PackageReferenceHandler : ReferenceHandler
+        {
+            internal PackageReferenceHandler() : base(ProjectSystemReferenceType.Package, PackageReference.SchemaName, PackageReference.SchemaName, "Name")
```
drewnoakes, line 139:
```suggestion
            internal PackageReferenceHandler() : base(ProjectSystemReferenceType.Package, PackageReference.SchemaName, PackageReference.SchemaName, PackageReference.NameProperty)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+            { }
+
+            public override async Task<bool> RemoveReferenceAsync(ConfiguredProject configuredProject, ProjectSystemReferenceInfo reference)
+            {
+                Assumes.Present(configuredProject);
```
drewnoakes, line 144: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+    internal partial class ReferenceCleanupService : IProjectSystemReferenceCleanupService
```
davkean, line 19: Where does IProjectSystemReferenceCleanupService come from? The spec calls this IReferenceCleanupService. We try to avoid using "ProjectSystem" in API names, it should just be "IProjectReferenceCleanupService".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                await configuredProject.Services.PackageReferences.RemoveAsync(reference.ItemSpecification);
+
+                return true;
+            }
+        }
+
+        private class AssemblyReferenceHandler : ReferenceHandler
+        {
+            internal AssemblyReferenceHandler() : base(ProjectSystemReferenceType.Assembly, AssemblyReference.SchemaName, "Reference", "Identity")
```
drewnoakes, line 156:
```suggestion
            internal AssemblyReferenceHandler() : base(ProjectSystemReferenceType.Assembly, AssemblyReference.SchemaName, AssemblyReference.PrimaryDataSourceItemType, AssemblyReference.IdentityProperty)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            }
+            else
+            {
+                referenceHandler = _sdkReferenceHandler;
+            }
+
+            return referenceHandler;
+        }
+
+        public bool IsProjectCpsBased(string projectPath)
```
davkean, line 199: Where is this API used? I can't see it as part of the API definition. We avoid using "Cps" in public APIs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        private class SdkReferenceHandler : ReferenceHandler
+        {
+            internal SdkReferenceHandler() : base(ProjectSystemReferenceType.Unknown, SdkReference.SchemaName, SdkReference.SchemaName, "Name")
+            { }
+
+            public override Task<bool> RemoveReferenceAsync(ConfiguredProject configuredProject, ProjectSystemReferenceInfo reference)
+            {
+                // Do not remove Sdks
+                return Task.FromResult(false);
```
drewnoakes, line 181: This creates a new `Task` each time. We can use a cached one to avoid allocating memory.

```suggestion
                return TaskResult.False;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            var projectHierarchy = GetProjectHierarchy(projectPath);
```
davkean, line 201: IProjectService2 is much better way of doing this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    [Export(typeof(IReferenceCleanupService))]
+    internal partial class ReferenceCleanupService : IProjectSystemReferenceCleanupService
+    {
+        ReferenceHandler _projectReferenceHandler = new ProjectReferenceHandler();
+        ReferenceHandler _packageReferenceHandler = new PackageReferenceHandler();
+        ReferenceHandler _assemblyReferenceHandler = new AssemblyReferenceHandler();
+        ReferenceHandler _sdkReferenceHandler = new SdkReferenceHandler();
```
drewnoakes: If these were a single `Dictionary<string, ReferenceHandler> _handlerByReferenceType` instead:

- `GetProjectSnapshotAsync` could be a `foreach` over `Values`
- `GetReferences` could be a `Values.SelectMany(...).ToList()`
- `FindReferenceHandler` could be a `TryGetValue`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            references.AddRange(_projectReferenceHandler.GetReferences());
+            references.AddRange(_packageReferenceHandler.GetReferences());
+            references.AddRange(_assemblyReferenceHandler.GetReferences());
+            references.AddRange(_sdkReferenceHandler.GetReferences());
+
+            return references;
+        }
+
+        public async Task<bool> TryUpdateReferenceAsync(string projectPath, string targetFrameworkMoniker, ProjectSystemReferenceUpdate referenceUpdate, CancellationToken cancellationToken)
```
davkean, line 115: It passes a TFM, but you ignore it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+                {
+                    wasUpdated =
+                        await referenceHandler.RemoveReferenceAsync(activeConfiguredProject, referenceUpdate.ReferenceInfo);
+                }
+            }
+
+            return wasUpdated;
+        }
+
+        private static bool canRemoveReference(ProjectSystemReferenceUpdate referenceUpdate, ReferenceHandler referenceHandler)
```
davkean, line 160: **C**anRemoveReference

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            else if (referenceUpdate.ReferenceInfo.ReferenceType == ProjectSystemReferenceType.Assembly)
+            {
+                referenceHandler = _assemblyReferenceHandler;
+            }
+            else
+            {
+                referenceHandler = _sdkReferenceHandler;
+            }
+
+            return referenceHandler;
```
drewnoakes, line 198: Perhaps store the handlers in a dictionary keyed by reference type. Other places where you explicitly call each of them can then be converted to a `foreach` or `Select` etc.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            {
+                referenceHandler = _packageReferenceHandler;
+            }
+            else if (referenceUpdate.ReferenceInfo.ReferenceType == ProjectSystemReferenceType.Assembly)
+            {
+                referenceHandler = _assemblyReferenceHandler;
+            }
+            else
+            {
+                referenceHandler = _sdkReferenceHandler;
```
davkean, line 193: Why do we fall back to "SdkReference"?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            }
+
+            return referenceHandler;
+        }
+
+        public bool IsProjectCpsBased(string projectPath)
+        {
+            var projectHierarchy = GetProjectHierarchy(projectPath);
+            var isCps = projectHierarchy.IsCapabilityMatch("CPS");
```
drewnoakes, line 204: Perhaps add `ProjectCapability.CPS` constant, rather than using a literal here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            return activeConfiguredProject;
+        }
+
+        private UnconfiguredProject FindUnconfiguredProjectByPath(string projectPath)
+        {
+            var unconfigProjects = _configuredProject.Services.ProjectService.LoadedUnconfiguredProjects;
+
+            var unconfiguredProject = unconfigProjects.First(project =>
+                StringComparers.Paths.Equals((string)project.FullPath, projectPath));
```
drewnoakes, line 83: `First` throws if not found. Is that ok?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+            private static bool filter(IProjectVersionedValue<IProjectSubscriptionUpdate> obj)
+            {
+                return true;
+            }
+
+            public List<ProjectSystemReferenceInfo> GetReferences()
+            {
+                List<ProjectSystemReferenceInfo> references = new List<ProjectSystemReferenceInfo>();
+
+                foreach (var item in _snapshotReferences.Items)
```
davkean, line 57: Instead of using snapshots, please just use IXXXReferencesService.GetUnresolvedReferencesAsync

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            return unconfiguredProject;
+        }
+
+        private async Task<List<ProjectSystemReferenceInfo>> GetAllReferencesInConfiguredProjectAsync(ConfiguredProject selectedConfiguredProject)
```
drewnoakes: Does this have to be `async`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            return wasUpdated;
+        }
+
+        private static bool canRemoveReference(ProjectSystemReferenceUpdate referenceUpdate, ReferenceHandler referenceHandler)
+        {
+            var references = referenceHandler.GetReferences();
+
+            var reference = references.Where(c =>
+                c.ItemSpecification == referenceUpdate.ReferenceInfo.ItemSpecification).First();
```
drewnoakes, line 167: Ditto re `First`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                {
+                    string treatAsUsed = GetAttributeTreatAsUsed(item);
+                    string itemSpecification = GetAttributeItemSpecification(item);
+
+                    references.Add(new ProjectSystemReferenceInfo(_referenceType, itemSpecification, treatAsUsed == "True"));
```
drewnoakes, line 62: Here and elsewhere, consider replacing `False` with `bool.FalseString`, and `True` with `bool.TrueString`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            public async Task<bool> UpdateReferenceAsync(ConfiguredProject activeConfiguredProject, ProjectSystemReferenceUpdate referenceUpdate, CancellationToken cancellationToken)
+            {
+                bool wasUpdated = false;
+
+                string projectPath = activeConfiguredProject.UnconfiguredProject.FullPath;
+
+                string newValue = referenceUpdate.Action == ProjectSystemUpdateAction.TreatAsUsed ? "True" : "False";
+
+                await activeConfiguredProject.Services.ProjectLockService.WriteLockAsync(async access =>
```
davkean, line 91: @Lifengl What's the easiest way to set metadata on a IProjectItem? 

I don't think below is going to handle situations such as when the item comes from an import.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
```diff
@@ -93,4 +93,7 @@
                 ReadOnly="True"
                 Visible="False" />
 
+  <BoolProperty Name="TreatAsUsed"
+                Description="Indicates that this reference should be treated as if it were used."/>
```
drewnoakes, line 97: Is this supposed to be visible to the user in the properties pane? If so, it must have `DisplayName`. If not, set `Visible` to false and remove the `Description`. Ditto other rules in this PR.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
davkean: I think perhaps we need a way to set arbitrary metadata here on an IUnresolvedReference.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+            internal AssemblyReferenceHandler() : base(ProjectSystemReferenceType.Assembly, AssemblyReference.SchemaName, "Reference", "Identity")
+            { }
+
+            public override async Task<bool> RemoveReferenceAsync(ConfiguredProject configuredProject, ProjectSystemReferenceInfo reference)
+            {
+                Assumes.Present(configuredProject);
+                Assumes.Present(configuredProject.Services);
+                Assumes.Present(configuredProject.Services.AssemblyReferences);
+
+                AssemblyName assemblyName = new AssemblyName(reference.ItemSpecification);
```
davkean, line 165: ItemSpecification for reference can be in form `<Reference Include="System" />`, `<Reference Include="System, Version=4.0.0.0, PublicKeyToken=null" />` or `<Reference Include="C:\Foo.dll" />`. AssemblyName will throw on certain file names.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                await configuredProject.Services.AssemblyReferences.RemoveAsync(assemblyName, null);
+
+                return true;
+            }
+        }
+
+        private class SdkReferenceHandler : ReferenceHandler
```
davkean, line 173: I don't think SdkReferenceHandler should exist unless Roslyn specifically handles these reference items, I not sure why we're falling back to it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
davkean: I think this is supposed to be a normal MEF (non-CPS MEF component) in which case you can't import ConfiguredProject. You should instead be importing IProjectServiceAccessor and getting IProjectService from it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            GetProjectSnapshotAsync(selectedConfiguredProject);
+
+            var references = GetReferences();
+
+            return references;
+        }
+
+        private void GetProjectSnapshotAsync(ConfiguredProject selectedConfiguredProject)
+        {
+            _projectReferenceHandler.GetProjectSnapshot(selectedConfiguredProject);
```
davkean, line 99: This pattern would be much nicer if this was a dictionary of reference type -> to the handler, and you just walked the dictionary instead of hanving "_projectReferenceHandler", "_packageReferenceHandler", etc. Then adding a new reference type is a 1 line change to this class instead of changing all usages.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+            references.AddRange(_sdkReferenceHandler.GetReferences());
+
+            return references;
+        }
+
+        public async Task<bool> TryUpdateReferenceAsync(string projectPath, string targetFrameworkMoniker, ProjectSystemReferenceUpdate referenceUpdate, CancellationToken cancellationToken)
+        {
+            bool wasUpdated = false;
+
+            if (referenceUpdate.Action == ProjectSystemUpdateAction.None)
```
davkean, line 121: Why would we be passed this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,252 @@
...
+                        await referenceHandler.RemoveReferenceAsync(activeConfiguredProject, referenceUpdate.ReferenceInfo);
+                }
+            }
+
+            return wasUpdated;
+        }
+
+        private static bool canRemoveReference(ProjectSystemReferenceUpdate referenceUpdate, ReferenceHandler referenceHandler)
+        {
+            var references = referenceHandler.GetReferences();
```
davkean, line 164: This would be much more efficient to just go through the GetUnresolvedReferenceAsync API than calculating all references only to grab the first one.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
davkean: Avoid the "ProjectSystem" prefix completely.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
davkean: We copied/were forced to use **Abstract**XXX pattern from Roslyn.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
davkean: We should be using PropertySerializer.SimpleTypes.ToString(value).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
davkean: These have to be added to the ResolvedXXX versions of these rules. I'm surprised we don't have tests for this.

davkean: I think we should sync with .NET Core reference trimming feature to see if they have a similar feature for this (to indicate that a reference should not be removed during trimming), so there's only one concept for this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: it seems that the full path to the project is enough. In which cases should we use fullPath + TFM ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: Do you mean   `IProjectService2.GetLoadedProject(projectPath).Capabilities.Current.Value.IsProjectCapabilityPresent("CPS")` ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: I was using GetUnresolvedReferencesAsync, but the unit-test was a little more complicated due to the static methods so I used snapshots.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: Can we add a warning rule for methods to be Pascal ? I added this, but didn't detect anything.

image

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done. Changed to match pattern from Roslyn

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done. Changed to use bool.FalseString/bool.TrueString

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: when should we use one or the other ?
Assumes for services ? and Requires for values ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: removed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
ocallesp: Done. Added to ResolvedProjectReference, ResolvedPackageReference, ResolvedAssemblyReference and removed from SdkReference.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done. Removed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: Roslyn should not send this value, and we can ignore it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: Yes. I was thinking Roslyn should handle the case when the project is incorrect/typo/not found.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: In this case we should return False, since we cannot make changes to a reference that doesn't exist.
Done. Handled with try/catch

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
drewnoakes: `Assumes.Present` is for services. `Requires.*` is for argument validation.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,235 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    [Export(typeof(IProjectSystemReferenceCleanupService))]
+    internal partial class ReferenceCleanupService : IProjectSystemReferenceCleanupService
+    {
+        private readonly AbstractReferenceHandler _projectAbstractReferenceHandler = new ProjectAbstractReferenceHandler();
+        private readonly AbstractReferenceHandler _packageAbstractReferenceHandler = new PackageAbstractReferenceHandler();
+        private readonly AbstractReferenceHandler _assemblyAbstractReferenceHandler = new AssemblyAbstractReferenceHandler();
```
drewnoakes, line 25: I don't think you need these fields. In the methods below, can't you foreach over `_mapReferenceTypeToHandler.Values.Where(handler => handler != null)`

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,235 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+#nullable disable
```
drewnoakes, line 3: Is there a reason this isn't annotated?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,235 @@
...
+            var references = abstractReferenceHandler.GetReferences();
+            try
+            {
+                referenceInfo = references.First(c =>
+                    c.ItemSpecification == referenceUpdate.ReferenceInfo.ItemSpecification);
+            }
+            catch
+            {
+                return false;
+            }
```
drewnoakes, line 177: Avoid using exceptions for control flow -- they have a performance impact and make debugging harder. Could use `FirstOrDefault` here then check for null.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: removed, it was to differentiate zero references in a non-cps project vs a project with no references.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done. replaced with SetPropertyValueAsync

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: the refactor now makes this method async

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done. Try/catch removed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: done. Enable nullable

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/References/ReferenceCleanupServiceTests.cs
```diff
@@ -0,0 +1,256 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Exceptions;
+using Moq;
+using Xunit;
+using System.Collections.Immutable;
+using System.Reflection;
+using Microsoft.VisualStudio.Mocks;
+using Microsoft.VisualStudio.ProjectSystem.References;
+using Microsoft.VisualStudio.ProjectSystem.VS.References.Roslyn;
```
ocallesp, line 13: this line will be changed by the real Interface provided by Roslyn

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+            internal AssemblyReferenceHandler() :
+                base(ProjectSystemReferenceType.Assembly)
+            { }
+
+            protected override Task RemoveReferenceAsync(ConfiguredProjectServices services,
+                ProjectSystemReferenceInfo referencesInfo)
+            {
+                Assumes.Present(services.AssemblyReferences);
+
+                var assemblyName = new AssemblyName(referencesInfo.ItemSpecification);
```
ocallesp, line 172: @davkean 
UnevaluatedInclude will break AssemblyName ? Should we use full path instead ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
JoeRobich: These interfaces are in the Project System ExternalAccess portion of Roslyn which sits in the VisualStudio layer. I do have a IReferenceCleanupService interface but it is internal and in the Features layer where you do not have IVTs.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
davkean: ItemSpecification is not always an "assembly name":

``` XML
<Reference Include="C:\Foo\System.dll"/>
```

You'll need to handle both cases.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                if (projectItems is null)
+                {
+                    return references;
+                }
+
+                foreach (var item in projectItems)
+                {
+                    string? treatAsUsed = await GetAttributeTreatAsUsedAsync(item.Metadata);
+                    string? itemSpecification = item.UnevaluatedInclude;
```
davkean, line 76: ItemSpecification == EvaluatedInclude and not UnevaluatedInclude. Please use this consistently.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                    if (!string.IsNullOrEmpty(itemSpecification))
+                    {
+                        references.Add(new ProjectSystemReferenceInfo(_referenceType, itemSpecification, treatAsUsed == bool.TrueString));
+                    }
+                }
+
+                return references;
+            }
+
+            private static async Task<string?> GetAttributeTreatAsUsedAsync(IProjectProperties metadata)
```
davkean, line 87: This should be typed as "bool".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                string? value = await metadata.GetUnevaluatedPropertyValueAsync(ProjectReference.TreatAsUsedProperty);
```
davkean, line 89: This must use "Evaluated" not unevaluated.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var projectItems = await GetUnresolvedReferencesAsync(selectedConfiguredProject);
+
+                var item = projectItems
+                    .FirstOrDefault(c => c.UnevaluatedInclude == referenceUpdate.ReferenceInfo.ItemSpecification);
+
+                if (item != null)
+                {
+                    string newValue = referenceUpdate.Action == ProjectSystemUpdateAction.TreatAsUsed ? bool.TrueString : bool.FalseString;
```
davkean: Please use PropertySerializer.SimpleTypes.ToString to convert from "bool" to "string", not bool.TrueString or bool.FalseString.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                {
+                    return references;
+                }
+
+                foreach (var item in projectItems)
+                {
+                    string? treatAsUsed = await GetAttributeTreatAsUsedAsync(item.Metadata);
+                    string? itemSpecification = item.UnevaluatedInclude;
+
+                    if (!string.IsNullOrEmpty(itemSpecification))
```
davkean, line 78: UnevaluatedInclude is non-nullable and will never return an empty value.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                return value ?? bool.FalseString;
+            }
+
+            internal async Task<bool> UpdateReferenceAsync(ConfiguredProject selectedConfiguredProject, ProjectSystemReferenceUpdate referenceUpdate, CancellationToken cancellationToken)
+            {
+                bool wasUpdated = false;
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var projectItems = await GetUnresolvedReferencesAsync(selectedConfiguredProject);
```
davkean, line 99: Please make sure you fix all Messages/Warnings in the Error List, these files aren't following the .editorconfig code styles.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                    await item.Metadata.SetPropertyValueAsync(ProjectReference.TreatAsUsedProperty, newValue, null);
+
+                    wasUpdated = true;
+                }
+
+                return wasUpdated;
+            }
+        }
+
+        private class ProjectReferenceHandler : AbstractReferenceHandler
```
davkean, line 117: We place nested classes in their own file.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    { ProjectSystemReferenceType.Project, new ProjectReferenceHandler() },
+                    { ProjectSystemReferenceType.Package, new PackageReferenceHandler() },
+                    { ProjectSystemReferenceType.Assembly, new AssemblyReferenceHandler()},
+                    { ProjectSystemReferenceType.Unknown, null }
+                };
+        }
+
+        public async Task<ImmutableArray<ProjectSystemReferenceInfo>> GetProjectReferencesAsync(string projectPath, string targetFrameworkMoniker, CancellationToken cancellationToken)
+        {
```
davkean, line 39: This takes a TFM but you ignore it. What does this TFM represent?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
davkean: I don't follow. What interface to do we need to implement?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: this enum value was removed from ProjectSystemUpdateAction

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/References/ReferenceCleanupServiceTests.cs
ocallesp: removed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,146 @@
...
+            _projectServiceAccessor = projectServiceAccessor;
+
+            _mapReferenceTypeToHandler =
+                new Dictionary<ProjectSystemReferenceType, AbstractReferenceHandler>()
+                {
+                    { ProjectSystemReferenceType.Project, new ProjectReferenceHandler() },
+                    { ProjectSystemReferenceType.Package, new PackageReferenceHandler() },
+                    { ProjectSystemReferenceType.Assembly, new AssemblyReferenceHandler()},
+                    { ProjectSystemReferenceType.Unknown, null }
```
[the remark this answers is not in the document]
ocallesp: removed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
```diff
@@ -0,0 +1,185 @@
...
+                return referenceInfo.TreatAsUsed != true;
+            }
+
+            internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(ConfiguredProject selectedConfiguredProject, CancellationToken cancellationToken)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var references = new List<ProjectSystemReferenceInfo>();
+
+                var projectItems = await GetUnresolvedReferencesAsync(selectedConfiguredProject);
```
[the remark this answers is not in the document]
ocallesp: done. Removed
The returned value from GetUnresolvedReferencesAsync() is not nullable.

Checking for null is not needed here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: Done.
Moved AbstractReferenceHandler to its own file.

I wanted to keep this class private, so only ReferenceCleanupService will use it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,146 @@
...
+            return activeConfiguredProject;
+        }
+
+        private UnconfiguredProject FindUnconfiguredProjectByPath(string projectPath)
+        {
+            var projectService = _projectServiceAccessor.GetProjectService();
+
+            var unconfigProjects = projectService.LoadedUnconfiguredProjects;
+
+            var unconfiguredProject = unconfigProjects.First(project =>
```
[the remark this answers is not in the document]
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: this will be removed from the interface  https://github.com/dotnet/roslyn/pull/48990

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: this will be removed from the interface  https://github.com/dotnet/roslyn/pull/48990

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
JoeRobich: We need you to implement the IProjectSystemReferenceCleanupService.

The naming is tricky here because a Project can have references and those references can be ProjectReferences. Leading to "IProjectReferenceCleanupService" having potentially two meanings; "Oh, a service for cleaning up ProjectReferences" and "Oh, a service for cleaning up a Project's references".

If this is a blocker for you, we can workshop a better name.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: Which errors ? I see there are 0 warning/errors

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: I added code to check when ItemSpecification is path or name.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.ReferenceHandlers.cs
ocallesp: added code to check when Include is a path or name

build/import/Packages.targets
```diff
@@ -14,15 +14,18 @@
...
       https://api.nuget.org/v3/index.json;
-      https://dotnet.myget.org/F/symreader-converter/api/v3/index.json
+      https://dotnet.myget.org/F/symreader-converter/api/v3/index.json;
     </RestoreSources>
   </PropertyGroup>
 
   <ItemGroup>
 
     <!-- Toolset -->
+    <PackageReference Include="Nerdbank.Streams"                                                      Version="2.6.81" />
```
davkean, line 27: Please use the same pattern as all the other package references. Update goes in this file, the reference goes in the layer appropriate file (VisualStudio.props, etc).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -0,0 +1,106 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+#nullable enable
```
davkean, line 3: This is unneeded, please remove.

build/import/Packages.targets
ocallesp: done

dmonroym (approved):

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IProjectServiceFactory.cs
```diff
@@ -23,5 +24,21 @@ public static IProjectService Create(ProjectServices? services = null, IProjectC
...
+        public static IProjectService Create(IEnumerable<UnconfiguredProject> loadedUnconfiguredProjects, ProjectServices? services = null, IProjectCapabilitiesScope? scope = null)
+        {
+            var mock = new Mock<IProjectService>();
+
+            services ??= ProjectServicesFactory.Create(projectService: mock.Object);
+
+            mock.Setup(p => p.Services)
+                .Returns(services);
+
+
```
MiYanni, line 37: nit: extra newline

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,133 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var unconfiguredProjectPath = ProjectService.GetLoadedProject(projectPath);
+
+            if (unconfiguredProjectPath is null)
+            {
+                throw new InvalidProjectFileException();
+            }
+
+            var activeConfiguredProject = await unconfiguredProjectPath.GetSuggestedConfiguredProjectAsync();
```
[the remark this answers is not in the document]
drewnoakes: With this change, the exception changes from `InvalidProjectFileException` to `NullReferenceException`. You cannot `await` a `null` value.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -0,0 +1,133 @@
...
+            var activeConfiguredProject = await GetActiveConfiguredProjectByPathAsync(projectPath, cancellationToken);
+
+            var referenceHandler = s_mapReferenceTypeToHandler[referenceUpdate.ReferenceInfo.ReferenceType];
+
+            if (referenceHandler != null)
+            {
+                wasUpdated = await ApplyActionAsync(referenceUpdate, referenceHandler, activeConfiguredProject, cancellationToken);
+            }
+
+            return wasUpdated;
```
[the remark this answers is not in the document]
drewnoakes: Quite a bit of code in this repo has generous white space.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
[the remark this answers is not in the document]
MiYanni: Ah. I see. Then, it would need to be more annoying, like:
```
await ProjectService.GetLoadedProject(projectPath)?.GetSuggestedConfiguredProjectAsync()?.AsTask() ?? Task.CompletedTask;
```

Fun times with C#. 😝

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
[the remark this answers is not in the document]
MiYanni: I tend to use whitespace to separate logically grouped things. Always found it odd to use them in excess.

PR #6702, opened 2020-10-22
Remove condition on code coverage which would work with NET 5 TFM

### LINKED ISSUE #6666: Audit TargetFramework usage for .NET 5 TFM change
With .NET 5 we are changing target framework from the netcoreappX.X style to netX.X. With this change, it appears this repo has code that could possibly break. Please check out https://github.com/dotnet/project-system/blob/master/build/import/Coverage.targets#L8 to ensure that this logic works as intended for .NET 5 TFMs, noting that $(TargetFramework.StartsWith('netcoreapp')) will no longer be true with net5.0, thanks!

dmonroym:
Remove condition on code coverage which would work with NET 5 TFM

Fixes #6666

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6702)


dmonroym:
@davkean I removed the condition but seeing as this is dead code should I remove the file and its references entirely or is there a plan to use this again?

build/import/Coverage.targets
```diff
@@ -4,10 +4,7 @@
   <Target Name="ExtractWindowsPdb" BeforeTargets="Test">
 
     <PropertyGroup>
-      <_RunOnCore>false</_RunOnCore>
-      <_RunOnCore Condition="$(TargetFramework.StartsWith('netcoreapp'))">true</_RunOnCore>
       <_UseOpenCover>$(UseOpenCover)</_UseOpenCover>
```
drewnoakes, line 7: Assuming this file is not being removed, `_UseOpenCover` can be inlined.

dmonroym:
Instead of modifying this code to fix #6666 I just removed it entirely as it's not being used.

drewnoakes (approved):

PR #6706, opened 2020-10-23
Adapt to Project Query API changes

tmeschter:
A recent change in the Project Query API ([PR 281922](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/281922) in CPS) changed the `SearchTerms` and `DependsOn` properties of `UIProperty` to return `string` instead of `ImmutableList<string>`. Here we adopt the latest version of the APR, and adapt the implementation to the updated return types. In practice these means that we just read these values from metadata in the .xaml file and pass it through unchanged, rather than splitting it up into a list. If the UI needs to do so, it will handle the splitting.

One odd aspect of this change is the need to move the references to PresentationCore, PresentationFramework, and WindowsBase from VisualStudio.props to HostAgnostic.props. The Microsoft.VisualStudio.ProjectSystem.Managed project has code that depends on the WPF `Control` and `UserControl` types and yet it did not have explicit references to them. I speculate that they used to be brought in by the previous version of
Microsoft.VisualStudio.ProjectSystem.SDK and the new version no longer does so. Regardless, since we need these types in both Microsoft.VisualStudio.ProjectSystem.Managed and Microsoft.VisualStudio.ProjectSystem.Managed.VS we can move the references into HostAgnostic.props, which is pulled in by both projects.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6706)

### DIFF
--- a/build/import/HostAgnostic.props
+++ b/build/import/HostAgnostic.props
@@ -6,11 +6,14 @@
   
   <ItemGroup>
     <!-- Framework -->
+    <Reference Include="PresentationCore" />
+    <Reference Include="PresentationFramework" />
     <Reference Include="System" />
     <Reference Include="System.ComponentModel.Composition" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Xml.Linq" />
+    <Reference Include="WindowsBase" />
     
     <!-- Toolset -->
     <PackageReference Include="MicroBuild.Core"                     ExcludeAssets="All" />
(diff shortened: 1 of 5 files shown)


tmeschter:
Yeah, the references to PresentationFramework et al used to come from the Microsoft.VisualStudio.ProjectSystem package; it no longer pulls those in.

ocallesp (approved):

drewnoakes (approved):

PR #6709, opened 2020-10-26
Properly support configuration-independent properties

tmeschter:
Currently when retrieving the values for a property we get them across all configurations, even for properties that are configuration-independent. At best this is inefficient as we will retrieve and transmit many duplicate values that the UI will not use. At worse it breaks the query when a particular property implementation expects to only be asked about its value (or supported values) in the currently active configuration (I'm looking at you, `TargetFrameworkMoniker`).

Here we update the `UIPropertyValueDataProducer` to only return a single value for these properties. The `IPropertyPageQueryCache` has been updated to return the "suggested" `ProjectConfiguration`, and this is the one we use to retrieve the property value as it is likely to have already been loaded.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6709)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/IPropertyPageQueryCache.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/IPropertyPageQueryCache.cs
@@ -41,5 +41,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
     {
         Task<IRule?> BindToRule(ProjectConfiguration projectConfiguration, string schemaName);
         Task<IImmutableSet<ProjectConfiguration>?> GetKnownConfigurationsAsync();
+        Task<ProjectConfiguration?> GetDefaultConfigurationAsync();
     }
 }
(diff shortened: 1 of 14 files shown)


tmeschter:
Note that `TargetFrameworkMoniker` currently throws an exception if you ask it for its supported values for any configuration other than the active configuration, as best I can tell. This change fixes that particular issue, in addition to being the right thing to do in general.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/IPropertyPageQueryCache.cs
```diff
@@ -41,5 +41,6 @@ internal interface IPropertyPageQueryCache
     {
         Task<IRule?> BindToRule(ProjectConfiguration projectConfiguration, string schemaName);
         Task<IImmutableSet<ProjectConfiguration>?> GetKnownConfigurationsAsync();
+        Task<ProjectConfiguration?> GetDefaultConfigurationAsync();
```
swesonga, line 44: >GetDefaultConfigurationAsync [](start = 36, length = 28)

wouldn't GetSuggestedConfigurationAsync be a more accurate name for this method? Defaults sound like things that don't need to be computed

drewnoakes (changes requested):
I played with this a little today and discovered an issue.

With this change configuration independent properties return only a single property value, and that value has arbitrary dimensions. This causes the UI to display the value as specific to that configuration.

It is fine to use any arbitrary configuration to query the value, but the value returned by the query should probably specify no configuration dimensions at all, indicating it applies to all configurations.

A specific example, the `AssemblyName` property has a single item in its `Values` array which specifies dimensions `Debug|AnyCPU|netcoreapp3.0`.

tmeschter:
@drewnoakes I've fixed that issue. If the property is configuration-independent it will no longer report any dimensions associated with the value.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/IPropertyPageQueryCache.cs
tmeschter: Updated.

drewnoakes (approved):
Thanks @tmeschter. I've tested this and it works as expected 👍

PR #6712, opened 2020-10-27
Remove target framework moniker parsing code

### LINKED ISSUE #6369: Stop using NuGet's IVsFrameworkParser in Microsoft.VisualStudio.ProjectSystem.VS.TargetFrameworkProvider
### Background

NuGet's `IVsFrameworkParser` API provides two APIs:

```cs
public interface IVsFrameworkParser
{
    FrameworkName ParseFrameworkName(string shortOrFullName);
    string GetShortFrameworkName(FrameworkName frameworkName);
}
```

The problem is that `System.Runtime.Versioning.FrameworkName` does not have properies that allows us to distingish `net5.0` and `net5.0-windows`. Therefore, these APIs will not work with net5.0 projects. Full net5.0 TFM spec: https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md

NuGet itself is going to be moving away from parsing the project's `TargetFramework` and will determine the project's TFM using the `TargetFrameworkIdentifier`, `TargetFrameworkVersion`, `TargetPlatformIdentifier` and `TargetFrameworkVersion`. This means that TFM aliasing is becoming increasingly close to a possibility, so it will be unsafe to assume that the project's `TargetFramework` will be parsable.

### Proposal

@tmeschter mentioned to me on Teams that this class looks like it might be used in the dependencies node in Solution Explorer. In this case, I recommend using the `TargetFramework` (either directly as the user provided, or with `ToLowerInvariant`, as [the SDK itself does](https://github.com/dotnet/sdk/blob/b8688e3db746fcd5117b78d1538936defde79e87/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets#L215)). This would allow projects to "just work" when/if [TFM aliasing gets implemented](https://github.com/dotnet/sdk/issues/1561#issuecomment-327051425).

### LINKED ISSUE #6370: Stop using NuGet's IVsFrameworkParser in Microsoft.VisualStudio.ProjectSystem.VS.Properties.TargetFrameworkMonikerValueProvider
### Background

NuGet's `IVsFrameworkParser` API provides two APIs:

```cs
public interface IVsFrameworkParser
{
    FrameworkName ParseFrameworkName(string shortOrFullName);
    string GetShortFrameworkName(FrameworkName frameworkName);
}
```

The problem is that `System.Runtime.Versioning.FrameworkName` does not have properies that allows us to distingish `net5.0` and `net5.0-windows`. Therefore, these APIs will not work with net5.0 projects. Full net5.0 TFM spec: https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md

NuGet itself is going to be moving away from parsing the project's `TargetFramework` and will determine the project's TFM using the `TargetFrameworkIdentifier`, `TargetFrameworkVersion`, `TargetPlatformIdentifier` and `TargetFrameworkVersion`. This means that TFM aliasing is becoming increasingly close to a possibility, so it will be unsafe to assume that the project's `TargetFramework` will be parsable.

### Proposal

@tmeschter mentioned to me on Teams that this class looks like it might be used to set the TargetFramework in the project/application properties page.  From the .NET 5 design spec: https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#tfms-in-the-ui

> For cases where we use branded display names, we should use the name .NET 5.

NuGet won't be able to parse ".NET 5" in any case (just as it can't parse ".NET Core 3.1"). So, I'm not sure how project system converted ".NET Core 3.1" into a string that `IVsFrameworkParser.ParseFrameworkName` could use, but perhaps instead of round-tripping with the two `IVsFrameworkParser` APIs, project-system could just use `net5.0` directly?  Keep in mind that developers who want to use WPF will need their project to target `net5.0-windows`, so the value in `TargetFramework` in the csproj is different, unlike using WPF in `netcoreapp3.1`.

Talk to me, or others in the NuGet team, if APIs are needed for this use-case.

### LINKED ISSUE #6674: net5.0 platform not displayed in dependencies tree
**Visual Studio Version**: 16.9.0 Preview 1.0 [30612.141.main]

**Summary**:
When loading up a project that has a TFM of `net5.0-windows10.0.19041`, VS only shows `net5.0` in the project's dependencies list. This is confusing since there could be multiple net5.0-based TFM's.

**Steps to Reproduce**:

1. Use a TFM of `net5.0-windows10.0.19041`

2. 

3. 

**Expected Behavior**:
VS should show `net5.0-windows10.0.19041`

**Actual Behavior**:
VS shows `net5.0`

image

**User Impact**:
Confusion

### LINKED ISSUE #6826: Dependency tree doesn't correctly support .NET 5+ platform specifiers
Version 16.9.0 Preview 4.0 [30909.6.main]

Given:

```xml
  <PropertyGroup>
    <TargetFrameworks>net5.0;net5.0-windows</TargetFrameworks>
  </PropertyGroup>
</Project>
```

This project has two TFMs. It should show two sub-graphs under the dependencies tree, but only shows one:

image

swesonga:
Fixes #6369.
Fixes #6370.
Fixes #6674.
Fixes #6826.

This change removes the calls to the `IVsFrameworkParser` API. This makes it possible to differentiate between TFMs with and without OS versions. It also addresses #6674 by displaying the full TFM in the dependencies node.

The change also replaces some references to `ConfigurationGeneral.TargetFramework` with `ConfigurationGeneral.TargetFrameworkMoniker` since the latter is the correct way to get the TFM for projects that are not cross-targeting (see `AggregateCrossTargetProjectContextProvider` and `DependenciesSnapshotProvider.ContextTracker`)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6712)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/TargetFramework.cs
```diff
@@ -19,36 +18,17 @@ internal sealed class TargetFramework : IEquatable<TargetFramework?>
...
+            TargetFrameworkMoniker = moniker;
         }
 
-        /// <summary>
-        /// Gets the full moniker (TFM).
-        /// </summary>
-        public string FullName { get; }
-
         /// <summary>
         /// Gets the short name.
```
drewnoakes, line 29: nit: comment needs updating

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.VisualStudio.ProjectSystem
     internal interface ITargetFrameworkProvider
     {
         /// <summary>
-        /// Parses full tfm or short framework name and returns a corresponding <see cref="TargetFramework"/>
-        /// instance, or <see langword="null" /> if the framework name has an invalid format.
+        /// Returns a <see cref="TargetFramework"/> instance for the target framework moniker
+        /// or <see langword="null" /> if the framework moniker is null or empty.
         /// </summary>
-        TargetFramework? GetTargetFramework(string? shortOrFullName);
+        TargetFramework? GetTargetFramework(string? targetFrameworkMoniker);
```
drewnoakes, line 14: A quick scan of the code suggests that we never actually pass a nullable string to this method. While you're cleaning this up, consider removing the null propagation. Alternatively, this could be annotated with `[return: NotNullIfNotNull("targetFrameworkMoniker")]`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -178,7 +178,7 @@ public void Dispose()
                 // Scan the snapshot and tally dependencies
                 foreach ((TargetFramework targetFramework, TargetedDependenciesSnapshot targetedSnapshot) in _dependenciesSnapshot.DependenciesByTargetFramework)
                 {
-                    var targetData = new TargetData(targetFramework.ShortName);
+                    var targetData = new TargetData(targetFramework.TargetFrameworkMoniker);
```
drewnoakes, line 181: @jjmew this will potentially change the representation of TFM in telemetry data.

i.e. before `net472`, after `.NETFramework,Version=v4.7.2`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Models/TargetDependencyViewModel.cs
```diff
@@ -16,15 +16,15 @@ internal sealed class TargetDependencyViewModel : IDependencyViewModel
 
         public TargetDependencyViewModel(TargetFramework targetFramework, DiagnosticLevel diagnosticLevel)
         {
-            Caption = targetFramework.ShortName;
+            Caption = targetFramework.TargetFrameworkMoniker;
```
drewnoakes, line 19: If this truly is the TFM, I would expect the display string to change from `net472` to `.NETFramework,Version=v4.7.2`, which probably isn't what we want across all target frameworks. Am I missing something?

tests/Microsoft.VisualStudio.ProjectSystem.IntegrationTests/DependencyNodeIntegrationTests.cs
```diff
@@ -26,14 +26,14 @@ public void SingleTarget_NetCoreApp21()
         [TestMethod]
         public void MultiTarget_NetCoreApp21_Net45()
         {
-            CreateProject(new Project("netcoreapp2.1;net45"));
+            CreateProject(new Project(";net45"));
```
drewnoakes, line 29: Is this change intentional?

Thanks for updating the integration tests.

drewnoakes (commented):
Some initial feedback.

Main question is around the display string. Have you tested this using a range of different target frameworks, not just .NET 5?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/CrossTarget/AggregateCrossTargetProjectContextProvider.cs
```diff
@@ -86,7 +86,7 @@ public async Task<AggregateCrossTargetProjectContext> CreateProjectContextAsync(
         {
             if (string.IsNullOrEmpty(shortOrFullName))
             {
-                object? targetObject = await configurationGeneralProperties.TargetFramework.GetValueAsync();
+                object? targetObject = await configurationGeneralProperties.TargetFrameworkMoniker.GetValueAsync();
```
davkean, line 89: I think this is wrong in the case of `net5.0-windows`; this results in:

```
TargetFrameworkMoniker; .NETCoreApp, Version=v5.0
TargetPlatformMoniker; Windows, Version=7.0
```

Either we represent both of these, or we need to continue to read this value. This really should be coming from the "TargetFramework" dimension from the ProjectConfiguration of the current project.

swesonga: I had started out expecting the "TargetFramework" dimension in configuredProject.ProjectConfiguration.Dimensions to be non-empty but that isn't the case. I thought that was by design. Let me dig into why that's happening.

davkean: That is by design in non-multi-TFM projects. Maybe we need to sync up on this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Frameworks/ITargetFrameworkProvider.cs
drewnoakes: I was wrong. The method can return `null` when an input is non-null (empty string).

drewnoakes (approved):

drewnoakes:
@swesonga we should get this PR in. I've made some changes to help with that:

- Fix merge conflicts.
- Replace the use of `TargetFrameworkMoniker` with `TargetFramework` and verify this behaves as expected.
- Remove the annotation I incorrectly suggested you add. The method _can_ return `null` for a non-`null` input (an empty string).
- Update the PR description to automatically close four bugs when this is merged.

I've given this a fair bit of testing and suggest we merge it ASAP.

AraHaan:
merge it now please.

PR #6714, opened 2020-10-28
Add Edit Project File command to menu Project in Nexus client.

ocallesp:
Fixes [AB#1218044](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1218044)

Add missing command.
image

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
@@ -74,6 +74,7 @@
       <IDSymbol name="idmWSE_ContextMenu" value="0x0002"/>
     </GuidSymbol>
 
+    <!-- Guid and IDs copied from Visual Studio Shell -->
     <GuidSymbol name="guidManagedProjectSystemClientMainMenuCmdSet" value="{d309f791-903f-11d0-9efc-00a0c911004f}">
       <IDSymbol name="groupidCtxtMenuEditProjectFile" value="0x0237"/>
     </GuidSymbol>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
@@ -53,6 +53,10 @@
       <Parent guid="guidManagedProjectSystemClientProjectCommandSet" id="groupidEditProjectFile" />
     </CommandPlacement>
 
+    <CommandPlacement guid="guidManagedProjectSystemClientProjectCommandSet" id="cmdidEditProjectFile" priority="0x0100">
+      <Parent guid="guidManagedProjectSystemClientMainMenuCmdSet" id="groupidCtxtMenuEditProjectFile"/>
+    </CommandPlacement>
+
   </CommandPlacements>
 
   <Symbols>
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
```diff
@@ -69,6 +73,10 @@
     <GuidSymbol name="guidWorkspaceExplorerToolWindowPackageCmdSet" value="{cfb400f1-5c60-4f3c-856e-180d28def0b7}">
       <IDSymbol name="idmWSE_ContextMenu" value="0x0002"/>
     </GuidSymbol>
+
+    <GuidSymbol name="guidManagedProjectSystemClientMainMenuCmdSet" value="{d309f791-903f-11d0-9efc-00a0c911004f}">
+      <IDSymbol name="groupidCtxtMenuEditProjectFile" value="0x0237"/>
```
tmeschter, line 78: Can you add a comment explaining where else these two values are defined?

jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
ocallesp: done

davkean:
This doesn't look correct. The shell has already placed these commands in the right places, they aren't showing up because the command hasn't been opt'd into Nexus (which you had a PR out for this here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/282545) and because we're not handling cmdidEditProjectFile: https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/SolutionExplorer/ProjectNodeCommandHandler.cs#L50.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
```diff
@@ -69,6 +73,11 @@
     <GuidSymbol name="guidWorkspaceExplorerToolWindowPackageCmdSet" value="{cfb400f1-5c60-4f3c-856e-180d28def0b7}">
       <IDSymbol name="idmWSE_ContextMenu" value="0x0002"/>
     </GuidSymbol>
+
+    <!-- Guid and IDs copied from Visual Studio Shell -->
+    <GuidSymbol name="guidManagedProjectSystemClientMainMenuCmdSet" value="{d309f791-903f-11d0-9efc-00a0c911004f}">
```
davkean, line 78: guidSHLMainMenu would be a better description of this. This GUID has nothing to do with us.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS.Client/Menus.vsct
davkean: Next time, please use the names as they are defined within the shell, for example, IDG_VS_CTXT_PROJECT_EDITFILE.

ocallesp: I wanted to use the original name, but it was going to look different from the rest of the file.

PR #6715, opened 2020-10-28
Use FWLinks for HelpUrls

### LINKED ISSUE #6541: Update HelpUrl links to use FWLink
A number of the properties in the .xaml files for our project property pages specify `HelpUrl` attributes with links to more information. We need to replace all of these with FWLinks that redirect to the proper pages. This way we don't need to update the product when these pages move to new locations or have significant changes; we can instead just update the FWLinks.

tmeschter:
Fixes #6541.

Properties declared in .xaml `Rule` files may include a `HelpUrl` field providing the web address of documentation for that property. Here we update the existing `HelpUrl`s to go through our redirection service (FWLink). This will allow us to update the URLs in the future without needing to service the product; we can just update the entry in the FWLink redirection service instead.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6715)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
@@ -24,17 +24,17 @@
   <StringProperty Name="PackageId"
                   DisplayName="Package ID"
                   Description="The case-insensitive package identifier, which must be unique across nuget.org or whatever gallery the package resides in. IDs may not contain spaces or characters that are not valid for a URL, and generally follow .NET namespace rules."
-                  HelpUrl="https://docs.microsoft.com/nuget/create-packages/creating-a-package#choose-a-unique-package-identifier-and-setting-the-version-number" />
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147131" />
 
   <StringProperty Name="Version"
                   DisplayName="Package Version"
                   Description="The version of the package, following the major.minor.patch pattern. Version numbers may include a pre-release suffix."
-                  HelpUrl="https://docs.microsoft.com/nuget/concepts/package-versioning" />
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147132" />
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -74,27 +74,27 @@
...
                   DisplayName="Repository URL"
                   Description="Specifies the URL for the repository where the source code for the package resides and/or from which it's being built. For linking to the project page, use the 'Project URL' field, instead."
-                  HelpUrl="https://docs.microsoft.com/nuget/reference/nuspec#repository" />
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147241" />
 
   <StringProperty Name="RepositoryType"
                   DisplayName="Repository type"
                   Description="Specifies the type of the repository. Default is 'git'."
-                  HelpUrl="https://docs.microsoft.com/nuget/reference/nuspec#repository" />
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147241" />
```
jjmew, line 87: >HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147241" /> [](start = 18, length = 60)

This is the same than the one above. Is this by design?

tmeschter: Yes; the same section of documentation covers both.

dmonroym (approved):

drewnoakes (approved):
Who has permission to update these links? Is there any chance someone leaves the company and leaves the future owners in a lurch?

A consequence of using FWLinks is that the URL we currently show in the UI becomes less scrutable. This is a tool tip on the help icon (the mouse isn't visible in the screenshot):

image

Perhaps we change the tool tip text to something like "Click to open documentation about this property in a browser.".

drewnoakes:
> Who has permission to update these links?

Ah, missed the conversation on Teams.

PR #6722, opened 2020-10-29
Add null pointer check for startupProjectsListService.

ocallesp:
GetValueAsyn() might return null if the service is not present.

[AB#1229073](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1229073)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6722)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
@@ -1,7 +1,5 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-#nullable enable
-
 using System;
 using System.ComponentModel.Composition;
 using System.Threading;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
@@ -1,5 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+#nullable enable
+
 using System;
 using System.ComponentModel.Composition;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -1,5 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+#nullable enable
+
```
drewnoakes, line 4:
```suggestion
```

This shouldn't be needed. This project enables nullable by default.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/StartupProjectRegistrar.cs
```diff
@@ -87,17 +89,17 @@ internal Task OnProjectChangedAsync(IProjectVersionedValue<IProjectSubscriptionU
             {
                 bool isDebuggable = await IsDebuggableAsync();
 
-                IVsStartupProjectsListService startupProjectsListService = await _startupProjectsListService.GetValueAsync();
+                IVsStartupProjectsListService? startupProjectsListService = await _startupProjectsListService.GetValueAsync();
```
drewnoakes, line 92: If this can return `null`, then it should be annotated correctly, which would have caught this.

One option is to define `_startupProjectsListService` as having type `IVsService<IVsStartupProjectsListService?>`.

The other option is to annotate `IVsService<T>.GetValueAsync(...)` as returning `Task<T?>`. @davkean I think we discussed this before and the thinking was that most services are expected to be present. However the current approach means we have to remember to add a nullable type parameter here. There are only 14 usages of that method in our solution. Now that we've had more time with C# nullable reference types, do you feel differently? I suggest we change the return to nullable and introduce `Assumes.Present` calls for services we expect to be always present. This will create a safer default behaviour for new usages of `IVsService<>`.

davkean: I would like to figure out why this returns null, instead of hiding it and introducing a bug where we no longer show up in the Startup list. This would affect every single project in the solution.

This service should always be present, we need to figure out why its not.

davkean: I still feel strongly that this should be non-nullable in most cases. If we checked for null here, we would be hiding a bug and telemetry wouldn't be telling us about it.

ocallesp:
More investigation is needed to find the root cause.

Handling this null pointer might hide another weird bug.

Issue #6726, opened 2020-11-02
Consider using project snapshot instead of live project in WindowsFormsEditorProvider 


lifengl:
The code read SubType from live project (through project lock), and it is shown in UI delay stack in some high contention scenario (like common imported msbuild file is updated).

It is called item preview logic during selection change event, which blocks the UI thread to evaluate the project. It is also a potential problem if we load project with a cached project snapshot.

![image](https://user-images.githubusercontent.com/11638466/97824333-75035680-1c70-11eb-9c6f-b1374093b05f.png)

PR #6734, opened 2020-11-05
Use snapshot data in WindowsFormsEditorProvider

davkean:
Fixes: #6726

Under high project lock contention, such as during a branch switch or a change to a props/targets shared by an entire solution, we can hit UI delays selecting items in Solution Explorer trying to figure out if we can "preview" files.

Use snapshot data if available in WindowsFormsEditorProvider to prevent that.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6734)


jjmew (approved):
:shipit:

PR #6736, opened 2020-11-05
Code cleanup - handle key not found

ocallesp:
The map Properties may not contain the key value.
The current logic checks for existing key is present in ChangedProperties, but it gets the value from Properties causing to throw an exception.

This change only adds logic to check if Properties contains the key before getting the value.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6736)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectFilePathAndDisplayNameEvaluationHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectFilePathAndDisplayNameEvaluationHandler.cs
@@ -41,9 +41,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
 
             VerifyInitialized();
 
-            if (projectChange.Difference.ChangedProperties.Contains(ConfigurationGeneral.MSBuildProjectFullPathProperty))
+            if (projectChange.Difference.ChangedProperties.Contains(ConfigurationGeneral.MSBuildProjectFullPathProperty) &&
+                projectChange.After.Properties.TryGetKey(ConfigurationGeneral.MSBuildProjectFullPathProperty, out string projectFilePath))
             {
-                string projectFilePath = projectChange.After.Properties[ConfigurationGeneral.MSBuildProjectFullPathProperty];
                 string displayName = GetDisplayName(projectFilePath);
 
                 logger.WriteLine("DisplayName: {0}", displayName);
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/ProjectFilePathAndDisplayNameEvaluationHandler.cs
```diff
@@ -41,9 +41,9 @@ public void Handle(IComparable version, IProjectChangeDescription projectChange,
 
             VerifyInitialized();
 
-            if (projectChange.Difference.ChangedProperties.Contains(ConfigurationGeneral.MSBuildProjectFullPathProperty))
+            if (projectChange.Difference.ChangedProperties.Contains(ConfigurationGeneral.MSBuildProjectFullPathProperty) &&
+                projectChange.After.Properties.TryGetKey(ConfigurationGeneral.MSBuildProjectFullPathProperty, out string projectFilePath))
```
drewnoakes, line 45:
```suggestion
                projectChange.After.Properties.TryGetValue(ConfigurationGeneral.MSBuildProjectFullPathProperty, out string projectFilePath))
```

jjmew (approved):

davkean:
This is a pretty important property, without it, language service integration will be broken. I'll like to understand the cause a little more. Please hold off while I investigate this. I suspect a broken project file in which case I would like instead to embed the rule so that this never happens.

ocallesp:
> This is a pretty important property, without it, language service integration will be broken. I'll like to understand the cause a little more. Please hold off while I investigate this. I suspect a broken project file in which case I would like instead to embed the rule so that this never happens.

@davkean Do you mean Embedding the xaml in your assembly and exposing it via a MEF export ?
I see `ConfigurationGeneral.xaml` is already in [Microsoft.VisualStudio.ProjectSystem.Managed.csproj](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj#L328)

davkean:
This is a fix that fixes the underlying cause; the rule missing. https://github.com/dotnet/project-system/pull/6740

davkean:
@ocallesp Was this deliberately merged in? Can you revert it in lieu of the the change that fixed the underlying change: https://github.com/dotnet/project-system/pull/6740?

PR #6740, opened 2020-11-06
Embed and stop localizing ConfigurationGeneral.xaml

davkean:
- We only localize rules intended for display in UI. This rule is not displayed in a UI and hence we'll stop localizing it.
- This embeds configurationgeneral and stops picking up the version from the design-time targets, which avoid situations like in [1228912](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1228912) where it doesn't exist.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6740)


ocallesp (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -123,30 +116,24 @@
...
-    <EnumValue Name="AppContainerExe"
-               DisplayName="Windows Store app" />
-    <EnumValue Name="WinMDObj"
-               DisplayName="WinMD" />
+                Visible="False" >
+    <EnumValue Name="Library" />
+    <EnumValue Name="Exe" />
+    <EnumValue Name="WinExe" />
+    <EnumValue Name="AppContainerExe" />
+    <EnumValue Name="WinMDObj" />
```
drewnoakes, line 136: This loses some information. It's not clear from the name that `AppContainerExe` is a Windows Store app.

```suggestion
    <EnumValue Name="Library" />
    <EnumValue Name="Exe" />                 <!-- Console app -->
    <EnumValue Name="WinExe" />
    <EnumValue Name="AppContainerExe" />     <!-- Windows Store app -->
    <EnumValue Name="WinMDObj" />
```

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
davkean: AppContainerExe is exe that lives in an "AppContainer", Windows Store app is a concept that existed in Windows 8 timeframe that doesn't exist. There is no better description other than "used by UWP projects"

PR #6741, opened 2020-11-06
Add some property page search terms

drewnoakes:
Increases the chance of finding properties using related words or alternate spellings.

It seems SearchTerms metadata is not localised. This defeats their purpose in other cultures. @tmeschter do you have a sense of what's involved here? Should I file an issue for this?

I'd like some examples of this feature in the data for testing/demo purposes. I hit this organically typing "optimise" and getting different results to "optimize".

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6741)


tmeschter:
Yes, please file an issue in our repo and tag me in it, and I'll update it with what needs to be done to support localization when I get back from vacation.

drewnoakes:
@tmeschter, I filed #6742.

tmeschter (approved):

melytc (approved):

Issue #6627, opened 2020-09-23
Project Property Pages


drewnoakes:
Tracking issue for work on the Project Property Pages UI and back end. The work is split between this repo, CPS and the VS repo.

---

# Project properties

- [x] Review backlog for additional properties (#6720)

### Essential for Preview

- [x] Model display name/underlying value for enums <kbd>0.2</kbd>
- [x] Allow enum property's value to not be in the set of supported values <kbd>0.5</kbd>
- [x] Complete support for essential editor types <kbd>0.5</kbd>
- [x] Support async loading of backing data <kbd>0.5</kbd>
- [x] Get new editor to work in non-Codespaces scenario, for easier development <kbd>1</kbd>
- [x] Add and use a new feature flag <kbd>2</kbd>
- [x] Move code to CPS and get it building <kbd>1</kbd>
  - [x] Use platform's `HighlightTextBlock` <kbd>0.2</kbd>
- [x] Call Project Query API and map data to model <kbd>2</kbd>
- [x] Use VS's image catalog instead of bundling images <kbd>0.5</kbd>
- [x] Reduce property values across invariant dimensions <kbd>1</kbd>
- [x] Integrate help URI <kbd>0.5</kbd>
- [x] Enforce dimension ordering in UI when multiple selected <kbd>0.2</kbd>
- [x] Add support for sending property updates to the server <kbd>2</kbd>
  - [x] Call server
  - [x] Control when update sent (on update, on lose focus, on <kbd>Enter</kbd> etc)
- [x] Conditional property visibility ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/287643))
  - [x] Model in rule files
  - [x] Evaluate condition expressions on client and update visibility
  - [ ] ~Extend visibility conditions~ (moved to #6907)
    - [ ] ~`(is-sdk-style)`~ (moved to #6907)
    - [ ] ~`(has-capability "GenerateDocumentationFile")` ([SDK source](https://github.com/dotnet/sdk/blob/955c0fc7b06e2fa34bacd076ed39f61e4fb61716/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L262-L265))~ (moved to #6907)
    - [ ] ~`(supports-lang-version 8.0)`~ (moved to #6907)
  - [ ] ~Custom functions via MEF?~ (moved to #6895)
- [x] Lock dependent properties during updates and refresh them
[...]

PR #6746, opened 2020-11-10
Make use of VisibilityCondition in the new project property pages

tmeschter:
Add `VisibilityCondition` metadata to our new project property pages.

Now that support for evaluating the `VisibilityCondition` expressions is in CPS, we can make use of it here. This involves adding the expressions themselves to the .xaml `Rule` files of course, but a larger part of the change involves the addition and handling of various "synthetic" properties referenced in the expressions.

By "synthetic" we mean that the properties have no backing storage of their own, and thus aren't meaningful properties in and of themselves. For example, we want to have a "Target multiple frameworks" check box; when it is checked, the "Target framework" drop down is hidden and a "Target frameworks" text box is shown instead. We need a .xaml property for that check box, but its value isn't stored to the project.

However, just because the value isn't stored doesn't mean the back end support for the property is simple. We need to intercept the reads and writes of the property appropriately. And in many cases we want the synthetic property to affect the values of other properties in complex ways. Going back to the "Target multiple frameworks" check box, when it is toggled on we want to copy the current `TargetFramework` into `TargetFrameworks` and then delete `TargetFramework`, and vice versa when it is turned off. Much of this change is the addition of interceptors for various properties that work in similar ways.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6746)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -32,6 +32,18 @@
...
 
+  <BoolProperty Name="TargetMultipleFrameworks"
+                DisplayName="Target multiple frameworks"
+                Category="General">
+    <BoolProperty.DataSource>
+      <DataSource Persistence="ProjectFileWithInterception"
+                  HasConfigurationCondition="False" />
+    </BoolProperty.DataSource>
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="TargetFrameworkMoniker;TargetFrameworks" />
```
drewnoakes, line 43: What do we do about dependencies across pages?

1. Assume names are unique across pages
2. Require the page be specified via some syntax (e.g. `Application::TargetFrameworkMoniker`)
3. Something else

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -44,8 +56,23 @@
                   Persistence="ProjectFileWithInterception"
                   HasConfigurationCondition="False"/>
     </DynamicEnumProperty.DataSource>
+    <DynamicEnumProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition" Value="(not (evaluated TargetMultipleFrameworks))" />
```
drewnoakes, line 60: Again, I think we need to specify the property's page.

```suggestion
      <NameValuePair Name="VisibilityCondition" Value="(not (evaluated Application TargetMultipleFrameworks))" />
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/TargetMultipleFrameworksValueProvider.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties
+{
+    [ExportInterceptingPropertyValueProvider("TargetMultipleFrameworks", ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class TargetMultipleFrameworksValueProvider : InterceptingPropertyValueProviderBase
```
drewnoakes, line 9: Returned tasks here could be cached on [`TaskResult`](https://github.com/dotnet/project-system/blob/c1a5bb26a42a83500c87529bdb3d47e1fe575001/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/TaskResult.cs).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/TargetMultipleFrameworksValueProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public override Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+        {
+            return Task.FromResult<string?>(null);
+        }
+
+        public override Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
+        {
+            return Task.FromResult(
+                string.IsNullOrEmpty(evaluatedPropertyValue)
```
drewnoakes, line 19: If this parameter (or the uneval equivalent below) may be `null`, we should annotate `InterceptingPropertyValueProviderBase`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -32,6 +32,18 @@
                   Description="Specifies the base namespace for files added to the project."
                   Category="General" />
 
+  <BoolProperty Name="TargetMultipleFrameworks"
+                DisplayName="Target multiple frameworks"
+                Category="General">
```
drewnoakes, line 37: Bool properties should have a description. We use it as the label on the check box. Without this, the header is used instead and no header is visible for the property.

```suggestion
  <BoolProperty Name="TargetMultipleFrameworks"
                DisplayName="Target multiple frameworks"
                Description="Build this project for multiple target frameworks."
                HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147236"
                Category="General">
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -44,8 +56,23 @@
...
                   HasConfigurationCondition="False"/>
     </DynamicEnumProperty.DataSource>
+    <DynamicEnumProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition" Value="(not (evaluated TargetMultipleFrameworks))" />
+    </DynamicEnumProperty.Metadata>
   </DynamicEnumProperty>
 
+  <StringProperty Name="TargetFrameworks"
+                  DisplayName="Target frameworks"
+                  Category="General">
```
drewnoakes, line 66:
```suggestion
  <StringProperty Name="TargetFrameworks"
                  DisplayName="Target frameworks"
                  Description="Specifies the semicolon-delimited list of frameworks that this project will target."
                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147236"
                  Category="General">
```

tmeschter:
@drewnoakes @melytc It occurs to me that the correct functioning of the `VisibilityCondition` depends a lot on when the client sends changes back to the server. For example, say the project file currently has `<TargetFramework>` defined, and the user opens up the property page. The "Target framework" drop down will show as usual, and "Target multiple frameworks" will be unchecked.

Then the user checks "Target multiple frameworks". If the client sends that change right away *and* requeries all the properties, then the check box will immediately go back to being unchecked. If, however, we wait for an explicit save then the user has a change to set the list of target frameworks first.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
tmeschter: Do we have any examples of cross-page dependencies? Or even possible cross-page dependencies? Arguably, if there is a dependency relationship between two properties then maybe they should be on the same page.

What if we just didn't support cross-page dependencies or visibility conditions for v1?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
tmeschter: *If* we support cross-page visibility conditions I would propose we support both of the following:

`(evaluated PropertyName)` returns the evaluated value of the property PropertyName on the same page.
`(evaluated PageName PropertyName)` returns the evaluated value of the property PropertyName on the page PageName.

tmeschter: To broaden this topic of discussion, what if we supported the following?

1. `Value="TargetMultipleFrameworks"`: produces the evaluated value of "TargetMultipleFrameworks" from the same page.
2. `Value="(evaluated TargetMultipleFrameworks)"`: produces the evaluated value of "TargetMultipleFrameworks" from the same page.
3. `Value="(unevaluated TargetMultipleFrameworks)"`: produces the unevaluated value of "TargetMultipleFrameworks" from the same page.
4. `Value="(evaluated OtherPage TargetMultipleFrameworks)"`: produces the evaluated value of "TargetMultipleFrameworks" from the "OtherPage" page.
5. `Value="(unevaluated OtherPage TargetMultipleFrameworks)"`: produces the unevaluated value of "TargetMultipleFrameworks" from the "OtherPage" page.

The idea here is to simplify the most common case (getting an evaluated value from a property on the same page), while still leaving open the option to get unevaluated values, or values from other pages.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: We have examples of this. For example, changing `AssemblyName` impacts the `Packaging` page. We cannot assume or enforce relationships existing only within a single page.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: I believe we should support cross-page visibility conditions.

A motivating example of something I would like to add based on customer feedback is a bool property that forces a single "version" across the project (file, assembly, package, etc). This would likely be on the Application page, and could hide something on the Packaging page.

My preference is to require the page name in the expression. It makes the evaluation simpler, as there's no per-property context, so we can reuse the same environment across all properties when we update visibilities, avoiding a bunch of GC.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
tmeschter: Ah, yes. I remember that one now.

OK, I would propose `PageName:PropertyName` (with just one color) for the syntax in `DependsOn`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
tmeschter: Hmm, fair enough. OK, we can always add support for the simplified forms later. I'll update the PR.

drewnoakes: Your second comment wasn't visible to me when I wrote my reply.

I'm not sure we need to simplify the expressions here. Having a consistent form will help the occasional unfamiliar author stumble upon the correct pattern.

I'd vote for using the same syntax for qualified property names in both DependsOn and these expressions.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: I don't feel strongly either way, but `::` is a common operator for resolving scopes (C#, Java, C++, Rust, PHP, Ruby).

drewnoakes:
What if setting "Target multiple frameworks" actually changed the project, moving values between `<TargetFramework>` and `<TargetFrameworks>`? A downside to this would be that unchecking then rechecking the property would remove all but one target framework.

Alternatively we could add metadata to the property that causes the client to not requery that particular property on update. Conceptually this would indicate that the client owns the state of this property, and only queries it once at startup.

drewnoakes:
What do we want to do about properties which may have multiple values? In the prototype we had simple conditions (some property has some value) and we consider it true if _any_ of that property's configured values match the specified value.

Using `(eq (evaluated Page::Prop) "My Value")` doesn't allow for that, as the `evaluated` function would return a single value. For some properties, we may have multiple values.

Perhaps `(has-evaluated-value Page::Prop "My Value")` is a better fit, as it can handle multiple values of a property.

tmeschter:
> What if setting "Target multiple frameworks" actually changed the project, moving values between <TargetFramework> and <TargetFrameworks>? A downside to this would be that unchecking then rechecking the property would remove all but one target framework.

That could be done. `TargetMultipleFrameworksValueProvider` could even store the previous values of `TargetFramework` and `TargetFrameworks` so that they could be restored when toggling back and forth.

This would mean that `TargetMultipleFrameworks` would DependOn both `TargetFrameworkMoniker` and `TargetFrameworks`, and they would in turn DependOn `TargetMultipleFrameworks`. Would that pose any problems for the UI?

tmeschter:
> Perhaps (has-evaluated-value Page::Prop "My Value") is a better fit, as it can handle multiple values of a property.

What if we supported both `has-evaluated-value` and `evaluated`, but it would be an error to use `evaluated` with a configuration-dependent property? Otherwise, we lose the ability to use the values of boolean properties directly and will always have to explicitly compare them to true or false. E.g.,

`(not (evaluated Application::TargetMultipleFrameworks))`

would have to become

`(not (has-evaluated-value Application::TargetMultipleFrameworks true))`

That's workable, but not great for properties that can only have one value.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -32,20 +32,75 @@
...
+    <StringProperty.DataSource>
+      <DataSource PersistedName="TargetFrameworksLink"
+                  Persistence="ProjectFileWithInterception"
+                  HasConfigurationCondition="False" />
+    </StringProperty.DataSource>
+    <StringProperty.ValueEditors>
+      <ValueEditor EditorType="LinkAction">
+        <ValueEditor.Metadata>
+          <NameValuePair Name="Action" Value="URL" />
+          <NameValuePair Name="URL" Value="https://go.microsoft.com/fwlink/?linkid=2147236" />
```
tmeschter, line 93: @drewnoakes One thing I found myself wanting was a way to set the link text separate from the title or description. That is, I'd like it if the control showed the title, then the description, then the link with the link text coming from ValueEditor metadata. That may be something to come back to later as it would require more loc support.

drewnoakes: For the case where we only want to show a hyperlink, using a separate field for that string would require setting an empty display name. I can imagine possible negative knock-on effects from that.

drewnoakes:
The final commit e2a713cfdc893159a3c62a9e056234ffec342b07 two breaking issues:

1. It inverts the direction of inter-property dependencies. Property A indicates it depends upon property B, meaning A refreshes when B is updated. This allows properties to be added in future, potentially via other mechanisms, and to refer to properties that don't know it exists.

2. It uses an invalid form of `DependsOn` which will throw exceptions now that https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/287643 has merged. We are consuming this value today.

Can we hold off on removing the target framework check box in e2a713cfdc893159a3c62a9e056234ffec342b07 for now? I'd like people to that UI during dogfooding. @davkean has indicated he may be able to unblock the scenario, and we still have plenty of time to do so.

In our sync meeting today @tmeschter expressed concern we could forget to remove this if it has significant bugs. Let me just say that I won't let that happen. It's captured on #6627 and is something I'm personally quite keen to address before shipping. I'll also capture a work item to produce test scripts for CTI.

tmeschter:
@drewnoakes Sure, I can back out that change.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.Properties
+{
+    [ExportInterceptingPropertyValueProvider("ResourceSpecificationKind", ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class ResourceSpecificationKindValueProvider : InterceptingPropertyValueProviderBase
+    {
+        // TODO should the rule file generate property and enum value constants that we can use here instead of these string literals?
+
+        public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+        {
+            if (unevaluatedPropertyValue == "Win32Resource")
```
tmeschter, line 15: Should this be "IconAndFile"?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                await defaultProperties.DeletePropertyAsync("PackageLicenseFile");
```
tmeschter, line 17: And should this be "Win32Resource"?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
```diff
@@ -0,0 +1,128 @@
...
+            else if (string.Equals(unevaluatedPropertyValue, NoManifestValue, StringComparison.InvariantCultureIgnoreCase))
+            {
+                await SaveCurrentApplicationManifestValueAsync(defaultProperties);
+
+                await defaultProperties.DeletePropertyAsync(ApplicationManifestMSBuildProperty);
+                await defaultProperties.SetPropertyValueAsync(NoManifestMSBuildProperty, "true");
+            }
+            else if (string.Equals(unevaluatedPropertyValue, CustomManifestValue, StringComparison.InvariantCultureIgnoreCase))
+            {
+                await RestoreApplicationManifestValueAsync(defaultProperties);
```
tmeschter, line 93: @drewnoakes What do you think of this approach of storing the previous values of properties when we clear them, and then restoring them later when/if the user re-enables them?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
drewnoakes: Yes it looks like I put this and the one below backwards. Can you fix this up, or give me write access to your fork (I don't often force push, but when I do I always use with-lease 😄).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
drewnoakes: I haven't reviewed the code yet but I like the idea in principle. I worry that users won't expect toggling settings to have destructive changes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <summary>
+        /// Save the current value of the "ApplicationManifest" MSBuild property, if any.
+        /// </summary>
+        private async Task SaveCurrentApplicationManifestValueAsync(IProjectProperties defaultProperties)
+        {
+            string? currentApplicationManifestPropertyValue = await defaultProperties.GetUnevaluatedPropertyValueAsync(ApplicationManifestMSBuildProperty);
+            if (!string.IsNullOrEmpty(currentApplicationManifestPropertyValue))
+            {
+                _temporaryPropertyStorage.AddOrUpdatePropertyValue(ApplicationManifestMSBuildProperty, currentApplicationManifestPropertyValue!);
+            }
```
drewnoakes, line 111: Using `Strings` here avoids the need for `!`.

```suggestion
            if (!Strings.IsNullOrEmpty(currentApplicationManifestPropertyValue))
            {
                _temporaryPropertyStorage.AddOrUpdatePropertyValue(ApplicationManifestMSBuildProperty, currentApplicationManifestPropertyValue);
            }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
tmeschter: I'll fix it up.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
tmeschter: Yeah, that was my motivation: you toggle a check box or change a drop down just to see what the other options are, and now you've lost your setting.

drewnoakes:
@tmeschter this looks good to me. Can we get it merged so it inserts and unbreaks dogfooding?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
drewnoakes: Took a deeper look at the approach and it looks good. So long as the set of properties applicable under each of some parent's enum values are mutually exclusive, this approach should work well as far as I can tell.

drewnoakes (approved):

PR #6748, opened 2020-11-11
Remove/fix suppressions

davkean:
This ensures that Debug/Release are free from any code analysis warnings.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6748)


swesonga (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/TypeForwards.cs
```diff
@@ -2,14 +2,10 @@
 
 using System.Runtime.CompilerServices;
 
-#pragma warning disable RS0016 //  Type Forwarding may be removed in the future, when we accept breaking changes.
```
drewnoakes, line 5: Looks like this suppression is needed (or we add these to the API files).

davkean: I think this is a case of VS analysis being different to the compiler, trying to see if updating the compiler fixes this.

PR #6754, opened 2020-11-18
Temporarily save property values

tmeschter:
Temporarily save certain property values when:
- toggling between icons/manifests and resource files
- toggling documentation file generation
- toggling between package license files and package license expressions

The idea here is that these operations should be non-destructive. That is, if you toggle documentation file generation off, and then back on, we should also restore the provided doc file path (if any). Similarly, if you switch from a package license expression to a file and then back, we should restore the license expression.

We accomplish this by saving the values of the properties to `ITemporaryPropertyStorage` before deleting them, and restore them if/when they later become applicable again.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6754)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
@@ -76,22 +76,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         {
             if (string.Equals(unevaluatedPropertyValue, DefaultManifestValue, StringComparison.InvariantCultureIgnoreCase))
             {
-                await SaveCurrentApplicationManifestValueAsync(defaultProperties);
-
+                await defaultProperties.RememberValueIfCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(ApplicationManifestMSBuildProperty);
                 await defaultProperties.DeletePropertyAsync(NoManifestMSBuildProperty);
             }
             else if (string.Equals(unevaluatedPropertyValue, NoManifestValue, StringComparison.InvariantCultureIgnoreCase))
             {
-                await SaveCurrentApplicationManifestValueAsync(defaultProperties);
-
+                await defaultProperties.RememberValueIfCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(ApplicationManifestMSBuildProperty);
                 await defaultProperties.SetPropertyValueAsync(NoManifestMSBuildProperty, "true");
             }
             else if (string.Equals(unevaluatedPropertyValue, CustomManifestValue, StringComparison.InvariantCultureIgnoreCase))
             {
-                await RestoreApplicationManifestValueAsync(defaultProperties);
-
+                await defaultProperties.RestoreValueIfNotCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(NoManifestMSBuildProperty);
             }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
@@ -1,6 +1,5 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
@@ -56,7 +55,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
             }
 
             string noManifestPropertyValue = await defaultProperties.GetEvaluatedPropertyValueAsync(NoManifestMSBuildProperty);
-            if (noManifestPropertyValue?.Equals("true", StringComparison.InvariantCultureIgnoreCase) == true)
+            if (StringComparers.PropertyLiteralValues.Equals(noManifestPropertyValue, "true"))
             {
                 return NoManifestValue;
             }
@@ -74,52 +73,26 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
             IProjectProperties defaultProperties,
             IReadOnlyDictionary<string, string>? dimensionalConditions = null)
         {
-            if (string.Equals(unevaluatedPropertyValue, DefaultManifestValue, StringComparison.InvariantCultureIgnoreCase))
+            if (StringComparers.PropertyLiteralValues.Equals(unevaluatedPropertyValue, DefaultManifestValue))
             {
-                await defaultProperties.RememberValueIfCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
+                await defaultProperties.SaveValueIfCurrentlySetAsync(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(ApplicationManifestMSBuildProperty);
                 await defaultProperties.DeletePropertyAsync(NoManifestMSBuildProperty);
             }
-            else if (string.Equals(unevaluatedPropertyValue, NoManifestValue, StringComparison.InvariantCultureIgnoreCase))
+            else if (StringComparers.PropertyLiteralValues.Equals(unevaluatedPropertyValue, NoManifestValue))
             {
-                await defaultProperties.RememberValueIfCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
+                await defaultProperties.SaveValueIfCurrentlySetAsync(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(ApplicationManifestMSBuildProperty);
                 await defaultProperties.SetPropertyValueAsync(NoManifestMSBuildProperty, "true");
             }
-            else if (string.Equals(unevaluatedPropertyValue, CustomManifestValue, StringComparison.InvariantCultureIgnoreCase))
+            else if (StringComparers.PropertyLiteralValues.Equals(unevaluatedPropertyValue, CustomManifestValue))
             {
-                await defaultProperties.RestoreValueIfNotCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
+                await defaultProperties.RestoreValueIfNotCurrentlySetAsync(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(NoManifestMSBuildProperty);
             }
 
             // We don't want to store a value for this so return null.
             return null;
-        }
-
-        /// <summary>
-        /// Save the current value of the "ApplicationManifest" MSBuild property, if any.
-        /// </summary>
-        private async Task SaveCurrentApplicationManifestValueAsync(IProjectProperties defaultProperties)
-        {
-            string? currentApplicationManifestPropertyValue = await defaultProperties.GetUnevaluatedPropertyValueAsync(ApplicationManifestMSBuildProperty);
-            if (!string.IsNullOrEmpty(currentApplicationManifestPropertyValue))
-            {
-                _temporaryPropertyStorage.AddOrUpdatePropertyValue(ApplicationManifestMSBuildProperty, currentApplicationManifestPropertyValue!);
-            }
-        }
-
-        /// <summary>
-        /// If the "ApplicationManifest" MSBuild property does not currently have a value
-        /// then restore the previously saved value, if any.
-        /// </summary>
-        private async Task RestoreApplicationManifestValueAsync(IProjectProperties defaultProperties)
-        {
-            string? currentApplicationManifestPropertyValue = await defaultProperties.GetUnevaluatedPropertyValueAsync(ApplicationManifestMSBuildProperty);
-            if (string.IsNullOrEmpty(currentApplicationManifestPropertyValue)
-                && _temporaryPropertyStorage.GetPropertyValue(ApplicationManifestMSBuildProperty) is string previousValue)
-            {
-                await defaultProperties.SetPropertyValueAsync(ApplicationManifestMSBuildProperty, previousValue);
-            }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/ProjectPropertiesExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    internal static class ProjectPropertiesExtensions
+    {
+        /// <summary>
+        /// Saves the unevaluated value of <paramref name="propertyName"/>, if any, to <paramref name="storage"/>.
+        /// </summary>
+        public static async Task RememberValueIfCurrentlySet(this IProjectProperties properties, string propertyName, ITemporaryPropertyStorage storage)
+        {
+            string? currentPropertyValue = await properties.GetUnevaluatedPropertyValueAsync(propertyName);
+            if (!Strings.IsNullOrEmpty(currentPropertyValue))
```
swesonga, line 15: >Strings [](start = 17, length = 7)

why Strings here and not in the other method?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
```diff
@@ -1,47 +1,66 @@
...
+        [ImportingConstructor]
+        public ResourceSpecificationKindValueProvider(ITemporaryPropertyStorage temporaryPropertyStorage)
+        {
+            _temporaryPropertyStorage = temporaryPropertyStorage;
+        }
 
         public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
         {
-            if (unevaluatedPropertyValue == "IconAndManifest")
+            if (unevaluatedPropertyValue == IconAndManifestValue)
```
swesonga, line 28: >unevaluatedPropertyValue == IconAndManifestValue [](start = 16, length = 48)

Should there be an explicit string comparison specified instead of the == operator?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/ProjectPropertiesExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties
+{
+    internal static class ProjectPropertiesExtensions
+    {
+        /// <summary>
+        /// Saves the unevaluated value of <paramref name="propertyName"/>, if any, to <paramref name="storage"/>.
+        /// </summary>
+        public static async Task RememberValueIfCurrentlySet(this IProjectProperties properties, string propertyName, ITemporaryPropertyStorage storage)
```
swesonga, line 12: >RememberValueIfCurrentlySet [](start = 33, length = 27)

I think SaveValueIfCurrentlySet would be more descriptive about what's going on when reading the code, but that might just be me

swesonga (approved):
:shipit:

drewnoakes (commented):
Thinking about this some more I am concerned that the approach is not watertight due to race conditions and differing lifetimes.

An example:

1. The user changes `ResourceSpecificationKind` from `IconAndManifest` to `ResourceFile`.
1. The client sends the updated value for that property.
1. The client re-evaluates visibility conditions and _hides_ the icon and manifest properties, however they still have their old values.
1. The server deletes the `ApplicationIcon` and `ApplicationManifest` property values from the project.

At this point, the UI and project are out of sync.

Let's consider the reversion case:

1. The user changes `ResourceSpecificationKind` back to `IconAndManifest`.
1. The client sends the updated value for that property.
1. The server unstashes the saved value.

Now the client and project are back in sync, but this is mostly a coincidence.

This state is fragile and will break:

- when we start pushing updates from the server to the client, or
- if the user closes and re-opens the editor (the lifetime of the editor is not tied to `ITemporaryPropertyStorage`).

These situations seem to be symptoms of not modelling the client's workspace on the server.

Stashing values on the server will force us to decide between:

- the UI state exactly matching the project state, versus
- the UI state matching the project plus any stashed values for hidden properties.

Potential solutions:

1. Have the server return any stashed values in results produced by the Project Query API. This means a newly-opened editor will have stashed values ready to display in hidden properties. This requires stashing both evaluated and unevaluated values. It still fails once we push updates from the server to the client.
1. Move this deletion logic from the server to the client, so it deletes property values that no longer apply in response to user changes. Perhaps we add bool `DeleteValueWhenConditionallyInvisible` metadata. The client can remember and restore the value if it becomes visible again, and the stored value evaporates if the editor is closed. To prevent race conditions we would need a compare-and-set style update function from the server, so the client only restores the value if it still missing.
1. Have every property that becomes conditionally visible update itself under a lock. This is likely to create frustration, interfere with tab navigation, etc.

I'm starting to work on dependent property hook-up, so will gain a deeper understanding of these interactions over the next while. I am concerned however that we have a design which allows data to get out of sync and requires careful management to prevent race conditions between the user and the server. This feels too complicated for a brand new design to me. I would prefer us to design these issues out of the system altogether.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
drewnoakes: Yes, my bad. Should have been `StringComparers.PropertyLiteralValues` (I believe. See [Teams chat](https://teams.microsoft.com/l/message/19:ba67a1e4184a4f1da5ee42813032ead0@thread.skype/1605754601487?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&parentMessageId=1605754601487&teamName=.NET%20Developer%20Experience&channelName=Project%20System&createdTime=1605754601487) about distinction between `PropertyLiteralValues` and `PropertyValues`).

tmeschter:
> Thinking about this some more I am concerned that the approach is not watertight due to race conditions and differing lifetimes.
> 
> An example:
> 
> 1. The user changes `ResourceSpecificationKind` from `IconAndManifest` to `ResourceFile`.
> 2. The client sends the updated value for that property.
> 3. The client re-evaluates visibility conditions and _hides_ the icon and manifest properties, however they still have their old values.
> 4. The server deletes the `ApplicationIcon` and `ApplicationManifest` property values from the project.
> 
> At this point, the UI and project are out of sync.

One problem is that we're missing the proper `DependsOn` metadata for several of these items; I'll address that.

Re-using your example, I think the following would be a workable set of steps and provide a reasonable user experience:

1. The user changes `ResourceSpecificationKind` from `IconAndManifest` to `ResourceFile`.
2. Further edits to `ResourceSpecificationKind` and all properties that depend on it are temporarily blocked (regardless of their visibility).
3. The client re-evaluates visibility conditions, _hides_ the icon and manifest properties, and _shows_ the resource file property, however they still have their old values.
4. The client sends the updated value for that property.
5. The client re-queries the values of the dependent properties.
6. The client re-evaluates visibility conditions again in light of the updated property values.
7. Edits to `ResourceSpecificationKind` and all properties that depend on it are unblocked.

> 
> Potential solutions:
> 
> 1. Have the server return any stashed values in results produced by the Project Query API. This means a newly-opened editor will have stashed values ready to display in hidden properties. This requires stashing both evaluated and unevaluated values. It still fails once we push updates from the server to the client.
> 2. Move this deletion logic from the server to the client, so it deletes property values that no longer apply in response to user changes. Perhaps we add bool `DeleteValueWhenConditionallyInvisible` metadata. The client can remember and restore the value if it becomes visible again, and the stored value evaporates if the editor is closed. To prevent race conditions we would need a compare-and-set style update function from the server, so the client only restores the value if it still missing.
> 3. Have every property that becomes conditionally visible update itself under a lock. This is likely to create frustration, interfere with tab navigation, etc.

I don't favor options 1 and 2; the `IInterceptingPropertyValueProvider` and related types exist to handle these complex interactions between properties and I don't want to reproduce that logic on the client side. Nor do I think we can in the general case; the relationship between properties may be more complex than just "A has a value and B doesn't, or B has a value and A doesn't". That's why we have `DependsOn`: if property _A_ depends on property _B_, then any time the user changes B we *have* to wait for that change to be returned to the server, applied, and the new value of A queried before the client can be certain of the state of the world.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/ProjectPropertiesExtensions.cs
tmeschter: Just an oversight. I'll make it consistent.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
tmeschter: I'll fix that.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/ProjectPropertiesExtensions.cs
tmeschter: 👍🏻

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationManifestKindValueProvider.cs
```diff
@@ -76,22 +76,19 @@ public override async Task<string> OnGetEvaluatedPropertyValueAsync(string prope
         {
             if (string.Equals(unevaluatedPropertyValue, DefaultManifestValue, StringComparison.InvariantCultureIgnoreCase))
             {
-                await SaveCurrentApplicationManifestValueAsync(defaultProperties);
-
+                await defaultProperties.RememberValueIfCurrentlySet(ApplicationManifestMSBuildProperty, _temporaryPropertyStorage);
                 await defaultProperties.DeletePropertyAsync(ApplicationManifestMSBuildProperty);
                 await defaultProperties.DeletePropertyAsync(NoManifestMSBuildProperty);
```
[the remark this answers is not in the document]
tmeschter: It should either be set to `true` or not present; I can't come up with any situations where I would really want to use a saved value.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ResourceSpecificationKindValueProvider.cs
```diff
@@ -1,47 +1,66 @@
...
     internal sealed class ResourceSpecificationKindValueProvider : InterceptingPropertyValueProviderBase
     {
-        // TODO should the rule file generate property and enum value constants that we can use here instead of these string literals?
+        private readonly ITemporaryPropertyStorage _temporaryPropertyStorage;
+
+        private const string Win32ResourceMSBuildProperty = "Win32Resource";
+        private const string ApplicationIconMSBuildProperty = "ApplicationIcon";
+        private const string ApplicationManifestMSBuildProperty = "ApplicationManifest";
+        private const string IconAndManifestValue = "IconAndManifest";
+        private const string ResourceFileValue = "ResourceFile";
```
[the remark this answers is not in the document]
tmeschter: I took a brief look, but couldn't figure out if there was a way to achieve this.

jjmew (approved):
:shipit:

drewnoakes:
> we're missing the proper DependsOn metadata for several of these items

This is an easy mistake to make.

Further, the user can add expression `$(SomeOtherProperty)` in the UI in a property that doesn't depend on `SomeOtherProperty`, meaning we wouldn't correctly refresh the UI when `SomeOtherProperty` changes.

Perhaps we infer dependencies from both visibility conditions and unevaluated values. We would then only need to explicitly state `DependsOn` metadata where there was some other relationship at play. If we want to solve this today, then client would need to do this, as the dependencies would be dynamic. Once we can push updates from the server to the client, the work could be done on the server.

> Re-using your example, I think the following would be a workable set of steps and provide a reasonable user experience:
> 1. The user changes ResourceSpecificationKind from IconAndManifest to ResourceFile.
> 2. Further edits to ResourceSpecificationKind and all properties that depend on it are temporarily blocked (regardless of their visibility).
> 3. The client re-evaluates visibility conditions, hides the icon and manifest properties, and shows the resource file property, however they still have their old values.
> 4. The client sends the updated value for that property.
> 5. The client re-queries the values of the dependent properties.
> 6. The client re-evaluates visibility conditions again in light of the updated property values.
> 7. Edits to ResourceSpecificationKind and all properties that depend on it are unblocked.

If we are going to go this route, I would change step 3 to:

3. The client re-evaluates visibility conditions, hides the icon and manifest properties, and shows the resource file property, ~however they still have their old values~ then refreshes hidden values.

drewnoakes (approved):
I'd like to see the operations I called out run concurrently but won't block merging for that.

tmeschter:
> Further, the user can add expression `$(SomeOtherProperty)` in the UI in a property that doesn't depend on `SomeOtherProperty`, meaning we wouldn't correctly refresh the UI when `SomeOtherProperty` changes.
> 
> Perhaps we infer dependencies from both visibility conditions and unevaluated values. We would then only need to explicitly state `DependsOn` metadata where there was some other relationship at play. If we want to solve this today, then client would need to do this, as the dependencies would be dynamic. Once we can push updates from the server to the client, the work could be done on the server.

If you mean parsing the unevaluated values for MSBuild property references and matching those property references up to properties returned by the Project Query API, that will work in some cases but does not solve the problem in general. For one thing, you'll have the property name but not the page name, so the property may be ambiguous. For another, the MSBuild property name may not even match the Project Query property name and you won't be able to match them up at all.

I don't think there is a fully general solution to this problem short of the server pushing updates.

Also, I think it's fine for the _UI_ to visibility conditions and unevaluated values to augment the `DependsOn` metadata, but I don't want to make it part of the _contract_. If you know that the value of one property surfaced in the UI depends on another, then the author of the .xaml file should capture that in `DependsOn` metadata regardless of anything else.

Also, note that there are places where we have a `VisibilityCondition` that does not imply a dependency, e.g. with the startup object property. We hide that property when the output type is set to "class library", but the value does not depend on the output type. It wouldn't be wrong to infer a dependency between the two, but it also isn't necessary.

> If we are going to go this route, I would change step 3 to:
> 
> 1. The client re-evaluates visibility conditions, hides the icon and manifest properties, and shows the resource file property, ~however they still have their old values~ then refreshes hidden values.

Do you mean simply that the text box for the resource file would be empty at that point? Or do you mean that the client would throw away the value it currently had for those properties? My thought is that those values (hidden or not) would be refreshed as part of step 5. Regardless, I think the most important part of step 3 is that, whatever the resource file control shows, it is disabled until step 7.

PR #6760, opened 2020-11-20
Replace tuples with named types

tmeschter:
Currently, when we create a Project Query API entity (representing a property page, property, etc.) we communicate the important state to the producers of the child entities using a tuple. As we add more producers and further refactor this code it has become clear that the use of a tuple here will lead to errors as different producers may accidentally come to expect different tuple shapes. Remembering the correct shape often means going back to the point where the tuple is created, if you can find it.

This commit replaces the tuples with simple named types.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6760)


jjmew (approved):
:shipit:

PR #6761, opened 2020-11-21
Extract boilerplate

tmeschter:
When asked to retrieve an `IEntityValue` (for a property, page, etc.) given an `EntityIdentity` we always do the same two basic operations: first, we expand the `EntityIdentity` into the relevant pieces. If that works, we then pass those pieces along to another method to actually create the `IEntityValue`. Here we extract out the duplicated code into a base class for the four producers that can handle retrieving a value by key.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6761)


tmeschter:
Yeah, I realize this actually adds more code than it deletes, but it is a stepping stone to further refactorings.

drewnoakes (approved):
Any reason you haven't use a `record` for `KeyData` classes here (also for the new types in PR #6760)?

jjmew (approved):
:shipit:

tmeschter:
> Any reason you haven't use a `record` for `KeyData` classes here (also for the new types in PR #6760)?

I was going to do that. However, in the course of refactoring I realized that the two steps I mentioned in the description could be collapsed into one, and then the `KeyData` types go away completely. That will be in the next PR.

PR #6763, opened 2020-11-23
Create new shared assembly

dmonroym:
It's currently just a placeholder as it doesn't contain nor do anything but we'll eventually need it so adding a stub for now.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6763)


drewnoakes (approved):
Looks good to me.

PR #6770, opened 2020-11-30
Simple Debug page support

tmeschter:
This PR includes all the changes needed to get properties for debug pages to show up in the new project property pages. For background, we already define the ParentDebugPropertyPage.xaml which includes properties for the active launch profile, and the active launch profile's launch target. These will always be shown. In addition to that, we have individual .xaml Rule files for each launch target, specifying all the properties relevant to that target. We also have code that can read and write the values of the active launch profile to the launchSettings.json file.

Our goal in this change is to take the parent Rule and all the child Rules and combine them into a single property page, as exposed through the Project Query API. In practice what this means is re-writing the names of properties and categories on those child pages (along with their entity IDs) to make them appear as if they belong to the parent page. This re-writing also involves embedding the original page name in the category or property name so that we can retrieve the underlying data later.

This change hard-codes in this adjustment everywhere it matters. This is messy. I can imagine better architectures for this, but I expect this support for debug properties in the UI will be temporary, only lasting until we can create a dedicated UI. As such, I don't want to introduce architectural complexity when we don't have a long-term use case for it. To the extent possible, the Debug-page-specific logic has been consolidated to a single file. This should make it easier to remove later.

Beyond that, we also need to tweak the `DependsOn` and `VisibilityCondition` metadata on the properties from the child pages. `DependsOn` is updated to indicate that these properties depend on a couple of well-known properties on the parent page. The `VisibilityCondition` is updated so that the properties for a particular child page will only show when the launch target property (from the parent page) will have the correct value.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6770)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/DebugUtilities.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/DebugUtilities.cs
@@ -6,13 +6,22 @@ using Microsoft.VisualStudio.ProjectSystem.Properties;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 {
+    /// <summary>
+    /// Helper methods to convert back and forth between the page, category, and property names
+    /// used by the Debug page in the Project Query API, and the names used by the underlying
+    /// property system.
+    /// </summary>
     internal static class DebugUtilities
     {
+        private const string CommandNameBasedDebuggerPageTemplate = "commandNameBasedDebugger";
+        private const string DebuggerParentPrefix = "#debuggerParent#";
+        private const string DebuggerChildPrefix = "#debuggerChild#";
+
         public static string ConvertDebugPageNameToRealPageName(string pageName)
         {
-            if (pageName.StartsWith("#debuggerParent#"))
+            if (pageName.StartsWith(DebuggerParentPrefix))
             {
-                return pageName.Substring("#debuggerParent#".Length);
+                return pageName.Substring(DebuggerParentPrefix.Length);
             }
             else
             {
@@ -22,14 +31,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 
         public static string ConvertRealPageNameToDebugPageName(string pageName)
         {
-            return "#debuggerParent#" + pageName;
+            return DebuggerParentPrefix + pageName;
         }
 
         public static (string pageName, string categoryName) ConvertDebugPageAndCategoryToRealPageAndCategory(string pageName, string categoryName)
         {
-            if (categoryName.StartsWith("#debuggerChild#"))
+            if (categoryName.StartsWith(DebuggerChildPrefix))
             {
-                categoryName = categoryName.Substring("#debuggerChild#".Length);
+                categoryName = categoryName.Substring(DebuggerChildPrefix.Length);
 
                 int index = categoryName.IndexOf("#");
                 if (index >= 0)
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/DebugUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+        private const string DebuggerParentPrefix = "#debuggerParent#";
+        private const string DebuggerChildPrefix = "#debuggerChild#";
+
+        public static string ConvertDebugPageNameToRealPageName(string pageName)
+        {
+            if (pageName.StartsWith(DebuggerParentPrefix))
+            {
+                return pageName.Substring(DebuggerParentPrefix.Length);
+            }
+            else
```
MiYanni, line 26: nit: redundant `else` condition.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
```diff
@@ -54,10 +58,19 @@ public static IEntityValue CreatePropertyPageValue(IEntityRuntimeModel runtimeMo
...
             {
-                newPropertyPage.Kind = rule.PageTemplate;
+                if (rule.PageTemplate == "debuggerParent")
+                {
+                    newPropertyPage.Kind = "generic";
+                }
+                else
+                {
+                    newPropertyPage.Kind = rule.PageTemplate;
+                }
```
MiYanni, line 68: nit: can use a ternary here like other assignment statements of this type in the code.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
```diff
@@ -99,16 +120,38 @@ public static IEntityValue CreatePropertyPageValue(IEntityRuntimeModel runtimeMo
...
                     if (projectCatalog.GetSchema(schemaName) is Rule rule
                         && !rule.PropertyPagesHidden)
                     {
-                        IEntityValue propertyPageValue = CreatePropertyPageValue(parent, propertyPageQueryContext, rule, requestedProperties);
-                        yield return propertyPageValue;
+                        if (rule.PageTemplate == "debuggerParent")
+                        {
+                            parentDebuggerPageRule = rule;
+                        }
+                        else if (rule.PageTemplate == "commandNameBasedDebugger")
```
MiYanni, line 135: nit: you can put this condition (inversed) in the outer if-statement so it is ignored upfront.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/DebugUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            return null;
+        }
+
+        public static string? UpdateDebuggerDependsOnMetadata(Rule containingRule, string? dependsOnString)
+        {
+            if (containingRule.PageTemplate == CommandNameBasedDebuggerPageTemplate)
+            {
+                dependsOnString = dependsOnString is not null
```
MiYanni, line 123: `not` 👀 😄 (I'm just happy to see new features of the language being used)

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
tmeschter: Done

jjmew (approved):

PR #6772, opened 2020-12-01
Preserve values of a synthetic property

tmeschter:
In the new project property pages we allow the user to specify if they want to supply separate icon and manifest files for their application, or if they want to supply a resource file containing the information. This selection is done via a drop down. Depending on which option the user chooses, we show or hide the controls for the relevant properties.

This "resource specification kind" is a "synthetic" property that has no representation of its own in the project file (or any other storage). Instead, we assume that if the "Win32Resource" property has a value, the user has opted to use a resource file; otherwise, they have opted to use an icon and manifest.

This currently leads to problems when trying to switch between the two. If the project does not specify the "Win32Resource" property and the user tries to change the drop down from "Icon and Manifest" to "Resource file" the property page will send the new value ("ResourceFile") to the server. It then refreshes its value--and gets back "IconAndManifest". At this point the drop down reverts to its original value, and the controls for the resource file path are hidden.

To fix this, we need to remember the user's choice for the resource specification kind when they have made an explicit change. Here we make use of `ITemporaryProjectStorage` to do so, and we use the following logic when the resource specification kind is retrieved:

1. If the application icon or application manifest are specified, return "IconAndManifest".
2. If the resource file is specified, return "ResourceFile".
3. If none of those properties are specified but we have a saved value, use that.
4. If all else fails, return "IconAndManifest" as the default.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6772)

### DIFF
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/ITemporaryPropertyStorageFactory.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/ITemporaryPropertyStorageFactory.cs
@@ -10,10 +10,7 @@ namespace Microsoft.VisualStudio.ProjectSystem
     {
         public static ITemporaryPropertyStorage Create(Dictionary<string, string>? values = null)
         {
-            if (values is null)
-            {
-                values = new();
-            }
+            values ??= new();
 
             var mock = new Mock<ITemporaryPropertyStorage>();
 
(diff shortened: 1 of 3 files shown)


jjmew (approved):
:shipit:

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/ITemporaryPropertyStorageFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    internal static class ITemporaryPropertyStorageFactory
+    {
+        public static ITemporaryPropertyStorage Create(Dictionary<string, string>? values = null)
+        {
+            if (values is null)
+            {
+                values = new();
+            }
```
drewnoakes, line 16: Could use null coalescing assignment:

```suggestion
            values ??= new();
```

drewnoakes (approved):
Looks good to me! Thanks for sorting this out so quickly.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/ITemporaryPropertyStorageFactory.cs
tmeschter: I always forget about that operator!

PR #6775, opened 2020-12-02
Only set config-independent properties once

tmeschter:
Currently, when setting a property value through the Project Query API we first find all the applicable `ConfiguredProject`s, then bind the requested rule in each, and finally retrieve and set the property in each bound rule.

This makes sense for configuration-dependent properties, where we may want to set the values in some configurations but not others. However, most of our properties are configuration-independent and it doesn't matter which `ConfiguredProject` we go through or how many: ultimately, we'll end up with just one definition in the project file, with no configuration conditions. To avoid unnecessary work we should only set the value of a configuration-independent property once.

This commit adds the necessary checks before we start retrieving the `ConfiguredProject`s.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6775)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
@@ -3,11 +3,9 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Build.Framework.XamlTypes;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
-using Microsoft.VisualStudio.ProjectSystem.Query;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
@@ -79,9 +79,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                         // all matching configurations.
                         if (await propertyPageCache.GetKnownConfigurationsAsync() is IImmutableSet<ProjectConfiguration> knownConfigurations)
                         {
-                            foreach (ProjectConfiguration knownConfiguration in knownConfigurations.Where(config => config.MatchesDimensions(_dimensions)))
+                            foreach (ProjectConfiguration knownConfiguration in knownConfigurations)
                             {
-                                if (await propertyPageCache.BindToRule(knownConfiguration, _pageName) is IRule boundRule)
+                                if (knownConfiguration.MatchesDimensions(_dimensions)
+                                    && await propertyPageCache.BindToRule(knownConfiguration, _pageName) is IRule boundRule)
                                 {
                                     projectRules.Add(boundRule);
                                 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
```diff
@@ -64,21 +65,39 @@ public async Task OnBeforeExecutingBatchAsync(IEnumerable<UnconfiguredProject> t
...
-                        foreach (ProjectConfiguration knownConfiguration in knownConfigurations.Where(config => config.MatchesDimensions(_dimensions)))
+                        // The property is configuration-dependent; we need to collect the bound rules for
+                        // all matching configurations.
+                        if (await propertyPageCache.GetKnownConfigurationsAsync() is IImmutableSet<ProjectConfiguration> knownConfigurations)
                         {
-                            if (await propertyPageCache.BindToRule(knownConfiguration, _pageName) is IRule boundRule)
+                            foreach (ProjectConfiguration knownConfiguration in knownConfigurations.Where(config => config.MatchesDimensions(_dimensions)))
                             {
-                                projectRules.Add(boundRule);
+                                if (await propertyPageCache.BindToRule(knownConfiguration, _pageName) is IRule boundRule)
```
MiYanni, line 84: Since you have this 'filter' if-statement, you can either combine it with the Where, or (to save allocations) remove the Where and put an && condition in this if-statement for the `MatchesDimensions` clause. Either will remove double filtering.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
```diff
@@ -64,21 +65,39 @@ public async Task OnBeforeExecutingBatchAsync(IEnumerable<UnconfiguredProject> t
...
-                    if (await propertyPageCache.GetKnownConfigurationsAsync() is IImmutableSet<ProjectConfiguration> knownConfigurations)
+                    if (configurationDependent)
                     {
-                        foreach (ProjectConfiguration knownConfiguration in knownConfigurations.Where(config => config.MatchesDimensions(_dimensions)))
+                        // The property is configuration-dependent; we need to collect the bound rules for
+                        // all matching configurations.
+                        if (await propertyPageCache.GetKnownConfigurationsAsync() is IImmutableSet<ProjectConfiguration> knownConfigurations)
                         {
-                            if (await propertyPageCache.BindToRule(knownConfiguration, _pageName) is IRule boundRule)
+                            foreach (ProjectConfiguration knownConfiguration in knownConfigurations.Where(config => config.MatchesDimensions(_dimensions)))
```
drewnoakes, line 82: Can potentially improve perf here by running the `BindToRule` calls concurrently, rather than serializing them in a `foreach`.

Something like:

```c#
var tasks = knownConfigurations
    .Where(config => config.MatchesDimensions(_dimensions))
    .Select(config => propertyPageCache.BindToRule(config, _pageName))
    .ToList();

await Task.WhenAll(tasks);

projectRules.AddRange(tasks.Select(task => task.GetAwaiter().GetResult());
```

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
tmeschter: The call to `BindToRule` may cause us to call `UnconfiguredProject.LoadConfiguredProjectAsync`, and that method is likely to take a lock. As such, the potential benefits here may not be as great as we could hope.

It may still be a worthwhile optimization, but I want to wait until we have actual data traces before I start with perf optimizations--especially those that introduce concurrency.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUIPropertyValueActionCore.cs
tmeschter: Done.

jjmew (approved):

PR #6778, opened 2020-12-03
Extend up-to-date check telemetry

### LINKED ISSUE #6776: Extend fast up-to-date check telemetry
Currently we only capture whether VS's fast up-to-date check succeeded or failed, and if failed then what the reason was (a simple string).

We should extend this to include:

- How much time was spent on the up-to-date check
- How many files were sniffed on disk
- Potentially other items (please comment)

Ideally we would also report on cases where the check decided the project was not up to date so scheduled a build, but MSBuild then concluded the project actually was up to date. @rainersigwald do you know of a way we could capture this information?

drewnoakes:
Fixes #6776.

Extends the existing telemetry events for pass/fail to include two new properties each:

1. The duration of the check, in milliseconds
2. The number of file system timestamp queries

## Not up-to-date

Event `vs/projectsystem/managed/uptodatecheck/fail`:

| property | value |
|----|---|
|vs.projectsystem.managed.uptodatecheck.durationmillis|22.5333|
|vs.projectsystem.managed.uptodatecheck.fail.reason|Outputs|
|vs.projectsystem.managed.uptodatecheck.filecount|6|

## Up-to-date

Event `vs/projectsystem/managed/uptodatecheck/success`:

| property | value |
|----|---|
|vs.projectsystem.managed.uptodatecheck.durationmillis|16.1586|
|vs.projectsystem.managed.uptodatecheck.filecount|157|

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6778)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Log.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Log.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using Microsoft.VisualStudio.Telemetry;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Log.cs
```diff
@@ -52,14 +57,30 @@ private static void ConvertToLocalTimes(object[] values)
...
             {
+                _stopwatch.Stop();
+
                 Minimal(message, values);
-                _telemetryService.PostProperty(TelemetryEventName.UpToDateCheckFail, TelemetryPropertyName.UpToDateCheckFailReason, reason);
+
+                _telemetryService.PostProperties(TelemetryEventName.UpToDateCheckFail, new[]
+                {
+                    (TelemetryPropertyName.UpToDateCheckFailReason, (object)reason),
+                    (TelemetryPropertyName.UpToDateCheckDurationMillis, _stopwatch.Elapsed.TotalMilliseconds),
```
panopticoncentral, line 67: >UpToDateCheckDurationMillis [](start = 43, length = 27)

It's a little bit of a shame that the project system telemetry service doesn't support telemetry events with duration (i.e. a start and an end) because then this would just be recorded naturally by the telemetry event.

panopticoncentral (approved):
:shipit:

PR #6779, opened 2020-12-03
Adding support for WebView2 debugging in .Net Core projects

rakatyal:
This PR adds an option to the Properties page for users to select WebView2 debugging. For now, we will support JavaScript debugging as stand alone and not in conjunction  with other debuggers. So when the WebView2 option is selected, we disable the Native and SQL debugging options. An additional text block is also displayed, informing the user of the needed component dependency inside VS.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6779)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=6779) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=6779) before we can accept your contribution.<br/><br/>:x: rakatyal [sign now](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=6779)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/project-system?pullRequest=6779) it.</sub>

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -399,6 +401,23 @@ private async Task<DebugLaunchSettings> GetConsoleTargetForProfile(ILaunchProfil
...
+                          new JProperty("type", "pwa-msedge"),
+                          new JProperty("runtimeExecutable", finalExecutable),
+                          new JProperty("webRoot", workingDir), // We use the Working Directory debugging option as the WebRoot, to map the urls to files on disk
+                          new JProperty("useWebView", true),
+                          new JProperty("runtimeArgs", finalArguments)
+                      );
+
+                  settings.Options = JsonConvert.SerializeObject(debuggerLaunchOptions);
+              }
+
```
drewnoakes, line 420: Please fix the indentation here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageViewModel.cs
```diff
@@ -394,6 +414,7 @@ private bool TrySetOtherProperty<T>(string propertyName, T value, T defaultValue
 
         public bool SupportNativeDebugging       => ActiveProviderSupportsProperty(UIProfilePropertyName.NativeDebugging);
         public bool SupportSqlDebugging          => ActiveProviderSupportsProperty(UIProfilePropertyName.SqlDebugging);
+        public bool SupportJSWebView2Debugging => ActiveProviderSupportsProperty(UIProfilePropertyName.JSWebView2Debugging);
```
drewnoakes, line 417: Please fix the indentation here.

```suggestion
        public bool SupportJSWebView2Debugging   => ActiveProviderSupportsProperty(UIProfilePropertyName.JSWebView2Debugging);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Converters.cs
```diff
@@ -30,4 +30,17 @@ public object[] ConvertBack(object value, Type[] targetTypes, object parameter,
...
+    internal sealed class InverseBooleanConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return !(bool)value;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException("ConvertBack should NOT be invoked");
```
drewnoakes, line 43:
```suggestion
            return !(bool)value;
```

drewnoakes (approved):
Looks good, modulo a few nits.

rakatyal:
@drewnoakes: Thanks for the review. Comments addressed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageControl.xaml
```diff
@@ -541,12 +551,33 @@
...
+                        <Binding Path="IsProfileSelected" Mode="OneWay"/>
+                        <Binding Path="JSWebView2Debugging" Mode="OneWay" Converter="{StaticResource InverseBooleanConverter}"/>
+                    </MultiBinding>
+                </CheckBox.IsEnabled>
+            </CheckBox>
+
+            <!-- JavaScript WebView2 Debugging -->
+            <CheckBox
+                Grid.Row="11"
+                Grid.Column="0"
```
tmeschter, line 568: Could you update this to have the checkbox span multiple columns so that it doesn't push the other content so far over? Please also update the screenshot in the PR.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/DebugPageControl.xaml
drewnoakes: We should apply the same change to the other checkboxes, as they may have longer strings in other languages.

tmeschter (changes requested):
FYI, per conversation with @rakatyal, the correct functioning of this feature requires that the Javascript Diagnostic component be installed, and it is currently an optional component of the desktop workload. If the user checks the box but it is not installed, when they debug they'll get an error message from the debugger. 

Before this goes in, I'd like to see some text in the UI leading the user to install the necessary component.

drewnoakes:
> Before this goes in, I'd like to see some text in the UI leading the user to install the necessary component.

Is it possible to only show the checkbox when that component is installed?

rakatyal:
@tmeschter : I added a text block as per our conversation (updated the image in the description). 
@drewnoakes: I am not aware of an easy way to do this.

tmeschter:
> > Before this goes in, I'd like to see some text in the UI leading the user to install the necessary component.
> 
> Is it possible to only show the checkbox when that component is installed?

I believe there are VS APIs for checking for installed components, however I am not familiar with them and I don't want to send @rakatyal down a rabbit hole for that.

Also, on further reflection I don't think we want to hide the check box even if we could. If the user opens this property page expecting to see it and they don't then they will just be stuck. Listing the requirement after they check the box is imperfect, but at least it provides *some* direction.

What I would like to see in an ideal world is the following:
1. The check box is always present (at least for project types where it makes sense).
2. The check box would be disabled if the required component isn't installed.
3. In addition, if (and only if) the required component isn't installed we would also include a brief message (like the one  @rakatyal just added) explaining the requirement, _with a link that would open up the installer._

However, these are changes that we can make after the fact. I think this is good enough for now.

@rakatyal Another thought: could the WebView2 documentation encourage the use of [.vsconfig](https://devblogs.microsoft.com/setup/configure-visual-studio-across-your-organization-with-vsconfig/#:~:text=%20In%20Visual%20Studio%202019%2C%20you%20can%20create,want%20to%20make%20and%20click%20Export%20More%20) files to ensure that the required component is installed? That is, if it is already going to explain that this optional component be installed in order to debug the WebView2, I think it would be great if it could also explain how to add that component to a .vsconfig file.

tmeschter (approved):

PR #6781, opened 2020-12-07
Improve performance of implicit dependency handling

### LINKED ISSUE #3548: Memory allocation in DependenciesSnapshotProvider.UpdateDependenciesSnapshot.GetProjectItemSpecs

drewnoakes:
Fixes #3548

Removes a bunch of redundant work and memory allocation during the ingestion of dependency data.

## Background

Project dependency declarations fall into two categories:

1. Declared in the primary project file (e.g. `MyProject.csproj`).
2. Declared in an imported project file (e.g. `Directory.Build.props` or MSBuild/SDK props files).

Dependencies in the second category are referred to as _implicit_. Compared to items in the first category, implicit dependencies:

- have a padlock overlay displayed on their icons, and
- may not be removed from the project, so we hide the _Remove_ item from their context menus.

We determine whether a dependency is implicit or not via two means:

1. If the dependency has `IsImplicitlyDefined` metadata, we use that value.
2. Otherwise, if the dependency has `DefiningProjectFullPath` metadata, where compare that value with the primary project file. If they don't match, it is an implicit dependency.

## Problem

Previously, the `BuildUtilities.IsImported` extension method was used to query `DefiningProjectFullPath` metadata via `ProjectItemInstance.GetMetadataValue`. This method is implemented by MSBuild, and is quite expensive in both time and memory.

We would call this method for all items in the snapshot on every single update. This included items which would never be displayed in the tree, meaning that even for projects with very few dependencies there could be hundreds upon hundreds of calls. For large solutions, this adds up.

We would also pass an `IImmutableSet<string>` containing the project's non-implicit items around to several APIs during snapshot construction, rather than storing this information directly upon the dependencies themselves.

The code that tested whether an item existed in this set of non-impliict items was in an `IDependenciesSnapshotFilter`. We are working towards removing such filters as they spread logic controlling the handling of dependency state across areas and encourage excessive data sharing to support these logic islands.

## Solution

CPS already provides `DefiningProjectFullPath` metadata for items returned via evaluation. Rather than querying this data from the project instance, we can reuse the existing data in the rule.

Achieving this was slightly complex, as design-time build items do not include this metadata. Some refactoring occurred to make the evaluated metadata available when processing design-time build data.

## Further consolidation of _implicit_ handling

`IDependencyModel` contains various flags, including `Implicit`. It also contains various icons, including expanded/collapsed, normal/warning.

For unknown reasons:

- `IDependencyModel` does not specify the icon to use for implicit dependencies
  - Instead `IProjectDependenciesSubTreeProviderInternal` exists for this reason alone
- `IDependencyModel.Implicit` property does not appear to be trusted
  - Instead `ImplicitDependenciesSnapshotFilter` would modify earlier-produced dependencies to add this flag and update the icon

This PR:

- Removes `IProjectDependenciesSubTreeProviderInternal`
- Removes `ImplicitDependenciesSnapshotFilter`
- Updates `PackageRuleHandler` and `DependenciesRuleHandlerBase` to produce a correct value for `IDependencyModel.Implicit`
- Expands `DependencyIconSet` to include `ImplicitIcons` and `ImplicitExpandedIcon`
- Simplifies several interfaces by removing parameters which supported passing data between stages of the ingestion pipeline

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6781)


jjmew:
> using System;

Thank you for great explanation. This sounds like a good candidate to discuss in one of the team meetings.

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed/Build/BuildUtilities.cs:3 in 578e22c. [](commit_id = 578e22cf73ded98f85501f74fa7c2468c3b99734, deletion_comment = False)

jjmew (approved):
:shipit:

PR #6796, opened 2020-12-16
Only use the new properties pages for CPS projects

### LINKED ISSUE #6794: Enabling new Project Properties UI breaks properties editor for legacy projects
_This issue has a corresponding [ticket on Developer Community](https://developercommunity2.visualstudio.com/t/Updated-Project-Properties-UI-not-workin/1287500). Please vote and comment there to make sure your voice is heard._

---
[severity:I'm unable to use this version] [regression] [worked-in:16.9.0 Preview 1.0]

Creating a new project and trying to open the properties does nothing. This works just fine if Updated Project Properties UI is disabled, but as long as it's enabled, I'm unable to pull up properties for any project.

drewnoakes:
Fixes #6794.

This prevents opening the properties pages in csproj projects, which fails.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6796)

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerEditorFactory.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerEditorFactory.vb
@@ -179,7 +179,7 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
         Implements IVsEditorFactory.CreateEditorInstance
 
             ' If we're using the new project properties editor, delegate to its editor factory
-            If UseNewEditor Then
+            If UseNewEditor(Hierarchy) Then
                 Return GetNewEditorFactory().CreateEditorInstance(
                     vscreateeditorflags,
                     FileName,
@@ -225,7 +225,12 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
             Return hr
         End Function
 
-        Private Function UseNewEditor As Boolean
+        Private Function UseNewEditor(vsHierarchy As IVsHierarchy) As Boolean
+            If Not vsHierarchy.IsCapabilityMatch("CPS") Then
+                ' The new editor is only available for CPS-based projects
+                Return False
+            End If
+
             Dim featureFlags = TryCast(_siteProvider.GetService(GetType(SVsFeatureFlags)), IVsFeatureFlags)
 
             Return _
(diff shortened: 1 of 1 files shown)


MiYanni (approved):

swesonga (approved):

tmeschter (approved):

HTD:
Where can I find the feature flag? It's not available in settings (it disappeared and I'm stuck with the broken editor). I tried to search for this in VS configuration files but failed to find it. I use Preview 2 version. I need to fallback to VS 2019 to edit some properties (or edit the project file manually).

drewnoakes:
@HTD the feature flag has been removed, so if there's a shortcoming of the new experience it'd be good to hear about it so we can try and fix it before GA. What properties are broken for you? If you could file a new issue outlining any problems you face, it'd be great.

HTD:
I posted the bug as the Visual Studio 2022 feedback. The Project Property Editor doesn't work. It doesn't load any projects. In Visual Studio 2019 it works OK. For now I edit the `.csproj` files manually. BTW, is that correct place to report issues with Visual Studio? If not, I would be delighted to be directed to the correct discussion. I tested the new Property Editor in Preview 1 and it was quite nice, but still didn't have `Nullable` setting. If someone is doing new editor, the `Nullable` setting seems like a very good feature to add.

drewnoakes:
@HTD would you mind linking that feedback ticket here so I can take a look?

The VS feedback tool is the best general place to report issues about Visual Studio. In the case of the Project Properties Editor, the team that owns this repo also owns that new feature, and we track work for that here. So if you hit issues or have suggestions in that particular feature, feel free to open issues here. Feedback tickets generally take a little longer to reach us due to the triage process.

As for the `Nullable` property, it is definitely supported. It is defined [here](https://github.com/dotnet/project-system/blob/2c8c5a78e06a3ba0835f018637e3602db867aa24/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml#L56-L77). Note that it is only visible if you are using C# 8 or greater (just like the nullable feature itself). That property was definitely present in VS2022 preview 1 though.

HTD:
I know the `Nullable` and `Language version` are supported, but they are not visible in old Project Property Editor. I always set them in `.csproj` files manually. BTW, I need the UI mostly to check if all needed fields for `NuGet` packages are filled. When editing manually I often either forget to set an option, or make a typo in a file path (to documentation xml or readme). I won't forget to set `Nullable` because I get warnings about it each time it's used in code. There are generally 2 kinds of project properties - the metadata used by `NuGet` and added to assembly properties, and the project setup (target frameworks, language version, nullable, compiler options, dependencies). So it could be 2 different UI editor pages as well. If the project properties are not set correctly, there are either warnings or errors, if the metadata is not set I see it either on publishing, or when it's already published.

PR #6805, opened 2020-12-21
Fix duplicate SDK/package nodes on older .NET Core targets

### LINKED ISSUE #6802: For older .NET Core targets, dependencies tree shows both package and SDK nodes for Microsoft.NETCore.App
Version 16.9.0 Preview 3.0 [30820.7.main].

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>netcoreapp1.0;netcoreapp1.1;netcoreapp2.0;netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
  </PropertyGroup>

</Project>
```

For .NET Core App versions 1.0, 1.1, 2.0, 2.1 and 2.2, we see both a package and SDK entry for `Microsoft.NETCore.App`:

image

We should only show a single node for these targets, as we do for other targets (e.g. .NET Core 3.0+).

This does not occur in 16.8 and is likely a result of the change in #6781.

For those targets, the target framework pack was exposed by the SDK as a package reference, and the project system has special logic (`SdkAndPackagesDependenciesSnapshotFilter`) to pretend that these are actually SDK nodes. In newer targets these are treated as target framework references, not package references. Nowadays, SDK references are for "extension SDKs" which are very rare.

Note that the package node actually expands to show more information than the SDK node, so it may be preferable to keep the package item.

The special handling in `SdkAndPackagesDependenciesSnapshotFilter` is another example of the anti-pattern of 'filters' in the dependencies node code, that makes certain types of data handling complicated. It may be that we can:

- Avoid showing duplicate nodes when targeting these older .NET Core versions
- Show more information about the target packs via their package entries rather than their SDK entries
- Remove one of the three remaining 'filter' classes, getting us closer to removing filters altogether

drewnoakes:
Fixes #6802

Between .NET Core versions 1.0 and 2.2, the target framework pack was provided to the dependencies tree as a "real" package reference item, and a corresponding "fake" SDK reference. The dependencies tree then had special knowledge about these items and their relationship, and would juggle them to apply information from the package reference onto the fake SDK node.

This juggling occurred in `SdkAndPackagesDependenciesSnapshotFilter`, an export of `IDependenciesSnapshotFilter`. We are working to remove these filter implementations as they spread data processing logic across the dependenices tree's data pipeline, creating implicit relationships between seemingly unrelated pieces of code.

In general, the dependencies tree (and indeed then entire project system) should avoid hard-coding knowledge of specific details of the SDK's implementation where possible.

This change removes `SdkAndPackagesDependenciesSnapshotFilter` and code that was only used by it. This has a few consequences:

- We no longer show duplicate `Microsoft.NETCore.App` nodes, removing the SDK and keeping the package.
- The package provides more information, as it may be expanded to show contents and transitive dependencies.
- We remove one of the three remaining `IDependenciesSnapshotFilter` classes.

Note that when targeting .NETCoreApp version 3.0 and later, these targeting packs are provided by the SDK as `FrameworkReference` items, not `SdkReference` nor `PackageReference` items. These behave as expected. This fix is only for projects targeting older versions of .NET Core.

## Before

image

## After

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6805)


MiYanni (approved):
Changes like this make me 😊

swesonga (approved):

PR #6806, opened 2020-12-21
Source all unresolved references from evaluation

### LINKED ISSUE #6803: SdkReference should be sourced from evaluation, not design-time builds
Version 16.9.0 Preview 3.0 [30820.7.main].

The `SdkReference.xaml` rule file currently shows as its data source:

https://github.com/dotnet/project-system/blob/9ff0ae4290bca4257c77f0e77f28d04d965d954f/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/SdkReference.xaml#L9-L14

On line 11, `MSBuildTarget="CollectSDKReferencesDesignTime"` means that this rule's data is sourced from a design-time build target, rather than from evaluation. Design-time builds are much slower to complete than evaluation, so this configuration causes population of the dependency tree to be delayed until the first design-time build completes.

We must source unresolved dependency items from evaluation.

### LINKED ISSUE #6804: AnalyzerReference should be sourced from evaluation, not design-time builds
Version 16.9.0 Preview 3.0 [30820.7.main].

The `AnalyzerReference.xaml` rule file currently shows as its data source:

https://github.com/dotnet/project-system/blob/9ff0ae4290bca4257c77f0e77f28d04d965d954f/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AnalyzerReference.xaml#L9-L14

On line 11, `MSBuildTarget="CollectAnalyzersDesignTime"` means that this rule's data is sourced from a design-time build target, rather than from evaluation. Design-time builds are much slower to complete than evaluation, so this configuration causes population of the dependency tree to be delayed until the first design-time build completes.

We must source unresolved dependency items from evaluation.

drewnoakes:
Fixes #6803. Fixes #6804.

The dependencies tree consumes both unresolved and resolved versions of project dependency items.

Unresolved items are obtained quickly via evaluation in order to build the tree rapidly when a project is opened, and to quickly reflect changes to the tree as items are added, removed and updated over time.

Resolved items are obtained via design-time builds. They contain much richer information about dependencies than evaluation items, and the tree is updated with that information as it becomes available.

It's important that unresolved items are only sourced from evaluation. If their rule files request they come from design-time builds, then they the dependencies tree cannot be built until a design-time build completes.

Before this change, both SDK and Analyzer items were being sourced from design-time build targets. This meant that all other dependency types were being held up, waiting for the design-time build to complete. To the user, this means a delay in the construction of the dependencies tree when loading a project.

This change adds unit tests to validate the details of both resolved and unresolved dependency item data sources, to prevent this problem occurring again in future.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6806)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedFrameworkReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedFrameworkReference.xaml
@@ -5,7 +5,6 @@
       DisplayName="Framework Reference"
       PageTemplate="generic"
       xmlns="http://schemas.microsoft.com/build/2009/properties">
-  
   <Rule.DataSource>
     <DataSource HasConfigurationCondition="False"
                 ItemType="FrameworkReference"
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AnalyzerReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AnalyzerReference.xaml
@@ -8,10 +8,8 @@
   <Rule.DataSource>
     <DataSource HasConfigurationCondition="False"
                 ItemType="Analyzer"
-                MSBuildTarget="CollectAnalyzersDesignTime"
                 Persistence="ProjectFile"
-                SourceOfDefaultValue="AfterContext"
-                SourceType="TargetResults" />
+                SourceOfDefaultValue="AfterContext" />
   </Rule.DataSource>
   
   <StringProperty Name="BrowsePath"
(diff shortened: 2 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/ResolvedFrameworkReference.xaml
```diff
@@ -5,7 +5,6 @@
       DisplayName="Framework Reference"
       PageTemplate="generic"
       xmlns="http://schemas.microsoft.com/build/2009/properties">
-
```
drewnoakes, line 8: This blank line removal is on purpose. It makes the formatting of this file match all others in this folder.

MiYanni (approved):

swesonga (approved):

dmonroym (approved):

PR #6822, opened 2021-01-11
BuildUpToDateCheck subscribes for updates lazily

### LINKED ISSUE #6327: Delay initializing BuildUpToDateCheck and PackageRestoreProgressTracker for inactive project configurations
Current, inactive project configurations can be loaded in various conditions, either because the mismatching solution configurations, or when an extension or the developer access configured level properties (like appDesigner). AutoLoadComponents, especially dynamic loaded components are triggered in those configurations, which initializes dataflows, especially DT build dataflow, which is not essential until the configuration is activated. If we can delay initialize some of the auto load components until the configuration is activated/implicitly activated, it will reduce some of those initialization/overhead.  BuildUpToDateCheck and PackageRestoreProgressTracker are two of those components inside ML project system, which can be delayed to establish dataflow subscription, until the configuration is activated.

drewnoakes:
Fixes #6327.

The `BuildUpToDateCheck` component is constructed and initialised (via `IProjectDynamicLoadComponent`) for both active and inactive project configurations. Previously, this initialisation would trigger dataflow subscriptions for each project configuration, triggering CPS to construct a chain of dataflow blocks to service that subscription. For inactive configurations those chains linger in memory without actually being used.

This commit avoids this problem by deferring construction of the dataflow subscription until the first call to `IsUpToDateAsync`. This method is only called when the configuration is actually being used. In this way, the overhead associated with instances of `BuildUpToDateCheck` in inactive project configurations is minimised.

An `IProjectDynamicLoadComponent` may be loaded and unloaded multiple times, creating multiple lifetimes for the component. This change introduces a `Subscription` type into which all state associated with such a lifetime is moved: the state object and dataflow link. This makes the lifetime explicit and more clearly indicates lifetime events in the code.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6822)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
@@ -1,0 +1,182 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
+{
+    internal sealed partial class BuildUpToDateCheck
+    {
+        /// <summary>
+        /// Contains and tracks state related to a lifetime instance of <see cref="BuildUpToDateCheck"/>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// As the parent <see cref="BuildUpToDateCheck"/> is an <see cref="IProjectDynamicLoadComponent"/>, it may have multiple lifetimes.
+        /// This class contains all the state associated with such a lifetime: it's Dataflow subscription, tracking the first value to arrive,
+        /// and the <see cref="BuildUpToDateCheck.State"/> instance.
+        /// </para>
+        /// <para>
+        /// Initialization of the Dataflow subscription happens lazily, upon the first up-to-date check request.
+        /// </para>
+        /// <para>
+        /// Destruction of the Dataflow subscription happens when the parent component is disposed or unloaded.
+        /// </para>
+        /// </remarks>
+        private sealed class Subscription : IDisposable
+        {
+            /// <summary>
+            /// Completes when the first project update is received. Cancelled if the subscription is disposed.
+            /// </summary>
+            /// <remarks>
+            /// This field is also used to synchronise updates to <see cref="_link"/> and <see cref="State"/>.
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
```diff
@@ -0,0 +1,182 @@
...
+                // prevent redundant work/allocation for inactive project configurations.
+                // https://github.com/dotnet/project-system/issues/6327
+                //
+                // We don't pass the cancellation token here as initialization must be atomic.
+                EnsureInitialized();
+
+                token.ThrowIfCancellationRequested();
+
+                // Wait for the first state to be computed
+                await _dataReceived.Task.WithCancellation(token);
```
MiYanni, line 85: Do we not do `ConfigureAwait(false)` on all async calls in this repo? I thought any library used for application purposes (those that have UI threads) should always have `ConfigureAwait(false)`.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
drewnoakes: That is common advice, but we do not do so in this repo. Nor does CPS.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -75,70 +70,31 @@ internal sealed partial class BuildUpToDateCheck : OnceInitializedOnceDisposedUn
...
+        public void Dispose()
         {
-            _link?.Dispose();
+            if (Interlocked.CompareExchange(ref _isDisposed, 1, 0) != 0)
+            {
+                return;
+            }
 
-            return Task.CompletedTask;
+            RecycleSubscription();
```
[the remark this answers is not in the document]
drewnoakes: The intent was that the would always be a subscription object in that field, ready to go. This allows the swap to be lock-free (`Interlocked.Exchange`). The subscription itself is lazy, so the cost of having an unused one is minimal.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
davkean: ConfigureAwait(false) says "schedule the continuation on a thread-pool thread". This is mostly not appropriate in VS and should only be done deliberately.

PR #6840, opened 2021-01-18
Ban IVsFrameworkParser

drewnoakes:
As described in #6369 and #6370, `IVsFrameworkParser` uses the `FrameworkName` type which does not support the platform information added to target frameworks in .NET 5.

We have already done the work to remove usages of this interface. This change prevents us from attempting to use the type again in future.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6840)


jjmew (approved):
:shipit:

PR #6844, opened 2021-01-20
Adding command checks for WebView2 debugging

rakatyal:
We should only change the debugger settings when the command is set to 'Project' or 'Executable' since WebView2 debugging is only supported for those.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6844)


rakatyal:
@tmeschter: Kindly review.

drewnoakes (approved):

swesonga (approved):

tmeschter (approved):

PR #6847, opened 2021-01-20
Change the default value for Symbols Rename in a fresh VS installation.

ocallesp:
fixes [AB#1237959](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1237959)

In a fresh VS installation or repair, Dotnet ProjectSystem cannot get the value in the checkbox Tools | Options | Project and Solutions | Enable Symbol Rename. At this point VS has no knowledge of this value and that is how the system works. The user has to turn the checkbox off and then on to make it work and have a persisted value for future usages.

This small change sets the default value of EnableSymbolRename to true to match the value on a fresh VS Installation.

This change will enable Symbols Rename by default on a fresh installation.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
@@ -230,7 +230,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
         {
             ISettingsManager settings = await _settingsManagerService.GetValueAsync();
 
-            bool enableSymbolicRename = settings.GetValueOrDefault("SolutionNavigator.EnableSymbolicRename", false);
+            bool enableSymbolicRename = settings.GetValueOrDefault("SolutionNavigator.EnableSymbolicRename", true);
 
             await _projectVsServices.ThreadingService.SwitchToUIThread();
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
```diff
@@ -230,7 +230,7 @@ private async Task<bool> CheckUserConfirmationAsync(string oldFileName)
         {
             ISettingsManager settings = await _settingsManagerService.GetValueAsync();
 
-            bool enableSymbolicRename = settings.GetValueOrDefault("SolutionNavigator.EnableSymbolicRename", false);
+            bool enableSymbolicRename = settings.GetValueOrDefault("SolutionNavigator.EnableSymbolicRename", true);
```
swesonga, line 233: Consider adding a comment indicating that the default value needs to match the default value in the Tools > Options UI

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
ocallesp: Yeah, makes sense. I'll add a comment in the next commit.

dmonroym (approved):

PR #6852, opened 2021-01-21
Improve activation of the fast up-to-date check component within the active project configuration

### LINKED ISSUE #6850: Up-to-date check laziness is causing delays during first build
#6822 made the up-to-date check's initialization lazy, such that it deferred its dataflow subscriptions until its first invocation. The motivation for that change was to prevent dataflow subscriptions for inactive configurations. Exports for inactive configurations are unused, so initializing them is wasteful and negatively impacts solution load and retained memory usage.

That change had the negative consequence of deferring dataflow subscriptions until the first build, at which time a design-time build would be scheduled to obtain the needed information. This triggered several RPS regressions.

A better solution is to monitor the active project configuration and only activate the up-to-date check component associated with the active configuration.

drewnoakes:
Fixes #6850 and the RPS regression associated with it.

Best reviewed commit-by-commit.

---

### Rename `IImplicitlyActiveService` to `IImplicitlyActiveConfigurationComponent`

This paves the way for `IActiveConfigurationComponent` in the next commit, which makes more sense as a name than `IActiveService`. I use the suffix _component_ rather than _service_ to mirror the suffix of `IDynamicProjectLoadComponent` which facilitates a similar pattern, but for project capabilities alone.

---

### Add `IActiveConfigurationComponent`

Similar to `IImplicitlyActiveConfigurationComponent`, this type is intended for configured-project scope exports to allow them to be activated and deactivated in response to their configuration becoming active or inactive.

The existing `IImplicitlyActiveConfigurationComponent` does this for _implicitly_ active configurations, of which there may be many (when multi-targeting, for example). This alternative exists for the _singular_ active configuration, which is usually the first when multi-targeting.

This pattern can be useful for deferring initialisation of components until their configuration is activated, reducing work on solution load for example. It can also help clean up components which are no longer applicable when the configuration is changed.

---

### Activate and deactivate the up-to-date check

#6822 made the up-to-date check's initialization lazy, such that it deferred its dataflow subscriptions until its first call. The motivation for that change was to prevent dataflow subscriptions for inactive configurations. Exports for inactive configurations are unused, so initializing them is wasteful and negatively impacts solution load and retained memory usage.

This change had the negative consequence of deferring dataflow subscriptions until the first build, at which time a design-time build would be scheduled to obtain the needed information.

A better solution is to monitor the active project configuration and only activate the up-to-date check component associated with the active configuration. That is what this commit does.

It does so by replacing `IProjectDynamicLoadComponent` with `IActiveConfigurationComponent`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6852)


tmeschter (approved):

drewnoakes:
@davkean I'm going to merge this now to get RPS started and unblock our insertions. If you have any feedback I will file a follow up PR.

PR #6861, opened 2021-01-25
Add telemetry to track when properties pages are already opened.

melytc:
The goal is to be able to measure the success of opening the properties pages. For that, we already track when they are closed and asked to be opened; but we missed an event to track when they are already opened (focused or not focused).

The general flow is as follows:
IMG_8081

The diagram can be read like this:
- The user first right clicks the project. This is the context menu.
- The user then selects the "Properties" command.
- The ProjectFileDocumentManager calls the method `OpenWithSpecific`, which calls to the ApplicationDesignerWindowPane's `ActivateLogicalWindow`, `PopulateView`, which calls to create the AppDesignerView with the `InitView` method. This is where we get our first event out: we call `LogAppDesignerDefaultPageOpened` which sends vs/projectsystem/appdesigner/page-opened.
- We move on to adding the pages o tabs to the AppDesigner in `AddTabs`, where for each tab we send a vs/projectsystem/appdesigner/tabinfo event. This includes the tab title, tab guid, project extension and project guid.
- Additionally, we send information about "specialtabs", which they refer to the resource and settings page. This is captured in vs/projectsystem/appdesigner/tabinfo/specialtabs.
- Finally, we get to `OnInitializationComplete` where we call `ShowTab` to either show the default page if this is the first time opening the properties pages (i.e. after creating a project), or to show the last active page. This is reflected in the last event sent, vs/projectsystem/appdesigner/page-opened, by calling `LogAppDesignerPageOpened`.

As mentioned at the beginning, this flow only covers the scenarios of having the properties pages closed. When having the properties pages opened, the flow is shorter: 
- The ProjectFileDocumentManager calls `OpenWithSpecific`, which again calls the ApplicationDesignerWindowPane's `ActivateLogicalWindow`; but since the AppDesignerView already exists, jumps straight to showing it.

With these changes, after realizing that the AppDesignerView already exists, we call a new method `AppDesignerAlreadyLoaded` which calls the `LogAppDesignerPageOpened` with a flag to identify that the properties pages are opened. This would look something like this:

IMG_8082

Note that the event is now called three times and we can tell them apart from the properties and values being sent:
1. At the `InitView` when we create our properties pages. `page-opened` = 0.
2. At the `ShowTabs` when we actually display a page from it. `page-opened` = 2, a `pageguid` and the `tabtitle` = Application.
3. When we detect the page is already opened. `page-opened` = 2, a `pageguid`, the `tabtitle` = Application and the flag `alreadyopened` = True.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6861)

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
@@ -970,6 +970,11 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
             Return True
         End Function
 
+        Public Sub AppDesignerAlreadyLoaded()
+            Dim ActivePanel As ApplicationDesignerPanel = _designerPanels(_activePanelIndex)
+            Common.TelemetryLogger.LogAppDesignerPageOpened(ActivePanel.ActualGuid, ActivePanel.TabTitle, True)
+        End Sub
+
         ''' <summary>
         ''' Show the requested tab
         ''' </summary>
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
@@ -970,7 +970,7 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
             Return True
         End Function
 
-        Public Sub AppDesignerAlreadyLoaded()
+        Friend Sub AppDesignerAlreadyLoaded()
             Dim ActivePanel As ApplicationDesignerPanel = _designerPanels(_activePanelIndex)
             Common.TelemetryLogger.LogAppDesignerPageOpened(ActivePanel.ActualGuid, ActivePanel.TabTitle, True)
         End Sub
--- a/src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
@@ -643,12 +643,12 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
                 LogAppDesignerPageOpened(DEFAULT_PAGE)
             End Sub
 
-            Public Shared Sub LogAppDesignerPageOpened(pageGuid As Guid, Optional tabTitle As String = Nothing)
+            Public Shared Sub LogAppDesignerPageOpened(pageGuid As Guid, Optional tabTitle As String = Nothing, Optional alreadyOpened As Boolean = False)
                 Dim pageId = PageGuidToId(pageGuid)
-                LogAppDesignerPageOpened(pageId, pageGuid, tabTitle)
+                LogAppDesignerPageOpened(pageId, pageGuid, tabTitle, alreadyOpened)
             End Sub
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -660,6 +660,10 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
                     userTask.Properties("vs.projectsystem.appdesigner.page-opened.tabtitle") = tabTitle
                 End If
 
+                If alreadyOpened Then
+                    userTask.Properties("vs.projectsystem.appdesigner.page-opened.alreadyopened") = alreadyOpened
+                End If
```
drewnoakes, line 665: Would it be useful to add this property unconditionally? As it stands, the property is omitted when `alreadyOpened` is false, and will only be present with a true value. Maybe allow the value to be both true and false. I guess this comes down to how the telemetry data is processed, so may be a question for @jjmew.

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
```diff
@@ -970,6 +970,11 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
             Return True
         End Function
 
+        Public Sub AppDesignerAlreadyLoaded()
```
drewnoakes, line 973: Could this be [`Friend`](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/modifiers/friend) rather than `Public`?

drewnoakes (approved):
Looks good. Love the notes. Great idea to share them.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
jjmew: If I am reading this correctly we will get the Page-Opened event every time that the user attempts to launch it. This will be even if the properties pages is already opened. I think it fine like it is right now. 

---
In reply to: [563556072](https://github.com/dotnet/project-system/pull/6861#discussion_r563556072) [](ancestors = 563556072)

jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
tmeschter: It looks to me like it could be `Private`, unless there is some reason a method outside this class would need to call it?

tmeschter (changes requested):

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
melytc: The only place where we pass it as `true` is on the `AppDesignerAlreadyLoaded`, and it's `false` by default. I think Drew's comment makes sense and could better describe the event, letting us know explicitly when they were closed. Doesn't harm to have it 😉

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerView.vb
melytc: It's being called by the `ApplicationDesignerWindowPane`, so I think it makes sense changing it to `Friend`.

MiYanni (approved):

tmeschter (approved):

PR #6443, opened 2020-07-28
Fix startup show only debuggable projects

ocallesp:
This pr fixes #1308

The Startup project was showing NetCore Libraries which overwhelm users when using VS with large projects. This bug was not about whether Libraries should be in Startup or not, but to decide which projects are non-debuggable and only display debuggables.

Added the project capability `DisableBuiltInDebuggerServices `to disable CPS debuggers services for managed projects.

Implemented the missing functionality in `CanLaunch`() to decide which projects should appear in the Startup list.
The logic to decide if a project is non-debuggable is in `LaunchProfilesDebugLaunchProvider`

Fixed unit tests

By default, classes that do not implement `IDebugProfileLaunchTargetsProvider3 `will appear in the Startup drop down list.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6443)

PR #6871, opened 2021-01-26
Remove ProjectNotRunnableDirectlyException

MiYanni:
Fixes: [AB#1245470](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1245470)
Related: https://github.com/dotnet/project-system/pull/6443#pullrequestreview-458792319

Currently, `ProjectNotRunnableDirectlyException` is thrown when the project cannot be debugged, such as a class library. Then, this exception is explicitly caught in `CanBeStartupProjectAsync`. If caught, it cannot be a startup projects (returns false). Otherwise, the method returns true.

This exception is thrown in `GetTargetCommandAsync`. Working up the call stack, this is used in the 2 public methods, `QueryDebugTargetsForDebugLaunchAsync` and `QueryDebugTargetsAsync`. These implement a couple of provider interfaces for the `ProjectLaunchTargetsProvider`. So, anything that calls these methods has the potential of throwing a `ProjectNotRunnableDirectlyException`.

In the attached issue, the exception the user saw was correctly being thrown from `CanBeStartupProjectAsync` according to the provided call stack. However, it is not a good practice to use exceptions as the 'happy path' for logic, even if they are being caught and handled. Instead, I've used a sentinel value `null` to represent the 'bad path' for this call stack. It isn't the best design, but it is the simplest solution without redesign of the calling structure. In the PR, the changes mainly relate to passing `null` up the call stack until it reaches `CanBeStartupProjectAsync`. If `null` is given to that method, it returns false. Otherwise, it returns true. If `null` reaches the 2 public methods, they will return an empty array to adhere to the interface the methods implement.

Other changes in this PR are addressing @drewnoakes comments on the [PR that added this exception](https://github.com/dotnet/project-system/pull/6443#pullrequestreview-458792319). Plus, minor cleanup in files that I touched, such as tab alignment and extra whitespace being removed.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
@@ -23,6 +23,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
     internal class LaunchProfilesDebugLaunchProvider : DebugLaunchProviderBase, IDeployedProjectItemMappingProvider, IStartupProjectProvider
     {
         private readonly IVsService<IVsDebuggerLaunchAsync> _vsDebuggerService;
+        // Launch providers to enforce requirements for debuggable projects
         private readonly ILaunchSettingsProvider _launchSettingsProvider;
         private IDebugProfileLaunchTargetsProvider? _lastLaunchProvider;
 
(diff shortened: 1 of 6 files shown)


davkean:
Please test these two scenarios:

## Class Library
1. File -> New -> Class Library (.NET COre)
2. F5

Expected: Get this _exact_ error:

---------------------------
Microsoft Visual Studio
---------------------------
A project with an Output Type of Class Library cannot be started directly.


In order to debug this project, add an executable project to this solution which references the library project. Set the executable project as the startup project.
---------------------------
OK   
---------------------------


## Unit Test inherits launchsettings.json

1. File -> New -> Unit Test Project
2. Properties -> Debug -> Check "Enable native code debugging"
3. Save
4. Set a breakpoint in test so that it will stop
5. Right-click on same test and choose Debug

Expected: Native debugging is enabled (have a look at CallStack to see if you see native dlls on it it)

tmeschter (approved):
Approved, pending review of the test cases suggested by @davkean.

swesonga (approved):

MiYanni:
@davkean For Scenario 1:
Updated the 2 public methods to `throw new Exception(VSResources.ProjectNotRunnableDirectly);` when it is null. Now, it properly displays the dialog with the message:
image

For Scenario 2:
When debugging, I had to "Show External Code" in the Call Stack window. After that, I could see these 2 unmanaged DLLs.
image

davkean:
👍

PR #6873, opened 2021-01-27
Replace invalid use of the TargetFrameworkMoniker with target framework alias

swesonga:
This PR reverts #6859 in order to restore #6712 with an additional fix.

The DependenciesSnapshotProvider keeps track of an AggregateCrossTargetProjectContextProvider (in its ContextTracker) to enable it to detect changes to a project's target framework. More specifically, it calls the context.TryUpdateCurrentAggregateProjectContextAsync expecting a non-null AggregateCrossTargetProjectContext if and only if the existing AggregateCrossTargetProjectContext is for a target framework that does not match the one in its current project context. This is the case:

1. Before any AggregateCrossTargetProjectContext has been created.
2. When the TargetFramework in the project file is changed.

When a new AggregateCrossTargetProjectContext needs to be created for the first time,  AggregateCrossTargetProjectContextProvider.CreateProjectContextAsync uses the ConfigurationGeneral.TargetFramework property to create a TargetFramework instance (saving this string as the target framework alias). In order to detect changes to project target frameworks, these TargetFramework instances are compared in the TryUpdateCurrentAggregateProjectContextAsync method. Therefore, the logic that picks up the target framework from the new/incoming project properties needs to also read this exact same ConfigurationGeneral.TargetFramework property. The change in PR #6712 missed this, causing a mismatch between the logic that creates new TargetFramework instances and the logic that detects changes to the project target framework. This mismatch resulted in the dependency node falsely reacting to target framework changes (and consequently repeatedly creating new AggregateCrossTargetProjectContexts then adding subscriptions to configured project evaluation) resulting in a significant increase in allocations.

This change fixes the logic to correctly detect the target framework and also renames the TargetFramework.TargetFrameworkMoniker property to TargetFramework.TargetFrameworkAlias to reduce some confusion about exactly what it contains.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6873)

### DIFF
--- a/src/Common/BannedSymbols.txt
+++ b/src/Common/BannedSymbols.txt
@@ -13,6 +13,8 @@ T:Microsoft.VisualStudio.ProjectSystem.IProjectLockService; Using IProjectAccess
 T:Microsoft.VisualStudio.ProjectSystem.VS.IProjectGuidService; Using ISafeProjectGuidService avoids reading the GUID before it is safe to do so during initialisation
 T:Microsoft.VisualStudio.ProjectSystem.VS.IProjectGuidService2; Using ISafeProjectGuidService avoids reading the GUID before it is safe to do so during initialisation
 
+T:NuGet.VisualStudio.IVsFrameworkParser; FrameworkName does not support the platform information added in .NET 5
+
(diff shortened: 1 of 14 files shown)


dmonroym:
For tracking purposes this also restores #6840 which bans `IVSFrameworkParser`

dmonroym (approved):

drewnoakes (approved):

PR #6876, opened 2021-01-27
Update null annotations

tmeschter:
This is a follow up to #6872.

Currently, the `ILaunchSettingsProvider` interface claims that the `CurrentSnapshot` property and `WaitForFirstSnapshot` method will always return a non-null `ILaunchSettings`. Unfortunately, neither is true.

In practice, creating an `ILaunchSettings` snapshot requires knowing the active profile name, and this is stored in a property in the .user file (if it exists). Retrieving the property requires an evaluation; as such we can't produce an `ILaunchSettings` until at least one evaluation has occurred and we have a chance to process it.

It is possible for a consumer of `ILaunchSettingsProvider` to request the `CurrentSnapshot` before this has occurred, in which case they will receive `null`. Note that the implementation of this property (in the `LaunchSettingsProvider` class) calls `EnsureInitalized()`; this may give the appearance of guaranteeing that the snapshot is available, but all it does is ensure that we are signed up to receive, eventually, the data we need to produce that snapshot.

`WaitForFirstSnapshot` _could_ guarantee that a non-null `ILaunchSettings`--except that the caller may pass a time out, and if the time out expires before a snapshot is available, the call will return `null`.

This change updates the interface to reflect reality, and updates all callers in our code to respond accordingly. In most cases we were already handling the possibility of `null`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6876)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
@@ -730,7 +730,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
                 return CurrentSnapshot;
             }
 
-            await _firstSnapshotCompletionSource.Task.TryWaitForCompleteOrTimeout(timeout);
+            if (await _firstSnapshotCompletionSource.Task.TryWaitForCompleteOrTimeout(timeout))
+            {
+                Assumes.NotNull(CurrentSnapshot);
+            }
 
             return CurrentSnapshot;
         }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
@@ -723,7 +723,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
         /// This function blocks until a snapshot is available. It will return null if the timeout occurs
         /// prior to the snapshot is available
         /// </summary>
-        public async Task<ILaunchSettings> WaitForFirstSnapshot(int timeout)
+        public async Task<ILaunchSettings?> WaitForFirstSnapshot(int timeout)
         {
             if (CurrentSnapshot != null)
             {
@@ -732,7 +732,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
 
             await _firstSnapshotCompletionSource.Task.TryWaitForCompleteOrTimeout(timeout);
 
-            Assumes.NotNull(CurrentSnapshot);
             return CurrentSnapshot;
         }
 
@@ -865,7 +864,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
         /// </summary>
         public async Task<ILaunchSettings> GetSnapshotThrowIfErrors()
         {
-            ILaunchSettings currentSettings = await WaitForFirstSnapshot(WaitForFirstSnapshotDelayMillis);
+            ILaunchSettings? currentSettings = await WaitForFirstSnapshot(WaitForFirstSnapshotDelayMillis);
             if (currentSettings == null || (currentSettings.Profiles.Count == 1 && string.Equals(currentSettings.Profiles[0].CommandName, ErrorProfileCommandName, StringComparisons.LaunchProfileCommandNames)))
             {
                 string fileName = await GetLaunchSettingsFilePathAsync();
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileNameValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileNameValueProvider.cs
@@ -50,9 +50,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         /// </remarks>
         private async Task<string> GetPropertyValueAsync()
         {
-            ILaunchSettings launchSettings = await _launchSettings.WaitForFirstSnapshot(Timeout.Infinite);
+            ILaunchSettings? launchSettings = await _launchSettings.WaitForFirstSnapshot(Timeout.Infinite);
 
-            return launchSettings.ActiveProfile?.Name ?? string.Empty;
+            return launchSettings!.ActiveProfile?.Name ?? string.Empty;
         }
 
         public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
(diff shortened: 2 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -739,7 +739,6 @@ public async Task<ILaunchSettings> WaitForFirstSnapshot(int timeout)
 
             await _firstSnapshotCompletionSource.Task.TryWaitForCompleteOrTimeout(timeout);
 
-            Assumes.NotNull(CurrentSnapshot);
```
swesonga, line 742: Would it make sense to enforce this if TryWaitForCompleteOrTimeout returns true? The annotation in various callers seem to have this expectation.

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: That's reasonable/

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileNameValueProvider.cs
```diff
@@ -50,9 +50,9 @@ public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyN
         /// </remarks>
         private async Task<string> GetPropertyValueAsync()
         {
-            ILaunchSettings launchSettings = await _launchSettings.WaitForFirstSnapshot(Timeout.Infinite);
+            ILaunchSettings? launchSettings = await _launchSettings.WaitForFirstSnapshot(Timeout.Infinite);
 
-            return launchSettings.ActiveProfile?.Name ?? string.Empty;
+            return launchSettings!.ActiveProfile?.Name ?? string.Empty;
```
drewnoakes, line 55: Perhaps add a comment as to why this will not be null. Rather than suppress, consider validating. I really dislike using `!` as it carries no context and looks like design flaw. eg:

```suggestion
            // Infinite timeout means this will not be null
            Assumes.NotNull(launchSettings);
            
            return launchSettings.ActiveProfile?.Name ?? string.Empty;
```

Alternatively, create an overload (or extension method) without a timeout that returns a non-null instance. That way we don't have to replicate this comment in loads of places through the code.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/LaunchSettingsValueProviderBase.cs
```diff
@@ -45,9 +45,9 @@ public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyN
 
         public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
         {
-            ILaunchSettings launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
+            ILaunchSettings? launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
 
-            var writableLaunchSettings = launchSettings.ToWritableLaunchSettings();
+            IWritableLaunchSettings writableLaunchSettings = launchSettings!.ToWritableLaunchSettings();
```
drewnoakes, line 50: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/LaunchSettingsValueProviderBase.cs
```diff
@@ -61,9 +61,9 @@ public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyN
 
         private async Task<string> GetPropertyValueAsync(string propertyName)
         {
-            ILaunchSettings launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
+            ILaunchSettings? launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
 
-            return GetPropertyValue(propertyName, launchSettings) ?? string.Empty;
+            return GetPropertyValue(propertyName, launchSettings!) ?? string.Empty;
```
drewnoakes, line 66: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/LaunchTargetPropertyPageValueProvider.cs
```diff
@@ -60,10 +61,10 @@ public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyN
             {
                 _projectThreadingService.RunAndForget(async () =>
                 {
-                    ILaunchSettings launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
+                    ILaunchSettings? launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
 
-                    var writableLaunchSettings = launchSettings.ToWritableLaunchSettings();
-                    var activeProfile = writableLaunchSettings.ActiveProfile;
+                    IWritableLaunchSettings writableLaunchSettings = launchSettings!.ToWritableLaunchSettings();
```
drewnoakes, line 66: Ditto.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/LaunchTargetPropertyPageValueProvider.cs
```diff
@@ -80,25 +81,25 @@ public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyN
 
         private async Task<string> GetPropertyValueAsync()
         {
-            ILaunchSettings launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
-            var commandName = launchSettings.ActiveProfile?.CommandName;
+            ILaunchSettings? launchSettings = await _launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
+            string? commandName = launchSettings!.ActiveProfile?.CommandName;
```
drewnoakes, line 85: Ditto.

drewnoakes (commented):
I expect there are also a bunch of new warnings to address in `LaunchSettingsProviderTests` and `ILaunchSettingsProviderFactory`.

PR #6880, opened 2021-01-28
Post additional telemetry from project system components

swesonga:
The managed project system runs various operations to support scenarios such as intellisense, compilation in the IDE, etc. Two critical functions of are:

1. Requesting that NuGet run a restore operation on a project ("nominating a project for restore"). Without this the data we pass to the language service may be incomplete.
2. Passing project data to the language service. Without this, the language service will have no data (and thus no useful Intellisense).

This PR updates the components involved in these scenarios to make them submit telemetry indicating which operations have been started and completed. A few of the components for scenario (1) include the package restore data sources and progress tracker (e.g. PackageRestoreConfiguredInputDataSource, PackageRestoreUnconfiguredInputDataSource, PackageRestoreDataSource, and PackageRestoreProgressTracker.

The first three are parts of a data flow chain, with the first harvesting relevant data from the Configured projects and the third actually asking NuGet to perform the restore operation. The last handles reporting to the Operation Progress system.

For scenario (2), the relevant types are those associated with Roslyn's IWorkspaceProjectContext. These include classes like ApplyChangesToWorkspaceContext (responsible for actually calling the methods on that interface as appropriate, based on the project changes), and ApplyChangesToWorkspaceContext (which manages the lifetime of the IWorkspaceProjectContext, invokes the ApplyChangesToWorkspaceContext as necessary, and reports status back to the Operation Progress system).

This description is a paraphrasing of a write-up by @tmeschter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6880)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    internal class PackageRestoreOperationNames
```
drewnoakes, line 5:
```suggestion
    internal static class PackageRestoreOperationNames
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.cs
```diff
@@ -48,20 +52,38 @@ internal partial class WorkspaceProjectContextHost : AbstractMultiLifetimeCompon
...
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(_projectId))
+                {
+                    string? fullPath = _project?.UnconfiguredProject?.FullPath;
+                    _projectId = string.IsNullOrEmpty(fullPath) ? string.Empty : _languageServiceTelemetryService.HashValue(fullPath!);
+                }
+
+                return _projectId!;
```
drewnoakes, line 68: I think this would work to remove some `!`s:

```suggestion
                    _projectId = Strings.IsNullOrEmpty(fullPath) ? string.Empty : _languageServiceTelemetryService.HashValue(fullPath);
                }

                return _projectId;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/ILanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
```
drewnoakes, line 8: The implemetnation of this interface imports `UnconfiguredProject`, so this interface should declare itself as being in the `UnconfiguredProject` scope.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/LanguageServiceOperationNames.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    internal class LanguageServiceOperationNames
```
drewnoakes, line 5:
```suggestion
    internal static class LanguageServiceOperationNames
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreConfiguredInputDataSource.cs
```diff
@@ -58,6 +64,8 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreUpdate> tar
             // and someone is blocked on us on the same thread, the call proceeds
             JoinUpstreamDataSources(source);
 
+            _packageReferenceTelemetryService.LogPackageRestoreEvent(new PackageRestoreTelemetryEvent(PackageRestoreOperationNames.PackageRestoreConfiguredInputDataSourceLinkedToExternalInput, ContainingProject.FullPath));
```
drewnoakes, line 67: Personally I would remove the new event classes. They don't seem to add much value, and require a heap allocation.

Instead, I'd move their constructor arguments to the `Log*Event` method directly, making the call sites simpler too:

```suggestion
            _packageReferenceTelemetryService.LogPackageRestoreEvent(PackageRestoreOperationNames.PackageRestoreConfiguredInputDataSourceLinkedToExternalInput, ContainingProject.FullPath);
```

One reason for using a class is if you expect to add members over time, as that can then be done in a non-API-breaking fashion (i.e. get/set property). However these are internal types so that's not a consideration in this case.

I don't feel too strongly about this. Just something to consider.

Alternatively convert them to `record`s. But I still don't think we need the runtime overhead or the extra API space.

drewnoakes, line 67: Most calls to these `Log*Event` methods seem to be passing a value that's derived from either the `ConfiguredProject` or `UnconfiguredProject`. Consider whether you can move the `IPackageRestoreTelemetryService`/`ILanguageServiceTelemetryService` interfaces into one of those scopes and have their implementations import the relevant projects directly, so that callers don't need to pass them on every call.

drewnoakes (commented):
Some minor feedback below.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/ILanguageServiceTelemetryService.cs
swesonga: Which implementation class are you referring to? I don't see the UnconfiguredProject import.

---
In reply to: [566549849](https://github.com/dotnet/project-system/pull/6880#discussion_r566549849) [](ancestors = 566549849)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreConfiguredInputDataSource.cs
swesonga: I think this is a good idea. This was actually how I started out but thought about the API-breaking possibilities. You're right that this shouldn't be an issue since the API is internal.

---
In reply to: [566551743](https://github.com/dotnet/project-system/pull/6880#discussion_r566551743) [](ancestors = 566551743)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreTelemetryService.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <param name="packageRestoreOperationName">The name of the specific package restore operation.</param>
+        /// <param name="fullPath">The full path to the project that needs a package restore.</param>
+        public void PostPackageRestoreEvent(string packageRestoreOperationName, string fullPath)
+        {
+            string projectId = GetProjectId(fullPath);
+
+            _telemetryService.PostProperties(TelemetryEventName.ProcessPackageRestore, new (string propertyName, object propertyValue)[]
+                {
+                    ( TelemetryPropertyName.PackageRestoreOperation, packageRestoreOperationName),
+                    ( TelemetryPropertyName.PackageRestoreProjectId,  projectId),
```
MiYanni, line 32: Nit:
```suggestion
                    ( TelemetryPropertyName.PackageRestoreProjectId, projectId),
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreTelemetryService.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <param name="isRestoreUpToDate">Flag indicating whether the restore is up to date.</param>
+        public void PostPackageRestoreEvent(string packageRestoreOperationName, string fullPath, bool isRestoreUpToDate)
+        {
+            string projectId = GetProjectId(fullPath);
+
+            _telemetryService.PostProperties(TelemetryEventName.ProcessPackageRestore, new (string propertyName, object propertyValue)[]
+                {
+                    ( TelemetryPropertyName.PackageRestoreIsUpToDate, isRestoreUpToDate),
+                    ( TelemetryPropertyName.PackageRestoreOperation, packageRestoreOperationName),
+                    ( TelemetryPropertyName.PackageRestoreProjectId,  projectId),
```
MiYanni, line 50: Nit:
```suggestion
                    ( TelemetryPropertyName.PackageRestoreProjectId, projectId),
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreTelemetryService.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        private string GetProjectId(string fullPath) => string.IsNullOrEmpty(fullPath) ? string.Empty : _telemetryService.HashValue(fullPath);
+
+        /// <summary>
+        /// Posts a telemetry events from package restore components.
+        /// </summary>
+        /// <param name="packageRestoreOperationName">The name of the specific package restore operation.</param>
+        /// <param name="fullPath">The full path to the project that needs a package restore.</param>
+        public void PostPackageRestoreEvent(string packageRestoreOperationName, string fullPath)
```
MiYanni, line 25: Nit: `fullPath` might be better named as `projectFullPath` since all the calls use a project's path. Also, the description of the parameter says it is a project path too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreProgressTracker.PackageRestoreProgressTrackerInstance.cs
```diff
@@ -87,15 +91,20 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
                         linkOptions: DataflowOption.PropagateCompletion,
                         cancellationToken: cancellationToken);
 
+                _packageReferenceTelemetryService.PostPackageRestoreEvent(PackageRestoreOperationNames.PackageRestoreProgressTrackerInstanceInitialized, ConfiguredProject.UnconfiguredProject.FullPath);
+
                 return Task.CompletedTask;
             }
 
             internal void OnRestoreCompleted(IProjectVersionedValue<ValueTuple<IProjectSnapshot, RestoreData>> value)
```
MiYanni, line 99: Nit: Might be best to replace `ValueTuple` with named ValueTuple via the `( )` syntax as you run into them. Otherwise, you end up with the calls below like `value.Value.Item1` which is not easily read on first glance.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/ILanguageServiceTelemetryService.cs
swesonga: I got your point from the comment on the PackageRestoreConfiguredInputDataSource. Investigating this change.

---
In reply to: [566994525](https://github.com/dotnet/project-system/pull/6880#discussion_r566994525) [](ancestors = 566994525,566549849)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    internal static class PackageRestoreOperationNames
+    {
+        internal const string BeginNominateRestore = nameof(BeginNominateRestore);
+        internal const string EndNominateRestore = nameof(EndNominateRestore);
+        internal const string PackageRestoreConfiguredInputDataSourceLinkedToExternalInput = nameof(PackageRestoreConfiguredInputDataSourceLinkedToExternalInput);
+        internal const string PackageRestoreDataSourceLinkedToExternalInput= nameof(PackageRestoreDataSourceLinkedToExternalInput);
```
MiYanni, line 10: Nit:
```suggestion
        internal const string PackageRestoreDataSourceLinkedToExternalInput = nameof(PackageRestoreDataSourceLinkedToExternalInput);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreDataSourceLoading= nameof(PackageRestoreDataSourceLoading);
```
MiYanni, line 11: Nit:
```suggestion
        internal const string PackageRestoreDataSourceLoading = nameof(PackageRestoreDataSourceLoading);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreDataSourceUnloading= nameof(PackageRestoreDataSourceUnloading);
```
MiYanni, line 12: Nit:
```suggestion
        internal const string PackageRestoreDataSourceUnloading = nameof(PackageRestoreDataSourceUnloading);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreProgressTrackerActivating= nameof(PackageRestoreProgressTrackerActivating);
```
MiYanni, line 13: Nit:
```suggestion
        internal const string PackageRestoreProgressTrackerActivating = nameof(PackageRestoreProgressTrackerActivating);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreProgressTrackerDeactivating= nameof(PackageRestoreProgressTrackerDeactivating);
```
MiYanni, line 14: Nit:
```suggestion
        internal const string PackageRestoreProgressTrackerDeactivating = nameof(PackageRestoreProgressTrackerDeactivating);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreProgressTrackerInstanceInitialized= nameof(PackageRestoreProgressTrackerInstanceInitialized);
```
MiYanni, line 15: Nit:
```suggestion
        internal const string PackageRestoreProgressTrackerInstanceInitialized = nameof(PackageRestoreProgressTrackerInstanceInitialized);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreProgressTrackerRestoreCompleted= nameof(PackageRestoreProgressTrackerRestoreCompleted);
```
MiYanni, line 16: Nit:
```suggestion
        internal const string PackageRestoreProgressTrackerRestoreCompleted = nameof(PackageRestoreProgressTrackerRestoreCompleted);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreOperationNames.cs
```diff
@@ -0,0 +1,19 @@
...
+        internal const string PackageRestoreUnconfiguredInputDataSourceLinkedToExternalInput= nameof(PackageRestoreUnconfiguredInputDataSourceLinkedToExternalInput);
```
MiYanni, line 17: Nit:
```suggestion
        internal const string PackageRestoreUnconfiguredInputDataSourceLinkedToExternalInput = nameof(PackageRestoreUnconfiguredInputDataSourceLinkedToExternalInput);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/ILanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    internal interface ILanguageServiceTelemetryService
```
MiYanni, line 10: I don't know the convention here, but would `ILanguageServiceTelemetryEvents` be a more accurate name? I don't know if this interface itself represents a 'Service`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/ILanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        /// <summary>
+        /// Hashes personally identifiable information for telemetry consumption.
+        /// </summary>
+        /// <param name="value">Value to hashed.</param>
+        /// <returns>Hashed value.</returns>
+        string HashValue(string value);
```
MiYanni, line 17: Since this name is a bit generic, would something relating to PII make sense to put in the name? Like, `HashPiiValue`?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/IPackageRestoreTelemetryService.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    internal interface IPackageRestoreTelemetryService
+    {
+        /// <summary>
+        /// Posts a telemetry events from package restore components.
+        /// </summary>
+        /// <param name="packageRestoreOperationName">The name of the specific package restore operation.</param>
+        /// <param name="fullPath">The full path to the project that needs a package restore.</param>
+        void PostPackageRestoreEvent(string packageRestoreOperationName, string fullPath);
```
MiYanni, line 16: Note: Left the comment for `fullPath` at implementation (should have probably been here). 😜

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/ILanguageServiceTelemetryServiceFactory.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    internal static class ILanguageServiceTelemetryServiceFactory
+    {
+        public static ILanguageServiceTelemetryService Create()
+        {
+            var mock = new Mock<ILanguageServiceTelemetryService>();
+
+            return mock.Object;
```
MiYanni, line 13: Is there a reason we prefer this boilerplate pattern over simply calling `new Mock` at the callsite?

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IPackageRestoreTelemetryServiceFactory.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.VisualStudio.Telemetry
+{
+    internal static class IPackageRestoreTelemetryServiceFactory
+    {
+        public static IPackageRestoreTelemetryService Create()
+        {
+            var mock = new Mock<IPackageRestoreTelemetryService>();
+
+            return mock.Object;
```
MiYanni, line 13: Same boilerplate here as above comment.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/LanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <returns>Hashed value.</returns>
+        public string HashValue(string value)
+        {
+            return _telemetryService.HashValue(value);
+        }
+
+        /// <summary>
+        /// Post a telemetry event when a design time build fails.
+        /// </summary>
+        /// <param name="projectId">A unique identifier for the project.</param>
```
drewnoakes, line 32: For interface implementation methods, if the docs are identical to those of the base I'd just remove them. There's less chance that they fall out of sync then.. The IDE will make documentation from the base visible here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/LanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void PostDesignTimeBuildFailureEvent(string projectId)
+        {
+            _telemetryService.PostProperties(TelemetryEventName.DesignTimeBuildComplete, new (string propertyName, object propertyValue)[]
+                {
+                    ( TelemetryPropertyName.DesignTimeBuildCompleteSucceeded, false),
```
drewnoakes, line 37: This `false` will be boxed. Might be good to reuse a boxed value by assigning it to a field.

CPS has an internal `Boxed` class for such values. Maybe we create one in this repo too.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.cs
```diff
@@ -48,20 +52,38 @@ internal partial class WorkspaceProjectContextHost : AbstractMultiLifetimeCompon
...
+            get
+            {
+                if (Strings.IsNullOrEmpty(_projectId))
+                {
+                    string? fullPath = _project?.UnconfiguredProject?.FullPath;
+                    _projectId = Strings.IsNullOrEmpty(fullPath) ? string.Empty : _languageServiceTelemetryService.HashValue(fullPath);
+                }
+
+                return _projectId;
+            }
```
drewnoakes, line 69: As I understand it, this value is only used for telemetry, but its name suggests something grander.

If the language service telemetry component was in the `ConfiguredProject` scope, it could import the `ConfiguredProject` via its constructor and cache this value itself, keeping this telemetry concept out of the language service code. It seems you could then remove `HashValue` from the interface as well.

Looking further, it seems that this usage is in the configured project, and a usage below is in the unconfigured project scope, so the approach I describe here would only work if you split that service in two. I'm less sold on the idea, but thought I'd leave this thought here in case it prompts something.

At the very least, consider renaming to something like `TelemetryId`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/PackageRestoreTelemetryService.cs
drewnoakes: If the `UnconfiguredProject` was imported into the constructor, I think these `fullPath` parameters could be removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.cs
swesonga: I have split the interface into 2 in [https://github.com/dotnet/project-system/commit/d718c99e5c46290d4ee63327c83188346ece0b9f](https://github.com/dotnet/project-system/commit/d718c99e5c46290d4ee63327c83188346ece0b9f) so that unconfigured and configured scopes are handled separately since the most likely change that could be needed is reporting additional telemetry and this seems like the cleaner way to handle that.

---
In reply to: [567238918](https://github.com/dotnet/project-system/pull/6880#discussion_r567238918) [](ancestors = 567238918)

swesonga: Still working on a fix for the duplication of the ProjectTelemetryId property though.

---
In reply to: [567490682](https://github.com/dotnet/project-system/pull/6880#discussion_r567490682) [](ancestors = 567490682,567238918)

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/ILanguageServiceTelemetryServiceFactory.cs
swesonga: I did this to be consistent with the other tests I looked at. My interpretation was that the goal of this usage of the factory pattern is to abstract the exact manner in which an implementation of the returned interface is created so that, for example, a switch from XUnit to Moq to something else doesn't need the individual tests to be modified. Perhaps other folks might have some thoughts about this?

---
In reply to: [567169220](https://github.com/dotnet/project-system/pull/6880#discussion_r567169220) [](ancestors = 567169220)

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/ILanguageServiceTelemetryService.cs
swesonga: Out of curiosity, what do you consider a service? Would this apply to the existing ITelemetryService as well?

---
In reply to: [567168061](https://github.com/dotnet/project-system/pull/6880#discussion_r567168061) [](ancestors = 567168061)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/ConfiguredProjectLanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            _project = project;
+            _telemetryService = telemetryService;
+        }
+
+        private string ProjectTelemetryId
+        {
+            get
+            {
+                if (Strings.IsNullOrEmpty(_projectHash))
```
drewnoakes, line 27: Should this just be a null check? It can be assigned `""` in this block. I wouldn't expect `project.FullPath` to become non-null at any point. Ditto in other service classes.

This chunk of code is repeated a few times in this PR. Perhaps you could push it into the `HashValue` method (under a slightly different name), making this property something like:

```c#
private string ProjectTelemetryId => _projectHash ??= _telemetryService.GetProjectId(_project?.UnconfiguredProject);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/ConfiguredProjectLanguageServiceTelemetryService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                return _projectHash;
+            }
+        }
+
+        public void PostDesignTimeBuildFailureEvent()
+        {
+            _telemetryService.PostProperties(TelemetryEventName.DesignTimeBuildComplete, new (string propertyName, object propertyValue)[]
+                {
+                    ( TelemetryPropertyName.DesignTimeBuildCompleteSucceeded, false),
```
drewnoakes, line 41: nit: spaces after opening paren here and elsewhere.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Telemetry/ConfiguredProjectLanguageServiceTelemetryService.cs
swesonga: I assumed that any spacing not changed by Edit > Advanced > Format Document was fine. Will fix.

---
In reply to: [567515572](https://github.com/dotnet/project-system/pull/6880#discussion_r567515572) [](ancestors = 567515572)

drewnoakes: That's fair. I guess either our code style rules are underspecified, or it's not possible to control this particular styling quirk. It's not a big deal, so don't block merging for this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreProgressTracker.PackageRestoreProgressTrackerInstance.cs
swesonga: makes sense. I will change that in a follow-up PR though.

---
In reply to: [567164943](https://github.com/dotnet/project-system/pull/6880#discussion_r567164943) [](ancestors = 567164943)

tmeschter (changes requested):
What I would like to see before this is merged in is an explanation about why we're capturing this information, and how we're going to use each event. I want to make sure we understand why it is important to us, and that we've thought through how we're going to process and report the data. For example, how are we going to use the telemetry related to the workspace project context to figure out if there are problems passing data to the language service?

I think the doc comments for each specific `Log*Event` method would be a good place to capture this information.

swesonga:
> What I would like to see before this is merged in is an explanation about why we're capturing this information, and how we're going to use each event. I want to make sure we understand why it is important to us, and that we've thought through how we're going to process and report the data. For example, how are we going to use the telemetry related to the workspace project context to figure out if there are problems passing data to the language service?
> 
> I think the doc comments for each specific `Log*Event` method would be a good place to capture this information.

@jjmew  has provided the same feedback. I'm writing this up.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/IConfiguredProjectLanguageServiceTelemetryService.cs
```diff
@@ -6,26 +6,73 @@
 
 namespace Microsoft.VisualStudio.Telemetry
 {
+    /**
```
drewnoakes, line 9: `/**` has no special meaning in C# (unlike in Java). This could be a `///` style comment instead. A few other instances below too.

drewnoakes (approved):

swesonga:
@tmeschter I've called out the key events in the IConfiguredProjectLanguageServiceTelemetryService and IConfiguredProjectPackageRestoreTelemetryService interfaces. I think using the events called out there is rudimentary but probably sufficient for a first iteration.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/IConfiguredProjectLanguageServiceTelemetryService.cs
swesonga: wow, muscle memory - definitely didn't catch myself doing this :D

---
In reply to: [569849335](https://github.com/dotnet/project-system/pull/6880#discussion_r569849335) [](ancestors = 569849335)

tmeschter (approved):

PR #6839, opened 2021-01-15
Updated build.cmd to use version minimum from Version.props

MiYanni:
Related: https://github.com/dotnet/project-system/issues/6744

- Removed temporary workaround that has been fixed
- Updated RequiredVSVersion to pull the version number from Versions.props which sets the ProjectSystemVersion
  - The current version in that file is `16.8.1`.
- Added arguments for calling `powershell` to reduce potential failures and removed progress bar for downloading VSWhere
  - Progress bars (traditionally) have overhead in Windows PowerShell.
  - We are using Windows PowerShell since we call `powershell`. PowerShell Core uses `pwsh`.
- Updated version of VSWhere used to latest release

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/project-system/pull/6839)

PR #6885, opened 2021-02-01
Bump version to match with VS and update Readme

dmonroym:
Increase version to 16.10 to maintain parity with VS and add 16.10 badge to Readme

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6885)


build/import/Versions.props
```diff
@@ -1,7 +1,7 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ProjectSystemVersion>16.8.1</ProjectSystemVersion>
+    <ProjectSystemVersion>16.10.0</ProjectSystemVersion>
```
drewnoakes, line 4: What is this property used for? Does it matter that we skipped a 16.9 value?

MiYanni (approved):

build/import/Versions.props
tmeschter: Based on the errors in the build, we enforce that the specified version of Visual Studio is installed. If that is all we use it for, we should consider renaming it.

dmonroym: We used to have `VisualStudioVersion` and `ProjectSystemVersion` but they used to match always so we decided to keep them the same. 16.9 was skipped by mistake but it only affects our package names so just mildly annoying but not a big impact. The build machines aren't always updated with the latest VS version so it might be useful to bring the `VisualStudioVersion` back.

dmonroym:
https://github.com/dotnet/project-system/pull/6839 also has an impact in this not building properly. I'll create a `RequiredVSVersion` property instead

build/import/Versions.props
```diff
@@ -1,7 +1,8 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ProjectSystemVersion>16.8.1</ProjectSystemVersion>
+    <MinimumRequiredVSVersion>16.8.1</MinimumRequiredVSVersion> <!-- We should try to keep this up to date with the highest version available on build machines -->
```
swesonga, line 4: This property name is misleading since it is the minimum VS version required to build the code, not to run the resulting binaries. I think VisualStudioVersion would be misleading as well. Perhaps MinimumVSVersionForBuild?

PR #6890, opened 2021-02-01
Fix CanBeStartupProjectAsync exception

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/6918

This sets `validateSettings` to `false` for `QueryDebugTargetsAsync` call in `CanBeStartupProjectAsync` since this check is ran prior to debugging. Prior to https://github.com/dotnet/project-system/pull/6871, these checks were being ran in `CanBeStartupProjectAsync`, but the exceptions being thrown we're being caught and ignored. Now, we're not running those checks since they aren't supposed to be ran as part of this method in the first place.


drewnoakes:
Once #6891 is merged, that PR will automatically be forward-integrated to `main` which will conflict with this. I'd hold off on merging this until that flow completes, so you can resolve conflicts here rather than the FI PR.

drewnoakes:
@MiYanni the FI is complete so you can pick this up again now.

MiYanni:
@drewnoakes Remerged with main. Should be good now.

dmonroym (approved):

drewnoakes (approved):

PR #6891, opened 2021-02-01
Suppress exceptions in CanBeStartupProjectAsync

MiYanni:
Adds a catch for CanBeStartupProjectAsync to suppress all exceptions. This is the 16.9 fix of https://github.com/dotnet/project-system/pull/6890, which is the 16.10 fix. We're doing 2 different fixes since we don't know what other potential exceptions might be thrown. The 16.10 fix works specifically for the issues filed in that PR. The PR here is to simply add the catch that was previously there as to suppress any potential exceptions for the 16.9 release.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6891)


drewnoakes (approved):

swesonga (approved):

drewnoakes:
Relates to AB#1272508

Issue #6820, opened 2021-01-07
Roslyn is unable to add diagnostic nodes properly in multi-targeting projects


jasonmalinowski:
Roslyn is responsible for populating the individual diagnostics and source generators under an analyzer node. In a multi-targeting case, we have logic that attempts to walk up the hierarchy of nodes to figure out which TFM is active. For example, consider this screenshot:

![image](https://user-images.githubusercontent.com/201340/103954695-a0d00c00-50f9-11eb-8ce8-3d3e923f0818.png)

The "SampleSourceGenerators" node is produced by CPS. Roslyn then walks up that node and finds the netcoreapp3.1 node, which has some special capabilities, namely a capability that starts with `$TFM:` -- that is then used to figure out which TFM we're looking under.

Looking at the Roslyn code however, I'm guessing this has been broken since we removed the IVsHierarchy wrappers around each target a long time ago. The problem is we may know which TFM we're looking at, but we have no way to figure out which IWorkspaceProjectContext this actually matches. The code currently looks at each IVsHierarchy for our project, and asks the IVsHierarchy for it's TFM. That works if we had separate IVsHierachies for each TFM but we don't (and we shouldn't...getting rid of that code was good.) Right now you'll see the feature work in _one_ of the targets under Dependencies simply because when you ask an IVsHierarchy that multi-targets for it's TFM, you just get the first one.

Note the user impact here is worse now -- previously this only meant you couldn't browse diagnostics provided by an analyzer. This is also where we're showing documents from source generators, so you can't browse to other options.

My proposal is the project system should, rather than giving the `$TFM:` capability, give some capability that is:

1. The ID directly obtained from IWorkspaceProjectContext.Id that matches up to the Roslyn project created for that node. This means we could just match things directly. Note this option though presumes that the capabilities of a node can change if the design time build triggers a project reload.
2. The unique name given to IWorkspaceProjectContextFactory.CreateProjectContext(). This will be stable; there's also precedent for this being exposed as VSHPROPID_ActiveIntellisenseProjectContext from the IVsHierarchy. Risk here is if the name has spaces, since I guess capabilities can't have spaces?

Otherwise we need some way to directly be told the TFM for a project, but I don't like binding directly to TFM here since that will require this to be redesigned again if we ever have multiple Roslyn projects for the same TFM, say if we ever do the work to expose both debug/release, etc.

drewnoakes:
[...]

PR #6893, opened 2021-02-02
Set TargetFramework on IWorkspaceProjectContext

drewnoakes:
Relates to #6820

Roslyn needs a way to identify which workspace project context a given dependencies tree node relates to. Ideally we would store the project context ID on the node itself for Roslyn to query, however we are late in the 16.9 cycle and the magnitude of that change would likely make it too risky. For now, we pass the `TargetFramework` property value to the language service, so it can match the value against the `$TFM:net5.0`-style value found in the nodes capabilities property value.

---

@jasonmalinowski would this change be useful for you in the 16.9 at this point? In 16.10 we could revert this and make the context ID available on the browse object.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6893)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
@@ -86,6 +86,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
                                                                                     hostObject,
                                                                                     data.BinOutputPath);
 
+                context.SetProperty(ConfigurationGeneral.TargetFrameworkProperty, data.TargetFramework ?? "");
+
                 context.LastDesignTimeBuildSucceeded = false;  // By default, turn off diagnostics until the first design time build succeeds for this project.
 
                 return context;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
@@ -86,6 +86,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
                                                                                     hostObject,
                                                                                     data.BinOutputPath);
 
+                // Roslyn needs a way to identify which workspace project context a given dependencies tree node relates to.
+                // Ideally we would store the project context ID on the node itself for Roslyn to query, however we are late
+                // in the 16.9 cycle and the magnitude of that change would likely make it too risky. For now, we pass the
+                // TargetFramework property value to the language service, so it can match the value against the
+                // "$TFM:net5.0"-style value found in the nodes capabilities property value.
                 context.SetProperty(ConfigurationGeneral.TargetFrameworkProperty, data.TargetFramework ?? "");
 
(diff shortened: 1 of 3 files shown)


swesonga (approved):

davkean:
This is a user property, is TFM also contain the user property? The moniker (ie the "M" in "TFM") makes it sounds like its a canonical representation.

jasonmalinowski:
@drewnoakes: I'll try this out this afternoon. Thanks for taking a look!

drewnoakes:
> This is a user property, is TFM also contain the user property? The moniker (ie the "M" in "TFM") makes it sounds like its a canonical representation.

Yeah it used to be TFM, but because of the platform suffix in .NET 5 it became TF, just with the same `$TFM` prefix. We will tidy this up in 16.10. There will be a fix on the NuGet side too.

MiYanni (approved):

jasonmalinowski:
A quick smoke test with this PR is looking good from the Roslyn side.

drewnoakes:
@jasonmalinowski that's encouraging. Say the word and I'll start the machinery to get this put to the QB. Do you have an existing bug on AzDO?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
```diff
@@ -86,6 +86,8 @@ public async Task ReleaseProjectContextAsync(IWorkspaceProjectContextAccessor ac
                                                                                     hostObject,
                                                                                     data.BinOutputPath);
 
+                context.SetProperty(ConfigurationGeneral.TargetFrameworkProperty, data.TargetFramework ?? "");
```
tmeschter, line 89: I know we're going to back this out in 16.10 but it would be useful to have a comment explaining why we need to send the `TargetFramework`.

tmeschter (approved):

davkean:
> Yeah it used to be TFM, but because of the platform suffix in .NET 5 it became TF, just with the same $TFM prefix. We will tidy this up in 16.10. There will be a fix on the NuGet side too.

Just to clarify because I think you misunderstood my question; the TargetFramework property is not the TargetFrameworkMoniker/TargetPlatformMoniker. The TargetFramework is a user property that can be used by SDK to figure values for the latter but they are not same thing. The latter are canonical values and have a single representation for a framework/platform, the former can have many representations and really shouldn't be used for anything other than passing to NuGet so that it can write conditions correctly.

My question is; is the value that is passed to $TFM the _exact_ value of TargetFramework property? Or is it roundtripped through NuGet, trim'd or manipulated in anyway?

davkean:
As spoken offline:

1) Call this property something other than TargetFramework to avoid someone in Roslyn thinking this is the actual TargetFramework
2) You don't need to manually set this, just change this rule: https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/LanguageService.xaml.

jasonmalinowski:
@drewnoakes: I'm using https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1275676 for QB approval.

drewnoakes:
@jasonmalinowski after chatting with @davkean I'm going to change the implementation here a bit, and rename the property to something like `TemporaryDependencyNodeTargetIdentifier`, so that its purpose is clear.

drewnoakes:
This PR has been replaced by #6932 which sets the value via rule file (a more robust and simpler approach) and renames the property set on the context from `TargetFramework` to `TemporaryDependencyNodeTargetIdentifier` to discourage use of this value for other purposes and to indicate it will be removed.

PR #6930, opened 2021-02-07
Use correct project value when finding target node

drewnoakes:
In #6873 we changed the dependencies tree to no longer use the `TargetFrameworkMoniker` to identify per-target branches of the dependencies tree. This fixed an issue where `net5.0` and `net5.0-windows` would merge into one, because they have identical `TargetFrameworkMoniker` values.

Code in `DependenciesTreeProjectSearchContext` was missed during that change. This commit brings it up to date.

Missing this usage broke searching in the dependencies tree for multi-targeting projects, causing such searches to return no results within the dependencies tree.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6930)


MiYanni (approved):

tmeschter (approved):

tmeschter:
I don't suppose there's a way to add a unit test around this?

PR #6932, opened 2021-02-08
Allow language service to identify dependency tree target nodes

drewnoakes:
Relates to issue #6820.
Replaces PR #6893.

Roslyn needs a way to identify which workspace project context a given dependencies tree node relates to. Ideally we would store the project context ID on the node itself for Roslyn to query, however we are late in the 16.9 cycle and the magnitude of that change would likely make it too risky. For now, we pass the `TargetFramework` property value (aliased as `TemporaryDependencyNodeTargetIdentifier`) to the language service, so it can match the value against the `$TFM:net5.0`-style value found in the node's capabilities property value.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6932)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/LanguageService.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/LanguageService.xaml
@@ -32,4 +32,26 @@
   <StringProperty Name="RunAnalyzersDuringLiveAnalysis"
                   ReadOnly="True"
                   Visible="False" />
+
+  <!--
+      Roslyn needs a way to identify which workspace project context a given dependencies tree node relates to.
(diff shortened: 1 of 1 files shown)


drewnoakes:
@jasonmalinowski I've tested this locally and it sets the same value as before on the workspace context, though with the new property name.

Once this is approved and merged I'll get an insertion lined up and linked to your ask mode issue.

jjmew (approved):
:shipit:

dmonroym (approved):

tmeschter (approved):

MiYanni (approved):

drewnoakes:
Insertion PR https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/303039

Issue #6858, opened 2021-01-23
We need to understand the plan for deprecation of BinaryFormatter


jjmew:
We need to understand the plan for deprecation of BinaryFormatter. We use it for the Resource Designer but we need to move to a new solution as it will be deprecated.  You can see https://github.com/dotnet/runtime/issues/39292

MiYanni:
Current plan is to replace BinaryFormatter usage. It is used for undo/redo of the old property pages and used for drag & drop/copy-paste in the resource designer. I'm currently working on a solution using System.Text.Json.

PR #6969, opened 2021-02-17
Add telemetry entries for BinaryFormatter callsites

MiYanni:
Related: https://github.com/dotnet/project-system/issues/6858

The purpose of this PR is to add telemetry logging for all callsites of [BinaryFormatter](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter?view=net-5.0) to monitor if they are actually called. Initial testing could only activate one spot, [SerializeObject](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceSerializationService_Store.vb#L581-L585) in `ResourceSerializationService_Store.vb`. All other callsites were not able to be activated with a combination of difference projects and usage of the Resource Editor. Therefore, we are adding telemetry and will monitor if there is any valid usage of this code.

Currently, we only use `BinaryFormatter` in the `AppDesigner` and `Editors` projects. These are both VB projects. There are bits and pieces of code that seems copy-pasted between the two. Here, the `Utils` file in each project has a `TelemetryLogger` class. I've added the same functions to each for logging usage of `BinaryFormatter`. The only difference between the two functions is the telemetry path has the appropriate project, *appdesigner* and *editors* respectively.

Unfortunately, the site to access the Telemetry Logger application is down. I'll be testing these changes once that site comes back up. But, I wanted to get feedback ASAP since this is for 16.9.

Edit: Even though the site for the app is down, I pulled the code for the logger, built it, and was able to test my changes. Here is a snippet of the log to show the changes are working as expected:
image
image

Remember, that's the only function I've been able to activate. But the telemetry events use the same function, so I don't expect issues with the other callsites.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6969)


src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -664,6 +664,21 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
 
                 TelemetryService.DefaultSession.PostEvent(userTask)
             End Sub
+
+            Private Const BinaryFormatterEventName As String = "vs/projectsystem/appdesigner/binaryformatter"
+            Public Enum BinaryFormatterType
```
drewnoakes, line 669: `BinaryFormatterOperation`?

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -664,6 +664,21 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
...
+                Serialize = 0
+                Deserialize = 1
+            End Enum
+
+            Public Shared Sub LogBinaryFormatterEvent(functionName As String, className As String, type As BinaryFormatterType)
```
drewnoakes, line 674: You can use `CallerMemberName` so that call sites don't have to pass their function name. The compiler will populate the argument automatically. There's no `CallerTypeName` unfortunately though.

```suggestion
            Public Shared Sub LogBinaryFormatterEvent(className As String, type As BinaryFormatterType, <System.Runtime.CompilerServices.CallerMemberName> Optional functionName As String = Nothing)
```

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
```diff
@@ -1641,6 +1641,20 @@ Namespace Microsoft.VisualStudio.Editors.Common
                 TelemetryService.DefaultSession.PostEvent(userTask)
             End Sub
 
+            Private Const BinaryFormatterEventName As String = "vs/projectsystem/editors/binaryformatter"
+            Public Enum BinaryFormatterType
+                Serialize = 0
+                Deserialize = 1
+            End Enum
+
+            Public Shared Sub LogBinaryFormatterEvent(functionName As String, className As String, type As BinaryFormatterType)
```
drewnoakes, line 1650: Ditto, re `CallerMemberName `.

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
```diff
@@ -1641,6 +1641,20 @@ Namespace Microsoft.VisualStudio.Editors.Common
                 TelemetryService.DefaultSession.PostEvent(userTask)
             End Sub
 
+            Private Const BinaryFormatterEventName As String = "vs/projectsystem/editors/binaryformatter"
+            Public Enum BinaryFormatterType
```
drewnoakes, line 1645: Ditto `BinaryFormatterOperation`.

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropertyPageSerializationService_Store.vb
```diff
@@ -131,7 +132,9 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
             ''' <param name="Stream">The stream to load from</param>
             Public Shared Function Load(Stream As Stream) As PropertyPageSerializationStore
                 Dim f As New BinaryFormatter
-                Return DirectCast(f.Deserialize(Stream), PropertyPageSerializationStore)
+                Dim StreamObject As Object = f.Deserialize(Stream)
+                TelemetryLogger.LogBinaryFormatterEvent(NameOf(Load), NameOf(PropertyPageSerializationStore), TelemetryLogger.BinaryFormatterType.Deserialize)
+                Return DirectCast(StreamObject, PropertyPageSerializationStore)
```
drewnoakes, line 137: I see throughout this PR that you're deserialising before logging telemetry. Is there a reason for that? Logging first would make it a smaller diff.

drewnoakes (approved):

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -664,6 +664,21 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
...
+                Dim userTask = New UserTaskEvent(BinaryFormatterEventName, TelemetryResult.Success)
+                userTask.Properties("vs.projectsystem.appdesigner.binaryformatter.functionname") = functionName
```
swesonga, line 676: would it be better to use a constant string prefix like the TelemetryPropertyName class does?

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropertyPageSerializationService_Store.vb
swesonga: Logging after using the binary formatter also means that a failure will not result in a telemetry event (even though I think you would expect one)

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: I avoided this originally as I thought it used reflection. Looked it up; it does not. https://stackoverflow.com/a/22580624/294804

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropertyPageSerializationService_Store.vb
MiYanni: I don't know the standard practice for telemetry. I'm used to debug logging, in which, you'd log an event after it occurs, since the log entry indicates it *has* happened. If you log prior, you can't actually determine if the operation happened or not. Since this particular set of data doesn't have a strong relationship to the operation itself, I can log preemptively. I only really care if the function itself was called at this moment.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Sure. I can do that.

MiYanni:
After the changes above, the appropriate properties are still being recorded.
image
image

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Done.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Done.

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
MiYanni: Done.

drewnoakes (approved):

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropertyPageSerializationService_Store.vb
MiYanni: Updated. Also normalized BinaryFormatter calling.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Updated both TelemetryLogger classes to use this pattern for all logging functions.

dmonroym (approved):

Issue #6720, opened 2020-10-29
List of missing properties


tmeschter:
(For the list of properties requiring attention see #6907.)

We're missing the below property pages and/or properties from the new Project Properties Designer. Note that this is a list of everything we think is reasonable to consider, and so will contain both items that exist in the old property pages and new items.

- [x] Build events pages and corresponding properties (#5953, PR #6058, PR #7119)
- [ ] ~Code Analysis page~ (moved to #7242)
- [ ] Build page
  - [x] `WarningsNotAsErrors` (#2409)
  - [x] MSBuild "warnings as errors" properties
    - MSBuild has different properties to control when warnings are treated as errors. NuGet seems to re-use the same properties as the C# compiler.
    - [ ] ~`CodeAnalysisTreatWarningsAsErrors`~ (moved to #6793)
  - [x] ~`UseWPF` and `UseWindowsForms`~ (moved to #7239)
  - [x] Debug information (symbols), eg: full, portable (PR #7123)
  - [ ] ~`ProduceReferenceAssembly` (needs investigation)~ (moved to #7245)
- [ ] ~Packaging page~
  - [ ] ~Show new README property~ (Moved to #6457)
- [ ] ~Signing page~
  - [ ] ~"Sign the ClickOnce manifests" property~ (moved to #7244)
- [x] ~VB-specific pages/properties (especially the VB Build page)~ (moved to #6935)
- [x] ~F#-specific pages/properties~ (moved to #6936)
- [x] ~WinForms-specific properties~ (moved to #6937)
- [x] ~WPF-specific properties~ (moved to #6938)
- [ ] .NET Framework-specific properties
  - [x] ~"Auto-generate binding redirects" on the Application page~ (captured by #5940)
  - [ ] ~"Prefer 32-bit" on the Build page~ (moved to #6989)
- [ ] ~"Assembly Information" properties (note some of these are found on the "Package" page; that may not be the best place).~ (moved to #7246)
  - [ ] Title
  - [ ] Description
  - [ ] Company
  - [ ] Product
  - [ ] Copyright
  - [ ] Trademark
  - [ ] Assembly version
  - [ ] File version
  - [ ] GUID
  - [ ] Neutral language
  - [ ] Make assembly COM-Visible (is this .NET Framework-specific?)

drewnoakes:
Remaining items have been split out into separate issues. Closing this.

PR #6971, opened 2021-02-18
Implement WarningsNotAsErrors in the new property pages

tmeschter:
Related to #6720.

### Background

The C# compiler provides a switch allowing the user to specify whether or not warnings _in general_ should be treated as errors. In addition to that, specific warnings can be treated as errors, or _blocked_ from being treated as errors. In the project file, these settings are stored in the `TreatWarningsAsErrors`, `WarningsAsErrors`, and `WarningsNotAsErrors` properties, respectively.

Currently, the project property pages only expose the first two of these. This change adds support for the third, `WarningsNotAsErrors`.

### Explanation

The first thing to note is that all we have to do to get the new control to show up in the UI *and* to get the property persisted to the right place in the project file is to add a new `StringProperty` element to BuildPropertyPage.xaml. The `Name` attribute uniquely identifies the property within the Build page, _and_ specifies the property name that is read and written in the project file.

Second, note the `VisibilityCondition` metadata on the new `WarningsNotAsErrors` and existing `WarningsAsErrors` property. We only want to show one or the other of these properties in the UI, not both. If warnings _in general_ are being treated as errors it makes sense to only show `WarningsNotAsErrors`; but if warnings are not being treated as errors then we should only show `WarningsAsErrors`. The `VisibilityCondition` contains an expression that controls whether or not the control should be displayed.

Before I get into specifics of the expressions implemented here, it is important to note that `TreatWarningsAsErrors`, `WarningsAsErrors`, and `WarningsNotAsErrors` are all configuration dependent properties. That is, their values could vary from one configuration to another, and so the property effectively has multiple values--one per configuration. There are also many configuration independent properties; these have at most one value.

We want the `WarningsAsErrors` control to show up if `TreatWarningsAsErrors` is false in _any_ configuration, and this is achieved with the following expression:

```
(has-evaluated-value "Build" "TreatWarningsAsErrors" "false")
```

This looks at all of the evaluated values of the `TreatWarningsAsErrors` property on the `Build` page and checks if any of them is "false"; if so, the overall expression evaluates to "true" and the control is displayed.

If the properties were configuration independent we could have gone with

```
(eq (evaluated "Build" "TreatWarningsAsErrors") "true")
```

to check if the evaluated value of TreatWarningsAsErrors is "true", or even just

```
(evaluated "Build" "TreatWarningsAsErrors")
```

Third, let's leave the .xaml file for the moment and look and the one piece of code, TreatWarningsAsErrorsValueProvider.cs. When the user toggles the value of `TreatWarningsAsErrors` we don't just want to hide or show `WarningsAsErrors` and/or `WarningsNotAsErrors`, we want to clear out the property values that no longer make sense. E.g., if we're setting `TreatWarningsAsErrors` to false, we can delete `WarningsNotAsErrors` entirely as it has no effect.

We accomplish this by implementing an `IInterceptingPropertyValueProvider` for `TreatWarningsAsErrors`. An intercepting value provider gets a chance to handle a property set before it is passed on for default handling; this allows it to manipulate the value or even other properties. It also has a chance to manipulate the values _read_ from the default handler before they are passed on to the property pages, but we don't need this ability here.

However, if the user toggles `TreatWarningsAsErrors` back and forth we don't want to permanently delete whatever values they had for the other two properties. Rather, we squirrel them away in `ITemporaryPropertyStorage` before deleting them, and restore the value from there when switching back.

Fourth, go back to the .xaml file and take a look at the `Rule.DataSource` entry at the top of the page. The `Persistence` attribute indicates that the property values will read from and written to the "ProjectFile" handler which, of course, reads and writes to the project file itself (as opposed to the .user file or some other .props or .targets file). However, as we want to intercept the `TreatWarningsAsErrors` property we update its `DataSource` entry to instead specify the "ProjectFileWithInterception" persistence handler. This handler runs the property through an applicable interceptors before handing it off to the "ProjectFile" handler. Without this, the TreatWarningsAsErrorsValueProvider would never be called.

Fifth and finally, note that the values of `WarningsAsErrors` and `WarningsNotAsErrors` now depend on the value of `TreatWarningsAsErrors`. We need to tell the UI that when `TreatWarningsAsErrors` is updated, it should re-query the values of the other two. This is done via the `DependsOn` metadata added to both of those properties' entries. Without this, the UI may become out of sync with the actual values in the project as the user makes changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6971)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/TreatWarningsAsErrorsValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/TreatWarningsAsErrorsValueProvider.cs
@@ -1,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties
+{
+    [ExportInterceptingPropertyValueProvider("TreatWarningsAsErrors", ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class TreatWarningsAsErrorsValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string WarningsAsErrorsProperty = "WarningsAsErrors";
+        private const string WarningsNotAsErrorsProperty = "WarningsNotAsErrors";
+        private readonly ITemporaryPropertyStorage _temporaryPropertyStorage;
+
+        [ImportingConstructor]
+        public TreatWarningsAsErrorsValueProvider(ITemporaryPropertyStorage temporaryPropertyStorage)
+        {
+            _temporaryPropertyStorage = temporaryPropertyStorage;
+        }
+
+        public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+        {
+            if (StringComparers.PropertyLiteralValues.Equals(unevaluatedPropertyValue, "true"))
+            {
+                // When setting this to "true", remove WarningsAsErrors
+                await defaultProperties.SaveValueIfCurrentlySetAsync(WarningsAsErrorsProperty, _temporaryPropertyStorage);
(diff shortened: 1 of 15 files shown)


drewnoakes (approved):
My only hesitation with this change is that a user who wants "All except ..." has to know to select "All" before they'll be given the option to specify exclusions. The inverse is true ("None" except).

Another way of implementing this is to change the enum property to have four values ("None", "Only specific", "All except specific", "All"). That would be a more complicated implementation, involving a hidden synthetic property. It might also be an interesting example to document.

Please add a link to this PR in https://github.com/dotnet/project-system/blob/main/docs/repo/property-pages/property-specification.md. We should add a section for good reference PRs.

tmeschter:
Yeah, once we everything implemented we're going to want to go through all the properties and figure out if we're providing the right user experience around them. For example, we should probably update the display names and descriptions for these and `NoWarn` to be consistent and make it clear how they all interact.

PR #6975, opened 2021-02-19
Fix dependency node when no TargetFramework(s) property exists in the project

### LINKED ISSUE #6974: Dependency node fails to initialise when project does not define TargetFramework
As part of making the Dependencies node work for WAP projects, I uncovered an issue where the node's data subscriptions are not constructed correctly.

Currently, the logic is to construct these subscriptions _only_ when the `TargetFramework` or `TargetFrameworks` properties change.

For a regular .NET project this is fine as that property will be defined, and when it is first observed it will trigger the subscriptions that allow the Dependencies node to function correctly.

For a project type that does not define these properties (i.e. WAP projects) this logic is incorrect.

The real fix is to tackle #6183. In the meantime we need a simple fix to unblock WAP projects.

drewnoakes:
Fixes #6974

Currently, the logic is to construct these subscriptions _only_ when the `TargetFramework` or `TargetFrameworks` properties change.

For a regular .NET project this is fine as that property will be defined, and when it is first observed it will trigger the subscriptions that allow the Dependencies node to function correctly.

For a project type that does not define these properties (i.e. WAP projects) this logic is incorrect.

This fix ensures a `AggregateCrossTargetProjectContext` when the current instance is null.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6975)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
@@ -311,7 +311,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies.Subscriptions
         private Task OnConfiguredProjectEvaluatedAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> e)
         {
             // If "TargetFrameworks" property has changed, we need to refresh the project context and subscriptions.
-            if (HasTargetFrameworksChanged())
+            // If no context exists yet, create one.
+            if (HasTargetFrameworksChanged() || _context.Current == null)
             {
                 return UpdateProjectContextAndSubscriptionsAsync();
             }


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
```diff
@@ -311,7 +311,8 @@ private void DisposeCore()
         private Task OnConfiguredProjectEvaluatedAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> e)
         {
             // If "TargetFrameworks" property has changed, we need to refresh the project context and subscriptions.
-            if (HasTargetFrameworksChanged())
+            // If no context exists yet, create one.
+            if (HasTargetFrameworksChanged() || _context.Current == null)
```
swesonga, line 315: does the old behavior result in a yellow bar warning?

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependenciesSnapshotProvider.cs
drewnoakes: No, the old behaviour fails silently. It would never create the necessary data flow subscriptions, so the Dependencies node would appear empty indefinitely.

PR #6957, opened 2021-02-11
Added accessible name to AnalysisLevelComboBox.

melytc:
For some reason, adding it via the designer saved us other accessibility bugs around that combo box (if I added it directly into the resource file, I got this error twice in the Accessibility Insights tool: An onscreen element must not have a null BoundingRectangle property).

Accessibility pass
On whole pane:
image

On Analysis level combo box:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6957)

PR #6976, opened 2021-02-19
Add accessible name to AnalysisLevelComboBox

melytc:
This change would fix the accessibility bug "The Name property of a focusable element must not be null" in [AzDO#1247585](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1247585), and will not require a localization pass.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6976)


drewnoakes (approved):

drewnoakes:
Nice solution!

jjmew (approved):
:shipit:

dmonroym:
You should include the revert of #6957 in this PR. Alternatively we can also do what @drewnoakes had suggested regarding pushing back the dev16.9.x branch. Nothing of this has been integrated into main so either option is fine

drewnoakes:
@dmonroym I force pushed `dev16.9.x` back to 9528a30b084098b753c85532394afce4506bd5ee so that this can be merged and we produce a clean insertion PR.

It seems that the old tip of `dev16.9.x` was merged into `main` in https://github.com/dotnet/project-system/pull/6966, so we'll have to keep an eye on main for 16.10 to make sure we have the correct state there.

PR #6978, opened 2021-02-21
Replace BufferBlock with CPS's slim variety

### LINKED ISSUE #6977: Do not use BufferBlock in the import tree provider and dependencies tree
Please switch to use DataflowBlockSlim.CreateSimpleBufferBlock or prevent using it completely.

BufferBlock is more memory expensive & our dataflow tracking tool doesn't support it to detect potential problems.

"Show All Files" is used more often than we think, it is shown as top UI delay when we switch to it. (This issue is unrelated, but need find what blocks UI thread.)

drewnoakes:
Fixes #6977

CPS's implementations of Dataflow blocks are lighter weight, and better supported by internal tools for detection and diagnosis of problems.

This change will help assist a current investigation into a UI delay when enabling _Show all files_ in Solution Explorerl.

To ensure we don't accidentally introduce usages of `BufferBlock<>` again in future, this PR also bans the use of `BufferBlock` so that attempts to use it will generate an error.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6978)


drewnoakes:
Published as a draft until CPS inserts [PR 305607](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/305607) so that a new API is available for us to consume.

[This insertion PR](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/305772) is a candidate for unblocking this.

tmeschter (approved):

MiYanni (approved):

PR #6979, opened 2021-02-21
Use newer free-threaded language service API

### LINKED ISSUE #353: Remove language service "switch to UI thread" when Roslyn interaction is thread-safe
Our interaction with Roslyn is not thread-safe and is expected to occur on the ui thread. Once that is fixed, we can remove switch to UI threads.

drewnoakes:
Fixes #353

Roslyn recently implemented a free-threaded version of this API, meaning we no longer need to switch to the UI thread before calling it.

This commit bumps the package version, removes the thread switch and calls the new API.

cc @CyrusNajmabadi @jasonmalinowski

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6979)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
```diff
@@ -69,19 +67,17 @@ public async Task ReleaseProjectContextAsync(IWorkspaceProjectContextAccessor ac
...
-                IWorkspaceProjectContext context = _workspaceProjectContextFactory.Value.CreateProjectContext(
+                IWorkspaceProjectContext context = await _workspaceProjectContextFactory.Value.CreateProjectContextAsync(
                                                                                     data.LanguageName,
                                                                                     data.WorkspaceProjectContextId,
                                                                                     data.ProjectFilePath,
                                                                                     data.ProjectGuid,
                                                                                     hostObject,
-                                                                                    data.BinOutputPath);
+                                                                                    data.BinOutputPath,
+                                                                                    CancellationToken.None);
```
drewnoakes, line 80: I don't believe we have anything to pass here for cancellation.

build/import/Packages.targets
```diff
@@ -87,7 +87,6 @@
     <PackageReference Update="EnvDTE"                                                                 Version="16.8.30523.219" />
     <PackageReference Update="EnvDTE80"                                                               Version="16.8.30523.219" />
     <PackageReference Update="EnvDTE90"                                                               Version="16.8.30523.219" />
-    <PackageReference Update="StreamJsonRpc"                                                          Version="2.6.41-alpha" />
```
drewnoakes, line 90: We don't expliictly reference the `StreamJsonRpc` package anywhere in this solution, so this line had no effect.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/DebuggerTraceListener.cs
```diff
@@ -38,17 +38,17 @@ public Task InitializeAsync(IAsyncServiceProvider asyncServiceProvider)
 
         public override void Write(string message)
         {
-            if (Debugger.IsLogging())
+            if (System.Diagnostics.Debugger.IsLogging())
```
drewnoakes, line 41: A package upgrade must have introduced a `Debugger` class into a `Microsoft.*` namespace, as this broke. A using alias didn't help either, hence the fully qualified name here.

jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
CyrusNajmabadi: that's fine. we'll continue to have the parameter though in case this is something you hookup in the future.  for example, i could imagine a world where the user decides to close VS while we're still loading everything.  This would allow youto thread though cancellation if you wanted us to abort.

Our positionhere is: we will support cancellation up through the point where we're still doing computation but haven't done any mutation.  once we mutate our actual internal state, we ignore caller-requested cancellation and finish what we're doing.  That way our internal models are always in a consistent state.

CyrusNajmabadi (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/DebuggerTraceListener.cs
MiYanni: If you put `using System.Diagnostics;` inside the namespace (not above it like it is now), it would have used the right debugger class as an alternative to fully qualifying it.

PR #6986, opened 2021-02-25
Define a launch profiles editor service

tmeschter:
Define a launch profiles editor service and expose it through the `IBrokeredService`. This contains very little of the actual implementation--just the data transfer objects and interfaces needed to communicate across the brokered service boundary.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6986)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
@@ -1,0 +1,93 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.ComponentModel.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ServiceHub.Framework;
+using Microsoft.ServiceHub.Framework.Services;
+using Microsoft.VisualStudio.ProjectSystem.Debug.Designer;
+using Microsoft.VisualStudio.RpcContracts;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.ServiceBroker;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
+{
+    [Export(typeof(IPackageService))]
+    internal sealed class LaunchProfileEditorServiceFactory : IPackageService, IDisposable
+    {
+        private const string LaunchProfileEditorServiceName = "Microsoft.VisualStudio.ProjectSystem.Managed.LaunchProfileEditorService";
+        private const string ProjectGuidActivationArgumentName = "ProjectGuid";
+
+        internal static readonly ServiceJsonRpcDescriptor LaunchProfileEditorServiceDescriptorV1 = new(
+            new ServiceMoniker(LaunchProfileEditorServiceName, new Version(0, 1)),
+            ServiceJsonRpcDescriptor.Formatters.UTF8,
+            ServiceJsonRpcDescriptor.MessageDelimiters.HttpLikeHeaders);
+
+        private readonly IProjectServiceAccessor _projectServiceAccessor;
+        private IDisposable? _profferDisposable;
+        private ILaunchProfileEditorService? _launchProfileEditorService;
+
+        [ImportingConstructor]
+        public LaunchProfileEditorServiceFactory(IProjectServiceAccessor projectServiceAccessor)
+        {
+            _projectServiceAccessor = projectServiceAccessor;
+        }
+
+        public async Task InitializeAsync(IAsyncServiceProvider asyncServiceProvider)
+        {
+            IBrokeredServiceContainer brokeredServiceContainer = await asyncServiceProvider.GetServiceAsync<SVsBrokeredServiceContainer, IBrokeredServiceContainer>();
+
+            _profferDisposable = brokeredServiceContainer.Proffer(LaunchProfileEditorServiceDescriptorV1, CreateInstanceAsync);
+
+            async ValueTask<object?> CreateInstanceAsync(
+                ServiceMoniker moniker,
+                ServiceActivationOptions options,
+                IServiceBroker serviceBroker,
+                AuthorizationServiceClient authorizationServiceClient,
+                CancellationToken cancellationToken)
+            {
+                await authorizationServiceClient.AuthorizeOrThrowAsync(WellKnownProtectedOperations.CreateClientIsOwner(), cancellationToken);
+
+                if (options.ActivationArguments is null
+                    || !options.ActivationArguments.TryGetValue("ProjectGuid", out string projectGuidString))
+                {
+                    throw new InvalidOperationException($"Missing required activation argument \"{ProjectGuidActivationArgumentName}\".");
+                }
+
+                if (!Guid.TryParse(projectGuidString, out Guid projectGuid))
+                {
+                    throw new InvalidOperationException($"Unable to parse activation argument \"{ProjectGuidActivationArgumentName}\": \"{projectGuidString}\".");
+                }
+
+                if (options.ClientRpcTarget is not ILaunchProfileEditorClientSession clientSession)
+                {
+                    throw new InvalidOperationException($"Missing client RPC target \"{nameof(ILaunchProfileEditorClientSession)}\".");
+                }
+
+                if (_launchProfileEditorService is null)
+                {
+                    _launchProfileEditorService = _projectServiceAccessor
+                        .GetProjectService()
+                        .Services
+                        .ExportProvider
+                        .GetExport<ILaunchProfileEditorService>()
+                        .Value;
+                }
+
+                if (await _launchProfileEditorService.CreateConnectionAsync(projectGuid, clientSession) is not ILaunchProfileEditorConnection connection)
+                {
+                    throw new InvalidOperationException($"Unable to connect to project with GUID \"{projectGuid}\".");
+                }
+
+                return connection;
+            }
+        }
+
+        public void Dispose()
+        {
+            _profferDisposable?.Dispose();
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
@@ -60,7 +60,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
                 // When the client requests the service we expect it to pass the target project's
                 // GUID, which is how we figure out which list of launch profiles to provide.
                 if (options.ActivationArguments is null
-                    || !options.ActivationArguments.TryGetValue("ProjectGuid", out string projectGuidString))
+                    || !options.ActivationArguments.TryGetValue(ProjectGuidActivationArgumentName, out string projectGuidString))
                 {
                     throw new InvalidOperationException($"Missing required activation argument \"{ProjectGuidActivationArgumentName}\".");
                 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
@@ -14,12 +14,19 @@ using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
 {
+    /// <summary>
+    /// Reponsible for proffering the <see cref="ILaunchProfileEditorService"/> through
+    /// the <see cref="IServiceBroker"/>.
+    /// </summary>
     [Export(typeof(IPackageService))]
     internal sealed class LaunchProfileEditorServiceFactory : IPackageService, IDisposable
     {
         private const string LaunchProfileEditorServiceName = "Microsoft.VisualStudio.ProjectSystem.Managed.LaunchProfileEditorService";
         private const string ProjectGuidActivationArgumentName = "ProjectGuid";
 
+        /// <summary>
+        /// The actual service descriptor the client UI would use to request the service.
+        /// </summary>
         internal static readonly ServiceJsonRpcDescriptor LaunchProfileEditorServiceDescriptorV1 = new(
             new ServiceMoniker(LaunchProfileEditorServiceName, new Version(0, 1)),
             ServiceJsonRpcDescriptor.Formatters.UTF8,
@@ -50,6 +57,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
             {
                 await authorizationServiceClient.AuthorizeOrThrowAsync(WellKnownProtectedOperations.CreateClientIsOwner(), cancellationToken);
 
+                // When the client requests the service we expect it to pass the target project's
+                // GUID, which is how we figure out which list of launch profiles to provide.
                 if (options.ActivationArguments is null
                     || !options.ActivationArguments.TryGetValue("ProjectGuid", out string projectGuidString))
                 {
@@ -61,11 +70,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
                     throw new InvalidOperationException($"Unable to parse activation argument \"{ProjectGuidActivationArgumentName}\": \"{projectGuidString}\".");
                 }
 
+                // We also expect the client to provide an ILaunchProfileEditorClientSession; this
+                // is the interface we use when we want to communicate information back to the
+                // client (e.g., the list of launch profiles, updates to properties, etc.).
                 if (options.ClientRpcTarget is not ILaunchProfileEditorClientSession clientSession)
                 {
                     throw new InvalidOperationException($"Missing client RPC target \"{nameof(ILaunchProfileEditorClientSession)}\".");
                 }
 
+                // Obtain the MEF-exported ILaunchProfileEditorService so we can delegate to that.
                 if (_launchProfileEditorService is null)
                 {
                     _launchProfileEditorService = _projectServiceAccessor
@@ -76,6 +89,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
                         .Value;
                 }
 
+                // Create the new ILaunchProfileEditorConnection that we pass back to the client,
+                // so the client can add/remove/update the launch profiles.
                 if (await _launchProfileEditorService.CreateConnectionAsync(projectGuid, clientSession) is not ILaunchProfileEditorConnection connection)
                 {
                     throw new InvalidOperationException($"Unable to connect to project with GUID \"{projectGuid}\".");
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
@@ -1,7 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
-using System.Diagnostics.CodeAnalysis;
 using Microsoft.VisualStudio.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.ProjectSystem.VS;
@@ -28,8 +27,7 @@ namespace Microsoft.VisualStudio.Shell.Interop
         /// <summary>
         ///     Gets the value of the specified property if the hierarchy supports it, or throws an exception if there was an error.
         /// </summary>
-        [return: MaybeNull]
-        public static T GetProperty<T>(this IVsHierarchy hierarchy, VsHierarchyPropID property, T defaultValue = default)
+        public static T? GetProperty<T>(this IVsHierarchy hierarchy, VsHierarchyPropID property, T? defaultValue = default)
         {
             return GetProperty(hierarchy, HierarchyId.Root, property, defaultValue);
         }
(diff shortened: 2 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.VisualStudio.ProjectSystem.Debug.Designer;
+using Microsoft.VisualStudio.RpcContracts;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.ServiceBroker;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug.Designer
+{
+    [Export(typeof(IPackageService))]
+    internal sealed class LaunchProfileEditorServiceFactory : IPackageService, IDisposable
```
tmeschter, line 18: This type is responsible for proffering the service through the `IServiceBroker`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        private const string LaunchProfileEditorServiceName = "Microsoft.VisualStudio.ProjectSystem.Managed.LaunchProfileEditorService";
+        private const string ProjectGuidActivationArgumentName = "ProjectGuid";
+
+        internal static readonly ServiceJsonRpcDescriptor LaunchProfileEditorServiceDescriptorV1 = new(
```
tmeschter, line 23: This is the actual service descriptor the client UI would use to request the service.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+                ServiceMoniker moniker,
+                ServiceActivationOptions options,
+                IServiceBroker serviceBroker,
+                AuthorizationServiceClient authorizationServiceClient,
+                CancellationToken cancellationToken)
+            {
+                await authorizationServiceClient.AuthorizeOrThrowAsync(WellKnownProtectedOperations.CreateClientIsOwner(), cancellationToken);
+
+                if (options.ActivationArguments is null
+                    || !options.ActivationArguments.TryGetValue("ProjectGuid", out string projectGuidString))
```
tmeschter, line 54: When the client requests the service we expect it to pass the target project's GUID. This is how we figure out which list of launch profiles to provide.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    throw new InvalidOperationException($"Missing required activation argument \"{ProjectGuidActivationArgumentName}\".");
+                }
+
+                if (!Guid.TryParse(projectGuidString, out Guid projectGuid))
+                {
+                    throw new InvalidOperationException($"Unable to parse activation argument \"{ProjectGuidActivationArgumentName}\": \"{projectGuidString}\".");
+                }
+
+                if (options.ClientRpcTarget is not ILaunchProfileEditorClientSession clientSession)
```
tmeschter, line 64: We also expect the client to provide an `ILaunchProfileEditorClientSession`; this is the interface we use when we want to communicate information back to the client (e.g., the list of launch profiles, updates to properties, etc.).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    throw new InvalidOperationException($"Missing client RPC target \"{nameof(ILaunchProfileEditorClientSession)}\".");
+                }
+
+                if (_launchProfileEditorService is null)
```
tmeschter, line 69: Here we obtain/create the MEF-exported `ILaunchProfileEditorService` so we can delegate to that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/Designer/LaunchProfileEditorServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    _launchProfileEditorService = _projectServiceAccessor
+                        .GetProjectService()
+                        .Services
+                        .ExportProvider
+                        .GetExport<ILaunchProfileEditorService>()
+                        .Value;
+                }
+
+                if (await _launchProfileEditorService.CreateConnectionAsync(projectGuid, clientSession) is not ILaunchProfileEditorConnection connection)
```
tmeschter, line 79: Here we create the `ILaunchProfileEditorConnection` that we pass back to the client, so the client can add/remove/update the launch profiles.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
```diff
@@ -0,0 +1,44 @@
...
+        PropertyEditor? Editor,
+        string? SearchTerms,
+        string? HelpUrl,
+        string? VisibilityCondition);
+
+    internal sealed record SupportedValue(
+        string DisplayName,
+        object Value);
+
+    internal sealed record Property(
```
tmeschter, line 35: Note that these DTOs are property-centric. Instead of a multi-level hierarchy where we start with a list of profiles and from there can dig down to the properties on the individual page, we have a flat list of properties with enough metadata to determine which pages they belong to.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorClientSession.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Notifies an editor client of a change to the values of one or more properties.
+        /// </summary>
+        /// <returns>
+        /// A Task that completes when the client has received and scheduled the property update.
+        /// </returns>
+        Task ReceivePropertyUpdatesAsync(
+            Guid? correlationId,
+            ImmutableArray<PropertyUpdate> propertyUpdates);
```
tmeschter, line 28: Now that I've posted this, I think this is insufficient as an update mechanism. If the user adds a new profile, then we need to communicate a whole set of additional properties, metadata _and_ values, back to the client.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// Must be disposed when the connection is no longer required.
+    /// </remarks>
+    internal interface ILaunchProfileEditorConnection : IAsyncDisposable
+    {
+        Task SetValuesAsync(Guid correlationId, ImmutableArray<PropertyValueChangeRequest> changes);
+
+        /// <summary>
+        /// Returns the set of available debug command names.
+        /// </summary>
+        Task<ImmutableArray<string>> GetDebugCommandNamesAsync();
```
tmeschter, line 23: The assumption here is that once the debug command name is chosen, the user cannot change it. This is because the set of properties shown to the user depends on the debug command name.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Immutable;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug.Designer
+{
+    internal sealed record PropertyValueChangeRequest(
+        string ProfileName,
+        string PropertyName,
+        object Value);
+
+    internal sealed record Profile(string Name, string DisplayName);
```
tmeschter, line 12: Should the `Profile` include the debug command name? In the current design it is not an editable property, but it would be useful to display to the user.

tmeschter:
@drewnoakes I'm not necessarily going to put this into dotnet/project-system, but I would appreciate your feedback on the shape of the interfaces and data transfer objects.

In particular, I'd like feedback on the update mechanism. Currently the server reports only property value updates back to the client, but to support adding, removing, and renaming profiles we also need to report property additions (including metadata) and deletions.

Alternatively, the server could provide a full snapshot after each update and leave it to the client side to perform the diff.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    internal sealed record Category(string Name, string DisplayName);
+
+    internal sealed record PropertyEditor(string Name, ImmutableDictionary<string, string> Metadata);
+
+    internal sealed record PropertyMetadata(
+        string Name,
+        string DisplayName,
+        Profile Profile,
```
drewnoakes, line 21: Please don't put the `Profile` on the property's metadata. We can identify the property's scope elsewhere, just like we identify the project to which a project property exists. I'd like to have as much similarity between the property type used for launch profiles and that used for project properties as possible, to maximise code re-use. In the project properties we have page/category, which are fine for launch profiles too. I don't think we'd show navigation for launch profiles (though we could have category headers).

As mentioned elsewhere, the client will treat each launch profile as its own entity, sealed off from any other. Ideally the client receives the initial state and any updates for each profile individually.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
drewnoakes: I've commented elsewhere that I'm not sure about this approach. Conceptually, each profile is an isolated set of properties, unlike in the case of project properties where there are relationships between each. I describe elsewhere how some changing this can make things like notification of new profiles more straightforward.

drewnoakes (commented):
Looks like a good start. I've left some feedback. The main concern I have is around being property centric, rather than profile centric. I've gone into detail in targeted comments.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorConnection.cs
tmeschter: In the current implementation, the "command name" uniquely identifies who gets to convert the launch settings into actual debug settings, and the "owner" of the command gets to decide what properties to expose in the UI. This is a MEF-extensible system; we (.NET Project System) provide handlers for "Project" and "Executable", the Xamarin team provides a handler for "Xamarin", and the Web Tools team provides handlers for "IIS" and "IISExpress", for example. So we already have the possibility for a collision though in practice it is unlikely.

I'm not sure why `ILaunchProfile.CommandName` is nullable. It may be to allow us to handle the case where the command name is not present in the launchSettings.json file for whatever reason. Regardless, at this layer we can count on the command name being non-null--the implementation can always provide "Project" as a reasonable default. 

All that being said, I'll update the API to use `ProfileType` as you've laid it out here as it does make sense to provide help, a display name, and an icon.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
tmeschter: The command name is not localizable, but we could associate it with a localizable string.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Adds a profile with the given name and debug command name.
+        /// </summary>
+        Task AddAsync(Guid correlationId, string profileName, string debugCommandName);
+
+        /// <summary>
+        /// Removes the profile with the given name.
+        /// </summary>
+        Task RemoveAsync(Guid correlationId, string profileName);
```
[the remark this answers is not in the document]
tmeschter: At this layer it can be assumed to be unique. Nothing stops you from editing the launchSettings.json by hand to create more than one profile with the same name, but even in the existing system we will effectively only see the first profile with a given name.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
tmeschter: That's fine. I can switch this so a profile consists of profile metadata (the `ProfileType` you laid out elsewhere) and a set of properties.

I don't think we need the concept of a page separate from the profile--any given profile will only have one "page" of properties (possibly organized into categories, of course).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
tmeschter: I've now updated this to be profile-centric rather than property-centric. The client now receives profiles one at time and each profile contains its properties, rather than the client receiving all properties at once and using property metadata to divide them into pages.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorClientSession.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug.Designer
+{
+    /// <summary>
+    /// Proffered by the client, for notifications from the server.
+    /// </summary>
+    internal interface ILaunchProfileEditorClientSession
```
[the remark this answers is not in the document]
tmeschter: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorClientSession.cs
tmeschter: This is largely the approach I ended up with.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
tmeschter: OK, the `ProfileType` now associates the command name with a localizable string.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
```diff
@@ -28,39 +27,33 @@ public static Guid GetGuidProperty(this IVsHierarchy hierarchy, VsHierarchyPropI
         /// <summary>
         ///     Gets the value of the specified property if the hierarchy supports it, or throws an exception if there was an error.
         /// </summary>
-        [return: MaybeNull]
-        public static T GetProperty<T>(this IVsHierarchy hierarchy, VsHierarchyPropID property, T defaultValue = default)
+        public static T? GetProperty<T>(this IVsHierarchy hierarchy, VsHierarchyPropID property, T? defaultValue = default)
```
drewnoakes, line 30: Is the return only null if `defaultValue` is null? i.e. if I pass a non-null default, can I still get a null. If not, this can be annotated with `[return: NotNullIfNotNull("defaultValue")]`.

Same comment regarding the three changes further down in this file (for `out` parameters in two cases).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <param name="Name">
+    /// The unique name of the profile.
+    /// </param>
+    /// <param name="CommandName">
+    /// The debugger command associated with this profile. Corresponds to ProfileType.CommandName.
+    /// </param>
+    /// <param name="Properties">
+    /// The properties and values for this launch profile.
+    /// </param>
+    internal sealed record Profile(
```
drewnoakes, line 39: Perhaps rename to `LaunchProfile` so the name is more specific. I don't feel strongly about this. If it were to change, should also rename `ProfileType` to `LaunchProfileType`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorClientSession.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </para>
+        /// </summary>
+        /// <param name="correlationId">
+        /// The correlation ID provided by the client, if this profile is being added as a result of calling <see cref="ILaunchProfileEditorConnection.AddAsync(Guid, string, string)"/>.
+        /// </param>
+        /// <param name="profile">The newly-added <see cref="Profile"/> along with all metadata and properties.</param>
+        /// <returns>
+        /// A Task that completes when the client has received and scheduled the profile addition.
+        /// </returns>
+        Task LaunchProfileAddedAsync(Guid? correlationId, Profile profile);
```
drewnoakes, line 29: One thought here is that the UI won't know when it has received all of the initial profiles. There are two minor downsides to this:

1. We won't be able to show a "loading" message as it is not possible to differentiate between "no profiles exist" and "we haven't received the first yet" states.
2. If they trickle in while the UI is visible, each arrival will trigger UI updates. We expect only a few profiles (1-5) so this isn't likely to be a huge issue in practice.

A simple change could be to pass `ImmutableArray<Profile>` here, with the assumption that the first call on the client includes the full snapshot. Future calls would likely only include a single item in that array.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorClientSession.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <param name="oldProfileName">
+        /// The old profile name.
+        /// </param>
+        /// <param name="newProfileName">
+        /// The new profile name.
+        /// </param>
+        /// <returns>
+        /// A Task that completes when the client has received and scheduled the profile rename.
+        /// </returns>
+        Task LaunchProfileRenamedAsync(Guid? correlationId, string oldProfileName, string newProfileName);
```
drewnoakes, line 60: Given we are using the profile name as an identifier, and we are supporting renaming that identifier, we need strict guarantees about overlapping requests via this interface, and careful handling on both sides of the contract here. For example, the client may attempt to remove a profile using name A, which the server is attempting to rename from A to B. This is unlikely, but I'd prefer a design that didn't allow such a situation in the first place. Using a GUID for each profile, that's never shown to the user and potentially differs across editor sessions would solve this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/DataTransferTypes.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// assumes that the debugger command name associated with a Profile never changes.
+    /// The set of properties associated with a profile depends on the debugger command;
+    /// if the command can change than the set of properties on a profile can change,
+    /// not just their values as in the current design. If we adopt an alternative
+    /// design that allows changing the command we will also need to update ILaunchProfileClientEditorSession"
+    /// to support adding/removing properties.
+    /// </remarks>
+    /// <param name="Name">
+    /// The unique name of the profile.
+    /// </param>
```
drewnoakes, line 32: Nice! I was wondering how to document record parameters. Seems obvious in retrospect 😃

tmeschter: This is why I had to update the compiler package to a 3.9.0 version--this isn't supported in 3.8.0.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
tmeschter: No, it can also return `null` if the underlying call to `IVsHierarchy.GetProperty` returns an HRESULT indicating a failure.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/Designer/ILaunchProfileEditorClientSession.cs
tmeschter: The server is going to serialize all of these operations, so either A will be removed and the rename will turn into a no-op, or the rename will succeed and the removal will turn into a no-op. I don't see either outcome as being problematic.

drewnoakes: The problem is that the client is not always in sync with the server, and must relate the identity to the server. I think there's still a race here. Renaming identifiers seems like something to avoid. Is there not another identifier we could use? Even just an opaque numeric one?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
```diff
@@ -28,39 +27,33 @@ public static Guid GetGuidProperty(this IVsHierarchy hierarchy, VsHierarchyPropI
...
-#pragma warning disable CS8717 // Needs https://github.com/dotnet/roslyn/issues/38638
             return GetProperty(hierarchy, HierarchyId.Root, property, defaultValue, out result);
-#pragma warning restore CS8717 
         }
 
         /// <summary>
         ///     Gets the value of the specified property of the specified item if the hierarchy supports it, or returns a HRESULT if there was an error.
         /// </summary>
-        public static int GetProperty<T>(this IVsHierarchy hierarchy, HierarchyId item, VsHierarchyPropID property, T defaultValue, [MaybeNull] out T result)
+        public static int GetProperty<T>(this IVsHierarchy hierarchy, HierarchyId item, VsHierarchyPropID property, T? defaultValue, out T? result)
```
drewnoakes, line 56: Null forgiveness operator below can be removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/Interop/VsHierarchyExtensions.cs
drewnoakes: That doesn't appear to be true of the call below this one, where `Verify.HResult` presumably throws. Is this call not just delegating to that one? If so, perhaps the annotation can be improved as suggested.

Can `IVsHierarchy.GetProperty` produce a null `resultObject`? If so, then yeah there's nothing we can do to improve this.

tmeschter:
We're going to handle launch profiles through the Project Query API, so this change is no longer needed.

PR #6991, opened 2021-02-26
Fix two usability issues with the package restore telemetry:

swesonga:
1. The implementation assumes that the package restore being up to date is sufficient to determine whether operations completed successfully. However, the more important data point is whether or not the restore failed or succeeded. This PR updates IConfiguredProjectPackageRestoreTelemetryService to reflect this and its implementation to report this success/failure status when restore is up to date.

2. The PackageRestoreDataSource posts an event with the PackageRestoreOperationNames.EndNominateRestore operation regardless of whether the nomination of a project completes successfully. However, the actual success/failure status is required to determine whether the operation ran into any issues. This PR adds the status flag to the nomination telemetry event.


tmeschter:
I believe there are situations where we need to run restore multiple times on a single project before we get to some steady state. For example, if a NuGet package adds a .props file that itself includes a `<PackageReference>` item.

In that case, if the first restore succeeds but the second one fails, will we record the overall restore operation as a success or failure?

davkean:
Why do we own the telemetry for restore success? I was pretty sure that NuGet already sent that.

swesonga:
> Why do we own the telemetry for restore success? I was pretty sure that NuGet already sent that with way more information than we're providing.

Good point @davkean. This was an oversight on my part. I'm now reviewing the NuGet telemetry to determine how to proceed.

PR #7006, opened 2021-03-08
Update Project Query API data providers to understand the binding context

tmeschter:
When retrieving properties through the CPS properties system three pieces of data are needed:

- A configured project (represented by a `ConfiguredProject`) which represents the "world" in which the property retrieval occurs
- A `Rule` describing _which_ properties to retrieve, including metadata about how to retrieve them and how the consumer should interpret them
- A binding context that both identifies the entity that you want to investigate, and the context in which to investigate it. This is represented by an `IProjectPropertiesContext` (which is a somewhat misleading name for reasons I will get to in a moment).

The binding context requires further explanation. Some examples may help. An `IProjectPropertiesContext` may refer to:

- The project as a whole, in which case the properties retrieved are _project properties_.
- An imported .props or .targets file, in which case the properties are _project properties as defined in that import_.
- An item of a particular item type, in which case the properties are the _item properties for that specific definition of the item_.
- An item without a particular item type, in which case the properties are the _item properties for **some** definition of that item_.
- An item type, in which case the properties are the _default properties supplied to all that items of that type_.
- Various other combinations.

If you didn't follow all of that, that's OK. The important point is that the `Rule`s describing a set a properties can be largely independent of the source of those properties. The same `Rule` could be used, for example, to query/update the properties on a specific `<Reference>` item _or_ the default properties of **all** `<Reference>` items and the consuming code doesn't need to be concerned about which it is--the properties system will do the right thing depending on the context.

Of course, not all combinations of `Rule`s and `IProjectPropertiesContext`s are meaningful. There isn't much point in binding the `Rule` for the "Application" property page to a specific item, and it is ultimately up to the `IProjectPropertiesProvider` (representing the underlying storage for properties) to determine what a particular `IProjectPropertiesContext` represents.

With that background out of the way, we can move on to the specifics of this change. Currently, when the providers that implement the Project Query API for the new project property pages bind the `Rule`s representing the individual pages they implicitly use the "project file" context. Outside of that implicit context, however, nothing about the providers is specific to _project_ properties--all the logic could apply equally well to item properties.

Now we're planning to expose individual launch profiles through their own UI. By creating an `IProjectPropertiesProvider` that can read and write launch profile properties from the launchSettings.json file (in a separate PR) and replacing the implicit context with an explicit context (done here) we can re-use a great deal of code we've already created for the project properties pages:

- We can use mostly the same Project Query API model types for both project property pages and launch profiles...
- ...which means the UI layer can reuse many of its controls with minimal changes
- On the back end, the same Project Query API data providers can handle both project property pages and launch profiles
- Developers can use the same .xaml format for `Rule` files to describe both the property pages and the properties to expose for a launch profile

The bulk of this change is to define an implementation of `IProjectPropertiesContext` we can use while processing a query, and pass instances of that context between the different providers.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7006)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
@@ -17,7 +17,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
     /// </summary>
     internal static class PropertyPageDataProducer
     {
-        public static IEntityValue CreatePropertyPageValue(IEntityValue parent, IPropertyPageQueryCache cache, Rule rule, List<Rule>? debugChildRules, IPropertyPagePropertiesAvailableStatus requestedProperties)
+        public static IEntityValue CreatePropertyPageValue(IEntityValue parent, IPropertyPageQueryCache cache, QueryProjectPropertiesContext context, Rule rule, List<Rule>? debugChildRules, IPropertyPagePropertiesAvailableStatus requestedProperties)
         {
             Requires.NotNull(parent, nameof(parent));
             Requires.NotNull(rule, nameof(rule));
@@ -28,15 +28,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 
             var identity = new EntityIdentity(
                 ((IEntityWithId)parent).Id,
-                new KeyValuePair<string, string>[]
-                {
-                    new(ProjectModelIdentityKeys.PropertyPageName, name)
-                });
+                createKeys());
 
-            return CreatePropertyPageValue(parent.EntityRuntime, identity, cache, rule, debugChildRules, requestedProperties);
+            return CreatePropertyPageValue(parent.EntityRuntime, identity, cache, context, rule, debugChildRules, requestedProperties);
+
+            IEnumerable<KeyValuePair<string, string>> createKeys()
+            {
+                yield return new(ProjectModelIdentityKeys.PropertyPageName, name);
+
+                if (context.ItemType is not null)
+                {
+                    yield return new(ProjectModelIdentityKeys.SourceItemType, context.ItemType);
+                }
+
+                if (context.ItemName is not null)
+                {
+                    yield return new(ProjectModelIdentityKeys.SourceItemName, context.ItemName);
+                }
+            }
         }
 
-        public static IEntityValue CreatePropertyPageValue(IEntityRuntimeModel runtimeModel, EntityIdentity id, IPropertyPageQueryCache cache, Rule rule, List<Rule>? debugChildRules, IPropertyPagePropertiesAvailableStatus requestedProperties)
+        public static IEntityValue CreatePropertyPageValue(IEntityRuntimeModel runtimeModel, EntityIdentity id, IPropertyPageQueryCache cache, QueryProjectPropertiesContext context, Rule rule, List<Rule>? debugChildRules, IPropertyPagePropertiesAvailableStatus requestedProperties)
         {
             Requires.NotNull(rule, nameof(rule));
             var newPropertyPage = new PropertyPageValue(runtimeModel, id, new PropertyPagePropertiesAvailableStatus());
@@ -64,8 +76,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
             }
 
             ((IEntityValueFromProvider)newPropertyPage).ProviderState = debugChildRules is not null
-                ? new PropertyPageProviderState(cache, rule, debugChildRules)
-                : new PropertyPageProviderState(cache, rule);
+                ? new PropertyPageProviderState(cache, context, rule, debugChildRules)
+                : new PropertyPageProviderState(cache, context, rule);
 
             return newPropertyPage;
         }
@@ -75,13 +87,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
             EntityIdentity id,
             IProjectService2 projectService,
             IPropertyPageQueryCacheProvider queryCacheProvider,
-            string projectPath,
+            QueryProjectPropertiesContext context,
             string propertyPageName,
             IPropertyPagePropertiesAvailableStatus requestedProperties)
         {
             propertyPageName = DebugUtilities.ConvertDebugPageNameToRealPageName(propertyPageName);
 
-            if (projectService.GetLoadedProject(projectPath) is UnconfiguredProject project
+            if (projectService.GetLoadedProject(context.File) is UnconfiguredProject project
                 && await project.GetProjectLevelPropertyPagesCatalogAsync() is IPropertyPagesCatalog projectCatalog
                 && projectCatalog.GetSchema(propertyPageName) is Rule rule
                 && !rule.PropertyPagesHidden)
@@ -93,7 +105,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                 }
 
                 var propertyPageQueryCache = queryCacheProvider.CreateCache(project);
-                IEntityValue propertyPageValue = CreatePropertyPageValue(runtimeModel, id, propertyPageQueryCache, rule, debugChildRules, requestedProperties);
+                IEntityValue propertyPageValue = CreatePropertyPageValue(runtimeModel, id, propertyPageQueryCache, context, rule, debugChildRules, requestedProperties);
                 return propertyPageValue;
             }
 
@@ -116,8 +128,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
             IEnumerable<IEntityValue> createPropertyPageValuesAsync()
             {
                 Rule? parentDebuggerPageRule = null;
-                
-                var propertyPageQueryContext = queryCacheProvider.CreateCache(project);
+
+                IPropertyPageQueryCache propertyPageQueryCache = queryCacheProvider.CreateCache(project);
+                QueryProjectPropertiesContext context = new QueryProjectPropertiesContext(isProjectFile: true, project.FullPath, itemType: null, itemName: null);
                 foreach (string schemaName in projectCatalog.GetProjectLevelPropertyPagesSchemas())
                 {
                     if (projectCatalog.GetSchema(schemaName) is Rule rule
@@ -135,7 +148,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                         }
                         else
                         {
-                            IEntityValue propertyPageValue = CreatePropertyPageValue(parent, propertyPageQueryContext, rule, debugChildRules: null, requestedProperties);
+                            IEntityValue propertyPageValue = CreatePropertyPageValue(parent, propertyPageQueryCache, context, rule, debugChildRules: null, requestedProperties: requestedProperties);
                             yield return propertyPageValue;
                         }
                     }
@@ -144,7 +157,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                 if (parentDebuggerPageRule is not null)
                 {
                     List<Rule> childDebuggerPageRules = DebugUtilities.GetDebugChildRules(projectCatalog).ToList();
-                    IEntityValue propertyPageValue = CreatePropertyPageValue(parent, propertyPageQueryContext, parentDebuggerPageRule, debugChildRules: childDebuggerPageRules, requestedProperties);
+                    IEntityValue propertyPageValue = CreatePropertyPageValue(parent, propertyPageQueryCache, context, parentDebuggerPageRule, debugChildRules: childDebuggerPageRules, requestedProperties: requestedProperties);
                     yield return propertyPageValue;
                 }
             }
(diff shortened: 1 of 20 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/QueryProjectPropertiesContext.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// Project Query API implementation.
+    /// </summary>
+    /// <remarks>
+    /// The point here is to capture the context for a part of a query execution in a
+    /// manner that can be passed from one provider to another and is also suitable as a
+    /// key into a cache (such as the <see cref="IPropertyPageQueryCache"/>).
+    /// </remarks>
+    internal sealed class QueryProjectPropertiesContext : IProjectPropertiesContext, IEquatable<QueryProjectPropertiesContext>
+    {
+        public static readonly QueryProjectPropertiesContext ProjectFile = new(isProjectFile: true, file: string.Empty, itemType: null, itemName: null);
```
tmeschter, line 22: Note to self: add a doc comment.

This is meant to be a "well known" context representing the project file as a whole. Note that if an `IProjectPropertiesContext` has `IsProjectFile` property is set to `true` and the `ItemType` and `ItemName` properties are `null` then the properties system treats it as referring to the project file as a whole _regardless of the `File` property_. This lets us get away with setting it to the empty string and re-using the same instance across projects.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/QueryProjectPropertiesContext.cs
```diff
@@ -0,0 +1,86 @@
...
+            
+            if (ItemName is not null)
+            {
+                hashCode = (hashCode * -1521134295) + StringComparers.ItemNames.GetHashCode(ItemName);
+            }
+
+            return hashCode;
+        }
+
+        public static bool TryCreateFromEntityId(EntityIdentity id, [NotNullWhen(true)] out QueryProjectPropertiesContext? context)
```
tmeschter, line 72: Note to self: add a doc comment.

test.csx
```diff
@@ -0,0 +1 @@
+System.Console.WriteLine(5+5);
```
tmeschter, line 1: Whoops, this file was included by accident.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageQueryCache.cs
```diff
@@ -87,22 +87,22 @@ public PropertyPageQueryCache(UnconfiguredProject project)
...
-            if (_ruleCache.TryGetValue((projectConfiguration, schemaName), out IRule? cachedRule))
+            if (_ruleCache.TryGetValue((projectConfiguration, schemaName, context), out IRule? cachedRule))
             {
                 return cachedRule;
             }
 
             IRule? rule = null;
             if (await GetProjectLevelPropertyPagesCatalogAsync(projectConfiguration) is IPropertyPagesCatalog catalog)
             {
-                rule = catalog.BindToContext(schemaName, file: null, itemType: null, itemName: null);
```
tmeschter, line 102: This is where we used to implicitly bind to the project as a whole.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageQueryCache.cs
```diff
@@ -87,22 +87,22 @@ public PropertyPageQueryCache(UnconfiguredProject project)
...
+                rule = catalog.BindToContext(schemaName, context);
```
tmeschter, line 102: And now we bind to the explicit context.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/QueryProjectPropertiesContext.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as QueryProjectPropertiesContext);
+        }
+
+        public override int GetHashCode()
+        {
+            int hashCode = IsProjectFile.GetHashCode();
+            hashCode = (hashCode * -1521134295) + StringComparers.Paths.GetHashCode(File);
```
swesonga, line 69: >1521134295 [](start = 36, length = 10)

Should this implementation use System.HashCode.Combine?

tmeschter: That wasn't introduced until .NET Core 2.1.

swesonga: Ah, I didn't realize that!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
```diff
@@ -28,15 +28,27 @@ public static IEntityValue CreatePropertyPageValue(IEntityValue parent, IPropert
 
             var identity = new EntityIdentity(
                 ((IEntityWithId)parent).Id,
-                new KeyValuePair<string, string>[]
+                createKeys());
+
+            return CreatePropertyPageValue(parent.EntityRuntime, identity, cache, context, rule, debugChildRules, requestedProperties);
+
+            IEnumerable<KeyValuePair<string, string>> createKeys()
```
swesonga, line 35: >createKeys [](start = 54, length = 10)

What advantage does this have over the KeyValuePairs it's replacing? Is reading from the context expensive?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
```diff
@@ -116,8 +128,9 @@ public static IEntityValue CreatePropertyPageValue(IEntityRuntimeModel runtimeMo
             IEnumerable<IEntityValue> createPropertyPageValuesAsync()
             {
                 Rule? parentDebuggerPageRule = null;
-                
-                var propertyPageQueryContext = queryCacheProvider.CreateCache(project);
+
+                IPropertyPageQueryCache propertyPageQueryCache = queryCacheProvider.CreateCache(project);
+                QueryProjectPropertiesContext context = new QueryProjectPropertiesContext(isProjectFile: true, project.FullPath, itemType: null, itemName: null);
```
swesonga, line 133: >itemType: null, itemName: null [](start = 129, length = 30)

Would it make sense for  the constructor to automatically set these to null if isProjectFile is true?

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Query/QueryProjectPropertiesContextTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            Assert.True(a.Equals(b));
+            Assert.True(a.GetHashCode() == b.GetHashCode());
+        }
+
+        [Theory]
+        [MemberData(nameof(ContextsThatAreNotEqual))]
+        public void NotEqual(IProjectPropertiesContext a, IProjectPropertiesContext b)
+        {
+            Assert.False(a.Equals(b));
+            Assert.False(a.GetHashCode() == b.GetHashCode());
```
swesonga, line 24: >a.GetHashCode() == b.GetHashCode() [](start = 25, length = 34)

Hash codes of unequal objects can theoretically collide. Is this always guaranteed to not be the case for IProjectPropertiesContexts?

swesonga: oh, I see you're testing only the ones below so that might not be a problem here.

---
In reply to: [593537232](https://github.com/dotnet/project-system/pull/7006#discussion_r593537232) [](ancestors = 593537232)

swesonga (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
drewnoakes: When there are expected (and especially required) relationships between constructor arguments, it can be clearer and safer to create factory methods.

drewnoakes: @tmeschter your comment on `QueryProjectPropertiesContext.ProjectFile` makes it sound like that instance is useful generally, even without the specific project file path. Why not use that instance here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/QueryProjectPropertiesContext.cs
drewnoakes, line 69: Why this particular constant? I usually see these kinds of factors as prime numbers, and this isn't prime.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Query/QueryProjectPropertiesContextTests.cs
```diff
@@ -0,0 +1,50 @@
...
+                new QueryProjectPropertiesContext[] { new(true, @"C:\alpha\beta", "myItemType", null), new(true, @"C:\alpha\beta", "MyItemType", null) },
+                new QueryProjectPropertiesContext[] { new(true, @"C:\alpha\beta", null, "MyItemName"), new(true, @"C:\alpha\beta", null, "MYITEMNAME") }
+            };
+        }
+
+        public static IEnumerable<object[]> ContextsThatAreNotEqual()
+        {
+            return new QueryProjectPropertiesContext[][]
+            {
+                new QueryProjectPropertiesContext[] { new QueryProjectPropertiesContext(false, string.Empty, null, null), QueryProjectPropertiesContext.ProjectFile },
```
drewnoakes, line 43:
```suggestion
                new QueryProjectPropertiesContext[] { new(false, string.Empty, null, null), QueryProjectPropertiesContext.ProjectFile },
```

To match the other items in the array.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Query/QueryProjectPropertiesContextTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
+{
+    public class QueryProjectPropertiesContextTests
+    {
+        [Theory]
+        [MemberData(nameof(ContextsThatAreEqual))]
+        public void Equal(IProjectPropertiesContext a, IProjectPropertiesContext b)
+        {
+            Assert.True(a.Equals(b));
```
drewnoakes, line 15: `Equals` overrides are supposed to be symmetric (i.e. `A==B implies B==A`) but there's no guarantee of this. I like to validate this both ways. I'm not sure if MSTest's `Assert.AreEqual` does this for you.

```suggestion
            Assert.True(a.Equals(b));
            Assert.True(b.Equals(a));
```

Ditto below in `NotEqual`.

drewnoakes (approved):
Looks good. Only a few nits.

Thank you for the detailed PR description. It made reviewing this much easier, and I understand the back end implementation better now than before.

Learning more about how these APIs work will help me complete an unrelated task involving binding browse objects for target nodes in the dependencies tree.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
tmeschter: No, the point here is that `context.ItemType` and `context.ItemName` may be null; if they are we don't want to add them to the key that identifies this entity in the Project Query API.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
tmeschter: At this point I don't remember why I didn't use the singleton. I'll have to go back and puzzle it out.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/QueryProjectPropertiesContext.cs
tmeschter: I lifted this from somewhere else in the code base. Maybe we need an internal version of System.HashCode.Combine.

PR #7017, opened 2021-03-11
Use IProjectDependentFileChangeNotificationService

### LINKED ISSUE #7008: Stop sending ProjectChangeFileSystemEntityHint.UpdateProjectDependentFile, but switch to IProjectDependentFileChangeNotificationService
It turns out using UpdateProjectDependentFile within a write lock leads lock contentions & also blocked on UI thread during the solution loading time. It leads slow performance to get project to the ready state.

A new service has been added to the new CPS in 16.10, and we need update the NET Core code to use the new service.

+ @drewnoakes

drewnoakes:
Fixes #7008

Previously we used `ProjectChangeFileSystemEntityHint.UpdateProjectDependentFile` to notify CPS that the assets file might have changed so that it can be sure to process it (in case its own file monitoring fails). This API required a write lock on the project, which introduced significant lock contention and blocking of the UI thread during solution load.

CPS has introduced a new API for this notification in `IProjectDependentFileChangeNotificationService`. This change removes usage of the old API in favour of this new one.

@lifengl @panopticoncentral 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7017)


swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -148,30 +145,23 @@ private async Task<bool> RestoreCoreAsync(ProjectRestoreInfo restoreInfo)
...
-            await HintProjectDependentFileAsync(restoreInfo);
+            HintProjectDependentFile(restoreInfo);
 
             return success;
         }
 
-        private Task HintProjectDependentFileAsync(ProjectRestoreInfo restoreInfo)
+        private void HintProjectDependentFile(ProjectRestoreInfo restoreInfo)
         {
-            // Hint to CPS that the assets file "might" have changed and therefore
```
lifengl, line 158: That block of comments still makes sense after this change.

lifengl (approved):
changes look good to me.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
drewnoakes: I just moved it lower down.

Issue #6896, opened 2021-02-02
Dedicated UI for launch profiles (7)


drewnoakes:
(moved from #6627)

This issue is tracking work related to pulling the launch profile experience out of the project properties UI and into it's own dedicated window and experience.

Because there may be many launch profiles, they are not a good fit for the model we are using for properties in the new UI.

# UI

- [x] Use server endpoints for create/delete/duplicate profile <kbd>3</kbd> ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/329209))
- [x] Request new images for launch profile kinds if necessary <kbd>0.1</kbd>
- [x] Research requirements around dialog windows in VS and integrate <kbd>0.5</kbd>
- [x] Identify and intercept existing commands to launching debug settings <kbd>1</kbd> ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/316814))
- [x] Use VS's UI controls where necessary <kbd>0.5</kbd>
- [x] Use VS's image catalog instead of bundling images <kbd>0.5</kbd>

# Back end

- [x] Support push notifications <kbd>3</kbd> (#7247)
- [x] Implement a Nexus service for adding/removing/copying a launch profile
- [x] Improve profile name generation when duplicating <kbd>1</kbd> ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/325522))
- [x] Implement Project Query API changes needed to support debug settings UI
- [x] Update property pages to make use of VisibilityConditions

drewnoakes:
Will be closed by CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/329209

PR #7020, opened 2021-03-12
Implement basic launch profile support in the Project Query API

tmeschter:
Now that the Project Query API has been updated to represent launch profiles, this change implements basic support for querying the launch profiles. This involves the following pieces:

1. Implementing Project Query API data providers that allow us to map from:
    1. A project to a set of "launch profile types"--that is, the different kinds of launch profiles (e.g., Project, Executable, IIS, IISExpress, etc.).
    2. A project to the actual set of launch profiles--that is, the individual profiles in the launchSettings.json file as exposed through the `ILaunchSettingsProvider` interface.
    3. A launch profile to a set of properties
    4. A launch profile to a set of categories (for organizing the properties)
2. Implementing an `IProjectPropertiesProvider` that can expose the launch settings to the properties system.
3. Removing the temporary support we had for exposing the _active_ launch profile as the Debug page of the new project property pages.

### Launch Profile Types

The most important property of a profile is the _command name_. This controls which component gets to take the active launch profile and turn it into debugger settings. As such, the command name effectively determines which properties in the profile are relevant to debugging, and which ones should be surfaced to the user in the UI and how.

For the Launch Profiles UI, each command name will need to have a corresponding `Rule` defined in a .xaml file. This specifies which properties to expose for a profile utilizing that command, just as we use `Rule`s to define the properties that appear on the project property pages. Metadata on the `Rule` specifies the command name, an `ImageMoniker` to use as an icon, and a URL for further help. Producing the launch profile types is a matter of enumerating these `Rules` and returning objects with this metadata.

### Launch Profiles

Producing Project Query API entities is largely a matter of looping over the set of launch profiles. For each one we verify that we can find the corresponding `Rule` (as described above) and if so we create a `QueryProjectPropertiesContext` representing that launch profile within the properties system. If we can't find the corresponding `Rule` we filter out that profile, as we won't be able to show any UI anyway.

### Properties and Categories

At this point we've reduced the profile to a combination of a `Rule` and a `QueryProjectPropertiesContext`. These are the same concepts used to produce categories and properties for the project property pages, so we can reuse those providers almost as-is. We do need a couple of extra types just to represent the launch-profile-to-category and launch-profile-to-property relationship.

### IProjectPropertiesProvider

The role of an `IProjectPropertiesProvider` is to take a context (such as a `QueryProjectPropertiesContext`) and provide read and write access to that context's properties in a standardized fashion that abstracts away the underlying storage mechanism. Here we implement a provider that wraps the `ILaunchSettingsProvider`, effectively exposing the data on launch profiles to the larger properties system.

When we want to create the properties to return through the Project Query API, the properties system will take the `Rule` we've provided and read the data in its `DataSource.Persistence` property to determine which `IProjectPropertiesProvider` should handle the property access. The _set_ of properties to produce comes from the `Rule`, while the `IProjectPropertiesProvider` is given the context (the `QueryProjectPropertiesContext`) and asked to produce the property _values_ for that context.

### Removing the Debug page

This is essentially a matter of backing out the changes in #6770, which was always meant to be temporary. The code involved mangled the names of properties and categories associated with debugging; this was a useful and necessary hack at the time but is no longer necessary or desirable. This *does* mean that until the new Launch Profiles UI is up and running the only way to edit the debug settings will be through the old property pages, or by editing the launchSettings.json directly.

### Limitations

The implementation here only supports reading properties from launch profiles, not setting them. Also, only the properties directly exposed by `ILaunchProfile` (`ExecutablePath`, `CommandLineArgs`, `WorkingDirectory`, etc.) are exposed. Other properties ultimately get stored in either `ILaunchSettings.GlobalSettings` or `ILaunchProfile.OtherSettings`; defining, reading, and writing these properties will need an extension mechanism of some sort.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7020)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
@@ -1,0 +1,119 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Framework.XamlTypes;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.Query;
+using Microsoft.VisualStudio.ProjectSystem.Query.Frameworks;
+using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModel;
+using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModel.Implementation;
+using Microsoft.VisualStudio.ProjectSystem.VS.Utilities;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Query.PropertyPages
+{
+    internal class LaunchProfileFromProjectDataProducer : QueryDataFromProviderStateProducerBase<UnconfiguredProject>
+    {
+        private readonly ILaunchProfilePropertiesAvailableStatus _properties;
+        private readonly IPropertyPageQueryCacheProvider _queryCacheProvider;
+
+        public LaunchProfileFromProjectDataProducer(ILaunchProfilePropertiesAvailableStatus properties, IPropertyPageQueryCacheProvider queryCacheProvider)
+        {
+            _properties = properties;
+            _queryCacheProvider = queryCacheProvider;
+        }
+
+        protected override Task<IEnumerable<IEntityValue>> CreateValuesAsync(IEntityValue parent, UnconfiguredProject providerState)
+        {
+            return CreateLaunchProfileValuesAsync(parent, providerState);
+        }
+
+        private async Task<IEnumerable<IEntityValue>> CreateLaunchProfileValuesAsync(IEntityValue parent, UnconfiguredProject project)
+        {
+            if (project.Services.ExportProvider.GetExportedValueOrDefault<ILaunchSettingsProvider>() is ILaunchSettingsProvider launchSettingsProvider
+                && await project.GetProjectLevelPropertyPagesCatalogAsync() is IPropertyPagesCatalog projectCatalog
+                && await launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite) is ILaunchSettings launchSettings)
+            {
+                return createLaunchProfileValues();
+            }
+
+            return Enumerable.Empty<IEntityValue>();
+
+            IEnumerable<IEntityValue> createLaunchProfileValues()
+            {
+                IPropertyPageQueryCache propertyPageQueryCache = _queryCacheProvider.CreateCache(project);
+
+                Dictionary<string, Rule> debugRules = new();
+                foreach (Rule rule in DebugUtilities.GetDebugChildRules(projectCatalog))
+                {
+                    if (rule.Metadata.TryGetValue("CommandName", out object? commandNameObj)
+                        && commandNameObj is string commandName)
+                    {
+                        debugRules[commandName] = rule;
+                    }
+                }
+
+                foreach ((int index, ProjectSystem.Debug.ILaunchProfile profile) in launchSettings.Profiles.WithIndices())
+                {
+                    if (!Strings.IsNullOrEmpty(profile.Name)
+                        && !Strings.IsNullOrEmpty(profile.CommandName)
+                        && debugRules.TryGetValue(profile.CommandName, out Rule rule))
+                    {
+                        QueryProjectPropertiesContext context = new(
+                            isProjectFile: true,
+                            file: project.FullPath,
+                            itemType: LaunchProfilesProjectItemProvider.ItemType,
+                            itemName: profile.Name);
+
+                        IEntityValue launchProfileValue = CreateLaunchProfileValue(parent, context, rule, index, propertyPageQueryCache);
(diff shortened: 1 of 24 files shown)


tmeschter:
This includes the commits in #7006; I recommend reviewing that first.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (project.Services.ExportProvider.GetExportedValueOrDefault<ILaunchSettingsProvider>() is ILaunchSettingsProvider launchSettingsProvider
+                && await project.GetProjectLevelPropertyPagesCatalogAsync() is IPropertyPagesCatalog projectCatalog
+                && await launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite) is ILaunchSettings launchSettings)
+            {
+                return createLaunchProfileValues();
+            }
+
+            return Enumerable.Empty<IEntityValue>();
+
+            IEnumerable<IEntityValue> createLaunchProfileValues()
```
MiYanni, line 45: Instead of creating this local function, it might make sense to have the method return `IAsyncEnumerable` instead, so you can still `yield return` the results you want. Info here: https://agirlamonggeeks.com/2018/11/25/c-8-features-part-2-async-method-with-yield-return/

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileTypeDataProvider.cs
```diff
@@ -0,0 +1,141 @@
...
+        private async Task<IEnumerable<IEntityValue>> CreateLaunchProfileTypeValuesAsync(IEntityValue parent, UnconfiguredProject project)
+        {
+            if (await project.GetProjectLevelPropertyPagesCatalogAsync() is IPropertyPagesCatalog projectCatalog)
+            {
+                return createLaunchProfileTypeValues();
+            }
+
+            return Enumerable.Empty<IEntityValue>();
+
+            IEnumerable<IEntityValue> createLaunchProfileTypeValues()
```
MiYanni, line 70: Same as above on `IAsyncEnumerable`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
```diff
@@ -0,0 +1,119 @@
...
+                });
+
+            return CreateLaunchProfileValue(parent.EntityRuntime, identity, context, rule, order, propertyPageQueryCache);
+        }
+
+        private IEntityValue CreateLaunchProfileValue(IEntityRuntimeModel runtimeModel, EntityIdentity id, QueryProjectPropertiesContext context, Rule rule, int order, IPropertyPageQueryCache cache)
+        {
+            LaunchProfileValue newLaunchProfile = new(runtimeModel, id, new LaunchProfilePropertiesAvailableStatus());
+
+            if (_properties.Name)
```
MiYanni, line 95: What are these predicates? Is it some kind of new implicit null check I'm not aware of? I see them here and more in other places where the if-expression is simply a non-boolean member.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileProjectPropertiesProvider.cs
```diff
@@ -0,0 +1,269 @@
...
+            }
+
+            public Task DeletePropertyAsync(string propertyName, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+            {
+                throw new NotImplementedException();
+            }
+
+            public Task<IEnumerable<string>> GetDirectPropertyNamesAsync()
+            {
+                return GetPropertyNamesAsync();
```
MiYanni, line 223: Random question. I notice you use the arrow shorthand for get-only properties (using `=>` such as `public string FileFullPath => _context.File;`), but you don't use the arrow shorthand for single-line methods. Any reason for that? For example, this would instead be:
```C#
public Task<IEnumerable<string>> GetDirectPropertyNamesAsync() => GetPropertyNamesAsync();
```

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileProjectPropertiesProvider.cs
tmeschter: No reason other than personal inconsistency. :-D

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
tmeschter: The `ILaunchProfilePropertiesAvailableStatus` type (and more generally all `*AvailableStatus` types) are used by the Projcet Query API to indicate which parts of the model should be populated. `_properties.Name` is in fact a boolean; `LaunchProfileValue` and `ILaunchProfilePropertiesAvailableStatus` have the same member names, but all of those members are booleans on the latter.

tmeschter:
I'm merging this as it is blocking other back end work.

tmeschter:
Related to #6896.

PR #7027, opened 2021-03-16
New annotation attributes

drewnoakes:
Adds newer nullability attributes:

- `MemberNotNullAttribute`
- `MemberNotNullWhenAttribute`

Uses them in one class to avoid null forgiveness operation (`!`) via better annotations.

There will be more places we can use these new attributes to tidy and tighten logic.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7027)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Annotations.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Annotations.cs
@@ -85,4 +85,61 @@ namespace System.Diagnostics.CodeAnalysis
         /// <summary>Gets the condition parameter value.</summary>
         public bool ParameterValue { get; }
     }
+
+    /// <summary>Specifies that the method or property will ensure that the listed field and property members have not-null values.</summary>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property, Inherited = false, AllowMultiple = true)]
+    internal sealed class MemberNotNullAttribute : Attribute
+    {
+        /// <summary>Initializes the attribute with a field or property member.</summary>
+        /// <param name="member">
+        /// The field or property member that is promised to be not-null.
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Annotations.cs
```diff
@@ -85,4 +85,61 @@ internal sealed class DoesNotReturnIfAttribute : Attribute
         /// <summary>Gets the condition parameter value.</summary>
         public bool ParameterValue { get; }
     }
+
+    /// <summary>Specifies that the method or property will ensure that the listed field and property members have not-null values.</summary>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property, Inherited = false, AllowMultiple = true)]
+    internal sealed class MemberNotNullAttribute : Attribute
```
swesonga, line 91: For my information: how are these picked up for nullability analysis?

tmeschter: The C# compiler looks for attributes with a certain name and shape. If it finds them, it can use the information they provide as part of its nullability analysis. It doesn't care if they come from source or referenced assemblies.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/Annotations.cs
drewnoakes: These attributes are not needed when targeting .NET Core 3.0 or later, where they're included in the core framework. We're targeting .NET Framework so have to define them. The compiler will happily use them.

PR #6992, opened 2021-02-26
Update CPS references to latest 16.10 SDK

swesonga:
This will enable usage of the latest CPS APIs

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/6992)

PR #7029, opened 2021-03-16
Update the project system version to 16.10.

swesonga:
#6978 introduced a dependency on 16.10 CPS API (with the CPS SDK updated in #6992). This follow-up change updates the project system version accordingly.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7029)


swesonga:
We currently require the build system to have the VS indicated by ProjectSystemToolsVersion. However, it is theoretically possible that we may not want those to match, in which case a different property could be useful. It looks like we only change ProjectSystemToolsVersion cross major versions of VS since we also base our assembly versions on this property in build\import\OverrideRepoToolsetVersions.props. What I was looking for was a way to indicate which version of VS is required to successfully run the code built from this repo (technically, a specific branch). I do not see that right now so I'm closing this PR for now.

PR #7031, opened 2021-03-18
Handle actions for launch profiles

tmeschter:
Handles actions to add, remove, rename, and duplicate launch profiles.

The remaining action to handle is setting a property, but the API for that is in a bit of flux at the moment.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7031)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -93,9 +93,9 @@
     <PackageReference Include="VSSDK.TemplateWizardInterface"                                         Version="12.0.4"                          ExcludeAssets="All" />
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK"                               Version="16.10.162-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.162-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="16.10.162-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK"                               Version="16.10.137-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.137-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="16.10.137-pre" />
 
     <!-- Roslyn -->
     <PackageReference Update="Microsoft.VisualStudio.LanguageServices"                                Version="3.10.0-2.21121.14" />
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -93,9 +93,9 @@
     <PackageReference Include="VSSDK.TemplateWizardInterface"                                         Version="12.0.4"                          ExcludeAssets="All" />
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK"                               Version="16.10.127-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.127-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="16.10.127-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK"                               Version="16.10.162-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.162-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="16.10.162-pre" />
 
     <!-- Roslyn -->
     <PackageReference Update="Microsoft.VisualStudio.LanguageServices"                                Version="3.10.0-2.21121.14" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/AddLaunchProfileAction.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/AddLaunchProfileAction.cs
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.ProjectSystem.Debug;
 using Microsoft.VisualStudio.ProjectSystem.Query;
@@ -20,7 +21,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
             _executableStep = executableStep;
         }
 
-        protected override Task ExecuteAsync(ILaunchSettingsProvider launchSettingsProvider)
+        protected override Task ExecuteAsync(ILaunchSettingsProvider launchSettingsProvider, CancellationToken cancellationToken)
         {
             return launchSettingsProvider.AddOrUpdateProfileAsync(
                 new WritableLaunchProfile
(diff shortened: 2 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/AddLaunchProfileAction.cs
```diff
@@ -0,0 +1,34 @@
...
+        private readonly AddLaunchProfile _executableStep;
+
+        public AddLaunchProfileAction(AddLaunchProfile executableStep)
+        {
+            _executableStep = executableStep;
+        }
+
+        protected override async Task ExecuteAsync(ILaunchSettingsProvider launchSettingsProvider)
+        {
+            await launchSettingsProvider.AddOrUpdateProfileAsync(
```
drewnoakes, line 25: Could just return the task to avoid the state machine. `RemoveLaunchProfileAction` does this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/RenameLaunchProfileAction.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assumes.NotNull(launchSettings);
+
+            ILaunchProfile? existingProfile = launchSettings.Profiles.FirstOrDefault(p => StringComparers.LaunchProfileNames.Equals(p.Name, _executableStep.CurrentProfileName));
+            if (existingProfile is not null)
+            {
+                var writableProfile = new WritableLaunchProfile(existingProfile);
+                writableProfile.Name = _executableStep.NewProfileName;
+
+                await launchSettingsProvider.RemoveProfileAsync(_executableStep.CurrentProfileName);
+                await launchSettingsProvider.AddOrUpdateProfileAsync(writableProfile.ToLaunchProfile(), addToFront: false);
```
drewnoakes, line 37: Would there be any benefit to running these tasks concurrently?

tmeschter: No. The `ILaunchSettingsProvider` serializes these operations so the underlying state _can't_ be modified concurrently, but that's an implementation detail the caller shouldn't depend on. If performance is a concern, a dedicated `ILaunchSettingsProvider3.RenameProfileAsync` method would be my preferred way of dealing with this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
```diff
@@ -0,0 +1,36 @@
...
+            if (((IEntityValueFromProvider)result.Result).ProviderState is UnconfiguredProject project
+                && project.Services.ExportProvider.GetExportedValueOrDefault<ILaunchSettingsProvider>() is ILaunchSettingsProvider launchSettingsProvider)
+            {
+                await ExecuteAsync(launchSettingsProvider);
+            }
+
+            await ResultReceiver.ReceiveResultAsync(result);
+        }
+
+        protected abstract Task ExecuteAsync(ILaunchSettingsProvider launchSettingsProvider);
```
drewnoakes, line 34: Should pass `CancellationToken` here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/DuplicateLaunchProfileAction.cs
```diff
@@ -0,0 +1,41 @@
...
+        private readonly DuplicateLaunchProfile _executableStep;
+
+        public DuplicateLaunchProfileAction(DuplicateLaunchProfile executableStep)
+        {
+            _executableStep = executableStep;
+        }
+
+        protected override async Task ExecuteAsync(ILaunchSettingsProvider launchSettingsProvider)
+        {
+            ILaunchSettings? launchSettings = await launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite);
```
drewnoakes, line 27: ...and use `CancellationToken` here, for example.

tmeschter:
This is ready for review, but the actions will not function correctly until #7054 is merged.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/DuplicateLaunchProfileAction.cs
tmeschter: You mean pass the token to `WaitForFirstSnapshot` or check it for cancellation?

drewnoakes: Pass it, if accepted, or use `.WithCancellation` or whatever the extension method is called.

tmeschter: `WaitForFirstSnapshot` takes a time out (in milliseconds) rather than a `CancellationToken`.

drewnoakes: It's the infinite timeout that concerns me here. If that snapshot never arrives (for one of many reasons) any tasks waiting on this to complete will linger indefinitely.

Looking at the implementation in `LaunchSettingsProvider`, I think it should call `TryCancel` on its `_firstSnapshotCompletionSource` in its `Dispose` method. Then if the provider is disposed while something is waiting for that first snapshot, the task will be cancelled and a leak will be avoided.

drewnoakes: https://github.com/dotnet/project-system/pull/7055 addresses part of my concern here.

The other part: if the execution is cancelled, the caller may still be stuck here waiting for longer than necessary until their cancellation takes effect. If it's not possible to modify `ILaunchSettingsProvider` then consider adding `ILaunchSettingsProvider2` that accepts a cancellation token. The current implementation can trivially be adapted to support cancellation.

tmeschter: I think there's a bigger issue here in that `ILaunchSettingsProvider` in general has no support for cancellation. I've filed #7056 to address that.

build/import/Packages.targets
```diff
@@ -93,9 +93,9 @@
     <PackageReference Include="VSSDK.TemplateWizardInterface"                                         Version="12.0.4"                          ExcludeAssets="All" />
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK"                               Version="16.10.127-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.127-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="16.10.127-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK"                               Version="16.10.162-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.162-pre" />
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="16.10.162-pre" />
```
swesonga, line 98: Does this version allow building the unit tests? Not sure if CPS has already fixed the extraneous assembly in the package.

drewnoakes (approved):

build/import/Packages.targets
tmeschter: @drewnoakes found a slightly older version that has everything I need but doesn't have the extraneous assembly, and I've downgraded to that.

tmeschter:
Related to #6896.

PR #7032, opened 2021-03-18
Atomically read and update profiles and global settings

tmeschter:
The `LaunchSettingsProvider` allows multiple concurrent changes to a project's launch settings by forcing them into a serialized order (via the `SequentialTaskExecutor`). However, two different changes to the same launch profile or global setting could still stomp over each other as there is no way to read and update these settings as a single operation.

Say operation A and operation B want to update different parts of the same launch profile. They each grab the current state, P0. They then each make their own changes to produce PA and PB, respectively. Then they both call `AddOrUpdateProfileAsync` to apply their changes. This method replaces the entire profile at once, so depending on the order in which the calls are processed we'll either end up with PA (in which case B's changes are lost) or PB (A's changes are lost) when what we really want is PAB (with both changes).

Global settings have a similar update mechanism, and a similar problem.

To fix this, this change adds new update methods that take callbacks to make the actual changes. When the update is processed, the `LaunchSettingsProvider` invokes the callback, providing the current state of the profile or global settings. The callback makes its changes, and returns the resulting object which is applied to the settings immediately. Any caller making use of these methods is guaranteed to see the current state of the profile/global setting when they make their changes.

Eventually these methods will be used by the new Launch Settings UI back end, but we should also consider adopting it in other places as well.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7032)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider3.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider3.cs
@@ -41,8 +41,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
         ///     Supports the retrieval and update of a given global setting as a single
         ///     operation. When <paramref name="updateFunction"/> is called it is given the
         ///     current value of the <paramref name="settingName"/> global setting (which may be
-        ///     <c>null</c> if no such setting exists and returns the updated object (or
-        ///     <c>null</c> if the setting is to be removed).
+        ///     <see langword="null"/> if no such setting exists and returns the updated object (or
+        ///     <see langword="null"/> if the setting is to be removed).
         /// </summary>
         /// <param name="settingName"></param>
         /// <param name="updateFunction"></param>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider3.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider3.cs
@@ -1,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    ///     Provides an implementation of <see cref="ILaunchSettingsProvider"/> with
+    ///     additional methods for atomically updating a launch profile or global setting.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    ///     The existing <see cref="ILaunchSettingsProvider.AddOrUpdateProfileAsync"/> method
+    ///     does not work well in situations where multiple update operations may run
+    ///     concurrently because the retrieval of the current set of values is separated from
+    ///     the later update. If operations A and B each retrieve the current profile, make
+    ///     their changes, and then apply them there is a very good chance that A's changes
+    ///     will overwrite B's or the other way around. The new method introduced here makes
+    ///     retrieving and updating a profile an single operation.
+    /// </para>
+    /// <para>
+    ///     The <see cref="ILaunchSettingsProvider.AddOrUpdateGlobalSettingAsync"/> method has
+    ///     similar problems.
+    /// </para>
+    /// </remarks>
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    public interface ILaunchSettingsProvider3 : ILaunchSettingsProvider2
+    {
+        /// <summary>
+        ///     Supports the retrieval and update of a given launch profile as a single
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider3.cs
```diff
@@ -0,0 +1,52 @@
...
+        ///     appropriate and when it is done the profile is applied to the settings.
+        /// </summary>
+        Task UpdateProfileAsync(string profileName, Action<IWritableLaunchProfile> updateAction);
+
+        /// <summary>
+        ///     Supports the retrieval and update of a given global setting as a single
+        ///     operation. When <paramref name="updateFunction"/> is called it is given the
+        ///     current value of the <paramref name="settingName"/> global setting (which may be
+        ///     <c>null</c> if no such setting exists and returns the updated object (or
+        ///     <c>null</c> if the setting is to be removed).
```
drewnoakes, line 45: Using `<cref langword="null"/>` provides a better cross-language experience. For example, `Nothing` is shown in VB.NET.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -815,6 +816,47 @@ public Task RemoveProfileAsync(string profileName)
             });
         }
 
+        public Task UpdateProfileAsync(string profileName, Action<IWritableLaunchProfile> updateAction)
+        {
+            // Updates need to be sequenced
+            return _sequentialTaskQueue.ExecuteTask(async () =>
+            {
+                ILaunchSettings currentSettings = await GetSnapshotThrowIfErrors();
```
drewnoakes, line 824: Can we change `GetSnapshotThrowIfErrors` to have an `Async` suffix?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -815,6 +816,47 @@ public Task RemoveProfileAsync(string profileName)
...
+                        existingProfile = p;
+                        break;
+                    }
+                    insertionIndex++;
+                }
+
+                if (existingProfile is null)
+                {
+                    return;
+                }
```
drewnoakes, line 840: The caller might reasonably want to know that this failed, either via an exception or a return value.

drewnoakes (commented):
TIL about global settings :)

Regarding the previous methods that allowed for race conditions: are there some members that can be marked as obsolete, or at least have their documentation extended to point to these new members as better alternatives?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider3.cs
tmeschter: I always forget about `langword="null"`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Yes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Hmm. We could return a `bool` to indicate if the change was successfully applied. I'm not sure about the exception--I don't think it is exceptional that the profile might be renamed or deleted before the update is applied.

If this fails, what is the caller going to do?

tmeschter:
> TIL about global settings :)

Yeah. :-) At the Project Query API level they will be exposed as properties on the launch profile like all the others. It does mean that updating a property on a launch profile can effectively change _other_ launch profiles, but the Project Query API update mechanism should take care of that. I guess it will be a good test that we've wired everything up correctly. :-)

> Regarding the previous methods that allowed for race conditions: are there some members that can be marked as obsolete, or at least have their documentation extended to point to these new members as better alternatives?

The `AddOrUpdateProfileAsync` and `AddOrUpdateGlobalSettingAsync` methods are still useful for _adding_ a profile/global setting, but it might make sense to deprecate those entirely in favor of separate methods for adding and updating. We've obsoleted members of `ILaunchSettingsProvider` before--the `LaunchSettingsFile` property has been obsoleted in favor of `ILaunchSettingsProvider2.GetLaunchSettingsFilePathAsync`, for example.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
drewnoakes: > If this fails, what is the caller going to do?

Display a message to the user rather than fail silently. I can't think of anything better than that.

tmeschter: I've updated the method to return a `bool`.

drewnoakes (approved):

PR #7038, opened 2021-03-23
Reorder properties on Application page

drewnoakes:
Historically in legacy projects, the _Assembly Name_ and _Default Namespace_ properties were manually kept in sync with the project by the user.

With SDK-style projects, it's more common to go with the default behaviour of picking up the `MSBuildProjectName` property, which is computed from the project's file name.

Given these properties are far less likely to be user edited, we move them further down the UI, allowing more commonly modified properties to rise to the top.

| Before | After |
|-|-|
| image | image |

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7038)

### DIFF
--- a/docs/repo/property-pages/property-specification.md
+++ b/docs/repo/property-pages/property-specification.md
@@ -225,7 +225,8 @@ XAML files in the dotnet/project-system repo are configured for automatic locali
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


tmeschter (approved):
I'm on the fence about whether Output Type or Target Framework should go first, but either is an improvement.

drewnoakes:
> I'm on the fence about whether Output Type or Target Framework should go first, but either is an improvement.

Yes I also contemplated this. Once we get telemetry around the frequency of property updates, we can revisit ordering using actual data to inform these decisions.

jjmew (approved):
:shipit:

PR #7023, opened 2021-03-15
Add a link to show the new Launch Profile UI

drewnoakes:
This PR adds a link action that invokes a command in CPS to show the new Launch Profile UI. The link is displayed on the "Debug" page of the new Project Properties UI.

Sibling CPS PR (Internal): https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/310621

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7023)

PR #7040, opened 2021-03-23
Reinstate "Debug" property section with placeholder text

drewnoakes:
Adds placeholder text, explaining that launch profile editing is not currently supported in the new Project Properties UI, that it is coming soon, and how to work around the limitation if needed.

image

Will only display properly in combination with https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/312423, although it is safe to merge before then.

Additionally, add links from the documentation to commits in this PR for reference when making similar changes to the property system in future. These links are to commits, so this PR should not be rebased without updating those links (important because the doc updates may conflict, depending upon PR merge order).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7040)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/DebugPropertyPage.xaml
```diff
@@ -0,0 +1,31 @@
...
+      Description="Debug"
+      DisplayName="Debug"
+      PageTemplate="generic"
+      Order="500"
+      xmlns:sys="clr-namespace:System;assembly=mscorlib"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns="http://schemas.microsoft.com/build/2009/properties">
+
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFile"
```
tmeschter, line 13: If you change this to `ProjectFileWithInterception` then you don't need the `<DataSource>` element within the `DebugPagePlaceholderDescription` property.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/DebugPropertyPage.xaml
```diff
@@ -0,0 +1,31 @@
...
+                SourceOfDefaultValue="AfterContext"
```
tmeschter, line 14: Not strictly necessary since we don't need to worry about default values for the one property currently on this page.

tmeschter (approved):

drewnoakes:
@tmeschter I'll rationalise the data sources in #7023 which will remove this new description property and add a link action.

jjmew (approved):
:shipit:

PR #7043, opened 2021-03-23
Support reading/writing arbitrary values to launch profiles

tmeschter:
`ILaunchProfile` has a few pre-defined properties (like ExecutablePath, CommandLineArgs, LaunchBrowser, etc.) but it also supports storage of arbitrary objects via the `ImmutableDictionary<string, object>` accessible via the `OtherSettings` property, and in practice the bulk of the properties we care about are stored there. Separately, `ILaunchSettings.GlobalSettings` is another `ImmutableDictionary<string, object>` storing launch settings that are global (that is, not tied to a particular launch profile).

There are a few operations we need to support:
1. In each case the property value comes in as a `string`, but needs to be stored in the dictionary as an `object`. While we could store it as a `string` by default this would be incorrect in many cases, and so we need a converter for each property we wish to store.
2. We also need to convert back from an `object` to a string
3. Changing one property may also require changes to other properties, and so we also need a way to modify the launch settings (or global settings) as a whole.

These new interfaces provide the necessary abilities.

Implementation classes must be tagged with the `ExportLaunchProfileExtensionValueProviderAttribute` indicating the names of the properties they handle, as well as whether it applies to the launch profiles or the global settings. In the former case it must implement `ILaunchProfileExtensionValueProvider`; in the latter `IGlobalSettingExtensionValueProvider`. The extensions will be imported by the `LaunchProfileProjectPropertiesProvider` which will redirect all reads and writes of the related properties through them.

These interfaces do almost the same thing, but there are some important differences which lead to different API shapes. When accessing the properties of a launch profile we may want to know the current global settings, and so those are passed along. However, when modifying a global setting we would not expect it to depend on or modify values in individual launch profiles, so those are _not_ available.

This change also adds an implementation of `ILaunchProfileExtensionValueProvider` to handle the extension properties we utilize for profiles using the Project and Executable debug commands.

Note that `LaunchProfileProjectProperties` will use the extension points for reading properties, but it does not yet support setting properties at all.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7043)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ExportLaunchProfileExtensionValueProviderAttribute.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ExportLaunchProfileExtensionValueProviderAttribute.cs
@@ -1,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.ComponentModel.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    /// Exports a <see cref="ILaunchProfileExtensionValueProvider" /> or <see cref="IGlobalSettingExtensionValueProvider"/>.
+    /// </summary>
+    [MetadataAttribute]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = false, Inherited = false)]
+    public sealed class ExportLaunchProfileExtensionValueProviderAttribute : ExportAttribute
+    {
+        public string[] PropertyNames { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExportLaunchProfileExtensionValueProviderAttribute"/>
+        /// class for a single intercepted property.
+        /// </summary>
+        public ExportLaunchProfileExtensionValueProviderAttribute(string propertyName, ExportLaunchProfileExtensionValueProviderScope scope)
+            : this(new[] { propertyName }, scope)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExportLaunchProfileExtensionValueProviderAttribute"/>
+        /// class for multiple intercepted properties.
+        /// </summary>
+        public ExportLaunchProfileExtensionValueProviderAttribute(string[] propertyNames, ExportLaunchProfileExtensionValueProviderScope scope)
+            : base(GetType(scope))
+        {
+            PropertyNames = propertyNames;
+        }
+
+        private static Type? GetType(ExportLaunchProfileExtensionValueProviderScope scope)
+        {
+            return scope switch
+            {
+                ExportLaunchProfileExtensionValueProviderScope.LaunchProfile => typeof(ILaunchProfileExtensionValueProvider),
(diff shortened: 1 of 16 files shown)


jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ExportLaunchProfileExtensionValueProviderAttribute.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.ComponentModel.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    /// Exports a <see cref="ILaunchProfileExtensionValueProvider" /> or <see cref="IGlobalSettingExtensionValueProvider"/>.
```
swesonga, line 9: nit: this isn't the class that exports those interfaces, right?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchProfileExtensionValueProviderMetadata.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    /// Metadata mapping interface for the <see cref="ExportLaunchProfileExtensionValueProviderAttribute"/>.
+    /// </summary>
+    public interface ILaunchProfileExtensionValueProviderMetadata
+    {
+#pragma warning disable CA1819 // Properties should not return arrays
```
swesonga, line 10: Why not use IReadOnlyCollection instead of suppressing this?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ExportLaunchProfileExtensionValueProviderAttribute.cs
tmeschter: If you want to export an `ILaunchProfileExtensionValueProvider` or an `IGlobalSettingExtensionValueProvider` you put this attribute on the type, yes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchProfileExtensionValueProviderMetadata.cs
tmeschter: This is consistent with the other places we define types for collecting export attribute metadata. That's not really a good _technical_ reason, of course. :-)

I'll look into `IReadOnlyCollection` and see if that would be possible.

tmeschter: My understanding is that these need to be arrays in order to match the properties in the related attributes.

PR #7044, opened 2021-03-24
Report versions via Project Query API

tmeschter:
In executing a query via the Project Query API we may retrieve data from multiple and varied sources. The query engine monitors known sources for changes in their versions, so if we report a version number for each source utilized during the query the engine can determine when a query needs to be re-run to acquire updated information.

In our case we use two well-known, versioned data sources for the project property pages: the `UnconfiguredProject` (as the source of the pages, properties, and metadata) and the `ConfiguredProject` (as the source of values and possible supported values). Here we update key parts of our Project Query API to report the appropriate version numbers.

- If we have an `UnconfiguredProject` handy, we can ask it for its version directly.
- `IPropertyPageQueryCache` has also been updated to return version numbers for both the `UnconfiguredProject` and `ConfiguredProject` as it is already responsible for holding on to those for the duration of a query.

Note that we don't report version numbers for every kind of entity we create. For example, you _could_ query for just the editor metadata on a particular editor on a particular property, separate from querying for the property itself--but there doesn't seem to be a good reason to do so. Ideally, the editors and their metadata would not be separate queryable entities but just part of the property entity itself, but we can't do that with the current implementation of the Project Query API. Since it is not convenient to retrieve the version information when producing editor metadata, nor particularly useful, we just skip it here. The same goes for other entities.

When setting a property value through the API, we also need to report the new version number to the query engine.

Note that this commit does not handle versioning of launch profiles. Changes to those are not captured in the versioning of the `UnconfiguredProject`, so we will need to create a versioning scheme for them and then expose it as a known versioned data source to the query engine.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7044)


drewnoakes (approved):
LGTM though I'm not very familiar with the changes in requirements for reporting versions.

tmeschter:
> LGTM though I'm not very familiar with the changes in requirements for reporting versions.

Neither am I. :-) This is heavily based on similar data providers in the CPS code, with me doing my best to reason about when and where we do or do not need to report versions. There will almost certainly be bugs once we have push notifications working end to end.

Issue #6936, opened 2021-02-09
Identify pages and properties missing in the F# experience


tmeschter:
We need to identify which F#-specific pages and properties are missing from the new project property pages and file issues to address those.

cartermp:
I can share a screenshot, but the only ones that come to mind are in the Build page (debug config):

|Property|Current UI|What it does|Context|
|--------|----------|------------|-------|
|OtherFlags|Textbox|Plops `<OtherFlags>` in the project file properties|General dumping ground for F# compiler options|
|Optimize code|Checkbox|Plops `--optimize` in `<OtherFlags>`.|On by default in release|
|Generate tail calls|Checkbox|Plops `--tailcalls` in `<OtherFlags>`|On by default in release|

tmeschter:
@cartermp So `OtherFlags` looks like this in the project file?

``` XML
<OtherFlags>--optimize --foo --bar --tailcalls</OtherFlags>
```

cartermp:
Yep!

tmeschter:
@cartermp Can I close this in favor of dotnet/fsharp#11532?

cartermp:
Yep, sounds fair.

Issue #6935, opened 2021-02-09
Identify pages and properties missing in the VB experience


tmeschter:
The work here is to _identify_ which VB-specific pages and properties are missing from the new project property pages and file issues to address those.

paul1956:
UseWindowsForms and UseWPF when you want to set both.
MinimumSplashScreenDisplayTime
Possibility UseCompatibleTextRendering
If and when Code Cleanup for VB gets integrated, separate profiles/configurations for VB and C# under Analyzer/Code CleanUp.
When configuring CodeCleanup the list of available cleanups should be by Language (especially important when names are the same).

melytc:
Notes on our review:
- Icon is in the resources pages.
- View windows settings (creates if needed, and shows app.manifest). This is might be covered by the Manifest property > Use a custom manifest.
- `Compile` page. The warnings are shown in a different way (list vs table). How to encourage users to go to editor config for warnings. 
- How dynamic are the warning configurations lists VB has? 
- `Advanced Compiler Settings`
- `Remove integer overflows checks` is in intermediate mode.
- Compilation constants (VB) and conditional compilation symbols (C#) are different but stored in the same msbuild property.

tmeschter:
At this point I believe I've split off all the individual work items identified here. Closing.

PR #7047, opened 2021-03-25
Only enable the new Project Properties UI for C# projects

drewnoakes:
So far the new Project Properties UI has good support for most C# project types. VB (#6935) and F# (#6936) support has not been addressed yet.

This change prevents delegation from the legacy editor to the new editor unless the `CSharp` capability is present. This will be important as we start to roll out by default to more MS internal folks.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7047)


jjmew (approved):
:shipit:

Issue #7052, opened 2021-03-25
New Property Pages: Allow visibility conditions per variation


MiYanni:
#### Summary
Some properties are directly related to other properties. Meaning, the state of one property determines something about another property. In this case here, `Treat warnings as errors` determines if one property can be shown (`Treat specific warnings as errors`) versus showing another property (`Exclude specific warnings as errors`). It works like a light-switch, when it is off, you see `Treat specific warnings as errors`, and when it is on, you see `Exclude specific warnings as errors`.

However, there is a problem. `Treat warnings as errors` can be varied by Configuration. Meaning, each configuration variation (`Release`, `Debug`, etc.) can have their own value for this light-switch. If one is off and one is on, you'll see both `Treat specific warnings as errors` and `Exclude specific warnings as errors`. The relationship to which variation those apply is the problem.

Each of these properties are, therefore, able to be varied by Configuration. But the UI doesn't display this properly. Firstly, it does not automatically go into `Vary value by Configuration` mode. This means, it shows a single edit box instead of one edit box per Configuration variation. If you switch it to `Vary value by Configuration` mode, it will show Configurations for which this value does not apply! This adds to the confusion. The ask here is to make dependent properties (those whom depend on other properties to be visible) to automatically react to varied parent properties.

#### Expected Behavior
| Before      | After
| ----------- | -----------
| Going from this state, all variation values of the parent property are the same [unchecked]. | To this state [checked], will properly show varied UI for dependent properties. I've make a mock-up here.
| ![image](https://user-images.githubusercontent.com/17788297/112550565-cd043c00-8d7c-11eb-9df4-f1587dab8eb7.png)   | ![image](https://user-images.githubusercontent.com/17788297/112550975-6cc1ca00-8d7d-11eb-9d70-f4f3c77dba7d.png)

#### Actual Behavior
| Before      | After          | Additional
| ----------- | ----------- | -----------
| Going from this state, all variation values of the parent property are the same [unchecked]. | To this state [checked], does not show variation UI for dependent properties. | Additionally, enabling variation UI for dependent properties improperly shows variations that are not applicable for those properties, given the parent property's state.
[...]

Issue #7050, opened 2021-03-25
New Property Pages: Switching dimensional properties between varied/unvaried states causes non-project file values to be written, causes self-referencing string expression duplication


MiYanni:
#### Summary
Some strings contain expressions. These expression-based strings are shown in the editing fields. However, these expressions don't exist in the project file and likely come from the SDK. For strings that have dimensions (configuration, platform, etc.), you can make the UI switch between varied and unvaried states. This will show a control per variation for each dimension of that property. However, switching from a varied state to unvaried will write the values of each variation into the project file. If that string expression references the property itself, this causes duplication of any terms beyond the expression part of the string.

#### Steps to Reproduce

1. Pick a string property that the SDK gives a self-referencing expression (such as `Conditional compilation symbols`). Observe the evaluated value (under the edit box).
2. Use the ⚙ to select `Vary value by Configuration`. This will split the edit box into one per variation of Configuration.
3. Use the ⚙ to select `Use the same value across all configurations`. This will change the UI back to a single edit field.
4. Observe that the evaluated value is now different from what it originally was.
5. Open the project file directly (double click on it) and observe the file now has the original string value written to it, once per each variation of Configuration.

#### Expected Behavior
Since the value was not edited, it should remain the same after flipping between `Vary value by Configuration` and `Use the same value across all configurations`. There should not be anything written to the project file.

#### Actual Behavior
The values of each variation of Configuration are written to the project file. This causes the evaluated value of the property to change since it uses a self-referencing expression.

#### Clip of Issue
![EvaluatedStringDuplicationIssue](https://user-images.githubusercontent.com/17788297/112545416-6cbdcc00-8d75-11eb-81c6-c87de8382a7b.gif)

drewnoakes:
This applies to any property which is defined in terms of itself (i.e. also conditional compilation symbols).

When we select "Use the same value across all configurations", we take the first value from the N varied values and apply it across the board. In this case, we send that update to the server, which bakes it into itself and produces the strange value seen above.

Any fix for this should also cover #6931.

Issue #7049, opened 2021-03-25
New Property Pages: Double click (fast click) dimensional checkbox causes desync, UI splits into varied state


MiYanni:
#### Summary
If you have a checkbox that can vary by dimensions (configuration, platform, etc.), it can be set/unset so fast (while non-varied), the value can be misunderstood by the UI. This causes the UI to think that at least 1 of the dimensions is a different value than the others; causing the UI to split into the varied display of the control. This is performed while the control is unvaried, and the result is the control become unintentionally varied.

#### Steps to Reproduce

1. Find a control (such as `Allow unsafe code`) that is a checkbox and allows dimensional varying. Make sure the control is in the non-varied (single checkbox) state.
2. Check and uncheck the control slowly. Observe that it works as intended.
3. Now, quickly (such as double click) click the checkbox. Observe that it now has changed to being the varied display of the control, showing multiple dimensions.

#### Expected Behavior
Slow un/checking of the control operates the same way as fast un/checking of the control.

#### Actual Behavior
When un/checked quickly, the control goes from a non-varied state into a varied state.

#### Clip of Issue
![DoubleClickCheckboxIssue](https://user-images.githubusercontent.com/17788297/112530625-e77deb80-8d63-11eb-9946-40d3add5eb5f.gif)

Issue #6989, opened 2021-02-25
High-priority list of properties requiring attention (3)


tmeschter:
_This work item has been split off of #6907 to contain the highest priority items._

The following properties are currently included in the UI and require some sort of additional treatment to make them work correctly.

- [x] Order of values in target framework drop down (both .NET Core and .NET Framework families) (possibly CPS, maybe [this](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/298982))
  - This may be a bug in our SupportedTargetFrameworksProvider; we may need to do some explicit sorting.
- [x] `Nullable` should only be visible when using C# 8 or greater
  - Possible solution: expose the language version as a hidden property, and condition the visibility of `Nullable` on that.
- [x] ~~`Nullable` appearing with nothing selected~~ Opened https://github.com/dotnet/project-system/issues/7360
  - This is because the SDK doesn't provide a default value for the `<Nullable>` property. We don't currently have a good _general_ solution to this, but it could be handled with an `IInterceptingPropertyValueProvider`: if the property isn't set it could return `disabled` for the value.
- [x] Toggling "use multi-targeting" shows a project fault (fix or remove these options)
  - We should remove the check box, and condition the visibility of the `TargetFramework` and `TargetFrameworks` properties on those properties actually being defined. That way only one or the other will show.
- [x] Rationalize `TreatWarningsAsErrors`, `NoWarn`, `WarningsAsErrors` and `WarningsNotAsErrors` (see #2409 for some context)
  - These properties are all related, so we should make sure the user understands how they work together.
- [ ] "Prefer 32-bit" should show, but only when targeting .NET Framework (PR #7362)
  - When single-targeting we should show this if the target framework is from the .NET Framework family.
  - When multi-targeting, we should show this if *any* of the target frameworks is from the .NET Framework family.

tmeschter:
@jjmew Here's the issue tracking the property page tweaks that I think someone other than me or @drewnoakes could make. I've excluded items that I view as low priority (e.g., changes specifically to support non-SDK projects) or are complex.

@drewnoakes I split these out of #6907. Let me know if there are other things we could include here.

drewnoakes:
@tmeschter @jjmew I think it would be easier to track if these were broken out into separate issues. Priorities can be fluid, and these may be assigned to different people.

tmeschter:
@drewnoakes Per @jjmew's request, I've kept these in a single issue.

jjmew:
[...]

PR #7053, opened 2021-03-25
New Property Pages: Update warnings-related property descriptions, change Treat warnings as Errors into a checkbox

MiYanni:
Related: https://github.com/dotnet/project-system/issues/6989

I looked at the warnings-related properties and attempted to make the descriptions/names more uniform in terms of nomenclature. Additionally, I've changed `Treat warnings as errors` to a boolean property (checkbox) so that it is obvious this only has 2 states (on and off, in essence). When checking the checkbox, you'll see `Exclude specific warnings as errors` replace the `Treat specific warnings as errors` property. Thus, these are mutually exclusive based on the `Treat warnings as errors` state. Lastly, none of the help links were working. I searched and found some reference docs for those items. Unfortunately, most of them do not have conceptual docs.

During these changes, I've found a few issues and a suggestion:
- https://github.com/dotnet/project-system/issues/7049
- https://github.com/dotnet/project-system/issues/7050
- https://github.com/dotnet/project-system/issues/7052

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7053)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -179,50 +179,48 @@
                DisplayName="5" />
   </EnumProperty>
 
+  <!-- TODO: Create fwlink for this HelpUrl -->
```
drewnoakes, line 182: @tmeschter created the last batch of fwlinks. We should create them in the same way, with owners configured so that we can easily update them in future.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/TreatWarningsAsErrorsValueProvider.cs
```diff
@@ -21,21 +21,19 @@ public TreatWarningsAsErrorsValueProvider(ITemporaryPropertyStorage temporaryPro
...
+                return null;
             }
 
+            // When true, remove WarningsAsErrors. Otherwise, remove WarningsNotAsErrors.
+            string removePropertyName = value ? WarningsAsErrorsProperty : WarningsNotAsErrorsProperty;
+            string restorePropertyName = value ? WarningsNotAsErrorsProperty : WarningsAsErrorsProperty;
+
+            await defaultProperties.SaveValueIfCurrentlySetAsync(removePropertyName, _temporaryPropertyStorage);
+            await defaultProperties.DeletePropertyAsync(removePropertyName, dimensionalConditions);
+            await defaultProperties.RestoreValueIfNotCurrentlySetAsync(restorePropertyName, _temporaryPropertyStorage);
```
drewnoakes, line 35: Is there a change here, or just refactoring?

MiYanni: The primary acquisition of `value` changed from being a string to a boolean. I realized each group of statements (within brackets) was the same. So, I just wrote it in a way where there was only 1 set of statements. I could go on a big philosophical tangent, but if you want me to change it back to if-else, let me know. Personally, I haven't written an if-else statement in several years unless I have to. I like collapsing forking decisions into expressions and assignments.

drewnoakes: I couldn't see a functional change, so wanted to double check this was just a refactoring and that I wasn't missing anything. It can be useful to split refactoring into separate commits or PRs to make review easier. Or if your PR has chunks of code that are moved without changes, call that out in the PR to make review easier.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
drewnoakes: @MiYanni are you going to set up these fwlinks before merging this? If not, please file an issue so we don't forget to clean this up.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -179,50 +179,48 @@
...
-                  DisplayName="Suppress warnings"
-                  Description="Blocks the compiler's ability to generate one or more warnings. Separate multiple warning numbers with a comma (',') or semicolon (';')."
-                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147300"
+                  DisplayName="Suppress specific warnings"
+                  Description="Blocks the compiler from generating the specified warnings. Separate multiple warning numbers with a comma (',') or semicolon (';')."
+                  HelpUrl="https://docs.microsoft.com/en-us/visualstudio/ide/how-to-suppress-compiler-warnings"
                   Category="ErrorsAndWarnings" />
 
-  <EnumProperty Name="TreatWarningsAsErrors"
-                HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147301"
```
tmeschter, line 189: This fwlink is working for me and redirects to the expected compiler documentation for this topic.

In general if there is a problem with a fwlink we should update the forwarding, rather than replacing the fwlink, as this allows us to fix the redirection for shipped versions of the product as well.

tmeschter (changes requested):
Please restore the fwlinks. Otherwise looks good.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
MiYanni: Interesting. I had tested all of the fwlinks and they did not work. These links are different from the ones I tested. Were they added recently?

tmeschter: I'm only aware of the fwlinks I've created for the property pages. Those were created back in Oct. 2020 and haven't been updated since then.

The ones I created are jointly owned by me and @jjmew. Ideally we would set up a security group to own them but I haven't gotten around to that.

MiYanni: Huh. I looked back at my Chrome history and the link failures were similar to this: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warning
Either way, I'll revert these changes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
MiYanni: HelpUrl changes were reverted.

tmeschter (approved):

drewnoakes (approved):

PR #7064, opened 2021-03-29
Enable indexing of sources when publishing symbols

swesonga:
Indexing of sources in this pipeline needs to be enabled until there is a different process in place that handles this. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7064)

### DIFF
--- a/build/ci/official.yml
+++ b/build/ci/official.yml
@@ -86,7 +86,7 @@ steps:
       '**/*.pdb
       **/*.dll
       **/*.exe'
-    indexSources: false
+    indexSources: true
     publishSymbols: true
     SymbolServerType: TeamServices
     SymbolsProduct: 'Roslyn Project System'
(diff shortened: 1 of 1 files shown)


build/ci/official.yml
```diff
@@ -86,7 +86,7 @@ steps:
       '**/*.pdb
       **/*.dll
       **/*.exe'
-    indexSources: false
+    indexSources: true
```
dmonroym, line 89: the default value is true, can we just remove this line instead?

drewnoakes (approved):

dmonroym (approved):

swesonga:
Note: see https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/index-sources-publish-symbols?view=azure-devops for more on how indexing sources works.

PR #7066, opened 2021-03-29
Add new CachedProjectEvaluation capability to default capabilities

arkalyanms:
Depends on https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/313873.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7066)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs
@@ -25,7 +25,8 @@ namespace Microsoft.VisualStudio.Packaging
                                       ProjectCapability.PreserveFormatting + "; " +
                                       ProjectCapability.ProjectConfigurationsDeclaredDimensions + "; " +
                                       ProjectCapability.LanguageService + "; " +
-                                      ProjectCapability.DotNet;
+                                      ProjectCapability.DotNet + "; " +
+                                      ProjectCapability.CachedProjectEvaluation;
 
         /// <summary>
         ///     Represents F#'s (fsproj) set of capabilities that are always present ("fixed").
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs
```diff
@@ -26,7 +26,7 @@ internal class ProjectTypeCapabilities
                                       ProjectCapability.ProjectConfigurationsDeclaredDimensions + "; " +
                                       ProjectCapability.LanguageService + "; " +
                                       ProjectCapability.DotNet + "; " +
-                                      ProjectCapability.CachedProjectEvaluation;
```
davkean: Are Shared Projects getting the same capability?

arkalyanms: Yes, that's the plan. Just adding more testing to bring shared projects into the fold.

davkean: Have you verified F# projects work under this mode (you've opt'd them in)? They have a different way of displaying the tree so might be affected by this change.

arkalyanms: Not yet, F# is in the test matrix for the cache service checkin acceptance. So far only been using CPS, project-system, orchardcore and some synthetic test solutions of varying size for the tests.

How are F# different? So I know what to test for.

davkean: Then specifically only opt in C# and VB: https://github.com/dotnet/project-system/blob/be0048c2e264608488391dee1a23921d47cb25fe/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeCapabilities.cs#L46-L57.

F# opt's into SortByDisplayOrder which requires evaluation to provide items in the order they appear in the project file.

drewnoakes (approved):
@arkalyanms explained that the feature this capability lights up is behind a feature flag, at least initially. Adding this now unblocks other work.

drewnoakes:
> error CS0117: 'ProjectCapabilities' does not contain a definition for 'UseProjectEvaluationCache'

Looks like the CPS package needs updating here.

PR #7069, opened 2021-03-30
Support looking up a Launch Profile by ID

tmeschter:
The `LaunchProfileFromProjectDataProducer` can take an `UnconfiguredProject` and produce a set of Project Query API entities representing the launch profiles in that project. However, we currently lack a way to map from an entity ID representing a launch profile to the actual entity. This is necessary in order to support actions on launch profiles via the Project Query API, namely, setting the value of a property in a launch profile. The query engine first tries to look up the launch profile based on its ID, and then applies the action.

The code to create and populate the actual `LaunchProfileValue` is needed in both scenarios (project -> launch profiles, and entity ID -> launch profile) and has been extracted into a new `LaunchProfileDataProducer` type.

With this change, the new property value can make its way from the Launch Profiles front end to the back end and on to the `LaunchProfileProjectProperties` type. A separate change to the `LaunchProfileProjectProperties` is needed to store the value properly.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7069)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileByIdDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileByIdDataProducer.cs
@@ -1,0 +1,77 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Framework.XamlTypes;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.Query;
+using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModel;
+using Microsoft.VisualStudio.ProjectSystem.Query.QueryExecution;
+using Microsoft.VisualStudio.ProjectSystem.VS.Utilities;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Query.PropertyPages
+{
+    internal class LaunchProfileByIdDataProducer : QueryDataByIdProducerBase
+    {
+        private readonly ILaunchProfilePropertiesAvailableStatus _properties;
+        private readonly IProjectService2 _projectService;
+        private readonly IPropertyPageQueryCacheProvider _queryCacheProvider;
+
+        public LaunchProfileByIdDataProducer(ILaunchProfilePropertiesAvailableStatus properties, IProjectService2 projectService, IPropertyPageQueryCacheProvider queryCacheProvider)
+        {
+            _properties = properties;
+            _projectService = projectService;
+            _queryCacheProvider = queryCacheProvider;
+        }
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileByIdDataProducer.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            _properties = properties;
+            _projectService = projectService;
+            _queryCacheProvider = queryCacheProvider;
+        }
+
+        protected override Task<IEntityValue?> TryCreateEntityOrNullAsync(IQueryExecutionContext executionContext, EntityIdentity id)
+        {
+            if (QueryProjectPropertiesContext.TryCreateFromEntityId(id, out QueryProjectPropertiesContext? context)
+                && StringComparers.ItemTypes.Equals(context.ItemType, "LaunchProfile"))
```
drewnoakes, line 31: Do we not have a constant for this item type?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileByIdDataProducer.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            return NullEntityValue;
+        }
+
+        private async Task<IEntityValue?> CreateLaunchProfileValueAsync(IQueryExecutionContext executionContext, EntityIdentity id, QueryProjectPropertiesContext context)
+        {
+            if (_projectService.GetLoadedProject(context.File) is UnconfiguredProject project
+                && project.Services.ExportProvider.GetExportedValueOrDefault<ILaunchSettingsProvider>() is ILaunchSettingsProvider launchSettingsProvider
+                && await project.GetProjectLevelPropertyPagesCatalogAsync() is IPropertyPagesCatalog projectCatalog
+                && await launchSettingsProvider.WaitForFirstSnapshot(Timeout.Infinite) is ILaunchSettings launchSettings)
```
drewnoakes, line 44: Is there a cancellation token in scope here, perhaps in the `executionContext`?

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileByIdDataProducer.cs
tmeschter: The `executionContext` carries a `CancellationToken`, yes. I need to go back and update *all* the Project Query API code to check for that, so I'm going to do it as a separate change. With the number of pieces in flight right now I'm worried I'll miss something if I do it piecemeal.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileByIdDataProducer.cs
tmeschter: No, and we definitely need one.

PR #7078, opened 2021-04-02
Fix dependency tree node customization via extensions

### LINKED ISSUE #6604: Allow project types to rename sub nodes into the dependency tree
MSIX deployment projects want to rename Projects -> Applications. We should let custom project types customize these nodes.

The References tree does this via the same mechasim that you customize the real tree:

``` C#
        [ImportMany(ReferencesProjectTreeCustomizablePropertyValues.ContractName)]
        protected OrderPrecedenceImportCollection<IProjectTreePropertiesProvider> ProjectTreePropertiesProviders { get; private set; }
```
``` C#
    [ProjectSystemContract(
        ProjectSystemContractScope.UnconfiguredProject,
        ProjectSystemContractProvider.Extension,
        ContractName = ReferencesProjectTreeCustomizablePropertyValues.ContractName,
        ContractType = typeof(IProjectTreePropertiesProvider))]
    public class ReferencesProjectTreeCustomizablePropertyValues : IProjectTreeCustomizablePropertyValues, IProjectTreeCustomizablePropertyValues2
    {
        /// <summary>
        /// Gets or sets the value which will be used in <see cref="IProjectTree.Caption"/>.
        /// </summary>
        public string? Caption { get; set; }

```

``` C#
        private void ApplyProjectTreePropertiesCustomization(
            IProjectTreeCustomizablePropertyContext context,
            ReferencesProjectTreeCustomizablePropertyValues values)
        {
            foreach (var provider in this.ProjectTreePropertiesProviders.ExtensionValues())
            {
                provider.CalculatePropertyValues(context, values);
            }
        }
```

Dependencies node should just do the same.

drewnoakes:
Fixes #6604

Previously `IProjectTreePropertiesProvider` implementations would only be invoked on _updates_ to tree nodes. With this change, they are invoked on both _creation_ and updates.

This change also improves performance in the common case by avoiding allocating instances of `ProjectTreeCustomizablePropertyContext` and `ReferencesProjectTreeCustomizablePropertyValues` when no relevant `IProjectTreePropertiesProvider` imports exist. Most projects will not have such an import.

Finally, we add a documentation file for the dependencies node, and include an example showing how to customize nodes.

---

This change supports the WAP team who are integrating with the dependencies tree in 16.10.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7078)

### DIFF
--- a/docs/repo/dependencies-node.md
+++ b/docs/repo/dependencies-node.md
@@ -1,0 +1,35 @@
+# Dependencies Node
+
+A high-level overview of this feature is available in [Dependencies Node Roadmap](dependencies-node-roadmap.md).
+
+## Customizing nodes
+
+It is possible to modify the caption, icon and `ProjectTreeFlags` of nodes in the dependencies tree.
+
+To do so, export an instance of `IProjectTreePropertiesProvider` as shown below. This example will rename the "Projects" node to "Applications":
+
+```c#
+[Export(ReferencesProjectTreeCustomizablePropertyValues.ContractName, typeof(IProjectTreePropertiesProvider))]
+[AppliesTo(ProjectCapabilities.AlwaysApplicable)]
+internal sealed class MyDependenciesTreePropertiesProvider : IProjectTreePropertiesProvider
+{
+    private static readonly ProjectTreeFlags ProjectDependencyGroup = ProjectTreeFlags.Create("ProjectDependencyGroup");
+
+    public void CalculatePropertyValues(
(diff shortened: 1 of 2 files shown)


edwardskrod (approved):
I implemented this in WAP.   It works perfectly.  
ApplicationsTree

jjmew (approved):

PR #7082, opened 2021-04-02
Update packages regarding Merged PIAs

dmonroym:
Updating more packages as they become available. Additionally we're renaming our packages to have the 17.0 suffix.

This is still not functional, it's WIP but publishing it to unblock others

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7082)


jjmew (approved):
:shipit:

PR #7090, opened 2021-04-05
Code cleanup.

ocallesp:
Handle null values properly to remove squiggles.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7090)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
@@ -54,7 +54,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Build
             if (!TryExtractErrorListDetails(error.BuildEventArgs, out ErrorListDetails details) || string.IsNullOrEmpty(details.Code))
                 return AddMessageResult.NotHandled;
 
-            bool handled = await _projectContextHost.OpenContextForWriteAsync(accessor =>
+            bool? handled = await _projectContextHost.OpenContextForWriteAsync(accessor =>
             {
                 var errorReporter = (IVsLanguageServiceBuildErrorReporter2)accessor.HostSpecificErrorReporter;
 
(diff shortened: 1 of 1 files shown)


dmonroym (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
```diff
@@ -54,7 +54,7 @@ private async Task<AddMessageResult> AddMessageCoreAsync(TargetGeneratedError er
             if (!TryExtractErrorListDetails(error.BuildEventArgs, out ErrorListDetails details) || string.IsNullOrEmpty(details.Code))
                 return AddMessageResult.NotHandled;
 
-            bool handled = await _projectContextHost.OpenContextForWriteAsync(accessor =>
+            bool? handled = await _projectContextHost.OpenContextForWriteAsync(accessor =>
```
drewnoakes, line 57: @ocallesp was there a warning here before? The method appears to return `Task<T?>`, so I would have expected the old code to fail the build.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
```diff
@@ -78,7 +78,7 @@ private async Task<AddMessageResult> AddMessageCoreAsync(TargetGeneratedError er
                 return TaskResult.False;
             });
 
-            return handled ? AddMessageResult.HandledAndStopProcessing : AddMessageResult.NotHandled;
+            return handled ?? false ? AddMessageResult.HandledAndStopProcessing : AddMessageResult.NotHandled;
```
MiYanni, line 81: I'm confused by this. It says, "If `handled` is not null, return `handled`. If `handled` is null, run the ternary. The ternary asks, "I'm always `false`, so return `AddMessageResult.NotHandled`".
This code is equivalent to:
```suggestion
            return handled ?? AddMessageResult.NotHandled;
```

PR #7099, opened 2021-04-08
Localized file check-in by OneLocBuild Task: Build definition ID 981: Build ID 1077211

MiYanni:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7099)


src/Microsoft.VisualStudio.Editors/ResourceEditor/xlf/OpenFileWarningDialog.zh-Hant.xlf
```diff
@@ -1,37 +1,37 @@
...
-        <note />
-      </trans-unit>
-      <trans-unit id="$this.Text">
-        <source>Open File</source>
-        <target state="translated">開啟檔案</target>
-        <note />
-      </trans-unit>
-    </body>
-  </file>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
drewnoakes, line 1: What is the actual difference in these files? I don't think these should all be displaying diffs.

PR #7104, opened 2021-04-09
Add properties to support CPS VS hierarchy queries

### LINKED ISSUE #7103: Include CmdUIContextGuid to the ConfigurationGeneral rule of the NET Core project
CPS gets the value of this property for VSHPROPID_CmdUIGuid. It prefers to use the value in the project snapshot. Unfortunately, because this is not defined in the NET Core project rule. It won't find it, and will fall back to a backup rule and uses the live project.

When thread pool is busy, or project hasn't been evaluated, this evaluation can be blocked, and leads into UI delays.  Due to this blocks context menu, it can slowdown the user experience especially during branch switching time.

This should be a straight change to prevent this UI delay issue.

+ @davkean

drewnoakes:
Fixes #7103

CPS's `ProjectNode` handles VS hierarchy requests on behalf of the project. For several of those hierarchy properties, the value comes from the project evaluation.

In order for CPS to retrieve the evaluated value, the corresponding MSBuild property must be defined in our ConfigurationGeneral.xaml rule file.

This commit adds several missing properties so that CPS can provide the correct value in response to hierarchy queries.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7104)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
@@ -46,6 +46,10 @@
   <StringProperty Name="BaseIntermediateOutputPath"
                   Visible="False" />
 
+  <StringProperty Name="CmdUIContextGuid"
+                  ReadOnly="True"
+                  Visible="False" />
+
   <StringProperty Name="Configuration"
                   Visible="False" />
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -207,6 +223,14 @@
     </BoolProperty.DataSource>
   </BoolProperty>
 
+  <StringProperty Name="SupportedMyApplicationTypes"
+                  ReadOnly="True"
+                  Visible="False" />
+
+  <StringProperty Name="SupportedOutputTypes"
```
davkean, line 230: CPS defines this as a StringListProperty.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ConfigurationGeneral.xaml
```diff
@@ -58,6 +62,18 @@
   <StringProperty Name="DefaultPlatform"
                   Visible="False" />
 
+  <StringProperty Name="DesignerFunctionVisibility"
```
davkean, line 65: These should match the CPS definition: https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2Fshared%2FRules%2Fgeneral.xaml&version=GBmain&line=101&lineEnd=102&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

drewnoakes: Thanks, updated.

davkean (approved):

davkean:
@drewnoakes This resolves https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1274309.

Issue #6985, opened 2021-02-24
Update localization infrastructure


tmeschter:
**Summary**:

Our internal localization team is updating their infrastructure, and this requires that we make some changes to our repo and our build process.

More information can be found at the [Migrating out of SimpleLoc](https://ceapex.visualstudio.com/CEINTL/_wiki/wikis/CEINTL.wiki/1481/Migrating-out-of-SimpleLoc) internal wiki page.

This work needs to be completed by Mar. 31st, 2021 or we will lose the ability to update our translations.

tmeschter:
I believe @jonfortescue is handling related work for repos that make use of Arcade, and may be able to provide some help if we get stuck.

drewnoakes:
This PR may be a useful reference (or may not be!) https://github.com/microsoft/vs-validation/pull/89

jjmew:
@dmonroym and @MiYanni  this is something that is time sensitive. Can you take a look at this next week?

PR #7108, opened 2021-04-09
Add OneLocBuild process and pipeline for localization

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/6985

NOTE: When reviewing this PR, I highly recommend you `Hide whitespace changes` and use the `File filter` to ignore `.lcl` files.
image

To preface, I'll end up scheduling a meeting to go over these things. The TLDR is:
- Adds a pipeline, `one-loc-build`, which will run on PR merge, similar to our official builds (the ones that create signed packages).
- Adds a project, `OneLocBuildSetup.csproj`, which runs as part of that pipeline.
  - It creates language-neutral named XLF files out of existing XLF files (via copy/paste). I use the Spanish (`es`) files for this purpose.
  - It generates a `LocProject.json` file which drives the OneLocBuild task in the pipeline.
- Adds `.lcl` files to the repo. These files hold the translation info for the OneLocBuild service.
- Put the VSSDK packages into a conditional ItemGroup so they can be excluded from `OneLocBuildSetup.csproj`.
  - They are set to `Include` and don't have .NET 5 packages, and they are not needed for the project.

An example of a successful pipeline run can be found here: https://dev.azure.com/dnceng/public/_build/results?buildId=1080894&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=12f1170f-54f2-53f3-20dd-22fc7dff55f9
The artifacts published by the pipeline are necessary for the OneLocBuild process. When the pipeline runs, it will create a PR with changes to the XLF files based on the translations. An example PR can be seen here: https://github.com/dotnet/project-system/pull/7107
The artifacts are consumed by the OneLocBuild service independent of the pipeline. At some point, the OneLocBuild service will create another PR itself in the repo that will have updates to the LCL files. These are the translation updates. 

Currently, we use XLF files produced by the `RoslynTools.RepoToolset` package. This package uses the `XliffTasks` package that generates the XLF files on-build. These files are simply an intermediary format for translations. In the not-too-distant future, we will be changing this process to no longer produce those files. The OneLocBuild process is currently using the XLF files to do translations, but would change to use the raw files (`.resx`, `.xaml`, `.vsct`) files. The support for `.vsct` files in currently in the works and we would change our process when that becomes stable.

This localization process had multiple iterations. One of the (more annoying) aspects is that the OneLocBuild needs a language-neutral named XLF file. Meaning, it cannot have a language in the filename (like `.es`, `.de`, `.fr`, etc). This file becomes the 'structure' needed to do the translation process. I don't know why this restriction is in place, but I tried different ways of doing this. I eventually arrived at creating the `OneLocBuildSetup.csproj` to do this. The other reason to create this code was because we needed to generate a `LocProject.json` file.

When learning this stuff, I originally wrote the `LocProject.json` file by-hand. At first, I thought this file was supposed to target the original source of the translations. It wasn't obvious to me that we were going to use the XLF files for this, so I originally did the `.resx`, `.xaml`, and `.vsct` files in the `LocProject.json`. This didn't work because `.vsct` is currently unsupported. When it is supported in the future, we will no longer create and use XLF files. After changing the `LocProject.json` to target XLF files and making sure the entire process 'worked', I created a way to generate this file since it was a simple structure. This file is produced in the pipeline and is not checked in. It is included in the build artifacts in the pipeline.

I ran into issues originally because I did not have the LCL files in this repo. The OneLocBuild task in the pipeline needs these files to do translations. These LCL files originally were in a separate repo entirely, had their own pipeline to create a NuGet package, and then had this pipeline acquire and consume the NuGet package. The issue was, the file path created from extracting these files was exceeding Windows path length. The bug specifically was that the directory path length was less than 248 characters (max directory length on Windows), but equal to/above max file path length (260 on Windows). This caused the entire process to fail since it couldn't copy the extracted LCL files to their proper destination for the OneLocBuild task to use.

I felt having a separate repo and package for only a single purpose was a bit overkill. So instead, I checked the LCL files into this repo and the OneLocBuild task just uses them from there. It saves time when running this process and avoids the path length issues (since there is no package extraction and copying involved). The main downside is we'll get PRs from the system for localization updates. But it does keep visibility (for us) on these updates. When they were in a separate repo, we wouldn't have that visibility.

Overall, it is a bit of a roundabout process, but it is what we'll be using going forward. Again, I'll explain this stuff and answer questions in a meeting. Before merging this PR, I will update the pipeline itself to target `main` and have it run on PR merges. After that, a PR should be created similar to my example PR, and we'll merge that into the repo.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7108)


drewnoakes:
> I highly recommend you Hide whitespace changes and use the File filter to ignore .lcl files.

You might try adding an entry for `.lcl` files in the `.gitattributes` file, similar to the one we have for `.xlf` files:

https://github.com/dotnet/project-system/blob/448b726d648bd9d458f8a982f91eae20585ce07c/.gitattributes#L8-L9

jjmew (approved):

build/loc/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+// ReSharper disable SuggestVarOrType_BuiltInTypes
+// ReSharper disable SuggestVarOrType_Elsewhere
+
+namespace OneLocBuildSetup
+{
+    [SuppressMessage("Style", "IDE0008:Use explicit type")]
+    internal class Program
+    {
+        private const string XlfExtension = ".xlf";
+        private const string SpanishXlfExtension = ".es" + XlfExtension;
```
swesonga, line 19: >SpanishXlfExtension [](start = 29, length = 19)

what was special about Spanish? a comment here would help.

build/loc/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public static void Main(string[] args) => Parser.Default.ParseArguments<Arguments>(args).WithParsed(RunSetup);
+
+        private static void RunSetup(Arguments args)
```
swesonga, line 23: >RunSetup [](start = 28, length = 8)

minor suggestion: I think having comments about what this code is doing (similar to the PR description) here in the code would be great so that we don't need to read the commit history to understand what the goal of this code is.

build/loc/Program.cs
MiYanni: There is nothing special about Spanish. It could be any language. I just picked one at random by writing every language onto slips of paper, putting them into a hat, and then pulling one out at random... or not. 😝 The OneLocBuild system uses the structure of the file, not the content.

build/loc/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+                .OrderBy(pg => pg.Key)
+                .Select(pg => new Project(pg
+                    .Select(p => CreateLocItem(p.xlfPath))
+                    .OrderBy(li => li.SourceFile)
+                    .ToArray()))
+                .ToArray());
+
+            var locPath = Path.Combine(args.OutputPath, "loc");
+            Directory.CreateDirectory(locPath);
+            var locProjectPath = Path.Combine(locPath, "LocProject.json");
```
swesonga, line 48: >"LocProject.json" [](start = 55, length = 17)

nit: consider using class constants for these paths

build/loc/Program.cs
swesonga: Makes sense, this would be great to mention in the code though otherwise future maintainers might expect there to be something magical about ".es"

---
In reply to: [613569792](https://github.com/dotnet/project-system/pull/7108#discussion_r613569792) [](ancestors = 613569792)

swesonga:
@MiYanni Which repo contained the LCL files that are now moving into this one?

swesonga (approved):

MiYanni:
> Which repo contained the LCL files that are now moving into this one?

@swesonga Originally was this one: https://dev.azure.com/ceapex/CEINTL/_git/LCL-JUNO-PROD-PRJSYSTEM
Then was this one to test the path length issue (this one is now disabled): https://dev.azure.com/ceapex/CEINTL/_git/LCL-JUNO-PROD-PRJSYS
The pipelines are:
- https://dev.azure.com/ceapex/CEINTL/_build?definitionId=3491
- https://dev.azure.com/ceapex/CEINTL/_build?definitionId=3526

The packages were published here: https://dev.azure.com/ceapex/CEINTL/_packaging?_a=feed&feed=LCL

build/loc/Program.cs
MiYanni: Added comments above this and link for explanation.

build/loc/Program.cs
MiYanni: Added constants for the folder names and file name.

build/loc/Program.cs
MiYanni: Split each process into separate methods and added comments with links for what each process does.

build/loc/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using CommandLine;
+
+// ReSharper disable SuggestVarOrType_BuiltInTypes
```
dmonroym, line 11: nit remove these comments

MiYanni: Removed comments. Used `SuppressMessage` attribute for all suppressions to be consistent.

PR #7109, opened 2021-04-10
Use most up to date PIA Packages

dmonroym:
Updates packages and adapts code to them.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7109)


dmonroym:
We use a private feed so these tests builds are expected to fail. Packages should shortly be signed and pushed to a public feed

PR #7115, opened 2021-04-12
Restore tools from nuget.config

dmonroym:
We are currently using private feeds as workarounds for dev 17. Our current restore feed has a feed specified which doesn't include vssdk-unifiedPIAs. By using this task we can specify which feeds to use in the restore.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7115)


dmonroym:
unit-tests are expected to fail as they'll attempt to read from private feed. This update is only meant to work on the official build.

Internal users can validate with a successful run here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4645793&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9

PR #7118, opened 2021-04-13
Document the use of monospace fonts in multi-line string editors

drewnoakes:
Adds documentation for the capability added in https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/317111.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7118)

### DIFF
--- a/docs/repo/property-pages/property-specification.md
+++ b/docs/repo/property-pages/property-specification.md
@@ -145,13 +145,27 @@ Most `StringProperty` properties are expected to have short values that fit on o
 If a property is expected to have multiple lines of text, the editor should be changed to `MultiLineString`. For example:
 
 ```xml
-<StringProperty Name="Description"
-                DisplayName="Description"
-                Description="A description of the package for UI display."
-                HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147238"
-                Category="General">
+<StringProperty Name="MyMultiLineProperty"
+                DisplayName="A Multi-line Property"
+                Description="A property that may contain multiple lines of text.">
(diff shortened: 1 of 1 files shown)


docs/repo/property-pages/property-specification.md
```diff
@@ -145,17 +145,31 @@ Most `StringProperty` properties are expected to have short values that fit on o
...
+<StringProperty Name="MyMultiLineProperty"
+                DisplayName="A Multi-line Property"
+                Description="A property that may contain multiple lines of text.">
   <StringProperty.ValueEditors>
     <ValueEditor EditorType="MultiLineString" />
   </StringProperty.ValueEditors>
 </StringProperty>
 ```
 
+If the text within this editor should be displayed with a monospace (fixed width) font, as would be common for code or other text which might be expected to align by column, add `UseMonospaceFont` following metadata with a value of `True`:
```
MiYanni, line 157: What fonts are these fonts? Like, both the monospace and non-monospace. Not sure if it needs mentioned in the docs, but I'm interested. I'm sure someone else is too.

MiYanni (approved):

melytc (approved):

docs/repo/property-pages/property-specification.md
drewnoakes: Ideally it would be the default editor font. Until I work out how to do that, it's Consolas.

PR #7121, opened 2021-04-13
Select statement failure in StartupObjectEnumGenerator

MiYanni:
Fixes: [AB#1280068](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1280068)

Combines the select statements and makes them aware of possible null `entryPoints`, which is likely the cause of this failure.


drewnoakes (approved):

jjmew (approved):
:shipit:

swesonga (approved):

PR #7122, opened 2021-04-14
Use `ProduceReferenceAssembly` for faster incremental builds

drewnoakes:
This produces a `ref` folder (e.g. `bin/Debug/ref/Project.dll`) that contains the surface API of the project. Other projects which reference that assembly may avoid rebuilding themselves when the referenced project's surface area did not change, even if its internals did. This can help prevent cascading project builds when simple changes are made at low levels in the dependency graph.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7122)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -36,6 +36,8 @@
 
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <ForceGenerationOfBindingRedirects>true</ForceGenerationOfBindingRedirects>
+
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
   </PropertyGroup>
 
   <!-- 


KirillOsenkov:
One thing I just ran into is how do you ensure that if B has a P2P to A, and A changes but the A ref asm doesn’t change, how do you still copy the A impl dll into B output to make sure it’s not stale? Since B doesn’t build, who does the copying?

KirillOsenkov:
This is not an issue if all projects share the same common output directory.

drewnoakes:
We are using a single output directory here.

However I tested this in another solutionwhich does not use a single output directory and it _didn't not_ work. I changed some core code, then ran a unit test in another project and the test failed. Then, reverted the change and ran the test, and it passed. The output window does make it look like the test project built, though it _seemed_ faster. I will check the logs next time I'm working in that solution.

swesonga (approved):

MiYanni (approved):
I like this change, but should this go into the `src/Directory.Build.props` instead? There are projects in other non-src folders that this change doesn't really apply to. And the test projects shouldn't need this either, right?

drewnoakes:
@MiYanni it has value for tests too. There is at least one library in the test folder (on a phone now).

PR #7123, opened 2021-04-14
Add the DebugType property to control debug symbol generation

drewnoakes:
Relates to #6720

Adds a drop down list with options that controls how debug symbols are omitted.

There are currently no good docs on the `DebugType` variable to link to. I've left a TODO to come back to that, as we should do a pass to review all the help links before GA.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7123)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -98,7 +98,8 @@
                 HelpUrl="https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/CommandLine.md"
                 Category="General">
     <EnumProperty.DataSource>
-      <DataSource HasConfigurationCondition="False" />
+      <DataSource HasConfigurationCondition="False"
+                  Persistence="ProjectFileWithInterception" />
     </EnumProperty.DataSource>
     <EnumValue Name="none"
                DisplayName="No symbols are emitted" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -91,6 +91,30 @@
     </BoolProperty.Metadata>
   </BoolProperty>
 
+  <!-- TODO create fwlink for this HelpUrl, and find or create a better documentation source -->
+  <EnumProperty Name="DebugType"
+                DisplayName="Debug symbols"
+                Description="Specifies the kind of debug symbols produced during build."
+                HelpUrl="https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/CommandLine.md"
+                Category="General">
+    <EnumProperty.DataSource>
+      <DataSource HasConfigurationCondition="False" />
+    </EnumProperty.DataSource>
+    <EnumValue Name="none"
+               DisplayName="No symbols are emitted" />
+    <!--
+    Note that 'pdbonly' is the same as 'full'.
+    <EnumValue Name="pdbonly"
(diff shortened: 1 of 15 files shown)


davkean:
"Current platform" should probably be "Windows only".

drewnoakes:
> "Current platform" should probably be "Windows only".

On Windows it makes Windows-only. On other platforms, it's portable.

davkean:
> On Windows it makes Windows-only. On other platforms, it's portable.

That wasn't the case in the .NET Core 1.x era, has this changed? How did you figure this out?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -91,6 +91,30 @@
...
+                HelpUrl="https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/CommandLine.md"
+                Category="General">
+    <EnumProperty.DataSource>
+      <DataSource HasConfigurationCondition="False" />
+    </EnumProperty.DataSource>
+    <EnumValue Name="none"
+               DisplayName="No symbols are emitted" />
+    <!--
+    Note that 'pdbonly' is the same as 'full'.
+    <EnumValue Name="pdbonly"
```
davkean, line 107: This might show up by default for the web SDK, if I'm reading this correctly.

https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.props#L18

drewnoakes:
> That wasn't the case in the .NET Core 1.x era, has this changed? How did you figure this out?

Based on this: https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/CommandLine.md

> This might show up by default for the web SDK, if I'm reading this correctly.

Perhaps we intercept this and convert the value for the UI.

davkean:
I would call it out like the compiler does, to make it clear that its Windows PDB on windows, and Portable everywhere else.

drewnoakes:
I considered that but wanted to keep the text concise. I wonder if radio buttons would be a better fit here. There are a few of these properties where items in the combo box would be best served with descriptions. Another option is to describe all the options in the property's description, but that's not great either.

What text would you use for the item?

MiYanni:
If you were to have a combo box with description, would the description update as you select different combo box items? The description would only reflect the current item selected, so each item would have a description bound to it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
drewnoakes: Good find. I'll look into converting `pdbonly` into `full` on the way from the project file to the UI.

I also filed https://github.com/dotnet/sdk/pull/16927 to change the web SDK's default `DebugType`.

drewnoakes:
> If you were to have a combo box with description, would the description update as you select different combo box items? The description would only reflect the current item selected, so each item would have a description bound to it.

We could do that. The challenge there is that you'd have to select the item before you knew what it does. The fear of "pressing a wrong button" would put some people off. If we want to show lengthier descriptions alongside options, then radio buttons might be a better fit than a drop down list.

In this instance I'm happy enough with the state of things. I'd like the help button to link to more useful documentation however.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
drewnoakes: I added interception that allows the UI to read `pdbonly` and display the equivalent value as `full` without having duplicate entries in the drop down. The project file remains unchanged unless the user changes the selection.

Issue #7025, opened 2021-03-15
Properly address property pages item ordering


MiYanni:
Related: https://github.com/dotnet/project-system/issues/6989

Currently, CPS does not seem to maintain item ordering. This can be seen in the Target Framework dropdown. For now, we are putting in a hotfix to manually order these items. However, the proper fix has a couple options:

1. Investigate and fix CPS item ordering preservation
  a. For target framework, the items are declared here and this ordering should be preserved: https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedTargetFrameworks.props
2. Add functionality to the Rule definitions as to allow custom ordering.

These two options could both be completed, where the default is preserved ordering and it is optional to add your own ordering to the Rule definitions. When this PR is completed, please revert the hotfix here:

tmeschter:
A third option would be to update Microsoft.NET.SupportedTargetFrameworks.props to include explicit ordering information separate from the order of items in the .props.

tmeschter:
Fourth option: parse the Include for each item from that .props (e.g. `.NETCoreApp,Version=v2.1`) to group items by the framework family (e.g. .NETCoreApp) and then order by version number within that family.

drewnoakes:
I would vote for adding specific `DisplayOrder` metadata to the items returned by the SDK.

melytc:
> Fourth option: parse the Include for each item from that .props (e.g. `.NETCoreApp,Version=v2.1`) to group items by the framework family (e.g. .NETCoreApp) and then order by version number within that family.

To avoid parsing, this is "already done" in the SDK side: instead of capturing all of the values in SupportedTargetFramework item, we could store and use each framework's ItemGroup [Microsoft.NET.SupportedTargetFrameworks.props](https://github.com/sfoslund/sdk/blob/36876246b7dd181fe6d69d7fa68ba454c916aed4/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedTargetFrameworks.props).

drewnoakes:
CPS now has a way for us to obtain the original item ordering, which may fix this.

We'll need code such as:

```c#
var after = projectChangeDescription.After.Items as IDataWithOriginalSource<KeyValuePair<string, IImmutableDictionary<string, string>>>;
```

This gives an `ICollection<KeyValuePair<string, IImmutableDictionary<string, string>>>` which has the original MSBuild order.

[...]

PR #7132, opened 2021-04-20
Introduce target platform, platform version and minimum version to the Properties Pages.

melytc:
We are introducing three properties into the Project Properties UI: `TargetPlatfiorm`, `TargetPlatformVersion` and `MinimumPlatformVersion`.

#### `TargetPlatform`
The supported platforms, or OS, the project can target, i.e. Windows, iOS, Android. The SDK has worked to supply these values through the `SdkSupportedTargetPlatformIdentifier`, which will be produced on evaluation (https://github.com/dotnet/sdk/pull/16208). As a way of testing this, I added a couple of these items on the project file to mock up the behavior.
image

#### `TargetPlatformVersion`
The supporting platform versions the project can target, i.e. when Windows is selected, will get 7.0, 8.0, 10.0.x. The SDK provides these values when a target platform is selected, and does that through the `SdkSupportedTargetPlatformVersion` item.
As it stands now, we have to close the Project Properties UI when selecting a platform to be able to see the supported platform versions.
image

#### `MinimumPlatformVersion`
The minimum platform version the project will support, i.e. when Windows is selected, will get 7.0, 8.0, 10.0.x.
As in the case of `TargetOSVersion`, the SDK provides these values when a target platform is selected and uses the same values from the `SdkSupportedTargetPlatformVersion` item.
image

Notes:
- Feedback on wording (i.e. descriptions) is appreciated.

Missing details:
- [x] Add conditional visibility to `TargetOSVersion` and `MinimumOSVersion` so that they are visible only if a platform is selected.
- [x] Scenario: provide a way for the user to unselect a value from these 3 properties.
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7132)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
@@ -4,14 +4,8 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.Composition;
-using System.Linq;
-using System.Threading.Tasks;
-using System.Threading.Tasks.Dataflow;
 using Microsoft.Build.Framework.XamlTypes;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
-using Microsoft.VisualStudio.Threading;
-using EnumCollection = System.Collections.Generic.ICollection<Microsoft.VisualStudio.ProjectSystem.Properties.IEnumValue>;
-using EnumCollectionProjectValue = Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<System.Collections.Generic.ICollection<Microsoft.VisualStudio.ProjectSystem.Properties.IEnumValue>>;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
 {
@@ -20,67 +14,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
     /// </summary>
     [ExportDynamicEnumValuesProvider("SdkSupportedTargetPlatformIdentifierEnumProvider")]
     [AppliesTo(ProjectCapability.DotNet)]
-    internal class SdkSupportedTargetPlatformIdentifierProvider : ChainedProjectValueDataSourceBase<EnumCollection>, IDynamicEnumValuesProvider, IDynamicEnumValuesGenerator
+    internal class SdkSupportedTargetPlatformIdentifierProvider : SupportedValuesProvider
     {
-        private readonly IProjectSubscriptionService _subscriptionService;
+        protected override string RuleName => SdkSupportedTargetPlatformIdentifier.SchemaName;
 
         [ImportingConstructor]
         public SdkSupportedTargetPlatformIdentifierProvider(
             ConfiguredProject project,
             IProjectSubscriptionService subscriptionService)
-            : base(project, synchronousDisposal: true, registerDataSource: false)
-        {
-            _subscriptionService = subscriptionService;
-        }
+            : base(project, subscriptionService, useNoneValue: true) { }
 
-        [ConfiguredProjectAutoLoad]
-        [AppliesTo(ProjectCapability.DotNet)]
-        public void Load()
-        {
-            // To avoid UI delays when opening the AppDesigner for the first time, 
-            // we auto-load so that we are included in the first design-time build
-            // for the project.
-            EnsureInitialized();
-        }
-
-        protected override IDisposable? LinkExternalInput(ITargetBlock<EnumCollectionProjectValue> targetBlock)
-        {
-            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _subscriptionService.ProjectRuleSource;
-
-            // Transform the changes from evaluation -> supported target OS
-            DisposableValue<ISourceBlock<EnumCollectionProjectValue>> transformBlock = source.SourceBlock.TransformWithNoDelta(
-                update => update.Derive(Transform),
-                suppressVersionOnlyUpdates: false,
-                ruleNames: SdkSupportedTargetPlatformIdentifier.SchemaName);
-
-            // Set the link up so that we publish changes to target block
-            transformBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
-
-            // Join the source blocks, so if they need to switch to UI thread to complete 
-            // and someone is blocked on us on the same thread, the call proceeds
-            JoinUpstreamDataSources(source);
-
-            return transformBlock;
-        }
-
-        private static EnumCollection Transform(IProjectSubscriptionUpdate input)
-        {
-            IProjectRuleSnapshot snapshot = input.CurrentState[SdkSupportedTargetPlatformIdentifier.SchemaName];
-
-            var list = new List<IEnumValue>(capacity: 1 + snapshot.Items.Count);
-
-            list.Add(new PageEnumValue(new EnumValue()
-            {
-                Name = string.Empty,
-                DisplayName = Resources.Property_NoneValue
-            }));
-
-            list.AddRange(snapshot.Items.Select(ToEnumValue));
-
-            return list.OrderBy(e => e.DisplayName).ToArray();
-        }
-
-        private static IEnumValue ToEnumValue(KeyValuePair<string, IImmutableDictionary<string, string>> item)
+        protected override IEnumValue ToEnumValue(KeyValuePair<string, IImmutableDictionary<string, string>> item)
         {
             return new PageEnumValue(new EnumValue()
             {
@@ -92,23 +36,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
             });
         }
 
-        public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>? options)
+        protected override int SortValues(IEnumValue a, IEnumValue b)
         {
-            return Task.FromResult<IDynamicEnumValuesGenerator>(this);
+            return StringComparer.OrdinalIgnoreCase.Compare(a.DisplayName, b.DisplayName);
         }
-
-        public async Task<EnumCollection> GetListedValuesAsync()
-        {
-            using (JoinableCollection.Join())
-            {
-                EnumCollectionProjectValue snapshot = await SourceBlock.ReceiveAsync();
-
-                return snapshot.Value;
-            }
-        }
-
-        bool IDynamicEnumValuesGenerator.AllowCustomValues => false;
-
-        Task<IEnumValue?> IDynamicEnumValuesGenerator.TryCreateEnumValueAsync(string userSuppliedValue) => TaskResult.Null<IEnumValue>();
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
@@ -67,7 +67,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
         {
             IProjectRuleSnapshot snapshot = input.CurrentState[SdkSupportedTargetPlatformIdentifier.SchemaName];
 
-            var list = new List<IEnumValue>();
+            var list = new List<IEnumValue>(capacity: 1 + snapshot.Items.Count);
 
             list.Add(new PageEnumValue(new EnumValue()
             {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
@@ -67,8 +67,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
         {
             IProjectRuleSnapshot snapshot = input.CurrentState[SdkSupportedTargetPlatformIdentifier.SchemaName];
 
-            return snapshot.Items.Select(ToEnumValue)
-                                    .ToList();
+            var list = new List<IEnumValue>();
+
+            list.Add(new PageEnumValue(new EnumValue()
+            {
+                Name = string.Empty,
+                DisplayName = Resources.Property_NoneValue
+            }));
+
+            list.AddRange(snapshot.Items.Select(ToEnumValue));
+
+            return list;
         }
 
         private static IEnumValue ToEnumValue(KeyValuePair<string, IImmutableDictionary<string, string>> item)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
@@ -1,0 +1,126 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+using Microsoft.Build.Framework.XamlTypes;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.Threading;
+using EnumCollection = System.Collections.Generic.ICollection<Microsoft.VisualStudio.ProjectSystem.Properties.IEnumValue>;
+using EnumCollectionProjectValue = Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<System.Collections.Generic.ICollection<Microsoft.VisualStudio.ProjectSystem.Properties.IEnumValue>>;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
+{
+    /// <summary>
+    ///     Responsible for producing valid values for the SdkSupportedTargetPlatformIdentifier property from a design-time build.
+    /// </summary>
+    [ExportDynamicEnumValuesProvider("SdkSupportedTargetPlatformIdentifierEnumProvider")]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class SdkSupportedTargetPlatformIdentifierProvider : ChainedProjectValueDataSourceBase<EnumCollection>, IDynamicEnumValuesProvider, IDynamicEnumValuesGenerator
+    {
+        private readonly IProjectSubscriptionService _subscriptionService;
+
+        [ImportingConstructor]
+        public SdkSupportedTargetPlatformIdentifierProvider(
+            ConfiguredProject project,
+            IProjectSubscriptionService subscriptionService)
+            : base(project, synchronousDisposal: true, registerDataSource: false)
+        {
+            _subscriptionService = subscriptionService;
+
+            ReadyToBuild = new OrderPrecedenceImportCollection<IConfiguredProjectReadyToBuild>(projectCapabilityCheckProvider: project);
+        }
+
+        [ImportMany]
+        public OrderPrecedenceImportCollection<IConfiguredProjectReadyToBuild> ReadyToBuild
+        {
+            get;
+        }
+
+        [ConfiguredProjectAutoLoad]
+        [AppliesTo(ProjectCapability.DotNet)]
+        public void Load()
+        {
+            // To avoid UI delays when opening the AppDesigner for the first time, 
+            // we auto-load so that we are included in the first design-time build
+            // for the project.
+            EnsureInitialized();
+        }
+
+        protected override IDisposable? LinkExternalInput(ITargetBlock<EnumCollectionProjectValue> targetBlock)
+        {
+            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _subscriptionService.ProjectRuleSource;
+
+            // Transform the changes from design-time build -> supported target OS
+            DisposableValue<ISourceBlock<EnumCollectionProjectValue>> transformBlock = source.SourceBlock.TransformWithNoDelta(
+                update => update.Derive(Transform),
+                suppressVersionOnlyUpdates: false,
+                ruleNames: SdkSupportedTargetPlatformIdentifier.SchemaName);
+
+            // Set the link up so that we publish changes to target block
+            transformBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+
+            // Join the source blocks, so if they need to switch to UI thread to complete 
+            // and someone is blocked on us on the same thread, the call proceeds
+            JoinUpstreamDataSources(source);
+
+            return transformBlock;
+        }
+
+        private static EnumCollection Transform(IProjectSubscriptionUpdate input)
+        {
+            IProjectRuleSnapshot snapshot = input.CurrentState[SdkSupportedTargetPlatformIdentifier.SchemaName];
+
+            return snapshot.Items.Select(ToEnumValue)
+                                    .ToList();
+        }
+
+        private static IEnumValue ToEnumValue(KeyValuePair<string, IImmutableDictionary<string, string>> item)
+        {
+            return new PageEnumValue(new EnumValue()
+            {
+                // Example: <SdkSupportedTargetPlatformIdentifier Include="windows" DisplayName="Windows"/>
+                //          <SdkSupportedTargetPlatformIdentifier Include="ios" DisplayName="iOS"/>
+
+                Name = item.Key,
+                DisplayName = item.Value[SdkSupportedTargetPlatformIdentifier.DisplayNameProperty]
+            });
+        }
+
+        public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>? options)
+        {
+            return Task.FromResult<IDynamicEnumValuesGenerator>(this);
+        }
+
+        public async Task<EnumCollection> GetListedValuesAsync()
+        {
+            if (!IsReadyToBuild())
+                throw new InvalidOperationException("This configuration is not set to build");
+
+            // NOTE: This has a race, if called off the UI thread, the configuration could become 
+            // inactive underneath us and hence not ready for build, causing below to block forever.
+
+            using (JoinableCollection.Join())
+            {
+                EnumCollectionProjectValue snapshot = await SourceBlock.ReceiveAsync();
+
+                return snapshot.Value;
+            }
+        }
+
+        private bool IsReadyToBuild()
+        {
+            IConfiguredProjectReadyToBuild? readyToBuild = ReadyToBuild.FirstOrDefault()?.Value;
+
+            return readyToBuild?.IsValidToBuild == true;
+        }
+
+        bool IDynamicEnumValuesGenerator.AllowCustomValues => false;
+
+        Task<IEnumValue?> IDynamicEnumValuesGenerator.TryCreateEnumValueAsync(string userSuppliedValue) => TaskResult.Null<IEnumValue>();
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
@@ -16,7 +16,7 @@ using EnumCollectionProjectValue = Microsoft.VisualStudio.ProjectSystem.IProject
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
 {
     /// <summary>
-    ///     Responsible for producing valid values for the SdkSupportedTargetPlatformIdentifier property from a design-time build.
+    ///     Responsible for producing valid values for the SdkSupportedTargetPlatformIdentifier property from evaluation.
     /// </summary>
     [ExportDynamicEnumValuesProvider("SdkSupportedTargetPlatformIdentifierEnumProvider")]
     [AppliesTo(ProjectCapability.DotNet)]
@@ -31,14 +31,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
             : base(project, synchronousDisposal: true, registerDataSource: false)
         {
             _subscriptionService = subscriptionService;
-
-            ReadyToBuild = new OrderPrecedenceImportCollection<IConfiguredProjectReadyToBuild>(projectCapabilityCheckProvider: project);
-        }
-
-        [ImportMany]
-        public OrderPrecedenceImportCollection<IConfiguredProjectReadyToBuild> ReadyToBuild
-        {
-            get;
         }
 
         [ConfiguredProjectAutoLoad]
@@ -55,7 +47,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
         {
             IProjectValueDataSource<IProjectSubscriptionUpdate> source = _subscriptionService.ProjectRuleSource;
 
-            // Transform the changes from design-time build -> supported target OS
+            // Transform the changes from evaluation -> supported target OS
             DisposableValue<ISourceBlock<EnumCollectionProjectValue>> transformBlock = source.SourceBlock.TransformWithNoDelta(
                 update => update.Derive(Transform),
                 suppressVersionOnlyUpdates: false,
@@ -98,25 +90,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
 
         public async Task<EnumCollection> GetListedValuesAsync()
         {
-            if (!IsReadyToBuild())
-                throw new InvalidOperationException("This configuration is not set to build");
-
-            // NOTE: This has a race, if called off the UI thread, the configuration could become 
-            // inactive underneath us and hence not ready for build, causing below to block forever.
-
             using (JoinableCollection.Join())
             {
                 EnumCollectionProjectValue snapshot = await SourceBlock.ReceiveAsync();
 
                 return snapshot.Value;
             }
-        }
-
-        private bool IsReadyToBuild()
-        {
-            IConfiguredProjectReadyToBuild? readyToBuild = ReadyToBuild.FirstOrDefault()?.Value;
-
-            return readyToBuild?.IsValidToBuild == true;
         }
 
         bool IDynamicEnumValuesGenerator.AllowCustomValues => false;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
@@ -77,7 +77,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
 
             list.AddRange(snapshot.Items.Select(ToEnumValue));
 
-            return list;
+            return list.OrderBy(e => e.DisplayName).ToArray();
         }
 
         private static IEnumValue ToEnumValue(KeyValuePair<string, IImmutableDictionary<string, string>> item)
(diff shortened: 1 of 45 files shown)


davkean:
Why the "OS" terminology? These all end up mapping to "TargetPlatformXXX" under the covers, right?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>? options)
+        {
+            return Task.FromResult<IDynamicEnumValuesGenerator>(this);
+        }
+
+        public async Task<EnumCollection> GetListedValuesAsync()
+        {
+            if (!IsReadyToBuild())
```
davkean, line 100: These are evaluations, these checks for or comments about "Build" are not needed.

melytc:
> Why the "OS" terminology? These all end up mapping to "TargetPlatformXXX" under the covers, right?

It came up in a meeting that calling it 'OS' in the UI might be more clear, but that's still open to changes. Also, there is already a "Platform target" property in the Build page that might confuse users (?), so I wanted to use something different.

drewnoakes:
We can add search aliases using the word "platform" so that these OS properties appear if you search "platform" too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+            IProjectSubscriptionService subscriptionService)
+            : base(project, synchronousDisposal: true, registerDataSource: false)
+        {
+            _subscriptionService = subscriptionService;
+
+            ReadyToBuild = new OrderPrecedenceImportCollection<IConfiguredProjectReadyToBuild>(projectCapabilityCheckProvider: project);
+        }
+
+        [ImportMany]
+        public OrderPrecedenceImportCollection<IConfiguredProjectReadyToBuild> ReadyToBuild
```
davkean, line 39: These are not needed as this is entirely evaluation (not build) based.

davkean:
It's called "Target Platform" everywhere else, to have the UI show different terms would result in other confusions.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+using EnumCollectionProjectValue = Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<System.Collections.Generic.ICollection<Microsoft.VisualStudio.ProjectSystem.Properties.IEnumValue>>;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Frameworks
+{
+    /// <summary>
+    ///     Responsible for producing valid values for the SdkSupportedTargetPlatformVersion property from a design-time build.
+    /// </summary>
+    [ExportDynamicEnumValuesProvider("SdkSupportedTargetPlatformVersionEnumProvider")]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class SdkSupportedTargetPlatformVersionProvider : ChainedProjectValueDataSourceBase<EnumCollection>, IDynamicEnumValuesProvider, IDynamicEnumValuesGenerator
```
davkean, line 23: There's a lot of common code here, between these two, perhaps introduce a base?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// <summary>
+    /// The complex Target Framework property consists of three UI-exposed properties: target framework, target OS and target OS version.
+    /// Those combine and produce a single value like this: [target framework]-[target OS][target OS version].
+    /// This class intercepts those values from the UI to saved them into the TargetFramework property, as well as retrive those values from that same property.
+    /// </summary>
+    [ExportInterceptingPropertyValueProvider(
+        new[]
+        {
+            TargetFrameworkProperty,
+            TargetOSProperty,
```
davkean, line 17: I think calling these different from the underlying values they represent; TargetPlatformIdentifier and TargetPlatformVersion is just resulting in confusion. While they get persisted in TargetFramework user property; they get translated to TargetPlatformXXX properties.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+            ComplexTargetFramework result = new ComplexTargetFramework();
+
+            if (storedComplexTargetFramework != null)
+            {
+                result.Platform = null;
+                result.PlatformVersion = null;
+            
+                if (storedComplexTargetFramework.IndexOf('-') != -1)
+                {
+                    result.TargetFramework = storedComplexTargetFramework.Split('-')[0];
```
davkean, line 134: The value 'windows-' will break this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <summary>
+        /// Takes the combined string and separates the values into the ComplextTargetFramework structure.
+        /// TargetFramework must always be present, but Platform and PlatformVersion can be null.
+        /// <para>i.e. net5.0-windows10.0 => { net5.0, windows, 10.0 }; netcoreapp3.1 => { netcoreapp3.1, null, null }</para>
+        /// </summary>
+        /// <param name="storedComplexTargetFramework"></param>
+        /// <returns></returns>
+        private static ComplexTargetFramework BreakDownValue(string? storedComplexTargetFramework)
+        {
+            ComplexTargetFramework result = new ComplexTargetFramework();
```
davkean, line 125: This string is a user property and should not be parsed; https://github.com/dotnet/project-system/blob/main/docs/repo/coding-conventions.md#data. Read the associated properties instead.

davkean: `TargetFramework` can contain `FooBar` and as long as the those previous properties are set, everything should just work.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+            ComplexTargetFramework values = BreakDownValue(storedComplexValue);
+            return await base.OnGetUnevaluatedPropertyValueAsync(propertyName, ExtractValue(values, propertyName) ?? "", defaultProperties);
+        }
+
+        //private static string? CommonCode(string propertyName, IProjectProperties defaultProperties)
+        //{
+        //    string? storedComplexValue = defaultProperties.GetUnevaluatedPropertyValueAsync(PropertyToBeSavedOnProjectFile); // TODO: change to TargetFrameworkProperty
+        //    ComplexTargetFramework values = BreakDownValue(storedComplexValue);
+        //    return ExtractValue(values, propertyName);
+        //}
```
MiYanni, line 79: Commented code.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <param name="platformVersion">The platform version value.</param>
+        /// <returns></returns>
+        private static string ComputeValue(string targetFramework, string? targetPlatform, string? platformVersion)
+        {
+            if (targetPlatform != null) // TODO: Ensure TargetFramework is not null.
+            {
+                return targetFramework + "-" + targetPlatform + platformVersion;
+            }
+            return targetFramework;
+        }
```
MiYanni, line 114: Could be written as:
```suggestion
        private static string ComputeValue(string targetFramework, string? targetPlatform, string? platformVersion) => targetPlatform != null ? $"{targetFramework}-{targetPlatform}{platformVersion}" : targetFramework;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+                return snapshot.Value;
+            }
+        }
+
+        private bool IsReadyToBuild()
+        {
+            IConfiguredProjectReadyToBuild? readyToBuild = ReadyToBuild.FirstOrDefault()?.Value;
+
+            return readyToBuild?.IsValidToBuild == true;
+        }
```
MiYanni, line 119: Could be written as:
```suggestion
        private bool IsReadyToBuild() => ReadyToBuild.FirstOrDefault()?.IsValidToBuild ?? false;
```

tmeschter:
Regarding "OS" vs. "platform", I recommend we use "OS" in user-facing text (i.e. the `DisplayName` and `Description`) and "platform" in the internal identifiers (e.g. in the property `Name` and the interceptors/enum providers class names).

"Platform" is over-used and I doubt it would be terribly meaningful to end users in this context. When we're giving users a choice between "Windows", "Android", "iOS", etc., "OS" is a much more helpful term, even if not entirely accurate.

Note that the Build page already has "Platform target" so we should definitely _not_ have a property named "Target platform". :-)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
drewnoakes: I'm not sure this is easier to read.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
drewnoakes: Three, including the existing `SupportedTargetFrameworksProvider`.

sfoslund:
Looks good to me. I don't have strong feelings about OS vs platform terminology but @dsplaisted or @KathleenDollard might.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
MiYanni: Interesting. It is easier for me to read as it is a single expressions. IMO, single expressions are generally easier to read than separate statements; mainly due to the statements may/may not have relationships with each other. I generally like minimizing 'statement bloat' as it causes loss of intent. Honestly, I'd probably style this as 2 lines to make it more obvious.
```suggestion
        private static string ComputeValue(string targetFramework, string? targetPlatform, string? platformVersion) =>
            targetPlatform != null ? $"{targetFramework}-{targetPlatform}{platformVersion}" : targetFramework;
```

Anyway, readability is always preference as long as the logic is the same.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
melytc: This is addressed in the new commits, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
melytc: This is addressed in the new commits, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: Removed on the new commits, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: At the end, the logic ended up being more complex. Thanks for the feedback, though.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
melytc: This is a good point. I'm planning to address this in following commits. Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: I agree that naming consistency helps bring clarity into the code. At the end, I went ahead and changed every reference to those properties to a platform-related name (and follow the naming of existing associated properties), and kept the OS references in the actual string being shown in the UI. Hope this helps :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: This is addressed in the new commits, thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: This was _super_ valuable, and it's addressed in the new commits, thanks for the feedback!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
melytc: This code ended up being deleted, but thanks.

melytc:
I've pushed a couple of commits I've been working on, but I'm still missing the following items:
-  Show TargetFramework value in the UI (i.e. FooBar to be visible in said property).
- Condition this experience to only .NET Core apps targeting 5.0 or higher.
     - This probably means I need to change the ItemGroup I'm reading from from `SupportedTargetFramework` to the specific `SupportedNETCoreAppTargetFramework`. 
           - And make sure I'm not breaking the experience in the old Properties pages regarding the supported TargetFramework list.
     - Also, hide the platform properties from the UI when the project isn't targeting .NET5+.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+            if (!Strings.IsNullOrEmpty(complexTargetFramework.TargetFrameworkMoniker))
+            {
+                string targetFrameworkAlias = await GetTargetFrameworkAlias(complexTargetFramework.TargetFrameworkMoniker);
+
+                if (string.IsNullOrEmpty(targetFrameworkAlias))
+                {
+                    return string.Empty;
+                }
+
+                if (targetFrameworkAlias.Contains("Unsupported"))
```
drewnoakes, line 154: Where does this `"Unsupported"` value come from? I don't remember seeing this in the SDK's items. It looks like `GetTargetFrameworkAlias` returns `""` for unknown values.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+                {
+                    // The value on the TargetFramework is not on the supported list and we shouldn't try to parse it.
```
drewnoakes, line 156: Should this be:

```suggestion
                    // The value on the TargetFrameworkMoniker is not on the supported list and we shouldn't try to parse it.
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+        {
+            IProjectSubscriptionService? subscriptionService = _configuredProject.Services.ProjectSubscription;
+            Assumes.Present(subscriptionService);
+
+            IImmutableDictionary<string, IProjectRuleSnapshot> sdkSupportedTargetPlatformIdentifiers = await subscriptionService.ProjectRuleSource.GetLatestVersionAsync(_configuredProject, new string[] { SdkSupportedTargetPlatformIdentifier.SchemaName });
+
+            IProjectRuleSnapshot sdkSupportedTargetPlatformIdentifierRuleSnapshot = sdkSupportedTargetPlatformIdentifiers[SdkSupportedTargetPlatformIdentifier.SchemaName];
+
+            // The SdkSupportedTargetPlatformIdentifier rule stores the alias in the key value.
+            if (sdkSupportedTargetPlatformIdentifierRuleSnapshot.Items.TryGetKey(targetPlatformIdentifier, out string targetPlatformAlias))
```
drewnoakes, line 220: `TryGetKey` looks suspicious here. The first argument is a key. So this means "find the dictionary entry that has key `targetPlatformIdentifier` and return that key's value as `targetPlatformAlias`". It doesn't make sense to mix TPI and TPA.

I think you need something more like:

```c#
string targetPlatformAlias = sdkSupportedTargetPlatformIdentifierRuleSnapshot.Items
    .FirstOrDefault(entry => StringComparers.PropertyValues.Equals(entry.Value, targetPlatformIdentifier))
    ?.Key;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/xlf/Resources.cs.xlf
```diff
@@ -96,6 +91,11 @@ Tento projekt se načetl pomocí nesprávného typu projektu, pravděpodobně v
         <target state="translated">Projekty</target>
         <note />
       </trans-unit>
+      <trans-unit id="Property_NoneValue">
+        <source>(None)</source>
+        <target state="new">(None)</target>
+        <note>Special value to show in the project property pages when a property has no value.</note>
+      </trans-unit>
```
drewnoakes, line 98: It would have been possible here to preserve the prior translations via replace-in-files. As done here, they will need to be re-translated.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ApplicationPropertyPage.xaml.cs.xlf
```diff
@@ -42,19 +72,34 @@
         <target state="translated">Spouštěcí objekt</target>
         <note />
       </trans-unit>
-      <trans-unit id="DynamicEnumProperty|TargetFramework|Description">
-        <source>Specifies the version of .NET that the application targets. This option can have different values depending on which versions of .NET are installed on your computer.</source>
-        <target state="translated">Určuje verzi rozhraní .NET, na kterou aplikace cílí. Tato možnost může mít různé hodnoty v závislosti na tom, které verze rozhraní .NET jsou na počítači nainstalované.</target>
```
drewnoakes, line 47: These translations could also have been preserved. It's a shame our tooling doesn't do a better job of tracking this ID renames, in cases where the source strings are the same.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+            TargetPlatformVersionProperty
+        },
+        ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class ComplexTargetFrameworkValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string InterceptedTargetFrameworkProperty = "InterceptedTargetFramework";
+        private const string TargetPlatformProperty = "TargetPlatform";
+        private const string TargetPlatformVersionProperty = "TargetPlatformVersion";
+        private const string TargetFrameworkProperty = "TargetFramework";
+        private const string MinimumPlatformVersionProperty = "MinimumPlatformVersion";
```
drewnoakes, line 30: Could use the existing constants instead:

- `ConfigurationGeneral.TargetPlatformVersionProperty`
- `ConfigurationGeneral.TargetFrameworkProperty`
- `ConfigurationGeneral.MinimumPlatformVersionProperty`

Using the existing constants makes it easier to find where we're using a given XAML rule property throughout the code (via Find All References).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            return string.Empty;
+        }
+
+        /// <summary>
+        /// Retrieves the target framework alias (i.e. net5.0) from the project's subscription service.
+        /// </summary>
+        /// <returns></returns>
+        private async Task<string> GetTargetFrameworkAlias(string targetFrameworkMoniker)
```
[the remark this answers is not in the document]
melytc: Added in next commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: True, this validation was used before I added it in the `GetTargetFrameworkAlias`, where I could get something like this: 
image
Removed on next commit, thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: Agree, added in the next commit. Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: This is a tricky one, the structure ends up looking like this:
image
The `RuleSnapshot` has a list of Items, which the first element contains the key-value: "windows", and the list of other properties like DisplayName ("Windows"), ProjectFullPath and ProjectDirectory.

I tried with the code you suggested and when asked to get the TPA for 'windows' it returned an empty value. I'll come back to further investigate this one.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            var list = new List<IEnumValue>();
+
+            list.Add(new PageEnumValue(new EnumValue()
+            {
+                Name = string.Empty,
+                DisplayName = Resources.Property_NoneValue
+            }));
+
+            list.AddRange(snapshot.Items.Select(ToEnumValue));
```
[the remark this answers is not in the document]
melytc: Ordering would be nice, I think the three providers would benefit from it. They are all related to #7025. I am able to order the platforms, for now. I'll add them in the next commit. Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/xlf/Resources.cs.xlf
melytc: I did change the note to reflect its use in a more generic context (it was previously used on the NeutralLanguage property), so that triggered changes in the translations.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ApplicationPropertyPage.xaml.cs.xlf
melytc: I agree, it's weird how it picks up the changes and triggers localization updates.

src/Microsoft.VisualStudio.ProjectSystem.Managed/xlf/Resources.cs.xlf
melytc: Was able to save these translations when I fixed the merge conflicts this was having with main :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ApplicationPropertyPage.xaml.cs.xlf
melytc: Was able to save these translations when I fixed some merge conflicts just now (with main).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformVersionProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            return transformBlock;
+        }
+
+        private static EnumCollection Transform(IProjectSubscriptionUpdate input)
+        {
+            IProjectRuleSnapshot snapshot = input.CurrentState[SdkSupportedTargetPlatformVersion.SchemaName];
+
+            return snapshot.Items.Select(ToEnumValue)
+                                    .ToList();
```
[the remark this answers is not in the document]
drewnoakes: The screenshots in the PR description show the version numbers appearing in "random" order.

To order strings containing numbers, there's a "natural order" comparer I wrote several years ago and am happy to contribute to this repo: https://stackoverflow.com/a/41168219/24874

drewnoakes (commented):
This is getting very close. A few comments below for you to consider.

If you'd like, we can look at this when we sync up tomorrow.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,234 @@
...
+                TargetFramework = (string?)await configuration.TargetFramework.GetValueAsync()
+            };
+
+            return storedValues;
+        }
+
+        /// <summary>
+        /// Extracts the specified value from the ComplexTargetFramework structure.
+        /// </summary>
+        /// <param name="values">The strucutre were all the values are.</param>
```
tmeschter, line 121: Nit: "structure where"

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
+        {
+            ConfigurationGeneral configuration = await _properties.GetConfigurationGeneralPropertiesAsync();
+            ComplexTargetFramework storedValues = await GetStoredComplexTargetFramework(configuration);
+
+            if (storedValues.TargetFrameworkMoniker != null && storedValues.TargetFrameworkMoniker.Contains("Unsupported"))
+            {
+                return await base.OnGetEvaluatedPropertyValueAsync(propertyName, "", defaultProperties);
```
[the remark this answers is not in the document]
tmeschter: (This is a minor question and I see no reason to hold the PR for this.)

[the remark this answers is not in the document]
melytc: Addressed in the last commit, as we discussed offline, thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            if (storedValues.TargetFrameworkMoniker != null && storedValues.TargetFrameworkMoniker.Contains("Unsupported"))
+            {
+                return await base.OnGetUnevaluatedPropertyValueAsync(propertyName, "", defaultProperties);
+            }
+
+            return await base.OnGetUnevaluatedPropertyValueAsync(propertyName, ExtractValue(storedValues, propertyName) ?? "", defaultProperties);
+        }
+
+        private static async Task<ComplexTargetFramework> GetStoredComplexTargetFramework(ConfigurationGeneral configuration)
```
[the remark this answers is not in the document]
melytc: Addressed on last commit, thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SdkSupportedTargetPlatformIdentifierProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            JoinUpstreamDataSources(source);
+
+            return transformBlock;
+        }
+
+        private static EnumCollection Transform(IProjectSubscriptionUpdate input)
+        {
+            IProjectRuleSnapshot snapshot = input.CurrentState[SdkSupportedTargetPlatformIdentifier.SchemaName];
+
+            var list = new List<IEnumValue>();
```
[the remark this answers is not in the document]
melytc: Added in next commit, thanks!

sfoslund:
General comment based on the screenshot: @melytc can we order the target platform versions from lowest to highest?

melytc:
> General comment based on the screenshot: @melytc can we order the target platform versions from lowest to highest?

#7025 is tracking this for other properties in the Project Properties UI, like the target framework. I will incorporate @drewnoakes's contribution for now, but a broader solution might replace it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Frameworks/SupportedTargetFrameworksProvider.cs
```diff
@@ -53,13 +53,13 @@ public void Load()
...
-            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _subscriptionService.ProjectBuildRuleSource;
+            IProjectValueDataSource<IProjectSubscriptionUpdate> source = _subscriptionService.ProjectRuleSource;
 
-            // Transform the changes from design-time build -> Supported target frameworks
+            // Transform the changes from evaluation -> Supported target frameworks
             DisposableValue<ISourceBlock<EnumCollectionProjectValue>> transformBlock = source.SourceBlock.TransformWithNoDelta(
                 update => update.Derive(Transform),
                 suppressVersionOnlyUpdates: false,
-                ruleNames: SupportedTargetFrameworkAlias.SchemaName);
+                ruleNames: SupportedTargetFramework.SchemaName);
```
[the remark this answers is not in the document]
melytc: I believe this doesn't have other properties using these values, other than the one I'm changing. The provider name is the same so the `GeneralBrowseObject` will continue to produce the `SupportedTargetFrameworks` property.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/RuleExporter.cs
```diff
@@ -28,11 +28,29 @@ private static class ProjectRules
             ///     Represents the design-time build items containing the supported (possible) target frameworks
             ///     for a project.
             /// </summary>
-            [ExportRule(nameof(SupportedTargetFrameworkAlias), PropertyPageContexts.ProjectSubscriptionService)]
+            [ExportRule(nameof(SupportedTargetFramework), PropertyPageContexts.ProjectSubscriptionService)]
             [AppliesTo(ProjectCapability.DotNet)]
             [Order(Order.Default)]
             public static int SupportedTargetFrameworkAliasRule;
```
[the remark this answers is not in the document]
melytc: That makes sense, I tested it and it seems okay to make that change.

melytc:
Ordering is fixed now:
image

mhutch:
How does `MinimumPlatformVersion` relate to the `SupportedOSPlatformVersion` property we standardized on? It seems to me to be redundant. /cc @terrajobst

mhutch:
Context: https://github.com/dotnet/designs/search?q=SupportedOSPlatformVersion

melytc:
> How does `MinimumPlatformVersion` relate to the `SupportedOSPlatformVersion` property we standardized on? It seems to me to be redundant. /cc @terrajobst

I'm not sure, I think this is more a question for the SDK team as we just saved the property with the name they asked us to - @sfoslund

sfoslund:
I'm not sure, @dsplaisted will know

dsplaisted:
It looks like the SDK no longer uses `MinimumPlatformVersion` version at all.  So I think this UI needs to be updated to use `SupportedOSPlatformVersion`.  @melytc

drewnoakes:
Thanks @dplaisted. File https://github.com/dotnet/project-system/issues/7502 to track this.

PR #7138, opened 2021-04-21
Handle exception when invalid document id.

ocallesp:
[AB#1243638](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1243638)

When a resx file is open and then moved to a folder, the resx designer holds a reference to a DTE.ProjectItem that becomes stale when the item id changes. Attempting to use it when stale will cause this exception.

This is a CPS limitation and Dotnet ProjectSystem should handle this exception.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7138)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/Common/DTEUtils.vb
+++ b/src/Microsoft.VisualStudio.Editors/Common/DTEUtils.vb
@@ -113,9 +113,15 @@ Namespace Microsoft.VisualStudio.Editors.Common
         ''' </summary>
         ''' <param name="PropertyName">The name of the property to retrieve.</param>
         Public Shared Function GetProjectItemProperty(ProjectItem As ProjectItem, PropertyName As String) As [Property]
-            If ProjectItem.Properties Is Nothing Then
+            Try
+                If ProjectItem.Properties Is Nothing Then
+                    Return Nothing
+                End If
+            Catch ex As KeyNotFoundException
+                ' The issue is a limitation in CPS. The main cause is that CPS, in some cases,
+                ' it is not able to maintain the same item id for items, causing them to not be found
                 Return Nothing
-            End If
+            End Try
 
             For Each Prop As [Property] In ProjectItem.Properties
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/Common/DTEUtils.vb
```diff
@@ -113,9 +113,15 @@ Namespace Microsoft.VisualStudio.Editors.Common
...
+            Try
+                If ProjectItem.Properties Is Nothing Then
+                    Return Nothing
+                End If
+            Catch ex As KeyNotFoundException
+                ' The issue is a limitation in CPS. The main cause is that CPS, in some cases,
+                ' it is not able to maintain the same item id for items, causing them to not be found
                 Return Nothing
-            End If
+            End Try
```
MiYanni, line 124: I don't suggest we eat the exception. It is always a bad practice unless there is *no* other way. In this case, it looks like `OAProjectItem` in CPS implements this interface. It only calls `Create` in `OAProperties` so the root cause shouldn't be too complex to find. In this case, accessing a property shouldn't throw `KeyNotFoundException`. As this code expects, it should be null (Nothing) if it isn't present. Eating exceptions generally causes a lot more issues than it fixes in the long term.

MiYanni (changes requested):

src/Microsoft.VisualStudio.Editors/Common/DTEUtils.vb
ocallesp: in AB#1243638 Adrian describes the CPS limitation and Dotnet ProjectSystem should handle this exception.

drewnoakes (approved):

davkean:
There's two bugs here neither of which this fixes:

1) Moving/deleting the file is supposed to close the designer. CPS owns this.
2) Renaming the file, the designer is supposed to handle.

You've handled the crash, but the very next time you save the file - it will be recreated.

PR #7139, opened 2021-04-21
Add commands for Unused References.

ocallesp:
The new interface IProjectSystemReferenceCleanupService2 replaces TryUpdateReferenceAsync() with GetUpdateReferenceOperationAsync().

This new method is needed to implement the command pattern where the commands are created in DotNet ProjectSystem and Roslyn executes them when needed.
dotnet/roslyn#52434

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7139)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AssemblyReferenceHandler.cs
```diff
@@ -16,25 +16,45 @@ internal class AssemblyReferenceHandler : AbstractReferenceHandler
...
+            string? assemblyPath = null;
+
+            if (Path.IsPathRooted(itemSpecification))
+            {
+                assemblyPath = itemSpecification;
+            }
+            else
+            {
+                assemblyName = new AssemblyName(itemSpecification);
+            }
```
MiYanni, line 52: I put this on the other PR as a suggestion:
```suggestion
            bool isPathRooted = Path.IsPathRooted(itemSpecification);
            AssemblyName? assemblyName = isPathRooted ? null : new AssemblyName(itemSpecification);
            string? assemblyPath = isPathRooted ? itemSpecification : null;
```

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AssemblyReferenceHandler.cs
drewnoakes: I find the conditional dependence harder to see between the two ternaries here, compared to the original code.

If you really wanted to reduce this, perhaps:

```c#
(AssemblyName? assemblyName, string? assemblyPath) = Path.IsPathRooted(itemSpecification)
    ? (null, new AssemblyName(itemSpecification))
    : (itemSpecification, null);
```

But I'm not sure it's really better than the original.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (item != null)
+            {
+                await item.Metadata.SetPropertyValueAsync(ProjectReference.TreatAsUsedProperty, PropertySerializer.SimpleTypes.ToString(true), null);
+            }
+
+            return true;
+        }
+
+        private async Task<IProjectItem> GetProjectItem()
```
drewnoakes, line 47: 1. This can return `null`
2. Should have an `Async` suffix

```suggestion
        private async Task<IProjectItem?> GetProjectItemAsync()
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public async Task<bool> ApplyAsync(CancellationToken cancellationToken)
+        {
+            IProjectItem item = await GetProjectItem();
+
+            if (item != null)
+            {
+                await item.Metadata.SetPropertyValueAsync(ProjectReference.TreatAsUsedProperty, PropertySerializer.SimpleTypes.ToString(false), null);
+            }
```
drewnoakes, line 30: If `item` is null, should this return `false` or flag an error somewhere? Ditto below in `RevertAsync`, and twice again in `SetAttributeCommand`. The only false return I see is in the null command.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
+                var value = await items.Metadata.GetEvaluatedPropertyValueAsync(property);
+                propertyValues.Add(string.Copy(property), string.Copy(value));
+            }
 
-                await item.Metadata.SetPropertyValueAsync(ProjectReference.TreatAsUsedProperty, newValue, null);
+            return propertyValues;
+        }
 
-                wasUpdated = true;
+        public async Task SetAttributes(ConfiguredProject selectedConfiguredProject, string itemSpecification, Dictionary<string, string>? projectPropertiesValues)
```
drewnoakes, line 125: nit: `Async` suffix.

Ideally we'd have an analyzer to catch these.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
+        {
+            if (projectPropertiesValues is null)
+            {
+                return;
             }
 
-            return wasUpdated;
+            IProjectItem items = await GetProjectItems(selectedConfiguredProject, itemSpecification);
+
+            if (items != null)
```
drewnoakes, line 134: If `items` is `null`, does that indicate a bug? If so, this will fail silently, which might make a bug go unnoticed or make a noticed bug hard to diagnose.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
+            return propertyValues;
+        }
 
-                wasUpdated = true;
+        public async Task SetAttributes(ConfiguredProject selectedConfiguredProject, string itemSpecification, Dictionary<string, string>? projectPropertiesValues)
+        {
+            if (projectPropertiesValues is null)
+            {
+                return;
             }
```
drewnoakes, line 130: Consider making the `projectPropertiesValues` parameter non-nullable and not calling this method when there's nothing to do. It feels strange to me to have a method that does nothing when given nothing to do. There's only one call site, so I think it would be clearer if that caller did the null check.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
         private static async Task<bool> GetAttributeTreatAsUsedAsync(IProjectProperties metadata)
         {
+            var propertyNames = await metadata.GetPropertyNamesAsync();
             string? value = await metadata.GetEvaluatedPropertyValueAsync(ProjectReference.TreatAsUsedProperty);
 
             return value != null && PropertySerializer.SimpleTypes.ToValue<bool>(value);
         }
 
-        internal async Task<bool> UpdateReferenceAsync(ConfiguredProject selectedConfiguredProject, ProjectSystemReferenceUpdate referenceUpdate, CancellationToken cancellationToken)
+        private async Task<IProjectItem> GetProjectItems(ConfiguredProject selectedConfiguredProject,
```
drewnoakes, line 80: A few issues with this method signature:

1. It can return `null` so should reflect that in the return type
2. It is async so should have an `Async` suffix
3. It returns a single item. The name suggests it returns multiple.

Instead:

```suggestion
        private async Task<IProjectItem?> GetProjectItemAsync(ConfiguredProject selectedConfiguredProject,
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
+        {
+            return new RemoveReferenceCommand(this, selectedConfiguredProject, referenceUpdate);
+        }
+
+        public async Task<Dictionary<string, string>> GetAttributesAsync(ConfiguredProject selectedConfiguredProject, string itemSpecification)
+        {
+            Dictionary<string, string> propertyValues = new ();
 
-            if (item != null)
+            IProjectItem items = await GetProjectItems(selectedConfiguredProject, itemSpecification);
```
drewnoakes, line 112: This is a single `item`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
-                wasUpdated = true;
+        public async Task SetAttributes(ConfiguredProject selectedConfiguredProject, string itemSpecification, Dictionary<string, string>? projectPropertiesValues)
+        {
+            if (projectPropertiesValues is null)
+            {
+                return;
             }
 
-            return wasUpdated;
+            IProjectItem items = await GetProjectItems(selectedConfiguredProject, itemSpecification);
```
drewnoakes, line 132: This is a single `item`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
-            if (item != null)
+            IProjectItem items = await GetProjectItems(selectedConfiguredProject, itemSpecification);
+
+            var propertyNames = await items.Metadata.GetPropertyNamesAsync();
+
+            foreach (var property in propertyNames)
             {
-                string newValue = PropertySerializer.SimpleTypes.ToString(referenceUpdate.Action == ProjectSystemUpdateAction.SetTreatAsUsed);
+                var value = await items.Metadata.GetEvaluatedPropertyValueAsync(property);
+                propertyValues.Add(string.Copy(property), string.Copy(value));
```
drewnoakes, line 119: I can't see a reason to copy the strings here.

```suggestion
                propertyValues.Add(property, value);
```

In .NET strings are immutable, except for in uncommon corner cases that don't appear to apply here. If they do apply here, they should be called out in a comment.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -68,32 +71,73 @@ internal async Task<List<ProjectSystemReferenceInfo>> GetReferencesAsync(Configu
...
 
-            return wasUpdated;
+            IProjectItem items = await GetProjectItems(selectedConfiguredProject, itemSpecification);
+
+            if (items != null)
+            {
+                foreach (var property in projectPropertiesValues)
+                {
+                    await items.Metadata.SetPropertyValueAsync(property.Key, property.Value, null);
+                }
```
drewnoakes, line 139: It's possible to destructure here. I like the fact that it gives control over the name (rather than `Key`/`Value`).

```suggestion
                foreach ((string propertyName, string propertyValue) in projectPropertiesValues)
                {
                    await items.Metadata.SetPropertyValueAsync(propertyName, propertyValue, null);
                }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (item != null)
+            {
+                await item.Metadata.SetPropertyValueAsync(ProjectReference.TreatAsUsedProperty, PropertySerializer.SimpleTypes.ToString(false), null);
+            }
+
+            return true;
+        }
+
+        private async Task<IProjectItem> GetProjectItem()
```
drewnoakes, line 47: 1. This can return `null`
2. Should have an `Async` suffix

```suggestion
        private async Task<IProjectItem?> GetProjectItemAsync()
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    internal class UnSetAttributeCommand : IProjectSystemUpdateReferenceOperation
```
drewnoakes, line 10: The `SetAttributeCommand` and `UnSetAttributeCommand` classes are almost identical. Consider pulling out a base class to avoid the duplication.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AssemblyReferenceHandler.cs
MiYanni: Again, it is preference. However, the ValueTuple version there (from my understanding) would do more allocations than just directly assigning the values. I tend to not create extraneous statement groups (if {} and else {}) unless absolutely necessary. I like binding variable declaration and initialization as close as possible when possible. Makes understanding (for me) more straightforward. And the extra statement groups encourages people (over time) to just inject random statements into them that have no relationship to their original intent. It is just a 'convenient' place to put it, and thus makes it more complex. I'm just very anti-imperative programming as I've seen it cause more bugs than it prevents. 😝

drewnoakes: > would do more allocations than just directly assigning the values

There are no heap allocations in the code I proposed, if that's what you're referring to.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
ocallesp: Done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
ocallesp: Done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            return true;
+        }
+
+        private async Task<IProjectItem?> GetProjectItemAsync()
+        {
+            var projectItems = await _referenceHandler.GetUnresolvedReferencesAsync(_selectedConfiguredProject);
+
+            var item = projectItems
+                .FirstOrDefault(c => c.EvaluatedInclude == _itemSpecification);
```
swesonga, line 56: >c.EvaluatedInclude == _itemSpecification [](http://example.com/codeflow?start=37&length=40)

Consider specifying an explicit string comparer instead of a raw == operator

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+        public async Task<bool> ApplyAsync(CancellationToken cancellationToken)
+        {
+            IProjectItem item = await GetProjectItem();
+
+            if (item == null)
+            {
+                return false;
+            }
+
+            await item.Metadata.SetPropertyValueAsync(ProjectReference.TreatAsUsedProperty, PropertySerializer.SimpleTypes.ToString(true), null);
```
swesonga, line 32: >PropertySerializer.SimpleTypes.ToString(true) [](http://example.com/codeflow?start=92&length=45)

Would it make sense to precompute these in the constructor to avoid repetitive work later?

swesonga, line 32: >null [](http://example.com/codeflow?start=139&length=4)

This is an optional value that is already set to null so it is not needed here. Same in the RevertAsync call

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    internal class SetAttributeCommand : IProjectSystemUpdateReferenceOperation
```
swesonga, line 10: >SetAttributeCommand [](http://example.com/codeflow?start=19&length=19)

SetAttributeCommand (and the unset variant) seem very generic to me given that it actually does. What do you think about a more specific name like (Un)SetTreatAsUsedAttributeCommand?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.References
+{
+    internal class UnSetAttributeCommand : IProjectSystemUpdateReferenceOperation
```
swesonga, line 10: >UnSet [](http://example.com/codeflow?start=19&length=5)

nit: as far as naming, isn't "Unset" one word? if so, the uppercase S in UnSet seems a bit off.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
```diff
@@ -87,47 +93,45 @@ private static async Task<List<ProjectSystemReferenceInfo>> GetAllReferencesInCo
...
 
             var activeConfiguredProject = await GetActiveConfiguredProjectByPathAsync(projectPath, cancellationToken);
 
             var referenceHandler = s_mapReferenceTypeToHandler[referenceUpdate.ReferenceInfo.ReferenceType];
 
+            IProjectSystemUpdateReferenceOperation command = s_nullCommand;
             if (referenceHandler != null)
             {
-                wasUpdated = await ApplyActionAsync(referenceUpdate, referenceHandler, activeConfiguredProject, cancellationToken);
+                command = CreateCommand(referenceUpdate, referenceHandler, activeConfiguredProject, cancellationToken);
```
swesonga, line 118: >command [](http://example.com/codeflow?start=16&length=7)

minor style observation you can ignore: this local variable isn't needed if you use a return statement here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
ocallesp: Actually I wanted to have only one exit point instead of having many returns.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
ocallesp: This means that the itemSpecification was not found

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/ReferenceCleanupService.cs
drewnoakes: That's fair, but it's not a commonly followed pattern in this repo (or in .NET in general).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
```diff
@@ -1,50 +1,16 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using System.Linq;
-using System.Threading.Tasks;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.References
 {
-    internal class SetAttributeCommand : IReferenceCommand
+    internal class SetAttributeCommand : BaseSetAttributeCommand
```
drewnoakes, line 7: @sawesonga suggested a better name for this class, and you said it was done, but I don't see it here. Ditto for the "Unset" class.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
```diff
@@ -1,50 +1,16 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using System.Linq;
-using System.Threading.Tasks;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.References
 {
-    internal class UnSetAttributeCommand : IReferenceCommand
+    internal class UnsetAttributeCommand : BaseSetAttributeCommand
```
drewnoakes, line 7: The file name's capitalization should also be updated.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -87,18 +87,19 @@ private static async Task<bool> GetAttributeTreatAsUsedAsync(IProjectProperties
...
         {
-            if (referenceUpdate.Action == ProjectSystemUpdateAction.SetTreatAsUsed)
-            {
-                return new SetAttributeCommand(this, selectedConfiguredProject, referenceUpdate.ReferenceInfo.ItemSpecification);
-            }
+            return new SetAttributeCommand(this, selectedConfiguredProject, referenceUpdate.ReferenceInfo.ItemSpecification);
+        }
 
-            return new UnSetAttributeCommand(this, selectedConfiguredProject, referenceUpdate.ReferenceInfo.ItemSpecification);
+        internal IProjectSystemUpdateReferenceOperation CreateUnSetAttributeCommand(ConfiguredProject selectedConfiguredProject,
```
drewnoakes, line 96: Capitalization of "UnSet" was not updated here (possibly elsewhere too).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/AbstractReferenceHandler.cs
```diff
@@ -108,33 +109,28 @@ private static async Task<bool> GetAttributeTreatAsUsedAsync(IProjectProperties
...
         {
-            if (projectPropertiesValues is null)
-            {
-                return;
-            }
-
-            IProjectItem items = await GetProjectItems(selectedConfiguredProject, itemSpecification);
+            IProjectItem items = await GetProjectItemsAsync(selectedConfiguredProject, itemSpecification);
 
             if (items != null)
```
drewnoakes, line 129: If this can return `null`, it should be annotated (both the local variable and the `GetProjectItemsAsync` method).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/SetAttributeCommand.cs
ocallesp: Done. "SetTreatAsUsedAttributeCommand"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/References/UnSetAttributeCommand.cs
ocallesp: If I change the name from Un**S**etAttributeCommand.cs to Un**s**etAttributeCommand.cs  Git doesn't detect this  change.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/References/ReferenceCleanupServiceTests.cs
```diff
@@ -73,57 +70,39 @@ public async Task GetProjectReferencesAsync_FoundReferences_ReturnAllReferences(
...
 
         [Fact]
-        public async Task UpdateReferenceAsync_TreatAsUsed_ReferencesChangedToTreatAsUnused()
+        public async Task UpdateReferencesAsync_GetUpdateReferenceOperationAsync_FailedToCreateOperation()
         {
-            s_item = null;
             var referenceCleanupService = Setup();
             var referenceUpdate1 =
-                new ProjectSystemReferenceUpdate(ProjectSystemUpdateAction.UnsetTreatAsUsed, new ProjectSystemReferenceInfo(ProjectSystemReferenceType.Package, _package3, true));
+                new ProjectSystemReferenceUpdate((ProjectSystemUpdateAction)3, new ProjectSystemReferenceInfo(ProjectSystemReferenceType.Package, "UnknownPackage", true));
```
drewnoakes, line 101: Why the literal `3` here?

PR #7140, opened 2021-04-21
Add a basic HOW TO on adding a new property page

tmeschter:
Adds documentation going through the basic steps of creating a new property page. Much of this is covered in the existing property-specification.md file, but I felt the need for something a bit more guided.

I plan to fill out the second option (embedding the XAML file) in a subsequent change; I started writing that section and then realized I didn't actually know enough of the details. :-)

I also plan to write similar HOW TOs for adding a launch profile, extending an existing property page or launch profile, localization, and other common operations.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7140)


docs/repo/property-pages/how-to-add-a-new-property-page.md
```diff
@@ -0,0 +1,88 @@
...
+
+Add a new XAML file named "MyPropertyPage.xaml" to your project. Depending on how the file is created you may end up with a `<Page>` item in your project but this is not what we want as we're not using the file to describe a piece of WPF UI.
+
+Update your project to replace the `<Page>` item with one of the following:
+
+- SDK-style projects:
+  ``` xml
+  <None Update="MyPropertyPage.xaml">
+    <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+  </None>
```
MiYanni, line 23: For SDK-style, this style is preferred:
```suggestion
  <None Update="MyPropertyPage.xaml" CopyToOutputDirectory="PreserveNewest" />
```

MiYanni (approved):

docs/repo/property-pages/how-to-add-a-new-property-page.md
```diff
@@ -0,0 +1,88 @@
...
+</Rule>
+```
+
+The format of the file is described in detail in [Property Specification](property-specification.md), but the most important points are:
+- The `Name` must be unique.
+- The `PageTemplate` attribute must have the value `"generic"`.
+
+You should now be able to build and see the MyPropertyPage.xaml copied as-is to the output directory.
+
+### Step 4: Define the `PropertyPageSchema` item
```
drewnoakes, line 70: Consider moving this step after step 1, or merging into step 1, so that all project file manipulation happens at the same time.

docs/repo/property-pages/how-to-add-a-new-property-page.md
```diff
@@ -0,0 +1,88 @@
...
+
+Next you need to update your .props or .targets files to properly reference the property page so Visual Studio can find it. Note that the creation and distribution of the .props and .targets files (as well as the distribution of MyPropertyPage.xaml itself) is beyond the scope of this document.
+
+Add the following item to your .targets file:
```
drewnoakes, line 74:
```suggestion
Add the following item to your .props or .targets file:
```

docs/repo/property-pages/README.md
```diff
@@ -12,6 +12,12 @@ This documentation details the updated Project Properties UI and associated back
 4. Extensibility. We need to support 3rd parties (both internal and external) adding both entirely new pages _and_ customizing the pages we provide. 
 5. Support new functionality. We want to enable scenarios such as searching for properties, editing the MSBuild expressions that underpin many properties, and collecting the most commonly used properties in one area.
 
+## How-To Guides
+
+These guides describe how to get specific scenarios up and running, step by step.
+
+- [HOW TO: Add a new property page](how-to-add-a-new-property-page.md)
+
```
drewnoakes, line 20: I'm not sure we need a dedicated 'how to' section. It splits our documentation for extenders/customisation in two.

I'd vote for just moving the _Customising Project Properties_ section above the _Architecture_ section and adding the new how-to document as a bullet point to that list.

docs/repo/property-pages/how-to-add-a-new-property-page.md
tmeschter: Well, it's better supported in SDK-style projects, but it's hard to call it preferred. If you update the property of a file through the Properties window (e.g. set "Copy to Output Directory" to "Copy if newer") we won't use the attribute style, even for SDK-style projects.

The only place it's really common right now is `PackageReference`.

drewnoakes (approved):

docs/repo/property-pages/how-to-add-a-new-property-page.md
MiYanni: So, a few years back, I migrated a bunch of projects from csproj-style to SDK-style. And, at that time, I found documentation on the format with a decent migration guide. Right now, I just spent some time searching, and literally everything I found from the past is no longer there. No idea where it went, but the internet has lost information as it does over time. However, I can find remnants of why I said this is 'preferred':
- https://stackoverflow.com/a/54272143/294804
- https://github.com/dotnet/msbuild/issues/2795
- https://stackoverflow.com/a/58600845/294804

I can find answers, posts, questions, issues, etc. showing it being used as an attribute versus a node. But, I honestly cannot find the source of behind it. I would not take the actions of VS to be an 'absolute' on the format. But, I do know there had been information on what should be used as attributes versus nodes. I just can't find it anymore. 😭

Issue #7035, opened 2021-03-21
Handle read-only properties. (1)


drewnoakes:
Properties in rule files may be marked as "read only".

```xml
<StringProperty
  Name="MyProperty"
  ReadOnly="True" />
```

Currently, such properties are presented to the user as writeable.

We should expose `IsReadOnly` metadata from the back end to the UI, and update the UI to allow showing properties in a read only mode.

A potential use case here is displaying the evaluated `LangVersion` value alongside the current message that explains the value is not user-editable. The Razor SDK also produces read only properties for the Razor language version and configuration name.

- [x] Add `IsReadOnly` to project model in CPS ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/319288))
- [x] Populate `IsReadOnly` in the .NET Project System (#7141)
- [x] Modify the UI to show read only properties somehow differently (either disabled, or with alternative templates, etc) ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/327180))

drewnoakes:
This issue was discovered while exploring the Project Properties UI for ASP.NET Web Application projects. In such projects, a "Razor Properties" page is present:

![image](https://user-images.githubusercontent.com/350947/111923248-50f6c380-8af2-11eb-9343-f798a87ee9dd.png)

The properties are coming from https://github.com/dotnet/sdk/blob/main/src/RazorSdk/Targets/Rules/RazorGeneral.xaml in which they are marked as read only.

I've filed https://github.com/dotnet/sdk/issues/16458 to work with the SDK to improve or remove that page.

drewnoakes:
Also assigning @tmeschter as this will require back end changes too.

drewnoakes:
https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/319288 adds `UIProperty.IsReadOnly` to CPS's project model.

tmeschter:
@drewnoakes The back-end changes were done in #7141, right? What's left is the UI side?

drewnoakes:
@tmeschter yes, just UI stuff now. I've unassigned you. Thanks.

drewnoakes:
https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/327180 makes read-only properties non-editable, which is the final step in this issue (for now at least).

drewnoakes:
All parts have been merged.

PR #7141, opened 2021-04-22
Populate UIProperty.IsReadOnly

drewnoakes:
Relates to #7035

The `UIProperty.IsReadOnly` property was recently added to CPS's project model. We extend our data provider to include this data.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7141)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
@@ -58,6 +58,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                 newUIProperty.ConfigurationIndependent = !property.IsConfigurationDependent();
             }
 
+            if (requestedProperties.IsReadOnly)
+            {
+                newUIProperty.IsReadOnly = property.ReadOnly;
+            }
+
             if (requestedProperties.HelpUrl)
             {
                 newUIProperty.HelpUrl = property.HelpUrl;
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -66,8 +66,8 @@
     <PackageReference Update="Microsoft.VisualStudio.Shell.Interop.15.8.DesignTime"                   Version="16.8.30406.65" />
     <PackageReference Update="Microsoft.VisualStudio.Shell.Interop.16.6.DesignTime"                   Version="16.8.30406.65-pre" />
     <PackageReference Update="Microsoft.VisualStudio.TemplateWizardInterface"                         Version="16.8.30403.137-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.Threading"                                       Version="16.9.51" />
-    <PackageReference Update="Microsoft.VisualStudio.Threading.Analyzers"                             Version="16.9.51" />
+    <PackageReference Update="Microsoft.VisualStudio.Threading"                                       Version="16.10.53-alpha" />
+    <PackageReference Update="Microsoft.VisualStudio.Threading.Analyzers"                             Version="16.10.53-alpha" />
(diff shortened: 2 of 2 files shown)


jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
```diff
@@ -58,6 +58,11 @@ public static IEntityValue CreateUIPropertyValue(IQueryExecutionContext executio
                 newUIProperty.ConfigurationIndependent = !property.IsConfigurationDependent();
             }
 
+            if (requestedProperties.IsReadOnly)
+            {
+                newUIProperty.IsReadOnly = property.ReadOnly;
+            }
```
MiYanni, line 64: The changes look good. I'm just commenting with a more... personal question. For this method, do you think it is easier to read as a series of if-statements (current format) or if there was a data map between the properties of `BaseProperty` and `UIPropertyValue` (think constructor or extension method or something like dependency injection)? I'm just asking out of personal preference. If-statements leave me with the, "If it isn't set (statement is false), then what is the value?"

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
drewnoakes: > If it isn't set (statement is false), then what is the value?

The semantics here are that any non-requested property should not be set, and by doing so any attempt to actually use the property will trigger a runtime exception. So there's no need to worry about an un-set value.

drewnoakes:
The build is failing again due to use of CPS packages that include reference assemblies. cc @lifengl @adrianvmsft 

> System.BadImageFormatException : Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. **Reference assemblies should not be loaded for execution.  They can only be loaded in the Reflection-only loader context.** (Exception from HRESULT: 0x80131058)
> ---- System.BadImageFormatException : Cannot load a reference assembly for execution.

lifengl:
there is a known issue blocking CPS SDK package to build without the extra unsafe.dll. The workaround for now is to only update the query package, and potentially threading package as its dependencies.  The CPS/Query doesn't cross reference each other, and doesn't have to be aligned.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
MiYanni: Interesting! That is a convention that is not conveyed well in this language. :-( Are the properties something similar to [`Lazy<T>`](https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization)? The benefit of those is thread safety and such. And, you could set them up so that if [IsValueCreated](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1.isvaluecreated?view=net-5.0) is false, then throw the runtime exception.

drewnoakes: It's a convention for this library. It asks a little of the author, but gives better performance than something like lazy.

It's similar to EF from what I remember of using it many years ago. If you attempt to use data you didn't request, it can error at runtime.

Ideally this would be a compile time check, but I can't see how that's possible with the C# type system.

PR #7144, opened 2021-04-24
All packages are finally consumed from public feeds

dmonroym:
I'm now consuming the proper packages from the correct feeds. Things still missing:

- Re-enable warningsaserrors once we get an update on Microsoft.VisualStudio.ProjectSystem.Query since the only type of warning now looks as such: 
`CSC : warning CS1762: A reference was created to embedded interop assembly 'Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' because of an indirect reference to that assembly created by assembly 'Microsoft.VisualStudio.ProjectSystem.Query, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Consider changing the 'Embed Interop Types' property on either assembly. [G:\project-system\tests\Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests\Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests.csproj]`
- Re-enable tests. IAsyncServiceProvider was changed and Mock is failing - this only affects our own unit tests which are temporarily disabled

With the latest changes though we enable functionality for .NET Core projects

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7144)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/VsContainedLanguageComponentsFactory.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/VsContainedLanguageComponentsFactory.cs
@@ -4,7 +4,6 @@ using System;
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
-using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TextManager.Interop;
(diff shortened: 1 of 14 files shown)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Mocks/IOleAsyncServiceProviderFactory.cs
```diff
@@ -8,13 +8,12 @@ namespace Microsoft.VisualStudio.Shell.Interop
...
-        public static IOleAsyncServiceProvider ImplementQueryServiceAsync(object? service, Guid clsid)
+        public static IOleAsyncServiceProvider ImplementQueryServiceAsync(object? service, Type clsid)
         {
             var mock = new Mock<IOleAsyncServiceProvider>();
 
-            //TODO: Merged PIAs
-            //mock.Setup(p => p.QueryServiceAsync(ref clsid))
-            //  .Returns(IVsTaskFactory.FromResult(service));
+            mock.Setup(p => p.GetServiceAsync(clsid))
+              .Returns((System.Threading.Tasks.Task<object?>)IVsTaskFactory.FromResult(service));
```
MiYanni, line 16: nit: fully-qualified type name

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/VsContainedLanguageComponentsFactory.cs
```diff
@@ -68,30 +67,37 @@ private async Task<(HierarchyId itemid, IVsHierarchy? hierarchy, IVsContainedLan
...
 
-            return (IVsContainedLanguageFactory)service;
+            if(null != service)
+            {
+                return (IVsContainedLanguageFactory)service;
+            }
+            else
+            {
+                return null;
+            }
```
MiYanni, line 93: Couple things here:
- The cast (boxing) does nothing for a return statement in general. Here, it is there only because the return type is `IVsContainedLanguageFactory?` and the compiler needs to know how to convert to that. The non-nullable version, `IVsContainedLanguageFactory`, is convertable to the nullable version.
- This entire block can just be `return service as IVsContainedLanguageFactory;`. The if-else only exists because of my first bullet point. `as IVsContainedLanguageFactory` automatically is understood as `IVsContainedLanguageFactory?`.

```suggestion
            return service as IVsContainedLanguageFactory;
```

dmonroym:
@MiYanni I'm addressing both of these in #7153

PR #7152, opened 2021-04-27
Move warnaserror+ to the end of the command line arguments.

ocallesp:
Current implementation gets the command line arguments from from the rules and this changes the order of the arguments which causes Build to behave differently.
When using the _WarningsNotAsErrors_ feature from Intellisense some errors and warnings were showed as warnings in the Error List, but as errors in the Output window.

We can get the correct order of command line arguments by subscribing to `ProjectBuildSnapshotService`, and then parse the output result from the build target, but since the structure implies iterating over all the elements, extract the arguments, and also changes some interfaces we are  not using this approach, unless we want to preserve ALL the arguments in the same order.

A simple change is to change move _warnaserror+_ to the end of the command line arguments to make sure _WarningsNotAsErrors_ behave correctly.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7152)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
@@ -153,12 +153,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             // The order of options matters and we should use warnaserror- and then warnaserror+
             // Depending which comes first, the compiler produces different diagnostics.
             // We just need to move warnaserror+ to the end.
-            var warnaserrorPlus = snapshot.Items.Where(a => a.Key.Contains("/warnaserror+")).FirstOrDefault();
-            ImmutableArray<string> commandLineArguments = warnaserrorPlus.Key is null ? snapshot.Items.Keys.ToImmutableArray() : 
-                snapshot.Items.Remove(warnaserrorPlus.Key).Keys.ToArray().Append(warnaserrorPlus.Key).ToImmutableArray();
- 
+            var warnaserrorPlus = snapshot.Items.FirstOrDefault(a => a.Key.Contains("/warnaserror+"));
+            IEnumerable<string> commandLineArguments = warnaserrorPlus.Key is null ? snapshot.Items.Keys :
+                    snapshot.Items.Remove(warnaserrorPlus.Key).Keys.Append(warnaserrorPlus.Key);
+
             // We just pass all options to Roslyn
-            _context.SetOptions(commandLineArguments);
+            _context.SetOptions(commandLineArguments.ToImmutableArray());
         }
 
         private Task ProcessCommandLineAsync(IComparable version, IProjectChangeDiff differences, ContextState state, CancellationToken cancellationToken)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
@@ -150,8 +150,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         {
             Assumes.NotNull(_context);
 
+            // The order of options matters and we should use warnaserror- and then warnaserror+
+            // Depending which comes first, the compiler produces different diagnostics.
+            // We just need to move warnaserror+ to the end.
+            var warnaserrorPlus = snapshot.Items.Where(a => a.Key.Contains("/warnaserror+")).FirstOrDefault();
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -150,8 +150,15 @@ private void ProcessOptions(IProjectRuleSnapshot snapshot)
         {
             Assumes.NotNull(_context);
 
+            // The order of options matters and we should use warnaserror- and then warnaserror+
+            // Depending which comes first, the compiler produces different diagnostics.
+            // We just need to move warnaserror+ to the end.
+            var warnaserrorPlus = snapshot.Items.Where(a => a.Key.Contains("/warnaserror+")).FirstOrDefault();
```
MiYanni, line 156: nit: `FirstOrDefault` allows a `Where` condition in it. So, this can just be:
```suggestion
            var warnaserrorPlus = snapshot.Items.FirstOrDefault(a => a.Key.Contains("/warnaserror+"));
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -150,8 +150,15 @@ private void ProcessOptions(IProjectRuleSnapshot snapshot)
...
+            ImmutableArray<string> commandLineArguments = warnaserrorPlus.Key is null ? snapshot.Items.Keys.ToImmutableArray() : 
+                snapshot.Items.Remove(warnaserrorPlus.Key).Keys.ToArray().Append(warnaserrorPlus.Key).ToImmutableArray();
+ 
             // We just pass all options to Roslyn
-            _context.SetOptions(snapshot.Items.Keys.ToImmutableArray());
+            _context.SetOptions(commandLineArguments);
```
MiYanni, line 161: nit: Can be simplified. `ToArray` resolves the query, `Append` creates a new query, and `ToImmutableArray` resolves that new query. So, it is inefficient for performance and memory.
```suggestion
            IEnumerable<string> commandLineArguments = warnaserrorPlus.Key is null ? snapshot.Items.Keys :
                snapshot.Items.Remove(warnaserrorPlus.Key).Keys.Append(warnaserrorPlus.Key)

            // We just pass all options to Roslyn
            _context.SetOptions(commandLineArguments.ToImmutableArray());
```
In this way, the call to `SetOptions` will resolve the query, independent of the ternary's result.

MiYanni (approved):

drewnoakes (changes requested):
I don't think we should be treating this property as special. This is fragile and will likely come back to bite us.

Did you investigate using the `ProjectBuildSnapshotService` to obtain items in the correct order?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -150,8 +150,15 @@ private void ProcessOptions(IProjectRuleSnapshot snapshot)
         {
             Assumes.NotNull(_context);
 
+            // The order of options matters and we should use warnaserror- and then warnaserror+
+            // Depending which comes first, the compiler produces different diagnostics.
+            // We just need to move warnaserror+ to the end.
+            var warnaserrorPlus = snapshot.Items.Where(a => a.Key.Contains("/warnaserror+")).FirstOrDefault();
+            ImmutableArray<string> commandLineArguments = warnaserrorPlus.Key is null ? snapshot.Items.Keys.ToImmutableArray() : 
+                snapshot.Items.Remove(warnaserrorPlus.Key).Keys.ToArray().Append(warnaserrorPlus.Key).ToImmutableArray();
+
```
davkean, line 159: This is very inefficient and will start pushing lots of data on the LOH heap, increasing GC time. This has been a large problem in the past, and was written to be much more efficient. If this approach is taken, please rewrite to avoid inbetween allocations (ToArray()) allocations

davkean (dismissed):
Per my comment.

ocallesp:
> I don't think we should be treating this property as special. This is fragile and will likely come back to bite us.
> 
> Did you investigate using the `ProjectBuildSnapshotService` to obtain items in the correct order?

Yes, and it seems that the structure/data is not straight forward. We still need to parse/copy all the elements to a new data structure.

jaredpar:
I do not believe this addresses the problem. The command line can contain `warnAsError` combinations in any order. It's perfectly legal for `/warnAsError-` to come before `/warnAsError+` and this change would produce a semantically different compilation in that request. 

The PR here is attempting to sort the command line. That is just not going to work here. The only correct way to process this information is to keep the command line in the same order it appeared in originally.

drewnoakes:
Yes I agree this should not have been merged as is. We need to fix the underlying ordering problem.

ocallesp:
let me undo this and submit the other proposal suggested by Drew

PR #7164, opened 2021-04-29
Use Scale Set pools for Build Pipelines fixes #7162

### LINKED ISSUE #7162: Update OneOffInsertion path on pipelines
There was an arcade update to roslyn-tools last night that changed the path of the OneOffInsertion script.
 
So in our insertion pipelineswe'll need to modify the OneOffInsertion task to get rid of the tfm in the path
So before

 `.\RIT\tools\net46\OneOffInsertion.ps1 ...args`

And after

 `.\RIT\tools\OneOffInsertion.ps1 ...args`

dmonroym:


dmonroym:
only affects official build so merging before unit tests run

dmonroym:
Validated via https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4708900&view=results

Issue #6996, opened 2021-03-01
Exclude Projects from publish


vsfeedback:
_This issue has a corresponding [ticket on Developer Community](https://developercommunity2.visualstudio.com/t/Exclude-Projects-from-publish/1239717?preview=true). Please vote and comment there to make sure your voice is heard._

---
<p>Hello!</p> <p>I want to use the command &quot;dotnet publish *.sln&quot; on our CI but stumpled over UnitTest-Projects which got published too. After some digging i found this <a href="https://github.com/dotnet/docs/issues/13365" target=_blank>Issue</a> which suggest to use &quot;IsPublishable&quot; in a .csproj file. Having a Option in the Project settings of Visual Studio would be nice.</p> <p>Using Visual Studio 2019 16.7.7 and .NET Core 3.1</p>

---
#### Feedback Bot on 11/1/2020, 11:35 PM: 

Thank you for taking the time to provide your suggestion.  We will do some preliminary checks to make sure we can proceed further.  We'll provide an update once the issue has been triaged by the product team.

cartermp:
This would need UI design work, but in general the ask seems fairly clear.

An alternative would be to put an ask on the actual build (msbuild?) to recognize test projects and not publish them by default.

tmeschter:
We can think about adding an option for this to the new property pages implementation. We'll need to make sure that the property is handled correctly within Visual Studio as well--it would be odd to have a publishing option in VS that wasn't honored when publishing from VS.

danielchalmers:
> An alternative would be to put an ask on the actual build (msbuild?) to recognize test projects and not publish them by default.

I would be a big fan of this! wonder why it's not already the default

PR #7165, opened 2021-04-29
Add IsPublishable to new property pages

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/6996

This PR adds the [`IsPublishable`](https://github.com/dotnet/sdk/blob/397a0beb1dd8ccbea15c59909314d648a91054db/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets#L17) property to the new property pages. This property only affects processes outside of VS that use the csproj files. It allows the [`Publish`](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-targets#default-build-targets) target to run, which is executed via the [`dotnet publish`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish) command. According to the targets documentation, VS uses `PublishOnly` target since it won't do a build. The CLI docs specifically state that they use the `Publish` target. And the linked targets file for `IsPublishable` shows that the property only affects the `Publish` target.

The only 'documentation' that exists is a comment made in this [GitHub thread](https://github.com/dotnet/docs/issues/13365#issuecomment-512152483). In the PR, I link to the MSBuild targets since this affects only the `Publish` target. If we think the `dotnet publish` docs are more appropriate, I'll update the link.

Also, if this is targeting the wrong branch in our repo (since this PR is non-critical), let me know and I can update that too.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7165)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -209,4 +209,10 @@
       </NameValuePair>
     </StringProperty.Metadata>
   </StringProperty>
+
+  <!-- TODO Documentation issue for this property: https://github.com/MicrosoftDocs/visualstudio-docs/issues/6643 -->
+  <BoolProperty Name="IsPublishable"
+                DisplayName="Publishing"
+                Description="Allow publishing outside of Visual Studio, such as via &quot;dotnet publish&quot;."
+                Category="Packaging" />
 </Rule>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -14,6 +14,9 @@
     <Category Name="Resources"
               DisplayName="Resources"
               Description="Resource settings for the application." />
+    <Category Name="Packaging"
+              DisplayName="Packaging"
+              Description="Packaging settings for the application." />
   </Rule.Categories>
 
   <Rule.DataSource>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
@@ -27,13 +27,6 @@
                 Description="Produce a package file during build operations."
                 Category="General" />
 
-  <!-- TODO add a fwlink for this help URL -->
-  <BoolProperty Name="IsPublishable"
-                DisplayName="Allow project publishing (Publish target)"
-                Description="When publishing outside of Visual Studio, this allows commands such as &quot;dotnet publish&quot; to execute on this project."
-                HelpUrl="https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-targets"
-                Category="General" />
-
   <StringProperty Name="PackageId"
                   DisplayName="Package ID"
                   Description="The case-insensitive package identifier, which must be unique across nuget.org or whatever gallery the package resides in. IDs may not contain spaces or characters that are not valid for a URL, and generally follow .NET namespace rules."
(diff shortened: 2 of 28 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -27,6 +27,13 @@
                 Description="Produce a package file during build operations."
                 Category="General" />
 
+  <!-- TODO add a fwlink for this help URL -->
+  <BoolProperty Name="IsPublishable"
+                DisplayName="Allow project publishing (Publish target)"
+                Description="When publishing outside of Visual Studio, this allows commands such as &quot;dotnet publish&quot; to execute on this project."
```
drewnoakes, line 33: `DisplayName` is the heading and `Description` is the check box's label. Most bool properties have a description which is more directive. For example:

```suggestion
                DisplayName="Publishing"
                Description="Allow publishing outside of Visual Studio, such as via &quot;dotnet publish&quot;."
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -27,6 +27,13 @@
...
+                HelpUrl="https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-targets"
```
drewnoakes, line 34: This URL is not very helpful. If there are no docs for this, I'd remove this URL and file an issue on the docs repo to add a description for this property, tagging the relevant folks to chime in with a correct description of it.

drewnoakes (changes requested):
Publishing and packaging are different things. This property does not belong on the packaging page.

Unless we expect several other properties related to publishing, I suggest creating a "Packaging" sub-category on the "Application" page for this property.

tmeschter:
So `dotnet publish` doesn't do anything unless you set `IsPublishable`?

MiYanni:
@tmeschter Correct. The [GitHub thread](https://github.com/dotnet/docs/issues/13365#issuecomment-512152483) I linked to was specifically for excluding projects when using `dotnet publish` against a solution file. The result was to set `IsPublishable` to false and that disables the Publish target from running, which is what `dotnet publish` runs.
image

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
MiYanni: I based the text on the Bool property above it, which looks like:
```xml
  <BoolProperty Name="GeneratePackageOnBuild"
                DisplayName="Generate NuGet package on build"
                Description="Produce a package file during build operations."
                Category="General" />
```
I remembered the 'action verb' style from previous conversations, but I applied it to the DisplayName and not the Description. :-(

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
MiYanni: Created an issue. Will link it in a comment for this member. https://github.com/MicrosoftDocs/visualstudio-docs/issues/6643

MiYanni: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
MiYanni: Updated the text.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -209,4 +209,10 @@
       </NameValuePair>
     </StringProperty.Metadata>
   </StringProperty>
+
+  <!-- TODO Documentation issue for this property: https://github.com/MicrosoftDocs/visualstudio-docs/issues/6643 -->
+  <BoolProperty Name="IsPublishable"
+                DisplayName="Publishing"
+                Description="Allow publishing outside of Visual Studio, such as via &quot;dotnet publish&quot;."
+                Category="Packaging" />
```
drewnoakes, line 217: Please also add a `Packaging` category at the start of this file.

MiYanni: Derp. I knew I missed something. Added.

drewnoakes (approved):

tmeschter (approved):

MiYanni:
The `license/cla` thing was stuck, but all the other checks passed. So, I just merged it.

PR #7170, opened 2021-05-03
Rename types

tmeschter:
Rename a few types related to the Project Query API to reflect that they are used by both Property Pages and Launch Profiles.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7170)


melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/CategoryFromRuleDataProducer.cs
```diff
@@ -9,19 +9,20 @@
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 {
     /// <summary>
-    /// Handles retrieving a set of <see cref="ICategory"/>s from an <see cref="IPropertyPage"/>.
+    /// Handles retrieving a set of <see cref="ICategory"/>s from an <see cref="IPropertyPage"/>
+    /// or <see cref="ILaunchProfile"/>.
     /// </summary>
-    internal class CategoryFromPropertyPageDataProducer : QueryDataFromProviderStateProducerBase<PropertyPageProviderState>
+    internal class CategoryFromRuleDataProducer : QueryDataFromProviderStateProducerBase<ContextAndRuleProviderState>
```
MiYanni, line 15: Just a Q. What is with the 'From' naming semantic? Wouldn't `RuleCategoryDataProducer` work just as well? Before, it would have been `PropertyPageCategoryDataProducer` which also works.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ContextAndRuleProviderState.cs
```diff
@@ -1,17 +1,18 @@
...
+    /// Hold a <see cref="Rule"/> and the context in which to bind it. Used by 
+    /// <see cref="LaunchProfileDataProducer"/> and <see cref="PropertyPageDataProducer"/>
+    /// to pass the state their child producers will need, but allows the actual binding
+    /// of the <see cref="Rule"/> to be delayed until needed.
     /// </summary>
-    internal sealed class PropertyPageProviderState
+    internal sealed class ContextAndRuleProviderState
     {
-        public PropertyPageProviderState(IPropertyPageQueryCache cache, QueryProjectPropertiesContext context, Rule rule)
+        public ContextAndRuleProviderState(IPropertyPageQueryCache cache, QueryProjectPropertiesContext context, Rule rule)
```
MiYanni, line 15: Similarly, wouldn't `RuleContextProviderState` work just as well?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/CategoryFromRuleDataProducer.cs
tmeschter: I wanted to emphasize that the category was being produced from the `Rule`--that is, there's a hierarchy in play here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ContextAndRuleProviderState.cs
tmeschter: The `*ProviderState` types are really just property bags holding all the data we may need or want to pass to some dependent data producer. Naming these types is hard, because they don't really have a meaning or purpose beyond being a container.

For a while I didn't bother with types; I just passed tuples around. Then I started running into problems when I changed the shape of the tuple created by the producer, but not the shape expected by the consumer. Explicit types prevent that kind of problem. But naming them is hard. :-)

PR #7172, opened 2021-05-03
Fix "Startup object" UI

tmeschter:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1320092.

The property pages for executables include a "Startup object" property where you can specify which `Main` method or `Form` to run at startup. In both the old and new UI, the drop down displays all applicable methods and forms. The old UI includes a "(Not set)" item representing the empty value; however, the new UI does not. If you set it to one of the other options there is no way to unset it.

The old UI gets the values for the drop down from the `StartupObjectsEnumProvider`, but then explicitly adds in the "(Not set)" value in the UI code. Since the new UI is generated, this option isn't available to us. Instead, we need `StartupObjectsEnumProvider` to include "(Not set)" along with all the other values, but only when the new UI is in use.

To achieve this we make use of the `DynamicEnumProperty.ProviderSettings`--a bag of name/value pairs that is passed along to the enum provider. ApplicationPropertyPage.xaml is updated to pass along an "IncludeEmptyValue" setting, and `StartupObjectsEnumProvider` is updated to check for the setting.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7172)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
@@ -1,6 +1,5 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
```diff
@@ -31,7 +30,18 @@ public StartupObjectsEnumProvider([Import(typeof(VisualStudioWorkspace))] Worksp
...
+            // We only include a value representing the "not set" state if requested. This is
+            // because the old property pages explicitly add the "(Not set)" value at the UI
+            // layer; the new property pages do not have that option and so the value must come
+            // from the enum provider.
+            // When this project system no longer needs to support the old property pages we can
+            // remove this and always include the "(Not set)" value.
+            bool includeEmptyValue = options?.Any(pair =>
+                pair.Name == "IncludeEmptyValue"
+                && bool.TryParse(pair.Value, out bool optionValue)
+                && optionValue) ?? false;
```
drewnoakes, line 42: Why not update the legacy pages to not add this item?

drewnoakes (approved):
Why not

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
tmeschter: The legacy property pages need to work with both the legacy project system _and_ dotnet/project-system. If this item isn't added at the UI level, then I would need to update the legacy project system to add it, just like I've done here. I don't want to touch either if I can avoid it, especially when most of the complexity I've added here (around whether or not to add the "(Not set)" item) can be removed in the Dev17 time frame.

PR #7174, opened 2021-05-04
Keep the order of command line arguments for build.

ocallesp:
What are the changes:
Get the command line arguments from IProjectBuildSnapshotService instead of IProjectSubscriptionService.

Subscribe to the ProjectBuildSnapshotService directly using the build targets, instead of rules.

What is the bug:
Getting the data from the rules changes the order of the commands which causes the Build to behave differently.

How this change fixes the bug:
Get the original order from ProjectBuildSnapshotService and pass it directly to roslyn.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7174)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.Build;
 using Microsoft.VisualStudio.ProjectSystem.Logging;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 
@@ -57,7 +58,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             EnsureInitialized();
         }
 
-        public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, ContextState state, CancellationToken cancellationToken)
+        public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, 
+            IProjectVersionedValue<IProjectBuildSnapshot> buildSnapshot,
+            ContextState state, 
+            CancellationToken cancellationToken)
         {
             Requires.NotNull(update, nameof(update));
 
@@ -69,9 +73,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             {
                 IComparable version = GetConfiguredProjectVersion(update);
 
+                ProcessOptions(buildSnapshot.Value);
                 await ProcessCommandLineAsync(version, projectChange.Difference, state, cancellationToken);
                 ProcessProjectBuildFailure(projectChange.After);
             }
+        }
+
+        private void ProcessOptions(IProjectBuildSnapshot buildSnapshot)
+        {
+            Assumes.NotNull(_context);
+
+            List<string> commandLineArguments = new List<string>();
+
+            buildSnapshot.TargetOutputs.TryGetValue("CompileDesignTime", out var keyValuePairsCommandLineOptions);
+
+            foreach (var kp in keyValuePairsCommandLineOptions)
+            {
+                commandLineArguments.Add(kp.Key);
+            }
+
+            // We just need to pass all options to Roslyn
+            _context.SetOptions(commandLineArguments.ToImmutableArray());
         }
 
         public Task ApplyProjectEvaluationAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, ContextState state, CancellationToken cancellationToken)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
@@ -69,7 +69,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             {
                 IComparable version = GetConfiguredProjectVersion(update);
 
-                ProcessOptions(projectChange.After);
                 await ProcessCommandLineAsync(version, projectChange.Difference, state, cancellationToken);
                 ProcessProjectBuildFailure(projectChange.After);
             }
@@ -146,12 +145,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             }
         }
 
-        private void ProcessOptions(IProjectRuleSnapshot snapshot)
+        public void SetCommandLineArgumentsToBuild(IEnumerable<string> arguments)
         {
             Assumes.NotNull(_context);
 
             // We just pass all options to Roslyn
-            _context.SetOptions(snapshot.Items.Keys.ToImmutableArray());
+            _context.SetOptions(arguments.ToImmutableArray());
         }
 
         private Task ProcessCommandLineAsync(IComparable version, IProjectChangeDiff differences, ContextState state, CancellationToken cancellationToken)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
@@ -20,7 +20,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
     /// </summary>
     /// <remarks>
     ///     This class is not thread-safe and it is up to callers to prevent overlapping of calls to
-    ///     <see cref="ApplyProjectBuildAsync(IProjectVersionedValue{IProjectSubscriptionUpdate}, ContextState, CancellationToken)"/> and
+    ///     <see cref="ApplyProjectBuildAsync(IProjectVersionedValue{IProjectSubscriptionUpdate}, IProjectBuildSnapshot, ContextState, CancellationToken)"/> and
     ///     <see cref="ApplyProjectEvaluationAsync(IProjectVersionedValue{IProjectSubscriptionUpdate}, ContextState, CancellationToken)"/>.
     /// </remarks>
     [Export(typeof(IApplyChangesToWorkspaceContext))]
@@ -59,7 +59,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         }
 
         public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, 
-            IProjectVersionedValue<IProjectBuildSnapshot> buildSnapshot,
+            IProjectBuildSnapshot buildSnapshot,
             ContextState state, 
             CancellationToken cancellationToken)
         {
@@ -73,7 +73,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
             {
                 IComparable version = GetConfiguredProjectVersion(update);
 
-                ProcessOptions(buildSnapshot.Value);
+                ProcessOptions(buildSnapshot);
                 await ProcessCommandLineAsync(version, projectChange.Difference, state, cancellationToken);
                 ProcessProjectBuildFailure(projectChange.After);
             }
@@ -83,17 +83,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
         {
             Assumes.NotNull(_context);
 
-            List<string> commandLineArguments = new List<string>();
+            buildSnapshot.TargetOutputs.TryGetValue(
+                "CompileDesignTime",
+                out IImmutableList<KeyValuePair<string, IImmutableDictionary<string, string>>> targetOutputs);
 
-            buildSnapshot.TargetOutputs.TryGetValue("CompileDesignTime", out var keyValuePairsCommandLineOptions);
+            var options = ImmutableArray.CreateBuilder<string>(targetOutputs.Count);
 
-            foreach (var kp in keyValuePairsCommandLineOptions)
+            foreach ((string option, _) in targetOutputs)
             {
-                commandLineArguments.Add(kp.Key);
+                options.Add(option);
             }
 
             // We just need to pass all options to Roslyn
-            _context.SetOptions(commandLineArguments.ToImmutableArray());
+            _context.SetOptions(options .MoveToImmutable());
         }
 
         public Task ApplyProjectEvaluationAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, ContextState state, CancellationToken cancellationToken)
@@ -165,14 +167,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
                 _logger.WriteLine(succeeded ? "Last design-time build succeeded, turning semantic errors back on." : "Last design-time build failed, turning semantic errors off.");
                 _context.LastDesignTimeBuildSucceeded = succeeded;
             }
-        }
-
-        public void SetCommandLineArgumentsToBuild(IEnumerable<string> arguments)
-        {
-            Assumes.NotNull(_context);
-
-            // We just pass all options to Roslyn
-            _context.SetOptions(arguments.ToImmutableArray());
         }
 
         private Task ProcessCommandLineAsync(IComparable version, IProjectChangeDiff differences, ContextState state, CancellationToken cancellationToken)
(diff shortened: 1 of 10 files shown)


ocallesp:
@drewnoakes this reverts the previous fix, and also make the changes as you suggested.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -146,19 +145,12 @@ private void ProcessProjectBuildFailure(IProjectRuleSnapshot snapshot)
...
-            // The order of options matters and we should use warnaserror- and then warnaserror+
-            // Depending which comes first, the compiler produces different diagnostics.
-            // We just need to move warnaserror+ to the end.
-            var warnaserrorPlus = snapshot.Items.FirstOrDefault(a => a.Key.Contains("/warnaserror+"));
-            IEnumerable<string> commandLineArguments = warnaserrorPlus.Key is null ? snapshot.Items.Keys :
-                    snapshot.Items.Remove(warnaserrorPlus.Key).Keys.Append(warnaserrorPlus.Key);
-
             // We just pass all options to Roslyn
-            _context.SetOptions(commandLineArguments.ToImmutableArray());
+            _context.SetOptions(arguments.ToImmutableArray());
```
drewnoakes, line 153: Please do the revert in a separate PR, to make this easier to review.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -103,6 +108,16 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
                             linkOptions: DataflowOption.PropagateCompletion,
                             cancellationToken: cancellationToken),
 
+                    ProjectDataSources.SyncLinkTo(
```
davkean, line 111: This needs to be sync-linked with the ProjectBuildRuleSource, otherwise, you are going to be pushing changes to Roslyn that are out of sync in version. ie The command-line arguments will be out of sync with the other changes we're pushing to Roslyn, such as source files, references, etc.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IApplyChangesToWorkspaceContext.cs
```diff
@@ -83,5 +83,14 @@ internal interface IApplyChangesToWorkspaceContext
...
         Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, ContextState state, CancellationToken cancellationToken);
+
+        /// <summary>
+        ///     Sets project build command line options to the underlying <see cref="IWorkspaceProjectContext"/>.
+        /// </summary>
+        /// <param name="arguments">List of command line arguments in the Build</param>
+        /// <remarks>
+        ///     Note: The compiler is sensitive to ordering of command-line arguments.
+        /// </remarks>
+        void SetCommandLineArgumentsToBuild(IEnumerable<string> arguments);
```
drewnoakes, line 94: Rather than adding a new member here, it'd be better to pass the project build snapshot as part of `ApplyProjectBuildAsync`. If you look at `ApplyProjectChangesUnderLockAsync` you will see it is called within a lock, and does a bunch of other scheduling (eg. start/end batch). Making another method here and calling it out of sync with `ApplyProjectBuildAsync` will lead to data races and potentially other kinds of issues.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -103,6 +108,16 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
...
+                        _activeConfiguredProjectProvider.ActiveConfiguredProjectBlock.SyncLinkOptions(),
+                        _projectBuildSnapshotService.SourceBlock.SyncLinkOptions(),
+                        target: DataflowBlockFactory.CreateActionBlock<IProjectVersionedValue<ValueTuple<ConfiguredProject, IProjectBuildSnapshot>>>(e =>
+                                ApplyBuildSnapshotChanged(e),
+                            _project.UnconfiguredProject,
+                            ProjectFaultSeverity.LimitedFunctionality),
+                        linkOptions: DataflowOption.PropagateCompletion,
+                        cancellationToken: cancellationToken),
```
drewnoakes, line 119: Instead of creating a new link here, add the project build snapshot into the above build link.

The new data source must be sync-linked with other sources to avoid races.

`OnProjectChangedAsync` is called for both evaluation and build results. You will likely need to add another member to the `ValueTuple` for the new data, and have it as `null` for evaluation data. Then you can pass the `IProjectBuildSnapshot` through to `ApplyProjectBuildAsync`.

drewnoakes (changes requested):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -162,6 +177,20 @@ public async Task<T> OpenContextForWriteAsync<T>(Func<IWorkspaceProjectContextAc
                 }
             }
 
+            private void ApplyBuildSnapshotChanged(IProjectVersionedValue<(ConfiguredProject project, IProjectBuildSnapshot buildSnapshot)> update)
+            {
+                List<string> commandLineArguments = new List<string>();
+
+                update.Value.buildSnapshot.TargetOutputs.TryGetValue("CompileDesignTime", out var keyValuePairsCommandLineOptions);
```
davkean, line 184: Please make this way more efficient; as spoken about earlier this is very sensitive to allocations and LOH allocations, the List above will be resized many times over. I think we want a more efficient way to go from the IEnumerable<string> -> ImmutableArray.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IApplyChangesToWorkspaceContext.cs
ocallesp: Done, please latest commit "Sync-link buildSnapshot with the ProjectBuildRuleSource."

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
ocallesp: main already has the revert.

This PR is a rebase on main having the revert.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
ocallesp: Done. This is implemented in the commit "Sync-link buildSnapshot with the ProjectBuildRuleSource"

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -57,7 +58,10 @@ public void Initialize(IWorkspaceProjectContext context)
             EnsureInitialized();
         }
 
-        public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, ContextState state, CancellationToken cancellationToken)
+        public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscriptionUpdate> update, 
+            IProjectVersionedValue<IProjectBuildSnapshot> buildSnapshot,
```
drewnoakes, line 62: This value does not need to be versioned, which will simplify calling code slightly.

```suggestion
            IProjectBuildSnapshot buildSnapshot,
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -69,11 +73,29 @@ public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscrip
...
+
+            buildSnapshot.TargetOutputs.TryGetValue("CompileDesignTime", out var keyValuePairsCommandLineOptions);
+
+            foreach (var kp in keyValuePairsCommandLineOptions)
+            {
+                commandLineArguments.Add(kp.Key);
+            }
+
+            // We just need to pass all options to Roslyn
+            _context.SetOptions(commandLineArguments.ToImmutableArray());
```
drewnoakes, line 96: To reduce allocations (written in a browser, so untested):

```suggestion
            buildSnapshot.TargetOutputs.TryGetValue(
                "CompileDesignTime",
                out IImmutableList<KeyValuePair<string, IImmutableDictionary<string, string>>> targetOutputs);

            var options = ImmutableArray.CreateBuilder<string>(targetOutputs.Count);

            foreach ((string option, _) targetOutputs)
            {
                options .Add(option);
            }

            // We just need to pass all options to Roslyn
            _context.SetOptions(options .MoveToImmutable());
```

By using a builder with a predefined capacity, and adding exactly that number of items, we can use `MoveToImmutable` which moves the backing array from the builder to the `ImmutableArray<string>`. This means there is only a single array allocated during this operation.

You need to handle the case where `TryGetValue` returns false, as then `targetOutputs` will be `null`, resulting in an exception.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -177,32 +168,19 @@ public async Task<T> OpenContextForWriteAsync<T>(Func<IWorkspaceProjectContextAc
...
 
-            private async Task ApplyProjectChangesUnderLockAsync(IProjectVersionedValue<(ConfiguredProject project, IProjectSubscriptionUpdate subscription)> update, bool evaluation, CancellationToken cancellationToken)
+            private async Task ApplyProjectChangesUnderLockAsync(IProjectVersionedValue<(ConfiguredProject project, IProjectSubscriptionUpdate subscription, IProjectBuildSnapshot buildSnapshot)> update, bool evaluation, CancellationToken cancellationToken)
             {
                 // NOTE we cannot call CheckForInitialized here, as this method may be invoked during initialization
                 Assumes.NotNull(_contextAccessor);
 
                 IWorkspaceProjectContext context = _contextAccessor.Context;
                 IProjectVersionedValue<IProjectSubscriptionUpdate> subscription = update.Derive(u => u.subscription);
+                IProjectVersionedValue<IProjectBuildSnapshot> buildSnapshot = update.Derive(u => u.buildSnapshot);
```
drewnoakes, line 183: As mentioned earlier, we don't need a versioned value for the build snapshot, so we can avoid the `Derive` here and reduce allocations.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
MiYanni: If you know the size beforehand, what is the advantage of using an `ImmutableArray<T>.Builder` over just using a normal array, populating it, and then calling `ImmutableArray<T>.Create()`? It seems more straightforward that way.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -69,12 +73,31 @@ public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscrip
...
 
+        private void ProcessOptions(IProjectBuildSnapshot buildSnapshot)
+        {
+            Assumes.NotNull(_context);
+
+            buildSnapshot.TargetOutputs.TryGetValue(
+                "CompileDesignTime",
+                out IImmutableList<KeyValuePair<string, IImmutableDictionary<string, string>>> targetOutputs);
+
+            var options = ImmutableArray.CreateBuilder<string>(targetOutputs.Count);
```
davkean, line 90: What makes TargetsOuputs.Count a good capacity here?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -146,19 +169,12 @@ private void ProcessProjectBuildFailure(IProjectRuleSnapshot snapshot)
             }
         }
 
-        private void ProcessOptions(IProjectRuleSnapshot snapshot)
+        public void SetCommandLineArgumentsToBuild(IEnumerable<string> arguments)
```
davkean, line 172: When is this method called?

davkean:
This looks much better. Please make sure you test this heavily and ensure that the bug you are fixing is fixed, plus we didn't regress passing options to Roslyn (such as Version, conditional compilation symbols, etc).

davkean (approved):
Assuming SetCommandLineArgumentsToBuild is being removed, this looks good.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
ocallesp: this was fixed, this is the code cleanup

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
drewnoakes: Yes, in order to `Move` we need to get the count right 100% of the time, otherwise it'll throw. There isn't a more efficient way than this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
```diff
@@ -69,12 +73,31 @@ public async Task ApplyProjectBuildAsync(IProjectVersionedValue<IProjectSubscrip
...
+
+            var options = ImmutableArray.CreateBuilder<string>(targetOutputs.Count);
+
+            foreach ((string option, _) in targetOutputs)
+            {
+                options.Add(option);
+            }
+
+            // We just need to pass all options to Roslyn
+            _context.SetOptions(options .MoveToImmutable());
```
drewnoakes, line 98:
```suggestion
            _context.SetOptions(options.MoveToImmutable());
```

drewnoakes (approved):
Agree that `SetCommandLineArgumentsToBuild` looks like it can be removed. Other than that, this looks like a good change (I haven't tested it locally).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
ocallesp: this method was removed

ocallesp:
cc. @jcouv

jcouv:
Awesome. Thanks much!
FYI @RikkiGibson

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/ApplyChangesToWorkspaceContext.cs
drewnoakes: @MiYanni I never replied.

`ImmutableArray<T>.Create(T[])` will copy the contents of the passed array to a new array, resulting in extra work and allocation. It does this copy so that you cannot modify the contents of the array. `ImmutableArray<>` is a wrapper around an array, and it is designed to make it very hard to modify the contents of the array, as the type's name would suggest.

The builder can encapsulate the array, so no one outside can modify it without its knowledge (via its API). The `MoveToImmutable` method "moves" the builder's array directly into the `ImmutableArray<>`.

PR #7180, opened 2021-05-05
Version launch profiles, and pass versions to the Project Query API

tmeschter:
The first commit adds versioning of launch profiles. The change introduces a new interface, `IVersionedLaunchSettings`, that returns a version number. The `LaunchSettings` type is updated to implement the new interface, and `LaunchSettingsProvider` now supplies and increments the version every time it produces a new instance of `LaunchSettings`.

The second commit uses these versions as the basis of an `IQueryDataSourceVersionProvider`. The `IQueryDataSourceVersionProvider` is responsible for informing the Project Query engine that there have been updates to data sources of interest to queries. When a client is observing a query (meaning they are interested in receiving push updates to the query data), the engine subscribes to the appropriate `IQueryDataSourceVersionProvider`s so it knows when it needs to re-run the query to get updated data and send it to the client.

There are three main parts to this implementation:

1. At the `UnconfiguredProject` level, the `LaunchSettingsTracker` continually watches for changes to the set of launch settings. As they change, it notifies the `LaunchSettingsQueryVersionProvider`.
2. The `LaunchSettingsQueryVersionProvider` is a singleton that aggregates the data from the trackers to create the total set of launch profile versions that may be needed by the Project Query API. It passes the aggregated data on to `LaunchSettingsQueryVersionProviderExport`.
3. `LaunchSettingsQueryVersionProviderExport` is the Project Query-facing counterpart to `LaunchSettingsQueryVersionProvider`. Logically, they work together as a single unit. However, we don't want the `LaunchSettingsQueryVersionProvider` to depend directly on types from the Project Query API as that would cause the query assemblies to load even in scenarios where we aren't using them. To avoid this we need the functionality into these two types, with communication occuring via the `ILaunchSettingsVersionPublisher` interface. The "export" type manages the set of subscribers and asynchronously passes along new version information.

The last piece, to come in a future PR, is to report version information for launch profiles along with query data. The query engine checks those reported versions against the versions supplied by the `IQueryDataSourceVersionProvider` to know when the query needs to be re-run.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProvider.cs
@@ -1,0 +1,88 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
+{
+    /// <summary>
+    /// Aggregates version information from the multiple <see cref="LaunchSettingsTracker" />s
+    /// and passes it on to the bound <see cref="ILaunchSettingsVersionPublisher"/>, if any.
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="LaunchSettingsTracker"/> loads automatically for any project
+    /// supporting launch profiles. However, we don't want to load all of the assemblies
+    /// associated with the Project Query API until they are actually needed, so <see cref="LaunchSettingsTracker"/>
+    /// can't refer to any of them directly. Instead, it communicates with this type,
+    /// which will pass on the version information only after being bound to an <see cref="ILaunchSettingsVersionPublisher"/>.
+    /// </remarks>
+    [Export]
+    internal sealed class LaunchSettingsQueryVersionProvider
+    {
+        private readonly object _lock = new();
+
+        private readonly Dictionary<string, LaunchSettingsTracker> _allLaunchSettingsTrackers = new(StringComparer.Ordinal);
+        private ImmutableDictionary<string, long> _versions = ImmutableDictionary<string, long>.Empty;
+
+        private ILaunchSettingsVersionPublisher? _versionPublisher;
+
+        /// <summary>
+        /// Sets the related <see cref="ILaunchSettingsVersionPublisher"/> and pushes
+        /// the current version information to it.
+        /// </summary>
+        internal void BindToVersionPublisher(ILaunchSettingsVersionPublisher versionPublisher)
+        {
+            lock (_lock)
+            {
+                _versionPublisher = versionPublisher;
+                _versionPublisher.UpdateVersions(_versions);
+            }
+        }
+
+        /// <summary>
+        /// Called when the <paramref name="tracker"/> becomes active.
+        /// </summary>
+        internal void OnLaunchSettingsTrackerActivated(LaunchSettingsTracker tracker)
+        {
+            lock (_lock)
+            {
+                string key = tracker.VersionKey;
+
+                _allLaunchSettingsTrackers.Add(key, tracker);
+                _versions = _versions.SetItem(key, tracker.CurrentVersion);
+
+                _versionPublisher?.UpdateVersions(_versions);
+            }
+        }
+
+        /// <summary>
+        /// Called when the <paramref name="tracker"/> has an updated version.
+        /// </summary>
+        internal void OnLaunchSettingsVersionUpdated(LaunchSettingsTracker tracker)
+        {
+            lock (_lock)
+            {
+                _versions = _versions.SetItem(tracker.VersionKey, tracker.CurrentVersion);
+                _versionPublisher?.UpdateVersions(_versions);
+            }
+        }
+
+        /// <summary>
+        /// Called when the <paramref name="tracker"/> is deactivated.
+        /// </summary>
+        internal void OnLaunchSettingsTrackerDeactivated(LaunchSettingsTracker tracker)
+        {
+            lock (_lock)
+            {
+                string key = tracker.VersionKey;
+
+                _versions = _versions.Remove(key);
+                _allLaunchSettingsTrackers.Remove(key);
+
+                _versionPublisher?.UpdateVersions(_versions);
+            }
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProviderExport.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProviderExport.cs
@@ -1,0 +1,218 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks.Dataflow;
+using Microsoft.VisualStudio.ProjectSystem.Query;
+using Microsoft.VisualStudio.ProjectSystem.Query.Providers;
+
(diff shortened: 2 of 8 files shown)


tmeschter:
Further notes:

- I'm not entirely happy with how the versioning is implemented here. For one thing, nothing below the Query API actually cares about launch settings versions. However, the future work to report the version used during a query may need to map from a specific set of launch settings to their version number; this is much easier if the relevant data is baked into the launch profile.
- The launch settings provider is chatty. When opening a project it pushes as many as ten copies of the launch settings through data flow, where I would have expected closer to two or three. That means more work for the `IQueryDataSourceVersionProvider` but also for every other component subscribed to that dataflow. If I have time I will look into why it is creating so many copies of the launch settings.

jjmew (approved):
:shipit:

drewnoakes (approved):

ocallesp (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsTracker.cs
```diff
@@ -0,0 +1,107 @@
...
+            UnconfiguredProject project,
+            LaunchSettingsQueryVersionProvider versionProvider)
+        {
+            _project = project;
+            _versionProvider = versionProvider;
+        }
+
+        public Task LoadAsync()
+        {
+            ILaunchSettingsProvider launchSettingsProvider = _project.Services.ExportProvider.GetExportedValue<ILaunchSettingsProvider>();
```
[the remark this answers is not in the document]
tmeschter: I'm no longer sure. I don't think there's a good reason; I'll import it via the constructor.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsTracker.cs
```diff
@@ -0,0 +1,107 @@
...
+                _launchSettingsProviderLink.Dispose();
+                _launchSettingsProviderLink = null;
+            }
+
+            _versionProvider.OnLaunchSettingsTrackerDeactivated(this);
+
+            return Task.CompletedTask;
+        }
+
+        internal void OnLaunchSettingsChanged(ILaunchSettings newSettings)
```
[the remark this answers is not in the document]
tmeschter: Yeah, a bunch of these methods can be `private`, or `public` if they should be generally accessible.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// Called when the <paramref name="tracker"/> becomes active.
+        /// </summary>
+        internal void OnLaunchSettingsTrackerActivated(LaunchSettingsTracker tracker)
+        {
+            lock (_lock)
+            {
+                string key = tracker.VersionKey;
+
+                _allLaunchSettingsTrackers.Add(key, tracker);
+                _versions = _versions.SetItem(key, tracker.CurrentVersion);
```
MiYanni, line 54: Does `SetItem` return a different instance? Is that why you need to reassign? (I notice this pattern applies to other methods too)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProviderExport.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// </remarks>
+        private QueryDataVersions _latestPostedVersions = QueryDataVersions.Empty;
+
+        [ImportingConstructor]
+        public LaunchSettingsQueryVersionProviderExport(LaunchSettingsQueryVersionProvider provider)
+        {
+            _provider = provider;
+#pragma warning disable RS0030 // Do not used banned APIs
+            _processingBlock = DataflowBlockSlim.CreateActionBlock<Action>(callback => callback(), nameFormat: nameof(LaunchSettingsQueryVersionProviderExport));
+#pragma warning restore RS0030 // Do not used banned APIs
```
MiYanni, line 63: Why is this API banned?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProviderExport.cs
```diff
@@ -0,0 +1,218 @@
...
+                _processingBlock.Post(() =>
+                {
+                    bool observerRequiresNotification = false;
+                    lock (SyncObject)
+                    {
+                        if (_observers.Contains(observer)
+                            && processedNotificationCountAtTimeOfSubscription == _processedNotificationCount)
+                        {
+                            observerRequiresNotification = true;
+                        }
```
MiYanni, line 108: Instead of an if-statement and then assigning `true`, couldn't you just assign it the predicate directly?
```suggestion
                        observerRequiresNotification = _observers.Contains(observer) && processedNotificationCountAtTimeOfSubscription == _processedNotificationCount);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProviderExport.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                        List<string>? expiredSources = null;
+                        foreach (string versionKey in _latestPostedVersions.Keys)
+                        {
+                            if (!newVersions.ContainsKey(versionKey))
+                            {
+                                if (expiredSources is null)
+                                {
+                                    expiredSources = new List<string>();
+                                }
```
MiYanni, line 179: nit: shorthand is:
```suggestion
                                expiredSources ??= new List<string>();
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -987,5 +990,12 @@ internal async Task<string> GetLaunchSettingsFilePathNoCacheAsync()
 
             return Path.Combine(folder, LaunchSettingsFilename);
         }
+
+        private long GetNextVersion() => _nextVersion++;
+
+        /// <summary>
+        /// Protected method for testing purposes.
+        /// </summary>
+        protected void SetNextVersion(long nextVersion) => _nextVersion = nextVersion;
```
MiYanni, line 999: 😁

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProviderExport.cs
tmeschter: One of the BannedSymbols.txt files has the answer:
> `Use DataflowBlockFactory.CreateActionBlock to ensure failed blocks are reported as non-fatal watsons and assigned to the correct component`

`DataflowBlockFactory.CreateActionBlock` creates an action block with additional machinery for catching and reporting failures, and possibly recovering. In almost all cases this is preferred over what I am doing here.

However, `DataflowBlockFactory.CreateActionBlock` takes an `UnconfiguredProject` as one of its parameters, but this type (`LaunchSettingsQueryVersionProviderExport`) isn't associated with any one project--so I can't actually use `DataflowBlockFactory.CreateActionBlock`.

I'll update the code to include a comment explaining why we are using the banned API here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchSettingsQueryVersionProvider.cs
MiYanni: If we are mutating an immutable dictionary, could we just use a mutable dictionary? I don't know enough about the code here, but I'm assuming we are just letting the original instance get garbage collected (it becomes unreferenced) at some point for every one of these reassignments. Maybe only create immutable dictionaries out of mutable ones when the interfaces (I'm assuming CPS) require an immutable dictionary.

PR #7188, opened 2021-05-10
Use constant strings for telemetry event names

drewnoakes:
Previously a method was used to compute these values. By converting these to constants, no allocation happens at runtime and we only keep a single copy of these strings in memory.


jjmew (approved):
:shipit:

MiYanni (approved):

Issue #7192, opened 2021-05-10
[Merged Interops] IAsyncServiceProvider 


dmonroym:
`IAsyncServiceProvider` changed from using `QueryServiceAsync` to `GetServiceAsync`. I think my implementation isn't fully correct or the tests need to be changed. As a workaround I disabled unit tests in `VsContainedLanguageComponentsFactoryTests.cs` and they're tagged with `Skip = "TODO: Merged PIAs  - uncomment this once IAsyncServiceProvider is updated"`

dmonroym:
Fixed in #7234

Issue #7191, opened 2021-05-10
[Merged Interops] Microsoft.DotNet.IBCMerge


dmonroym:
We had to change the way we did `/ibc` since now we're fetching `Microsoft.DotNet.IBCMerge` from a new feed which also requires Nuget.config file.

The current solution is perhaps a better method of doing this and given our lack of optprof we weren't really using ibc so it's hard to verify. I'll need to do a deeper analysis on this and clean up the code surrounding this

dmonroym:
We're currently getting this from `<add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />` 

When we attempt to build we get this error:
```
"C:\a\1\s\build\proj\Build.proj" (default target) (1) ->
       "C:\a\1\s\build\proj\CoreBuild.proj" (default target) (2) ->
       "C:\a\1\s\ProjectSystem.sln" (Build;Test;Pack target) (3:2) ->
       "C:\a\1\s\src\Microsoft.VisualStudio.AppDesigner\Microsoft.VisualStudio.AppDesigner.vbproj" (default target) (15:10) ->
       (ApplyOptimizations target) -> 
         C:\Users\AzDevOps\.nuget\packages\roslyntools.repotoolset\1.0.0-beta-62705-01\tools\OptimizationData.targets(51,5): error MSB3073: The command ""C:\Users\AzDevOps\.nuget\packages\microsoft.dotnet.ibcmerge\5.0.7-beta.20159.1\lib\net45\ibcmerge.exe" -q -f -partialNGEN -minify -mo "C:\a\1\s\artifacts\Release\obj\Microsoft.VisualStudio.AppDesigner\Microsoft.VisualStudio.AppDesigner.dll" -incremental "C:\Users\AzDevOps\.nuget\packages\RoslynDependencies.ProjectSystem.OptimizationData\2.6.0-beta1-62205-02\content\OptimizationData\Microsoft.VisualStudio.AppDesigner.pgo"" exited with code 3. [C:\a\1\s\src\Microsoft.VisualStudio.AppDesigner\Microsoft.VisualStudio.AppDesigner.vbproj]
```

tmeschter:
Note that [MicroBuild has support for OptProf](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/700/OptProf) and I strongly recommend we leverage that.

MiYanni:
[...]

Issue #7190, opened 2021-05-10
[Merged interops] Warnings on build


dmonroym:
The current build contains a version of `Microsoft.VisualStudio.ProjectSystem.Query` that causes the following build warning:
```
CSC : warning CS1762: A reference was created to embedded interop assembly 'Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' because of an indirect reference to that assembly created by assembly 'Microsoft.VisualStudio.Pr
ojectSystem.Query, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Consider changing the 'Embed Interop Types' property on either assembly. [G:\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj]
```

The solution is for CPS team to update this binary and then we'll be able to re-add `/warnaserror` into our build.cmd

dmonroym:
Solution consisted in switching `'%(Filename)' == 'Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime'` in `HostAgnostic.props` such that `<EmbedInteropTypes>false</EmbedInteropTypes>`

PR #7193, opened 2021-05-10
Merged Interop work

dmonroym:
This merges all the work associated with merged interops.

It still contains some warnings and there is additional work left to do but the work is stable and this will unblock partners

Opens issues:
#7190 
#7191 
#7192

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7193)


NuGet.config
```diff
@@ -0,0 +1,19 @@
...
+    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
+    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
+    <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="vssdk" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vssdk/nuget/v3/index.json" />
+    <add key="vs-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" />
+    <add key="myget-legacy" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json" />
+    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+  </packageSources>
+  <disabledPackageSources>
```
swesonga, line 16: For my info, which packages require disabledPackageSources to be cleared?

NuGet.config
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information. -->
+<configuration>
```
swesonga, line 3: Can this change not be made to https://github.com/dotnet/project-system/blob/main/build/import/Packages.targets#L5 instead of introducing a NuGet.config file?

dmonroym: Yes, that's part of #7191. I'm trying to get rid of this file and the workarounds surrounding ibc. The file at the time existed only in private feeds so our pipelines couldn't capture it and this file contained other feeds that gave access to it. I'll revert this PR back to a draft and fix that before it's merged

NuGet.config
dmonroym: none. This tag was copied from another file and wasn't cleaned up

NuGet.config
dmonroym: Fixed in #7233

dmonroym:
This PR is substituted by #7233

PR #7196, opened 2021-05-11
 Remove code dependent on CPS ReadOnly property

dmonroym:
Tracking this removal with issue #

We're waiting for CPS to update their package for dev17 containing this. Their PR is here: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/323370


jjmew (approved):
:shipit:

PR #7225, opened 2021-05-17
Remove usage of the 'evaluated' visibility condition function

### LINKED ISSUE #7021: Remove "evaluated" function from visibility conditions
Currently we have a visibility condition function `evaluated` which returns a single evaluated value for a property.

However a project may produce different evaluated values per configuration. Currently, if that happens, the `evaluated` function throws an exception.

We also have an `has-evaluated-value` function which returns true if any of the evaluated value matches an argument. This will never throw the above exception.

https://github.com/dotnet/project-system/blob/main/docs/repo/property-pages/visibility-conditions.md#function-reference

The original idea was that we would use `evaluated` for properties which self-identify as non-configurable (and therefore would only have a single value).

The problem is that a project can still be manually constructed that produces different values across configurations for any property. If we use `evaluated` then there will always be a "valid" project that produces an exception in the UI.

I will remove `evaluated` and switch everything to `has-evaluated-value` instead.

drewnoakes:
Fixes #7021

Currently we have a visibility condition function `evaluated` which returns a single evaluated value for a property.

However a project may produce different evaluated values per configuration. Currently, if that happens, the `evaluated` function throws an exception.

We also have an `has-evaluated-value` function which returns true if any of the evaluated value matches an argument. This will never throw the above exception.

The original idea was that we would use `evaluated` for properties which self-identify as non-configurable (and therefore would only have a single value).

The problem is that a project can still be manually constructed that produces different values across configurations for any property. If we use `evaluated` then there will always be a "valid" project that produces an exception in the UI.

This change removes usage of the `evaluated` function, and removes it from the documentation. Once this has merged and inserted, we can remove the definition of this function from CPS.

I have manually tested all the changed properties here to verify they still work as expected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7225)

### DIFF
--- a/docs/repo/property-pages/visibility-conditions.md
+++ b/docs/repo/property-pages/visibility-conditions.md
@@ -16,7 +16,7 @@ In a XAML rule file, a visibility condition is specified as metadata on the prop
 <StringProperty ...>
   <StringProperty.Metadata>
     <NameValuePair Name="VisibilityCondition">
-      <NameValuePair.Value>(has-evaluated-evalue "MyPage" "MyProperty" "Foo")</NameValuePair.Value>
+      <NameValuePair.Value>(has-evaluated-value "MyPage" "MyProperty" "Foo")</NameValuePair.Value>
     </NameValuePair>
   </StringProperty.Metadata>
 </StringProperty>
@@ -96,7 +96,7 @@ The following table details the default set of visibility expression functions:
 
 These functions are defined in class `VisibilityConditionEvaluator`.
 
-Note that there is no `evaluated` function. A propery may have multiple evaluated values, and as such it's not possible to reliably return a single value. Use `has-evaluated-value` instead.
+Note that there is no `evaluated` function. A property may have multiple evaluated values, and as such it's not possible to reliably return a single value. Use `has-evaluated-value` instead.
 
 ## Adding Functions
 
--- a/docs/repo/property-pages/visibility-conditions.md
+++ b/docs/repo/property-pages/visibility-conditions.md
@@ -16,7 +16,7 @@ In a XAML rule file, a visibility condition is specified as metadata on the prop
 <StringProperty ...>
   <StringProperty.Metadata>
     <NameValuePair Name="VisibilityCondition">
-      <NameValuePair.Value>(eq "Foo" (evaluated "MyPage" "MyProperty"))</NameValuePair.Value>
+      <NameValuePair.Value>(has-evaluated-evalue "MyPage" "MyProperty" "Foo")</NameValuePair.Value>
     </NameValuePair>
   </StringProperty.Metadata>
 </StringProperty>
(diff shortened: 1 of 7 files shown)


drewnoakes:
CPS PR, to be inserted after this: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/324704

docs/repo/property-pages/visibility-conditions.md
```diff
@@ -16,7 +16,7 @@ In a XAML rule file, a visibility condition is specified as metadata on the prop
 <StringProperty ...>
   <StringProperty.Metadata>
     <NameValuePair Name="VisibilityCondition">
-      <NameValuePair.Value>(eq "Foo" (evaluated "MyPage" "MyProperty"))</NameValuePair.Value>
+      <NameValuePair.Value>(has-evaluated-evalue "MyPage" "MyProperty" "Foo")</NameValuePair.Value>
```
melytc, line 19:
```suggestion
      <NameValuePair.Value>(has-evaluated-value "MyPage" "MyProperty" "Foo")</NameValuePair.Value>
```

docs/repo/property-pages/visibility-conditions.md
```diff
@@ -89,14 +89,15 @@ The following table details the default set of visibility expression functions:
...
 | `not`                    | 1        | Computes logical NOT of argument                                                                |
-| `evaluated`              | 2        | Returns the evaluated value of property on page `arg0` with name `arg1`                         |
 | `unevaluated`            | 2        | Returns the unevaluated value of property on page `arg0` with name `arg1`                       |
 | `has-evaluated-value`    | 3        | Returns true if property on page `arg0` with name `arg1` has an evaluated value matching `arg2` |
 | `is-codespaces-client`   | 0        | Returns true if the Project Properties UI is running in a Codespaces client                     |
 | `has-project-capability` | 1        | Returns true if the project has the specified capability.                                       |
 
 These functions are defined in class `VisibilityConditionEvaluator`.
 
+Note that there is no `evaluated` function. A propery may have multiple evaluated values, and as such it's not possible to reliably return a single value. Use `has-evaluated-value` instead.
```
melytc, line 99:
```suggestion
Note that there is no `evaluated` function. A property may have multiple evaluated values, and as such it's not possible to reliably return a single value. Use `has-evaluated-value` instead.
```

melytc (approved):

docs/repo/property-pages/visibility-conditions.md
drewnoakes: Thanks!

docs/repo/property-pages/visibility-conditions.md
drewnoakes: Thanks again!

tmeschter (approved):

PR #7231, opened 2021-05-18
Use new Project Properties UI by default in Dev17

### LINKED ISSUE #7171: Use new Property Pages UI in Dev17 by default
Currently, in Dev16 preview builds the new UI is only displayed if a feature flag is set.

In Dev17 the new UI should be displayed for all relevant projects, without checking the feature flag.

We should remove the feature flag from Dev17 as well (tracked by #7230).

drewnoakes:
Fixes #7171

In Dev16 we only enabled the new UI if a feature flag was enabled.

In Dev17 we will only support this new UI, so no longer need to check the feature flag. The flag will be removed once this change inserts.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7231)

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerEditorFactory.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerEditorFactory.vb
@@ -178,7 +178,7 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
         Implements IVsEditorFactory.CreateEditorInstance
 
             ' If we're using the new project properties editor, delegate to its editor factory
-            Dim shouldUseNewEditor As Boolean = UseNewEditor(Hierarchy)
+            Dim shouldUseNewEditor As Boolean = Hierarchy.IsCapabilityMatch("ProjectPropertiesEditor")
 
             Common.TelemetryLogger.LogEditorCreation(shouldUseNewEditor, FileName, PhysicalView)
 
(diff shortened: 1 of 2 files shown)


melytc (approved):

swesonga (approved):

dmonroym (approved):

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerEditorFactory.vb
```diff
@@ -229,17 +228,9 @@ Namespace Microsoft.VisualStudio.Editors.ApplicationDesigner
...
-            End If
-
-            Dim featureFlags = TryCast(_siteProvider.GetService(GetType(SVsFeatureFlags)), IVsFeatureFlags)
-
-            Return _
-                featureFlags IsNot Nothing AndAlso _
-                featureFlags.IsFeatureEnabled("CPS.UpdatedProjectPropertiesDesigner.Local", False)
+        Private Shared Function UseNewEditor(vsHierarchy As IVsHierarchy) As Boolean
+            ' The new editor is only available for CPS-based C# projects
+            Return vsHierarchy.IsCapabilityMatch("CPS & CSharp")
```
tmeschter, line 233: I wonder if we should check for a capability explicitly related to property pages, and then we can "light up" the property pages in other languages/project types just by adding that capability when it is ready. E.g. when the VB property page changes are ready VB projects add that capability.

tmeschter (approved):

MiYanni (approved):

src/Microsoft.VisualStudio.AppDesigner/ApplicationDesigner/ApplicationDesignerEditorFactory.vb
drewnoakes: I like this idea, thanks. I've added a new `ProjectPropertiesEditor` capability, and opted C# projects in by default.

PR #7233, opened 2021-05-18
Update main to include merged interop changes

### LINKED ISSUE #7191: [Merged Interops] Microsoft.DotNet.IBCMerge

dmonroym:
This merges all the work associated with merged interops and brings up to parity what is in 17.0 Preview 1. This also brings the package version up to `17.0.1` to be able to differentiate between P1 and P2 packages.

I decided to track the issues this opens separately and fix those afterwards. It opens:
 - #7191: The `/ibc` option is disabled and set to `/no-ibc`. Currently this doesn't affect our build since we don't use optprof but I had to introduce some changes surrounding the build of `RestoreIBCMerge.csproj` including temporarily adding a Nuget.Config file. This file is solely used for this project as we need to retrieve the package from a private feed so we can't include it in packages.targets
 - #7192 Creates some functional problems and is being addressed. Our tests caught this and they're currently disabled. This is fixed in #7234 and will forward integrate after this is merged as it was made into preview 1.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7233)


dmonroym:
Validated this with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4771137&view=results

build/import/Versions.props
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <!-- IMPORTANT: These assembly versions must always be 3 digits -->
     <MinimumRequiredVSVersion>16.8.1</MinimumRequiredVSVersion> <!-- We should try to keep this up to date with the highest version available on build machines -->
-    <ProjectSystemVersion>16.10.0</ProjectSystemVersion> <!-- This should always match with the VS release we're targeting -->
+    <ProjectSystemVersion>17.0.1</ProjectSystemVersion> <!-- This should always match with the VS release we're targeting -->
```
tmeschter, line 6: Should we bump this to 17.0.2 for Preview 2 and beyond?

src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.vbproj
```diff
@@ -281,9 +281,7 @@
       <DependentUpon>PackagePropPage.vb</DependentUpon>
       <SubType>UserControl</SubType>
     </Compile>
-    <Compile Update="PropPages\PackagePropPage.vb">
-      <SubType>UserControl</SubType>
```
tmeschter, line 285: Do we know why we lost the `SubType`s?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -262,9 +262,9 @@ internal static VsDebugTargetInfo4 GetDebuggerStruct4(IDebugLaunchSettings info)
             {
                 var processStartupInfo = new VsDebugStartupInfo
                 {
-                    hStdInput = unchecked((uint)info.StandardInputHandle.ToInt32()),
-                    hStdOutput = unchecked((uint)info.StandardOutputHandle.ToInt32()),
-                    hStdError = unchecked((uint)info.StandardErrorHandle.ToInt32()),
+                    hStdInput = unchecked((IntPtr)(uint)info.StandardInputHandle.ToInt32()),
```
tmeschter, line 265: Can we go straight from `info.StandardInputHandle` to `IntPtr` without going though Int32 and uint first?

That is, are we able to simply these expressions somehow?

tmeschter (approved):
Approved with comments.

build/import/Versions.props
dmonroym: Preview 1 is 17.0.0 so I just bumped it up one but would it make more sense to go to 17.0.2?

src/Microsoft.VisualStudio.Editors/Microsoft.VisualStudio.Editors.vbproj
dmonroym: Shouldn't have, thanks for catching that. I'll fix in next iteration.

build/import/Versions.props
tmeschter: No, that's fine. Just ensuring that they are unique (and incrementing) is enough. I don't want to have more version numbers than necessary. :-)

src/Microsoft.VisualStudio.AppDesigner/PropPages/PropPageUserControlBase.vb
```diff
@@ -2833,7 +2835,9 @@ NextControl:
 
             'First see if it is already in the project
             For Each ProjectItem As EnvDTE.ProjectItem In ProjectItems
-                If ProjectItem.FileNames(1).Equals(FileName, StringComparison.OrdinalIgnoreCase) Then
+#Disable Warning BC42025 ' Access of shared member, constant member, enum member or nested type through an instance
+                If ProjectItem.get_FileNames(1).Equals(FileName, StringComparison.OrdinalIgnoreCase) Then
+#Enable Warning BC42025 ' Access of shared member, constant member, enum member or nested type through an instance
```
dmonroym, line 2840: These warning wrappers should be removed after #7235 is integrated. Tracked by #7257

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb
```diff
@@ -669,7 +671,9 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
             Get
                 'First see if it is already in the project
                 For Each ProjectItem As ProjectItem In ProjectDesignerProjectItem.ProjectItems
-                    If ProjectItem.FileNames(1).Equals(MyAppFileNameWithPath, StringComparison.OrdinalIgnoreCase) Then
+#Disable Warning BC42025 ' Access of shared member, constant member, enum member or nested type through an instance
```
dmonroym, line 674: Also tracked by #7257

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
dmonroym: Yes, in fact `info.StandardInputHandle` is an `IntPtr` so none of these are needed.

PR #7235, opened 2021-05-18
Fix up VB code

tmeschter:
While the Dev17 interop work was ongoing we needed to make some adjustments to the VB code to make progress, notably referring to some interop methods and properties with different names and making use of `Option Strict Off`. Now that the interop work has settled a bit these are no longer needed and should be backed out.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1327354/.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7235)


dmonroym (approved):

jjmew (approved):
:shipit:

PR #7247, opened 2021-05-19
Report launch profile versions

tmeschter:
This change performs a significant amount of refactoring with the goal of reporting the versions of the launch settings used in queries.

There are some challenges with reporting these versions properly. For background and as a point of comparison, consider properties exposed in the Project Properties UI. The metadata for these properties exists at the unconfigured project level (that is, it comes from `Rule`s in .xaml files that we effectively assume will be the same across all configurations), and the values for the properties exist at the configured project level (so we can support properties that vary from one configuration to another; even where we don't expect a particular property to vary we still retrieve the value from a specific configuration). So the parts of our query implementation that return property metadata should report the version of the `UnconfiguredProject`, while the parts that return values (and the sets of supported values) should report versions from the particular `ConfiguredProject`s that were used.

For the Launch Profiles UI, the metadata about properties is still coming from `Rule`s that are managed by the project, but the actual set of profiles and the property values is generally coming from the launch profiles. Launch profiles are specific to the .NET Project System and are unknown to the `UnconfiguredProject` or `ConfiguredProject`, and do not affect their versioning. So for queries related to Launch Profiles, we need to report the `UnconfiguredProject` version for metadata, but the launch profile version for all values.

The difficulty is that much of our Project Query API implementation just layers on top of the generic CPS properties system, and it abstracts away the underlying data source. When we retrieve a property value, we don't know if it is coming from the project properties or the launch profiles.

To handle this, we now have two implementations of `IProjectState`: one for property pages, and one for launch profiles. They differ only in how they report the "data" version (formerly called the configured project version). We only retrieve the versions through the `IProjectState` in situations where we don't know if we're handling project properties or launch profiles; in cases where we do (such setting values or retrieving the list of property pages/launch profiles) we report the appropriate version directly. In those cases we generally don't have an `IProjectState` yet anyway.

This commit includes the following changes:

1. Reporting the version when adding, removing, renaming, duplicating, and updating launch profiles.
2. Move most of the implementation of `PropertyPageProjectState` down to a base class, `AbstractProjectState`, and adding `LaunchProfileProjectState` as a second implementation of that base class.
3. Rename `GetUnconfiguredProjectVersion` and `GetConfiguredProjectVersionAsync` in `IProjectState` to `GetMetadataVersionAsync` and `GetDataVersionAsync` to reflect their intent.
4. Remove `IProjectStateProvider` entirely. It largely existed to make it easier to inject mock implementations of `IProjectState` in unit testing, and it's not really helping the product code at the moment. We now just create instances of either `PropertyPageProjectState` or `LaunchProfileProjectState` as needed. Most of the code changes are related to this.
5. Remove the `QueryVersionExtensions` type. These extensions were being used inconsistently.
6. Refactor some tests. There were only a few tests that actually used the `IProjectStateProvider` to create a mock `IProjectState`. By updating some of the other mocks to be more functional we no longer need to do that.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7247)


ocallesp (approved):

dmonroym (approved):

tmeschter:
This includes part of the work for #6896.

PR #7259, opened 2021-05-24
Add ProduceReferenceAssembly property to Build section

### LINKED ISSUE #7245: Support the ProduceReferenceAssembly property in the new Project Properties UI
We should investigate the use of the `ProduceReferenceAssembly` property across different types of projects and potentially add it to the UI.

https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies

This property is enabled by default from .NET 5.0 onwards. It may still be useful for earlier targets, and we should offer it.

drewnoakes:
Fixes #7245

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7259)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -274,6 +274,17 @@
   </StringProperty>
 
   <!-- TODO create fwlink for this HelpUrl -->
+  <BoolProperty Name="ProduceReferenceAssembly"
+                DisplayName="Reference assembly"
+                Description="Produce a reference assembly containing the public API of the project."
+                HelpUrl="https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies"
+                Category="Output">
+    <BoolProperty.DataSource>
+      <DataSource HasConfigurationCondition="False" />
+    </BoolProperty.DataSource>
+  </BoolProperty>
+
+  <!-- TODO create fwlink for this HelpUrl -->
   <BoolProperty Name="GenerateDocumentationFile"
                 DisplayName="Documentation file"
                 Description="Generate a file containing API documentation."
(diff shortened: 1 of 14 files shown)


swesonga (approved):
Consider filing a bug for the fwlink issues since there are already other TODOs in the file (or linking the issue if it has already been created)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -273,6 +273,17 @@
...
   </StringProperty>
 
+  <!-- TODO create fwlink for this HelpUrl -->
+  <BoolProperty Name="ProduceReferenceAssembly"
+                DisplayName="Reference assembly"
+                Description="Produce a reference assembly containing the public API of the project."
+                HelpUrl="https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies"
+                Category="Output">
+    <BoolProperty.DataSource>
+      <DataSource HasConfigurationCondition="False" />
```
MiYanni, line 283: Looking at `HasConfigurationCondition`, it might make sense if we flip the default to false so controls need to opt into dimensionality. Also, by the name of this, it isn't obvious it has to do with configuration variants/dimensions. Maybe something like `IsVariedByConfiguration`?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
drewnoakes: I'm open to flipping the default if most properties on this page are not configured. Feel free to submit a PR. You set the default in the `Rule`-level `DataSource`.

`HasConfigurationCondition` is an MSBuild API which has been public for forever and cannot be changed. If you think of the property it creates though, it does end up having a configuration condition (on the property group), so the name does make some sense.

To be clear, the `Rule` object model is used in many places throughout VS.

drewnoakes:
> Consider filing a bug for the fwlink issues since there are already other TODOs in the file (or linking the issue if it has already been created)

Great point, thanks. https://github.com/dotnet/project-system/issues/7262

Issue #7243, opened 2021-05-19
Extend property visibility condition functions (2)


drewnoakes:
Several properties are only applicable for certain project configurations. We need to create more functions for use in visibility conditions.

- [x] `(has-net-framework)`
- [x] `(has-net-core-app)`
- [x] `(has-net-framework-version-or-greater "4.8")`
- [x] `(has-net-core-app-version-or-greater "5.0")`
- [x] `(has-csharp-lang-version-or-greater "7.3")`

drewnoakes:
Looking at this, there are two options.

1. Allow exporting visibility condition functions via MEF, then export the above functions with an implementation that can read from the project state. This would contribute to #6895.

2. Pull in extra project data (`LangVersion`, `TargetFrameworkIdentifier`, etc) into non-visible properties to be able to query then.
    - Requires new functions
      - Comparing version number strings (`LangVersion`) with special handling for values such as `latest`, `preview`, etc
      - Counting the number of strings in a semicolon-delimited list of values (`TargetFrameworks`)
    - Requires implementing support for non-visible properties
      - Server stops filtering non-visible properties
      - Add `IsVisible` metadata to the project model
      - Update client to hide non-visible properties
   - This approach would make for less readable expressions (i.e. `(version-greater-or-equal (unevaluated "Application" "LangVersion") "8.0")` versus `(csharp-lang-version-or-greater "8.0")`. Actually the first expression is incomplete, as it would also need to verify the project is a C# project.

I favour approach one because it's simplest and more flexible in future. The main downside is it brings more CPS code into the client, which originally was something we avoided. We already do this right now for checking project capabilities in visibility conditions, however.

drewnoakes:
https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/325289 adds the ability to export visibility condition functions via MEF, which is required for option 1 above.

drewnoakes:
CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/327513 adds functions that will allow this.

drewnoakes:
These new functions are documented in https://github.com/dotnet/project-system/pull/7282.

drewnoakes:
These have now merged into CPS.

Issue #7251, opened 2021-05-20
Plumb non-visible project properties through to the UI (1)


drewnoakes:
It would be useful for the client to have non-visible properties available for use in visibility conditions.

- [x] Add `IsVisible` to the project model in CPS ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/325905))
- [x] Stop filtering non-visible properties in the .NET Project System ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/327511))
- [x] Filter non-visible properties from the UI ([PR](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/327511))

tmeschter:
@drewnoakes Do you think this is a Must Have for Dev17 GA?

drewnoakes:
@tmeschter yes. It unblocks other must-have scenarios such as #7227.

drewnoakes:
This has been addressed by in-flight PR and can be closed once they merge.

drewnoakes:
All parts have been merged.

PR #7266, opened 2021-05-25
Include invisible properties

tmeschter:
Related to #7243.

On the properties pages back end, the `BaseProperty` type has a `Visible` property that can be `true` or `false` to indicate if the property should be shown in the UI. For the very first implementation of property pages in CodeSpaces (using the C++-style grid) we didn't have the ability to hide a property at the UI layer, so the back end simply filtered out hidden properties from the data it sent to the client. Even as we reworked the new Property Page UI in various ways this implementation detail persisted.

At the same time, we wanted the ability to _conditionally_ show or hide some properties based on the values of other properties; to support this we added the concept of `VisibilityCondition` metadata on the property. So if you wanted a property to be hidden but still available on the client to other properties, you would set the `VisibilityCondition` to `false`.

However, setting the `VisibilityCondition` is rather verbose, since it is stored in the `BaseProperty.Metadata` property bag rather than in a real member of the `BaseProperty` type. For properties that will be unconditionally hidden it would be much simpler to just set `Visible` to false.

In this commit we stop filtering properties based on `Visible`. However, if `Visible` is set to `false` we will force the `VisibilityCondition` of the created entity to `false`, ignoring whatever was in the metadata.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7266)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
@@ -106,8 +106,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 
             if (requestedProperties.VisibilityCondition)
             {
-                string? visibilityCondition = property.GetMetadataValueOrNull("VisibilityCondition");
-                newUIProperty.VisibilityCondition = visibilityCondition ?? string.Empty;
+                if (!property.Visible)
+                {
+                    newUIProperty.VisibilityCondition = "false";
+                }
+                else
+                {
+                    string? visibilityCondition = property.GetMetadataValueOrNull("VisibilityCondition");
+                    newUIProperty.VisibilityCondition = visibilityCondition ?? string.Empty;
+                }
             }
 
             ((IEntityValueFromProvider)newUIProperty).ProviderState = new PropertyProviderState(cache, property.ContainingRule, propertiesContext, property.Name);
@@ -119,11 +126,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
         {
             foreach ((int index, BaseProperty property) in rule.Properties.WithIndices())
             {
-                if (property.Visible)
-                {
-                    IEntityValue propertyValue = CreateUIPropertyValue(queryExecutionContext, parent, cache, propertiesContext, property, index, properties);
-                    yield return propertyValue;
-                }
+                IEntityValue propertyValue = CreateUIPropertyValue(queryExecutionContext, parent, cache, propertiesContext, property, index, properties);
+                yield return propertyValue;
             }
         }
 
@@ -143,8 +147,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 
                 if (await project.GetProjectLevelPropertyPagesCatalogAsync() is IPropertyPagesCatalog projectCatalog
                     && projectCatalog.GetSchema(propertyPageName) is Rule rule
-                    && rule.TryGetPropertyAndIndex(propertyName, out BaseProperty? property, out int index)
-                    && property.Visible)
+                    && rule.TryGetPropertyAndIndex(propertyName, out BaseProperty? property, out int index))
                 {
                     IProjectState? projectState = null;
                     if (StringComparers.ItemTypes.Equals(propertiesContext.ItemType, "LaunchProfile"))
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducerTests.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducerTests.cs
@@ -95,7 +95,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
             {
                 new TestProperty { Name = "Alpha" },
                 new TestProperty { Name = "Beta" },
-                new TestProperty { Name = "Gamma" },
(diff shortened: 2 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
```diff
@@ -106,8 +106,15 @@ public static IEntityValue CreateUIPropertyValue(IQueryExecutionContext queryExe
 
             if (requestedProperties.VisibilityCondition)
             {
-                string? visibilityCondition = property.GetMetadataValueOrNull("VisibilityCondition");
-                newUIProperty.VisibilityCondition = visibilityCondition ?? string.Empty;
+                if (!property.Visible)
```
swesonga, line 109: Could you please clarify why the Visible property remains? What is it visible to? The front end?

tmeschter: The data model that the back end exposes to the client doesn't have `Visible`, only `VisibilityCondition`. The `Visible` property is part of the general-purpose properties system implemented by CPS, and the back-end sits on top of that.

The end result of all this is that the property is available to the client code, but it won't be shown to the user.

swesonga: I see, so we want the property to always be available to the client but with the correct VisibilityCondition. Thank you for the clarification

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyDataProducer.cs
```diff
@@ -106,8 +106,15 @@ public static IEntityValue CreateUIPropertyValue(IQueryExecutionContext queryExe
...
+                {
+                    newUIProperty.VisibilityCondition = "false";
```
drewnoakes, line 111: I don't think setting this visibility condition is necessary.

A design for this was proposed in https://github.com/dotnet/project-system/issues/7251.

Can we populate the `IUIProperty.IsVisible` property instead? That was added to the project model in https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/325905

Using `IsVisible` is a clearer way to express to the UI that the property must never be displayed. We can then completely remove these properties from the set of properties for which we evaluate visibility. In the UI, a property that is hidden due to its visibility condition still has several resources allocated for it. I prefer using `IsVisible` to indicate that, rather than interpreting this expression as a constant.

swesonga: > We can then completely remove these properties ...
@drewnoakes exactly which properties are you referring to? the ones on the back end?

tmeschter: Ah, I think I was getting that CPS change confused with the `IsReadonly` change.

So on the front-end we would first check `IsVisible`, and only if that is `true` would we evaluate the `VisibilityCondition` (if any)?

My concern is that we now have two properties in the project model (`IsVisible` and `VisibilityCondition`) with overlapping functionality. We can't avoid that on the back end because `BaseProperty.Visible` is already part of the public API, but on the front-end I could imagine using only `VisibilityCondition`:

- If it is not set the property is always visible.
- If it is set to a constant `true` it is always visible.
- If it is set to a constant `false` it is always invisible.

In these cases you would never need to re-evaluate.

That being said I don't feel strongly about it.

tmeschter: I tried to make this change but as yet there doesn't seem to be a version of the Microsoft.VisualStudio.ProjectSystem.Query package that includes the `IsVisible` property.

drewnoakes: > @drewnoakes exactly which properties are you referring to? the ones on the back end?

@sawesonga sorry, I'll be more explicit.

The UI will need all properties, both those which are always non-visible, and those which are visible (with optional conditionality on that visibility).

When a property is conditionally hidden, the UI still allocates UI elements for it. These elements are just not displayed.

For the case where a property is never going to be displayed, we can avoid creating those elements at all. That is what I meant by "completely remove".

> So on the front-end we would first check IsVisible, and only if that is true would we evaluate the VisibilityCondition (if any)?

Sort of. The filtering happens at different layers. The process would be:

1. Get all properties into the `PropertyContext`
2. Create a view of properties to bind to the UI's `ItemsControl`, filtered to those where `IsVisible` is `true`
3. Evaluate visibility conditions on only those properties bound to the UI from then on

> If it is set to a constant true it is always visible

That could work too, and I'm open to that, though it feels a bit hacky to me (my comment earlier saying "rather than interpreting this expression as a constant"). If we go that road we would have to back out the CPS change that introduced `IsVisible` to the project model.

> I tried to make this change but as yet there doesn't seem to be a version of the Microsoft.VisualStudio.ProjectSystem.Query package that includes the IsVisible property.

I've asked the CPS folks for guidance here on Teams. cc @adrianvmsft

drewnoakes: Adrian pointed to a new package on Teams. I tested it out and ended up filing #7271 to show the approach I envisaged.

tmeschter:
Closing in favor of #7271.

Issue #7267, opened 2021-05-25
System.OverflowException in old Property Pages (csproj.dll-based) in Dev17


MiYanni:
I'm fairly certain this is a difference in integer sizes or something of that sort. This System.OverflowException is thrown many times when trying to use the old Property Pages for csproj.dll-based projects.
![image](https://user-images.githubusercontent.com/17788297/119573700-7eb5ee00-bd69-11eb-86f4-2ed133815edd.png)

This is getting thrown from `PropPageDesignerWindowPane.vb`. Only noticed this happening after beginning to debug in Dev17.

PR #7270, opened 2021-05-27
Fixes System.OverflowException being thrown from old csproj.dll-based property pages

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7267

With the change to 64-bit, it seems this `WParam` value can be above Int32 now. Good news is that `Keys.KeyCode` is an enum set to `65535` in *System.Windows.Forms*, so that isn't changing... ever. Therefore, doing a *binary-and* against `65535` means the value will always be equal to or below `65535`, well within the range of Int32. Here, I just *binary-and* the Int64 with the enum, and do `CInt` so that the cast to `Keys` is aware the value produced an Int32. This no longer causes the System.OverflowException to occur.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7270)

### DIFF
--- a/src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerWindowPane.vb
+++ b/src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerWindowPane.vb
@@ -51,7 +51,7 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
 
             Dim DesignerView As PropPageDesignerView = GetPropPageDesignerView()
             If DesignerView IsNot Nothing Then
-                Dim KeyCode As Keys = DirectCast(m.WParam.ToInt32(), Keys) And Keys.KeyCode
+                Dim KeyCode As Keys = DirectCast(CInt(m.WParam.ToInt64() And Keys.KeyCode), Keys)
                 'Is the message intended for a window or control in the property page?
                 If DesignerView.IsNativeHostedPropertyPageActivated AndAlso NativeMethods.IsChild(View.Handle, m.HWnd) Then
                     Common.Switches.TracePDMessageRouting(TraceLevel.Info, "  ... Message is for a child of the property page.  Calling MyBase.PreProcessMessage", m)


MiYanni:
I just realized this change might make sense in `dev17Preview1`. I can retarget. @dmonroym, sound reasonable?

drewnoakes (approved):
Have you tested this in both 32-bit and 64-bit VS? Or asked another way, do we have a potential problem if this flows back into 16.x?

dmonroym:
> I just realized this change might make sense in `dev17Preview1`. I can retarget. @dmonroym, sound reasonable?

@MiYanni "Preview 1 is open for check-ins for Preview blocking bugs, deprecations, and QB approved items." 
Despite this being a potential huge improvement I don't think it fits in those categories but if you disagree feel free to re-target. 

It might also be worth doing a test validation run before merging

MiYanni:
@drewnoakes You can't run `main` in 16.x. It won't load the project system packages (I tried it). AFAIK, changes in `main` don't go backwards into 16.x. And AFAIK, 17.x is only 64-bit.

MiYanni:
@dmonroym I'm not certain on the details of this, but if this is throwing an exception, how could it not affect the usage of the property pages? I haven't done like an extensive test analysis of this. I basically required this change or else I can't work on my BinaryFormatter changes. By "test validation run", do you mean the validation insertion pipeline?

tmeschter:
We could consider taking this for Dev17 Preview 1, but first we need to know what impact this has on the user (e.g. does VS crash, are changes to the property pages not persisted correctly, is keyboard navigation broken, etc.).

MiYanni:
I think the simplest strategy here would be to merge this PR into `main` and if we get feedback on this causing problems for Preview 1, we can investigate further at that time and merge the changes into there if needed. Sound good?

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerWindowPane.vb
```diff
@@ -51,7 +51,7 @@ Namespace Microsoft.VisualStudio.Editors.PropPageDesigner
 
             Dim DesignerView As PropPageDesignerView = GetPropPageDesignerView()
             If DesignerView IsNot Nothing Then
-                Dim KeyCode As Keys = DirectCast(m.WParam.ToInt32(), Keys) And Keys.KeyCode
+                Dim KeyCode As Keys = DirectCast(CInt(m.WParam.ToInt64() And Keys.KeyCode), Keys)
```
swesonga, line 54: The enum has an underlying type of int so no need to do the AND using the 64-bit value. I'm not sure if the 32-bit value gets extended to 64-bits to do this.

swesonga (approved):

src/Microsoft.VisualStudio.AppDesigner/PropPageDesigner/PropPageDesignerWindowPane.vb
swesonga: I think my assumptions here were wrong. Ignore this comment.

swesonga: This is not that important but it was bugging me enough that I decided to drill into the Release code to see what happens in x64 in a simple Winforms app. If you bitwise AND after casting to an int, only a (32-bit) AND instruction is used:

and         edx,0FFFFh 

In the current pattern, we generate 2 additional instructions - a load and a sign extension to a 64-bit value - before doing a 64-bit AND.

mov         ecx,0FFFFh  
movsxd      rcx,ecx  
and         rcx,qword ptr [rbp-8]  

Not recommending a change here, simply documenting the behavior I observed from what didn't need to become a rabbit hole :)

swesonga: My vote is we fix in Preview 1 as well. Seems like a bad issue to leave in Preview 1.

drewnoakes:
This does seem like something we would want to fix in preview one. It blocks use of the legacy property pages, which are still in use for most projects in preview 1.

MiYanni:
@drewnoakes @swesonga 
I did a little testing. This doesn't block the use of legacy property pages in Dev17. However, there is a separate issues which I've documented here with the property pages: https://github.com/dotnet/project-system/issues/7297. Given that this particular change doesn't affect property page usage, I'm only merging it into Preview 2.

PR #7271, opened 2021-05-27
Populate IsVisible and return non-visible properties

drewnoakes:
Part of #7251. Alternative to #7266.

In order to show/hide properties based on project-level metadata such as `TargetPlatformIdentifier` or `LangVersion`, we need to plumb that data through to the UI for use in visibility condition expressions. Our approach for this is to have non-visible properties in the UI whose data is available to the visibility condition evaluator.

CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/325905 added an `IsVisible` property to the Project Query API's project model.

This PR populates that property when queried.

It also changes the query provider to return non-visible properties.

Future work will be to add non-visible properties (probably in a new well-known `ConfigurationGeneralPropertyPage.xaml` in which all properties are non-visible), and then convenient visibility expressions to access them.

---

This PR bumps the CPS package version, which brings in newly annotated APIs. The bulk of the diff here is reacting to those annotations. Each step is pulled out into its own commit, in case that makes review easier.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7271)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -72,10 +72,10 @@
     
 
     <!-- CPS -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem"                                   Version="17.0.89-pre-geec7dc0def" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="16.10.127-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="17.0.89-pre-geec7dc0def" />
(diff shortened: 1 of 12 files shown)


MiYanni (approved):
Looks good. But, what about Tom's PR?

tmeschter (approved):

tmeschter:
> Looks good. But, what about Tom's PR?

I'll probably just close it. :-)

swesonga (approved):

PR #7273, opened 2021-05-27
Enforce use of Async suffixes for async methods

drewnoakes:
This is a theme that comes up regularly during code review. It's more productive for everyone if we use an analyzer to enforce this.

This commit enables the analyzer and fixes or suppresses the diagnostics it produces. Future violations will be flagged as errors and fixed before reaching code review.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7273)


src/Microsoft.VisualStudio.ProjectSystem.Managed/GlobalSuppressions.cs
```diff
@@ -15,3 +15,12 @@
...
+[assembly: SuppressMessage("Style", "VSTHRD200:Use \"Async\" suffix for async methods", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.UnconfiguredProjectTasksService.OnProjectFactoryCompleted~System.Threading.Tasks.Task")]
+[assembly: SuppressMessage("Style", "VSTHRD200:Use \"Async\" suffix for async methods", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.VisualStudio.Threading.Tasks.SequentialTaskExecutor.ExecuteTask(System.Func{System.Threading.Tasks.Task})~System.Threading.Tasks.Task")]
+[assembly: SuppressMessage("Style", "VSTHRD200:Use \"Async\" suffix for async methods", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.AbstractMultiLifetimeComponent`1.CreateInitializedInstanceAsync~System.ValueTuple{`0,Microsoft.VisualStudio.Threading.JoinableTask}")]
```
swesonga, line 24: This one shouldn't be changed as per the unit test failure:

D:\...\src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\AbstractMultiLifetimeComponent.cs(125,64): error VSTHRD200: Avoid "Async" suffix in names of methods that do not return an awaitable type. [D:\workspace\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed\Microsoft.VisualStudio.ProjectSystem.Managed.csproj]

swesonga (approved):

tmeschter (approved):

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/GlobalSuppressions.cs
drewnoakes: This seems to be a bug. If I remove the suppression, the diagnostic appears in the IDE. If I suppress it again, it generates exactly this same suppression, yet then fails to build.

PR #7274, opened 2021-05-27
Add ConfigurationGeneralPage

drewnoakes:
Part of #7243

This "page" contains non-visible properties. It is present to support visibility conditions based on things like the target framework or language version.

As it stands, this PR will produce a strange looking page in the UI. A separate PR will be filed in CPS to filter not just non-visible properties, but also pages/categories containing only non-visible properties from the navigation tree.

A future PR in CPS will start to add visibility condition functions that will use the values provided via this new rule (by convention).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7274)


swesonga (approved):

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
```diff
@@ -23,6 +23,7 @@ folder "InstallDir:MSBuild\Microsoft\VisualStudio\Managed"
   file source="$(VisualStudioXamlRulesDir)Compile.FSharp.xaml"
   file source="$(VisualStudioXamlRulesDir)Compile.xaml"
   file source="$(VisualStudioXamlRulesDir)COMReference.xaml"
+  file source="$(VisualStudioXamlRulesDir)ConfigurationGeneralPage.xaml"
```
tmeschter, line 26: We already have a ConfigurationGeneral.xaml; we may want to give this a name that further differentiates it from that one.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ConfigurationGeneralPage.xaml
```diff
@@ -0,0 +1,27 @@
...
+    <DataSource Persistence="ProjectFile"
+                SourceOfDefaultValue="AfterContext"
+                HasConfigurationCondition="True" />
+  </Rule.DataSource>
+
+  <StringProperty Name="TargetFrameworkIdentifier" Visible="False" ReadOnly="True" />
+
+  <StringProperty Name="TargetFrameworkVersion" Visible="False" ReadOnly="True" />
+  
+  <StringProperty Name="LangVersion" Visible="False" ReadOnly="True" />
```
tmeschter, line 25: I would expect `TargetFrameworkIdentifier` and `TargetFrameworkVersion` to be present in more or less all .NET projects, but I'm not sure about `LangVersion`. We should at least make sure it is applicable to C#, VB, and F#. Also, if we're going to have an intrinsic visibility expression that uses this we should make sure that it works with the values returned by those languages.

tmeschter (approved):

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/CommonFiles.swr
drewnoakes: The similarity was intentional. Do you think it's confusing? In both CPS and the .NET Project System we have `ConfigurationGeneral` to represent general configured data about the project, not intended for display, which is similar to what this does (only this is a "page" for the UI).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ConfigurationGeneralPage.xaml
drewnoakes: `LangVersion` doesn't appear to be available to VB or F#. We will need an additional means to identify the project's language in visibility conditions, I suspect.

PR #7275, opened 2021-05-27
Fix 'sequence has no elements' exception

drewnoakes:
While testing this in Dev17, this code was throwing an exception as the predicate did not find a match.

This change makes the code tolerant of this situation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7275)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


swesonga (approved):

tmeschter:
I wonder when this can happen. Perhaps if we open the property pages right after the solution has opened, and the workspace hasn't populated yet?

tmeschter (approved):

MiYanni (approved):
Code looks good but I gotta wonder what changed to make this situation happen.

drewnoakes:
> I wonder when this can happen. Perhaps if we open the property pages right after the solution has opened, and the workspace hasn't populated yet?

Seems likely, though why now and why so repeatedly on my machine? Thinking about this again the next day, I'm not sure this is the best fix. Will think about this some more before.

drewnoakes:
@davkean do you have some insight here?

This is the dynamic enum provider for the startup object, and it seems we can hit it when no project is found in the workspace.

Ideally we would wait for things to catch up, but it's not clear to me how that would work.

This PR at least stops there being an exception. But it will result in the "startup object" drop-down being empty.

davkean:
This code needs to wait for the language service to be initialized. It also needs to use the Roslyn project specific to the configuration that that it belongs to, currently its just picking the first. Can IDynamicEnumValuesProvider's be specific to configurations?

drewnoakes:
> This code needs to wait for the language service to be initialized

@davkean any tips on how to do this? Would `IActiveWorkspaceProjectContextHost.PublishAsync` achieve this?

> Can IDynamicEnumValuesProvider's be specific to configurations?

Yes. `LaunchTargetPropertyPageEnumProvider` is. However we don't support differing startup objects by configuration today. We can, of course, change that.

drewnoakes:
Never mind. I worked it out.

Issue #7279, opened 2021-05-28
Enhance Windows Forms project properties page


RussKie:
**Visual Studio Version**:
Dev17

**Summary**:

We have a design proposal https://github.com/dotnet/winforms/issues/4909 that seeks to streamline bootstrap of Windows Forms C# apps, and get developers to configure certain aspects of their app via MSBuild properties. E.g. instead of writing this code
```diff
    static class Program
    {
        [STAThread]
        static void Main()
        {
-           Application.EnableVisualStyles();
-           Application.SetDefaultFont(new Font(....));
-           Application.SetHighDpiMode(HighDpiMode.PerMonitorV2);
-           Application.SetCompatibleTextRenderingDefault(false);

            Application.Run(new MainForm());
       }
    }
```
developers will configure these as such:
```diff
// csproj / Directory.Build.props / etc.

  <PropertyGroup>
     <ApplicationIcon />
     <ApplicationManifest>app1.manifest</ApplicationManifest>
+    <ApplicationVisualStyles>[true|false, default=true]</ApplicationVisualStyles>
+    <ApplicationUseCompatibleTextRendering>[true|false, default=false]</ApplicationUseCompatibleTextRendering>
+    <ApplicationFontName>[Font family name, string, default='', i.e. Control.DefaultFont]</ApplicationFontName>
+    <ApplicationFontSize>[Font size, float, default=9]</ApplicationFontSize>
+    <ApplicationHighDpiMode>[dpi mode, string/HighDpiMode enum value, default=PerMonitorV2]</ApplicationHighDpiMode>
  </PropertyGroup>
```

It would be great to extend this experience to the Windows Forms apps property pages, which could looks something like (mock up):
![image](https://user-images.githubusercontent.com/4403806/119922957-2e7e9c00-bfb4-11eb-8086-c002c2315c58.png)

**Nitty gritty details**:

To facilitate the experience we'll need the following controls:
- checkbox for `ApplicationVisualStyles` (default=checked) and `ApplicationUseCompatibleTextRendering` (default=unchecked)
- a dropdown list with the list of enum values for `ApplicationHighDpiMode`
- a font picker for `ApplicationFontName` and `ApplicationFontSize`.
Nothing fancy, just to pick any Windows font: 
![image](https://user-images.githubusercontent.com/4403806/119923808-c16c0600-bfb5-11eb-8649-a08c8188b5ce.png)

The visibility of several controls is subject to the `TargetFramework`, and is as follows:
[...]

PR #7281, opened 2021-05-28
Add UseWindowsForms and UseWPF properties

### LINKED ISSUE #7239: Expose UseWindowsForms and UseWPF properties
I believe these properties are only meaningful on .NET Core 3.0/3.1, but it would still be useful to have them as they control whether or not WinForms and WPF functionality is available to the project.

The Application property page seems like a natural place for these. However, we only want them to show for projects that target .NET Core 3.0/3.1.

_This work was identified as part of #6935._

drewnoakes:
Fixes #7239. Contributes to #7279.

The `UseWindowsForms` and `UseWPF` properties are only applicable when targeting .NET Core 3.0 or later.

Until recently, we did not have a means to express a visibility condition based upon target framework.

Now that we have one, we can add these properties to the UI and have them behave correctly.

cc @RussKie

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7281)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -104,6 +104,42 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <!-- TODO use fwlink for HelpUrl -->
+  <BoolProperty Name="UseWindowsForms"
+                DisplayName="Windows Forms"
+                Description="Enable Windows Forms for this project."
+                HelpUrl="https://docs.microsoft.com/dotnet/core/project-sdk/msbuild-props-desktop"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::OutputType" />
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>
+          (and
+            (has-evaluated-value "Application" "OutputType" "WinExe")
+            (has-net-core-app-version-or-greater "3.0"))
+        </NameValuePair.Value>
+      </NameValuePair>
+    </BoolProperty.Metadata>
+  </BoolProperty>
+
+  <!-- TODO use fwlink for HelpUrl -->
+  <BoolProperty Name="UseWPF"
+                DisplayName="Windows Presentation Foundation"
+                Description="Enable WPF for this project."
+                HelpUrl="https://docs.microsoft.com/dotnet/core/project-sdk/msbuild-props-desktop"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::OutputType" />
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>
+          (and
+            (has-evaluated-value "Application" "OutputType" "WinExe")
+            (has-net-core-app-version-or-greater "3.0"))
+        </NameValuePair.Value>
+      </NameValuePair>
+    </BoolProperty.Metadata>
+  </BoolProperty>
+
   <DynamicEnumProperty Name="StartupObject"
                        DisplayName="Startup object"
                        Description="Defines the entry point to be called when the application loads. Generally this is set either to the main form in your application or to the 'Main' procedure that should run when the application starts. Class libraries do not define an entry point."
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ApplicationPropertyPage.xaml.cs.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ApplicationPropertyPage.xaml.cs.xlf
@@ -22,6 +22,26 @@
         <target state="translated">Cílit na více architektur</target>
         <note />
       </trans-unit>
+      <trans-unit id="BoolProperty|UseWPF|Description">
+        <source>Enable WPF for this project.</source>
+        <target state="new">Enable WPF for this project.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="BoolProperty|UseWPF|DisplayName">
+        <source>Windows Presentation Foundation</source>
+        <target state="new">Windows Presentation Foundation</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="BoolProperty|UseWindowsForms|Description">
+        <source>Enable Windows Forms for this project.</source>
+        <target state="new">Enable Windows Forms for this project.</target>
+        <note />
+      </trans-unit>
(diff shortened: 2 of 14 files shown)


RussKie:
Any screenshots to share?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -104,6 +104,42 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <!-- TODO use fwlink for HelpUrl -->
+  <BoolProperty Name="UseWindowsForms"
+                DisplayName="Windows Forms"
+                Description="Enable Windows Forms for this project."
+                HelpUrl="https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop"
```
RussKie, line 111: https://review.docs.microsoft.com/help/contribute/links-how-to?branch=master#locale-codes-in-external-links
```suggestion
                HelpUrl="https://docs.microsoft.com/dotnet/core/project-sdk/msbuild-props-desktop"
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -104,6 +104,42 @@
...
+        </NameValuePair.Value>
+      </NameValuePair>
+    </BoolProperty.Metadata>
+  </BoolProperty>
+
+  <!-- TODO use fwlink for HelpUrl -->
+  <BoolProperty Name="UseWPF"
+                DisplayName="Windows Presentation Foundation"
+                Description="Enable WPF for this project."
+                HelpUrl="https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop"
```
RussKie, line 129: ditto

drewnoakes:
> Any screenshots to share?

I should have captured some, but I've unravelled all the threads here in order to submit PRs. This change requires bringing together about 5 branches across two repos and three VSIX packages. It would take me a while to get them together.

But basically it's just:

```
Windows Forms
[ ] Enable Windows Forms for this project.

Windows Presentation Foundation
[ ] Enable WPF for this project.
```

And these items are only present if the `OutputType` is `WinExe` and the project targets .NET Core 3.0 or later. If multi-targeting, then only one target has to satisfy the constraint.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -104,6 +104,42 @@
...
+                DisplayName="Windows Forms"
+                Description="Enable Windows Forms for this project."
+                HelpUrl="https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::OutputType" />
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>
+          (and
+            (has-evaluated-value "Application" "OutputType" "WinExe")
```
RussKie, line 118: You probably want to force `WinExe` over `Exe` if either of these frameworks are selected: https://review.docs.microsoft.com/help/contribute/links-how-to?branch=master#locale-codes-in-external-links

drewnoakes: Is that the right link? I'm not sure what you mean by forcing here.

RussKie: Of course not 🤦‍♂️ Here's the right link: https://docs.microsoft.com/dotnet/core/compatibility/sdk/5.0/automatically-infer-winexe-output-type#change-description

image
This what prompted the comment about forcing - e.g. is a user checked `UseWindowsForms` or `UseWPF` we shouldn't allow switching to `Exe`. May be even remove it from the list?
Likewise may be show a checkbox for `DisableWinExeOutputInference`?

drewnoakes: Interesting. Offhand I'm not sure we can dynamically refresh the contents of the output type list.

What we _could_ do is have the `UseWindowsForms` and `UseWPF` properties be deleted from the project if the user selects an output type other than `WinExe`.

RussKie: Strictly speaking Exe is an accepted value for a Windows Forms app, it is the same as WinExe as of .NET 5.0.
Also one can build a DLL targeting Windows Forms, e.g. GitUI or GitCommands are both libraries targeting Windows Forms.

drewnoakes:
> Any screenshots to share?

image

And not present for .NET Framework:

image

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: This is awesome, thanks. Fixed in this PR, and I also filed https://github.com/dotnet/project-system/pull/7283 to fix it in a few other places. Appreciated.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
tmeschter: The Property Pages UI will reflect the final evaluated value of the OutputType property. So for WinForms or WPF apps if you try to set it to Exe in the UI the project file will get updated--and then the UI will reset itself because the value is overridden in the targets.

This means the `DependsOn` is backwards--the OutputType effectively depends on `UseWindowsForms` and `UseWPF` even though for the purposes of showing these controls we would like it to be the other way around.

I agree with @drewnoakes; let's get this in and see what the feedback is. It's better to have these checkboxes in some form now rather than wait until we've figured out the perfect implementation.

tmeschter (approved):

drewnoakes:
I'll merge this once the CPS bits are reviewed and merged.

> This means the `DependsOn` is backwards

I actually think we over-specify `DependsOn` in several places. It could be revisited.

MiYanni (approved):

PR #7282, opened 2021-05-28
Document new visibility condition functions

drewnoakes:
https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/327513 adds several new functions. They are documented here.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7282)

### DIFF
--- a/docs/repo/property-pages/visibility-conditions.md
+++ b/docs/repo/property-pages/visibility-conditions.md
@@ -93,10 +93,17 @@ The following table details the default set of visibility expression functions:
 | `has-evaluated-value`    | 3        | Returns true if property on page `arg0` with name `arg1` has an evaluated value matching `arg2` |
 | `is-codespaces-client`   | 0        | Returns true if the Project Properties UI is running in a Codespaces client                     |
 | `has-project-capability` | 1        | Returns true if the project has the specified capability.                                       |
+| `has-net-framework`                    | 0 | Returns true if the project targets .NET Framework in at least one configuration. |
+| `has-net-core-app`                     | 0 | Returns true if the project targets .NET Core or .NET 5+ in at least one configuration. |
+| `has-net-framework-version-or-greater` | 1 | Returns true if the project targets .NET Framework at the specified version or above in at least one configuration. |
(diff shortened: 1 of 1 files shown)


tmeschter (approved):
I'm approving the PR because this functionality is incredibly useful.

On the other hand, I feel very iffy about tying the visibility functions so strongly to .NET and even C# projects, especially that the UI now has "baked-in" knowledge of certain specific properties.

What would it look like to solve the same issues in a more generic way? I see the following requirements:
* Version comparison functions.
* A function that exposes the set of evaluated values.
* An `any` function that can apply another function to a set of values.

So `has-csharp-lang-version-or-greater` could be expressed as:

```
(any
  (evaluated-values "ConfigurationGeneralPage" "LangVersion")
  "version-greater-than" "8.0")

This is better in some ways and worse in others. :-) Anyway, just something to keep in mind.

**Comment by drewnoakes** [comment]
I understand your reservations here. It's similar to my concern a while back about binding to MSBuild concepts such as evaluated/unevaluated values.

One idea I'd considered was allowing rules to define their own macros for these things, to allow code re-use. We could always add this in future without breaking existing pages.

Note that if a usage of this property infrastructure doesn't need these .NET concepts, they would not need to populate `ConfigurationGeneralPage` with the properties expected by these functions. So long as they don't use these functions, there is no run-time error.

> So `has-csharp-lang-version-or-greater` could be expressed as:
>
>```
>(any
>  (evaluated-values "ConfigurationGeneralPage" "LangVersion")
>  "version-greater-than" "8.0")
>```

Actually that function also checks the language, so the equivalent would be more like:

```
(and
  (has-evaluated-value "ConfigurationGeneralPage" "Language" "C#")
  (any
    (evaluated-values "ConfigurationGeneralPage" "LangVersion")
    "version-greater-than" "8.0"))
```

Comparing this with `(has-csharp-lang-version-or-greater "8.0")`, I don't think (for me at least) that concerns here outweigh the communication and maintenance benefits of having such a succinct form. The definition of this function could always be moved so that it's only available in a .NET context in future.

MiYanni (approved):

drewnoakes:
These have now merged into CPS.

PR #7285, opened 2021-05-28
Restrict 'Nullable' property to C# 8+

drewnoakes:
Contributes to #6989.

Nullable reference types were introduced in C# 8.

This change uses a newly introduced visibility condition function to hide the `Nullable` property from the UI when `LangVersion` is less than 8.0.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7285)


tmeschter:
What if `LangVersion` is set to `default`, `latestmajor`, `preview`, or `latest`?

MiYanni (approved):

drewnoakes:
> What if `LangVersion` is set to `default`, `latestmajor`, `preview`, or `latest`?

Discussed in the CPS PR.

PR #7298, opened 2021-06-03
Add some basic information about launch profiles

tmeschter:
A brain dump of information about launch profiles and how they are used in VS.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7298)

### DIFF
--- a/docs/launch-profiles.md
+++ b/docs/launch-profiles.md
@@ -1,0 +1,65 @@
+﻿# Launch Profiles
+
+The _Launch Profiles_ feature of the .NET Project System allows developers to specify multiple ways of running their project, each with a different set of options, and quickly switch between them in Visual Studio. For example, when developing a web application you may have one profile that runs your project under the Kestrel web server, and another that launches it under IIS Express. Or, when creating a console application, different profiles may specify different command line arguments.
+
+## Example
+
+The launch profiles for a particular project (if any) are stored in the Properties\launchSettings.json file. Below is an example based on a launchSettings.json found in this repo. Note that within Visual Studio it is more common to define and update launch profiles through the UI (via the project properties in VS2019 and earlier, or the Launch Profiles UI in VS2022 and later) but the file can be edited directly.
+
+``` json
+{
+  "profiles": {
+    "Start (with native debugging)": {
+      "commandName": "Executable",
+      "executablePath": "$(DevEnvDir)devenv.exe",
+      "commandLineArgs": "/rootsuffix $(VSSDKTargetPlatformRegRootSuffix) /log",
+      "environmentVariables": {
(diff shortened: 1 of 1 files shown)


jjmew (approved):

docs/launch-profiles.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+This file contains a single launch profile named "Start (with native debugging)" containing various settings:
+
+- `commandName`: This is the most important setting, and the only required one. We'll come back to it in a moment.
+- `executablePath`: The path to an executable (.exe) to run when this launch profile is selected, and the project is "started". In this starting the project will run devenv.exe.
+- `commandLineArgs`: The arguments to supply to the specified executable.
+- `environmentVariables`: A collection of name/value pairs, each specifying an environment variable and value to set. A setting in a profile can contain any arbitrary data so long as it can be represented in JSON (and even that isn't always a requirement; see "In-memory settings and profiles", below).
+- `nativeDebugging`: Specifies that the native code debugging engine should be used in addition to the managed code debugging engine.
+
```
drewnoakes, line 36: To avoid someone who is skim reading, or who stopped reading here, taking this as a reference list of settings:

```suggestion

This is not an exhaustive list of settings, as we will discuss below.

```

gundermanc (approved):

docs/launch-profiles.md
```diff
@@ -0,0 +1,65 @@
...
+This functionality is available to other launch commands through the [`IDebugTokenReplacer`](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDebugTokenReplacer.cs) interface.
+
+### Command-specific UI
+
+Each launch command can have its own associated UI for updating properties; see the [Project Property Pages documentation](https://github.com/dotnet/project-system/tree/main/docs/repo/property-pages) for more information.
+
+### In-memory profiles and settings
+
+VS supports in-memory profiles that are not committed to the launchSettings.json. For example, if no launchSettings.json is present the .NET Project System will automatically create an in-memory profile with the project's name and the "Project" launch command; this allows running a project without cluttering your project with a trivial launchSettings.json file. Only when the profile is edited through the UI do we create the file and commit the profile.
```
drewnoakes, line 63: Can we cross reference the code that does this here?

drewnoakes (approved):

docs/launch-profiles.md
tmeschter: Oh, yes. Good catch. I went back and added those links for other things, but missed these.

PR #7299, opened 2021-06-03
Use CPS's IProjectDiagnosticOutputService for output logging

### LINKED ISSUE #7166: Use IProjectDiagnosticOutputService from CPS
In https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/321174, CPS is adding a service that duplicates the code we have in `IProjectOutputWindowPaneProvider`. Once it merges, we should remove our implementation and associated testing code, then delegate to CPS's service.

drewnoakes:
Fixes #7166

The .NET Project System does some diagnostic logging to the Output window in VS when running in a diagnostic mode.

Previously we had code to wire this up using `IVsOutputWindow` and other VS services.

Now, CPS provides `IProjectDiagnosticOutputService` for this purpose, so we can use it directly and delete a bunch of code from this repo.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7299)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Logging/ProjectDiagnosticOutputServiceExtensions.cs
```diff
@@ -91,15 +51,18 @@ public static void WriteLine(this IProjectLogger logger, string format, object?
         /// <exception cref="FormatException">
         ///     The format specification in <paramref name="format"/> is invalid.
         /// </exception>
-        public static void WriteLine(this IProjectLogger logger, string format, object? argument1, object? argument2)
+        public static void WriteLine(this IProjectDiagnosticOutputService logger, string format, object? argument1, object? argument2)
         {
             Requires.NotNull(logger, nameof(logger));
 
-            logger.WriteLine(new StringFormat(format, argument1, argument2));
+            if (logger.IsEnabled)
```
swesonga, line 58: >if (logger.IsEnabled) [](http://example.com/codeflow?start=12&length=21)

Should the IProjectDiagnosticOutputService be updated to check whether it is enabled before WriteLining? Seems odd to have to do this in the consumers.

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Logging/ProjectDiagnosticOutputServiceExtensions.cs
drewnoakes: That service only accepts a `string`. Here we have format arguments and want to avoid allocating a string if the logger is disabled.

The format arguments here are `object` so we may still have some boxing. In places where we do larger amounts of logging, the caller of these extension methods may also check `IsEnabled`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/RestoreLogger.cs
```diff
@@ -2,18 +2,17 @@
...
 {
     internal static class RestoreLogger
     {
-        public static void BeginNominateRestore(IProjectLogger logger, string fullPath, IVsProjectRestoreInfo2 projectRestoreInfo)
+        public static void BeginNominateRestore(IProjectDiagnosticOutputService logger, string fullPath, IVsProjectRestoreInfo2 projectRestoreInfo)
         {
             if (logger.IsEnabled)
             {
-                using IProjectLoggerBatch batch = logger.BeginBatch();
+                using var batch = new BatchLogger(logger);
```
tmeschter, line 15: Style question: Should we prefer `using BatchLogger batch = new(logger);`?

In general we try to avoid using `var`.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Logging/ProjectDiagnosticOutputServiceExtensions.cs
swesonga: Wouldn't having a WriteFormattedString in the logger and have the logger check whether it is enabled address these concerns?

drewnoakes: Potentially yes. In this case that'd be a change to the CPS API.

Though for performance it's simpler to have the client decide how it wants to handle it. For example, what API would that method have? Probably `WriteFormattedString(string format, params string[] args)` with a `params` array for the format args. That means that the array of arguments is allocated (and any struct args boxed) on every call, even if the logging is disabled.

Ideally we want "you don't pay for what you don't use" which is hard to achieve for this kind of logging with today's core types. I'd hoped that `FormattableString` might be able to help here somehow, but that is still a class and has its own limitations. In the .NET Project System we have [`StringFormat`](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/StringFormat.cs) as an attempt to avoid allocation of params arrays, but it's not a perfect pattern either, and we keep it `internal`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/Snapshots/RestoreLogger.cs
drewnoakes: I'm happy either way. My feeling is that this kind of thing should be enforced by analyzers where possible to maintain consistency. We do have a lot of use of `var` in the code today, especially where the type is evident.

PR #7300, opened 2021-06-04
Localized file check-in by OneLocBuild Task: Build definition ID 981: Build ID 1170541

MiYanni:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7300)


drewnoakes:
@MiYanni do we understand why the tooling that generates these PRs makes it look as though the whole file has changed? It's impossible to review these PRs as they are today.

Historically we've caught issues by reviewing diffs in localisation PRs. I'm concerned that we'll miss those with this new system.

MiYanni:
@drewnoakes For now, disable whitespace changes when viewing the PR. I'm talking to Cristiano about the whitespace changes.
image

MiYanni:
After merging other Loc PRs, this PR only contains whitespace changes now. Closing.

MiYanni:
@drewnoakes I never responded to this from before. I found out recently that Arcade has a custom layer on top of the PR system from the localization team. So, I'm not sure if the feature to update PRs is part of the localization team's process or part of the Arcade process. Honestly, I think we should simply transition to using Arcade so that we have less we need to manage and maintain. Not something I'm looking into right this second, but I think it should be on the plate for discussion among the team at some point.

PR #7316, opened 2021-06-11
Replace BinaryFormatter usage

MiYanni:
After many attempts and investigation, this solution to replace BinaryFormatter seems the most simple and covers 99% of prior scenarios. After reviewing my previous solution, @drewnoakes pointed this solution out to me. Original solutions I made used System.Text.Json as the main serialization driver, as it is the library I've used for the last 3 years or so. This solution doesn't use that, and uses a mix of BinaryReader/Writer and DataContractSerializer.

How is works is it encodes the AssemblyQualifiedName of the value's type as a string into the stream. When using BinaryReader/Writer, it encodes strings including the length of the string. That is important because of the second step. The second step is to use DataContractSerializer to serialize the actual value we want to keep. This is encoded as XML. I don't have a very strong understanding of the mechanics of this, since it must include some type information in the XML. I'm assuming it makes some schema of XML that is similar to how BinaryFormatter worked, but safer? Again, since type information is stored in these serialized values, this allows classes that define Object properties to deserialize (since Object can hold any type). My original solution(s) had to manually encapsulate these Object properties to store type information, since most serializers I'm used to do not use their own schema for serialization that includes type information.

On deserialization, the AssemblyQualifiedName string is read (using the known string length). This is then turned into a Type. What remains in the stream is the XML node of the class's value. This is deserialized using DataContractSerializer. This mechanism works because we only leave 1 XML root node in the stream. Trying to use DataContractSerializer for both the type and value information creates 2 root nodes, which won't work. The other option was to make a type that simply holds both a Type property and a Value property. My original solution did something to this, which I had called SerializationContainer. Here, we are using the BinaryReader/Writer mechanics to encode 2 different serialization types into the same stream, and read them each correctly. As long as the string is read first and then the XML, there aren't any issues.

**Note:** I will 'Squash and Merge' this PR so that it doesn't have my commit history of the multiple solutions I tried. If you're in for looking at my madness, go look through the commit history. Otherwise, it can be ignored.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7316)


src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -30,6 +30,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
     ''' </remarks>
     <Serializable>
     <TypeDescriptionProvider(GetType(ResourceTypeDescriptionProvider))>
+    <KnownType(GetType(ResXDataNode))>
```
MiYanni, line 33: This is the only thing I needed to add to any of the types being serialized. It seems that DataContractSerializer is slightly (very slightly) more strict and needs to know ahead of time that it will d/encode this particular type. All other types used did not need any additional information. This is why I said that DataContractSerialize will cover **99%** of the types used without any changes.

drewnoakes: Interesting. Have you checked other common types, like `Bitmap` and `Icon`?

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
```diff
@@ -1665,5 +1666,29 @@ Namespace Microsoft.VisualStudio.Editors.Common
 
         End Class
 #End Region
+
+        Public Class SerializationProvider
```
drewnoakes, line 1670: Rather than duplicate this source, your options are either:

- Add an `InternalsVisibleTo` (seems overkill just for this).
- Make it a top-level internal type in its own source file, and link the file into both projects.

I'd recommend the second approach.

drewnoakes, line 1670: Do you think unit tests would add value here? For example, to validate that the types we know about round trip successfully.

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
MiYanni: I used Bitmap as a primary testing scenario. In older implementations, I tried Icon, audio files, and other files. I need to test them again with this implementation. But, my assumption has been if Bitmap works, the other types will too.

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
MiYanni: There are multiple functions/subroutines in each Utils files that are duplicates. I even created one when doing the BinaryFormatter telemetry. I've been *really* wanting to clean up this code. If I end up making a shared source file, I'd clean up this whole file (Utils) to work in either project; unless there are types specific to a particular project. I think it would make sense to do that in a separate PR, though. I also need to make another PR to remove the BinaryFormatter telemetry since it won't be as necessary after this PR.

MiYanni: Yeah. I want to do that. But I also wanted to get this PR available to review. So, I'll be adding tests shortly.

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
drewnoakes: Please test those other common types before merging this.

drewnoakes (approved):
Consider squash merging this (or cleaning up history and force-pushing).

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
swesonga, line 1670: >SerializationProvider [](http://example.com/codeflow?start=21&length=21)

How about a name like ObjectSerializer? I think the Provider suffix here is confusing because this class has only static methods and doesn't appear to "provide" any object as the provider pattern does, assuming Provider is even a pattern [https://blog.ploeh.dk/2011/04/27/Providerisnotapattern/](https://blog.ploeh.dk/2011/04/27/Providerisnotapattern/)

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -705,5 +707,28 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
         End Class
 #End Region
 
+        Public Class SerializationProvider
+
+            Public Shared Sub Serialize(stream As Stream, value As Object)
+                If value Is Nothing Then
+                    Return
+                End If
+                Using writer As New BinaryWriter(stream, Encoding.UTF8, True)
```
drewnoakes, line 716: Using a named argument makes this clearer.

```suggestion
                Using writer As New BinaryWriter(stream, Encoding.UTF8, leaveOpen:=True)
```

Ditto below.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -705,5 +707,28 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
...
+                    Dim valueType = value.GetType()
+                    writer.Write(valueType.AssemblyQualifiedName)
+                    writer.Flush()
+                    Call New DataContractSerializer(valueType).WriteObject(stream, value)
+                End Using
```
drewnoakes, line 721: I would invert these lines personally. I think it's clearer:

```suggestion
                End Using
                Call New DataContractSerializer(valueType).WriteObject(stream, value)
```

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -705,5 +707,28 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
         End Class
 #End Region
 
+        Public Class SerializationProvider
+
+            Public Shared Sub Serialize(stream As Stream, value As Object)
+                If value Is Nothing Then
+                    Return
```
drewnoakes, line 714: In this case no data is written to the stream. Does a deserialize call on the same data return `null` correctly?

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
MiYanni: Unit tests have been added. I did a few basic types, a couple complex types, and null/empty stream testing.

src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
MiYanni: I added tests with some types. However, the top-level 'common' types (`Resource`, `MultipleValuesStore`) for this serialization implementation cannot be done as they require the context of Visual Studio to create. They have quite deep hierarchies of objects. I think mocking them would take longer than it is worth, considering this is legacy code that has existed this long without any testing.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -709,22 +709,28 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
 
         Public Class SerializationProvider
 
+            Private Shared ReadOnly KnownTypes As Type() = {GetType(Size)}
+
             Public Shared Sub Serialize(stream As Stream, value As Object)
-                If value Is Nothing Then
+                If stream Is Nothing Or value Is Nothing Then
```
drewnoakes, line 715: I don't think it makes sense to accept a null stream. I would use `Requires.NotNull` for that parameter.

drewnoakes: Instead, for null value, consider writing `"<null>"` (or some such string) to the stream in place of the type name. Then, when deserialising, you receive that string then you know to return null.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
```diff
@@ -709,22 +709,28 @@ Namespace Microsoft.VisualStudio.Editors.AppDesCommon
...
+                    Call New DataContractSerializer(valueType, KnownTypes).WriteObject(stream, value)
                 End Using
             End Sub
 
             Public Shared Function Deserialize(stream As Stream) As Object
+                If stream Is Nothing OrElse stream.Length = 0 Then
+                    Return Nothing
+                End If
                 Using reader As New BinaryReader(stream, Encoding.UTF8, True)
+                    reader.BaseStream.Position = 0
```
drewnoakes, line 731: A deserialize function should not assume it is reading from the start of the stream. If you need to reset the position in test code, do that in the test itself.

tests/Microsoft.VisualStudio.Editors.UnitTests/Microsoft.VisualStudio.Editors.UnitTests.vbproj
```diff
@@ -0,0 +1,40 @@
...
+    <Compile Remove="My Project\**" />
+    <EmbeddedResource Remove="My Project\**" />
+    <None Remove="My Project\**" />
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Remove="Resource1.resx" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="explorer.ico" />
+    <None Remove="img0.jpg" />
```
swesonga, line 17: remove unnecessary file reference

tests/Microsoft.VisualStudio.Editors.UnitTests/Microsoft.VisualStudio.Editors.UnitTests.vbproj
```diff
@@ -0,0 +1,40 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\DeployTestDependencies\DeployTestDependencies.csproj" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\src\Microsoft.VisualStudio.Editors\Microsoft.VisualStudio.Editors.vbproj" />
+    <ProjectReference Include="..\..\src\Microsoft.VisualStudio.AppDesigner\Microsoft.VisualStudio.AppDesigner.vbproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Update="Resource1.Designer.vb">
```
swesonga, line 34: Is this compile item necessary? Look like there are various files mentioned in here that aren't being added. Consider cleaning them up.

swesonga:
Why are you concerned about including your initial implementations in the commit history if you don't squash merge? I would have thought it would be helpful to see that evolution in the commit history.

swesonga:
>             TelemetryLogger.LogBinaryFormatterEvent(NameOf(ResourceEditorView), TelemetryLogger.BinaryFormatterOperation.Deserialize)

Shouldn't all this binary formatter telemetry be removed as well, or at least renamed to not be misleading since BinaryFormatter will no longer be in use?

---
Refers to: src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceEditorView.vb:2993 in 48a3c9e. [](commit_id = 48a3c9e94319336c4fee2b9d5c4bd31e1c9f152b, deletion_comment = False)

swesonga (approved):

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
MiYanni: At the time, I forgot our made-up 'provider' naming thing meant what the industry normally calls a factory. This is not a factory. However, it does *provide* serialization. But, I agree, `ObjectSerializer` would work just fine here.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Is this personal preference or a standard we do as a team? Over the last few years, I got used to [this feature](https://www.jetbrains.com/help/resharper/Inline_Parameter_Name_Hints.html#parameter-name-hints) of Resharper. Plus, usual code styling tools will make the `leaveOpen:=` grey out and say, "Remove unnecessary named parameter". It doesn't seem we have that styling, though. Basically, I've used named parameters when they are needed (to compile), not as code comments.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: The reason this is done is because deserialization initializes `valueType` in the using statement. Otherwise, you would need a sentinel value (such as Nothing) and then assign it in the using. Both serialization and deserialization were written in a way to appear similar to each other. So, I opted for that consistency over making the usings completely pure. Either is straightforward enough to read. It is 3 or 4 lines of code. Pulling `valueType` out of the using means it is declared earlier than it is used, which is my preference to not do.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: I've added a unit test for this and updated these guard clauses to be more encompassing.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: I've added these either way. But, my initial intuition would always lead me to believe they are necessary, without looking at all the method overloads; if I was seeing this in a code review or in the code itself.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Ah. I looked into it. Since this is a 'drop-in replacement' of BinaryFormatter, I'm going to do the same thing that [Serialize](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter.serialize?view=net-5.0) and [Deserialize](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter.deserialize?view=net-5.0) do in these situations. I can use `Requires.NotNull` for most of them. Thanks.

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
MiYanni: I've renamed it.

src/Microsoft.VisualStudio.AppDesigner/Common/Utils.vb
MiYanni: Removed this line.

MiYanni:
@swesonga Removing the telemetry would be in a separate PR. I've added an issue here: https://github.com/dotnet/project-system/issues/7349

tests/Microsoft.VisualStudio.Editors.UnitTests/Microsoft.VisualStudio.Editors.UnitTests.vbproj
MiYanni: This is the handiwork of Visual Studio not cleaning up the file properly. I've cleaned it myself now.

tests/Microsoft.VisualStudio.Editors.UnitTests/Microsoft.VisualStudio.Editors.UnitTests.vbproj
MiYanni: Removed these statements.

src/Microsoft.VisualStudio.Editors/Common/Utils.vb
MiYanni: Created an issue for this here: https://github.com/dotnet/project-system/issues/7351

jjmew (approved):
:shipit:

PR #7320, opened 2021-06-15
Introduce AutoGenerateBindingRedirects property.

### LINKED ISSUE #5940: New "Application" page needs "Auto-generate binding redirects" property (1)
#5920 added the back-end components for an "Application" project property page for VS Online scenarios. However, it does not currently support the "Auto-generate binding redirects" property for a couple of reasons:

1. The property is only meaningful when targeting the .NET Framework, so it isn't needed for the .NET Core/Standard scenarios we are currently targeting.
2. We would like to hide the control entirely by marking the property invisible unless we're targeting .NET Framework, but we don't currently have a way to dynamically adjust the visibility. We're going to need that for #5937, though.

Once we can dynamically update the visibility we should revisit this and add the property.

melytc:
Fixes #5940.

On .NET Framework projects, a user will now be able to enable or disable automatic binding redirection by setting the `AutoGenerateBindingRedirects` property in the Project Propeties UI. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7320)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -190,6 +190,20 @@
     </BoolProperty.Metadata>
   </BoolProperty>
 
+  <BoolProperty Name="AutoGenerateBindingRedirects"
+                DisplayName="Auto-generate binding redirects"
+                Description="Enables binding redirection to be added automatically to the app configuration file to override assembly unification."
+                HelpUrl="https://go.microsoft.com/fwlink/?linkid=2164379"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>
+        (has-net-framework)
+        </NameValuePair.Value>
+      </NameValuePair>
+    </BoolProperty.Metadata>
+  </BoolProperty>
+
   <DynamicEnumProperty Name="StartupObject"
                        DisplayName="Startup object"
                        Description="Defines the entry point to be called when the application loads. Generally this is set either to the main form in your application or to the 'Main' procedure that should run when the application starts. Class libraries do not define an entry point."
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -190,6 +190,20 @@
     </BoolProperty.Metadata>
   </BoolProperty>
 
+  <BoolProperty Name="AutoGenerateBindingRedirects"
+                DisplayName="Auto-generate binding redirects"
+                Description="Enables binding redirection to be added automatically to the app configuration file to override assembly unification."
```
swesonga, line 195: not sure what the best phrasing for this would be @mckennabarlow. How about a more direct description like "Enables automatic addition of binding redirects to the app configuration file to override assembly unification."

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
melytc: I agree wording is not the best, I think I ended taking what the documentation describes it as (see helpurl below). @mckennabarlow or @cartermp, feel free to propose.

dmonroym (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes, line 195:
```suggestion
                DisplayName="Auto-generate binding redirects"
                Description="Add binding redirects automatically to App.config."
```

For `BoolProperty` the `Description` is used as the label of the check box. In general check boxes have a "directive" statement. There are a few violations of this currently which I will propose fixes for separately.

drewnoakes (approved):

MiYanni (approved):

PR #7321, opened 2021-06-15
Add Hot Reload UI

tmeschter:
In general we expect Hot Reload to be enabled by default for projects using the "Project" launch command, but there may be situations where users would like to turn it off. This change includes the following:

1. Defines the `hotReloadEnabled` value that will be stored in a launch profile in the launchSettings.json file.
2. Updates the `ProjectLaunchProfileExtensionValueProvider`, responsible for exposing settings in `ILaunchProfile.OtherSettings` to the CPS properties system, to expose that value as the `HotReloadEnabled` property. If not defined in the launch profile, the value defaults to "true".
3. Updates the ProjectDebugPropertyPage.xaml to include a checkbox surfacing the value to the user.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7321)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/ProjectLaunchProfileExtensionValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/ProjectLaunchProfileExtensionValueProvider.cs
@@ -27,6 +27,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         new[]
         {
             AuthenticationModePropertyName,
+            HotReloadEnabledPropertyName,
             NativeDebuggingPropertyName,
             RemoteDebugEnabledPropertyName,
             RemoteDebugMachinePropertyName,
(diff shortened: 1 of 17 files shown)


BillHiebert (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/ProjectLaunchProfileExtensionValueProvider.cs
```diff
@@ -71,6 +74,10 @@ public void OnSetPropertyValue(string propertyName, string propertyValue, IWrita
                     TrySetOtherProperty(launchProfile, LaunchProfileExtensions.RemoteAuthenticationModeProperty, propertyValue, string.Empty);
                     break;
 
+                case HotReloadEnabledPropertyName:
+                    TrySetOtherProperty(launchProfile, LaunchProfileExtensions.HotReloadEnabledProperty, bool.Parse(propertyValue), true);
```
swesonga, line 78: How are invalid values handled for these properties? Does the exception from bool.Parse get translated into something else in the UI?

swesonga (approved):

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/ProjectLaunchProfileExtensionValueProvider.cs
tmeschter: I'm not entirely sure what will happen. Throwing an exception here is probably not the best idea. Probably it should just ignore the input if it can't be converted to a bool, just like the `OnGet*` method ignores the stored value if it can't be converted to a bool.

PR #7323, opened 2021-06-15
Implement Hot Reload Support

tmeschter:
Implement Hot Reload support for Console, WinForms, and WPF apps in Ctrl+F5 scenarios.

When Hot Reload is enabled, the project system is responsible for implementing the following functionality:

1. Configuring the application for Hot Reload. This involves checking that the application supports Hot Reload (e.g., targets .NET 6 or higher; is built with the correct code gen settings, etc.) and setting various environment variables that will cause the CLR to load the "delta applier" assembly when the app runs.
2. Managing the Hot Reload session lifetime. The project system creates the Hot Reload session, tells it to start once the app has launched, and tells it to stop if/when the app exits.
3. Managing the process lifetime. There are situations where we will want to stop and restart the application. For example, if the process is running and the user invokes Ctrl+F5 again we want to stop the current process and start it again. A similar situation arises if the user tries to make edits that aren't supported by Hot Reload. In this case we want to stop the process, rebuild, and launch it again.

So really the project system is only involved at the beginning and end of the Hot Reload scenario. Everything that happens in the middle (watching for changes to source files, applying code changes to the running process, etc.) is handled elsewhere.

This change includes several base components that will be used both by dotnet/project-system as well as other project systems that build on top of us (e.g. WinUI, web tooling):

1. `IProjectHotReloadSession` - Represents the Hot Reload session within the project system, and is responsible for communicating with the other parts of the feature. This includes methods for setting up environment variables prior to process launch (`ApplyLaunchVariablesAsync`), starting and stopping the Hot Reload session, and a method to signal that any pending changes should be applied to the application (`ApplyChangesAsync`).
2. `IProjectHotReloadSessionCallback` - Handles communication from the other parts of the Hot Reload feature back to the project system and allows the session creator to customize the experience in various ways.
3. `IProjectHotReloadAgent` - A singleton service responsible for creating an `IProjectHotReloadSession` for a given version of the runtime and callback. Returns `null` if a session cannot be created.

The implementations of these three interfaces largely abstract away the underlying Hot Reload machinery.

For our internal use, we also define and implement the `IHotReloadDiagnosticOutputService`. This abstracts away and simplifies our interaction with the Hot Reload output window pane shared by all Hot Reload components for diagnostic messages.

We use these base components to implement Hot Reload support for console, WinForms, and WPF apps via a number of changes to the `ProjectLaunchTargetsProvider`.

1. Update `GetConsoleTargetForProfileAsync` to perform various checks for Hot Reload support and create a new _pending_ Hot Reload session.
2. Update `OnBeforeLaunchAsync` to stop the _active_ Hot Reload session (if any) so that we can start a new one. This will cause the running process to exit as well as ending the session.
3. Update `OnAfterLaunchAsync` to "attach" the pending Hot Reload session to the just-launched app (which is mostly a matter of hooking up an event handler to watch for it exiting), start the session, and make the pending session the active session.
4. Implement a bare-bones version of the `IProjectHotReloadSessionCallback` interface.

There's a fair bit of clean-up that could be done, as well as a number of improvements that will come on later. However, we would like to get the basic implementation in as soon as possible.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7323)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -47,6 +47,7 @@
     <PackageReference Update="Microsoft.VisualStudio.DataDesign.Common"                               Version="17.0.0-preview-2-31223-026"/>
     <PackageReference Update="Microsoft.VisualStudio.Data.Services"                                   Version="9.0.21023" />
     <PackageReference Update="Microsoft.VisualStudio.DataTools.Interop"                               Version="17.0.0-preview-2-31223-026" />
+    <PackageReference Update="Microsoft.VisualStudio.Debugger.Contracts"                              Version="17.2.0-beta.21278.1" />
     <PackageReference Update="Microsoft.VisualStudio.Designer.Interfaces"                             Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VisualStudio.ImageCatalog"                                    Version="17.0.0-previews-1-31321-016" />
     <PackageReference Update="Microsoft.VisualStudio.ManagedInterfaces"                               Version="8.0.50728" />
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -44,9 +44,9 @@
     <PackageReference Update="Microsoft.VSSDK.BuildTools"                                             Version="17.0.1056-Dev17PIAs-g9dffd635" />
     <PackageReference Update="Microsoft.VisualStudio.ComponentModelHost"                              Version="17.0.154-preview"/>
     <PackageReference Update="Microsoft.VisualStudio.Composition"                                     Version="16.9.20"/>
-    <PackageReference Update="Microsoft.VisualStudio.Data.Core"                                       Version="9.0.21023" />
+    <PackageReference Update="Microsoft.VisualStudio.Data.Core"                                       Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VisualStudio.DataDesign.Common"                               Version="17.0.0-preview-2-31223-026"/>
-    <PackageReference Update="Microsoft.VisualStudio.Data.Services"                                   Version="9.0.21023" />
+    <PackageReference Update="Microsoft.VisualStudio.Data.Services"                                   Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VisualStudio.DataTools.Interop"                               Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VisualStudio.Debugger.Contracts"                              Version="17.2.0-beta.21278.1" />
     <PackageReference Update="Microsoft.VisualStudio.Designer.Interfaces"                             Version="17.0.0-preview-2-31223-026" />
@@ -69,8 +69,8 @@
     <PackageReference Update="Microsoft.VisualStudio.XmlEditor"                                       Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VSDesigner"                                                   Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="VsWebSite.Interop"                                                      Version="16.8.30523.219"/>
-    <PackageReference Include="Microsoft.VisualStudio.Interop"                                        Version="17.0.0-previews-2-31421-281"/>
-    <PackageReference Include="Microsoft.Internal.VisualStudio.Interop"                               Version="17.0.0-previews-1-31321-016"/>
+    <PackageReference Include="Microsoft.VisualStudio.Interop"                                        Version="17.0.0-previews-2-31423-289"/>
+    <PackageReference Include="Microsoft.Internal.VisualStudio.Interop"                               Version="17.0.0-previews-2-31423-289"/>
     
     <!-- https://github.com/dotnet/roslyn/issues/53877 -->
     <PackageReference Include="StreamJsonRpc"                                                         Version="2.8.21" />
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -12,6 +12,7 @@
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json;
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json;
+      https://pkgs.dev.azure.com/dnceng/internal/_packaging/dotnet-tools-internal/nuget/v3/index.json;
     </RestoreSources>
   </PropertyGroup>
 
@@ -41,7 +42,7 @@
 
     <!-- VS SDK -->
     <PackageReference Update="Microsoft.VSSDK.BuildTools"                                             Version="17.0.1056-Dev17PIAs-g9dffd635" />
-    <PackageReference Update="Microsoft.VisualStudio.ComponentModelHost"                              Version="17.0.67-g9e37b637e6"/>
+    <PackageReference Update="Microsoft.VisualStudio.ComponentModelHost"                              Version="17.0.154-preview"/>
     <PackageReference Update="Microsoft.VisualStudio.Composition"                                     Version="16.9.20"/>
     <PackageReference Update="Microsoft.VisualStudio.Data.Core"                                       Version="9.0.21023" />
     <PackageReference Update="Microsoft.VisualStudio.DataDesign.Common"                               Version="17.0.0-preview-2-31223-026"/>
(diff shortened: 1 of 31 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadAgent.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
+{
+    public interface IProjectHotReloadAgent
+    {
+        IProjectHotReloadSession? CreateHotReloadSession(string runtimeVersion, IProjectHotReloadSessionCallback callback);
```
tmeschter, line 7: @BillHiebert My intention here is that `CreateHotReloadSession` will return `null` if it can't create the session for whatever reason. E.g., it can't find a a delta applier that matches `runtimeVersion`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
+{
+    public interface IProjectHotReloadSession
+    {
+        bool IsSupported { get; }
```
tmeschter, line 11: @BillHiebert Is this needed? If Hot Reload isn't supported for some reason, I would just as soon have `IProjectHotReloadAgent.CreateHotReloadSession` return `null`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Task StartSessionAsync(CancellationToken cancellationToken);
+
+        Task StopSessionAsync(CancellationToken cancellationToken);
+
+        Task ApplyChangesAsync(CancellationToken cancellationToken);
+
+        Task<bool> ApplyLaunchVariablesAsync(IDictionary<string, string> envVars, string configuration, CancellationToken cancellationToken);
```
tmeschter, line 19: @BillHiebert What is `configuration` needed for?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionCallback.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
+{
+    public interface IProjectHotReloadSessionCallback
+    {
+        bool IsSupported { get; }
```
tmeschter, line 10: @BillHiebert I don't think we need this. Presumably the code implementing the callback is also the code calling `IProjectHotReloadAgent.CreateHotReloadSession`--so if it knows Hot Reload isn't supported it just wouldn't create a session to being with.

dmonroym (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadAgent.cs
BillHiebert: @tmeschter that seems reasonable.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
BillHiebert: @tmeschter we could remove this and do as you say.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
BillHiebert: @tmeschter it is only supported for debug builds, so the idea was that the caller could just blindly call this and have it return false if not supported.  We could remove it and delegate that responsibility to the caller

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionCallback.cs
BillHiebert: @tmeschter.  At first I was unsure where all these properties would live so was unsure which code would know whether it is supported or not.  But I agree, if the project already has a setting to disable it, it should read that and not call to start a hot reload session

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
drewnoakes, line 19: Does this need a mutable dictionary?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/HotReloadDiagnosticOutputService.cs
```diff
@@ -0,0 +1,74 @@
...
+            Guid paneGuid = s_hotReloadOutputWindowPaneGuid;
+
+            Verify.HResult(outputWindow.CreatePane(ref paneGuid, VSResources.HotReloadOutputWindowPaneName, fInitVisible: 1, fClearWithSolution: 0));
+
+            Verify.HResult(outputWindow.GetPane(ref paneGuid, out IVsOutputWindowPane pane));
+
+            return pane;
+        }
+
+        public async Task WriteLineAsync(string outputMessage)
```
drewnoakes, line 63: Is this called a lot? If it's regularly called from the UI thread then `ValueTask` might be a good fit.

tmeschter: It's not clear to me how much we'll be writing to the output window. A lot of this output is for diagnostic purposes while the feature is in development; once the feature is solid I could see turning it off by default.

One consideration is that other parts of the Hot Reload pipeline are going to be outputting text as well. It might be weird or unhelpful if some components are outputting information and some aren't.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
tmeschter: We may need to merge the value of existing environment variables with the Hot Reload values. Either we mutate the dictionary, or we take in the current set of environment variables as an immutable object, produce a new immutable object with the changes, and then let the caller merge that back in.

tmeschter: One concern I have is that we shouldn't rely on the configuration _name_ in this check (whether it is done here or in the caller). Do we know what settings in Hot Reload depend son in Debug configurations?

tmeschter:
@BillHiebert This now has all the major pieces in place, except for a dependency on the shipping delta applier components.

Also, I need to implement process lifetime management in order to support restart scenarios.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
tmeschter: FYI, it looks like we really need to check for the following:

``` xml
<Optimize>false</Optimize>
<EmitDebugInformation>true</EmitDebugInformation>
```

melytc (approved):

build/import/Packages.targets
```diff
@@ -41,34 +41,35 @@
...
+    <PackageReference Update="Microsoft.VisualStudio.Validation"                                      Version="17.0.12-alpha" />
     <PackageReference Update="Microsoft.VisualStudio.Workspace.VSIntegration"                         Version="16.7.47-preview-0001" />
     <PackageReference Update="Microsoft.VisualStudio.XmlEditor"                                       Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VSDesigner"                                                   Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="VsWebSite.Interop"                                                      Version="16.8.30523.219"/>
-    <PackageReference Include="Microsoft.VisualStudio.Interop"                                        Version="17.0.0-previews-1-31321-016"/>
-    <PackageReference Include="Microsoft.Internal.VisualStudio.Interop"                               Version="17.0.0-previews-1-31321-016"/>
+    <PackageReference Include="Microsoft.VisualStudio.Interop"                                        Version="17.0.0-previews-2-31423-289"/>
```
dmonroym, line 71: Seems like the packages with this version will always need to be synced. Can you define this version in `Versions.props` so we only have to change it once?

dmonroym (approved):

PR #7328, opened 2021-06-18
Add feature flag to disable sending build diagnostics to roslyn when the new LSP pull diagnostics system is enabled

dibarbet:
In d17, we will be starting an experiment to switch away from direct Roslyn integration with the error list / tagger and over to an LSP based implementation.  In the new LSP based implementation, Roslyn will no longer be publishing build diagnostics and instead they will be published by CPS.  Once the feature flag fully roles out, this provider can be deleted.

To allow the CPS IVsErrorListProvider to take over, we check the appropriate feature flag and return NotHandled (as the CPS provider has Order(1) but this provider has Order(10)).

I considered implementing the check in roslyn, but there is no way to report NotHandled via the IVsLanguageServiceBuildErrorReporter2 interface.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7328)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
@@ -4,6 +4,7 @@ using System;
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using Microsoft.Build.Framework;
+using Microsoft.Internal.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.ProjectSystem.LanguageServices;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Threading;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
```diff
@@ -54,6 +69,13 @@ private async Task<AddMessageResult> AddMessageCoreAsync(TargetGeneratedError er
             if (!TryExtractErrorListDetails(error.BuildEventArgs, out ErrorListDetails details) || string.IsNullOrEmpty(details.Code))
                 return AddMessageResult.NotHandled;
 
+            // When this feature flag is enabled, build diagnostics will be published by CPS and should not be passed to roslyn.
```
davkean, line 72: Who de-dups the results in that case?

davkean:
The change looks good, but am interested in how we maintain behavior (de-duping) from before if CPS pushes the results to the Error List.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/LanguageServiceErrorListProvider.cs
dibarbet: De-duping will be handled by the (newish) diagnostics brokered service on the editor side.  CPS and the LSP client provide diagnostics to the service, it then de-dupes and handles tagging + error list display

davkean: How does editor know that Roslyn hides certain build errors if they don't appear during background analysis, but shows others? For example, when IntelliSense + Build is selected, it continues to show missing entry point, but hides semantic errors? It used to filter out the semantic errors when we pass it over to them.

dibarbet: So essentially what Aditya is doing is de-duping based on diagnostic id.  The intellisense window only shows intellisense tagged errors (produced by LSP).  The build window only shows build errors (tagged as build by CPS).  The build + intellisense shows both, but always prefers diagnostics (based on id) that are produced by the live system (LSP).  

See this doc for more details - https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/23530/Error-List-Diagnostic-API-Platform-level-De-Duplication

Note that we might not get it perfectly correct to start with, so we'll try it out under the experiment and iterate to get it correct.

drewnoakes (approved):

swesonga (approved):

dibarbet:
@drewnoakes or @swesonga would either of you be able to merge?  I'm unable to

Issue #7327, opened 2021-06-17
Create extension point for default launch profiles


tmeschter:
Currently the `LaunchSettingsProvider.UpdateProfilesAsync` method ensures that there is always at least one launch profile associated with the project. It does this by creating a profile matching the project name and with a launch command of "Project" if the launchSettings.json file is missing or is present but contains no profiles.

Without at least one launch profile we don't know how to start the app; if you try you get cryptic error messages from the debugger. This default profile allows us to properly start console, WinForms, and WPF apps "out of the box" without forcing them to include a trivial launchSettings.json in their project template, or requiring the user to that trivial profile later.

This approach does not work so well for other project types, however. For example, the "Project" launch command is not particularly useful for .NET MAUI apps. Those generally use the "iOS" or "Android" launch commands to deploy to an emulator, and they generate a number of in-memory profiles targeting different devices to do this. The default profile is still hanging around, however, and if the user selects it and tries to run they'll get cryptic deployment errors.

The default profile could be deleted, but it would be automatically re-created the next time the project system noticed the launchSettings.json was empty or missing.

To better handle this scenario, we could create a new extension point for the `LaunchSettingsProvider` to support different default launch profiles, or even no default at all. I imagine the interface would look something like this:

``` C#
[ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Extension, Cardinality = ImportCardinality.ZeroOrMore)]
public interface IDefaultLaunchProfileProvider
{
    ILaunchProfile? CreateDefaultProfile();
}
```

Implementations of this interface would be tagged with `[AppliesTo(...)]` attributes so they are only picked up by the appropriate project types, as well as `[Order(...)]` attribute to specify a priority. The `LaunchSettingsProvider` would import all applicable implementations into an `OrderPrecedenceImportCollection`. When it is unable to find a launchSettings.json, it consults the highest-priority implementation (and only that one) to see if it wishes to create a default.

The project system itself would provide an implementation of this interface with the lowest-possible priority that creates the same default as today.

joj:
@tmeschter this is my first try at it: https://github.com/dotnet/project-system/pull/7329. It's failing tests because the project in those tests doesn't have capabilities. I need your help with that one, I'm not sure how to properly modify that test. Thanks! J

PR #7329, opened 2021-06-18
Added new IDefaultLaunchProfileProvider

joj:
As specified in https://github.com/dotnet/project-system/issues/7327, we
need a way to remove de default launch profile for project systems
where it doesn't represent a valid option. This commit adds an extension
point for providing a new provider that lets extensions proide a
different default that will make more sense for their projects, or no
provider at all.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7329)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/DefaultLaunchProfileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/DefaultLaunchProfileProvider.cs
@@ -1,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.IO;
+using System.ComponentModel.Composition;
+using ExportOrder = Microsoft.VisualStudio.ProjectSystem.OrderAttribute;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    [Export(typeof(IDefaultLaunchProfileProvider))]
+    [AppliesTo(ProjectCapability.LaunchProfiles)]
+    [ExportOrder(int.MinValue)] 
+    internal class DefaultLaunchProfileProvider : IDefaultLaunchProfileProvider
+    {
+        private readonly UnconfiguredProject _project;
+
+        [ImportingConstructor]
+        public DefaultLaunchProfileProvider(UnconfiguredProject project)
+        {
+            _project = project;
+        }
+
+        public ILaunchProfile? CreateDefaultProfile()
+        {
+            return new LaunchProfile { Name = Path.GetFileNameWithoutExtension(_project.FullPath), CommandName = LaunchSettingsProvider.RunProjectCommandName };
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/DefaultLaunchProfileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/DefaultLaunchProfileProvider.cs
@@ -8,7 +8,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
 {
     [Export(typeof(IDefaultLaunchProfileProvider))]
     [AppliesTo(ProjectCapability.LaunchProfiles)]
-    [ExportOrder(int.MinValue)] 
+    [ExportOrder(Order.Default)] 
     internal class DefaultLaunchProfileProvider : IDefaultLaunchProfileProvider
     {
         private readonly UnconfiguredProject _project;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
@@ -4,9 +4,15 @@ using Microsoft.VisualStudio.Composition;
 
 namespace Microsoft.VisualStudio.ProjectSystem.Debug
 {
+    /// <summary>
+    /// Provides a default <see cref="ILaunchProfile"/> to be displayed when no profiles are provided.
+    /// </summary>
     [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Extension, Cardinality = ImportCardinality.ZeroOrMore)]
     public interface IDefaultLaunchProfileProvider
     {
+        /// <summary>
+        /// Gets the default <see cref="ILaunchProfile"/>. Return null to remove the default profile.
+        /// </summary>
         ILaunchProfile? CreateDefaultProfile();
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
@@ -1,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Extension, Cardinality = ImportCardinality.ZeroOrMore)]
+    public interface IDefaultLaunchProfileProvider
+    {
+        ILaunchProfile? CreateDefaultProfile();
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
@@ -76,10 +76,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
                                                                                                                     project);
             SourceControlIntegrations = new OrderPrecedenceImportCollection<ISourceCodeControlIntegration>(projectCapabilityCheckProvider: project);
 
+            DefaultLaunchProfileProviders = new OrderPrecedenceImportCollection<IDefaultLaunchProfileProvider>(ImportOrderPrecedenceComparer.PreferenceOrder.PreferredComesFirst, project);
+
             _projectSubscriptionService = projectSubscriptionService;
             _appDesignerSpecialFileProvider = appDesignerSpecialFileProvider;
             _projectFaultHandler = projectFaultHandler;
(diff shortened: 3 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/DefaultLaunchProfileProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System.IO;
+using System.ComponentModel.Composition;
+using ExportOrder = Microsoft.VisualStudio.ProjectSystem.OrderAttribute;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    [Export(typeof(IDefaultLaunchProfileProvider))]
+    [AppliesTo(ProjectCapability.LaunchProfiles)]
+    [ExportOrder(int.MinValue)]
```
tmeschter, line 11: use `Order.Default` instead of `int.MinValue`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -90,6 +97,11 @@ internal class LaunchSettingsProvider : OnceInitializedOnceDisposed, ILaunchSett
         [ImportMany]
         protected OrderPrecedenceImportCollection<ISourceCodeControlIntegration> SourceControlIntegrations { get; set; }
 
+        [ImportMany]
+        protected OrderPrecedenceImportCollection<IDefaultLaunchProfileProvider> DefaultLaunchProfileProviders { get; set; }
+
+        private readonly Lazy<ILaunchProfile?> _defaultLaunchProfile;
```
tmeschter, line 103: This should go up with the other `private readonly` fields near the top.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -252,7 +264,9 @@ protected async Task UpdateProfilesAsync(string? updatedActiveProfileName)
                 // won't be called on F5 and the user will see a poor error message
                 if (launchSettingData.Profiles.Count == 0)
                 {
-                    launchSettingData.Profiles.Add(new LaunchProfileData() { Name = Path.GetFileNameWithoutExtension(_commonProjectServices.Project.FullPath), CommandName = RunProjectCommandName });
+                    var defaultLaunchProfile = _defaultLaunchProfile.Value;
+                    if (defaultLaunchProfile != null)
+                        launchSettingData.Profiles.Add(new LaunchProfileData() { Name = defaultLaunchProfile.Name, CommandName = defaultLaunchProfile.CommandName});
```
tmeschter, line 269: Use `LaunchProfileData.FromILaunchProfile()`; this will pick up all the data rather than just the `Name` and `CommandName`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -252,7 +264,9 @@ protected async Task UpdateProfilesAsync(string? updatedActiveProfileName)
                 // won't be called on F5 and the user will see a poor error message
                 if (launchSettingData.Profiles.Count == 0)
                 {
-                    launchSettingData.Profiles.Add(new LaunchProfileData() { Name = Path.GetFileNameWithoutExtension(_commonProjectServices.Project.FullPath), CommandName = RunProjectCommandName });
+                    var defaultLaunchProfile = _defaultLaunchProfile.Value;
```
tmeschter, line 267: Nit: prefer explicit typing to `var`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -252,7 +264,9 @@ protected async Task UpdateProfilesAsync(string? updatedActiveProfileName)
...
+                    if (defaultLaunchProfile != null)
```
tmeschter, line 268: Nit: The code is not consistent about this, but we generally prefer the `defaultLaunchProfile is not null` form over `defaultLaunchProfile != null` in new code.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -76,11 +76,18 @@ internal class LaunchSettingsProvider : OnceInitializedOnceDisposed, ILaunchSett
...
 
+            DefaultLaunchProfileProviders = new OrderPrecedenceImportCollection<IDefaultLaunchProfileProvider>(ImportOrderPrecedenceComparer.PreferenceOrder.PreferredComesFirst, project);
+
             _projectSubscriptionService = projectSubscriptionService;
             _appDesignerSpecialFileProvider = appDesignerSpecialFileProvider;
             _projectFaultHandler = projectFaultHandler;
             _launchSettingsFilePath = new AsyncLazy<string>(GetLaunchSettingsFilePathNoCacheAsync, commonProjectServices.ThreadingService.JoinableTaskFactory);
 
+            _defaultLaunchProfile = new Lazy<ILaunchProfile?> (() => {
+                    return DefaultLaunchProfileProviders?.First()?.Value?.CreateDefaultProfile();
```
tmeschter, line 87: You need `FirstOrNull` as `First` will throw if the collection is empty.

tmeschter (changes requested):
This is exactly what I had in mind, there are just a few things I'd like you to clean up before we merge it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
joj: I fixed the others (will send the commit in a minute). On this one, though, if no defaultlaunchprofileprovider exists, MEF will fail the import. We'll never get this far. I'm not checking if _defaultLaunchProfile is null when I use it for this reason. That's also what's failing the test right now. If you want I'll do FirstOrDefault, no problem... but I think First makes the fact that null is not an option very explicit.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/DefaultLaunchProfileProvider.cs
joj: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
joj: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
joj: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
joj: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
joj: Fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: The import shouldn't fail, even if there are no `IDefaultLaunchProfileProvider`s. The `ProjectSystemContract` attribute explicitly states that there can be zero or more implementations.

tmeschter: To fix the test I think you can just update `LaunchSettingsProviderTests.GetLaunchSettingsProvider` to explicitly add an instance of the `DefaultLaunchProfileProvider` to the created `LaunchSettingsUnderTest` object.

joj: Thanks Tom, I'll correct those two things and ping you again

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Extension, Cardinality = ImportCardinality.ZeroOrMore)]
+    public interface IDefaultLaunchProfileProvider
```
tmeschter, line 8: Please add doc comments for the type and method.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        ILaunchProfile? CreateDefaultProfile();
```
drewnoakes, line 10: Please explain when the return value will be `null` when you doc this.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
joj: I just added a "value can be null" comment there. I can provide a longer explanation if needed. The whole point of this interface is to be able to provide a null value in cases where we don't want a default launch profile (which happens for Xamarin but may happen for others).

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IDefaultLaunchProfileProvider.cs
drewnoakes: Thanks. I think explaining that a default launch profile is optional would make it clearer. The type already shows it may be null, but doesn't give any indication why.

joj:
Hi! @drewnoakes @tmeschter do we need more changes to merge this? Thanks! J

tmeschter:
@joj Merged. Thank you!

PR #7336, opened 2021-06-20
Treat 'folder' as an alias of 'directory'

### LINKED ISSUE #7331: "Working directory" does not have a browse button
The "Working directory" property in the Launch Profiles should have a browse button, but does not:

image

drewnoakes:
Fixes #7331

`StringProperty` may specify a `SubType`, which modifies how it is displayed in the UI. It turns out that rule files commonly use both `path` and `directory`, and these appear to be equivalent in semantics.

This is the root cause of the failure to show a "browse" button alongside the "working directory" property's editor.

Previously a value of `folder` would be ignored. This commit makes it equivalent to `directory`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7336)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
```diff
@@ -72,7 +72,7 @@ IEnumerable<IEntityValue> createEditorValues()
                     {
                         yield return CreateEditorValue(queryExecutionContext, parent, new ValueEditor { EditorType = "FilePath" }, properties);
                     }
-                    else if (stringProperty.Subtype == "directory")
+                    else if (stringProperty.Subtype == "folder" || stringProperty.Subtype == "directory")
```
swesonga, line 75: Don't we normally explicitly specify the type of string comparison?

swesonga (approved):

jjmew (approved):
:shipit:

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIPropertyEditorDataProducer.cs
drewnoakes: Thanks, I added one and made it case-insensitive.

PR #7339, opened 2021-06-21
Localized file check-in by OneLocBuild Task: Build definition ID 981: Build ID 1197416

MiYanni:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7339)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/BuildPropertyPage.xaml.tr.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/BuildPropertyPage.xaml.tr.xlf
@@ -4,12 +4,12 @@
     <body>
       <trans-unit id="BoolProperty|AllowUnsafeBlocks|Description">
         <source>Allow code that uses the 'unsafe' keyword to compile.</source>
-        <target state="needs-review-translation">'unsafe' anahtar sözcüğünü kullanan kodun derlenmesine izin verir.</target>
+        <target state="new">Allow code that uses the 'unsafe' keyword to compile.</target>
         <note />
       </trans-unit>
       <trans-unit id="BoolProperty|AllowUnsafeBlocks|DisplayName">
         <source>Unsafe code</source>
-        <target state="needs-review-translation">Güvenli olmayan koda izin ver</target>
+        <target state="new">Unsafe code</target>
         <note />
       </trans-unit>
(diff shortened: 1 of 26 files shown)


jjmew (approved):

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/BuildPropertyPage.xaml.tr.xlf
```diff
@@ -4,12 +4,12 @@
     <body>
       <trans-unit id="BoolProperty|AllowUnsafeBlocks|Description">
         <source>Allow code that uses the 'unsafe' keyword to compile.</source>
-        <target state="needs-review-translation">'unsafe' anahtar sözcüğünü kullanan kodun derlenmesine izin verir.</target>
+        <target state="new">Allow code that uses the 'unsafe' keyword to compile.</target>
```
drewnoakes, line 7: This looks suspicious. The translation has been undone.

There are a few others in this file. Possibly elsewhere too.

MiYanni: This text was changed from `Allows` to `Allow`. You reviewed that PR if I remember correctly. This undid the translation. The new LEGO PRs have the new translation for that (for some languages). So, it'll update the XLFs once that runs through the system. I just merged the PR: https://github.com/dotnet/project-system/pull/7343/files

The `TR` language specific translations are lagging behind, so they aren't in these new LEGO PRs. But, you can see how the translations for the other languages were updated. All translations for all languages don't happen all at the same time.

drewnoakes: > You reviewed that PR if I remember correctly

I wrote that PR :) I'm ok with the change to the English text. My assumption in the above comment was that this file contained the _source_ of the translation. You explained elsewhere that this is not the source, but rather an output from upstream source. So long as the original translators have the same source string, this is fine.

My concern was for a scenario where the previously translated string was lost, and the new translation becomes materially different to the original. I think that could be confusing as a consumer of the translation.

PR #7341, opened 2021-06-21
Expose process information

tmeschter:
Currently, the `LaunchProfilesDebugLaunchProvider` calls `IDebugProfileLaunchTargetsProvider.OnBeforeLaunchAsync` right before launching processes, and `IDebugProfileLaunchTargetsProvider.OnAfterLaunchAsync` right after. The idea is that, in addition to creating the launch settings for the processes, the launch targets provider may want to do something special right before or right after processes launch like turn on some logging or show some UI.

There is some information available that isn't currently passed to the launch targets provider. One is the set of `IDebugLaunchSettings` objects that is actually going to be passed to CPS (and then on to the debugger) to start the processes, and the other is the set of `VsDebugTargetProcessInfo`s providing info about the started processes. We're going to need these, especially the latter, for Hot Reload scenarios. For example, when we start a process with Hot Reload enabled we need to know when it exits so that we can end the Hot Reload session. To do this, we need the process ID from the `VsDebugTargetProcessInfo`.

This change introduces `IDebugProfileLaunchTargetsProvider4` which provides replacements for `IDebugProfileLaunchTargetsProvider.OnBeforeLaunchAsync` and `IDebugProfileLaunchTargetsProvider.OnAfterLaunchAsync` that pass along the relevant information.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7341)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
@@ -154,6 +154,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
 
             public void OnComplete(int hr, uint debugTargetCount, VsDebugTargetProcessInfo[] processInfoArray)
             {
+                if (_targetProfile is IDebugProfileLaunchTargetsProvider4 targetsProvider4)
+                {
+                    _threadingService.ExecuteSynchronously(() => targetsProvider4.OnAfterLaunchAsync(_launchOptions, _activeProfile, processInfoArray));
+                }
                 if (_targetProfile != null)
                 {
                     _threadingService.ExecuteSynchronously(() => _targetProfile.OnAfterLaunchAsync(_launchOptions, _activeProfile));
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/LaunchProfilesDebugLaunchProvider.cs
```diff
@@ -154,6 +154,10 @@ public LaunchCompleteCallback(IProjectThreadingService threadingService, DebugLa
 
             public void OnComplete(int hr, uint debugTargetCount, VsDebugTargetProcessInfo[] processInfoArray)
             {
+                if (_targetProfile is IDebugProfileLaunchTargetsProvider4 targetsProvider4)
+                {
+                    _threadingService.ExecuteSynchronously(() => targetsProvider4.OnAfterLaunchAsync(_launchOptions, _activeProfile, processInfoArray));
```
swesonga, line 159: Does this code need an else block to avoid calling _targetProfile.OnAfterLaunchAsync(...)?

drewnoakes: I wonder the same. Whatever the answer, it'd be good to document the behaviour on the interfaces in question.

drewnoakes: The below case has an else. I suppose both before/after should behave identically.

tmeschter: See this is why we have code reviews...

tmeschter: Yeah, there is supposed to be an `else`. Fixed.

drewnoakes (approved):

PR #7348, opened 2021-06-22
Support WebView2 debugging

### LINKED ISSUE #7333: New Property Page doesn't have the WebView2 debugging option
We added the support for WebView2 debugging in .Net Core apps in VS [here](https://github.com/dotnet/project-system/pull/6779). To enable it, users had to select the option in the debug properties page:

image

However, the new properties page doesn't show the UI option anymore. This will block our customers to enable and use JavaScript debugging in WebView2.

image

tmeschter:
Fixes #7333.

Update the ProjectDebugPropertyPage.xaml and ExecutableDebugPropertyPage.xaml to support WebView2 debugging. This is present in the legacy property pages but had not yet been implemented in the new Launch Profiles UI.

Screenshot:

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7348)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ExecutableDebugPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ExecutableDebugPropertyPage.xaml
@@ -77,4 +77,7 @@
                 DisplayName="Enable SQL Server debugging"
                 Description="Enable debugging of SQL scripts and stored procedures." />
 
+  <BoolProperty Name="WebView2Debugging"
+                DisplayName="Enable WebView2 debugging"
+                Description="Enable JavaScript debugger for Microsoft Edge (Chromium) based WebView2. Requires the JavaScript Diagnostics component." />
 </Rule>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/ProjectLaunchProfileExtensionValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/ProjectLaunchProfileExtensionValueProvider.cs
@@ -31,7 +31,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
             NativeDebuggingPropertyName,
             RemoteDebugEnabledPropertyName,
             RemoteDebugMachinePropertyName,
-            SqlDebuggingPropertyName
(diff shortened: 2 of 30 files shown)


tmeschter:
FYI @rakatyal.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ExecutableDebugPropertyPage.xaml
```diff
@@ -77,4 +77,7 @@
                 DisplayName="Enable SQL Server debugging"
                 Description="Enable debugging of SQL scripts and stored procedures." />
 
+  <BoolProperty Name="WebView2Debugging"
+                DisplayName="Enable WebView2 debugging"
+                Description="Enable JavaScript debugger for Microsoft Edge (Chromium) based WebView2. Requires the JavaScript Diagnostics component." />
```
drewnoakes, line 82: I wonder if we could have a visibility condition based upon VS workload/component availability...

drewnoakes (approved):

drewnoakes:
@tmeschter you also called out (https://github.com/dotnet/project-system/issues/7333#issuecomment-864264540) a follow on work item for the WebTools folks. Is that being tracked anywhere? cc @BillHiebert

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ExecutableDebugPropertyPage.xaml
tmeschter: Yeah, there are definitely improvements we could make here, but this brings us more or less back up to parity with the experience in the old property pages.

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ExecutableDebugPropertyPage.xaml
drewnoakes: Sure, it was more of a forward thinking idea. It seems like it would be a useful facility.

drewnoakes: Tracked by https://github.com/dotnet/project-system/issues/7358

swesonga (approved):

PR #7362, opened 2021-06-24
Add Prefer32Bit property

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/6989

Adds the Prefer32Bit property to the Build property page. This property only applies to .NET framework and executables. Information can be found on either of these pages:
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/output#platformtarget
- https://docs.microsoft.com/en-us/dotnet/framework/tools/corflags-exe-corflags-conversion-tool#parameters
- https://dzone.com/articles/what-anycpu-really-means-net
- https://stackoverflow.com/questions/12066638/what-is-the-purpose-of-the-prefer-32-bit-setting-in-visual-studio-and-how-does

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7362)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -76,6 +76,18 @@
                DisplayName="Annotations" />
   </EnumProperty>
 
+  <BoolProperty Name="Prefer32Bit"
+                DisplayName="Prefer 32-bit"
+                Description="Set the 32-bit preferred flag for the application. Only applies to executables."
+                HelpUrl="https://go.microsoft.com/fwlink/?linkid=2166468"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(has-net-framework)</NameValuePair.Value>
+      </NameValuePair>
+    </BoolProperty.Metadata>
+  </BoolProperty>
+
   <BoolProperty Name="AllowUnsafeBlocks"
                 DisplayName="Unsafe code"
                 Description="Allow code that uses the 'unsafe' keyword to compile."
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -85,6 +85,7 @@
       <NameValuePair Name="DependsOn" Value="Build::PlatformTarget" />
       <NameValuePair Name="DependsOn" Value="Application::OutputType" />
       <NameValuePair Name="VisibilityCondition">
+        <!-- Visibility based on: https://github.com/dotnet/msbuild/blob/9bcc06cbe19ae2482ab18eab90a82fd079b26897/src/Tasks/Microsoft.NETFramework.CurrentVersion.props#L87 -->
         <NameValuePair.Value>
           (and
             (has-net-framework)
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -76,6 +76,18 @@
                DisplayName="Annotations" />
   </EnumProperty>
 
+  <BoolProperty Name="Prefer32Bit"
+                DisplayName="Prefer 32-bit"
+                Description="Set the 32-bit preferred flag for the application. Only applies to executables."
```
drewnoakes, line 81: Also, it would be good if the `Description` explained a little of the effect this option has on the output. For example:

```suggestion
                Description="Run in 32-bit mode on systems that support both 32-bit and 64-bit applications."
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -76,6 +76,18 @@
...
+                HelpUrl="https://go.microsoft.com/fwlink/?linkid=2166468"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(has-net-framework)</NameValuePair.Value>
```
drewnoakes, line 86: It seems like we should extend the visibility condition to test whether `OutputType` is not `Library` (removing the need for "Only applies to executables" in the description.

drewnoakes, line 86: Doesn't Prefer32Bit only make sense with `AnyCPU`? I don't think it has any effect if the platform target has any other value (for example, `x86`, `x64`, `ARM`, ...). The visibility condition should probably be extended to include that also.

drewnoakes (commented):
I believe we should extend the visibility condition further here, as commented.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
MiYanni: That is something I couldn't get a concrete explanation on. There is clearly a `anycpu32bitpreferred` platform, but I don't know if this flag is directly related to that platform. I forgot to link this originally in the description of this PR:
https://dzone.com/articles/what-anycpu-really-means-net

Basically, `32bitpref` is just a flag on the metadata of the executable. This properly might link directly to setting that flag. Why I suspect this is because one would think that we would actually set the value `anycpu32bitpreferred` in the PlatformTarget property. In this PR, I linked to that documentation, but thinking about it now, linking to the flags documentation might be more appropriate.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
MiYanni: Updated.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
MiYanni: Added to visibility condition.

MiYanni: Added to visibility condition.

MiYanni:
I've also tested the new visibility condition for it. Gif here:
Prefer32BitWorking

jjmew (approved):
:shipit:

drewnoakes (approved):
Looks great.

Can you file a bug for the TFM fighting you saw in that gif?

swesonga (approved):

MiYanni:
@drewnoakes Created an issue here for the framework not switching: https://github.com/dotnet/project-system/issues/7367

PR #7368, opened 2021-06-28
Remove Microsoft.CodeAnalysis.FxCopAnalyzers

tmeschter:
This package has been deprecated, with the relevant analyzers now being included in the SDK itself. Here we remove the package, and enable the analyzers from the SDK using the `EnableNETAnalyzers` property--the analyzers are on by default for .NET 5.0 and up, but since we're targeting .NET Framework 4.7.2 we need to explicitly turn them on.

Other work I'm going is running into conflicts between the version of the package we currently use, and a different version used by another package. Rather than try to reconcile those, I'd just as soon move us to the supported distribution of these analyzers.

One difference between the package and the SDK is that the SDK turns on fewer of these analyzers by default. However, we already explicitly enable or disable many CA* rules in our .editorconfig files, and I could only identify one additional rule that needed to be turned back on (CA1819). It may be useful to review the available rules and decide which ones are appropriate for our code.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7368)


swesonga:
@tmeschter Do you want to file an issue to track reviewing the available rules? I'm guessing you're referring to the rules at https://docs.microsoft.com/en-us/visualstudio/code-quality/fxcop-rule-port-status?view=vs-2019

swesonga (approved):

drewnoakes (approved):

PR #7370, opened 2021-06-28
Project cache: Upgradeable read lock support in project cache mode

arkalyanms:
For upgradeable read lock support, in scenarios where the project intends to upgrade to a write lock, we need to ensure the project  is setup to take on writes. If the project is in cache mode, the new public api in ConfiguredProject2 will evaluate the project or no op if already evaluated. Write lock is implicitly setup to break out of cache mode.

CPS side changes that exposed the public api is here - https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/332239

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7370)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectAccessor.cs
```diff
@@ -105,6 +105,7 @@ public async Task OpenProjectXmlForWriteAsync(UnconfiguredProject project, Actio
             Requires.NotNull(project, nameof(project));
             Requires.NotNull(project, nameof(action));
 
+            await ((ConfiguredProject2)project).EnsureProjectEvaluatedAsync();
```
swesonga, line 108: Pardon my ignorance about this area, but why is it crucial for the project to have been evaluated at this point?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectAccessor.cs
```diff
@@ -125,6 +126,7 @@ public async Task OpenProjectForWriteAsync(ConfiguredProject project, Action<Pro
...
```
swesonga, line 129: why isn't this changed needed in EnterWriteLockAsync?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectAccessor.cs
arkalyanms: In the cache mode, even the msbuild construction model is cached for imports and properties from the imports. The xml/construction model write would need working with actual construction xmls. Perhaps it would make sense to decouple need to load the actual construction model vs need for evaluation given a construction model write would trigger an evaluation anyway down the road.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectAccessor.cs
arkalyanms: It's probably needed there too. Can you elaborate a little on what EnterWriteLockAsync is meant to be used for? I noticed it's not used at the moment. 

EnsureProjectEvaluatedAsync is meant to be used whenever we need an actual writeable evaluation result.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ProjectAccessor.cs
arkalyanms: The more I think about this, I feel this should be removed. I am planning to switch out the construction model selectively out of the cache mode on demand instead of full cache mode break out for the project.

drewnoakes, line 108: Is this right? It's casting an UNconfigured project to a configured project.

PR #7398, opened 2021-07-13
Remove unnecessary package

ocallesp:
Code cleanup

NuGet merged `NuGet.SolutionRestoreManager.Interop` into `NuGet.VisualStudio`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7398)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -58,7 +58,7 @@
     <PackageReference Update="Microsoft.VisualStudio.Shell.15.0"                                      Version="17.0.0-previews-2-31421-281" />
     <PackageReference Update="Microsoft.VisualStudio.Shell.Design"                                    Version="17.0.0-previews-2-31421-281" />
     <PackageReference Update="Microsoft.VisualStudio.Shell.Framework"                                 Version="17.0.0-previews-2-31421-281" />
-    <PackageReference Update="Microsoft.VisualStudio.TemplateWizardInterface"                         Version="17.0.0-preview-2-31223-026" />
+    <PackageReference Update="Microsoft.VisualStudio.TemplateWizardInterface"                         Version="17.0.0-previews-1-31321-016" />
     <PackageReference Update="Microsoft.VisualStudio.Threading"                                       Version="17.0.39-alpha" />
     <PackageReference Update="Microsoft.VisualStudio.Threading.Analyzers"                             Version="17.0.39-alpha" />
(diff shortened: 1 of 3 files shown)


build/import/Packages.targets
```diff
@@ -89,8 +89,7 @@
     <PackageReference Update="Roslyn.Diagnostics.Analyzers"                                           Version="3.3.2" />
 
     <!-- NuGet -->
-    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="5.8.0-preview.2.6776" />
-    <PackageReference Update="NuGet.VisualStudio"                                                     Version="6.0.0-preview.0.2" />
+    <PackageReference Update="NuGet.SolutionRestoreManager.Interop"                                   Version="6.0.0-preview.2.129" />
```
drewnoakes, line 92: This is just an update. You'll need to find the includes elsewhere and change them too.

zivkan: GitHub's not letting me start a new conversation on the same line of code :(

NuGet merged `NuGet.SolutionRestoreManager.Interop` into `NuGet.VisualStudio`, not the other way around. You need to keep `NuGet.VisualStudio`.

ocallesp: Thank you @zivkan

davkean (approved):

MSylvia:
@ocalles  @MiYanni Looks like its because of a naming convention change. The newer package left off the `s` in `previews`:
```
17.0.0-previews-1-31321-016
17.0.0-preview-2-31223-026
```

If both package names contain the `s` (or remove the `s`) it would not evaluate to a downgrade.

ocallesp:
> @ocalles @MiYanni Looks like its because of a naming convention change. The newer package left off the `s` in `previews`:
> 
> ```
> 17.0.0-previews-1-31321-016
> 17.0.0-preview-2-31223-026
> ```
> 
> If both package names contain the `s` (or remove the `s`) it would not evaluate to a downgrade.

Thank you.

I updated the commit with your suggested package version.

MSylvia:
@ocallesp I'm not suggesting one package version over the other. I was just pointing out the error was caused by nuget semvar parsing and the missing `s`.

I believe it would up to the author of  `Microsoft.VisualStudio.TemplateWizardInterface` to fix the naming convention so that nuget can correctly parse it, assuming that "preview 2" is the newer package of the two.

PR #7401, opened 2021-07-15
Remove unused IProjectAccessor method

swesonga:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7401)


drewnoakes:
Is the sense that this method would not be used in future? I'm generally for removing unused code, but I can imagine someone being surprised in future that the other methods exist but not this one, then wondering whether there's some fundamental reason it does not exist.

swesonga:
> Is the sense that this method would not be used in future? I'm generally for removing unused code, but I can imagine someone being surprised in future that the other methods exist but not this one, then wondering whether there's some fundamental reason it does not exist.

Agreed. Code for accessing the project shouldn't have to be rediscovered. I'll close this PR.

PR #7402, opened 2021-07-16
Add steps for using embedded XAML property pages

philliphoff:
Adds steps for using embedded XAML property pages when defining new launch profile types. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7402)

### DIFF
--- a/docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
+++ b/docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
@@ -103,125 +103,81 @@ And you're done. Projects that import the .targets file will now show this UI wh
 
 ## Option 2: Embedded XAML file
 
-In this option the XAML file is embedded in an assembly as a resource and discovered by means of a MEF export. Compared to Option 1, this requires more initial setup but does not require you to distribute an additional file. This may be an attractive option if you are already exporting MEF components for use in Visual Studio.
+In this option the XAML file is embedded in an assembly as a resource and discovered by means of a MEF export. Compared to Option 1, this requires more initial setup but does not require you to distribute an additional file. This may be an attractive option if you are already exporting MEF components for use in Visual Studio--for example, via an extension.
 
-### Step 1: Define the `XamlPropertyRule` item
+### Step 1: Add the Project System SDK
 
-Add the following item to your .csproj file:
+Use the NuGet Package Manager to add the Microsoft.VisualStudio.ProjectSystem.Sdk package package to your project.
+
+### Step 2: (optional): Add the Microsoft.Build.Framework package
+
+Use the NuGet Package Manager to add the Microsoft.Build.Framework package to your project. This is an optional step, but it will allow the XAML editor to find the [Rule](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.xamltypes.rule) type (and related types) and provide code completion, tool tips, Go to Definition, and other features while you type.
+
+### Step 3: Define the XAML file
+
+Add a new XAML file named "MyLaunchProfile.xaml" to your project. Depending on how the file is created you may end up with a `<Page>` item in your project but this is not what we want as we're not using the file to describe a piece of WPF UI.
+
+Update your project to replace the `<Page>` item with the following:
 
 ```xml
-  <ItemGroup>
-    <XamlPropertyRule Include="path\to\MyPropertyPage.xaml">
-      <Namespace>MyNamespace</Namespace>
-      <DataAccess>IRule</DataAccess>
-      <RuleInjectionClassName>MyPropertyPageProperties</RuleInjectionClassName>
-    </XamlPropertyRule>
-  </ItemGroup>
+<ItemGroup>
+  <XamlPropertyRule Include="MyLaunchProfile.xaml">
+    <CopyToOutputDirectory>Never</CopyToOutputDirectory>
+    <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
+    <SubType>Designer</SubType>
+    <DataAccess>None</DataAccess>
+    <RuleInjection>None</RuleInjection>
+  </XamlPropertyRule>
+</ItemGroup>
 ```
 
-### Step 2: Register the property page and properties
+### Step 4: Describe the property page
 
-Add a partial class that corresponds to the properties defined within the rule and generated during build:
+This is the same as Step 4 under "Option 1: XAML file on disk" (above). Please see that step for details.
+
+### Step 5: Register the property page
+
+Add a class to hold the MEF metadata describing the embedded XAML `Rule`:
 
 ```csharp
 using Microsoft.VisualStudio.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
-using System;
-using System.ComponentModel.Composition;
 
 namespace MyNamespace
 {
-    [Export]
-    [AppliesTo(MyProject.UniqueCapability)]
-    internal partial class MyPropertyPageProperties : StronglyTypedPropertyAccess
+    internal static class RuleExporter
     {
-        [ExportPropertyXamlRuleDefinition("<Assembly Name>, Version=<Assembly Version>, Culture=<Assembly Culture>, PublicKeyToken=<Assembly Public Key Token>", "XamlRuleToCode:MyPropertyPage.xaml", "Project")]
-        [AppliesTo(MyProject.UniqueCapability)]
-        private object MyRule { get { throw new NotImplementedException(); } }
-
-        [ImportingConstructor]
-        public MyPropertyPageProperties(ConfiguredProject configuredProject)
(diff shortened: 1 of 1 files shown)


philliphoff:
@tmeschter This draft represents notes I've collected as I'm going through the process.  Feel free to suggest changes, make corrections, massage as necessary, etc.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+using Microsoft.VisualStudio.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using System;
+using System.ComponentModel.Composition;
+
+namespace MyNamespace
+{
+    [Export]
+    [AppliesTo(MyProject.UniqueCapability)]
+    internal partial class MyPropertyPageProperties : StronglyTypedPropertyAccess
```
philliphoff, line 136: I'm not completely sure that, for the purposes of launch profile property pages, we actually need the generated rule properties code, though it's a convenient place to hang the rule registration from.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+    {
+        [ExportPropertyXamlRuleDefinition("<Assembly Name>, Version=<Assembly Version>, Culture=<Assembly Culture>, PublicKeyToken=<Assembly Public Key Token>", "XamlRuleToCode:MyPropertyPage.xaml", "Project")]
```
drewnoakes, line 138: Maybe call out that this can be simplified using an approach similar to that used in this repo's `ExportRuleAttribute`.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+            : base(configuredProject, project)
+        {
+        }
+    }
+}
+```
+
+### Step 3: Register the assembly codebase
+
+In order for Visual Studio to locate the rule, you must register the codebase of the assembly embedding the rule from the Visual Studio package via the `ProvideCodeBaseAttribute`.
```
drewnoakes, line 168: Is this true? The `dotnet/project-system` repo exports rules yet doesn't seem to use this attribute.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+    },
+    ExportLaunchProfileExtensionValueProviderScope.LaunchProfile)]
+    [AppliesTo(MyProject.UniqueCapability)]
+    internal sealed class MyLaunchProfileExtensionValueProvider : ILaunchProfileExtensionValueProvider
+    {
+        private const string ExecutablePathPropertyName = "ExecutablePath";
+
+        public string OnGetPropertyValue(string propertyName, ILaunchProfile launchProfile, ImmutableDictionary<string, object> globalSettings, Rule rule)
+        {
+            string propertyValue = null;
```
drewnoakes, line 205: It'd be good to use nullable reference types in these code samples. It's easier to remove the annotations than it is to add them, and most folks will be on a new enough compiler.

```suggestion
            string? propertyValue = null;
```

I'd stick `#nullable enable` at the top of these files too, just to make it work regardless of project-level settings.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+            switch (propertyName)
+            {
+                case ExecutablePathPropertyName:
+                    propertyValue = (string)launchProfile.OtherSettings["executablePath"];
```
drewnoakes, line 209: This should use `TryGetValue`. If the key is not present, the indexer will throw a `KeyNotFoundException` rather than returning `null`.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+
+            return propertyValue ?? string.Empty;
+        }
+
+        public void OnSetPropertyValue(string propertyName, string propertyValue, IWritableLaunchProfile launchProfile, ImmutableDictionary<string, object> globalSettings, Rule rule)
+        {
+            switch (propertyName)
+            {
+                case ExecutablePathPropertyName:
+                    launchProfile.OtherSettings["executablePath"] = propertyValue;
```
drewnoakes, line 221: Ditto.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+
+In order for Visual Studio to locate the rule, you must register the codebase of the assembly embedding the rule from the Visual Studio package via the `ProvideCodeBaseAttribute`.
+
+> Note: this is true *even when other MEF components are successfully loaded without a registered codebase*.
+
+```csharp
+using Microsoft.VisualStudio.Shell;
+
+[assembly: ProvideCodeBase(CodeBase = @"$PackageFolder$\Path\To\MyAssembly.dll")]
+
```
drewnoakes, line 176:
```suggestion
```

drewnoakes (commented):
Thanks very much for these docs @philliphoff. I'm not an expert on this area so would appreciate @tmeschter's input here as well, but this looks pretty much good to go, barring a few comments.

tmeschter:
I'm going to use this as a starting point and clean it up a bit today.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
tmeschter: #7462 _removed_ our use of CodeBase. I'll double-check this.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
```diff
@@ -103,6 +103,125 @@ And you're done. Projects that import the .targets file will now show this UI wh
...
+```
+
+### Step 4: Bind rule properties to launch settings values
```
tmeschter, line 179: I am likely to move this content to a separate HOWTO.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
tmeschter: Most of the code in this type is unnecessary for these purposes. I've updated the steps to eliminate this file and put the rule MEF metadata elsewhere.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
tmeschter: This is still I good idea, but I ended up leaving it off of this how-to for the sake of brevity. We should find a different place for this, though.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
tmeschter: I've ended up removing this section as it really belongs in a different HOW TO.

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
tmeschter: The important part is to get the assembly on the VS binding path because we end up loading the assembly to read the resources. I've replaced this with a brief note on how to do this.

tmeschter (approved):

tmeschter:
@philliphoff @drewnoakes I've made numerous updates and the PR is now ready to go.

drewnoakes (approved):

philliphoff:
Thanks all for fleshing this out and getting it in!

docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
philliphoff: Is it general practice (now) to use `ProvideBindingPathAttribute` for extensions? We use `ProvideCodeBaseAttribute` in all of our extensions, in part to prevent VS from blindly using assemblies it might find in the folder over the "official" versions shipped with VS core, which has caused occasional problems in the past.

Issue #6256, opened 2020-05-29
VS 2019 16.5.0 hangs when opening a project where a compatible .NET Core SDK is not available (with repro)


vsfeedback:
_This issue has a corresponding [ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/954681/vs-2019-1650-hangs-when-opening-a-project-where-a-1.html). Please vote and comment there to make sure your voice is heard._

---
[regression] [worked-in:16.4.6]
I have a project that has a `global.json` file in the root indicating that version `3.1.100` of the .NET Core SDK should be used. A compatible version of that SDK comes with VS 16.4.0. After upgrading VS to 16.5.0 and attempting to open my project, the IDE hangs indefinately. After some trial and error, `dotnet --list-sdks` revealed that no compatible .NET Core SDK was installed for 3.1.1xx. After installing .NET Core SDK 3.1.101, VS 16.5.0 opened the project successfully. This is likely because my machine was using the VS specific version of that SDK and it got upgraded, which I understand.

The issue is that the IDE hangs when opening the project. It would be great if an error indicating the issue could be displayed. A bonus would be to pop the installer to install the missing SDK.

---
#### Roi Chen [MSFT] on 3/18/2020, 10:21 PM: 

<p>Thank you for taking the time to log this issue!<br>
I’ve tried to reproduce and investigate using the description, and attachments already provided. Unfortunately those aren’t enough and more information is needed in order to investigate it further.<br>
The easiest way to provide all the information is to use the Visual Studio Feedback Tool. This will ensure that we collect the needed information for you without worrying about what to provide (recording, dump file or ETL trace).<br>
Since this issue is now marked as Need More Info, that workflow is enabled in the Feedback Tool:<br>
•	Open Visual Studio Feedback tool.<br>
•	Click the banner letting you know that you have problems requesting your attention.<br>
•	Click this problem from the list<br>
•	Click “View their request and respond” from the problem details banner<br>
•	Add a comment, in the Attachments/Record: click Start Recording<br>
•	When the Steps Recorder tool appears, perform the steps that reproduce the problem.<br>
•	When you’re done, choose the Stop Record button.<br>
•	Wait a few minutes for Visual Studio to collect and package the information that you recorded.<br>
•	Submit. You will be able to see the comment on Developer Community. For security reasons, your files come directly to us and don’t appear on Developer Community.<br>
[...]

PR #7403, opened 2021-07-19
Delete references to IDE dispose analyzers

xtqqczze:
The IDE dispose analyzer rules have been deleted from Roslyn:

https://github.com/dotnet/roslyn/commit/eeba499ecf839ec35bca25062d69d2fc5c4885b9

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7403)


xtqqczze:
See also: https://github.com/dotnet/runtime/commit/ca3b27963931172f0535fff644e1fae23f406193

xtqqczze:
The suppressions of `IDE0069` should be replaced with suppressions of [CA2213: Disposable fields should be disposed](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2213)

xtqqczze:
The suppressions of `IDE0068` should be replaced with suppressions of [CA2000: Dispose objects before losing scope](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2200)?

drewnoakes:
@xtqqczze are you happy with this PR?

Do you see any diagnostics in files after removing the suppressions?

MiYanni:
I've updated the branch and removed the suggested suppressions. They only existed for the .editorconfig of the tests. No warnings on build. I leafed through the files to see if those warnings where there, and didn't see them in the files I checked. This PR looks good to go.

MiYanni (approved):

xtqqczze:
My concern is that the `SuppressMessage` attributes were possibly added for a good reason (by @davkean in #6256), therefore we should keep these suppressions, but change the diagnostic identifiers to `CA2000`, `CA2213`, etc.

MiYanni:
@xtqqczze I don't see any code changes related to that issue. I see from the original ticket here (https://developercommunity.visualstudio.com/t/vs-2019-1650-hangs-when-opening-a-project-where-a-1/954681) that the issue has been resolved. But from the internal ticket, there is no link to the actual PR. 

Either way, the suppressions shouldn't affect the product's logic in any way. It would only affect building it, right? I specifically checked this one and there was no difference between the suppression existing versus not existing: `~M:Microsoft.VisualStudio.Editors.PropertyPages.ReferencePropPage.WndProc(System.Windows.Forms.Message@)`. I don't believe that removing these suppressions will cause harm.

drewnoakes:
The only downside to the suppressions would be getting a bunch of diagnostics during build (or in the IDE). If we don't see those, then it's fine to remove them.

drewnoakes:
Let's merge this. It builds cleanly and can be tweaked if needed. Thanks very much @xtqqczze.

xtqqczze:
> @xtqqczze I don't see any code changes related to that issue. I see from the original ticket here (https://developercommunity.visualstudio.com/t/vs-2019-1650-hangs-when-opening-a-project-where-a-1/954681) that the issue has been resolved. But from the internal ticket, there is no link to the actual PR.

I'm not sure how I came to link to that issue, I was thinking of #5019.
#5019

PR #7407, opened 2021-07-20
Support reading and writing properties directly to OtherSettings

### LINKED ISSUE #7384: Launch Profile UI does not persist properties added by custom launch profile
*TL;DR from @tmeschter*: Currently every property you wish to read or write in the Launch Profiles UI needs to be handled by some implementation of `ILaunchProfileExtensionValueProvider`. We should implement a fallback such that any property **not** handled by an `ILaunchProfileExtensionValueProvider` is simply passed through "as-is" and stored in the `ILaunchProfile.OtherSettings` dictionary. This would allow simple scenarios to have a very simple implementation.

*Original issue report continues below.*

**Visual Studio Version**: 17.0.0 Preview 2.0 [31430.505.main] (Dev17.IntPreview)

**Summary**: 
After following the [documentation](https://github.com/dotnet/project-system/blob/main/docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md) on adding a custom launch profile, I found that the Launch Profile UI fails to persist properties defined by the custom launch profile. Oddly, a newly-added property whose name matches that of one of the "standard" properties (`RemoteDebugEnabled`) *does* get persisted to `launchSettings.json`.

**Steps to Reproduce**:
[Repro project](https://github.com/dotnet/project-system/files/6751770/WpfApp1.zip)
New launch profile is defined in `<rootdir>\MyCustomLaunchProfile.xaml` and imported into the project via
```xml
<ItemGroup>
    <PropertyPageSchema Include="..\MyCustomLaunchProfile.xaml">
        <Context>Project</Context>
    </PropertyPageSchema>
</ItemGroup>
```
A `launchSettings.json` with a predefined profile using the newly defined launch command (`MyCustomLaunchProfile`) is included for convenience.

1. Open the Launch Profiles UI
2. (only needed if included `launchSettings.json` is deleted) Create a new profile based on "My Custom Launch Profile"
3. Change the value of each of the three listed properties (either by clicking on the checkbox or entering text into the TextBox).
4. Close and reopen the Launch Profiles UI

**Expected Behavior**:
Both checkboxes should be checked, and text entered into "Test String Property" should be preserved. Opening `launchSettings.json` will show values set on the three properties.

**Actual Behavior**:
The "Test bool property" checkbox will never be checked (if you click on it quickly enough you might notice a checkmark flicker in and out of existence, almost as if something is clobbering the checkbox's state). The text in the "Test string property" TextBox will be cleared as soon as the TextBox loses focus. Only the `RemoteDebugEnabled` property will have a value set in `launchSettings.json`.

**User Impact**:
This makes it impossible for a project system to add a functional custom launch profile.

tmeschter:
Fixes #7384.

Right now, if you want to create a .xaml file for a particular kind of launch profile you also need to create an implementation of `ILaunchProfileExtensionValueProvider` and/or `IGlobalSettingExtensionValueProvider` to handle reading and writing the property value. Most developers outside of the dotnet/project-system team who need to do this will just store their properties as-is in the `ILaunchProfiles.OtherSettings` dictionary; writing an extension value provider to do this isn't hard but does involve some amount of boilerplate.

Instead, in the absence of an extension value provider for a particular property we should just default reading and writing it to `ILaunchProfile.OtherSettings`. We do need to know the storage type of the property, but this can be determined based on the type information already in the .xaml `Rule` file.

There are a number of further improvements we could make, such as using default values from the `Rule` in cases where the property is being read but no value is found in `OtherSettings` or as a backup when attempting to set the value to an invalid value. However, this should be enough to unblock our partners.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7407)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
@@ -162,6 +162,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
                 }
             }
 
+            if (profile.OtherSettings is not null)
+            {
+                foreach ((string propertyName, _) in profile.OtherSettings)
+                {
+                    builder.Add(propertyName);
+                }
+            }
+
             return builder.ToImmutable();
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
@@ -240,13 +240,39 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
                 return;
             }
 
-            // Finally, check if a global setting extender can handle it.
+            // Then, check if a global setting extender can handle it.
 
             Func<ImmutableDictionary<string, object>, ImmutableDictionary<string, object?>>? globalSettingsUpdateFunction = GetPropertyValueSetterFromGlobalExtenders(propertyName, unevaluatedPropertyValue);
             if (globalSettingsUpdateFunction is not null)
             {
                 await _launchSettingsProvider.UpdateGlobalSettingsAsync(globalSettingsUpdateFunction);
                 return;
+            }
+
+            // Finally, store it in ILaunchProfile.OtherSettings.
+
+            object? valueObject = null;
+            if (_rule?.GetProperty(propertyName) is BaseProperty property)
+            {
+                valueObject = property switch
+                {
+                    BoolProperty => bool.TryParse(unevaluatedPropertyValue, out bool result) ? result : null,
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
```diff
@@ -240,6 +249,32 @@ public async Task SetPropertyValueAsync(string propertyName, string unevaluatedP
...
+                };
+            }
+            else
+            {
+                valueObject = unevaluatedPropertyValue;
+            }
+
+            if (valueObject is not null)
+            {
+                profileUpdateAction = p => p.OtherSettings[propertyName] = valueObject;
```
MiYanni, line 274: I'm assuming capturing `valueObject` in the closure is fine?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
tmeschter: Yeah, that should be fine.

Issue #7411, opened 2021-07-20
Don't hardcode dynamic file item type


davidwengier:
At the moment the DynamicItemHandler [hardcodes the item type](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/DynamicItemHandler.cs#L31) for Razor files which means there is no way to opt into the Razor editor lighting up, without also affecting build/publish/packaging. For large enterprise solutions (Bing, OrchardCore, etc.) where the Web SDK is not in use this means the Razor editor doesn't work correctly.

The Razor tooling team is wanting to enable this item type to come from a property that is read at evaluation so that in the project system design time targets this can be set to `None` and in the Web SDK (or possibly the Razor tooling targets) it can be set to `Content` to preserve the existing behaviour.

/cc @tmeschter @NTaylorMullen

davidwengier:
Spoke with @drewnoakes and using a customizable item type suffers from the chicken and egg problem: We need evaluation to determine which item types to read, but the item types need to be read from evaluation.

Instead I'm going to try creating a specific item type for the Razor editor, which the design time targets will add razor files to. This means the design time targets will match the `DynamicItemHandler` and means the project system will fully define both sides of the editor hookup, rather than at the moment when the Web SDK has the file glob.

PR #7417, opened 2021-07-22
Allow Razor files to appear in any source item type

davidwengier:
Fixes https://github.com/dotnet/project-system/issues/7411
Fixes https://github.com/dotnet/aspnetcore/issues/34188

This changes the DynamicItemHandler for Razor so that instead of being hardcoded to the `Content` item type, it now uses CPSes SourceItemsService (via an ISourceItemsHandler) so that it can see all items in any type that is considered a source file. Essentially my understanding is that this will now see any file that is in the project tree. This solves various issues from enterprise solutions:

* OrchardCore includes Razor files in the `EmbeddedResource` item type
* Bing has Razor files in the `None` item type (ie, the default from the SDK) and uses custom build targets and tasks to process them at build time
* MAUI uses a custom item type added by `AvailableItemName`

I'm still doing manual validation, but want to get a real build so may as well put this up too.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7417)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/DynamicItemHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/DynamicItemHandler.cs
@@ -36,6 +36,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
 
             foreach (var (_, projectChange) in projectChanges)
             {
+                if (!projectChange.Difference.AnyChanges)
+                    continue;
+
                 IProjectChangeDiff difference = HandlerServices.NormalizeRenames(projectChange.Difference);
 
                 foreach (string includePath in difference.RemovedItems)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/DynamicItemHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/DynamicItemHandler.cs
@@ -13,7 +13,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
     ///     Handles changes to dynamic items, such as Razor CSHTML files.
     /// </summary>
     [Export(typeof(IWorkspaceContextHandler))]
-    internal class DynamicItemHandler : AbstractWorkspaceContextHandler, IProjectEvaluationHandler
+    internal class DynamicItemHandler : AbstractWorkspaceContextHandler, ISourceItemsHandler
     {
         private const string RazorPagesExtension = ".cshtml";
         private const string RazorComponentsExtension = ".razor";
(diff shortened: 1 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/DynamicItemHandler.cs
```diff
@@ -26,48 +26,49 @@ public DynamicItemHandler(UnconfiguredProject project)
             _project = project;
         }
 
-        public string ProjectEvaluationRule
-        {
-            get { return Content.SchemaName; }
-        }
-
-        public void Handle(IComparable version, IProjectChangeDescription projectChange, ContextState state, IProjectDiagnosticOutputService logger)
+        public void Handle(IComparable version, IImmutableDictionary<string, IProjectChangeDescription> projectChanges, ContextState state, IProjectDiagnosticOutputService logger)
```
davidwengier, line 29: Recommend reviewing this file with whitespace turned off, otherwise GitHub overplays the changes.

drewnoakes:
Looks good.

As discussed separately, this may add Razor files to intellisense which are not present in the build.

It seems like it would be safer to have a property that controls which item types are considered dynamic. This could be added here I think, with a small tweak to your new ISourceItemHandler interface to also take ConfigurationGeneral, sync-linked to the items it receives today, then pulling from the required item type(s) alone.

davidwengier:
It works!
image

(Top is VS 2022 normal, bottom is with my changes)

davidwengier:
And verified in Project System Tools when dumping the Roslyn workspace that all of the right files are there. I think this is good to go from my point of view.

> this may add Razor files to intellisense which are not present in the build

For the Razor editor this is actually a good thing. Firstly it matches the experience of using the old editor, which is the main concern we have (we don't want users to notice that anything is different in VS 2022 with the new editor), and its also something I think we are happy with, with respect to lighting up the editor in any circumstances. If that changes in future we can always update.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -103,7 +106,18 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
...
+                                OnProjectChangedAsync(e, WorkspaceContextHandlerType.ProjectBuild),
+                                _project.UnconfiguredProject,
+                                ProjectFaultSeverity.LimitedFunctionality),
+                            linkOptions: DataflowOption.PropagateCompletion,
+                            cancellationToken: cancellationToken),
+
+                    ProjectDataSources.SyncLinkTo(
+                        _activeConfiguredProjectProvider.ActiveConfiguredProjectBlock.SyncLinkOptions(),
+                        _projectSubscriptionService.SourceItemsRuleSource.SourceBlock.SyncLinkOptions(),
+                        _projectBuildSnapshotService.SourceBlock.SyncLinkOptions(),
```
davkean, line 118: This is wrong it should be not sync'd with the build snapshot, otherwise, the results wait until the design time build has finished. I know that evaluation does this but this is a regression that needs to be fixed, and might explain why things are ending up in misc on first open.

davkean: I've opened https://github.com/dotnet/project-system/issues/7423 for evaluation.

davidwengier: Thanks. This is so annoying to change 😛

We need generic generics so I can define a method that takes a tuple of any size as long as the first item is of a specific type :)

davidwengier:
Not sure how you'll feel about the last couple of commits here.. but I thought it beat duplicating all those methods to handle a 2-tuple. If a larger refactoring is desired, you can always do it when you fix the evaluation bit 😛

davidwengier:
@dotnet/project-system can I get reviews on this? would _love_ to get this in to preview 3 so Bing can test it etc.

drewnoakes (approved):

davidwengier:
Thanks Drew!

PR #7418, opened 2021-07-22
Support multiple Hot Reload sessions

tmeschter:
1. Extract the logic for creating and tracking Hot Reload sessions to a new type, `IProjectHotReloadSessionManager`. The `ProjectLaunchTargetsProvider`'s role in Hot Reload to now reduced to requesting that a pending session be created, and informing the manager of when the pending session should be activated. Inside the manager, `HotReloadState` serves as to associate a session with a `Process` and also implements `IProjectHotReloadSessionCallback`. In general it redirects all operations to the manager for actual handling, though.
2. Support multiple sessions. Currently, If a Hot Reload session is active for a project and the user runs the project again with Ctrl+F5 we kill the previous process and session, and then run the new one. This is deviation from the previous (non-Hot-Reload) behavior where running with Ctrl+F5 multiple times will start multiple processes. Bringing support for this to Hot Reload is mostly a matter of never killing a process and maintaining a set of all the active sessions/processes.

In future the manager will likely have more responsibility. For example, if the user makes an edit that cannot be applied we would like to automate the process of killing the processes, building, and restarting.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7418)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
@@ -3,13 +3,10 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Buffers.PooledObjects;
-using Microsoft.VisualStudio.HotReload.Components.DeltaApplier;
 using Microsoft.VisualStudio.IO;
 using Microsoft.VisualStudio.ProjectSystem.Debug;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
@@ -17,7 +14,6 @@ using Microsoft.VisualStudio.ProjectSystem.Utilities;
 using Microsoft.VisualStudio.ProjectSystem.VS.HotReload;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.Threading;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Task = System.Threading.Tasks.Task;
@@ -36,8 +32,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         IDebugProfileLaunchTargetsProvider,
         IDebugProfileLaunchTargetsProvider2,
         IDebugProfileLaunchTargetsProvider3,
-        IDebugProfileLaunchTargetsProvider4,
-        IProjectHotReloadSessionCallback
+        IDebugProfileLaunchTargetsProvider4
     {
         private static readonly char[] s_escapedChars = new[] { '^', '<', '>', '&' };
         private readonly ConfiguredProject _project;
@@ -50,11 +45,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
         private readonly IProjectThreadingService _threadingService;
         private readonly IVsUIService<IVsDebugger10> _debugger;
         private readonly IRemoteDebuggerAuthenticationService _remoteDebuggerAuthenticationService;
-        private readonly IProjectHotReloadAgent _projectHotReloadAgent;
-        private readonly Lazy<IHotReloadDiagnosticOutputService> _hotReloadDiagnosticOutputService;
-
-        private HotReloadState? _pendingHotReloadSession;
-        private HotReloadState? _activeHotReloadSession;
+        private readonly Lazy<IProjectHotReloadSessionManager> _hotReloadSessionManager;
 
         [ImportingConstructor]
         public ProjectLaunchTargetsProvider(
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -121,16 +104,7 @@ public Task OnAfterLaunchAsync(DebugLaunchOptions launchOptions, ILaunchProfile
 
         public async Task OnAfterLaunchAsync(DebugLaunchOptions launchOptions, ILaunchProfile profile, IReadOnlyList<VsDebugTargetProcessInfo> processInfos)
         {
-            await TaskScheduler.Default;
```
drewnoakes, line 124: Does it make sense to still switch to a background thread here? I assume this is called on the UI thread, and it looks like `ActivateSessionAsync` does some non-trivial work.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -0,0 +1,233 @@
...
+                targetFrameworkVersion = targetFrameworkVersion.Substring(startIndex: 1);
+            }
+
+            return targetFrameworkVersion;
+        }
+
+        private async Task ProcessExitedAsync(HotReloadState hotReloadState)
+        {
+            Assumes.NotNull(hotReloadState.Session);
+            Assumes.NotNull(hotReloadState.Process);
```
drewnoakes, line 148: It might be hard to know which of this caused an exception, if we don't have line numbers. Perhaps use `Requires.NotNull`, passing the property name instead of the parameter name. Ditto below in `StopProjectAsync`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -0,0 +1,233 @@
...
+            public HotReloadState(ProjectHotReloadSessionManager sessionManager)
+            {
+                _sessionManager = sessionManager;
+            }
+
+#pragma warning disable VSTHRD100 // Avoid async void methods
+            internal async void OnProcessExited(object sender, EventArgs e)
+#pragma warning restore VSTHRD100 // Avoid async void methods
+            {
+                await _sessionManager.ProcessExitedAsync(this);
```
drewnoakes, line 209: Why not `JTF.RunAsync` here, to avoid `async void`?

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
tmeschter: Yeah, that could work, too.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            if (_pendingSessionState is not null)
+            {
+                Assumes.NotNull(_pendingSessionState.Session);
+
+                try
+                {
+                    Process? process = Process.GetProcessById(processId);
+
+                    await _hotReloadDiagnosticOutputService.Value.WriteLineAsync($"{_pendingSessionState.Session.Name}: Attaching to process '{processId}'.");
```
MiYanni, line 55: This diagnostic output is only for our (Microsoft) purposes, yes? Not really for the customer?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
tmeschter: Ah, yes. That was an accidental removal.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
tmeschter: If we've got symbols, we should have line numbers.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
tmeschter: It is primarily for development purposes at this point. What we want to display to the customer is TBD and also covers a bunch of Hot Reload components we don't own.

It all goes to the "Hot Reload" pane of the Output Window.

MiYanni: Ah, okay. So... since it is a 'log', I'm assuming we don't do translations on log information?

tmeschter: If these end up being hidden behind a flag or reg key or some such we don't need to translate them. If they are on by default (e.g. the current behavior) we will need to translate them because they are still user-facing. I'll probably pull them out into string resources because there's no real reason not to.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
tmeschter: I ended up using `IProjectFaultHandlerService.Forget` to run the async code but still capture exceptions.

PR #7419, opened 2021-07-22
Send telemetry on a thread pool thread

### LINKED ISSUE #7412: Solution Close Perf: DependencyTreeTelemetryService.Dispose sends telemetry on the UI thread
The MEF component was disposed on the UI thread, and blocks it to send elemetry during solution.close time.

Can we just do it in a thread pool thread?

It blocks UI thread for 80ms when OrchardCore.sln is closed.

drewnoakes:
Fixes #7412 

During shutdown we compute and publish telemetry about project dependencies. Previously this code would run on the UI thread, although that is not necessary.

This change uses the `OnceInitializedOnceDisposed` base class which uses async disposal by default, running that work on the thread pool.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7419)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
@@ -24,7 +24,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
     /// </remarks>
     [Export(typeof(IDependencyTreeTelemetryService))]
     [AppliesTo(ProjectCapability.DependenciesTree)]
-    internal sealed class DependencyTreeTelemetryService : IDependencyTreeTelemetryService, IDisposable
+    internal sealed class DependencyTreeTelemetryService : OnceInitializedOnceDisposed, IDependencyTreeTelemetryService
     {
         private const int MaxEventCount = 10;
 
@@ -58,6 +58,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.Dependencies
                 _telemetryService = telemetryService;
                 _stateByFramework = new Dictionary<TargetFramework, TelemetryState>();
             }
+        }
+
+        protected override void Initialize()
+        {
         }
 
         public void InitializeTargetFrameworkRules(ImmutableArray<TargetFramework> targetFrameworks, IReadOnlyCollection<string> rules)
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
```diff
@@ -170,6 +174,17 @@ public void ObserveSnapshot(DependenciesSnapshot dependenciesSnapshot)
         public void Dispose()
         {
             if (_telemetryService != null && _dependenciesSnapshot != null && _projectId != null)
+            {
+                _joinableTaskContext.Factory.Run(async () =>
```
davkean, line 178: See: [Avoid calling JTF.Run on the UI thread](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24027/Threading?anchor=%E2%9D%8C-**avoid**-calling-joinabletaskfactory.run-on-the-ui-thread).

This does not free up the UI thread, this is basically Task.Wait.

This actually does the reverse of what you are hoping for as increases the likelihood of this taking longer because now the UI thread needs to wait for a thread pool worker to service the request, which is about 1% of all delays in 16.9.

Instead, this class should derive from OnceInitializedOnceDisposeAsync and implement its logic in there.

drewnoakes: Converted to use `OnceInitializedOnceDisposed`. One concern here is that by not blocking, the MEF parts that this logic requires (e.g. `ITelemetryService`) may be disposed by the time the async cleanup code wants to use them. Do you think that is a concern now, with the change you suggest?

davkean: You had the same problem before or after this change; there is no order to MEF disposing. The underlying service is not disposable, so its okay now, but might be a problem if you ever made it disposable.

davkean (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/DependencyTreeTelemetryService.cs
drewnoakes: TIL. In an email thread with Lifeng and Andrew it became clear that scopes are disposed in order, but the components in each scope are essentially unordered. The actual order appears to be FIFO, which is the near-opposite of what I thought MEF did (dispose in reverse dependency order where possible).

davkean: If you want code to be guaranteed before project close, then you need to register it; http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem&leftSymbol=xahwgqrmhthr&file=Core%5CThreading%5CIProjectAsynchronousTasksService.cs&line=77. Note that Lifeng recently made a change to only run tasks critical to actually saving the project when the user closes Visual Studio. Telemetry is not critical to that.

davkean: LoadedProjectAsync extension method is basically sugar over that method and it passes ProjectOperation.Unload.

MiYanni (approved):

Issue #7376, opened 2021-07-01
Package Icon property needs to add item to project with metadata (3)


drewnoakes:
When an icon is added to a package, the icon file must be shipped inside the package itself.

In order to do this, the icon must be added to the project as an item, as well as being specified as a property. That item must have metadata to include it in the package, and to specify the path within the package.

The legacy project properties handles this correctly. We should reproduce that behaviour.

drewnoakes:
This is very similar to #6457, where the README file is included in the package. The legacy pages do not support README files in the package, but the same approach applies to both properties so the delta between fixing one versus two properties here is likely very small.

drewnoakes:
Notes to triage team: We added this as `must-have` on the assumption that it is similar to #7167, which is also `must-have`.

MiYanni:
Prior experience:
![image](https://user-images.githubusercontent.com/17788297/125527606-67e72ab6-47c1-4e4c-8ff4-728c13c27aa2.png)

Produces:
```xml
  <PropertyGroup>
    <PackageIcon>shell32_192.ico</PackageIcon>
    <PackageIconUrl />
  </PropertyGroup>

  <ItemGroup>
    <None Include="..\..\..\..\Code\project-system\tests\Microsoft.VisualStudio.Editors.UnitTests\shell32_192.ico">
      <Pack>True</Pack>
      <PackagePath></PackagePath>
    </None>
  </ItemGroup>
```

Issue #7421, opened 2021-07-22
New Property Pages: Property delay on update


MiYanni:
The information below is from testing I did related to the Package Icon property changes I did. Basically, the value gets filled in as an absolute path after initially selecting a value in the Browse dialog. Then, there is a delay between when that value is updated to be the proper relative path that is returned by the value interceptor.

> So, I'm not sure what to make of this. I never provide a fullpath to any of the value interceptors. However, it automatically fills in a fullpath. I'm assuming this is a feature of the subtype 'file' control. Then, there is a delay (on my very fast machine) of 3-6 seconds for it to update. If I run in debug, I've seen it go as long as 8-10 seconds before it shows the proper value. I'm assuming the delays would be even longer on a slower machine... or is there some kind of update timer? (GIF 1)
> 
> I tested manually putting in a path (not using Browse), making the edit field lose focus (click somewhere else), and the delay still occurs. It also makes me realize that I don't have code to validate if the path provided is valid.
> 
> Tested this on my work PC (which is slower than my home PC) and it updates faster(?) But the delay is definitely noticeable in certain situations. I explicitly show the .csproj and solution explorer in the GIF here to show them update nearly instantly, while the edit field is delayed in comparison. (GIF 2)

GIF 1:
![DelayedPropertyUpdate](https://user-images.githubusercontent.com/17788297/126710582-351fbdd9-cc70-47b3-a316-b33ed6cc740d.gif)

GIF 2:
![PackageIconUpdateDelayLess](https://user-images.githubusercontent.com/17788297/126710630-0f6519a3-bf9d-4ea9-a548-7a1692992025.gif)

drewnoakes:
@MiYanni is this still an issue? Can we close this?

MiYanni:
@drewnoakes Without retesting it, there is no way to know if it still an issue or not. AFAIK, there hasn't been any code change to directly address this behavior.

drewnoakes:
Could you test again to verify?

MiYanni:
@drewnoakes Just spent 10 minutes testing it. I couldn't reproduce this exact problem. However several times, the application (VS itself) would freeze for like 5 seconds (Window's style freeze, where you cannot interact with the app and the title bar flips theme). So, I'd say this specific issue is no longer an issue. I'll close it.

drewnoakes:
If you can repro a freeze like that, use the "Report a problem" feature to capture a trace and file it.

Issue #7420, opened 2021-07-22
New Property Pages: Add ability to specify extension information to Browse dialog (subtype "file") properties


MiYanni:
In the old property pages, the Browse dialog (file picker) allows you to specify file extension filters. These could be one or more custom named filters that appear in a dropdown near the Open/Select/Cancel buttons. For example, the old Package Icon property has a single file extension filter named "Icons" that has the value of `*.jpg, *.jpeg, *.png`.
![image](https://user-images.githubusercontent.com/17788297/126707722-c86524ec-6d02-48d5-aeab-9da81832e4ca.png)

This should be allowed for the new property pages as the same file extensions in the old property pages would apply to the same properties in the new pages.

PR #7422, opened 2021-07-22
Package Icon property fix

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7376
Related: https://github.com/dotnet/project-system/issues/7420
Related: https://github.com/dotnet/project-system/issues/7421

Currently, the package icon property was set up as a generic string, which won't work for the purpose of this property. This property has 2 values to it: the value stored in the property itself `<PackageIcon>` and the `Include` value stored in an associated `<None>` item. The former is a relative path to the icon file *inside the package*. The latter is the relative path (from the project) to the file on disk. There are additionally 2 metadata elements associated with the `<None>` item: `<Pack>` and `<PackagePath>`. The former is a boolean that states if the item should be packed into the package. The latter gives a folder path to the destination inside the package that the item should be stored.

In the old property pages, updating the property would convert the absolute file path provided into a relative path from the project file location. This causes the .csproj to write a new `<None>` item if no related item existed previously. If there was a previous `<None>` item associated with the package icon, it will update the `Include` value with the relative path to the file on disk. Then, it puts the filename into the `<PackageIcon>` property.

In this new property page implementation, the process is very similar. You select the absolute file path and it becomes converted to a relative file path. The delay seen in the GIF below is an issue, which I've created here: https://github.com/dotnet/project-system/issues/7420. Then, the .csproj is updated with a new `<None>` item is it previously did not exist, or update the `Include` if one previously existed. I match the `<None>` item with the `<PackageIcon>` by filename and if the item contains `<Pack>True</Pack>`. Additionally, it is now correctly recognizing `<PackagePath>` which the old property pages did not honor. This means a filename of `Icon.png` with a package path of `content` correctly becomes `content\Icon.png` in the `<PackageIcon>` property. Therefore, `<PackageIcon>` is updated to be the package path's value + the icon's filename. This is the same logic used in the .nuspec for which these properties are based.

### Related Docs
- https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageicon
- https://docs.microsoft.com/en-us/nuget/reference/nuspec#icon
- https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target
- https://github.com/NuGet/Home/wiki/Packaging-Icon-within-the-nupkg

### Old Property Pages GIF
PackageIconExample-Old

### New Property Pages GIF
PackageIconExample-New

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7422)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -91,16 +91,17 @@
...
 
-  <!-- TODO (tomescht): 
-       We should re-think this. This property is the relative path to the icon file _within the package_; setting it does not guarantee
-       that the corresponding file in the project is correctly copied into the package. A better alternative may be to expose a property
-       with the path to the file in the project instead, and update this one appropriately when that one changes.
-  -->
   <StringProperty Name="PackageIcon"
-                  DisplayName="Icon file"
-                  Description="The relative path from the package root to the icon file. In addition to setting this property you need to make sure that the file is included in the package. Image file size is limited to 1 MB. Supported file formats include JPEG and PNG. An image resolution of 128x128 is recommended."
+                  DisplayName="Icon"
```
MiYanni, line 95: Updated display name to match that of the similar application icon in the Application property page. Now, they both say `Icon` but are in their appropriate property page. I believe 'file' was there because there used to be an Icon URL, but that functionality is deprecated.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+            _sourceItemsProvider = sourceItemsProvider;
+            _unconfiguredProject = unconfiguredProject;
+        }
+
+        // https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageicon
+        private static string CreatePackageIcon(string filePath, string? packagePath)
+        {
+            string filename = Path.GetFileName(filePath);
+            // Make a slash-only value into empty string, so it won't get prepended onto the path.
+            packagePath = ("\\".Equals(packagePath, StringComparison.Ordinal) || "/".Equals(packagePath, StringComparison.Ordinal))
```
tmeschter, line 37: Check if there is an appropriate comparer in the `StringComparisons` type.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+                ? string.Empty : (packagePath ?? string.Empty);
+            // The assumption is that PackagePath does not contain a path to a file; only a directory path.
+            return Path.Combine(packagePath, filename);
+        }
+
+        private async Task<IProjectItem?> GetExistingNoneItemAsync(string existingPackageIcon)
+        {
+            foreach (IProjectItem noneItem in await _sourceItemsProvider.GetItemsAsync(NoneItem))
```
tmeschter, line 45: You may want to test and see what happens if the matching `None` item is found in a Directory.Build.props or other imported file.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                string pack = await noneItem.Metadata.GetEvaluatedPropertyValueAsync(PackPropertyName);
+                string itemIconFilename = Path.GetFileName(noneItem.EvaluatedInclude);
+                string existingIconFilename = Path.GetFileName(existingPackageIcon);
+                // Instead of doing pure equality between itemPackageIcon and existingPackageIcon, a user may update the PackagePath
+                // of the item and forget to update the PackageIcon to reflect those changes, or vice versa. Instead, if the filename
+                // of this packed None item and the filename of the PackageIcon match, consider those to be related to one another.
+                if (TrueValue.Equals(pack, StringComparison.OrdinalIgnoreCase) && itemIconFilename.Equals(existingIconFilename, StringComparison.Ordinal))
```
tmeschter, line 53: Again, check `StringComparisons` for an appropriate comparison.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+            }
+
+            return null;
+        }
+
+        public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+        {
+            string relativePath = PathHelper.MakeRelative(_unconfiguredProject, unevaluatedPropertyValue);
+            string existingPackageIcon = await defaultProperties.GetEvaluatedPropertyValueAsync(PackageIconPropertyName);
```
tmeschter, line 65: Is it possible for `existingPackageIcon` to be the empty string? What happens if it is?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+                {
+                    return itemPackageIcon;
+                }
+
+                await existingItem.SetUnevaluatedIncludeAsync(relativePath);
+                packagePath = itemPackagePath;
+            }
+            else
+            {
+                await _sourceItemsProvider.AddAsync(NoneItem, relativePath, new Dictionary<string, string> { { PackPropertyName, "True" }, { PackagePathPropertyName, packagePath } });
```
tmeschter, line 83: `TrueValue` instead of `"True"`?

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
+{
+    [ExportInterceptingPropertyValueProvider(PackageIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class PackageIconValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string PackageIconPropertyName = "PackageIcon";
+        private const string PackPropertyName = "Pack";
+        private const string PackagePathPropertyName = "PackagePath";
+        private const string NoneItem = "None";
+        private const string TrueValue = "True";
```
drewnoakes, line 18: You can use `bool.TrueString` and `bool.FalseString` rather than define your own constant.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
+{
+    [ExportInterceptingPropertyValueProvider(PackageIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class PackageIconValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string PackageIconPropertyName = "PackageIcon";
+        private const string PackPropertyName = "Pack";
+        private const string PackagePathPropertyName = "PackagePath";
+        private const string NoneItem = "None";
```
drewnoakes, line 17: When referring to an item's name in code, and when we have a corresponding rule file, you can use the rule to define the constant. For example in this case you'd use `None.SchemaName`.

This can make it easier to find places in code that are using a given item type. Also, while it's not likely we'd remove the `None` item's rule, it will cause compile errors when a rule is removed that would would otherwise need.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
+{
+    [ExportInterceptingPropertyValueProvider(PackageIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class PackageIconValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string PackageIconPropertyName = "PackageIcon";
+        private const string PackPropertyName = "Pack";
+        private const string PackagePathPropertyName = "PackagePath";
```
drewnoakes, line 16: These fields are "properties" of items. In MSBuild the term "property" has a special meaning, so in order to avoid confusion we refer to these as "metadata".

```suggestion
        private const string PackMetadataName = "Pack";
        private const string PackagePathMetadataName = "PackagePath";
```

Confusingly when we define a XAML rule for an MSBuild item, the metadata there is referred to as properties, but that's due to using rule files for more than one kind of entity. In general, using the term "metadata" for properties of items can avoid confusion.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
+{
+    [ExportInterceptingPropertyValueProvider(PackageIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class PackageIconValueProvider : InterceptingPropertyValueProviderBase
```
drewnoakes, line 12: It would be helpful for future generations if we added a comment explaining the intended final state of the project.

I wouldn't worry about making it an XML doc. Just a multi-line `//` comment within the class showing the MSBuild property and items at play here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        // https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageicon
+        private static string CreatePackageIcon(string filePath, string? packagePath)
+        {
+            string filename = Path.GetFileName(filePath);
+            // Make a slash-only value into empty string, so it won't get prepended onto the path.
+            packagePath = ("\\".Equals(packagePath, StringComparison.Ordinal) || "/".Equals(packagePath, StringComparison.Ordinal))
+                ? string.Empty : (packagePath ?? string.Empty);
+            // The assumption is that PackagePath does not contain a path to a file; only a directory path.
```
drewnoakes, line 39: Is this referring to the parameter or some other concept?

```suggestion
            // The assumption is that packagePath does not contain a path to a file; only a directory path.
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+            _unconfiguredProject = unconfiguredProject;
+        }
+
+        // https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageicon
+        private static string CreatePackageIcon(string filePath, string? packagePath)
+        {
+            string filename = Path.GetFileName(filePath);
+            // Make a slash-only value into empty string, so it won't get prepended onto the path.
+            packagePath = ("\\".Equals(packagePath, StringComparison.Ordinal) || "/".Equals(packagePath, StringComparison.Ordinal))
+                ? string.Empty : (packagePath ?? string.Empty);
```
drewnoakes, line 38: This took me a while to parse. Would this be clearer?

```suggestion
            if (packagePath is null or "\\" or "/")
            {
                packagePath = "";
            }
```
While the comparison isn't explicit, that doesn't matter for these particular literals.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
drewnoakes, line 53: Minor, but consider parsing rather than comparing boolean string values:

```suggestion
                if (bool.TryParse(pack, out bool packBool) && packBool  && itemIconFilename.Equals(existingIconFilename, StringComparison.Ordinal))
```

The parse function will handle any leading/trailing white space correctly. That is, `" true "` is still true.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+            return Path.Combine(packagePath, filename);
+        }
+
+        private async Task<IProjectItem?> GetExistingNoneItemAsync(string existingPackageIcon)
+        {
+            foreach (IProjectItem noneItem in await _sourceItemsProvider.GetItemsAsync(NoneItem))
+            {
+                string pack = await noneItem.Metadata.GetEvaluatedPropertyValueAsync(PackPropertyName);
+                string itemIconFilename = Path.GetFileName(noneItem.EvaluatedInclude);
+                string existingIconFilename = Path.GetFileName(existingPackageIcon);
```
drewnoakes, line 49: These calls will allocate strings for every none item in the project. If you check the `pack` value first, you can avoid those allocations for all items that do not have `Pack="true"`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+            string packagePath = string.Empty;
+            if (existingItem != null)
+            {
+                string itemPackagePath = await existingItem.Metadata.GetEvaluatedPropertyValueAsync(PackagePathPropertyName);
+                string itemPackageIcon = CreatePackageIcon(existingItem.EvaluatedInclude, itemPackagePath);
+                // The new filepath is the same as the current. No item changes are required.
+                if (relativePath.Equals(existingItem.EvaluatedInclude, StringComparison.Ordinal))
+                {
+                    return itemPackageIcon;
+                }
```
drewnoakes, line 76: Work can be avoided here in some cases:

```suggestion
                // The new filepath is the same as the current. No item changes are required.
                if (relativePath.Equals(existingItem.EvaluatedInclude, StringComparison.Ordinal))
                {
                    return CreatePackageIcon(existingItem.EvaluatedInclude, itemPackagePath);
                }
```

drewnoakes (approved):
Looks good! I left some performance and clarity feedback for you to consider.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: It gets passed into `GetExistingNoneItemAsync`. That method uses it via `Path.GetFilename`. Just checked in .NET Fiddle, and `System.IO.Path.GetFileName(string.Empty) == string.Empty` (they are equal). So, if `existingPackageIcon` is an empty string, the logic will end up using it as an empty string (which is an acceptable path). The first time you add a package icon, it will be an empty string. Thus, `existingItem` (in the next line after this) will be `null`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Fixed. I didn't have the constant at first and forgot to update usage of it here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
```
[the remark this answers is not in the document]
MiYanni: A majority of the files in this folder are `Microsoft.VisualStudio.ProjectSystem.Properties.Package`. The ones in red here are using that namespace. Only the 2 blue files use `Microsoft.VisualStudio.ProjectSystem.Properties`. Seems like we should do some spring cleaning and make sure our namespaces are what we want them to be. Personally, I love when the namespace matches the folder structure, so I wouldn't like either of these namespaces. 😝
image

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Holy moly. Took me a minute to understand what you were talking about. I thought you mean the rule file related to the Package property page (`PackagePropertyPage.xaml`). Instead, you meant that MSBuild items have a rule file associated with them. Thus, the `None.xaml` file. And, you can access the name (`<Rule Name="None" />`) of that file via `None.SchemaName`, which is what the rule files generate as part of compilation. Got it. Fixed.

Prior to this comment, I didn't know these rule files per item existed.
image

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Got it. I'm not sure why I wrote property here. I was thinking the generic xml-terms of `Element` or `Node`, but I used neither of those. My bad. Fixed. (And I updated the PR description accordingly)

Question: Does `Metadata` refer to it, independent if it is an xml `Element` or an `Attribute`? I know that for MSBuild, it generally treats both elements and attributes as the same. But, many xml-based languages differentiate between the two.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Done. Added a description similar to a shorter version of the description of this PR, and included example XML output.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Extracted the or-expression into a method to make it more explicit and easier to read left-to-right. Also, added to the comment above this line.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: Makes sense. I'm used to there being no direct relationship between the value stored in a data format language and a CLR type. So, I'd normally not use a CLR parser to evaluate a value stored in an XML file. But, I understand how it likely makes sense here since it is MSBuild information. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: I had that at first. After a few iterations, I realized I didn't need to have 2 consecutive if-statements. This is done strictly for human readability, because you can drop these method calls in-place in the predicate for the if-statement, and they'll only get executed as-needed because of the short-circuiting of the && operation. But if that's what is desired, I'll do that. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: This was also a product of refactoring. Originally, `itemPackageIcon` was used as part of the logic after this if-statement. Thus, it was used in 2 places. Now, it is not. Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
MiYanni: This was a great suggestion. I tested this an the results were... bizarre. The `SetUnevaluatedIncludeAsync` call doesn't function if the None item is not inside the project file. This would cause a desync between the in-memory representation of the project file and the one on-disk. You'd even see the Solution Explorer updating the included files, but there would be no new None item in the project file. After a couple iterations, I got the functionality working as it should.

Basically, I still need to locate the None item, as logic related to `PackagePath` uses that information. So, I only check if the existing None item is within the project file right before either updating the existing item or creating a new one. This way, the PackagePath information is preserved. Now, you might say, "Well, if I add the file that already exists in a `.props` file, what happens?" This is the magic. It will call `SetUnevaluatedIncludeAsync` and set it to the same Include path as that in the `.props` file's None item. Then, deduplication occurs on the `.csproj`, thus removing the None item in the `.csproj`! It is really clean, honestly. I've made a GIF to show you what I mean. So, issues related to this are now fixed.

PackageIconExternalFileExample

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
drewnoakes: Yes it's metadata regardless of whether it's expressed as an attribute or child element in the XML representation.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
drewnoakes: In general we should avoid allocating memory for no reason.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
drewnoakes: I still find this hard to read. And now we have an extra method, which makes it harder to reason about the type's behaviour as a whole.

If you want to use a function here, at least make it local to the caller.

MiYanni: Fixed.

PR #7424, opened 2021-07-23
Update cps packages

ocallesp:
The update is needed in order to use the new constructor in `ChainedProjectValueDataSourceBase `that will be used in the implementation of `IVsProjectRestoreInfoSource`.

The two warnings [VSTHRD110 ](https://github.com/microsoft/vs-threading/issues/233) detected were suppressed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7424)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.cs
@@ -44,7 +44,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages
 
         protected override Task<int> OnApplyAsync()
         {
+#pragma warning disable VSTHRD110 // Observe result of async calls
             return _control?.ApplyAsync() ?? Task.FromResult((int)HResult.Fail);
+#pragma warning restore VSTHRD110 // Observe result of async calls
         }
 
         protected override Task OnDeactivateAsync()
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.cs
```diff
@@ -44,7 +44,9 @@ protected override async Task OnSetObjectsAsync(bool isClosing)
 
         protected override Task<int> OnApplyAsync()
         {
+#pragma warning disable VSTHRD110 // Observe result of async calls
```
tmeschter, line 47: Do we know why this is necessary? On the surface it looks like a bug in the analyzer.

ocallesp: In this scenario I am not sure why this is needed.

Should suppress it at a higher level ?

tmeschter: No, this is the right place to suppress it, but it looks like there's a bug in the analyzer that prevents it from properly understanding this code pattern.

I suggest filing a bug at https://github.com/microsoft/vs-threading, and including the link to the bug with the suppression.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/WpfBasedPropertyPage.cs
ocallesp: @sharwell

sharwell: yes it's a bug

drewnoakes: https://github.com/microsoft/vs-threading/issues/880

PR #7425, opened 2021-07-23
Export IProjectSystemReferenceCleanupService2

ocallesp:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7425)


tmeschter (approved):

drewnoakes:
Is something explicitly importing version two of the interface? It seems more common that the unversioned interface is imported, then type cast to the versioned one.

ocallesp:
> Is something explicitly importing version two of the interface? It seems more common that the unversioned interface is imported, then type cast to the versioned one.

cc. @JoeRobich

JoeRobich:
> It seems more common that the unversioned interface is imported, then type cast to the versioned one.

Oh, If that is the precedent, then that is fine. I just wasn't aware.

PR #7428, opened 2021-07-24
Fast up-to-date check works on solution open

### LINKED ISSUE #3826: FastUpToDate check is wrong after initial Rebuild in Visual Studio
Steps to reproduce:
Pre-req: Options->.NET Core->Up to Date Checks: Check "Don't call MSBuild if a project appears to be up to date" (Note: I have logging level turned to Verbose)

1. Open a solution with one netStandard project.
2. Rebuild Solution (Results in "1 succeeded", and any pre/post build events are run)
3. Build Solution
Bug: 
- Results in "1 succeeded", and pre/post build events are run.
- Output: `FastUpToDate: The list of source items has changed since the last build, not up to date. (netstandardlib)`
4. Build Solution (Results in "1 up-to-date", no pre/post build events are run)
Output: `FastUpToDate: Project is up to date. (myprojectname)`

The expectation is that in Step 3 above, that the Result would be "1 up-to-date" with no pre/post build events being run (i.e. what happens in Step 4).

Notes:
- Subsequent Rebuilds/builds in the same session of Visual Studio have the expected behavior.  However, restarting Visual Studio, the bug appears again.  
- Here's a video demonstrating the issue: http://www.screencast.com/t/fCndQNvg

### LINKED ISSUE #6633: Up-to-date Check does not work properly on first solution opening
**Visual Studio Version**: 16.7.3

**Summary**:
Up-to-date check doesn't work when a solution is opened for the first time. 

**Steps to Reproduce**:

1. Create a SDK-style project and the corresponding solution

2. Build using command-line

3. Open Visual Studio and build 

**Expected Behavior**:
The project should be considered as "up-to-date" since it has already been built using command-line.

**Actual Behavior**:
MSBuild is triggered on the project :

> FastUpToDate: The up-to-date check has not yet run for this project. Not up-to-date. (SampleProject)
>------ Build started: Project: SampleProject, Configuration: Debug Any CPU ------
>SampleProject -> C:\Temp\SampleProject\bin\Debug\netcoreapp3.1\SampleProject.dll
>========== Build: 1 succeeded, 0 failed, **0 up-to-date**, 0 skipped ==========

**User Impact**:
We have huge solutions (300+ projects, mix of SDK-style C# projects and C++ projects) and our workflow requires to first build the repository using command-line. It's pretty inconvenient to recompile all the projects inside Visual Studio (even if it is not, I assume, a full recompilation) when opening the solution.

Is there anyway we could fix this behavior ? 
Thanks

drewnoakes:
Fixes #6633
Fixes #3826

Previously, the first up-to-date check that runs after a project loads always reports that a build is required. This avoided an uncommon yet serious potential issue.

Consider a project that brings in all `.cs` files via a glob, with the following steps:

1. Load the project
1. Build the project
1. Close the project
1. Delete one of the input `.cs` files from disk
1. Reopen the project

If the fast up-to-date check only considered the timestamps of input and output files, the project would appear up-to-date. However, it is not.

If an item is removed while Visual Studio has the project loaded, the change is observed and handled correctly.

Note that it is also possible to trigger this situation by adding a file, however usually when a file is added its timestamp is updated. Of course software can set the timestamp to whatever it wants. One easy way to do this is to undelete a file, which restores it with the original timestamp.

In order to correctly handle the above scenario, we need to persist some information about the project's inputs, so that when it is reloaded we can decide whether any inputs have been removed from disk.

That is what this change does. We introduce `IUpToDateCheckStatePersistence` which stores state in the `.vs` folder. That state contains a hash of the items in each project configuration.

With this change, we are now able to correctly identify whether a reloaded project is up-to-date or not, and avoid the precautionary build.

For OrchardCore, this cuts the time for first build after solution load from just over a minute to around one second, in the case that all 157 projects were already up-to-date.

This change will also reduce wait times when running unit tests after loading a solution, or launching a debugger. It will also reduce wait times when building after the solution being reloaded, perhaps due to the `.sln` file changing during a branch switch.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7428)

### DIFF
--- a/docs/repo/up-to-date-check-implementation.md
+++ b/docs/repo/up-to-date-check-implementation.md
@@ -7,6 +7,8 @@ The `IBuildUpToDateCheckProvider` interface (from CPS) has two members:
 - `IsUpToDateCheckEnabledAsync`, which is serviced by `IProjectSystemOptions.GetFastUpToDateLoggingLevelAsync`
 - `IsUpToDateAsync` which performs the checks described below
 
+We implement the derived `IBuildUpToDateCheckProvider2` version of the interface, which gives us access to the set of global properties used in the build. A build may use global properties to exlude certain "kinds" of inputs and outputs if necessary.
+
 ## What is checked
 
 There are several checks which must all pass. 
@@ -68,6 +70,15 @@ For each `_copiedOutputFiles` source/destination
 - :heavy_check_mark: ...the source must exist
 - :heavy_check_mark: ...the destination must exist
 - :heavy_check_mark: ...the destination must be older or the same age as the source
+
+## Persistence
+
+Our up-to-date check persists data to the `.vs` folder, in a binary file such as `.vs/MySolution/v17/.futdcache.v1`.
+
+This file contains, per project:
+
+1. A hash of the items in the project, so that we can tell whether the items have changed since the last time we loaded the project. If not, we can consider the project up-to-date on solution open, which can save a lot of time for the user's first build/test/debug operations.
+2. The time at which the last set of items was observed to have changed. This is also important on solution open, to correctly handle the first up-to-date check of a project.
 
 ## Implementation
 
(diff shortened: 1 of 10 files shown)


arkalyanms:
> If the fast up-to-date check only considered the timestamps of input and output files, the project would appear up-to-date. However, it is not.
@drewnoakes : I didn't quite follow when timestamps appear up-to-date, but the cache is actually not up-to-date. Would you please elaborate?

melytc (approved):

drewnoakes:
>> If the fast up-to-date check only considered the timestamps of input and output files, the project would appear up-to-date. However, it is not.
>
> @drewnoakes : I didn't quite follow when timestamps appear up-to-date, but the cache is actually not up-to-date. Would you please elaborate?

@arkalyanms by "cache" are you referring to MSBuild's cache?

The scenario here is when the project is built, closed, and source file or other input is removed. Because the VS FUTD check had no state retained from the last time it had the solution open, it wouldn't know that the input file was removed. It would still assume that all inputs are older than the outputs and determine the project as up-to-date.

PR #7429, opened 2021-07-25
Class libraries cannot be startup projects

### LINKED ISSUE #1308: Class Libraries show up in debug startup project list

drewnoakes:
Fixes #1308

The boolean argument for `validateSettings` controls whether class libraries are considered launchable or not.

The changed invocation here is only used when computing items to appear in the startup project list. When `validateSettings` is `false` (as before) then all projects appear, including class libraries. When `validateSettings` is `true` (as now) then they are excluded.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7429)


ocallesp (approved):

ocallesp:
This fix works. I am just wondering why initially `validateSettings `was set to false.

PR #7432, opened 2021-07-27
Add Support for Optional .NET Workloads

swesonga:
Addresses acquisition in the solution load scenario for issue #6080

# Background
- https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workloads.md
- https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workload-resolvers.md

# Description
Optional workloads have been introduced into .NET to enable the SDK to scale without shipping huge monolithic SDKs as .NET support of additional features and platforms (e.g. iOS and Android) grows. As a result, Visual Studio needs to detect when projects are loaded that require optional workloads that are not installed. The in-product acquisition experience should then present users with the list of all components that need to be installed in order to successfully load and use their projects.

# Design
The .NET SDK contains targets that detect when workloads required by the project are absent and then fail the build if there were missing workloads. The missing workloads are output as "SuggestedWorkload" MSBuild items. See https://github.com/dotnet/sdk/blob/release/6.0.1xx-preview7/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets for more details. Note that MSBuild project evaluation needs to happen to gather the list of missing workloads.

This change introduces the IWorkloadDescriptorDataSource to provide WorkloadDescriptors containing Visual Studio setup component IDs to the project system. These are then passed to a MissingWorkloadRegistrationService which aggregates these VS component IDs and registers them with the setup system. The setup system consequently displays a banner letting the user know that they need to install extra components. The setup system needs to present this list of missing components to the user when they click on the "Install" button. This is done via a new IMissingWorkloadRegistrationService, which is obtained using the service broker mechanism.

The IMissingWorkloadRegistrationService interface is defined in the RPC contracts in PR https://devdiv.visualstudio.com/DevDiv/_git/VS.RPC.Contracts/pullrequest/340386
IMissingWorkloadRegistrationService is implemented in the Visual Studio shell's setup Package in PR https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/340388

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7432)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/OptionalWorkloadProjectLoadVeto.cs
```diff
@@ -0,0 +1,107 @@
...
+                            return false;
+                        }
+
+                        return true;
+                    },
+                    cancellationToken);
+                }
+                finally
+                {
+                    (missingWorkloadRegistrationService as IDisposable)?.Dispose();
```
drewnoakes, line 91: Disposal of this is already covered.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/OptionalWorkloadProjectLoadVeto.cs
```diff
@@ -0,0 +1,107 @@
...
+                }
+            }
+        }
+
+        private IEnumerable<string> GetVSComponentIdsForMissingWorkloads(Project project)
```
drewnoakes, line 96: Consider returning `IReadOnlyList<>` here to convey that it's safe for consuming code to enumerate multiple times (as it currently does).

drewnoakes (approved):

docs/repo/acquisition/optional-workloads.md
```diff
@@ -0,0 +1,18 @@
+Add Support for Optional .NET Workloads
```
swesonga, line 1: This file is outdated, I'm still updating it with details of the current design.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/RuleExporter.cs
```diff
@@ -201,5 +201,16 @@ private static class BuildUpToDateCheckRules
...
         }
+
+        private static class OptionalWorkloadRules
+        {
+            /// <summary>
+            ///     Represents the evaluation items representing optional workloads.
+            /// </summary>
+            [ExportRule(nameof(SuggestedWorkload), PropertyPageContexts.ProjectSubscriptionService)]
+            [AppliesTo(ProjectCapability.DotNet)]
+            [Order(Order.Default - 1)]
```
swesonga, line 212:
```suggestion
            [Order(Order.Default)]
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/MissingWorkloadRegistrationService.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                return;
+            }
+
+            var missingWorkloadRegistrationService = await serviceBroker.GetProxyAsync<IMissingComponentRegistrationService>(
+                serviceDescriptor: VisualStudioServices.VS2022.MissingComponentRegistrationService);
+
+            using (missingWorkloadRegistrationService as IDisposable)
+            {
+                if (missingWorkloadRegistrationService != null)
```
MiYanni, line 119: Little confused by this. According to this, you shouldn't really create the variable outside of the using statement scope: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement#remarks

I don't think the `as Disposable` does anything... since you're making a temporary boxing of the variable that is then not stored... so I don't know if it is actually being disposed? And checking if it is non-null inside the using confuses me even more. If you check if it is null outside the using, then there's no reason to have a using? You can't dispose something that is null. Again, brain confusion. 😵

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/MissingWorkloadRegistrationService.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (projectConfigurationSet.Count == 0 && ShouldDisplayMissingComponentsPrompt())
+                {
+                    return DisplayMissingComponentsPromptAsync(cancellationToken);
+                }
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public void RegisterProjectConfiguration(Guid projectGuid, ProjectConfiguration projectConfiguration)
```
drewnoakes, line 72: I don't see a way to de-register these.

What happens if a configuration is removed? For example, if I change a target framework in a multi-targeting project.

What happens if I remove a project from the solution?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -234,6 +234,10 @@
       <Context>Project</Context>
     </PropertyPageSchema>
 
+    <PropertyPageSchema Include="$(ManagedXamlNeutralResourcesDirectory)SuggestedWorkload.xaml">
+      <Context>Project</Context>
```
drewnoakes, line 238: This context does not match that of the export (`ProjectSubscriptionService`).

Is this project item needed for exported rules?

drewnoakes (commented):
Looking good. Left some thoughts.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/MissingWorkloadDetector.cs
```diff
@@ -0,0 +1,115 @@
...
+        private void OnWorkloadDescriptorComputed(IProjectVersionedValue<(IProjectSnapshot projectSnapshot, WorkloadDescriptor workloadDescriptor)> pair)
+        {
+            if (!_enabled || _hasNoMissingWorkloads == true)
+            {
+                return;
+            }
+
+            if (WorkloadDescriptor.Empty.Equals(pair.Value.workloadDescriptor))
+            {
+                _hasNoMissingWorkloads = true;
```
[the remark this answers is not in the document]
swesonga: > The workload is empty, so shouldn't `_hasNoMissingWorkloads` be `false`?

I think that would be the case if the variable was named _hasMissingWorkloads

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WorkloadDescriptor.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        ///     Gets the name of the .NET workload.
+        /// </summary>
+        public string WorkloadName { get; }
+
+        /// <summary>
+        ///     Gets the Visual Studio setup component ID corresponding to the .NET workload.
+        /// </summary>
+        public string VisualStudioComponentId { get; }
```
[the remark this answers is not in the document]
swesonga: Good point, 

> Why is `VisualStudioComponentId` not considered for equality or hash codes?
> 
Good point, it should be considered as well.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/VisualStudioParentWorkloadTransformer.cs
```diff
@@ -0,0 +1,83 @@
...
+        };
+
+        private static readonly Dictionary<string, IReadOnlyCollection<string>> s_vsComponentIdToParentComponentsMap = new(StringComparers.VisualStudioSetupComponentIdComparer)
+        {
+            { MauiAndroidWorkloadName, s_mauiAndroidComponents },
+            { MauiIOSWorkloadName, s_mauiIOSComponents },
+            { MauiMacCatalystWorkloadName, s_mauiMacCatalystComponents }
+        };
+
+        public Task<IReadOnlyCollection<string>> TransformVisualStudioComponentIdsAsync(IReadOnlyCollection<string> vsComponentIds)
```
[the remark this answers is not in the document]
swesonga: > Consider returning `IReadOnlySet<T>` for a little more semantic information about the response. I don't believe there's any runtime difference here.

That's an internal CPS interface. Perhaps it should be exposed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
swesonga: > This context does not match that of the export (`ProjectSubscriptionService`).
> 
> Is this project item needed for exported rules?

I don't think it is. I have removed it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/MissingWorkloadRegistrationService.cs
swesonga: > I don't think the `as Disposable` does anything... since you're making a temporary boxing of the variable that is then not stored... so I don't know if it is actually being disposed? And checking if it is non-null inside the using confuses me even more. If you check if it is null outside the using, then there's no reason to have a using? You can't dispose something that is null. Again, brain confusion. 😵

@MiYanni The service does not derive from IDisposable. See the analyzer notes at https://dev.azure.com/devdiv/DevDiv/_git/DevCore?path=%2Fdoc%2Fanalyzers%2FISB001.md&_a=preview

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/MissingWorkloadRegistrationService.cs
swesonga: > I don't see a way to de-register these.

I have added an explicit Unregister method to the service for the MissingWorkloadDetector. Thanks for the feedback.

PR #7434, opened 2021-07-28
Move Windows Forms specific rules

RussKie:
Move Windows Forms specific rules required for Windows Forms designer.

The rules were originally introduced in https://github.com/dotnet/winforms-designer/pull/3509, however that resulted in [VS perf regression](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1359439). Moving the rules here to address the regression.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7434)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/WindowsFormsProjectSchema.xaml
```diff
@@ -0,0 +1,32 @@
+﻿<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
+<Rule Name="WindowsFormsConfigurationGeneral"
+      DisplayName="WindowsFormsConfigurationGeneral"
```
drewnoakes, line 3: Can we make the name consistent across this PR? Currently it uses both `WindowsFormsConfigurationGeneral` and `WindowsFormsProjectSchema`.

We use "configuration general" for general properties, and this is WinForms-specific. 

I think a more consistent name would be `WindowsFormsConfiguration`.

RussKie: I've renamed a bunch of them. Please check.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/RuleExporter.cs
```diff
@@ -201,5 +201,16 @@ private static class BuildUpToDateCheckRules
...
+
+        /// <summary>
+        ///     Contains rules for the Windows Forms designer component.
+        /// </summary>
+        private static class WindowsFormsConfigurationRules
+        {
+            [ExportRule(nameof(WindowsFormsConfiguration), PropertyPageContexts.Project)]
+            [AppliesTo(ProjectCapability.DotNet)]
+            [Order(Order.Default)]
+            public static int WindowsFormsConfigurationSchemaRule;
```
drewnoakes, line 213:
```suggestion
            public static int WindowsFormsConfigurationRule;
```

drewnoakes (approved):

RussKie:
ested locally be making necessary changes to the Project System, deploying those into Exp hive, and then running the designer and verifying that .NET 6.0 apps rendered with a custom font.

RussKie:
Thank you heaps for you help!

Issue #6457, opened 2020-07-31
Package Properties tab should expose the new Readme property


advay26-zz:
Similar to [#4905](https://github.com/dotnet/project-system/issues/4905), Embedded Readmes are being introduced and will be included in packages - [Embedding and displaying NuGet READMEs](https://github.com/NuGet/Home/wiki/Embedding-and-displaying-NuGet-READMEs). 

The behavior for Project Properties would be similar to Icon and License File support:

- Using a file picker, users can select a markdown file on disc (.md)

- On selection, the value updates to the file's relative path

- The project file is updated as below:

```
    <PropertyGroup>
        ...
        <PackageReadmeFile>readme.md</PackageReadmeFile>
        ...
    </PropertyGroup>

    <ItemGroup>
        ...
        <None Include="..\assets\readme.md" Pack="true" PackagePath=""/>
        ...
    </ItemGroup>
```

The UI view:

![image](https://user-images.githubusercontent.com/51111468/89075783-f7683b80-d34c-11ea-8f39-668555f9b7bb.png)

drewnoakes:
We will likely add support for editor metadata here that sets the FileOpenDialog default extension to `.md`, as required in the spec.

drewnoakes:
There is some complexity here, in that the interceptor will need to create/update/remove the `None` item, in addition to the property itself.

drewnoakes:
> There is some complexity here, in that the interceptor will need to create/update/remove the None item, in addition to the property itself.

That complexity also exists for `PackageIcon` which is not currently wired up to manage the project item. This works in the legacy pages. We need to handle this.

https://github.com/dotnet/project-system/blob/034b42b3031ab93a14697b918e52f2ee79a5c55d/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml#L94-L103

MiYanni:
I'll be implementing this into the new property pages.

PR #7440, opened 2021-07-29
Add package readme property

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/6457
Related: https://github.com/dotnet/project-system/pull/7422

This is a quick implementation of the [PackageReadmeFile](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packagereadmefile) property. It is copy-pasted logic from the PackageIcon property that I implemented recently. I'd recommend we extract this logic into some shared code between the 2 providers, but I'll do that in a follow-up PR. For now, this adds the PackageReadmeFile property and the associate None item (and metadata) to the project.

I also did some very minor cleanup to the PackageIcon value provider as I realized I can describe it more accurately and generically at the same time.

The package is loading both PackageReadmeFile and PackageIcon properly.
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7440)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
@@ -220,4 +220,16 @@
     </BoolProperty.Metadata>
   </BoolProperty>
 
+  <StringProperty Name="PackageReadmeFile"
+                  DisplayName="README"
+                  Description="The readme document for the package. Supported file formats include only Markdown (.md)."
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2168540"
+                  Category="General"
+                  Subtype="file">
+    <StringProperty.DataSource>
+      <DataSource Persistence="ProjectFileWithInterception"
+                  HasConfigurationCondition="False" />
+    </StringProperty.DataSource>
+  </StringProperty>
+
 </Rule>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
@@ -103,6 +103,18 @@
     </StringProperty.DataSource>
   </StringProperty>
 
+  <StringProperty Name="PackageReadmeFile"
+                  DisplayName="README"
+                  Description="The readme document for the package. Supported file formats include only Markdown (.md)."
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2168540"
+                  Category="General"
+                  Subtype="file">
+    <StringProperty.DataSource>
+      <DataSource Persistence="ProjectFileWithInterception"
+                  HasConfigurationCondition="False" />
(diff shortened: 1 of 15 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs
```diff
@@ -45,29 +45,28 @@ internal sealed class PackageIconValueProvider : InterceptingPropertyValueProvid
...
 
         // https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageicon
-        private static string CreatePackageIcon(string filePath, string? packagePath)
+        private static string CreatePackageIcon(string filePath, string packagePath)
         {
             string filename = Path.GetFileName(filePath);
             // Make a slash-only value into empty string, so it won't get prepended onto the path.
-            // Also, make sure that the packagePath is non-null, so Path.Combine doesn't throw an ArgumentNullException.
-            if (packagePath is null || "\\".Equals(packagePath, StringComparisons.Paths) || "/".Equals(packagePath, StringComparisons.Paths))
+            if ("\\".Equals(packagePath, StringComparisons.Paths) || "/".Equals(packagePath, StringComparisons.Paths))
```
MiYanni, line 52: Null check here was not necessary. Every call to this method will provide either a value or `string.Empty` for `packagePath`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -103,6 +103,18 @@
     </StringProperty.DataSource>
   </StringProperty>
 
+  <StringProperty Name="PackageReadmeFile"
```
MiYanni, line 106: I put this right after PackageIcon since that was the order in the NuGetPackageExplorer. I'm open to moving this around if there is a better spot for it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -103,6 +103,18 @@
...
+                  DisplayName="README"
+                  Description="The readme document for the package. Supported file formats include only Markdown (.md)."
+                  HelpUrl="https://go.microsoft.com/fwlink/?linkid=2168540"
```
drewnoakes, line 109: This fwlink redirects to a URL that contains `en-us` in it. Please update the fwlink from:

```diff
-https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packagereadmefile
+https://docs.microsoft.com/nuget/reference/msbuild-targets#packagereadmefile
```

drewnoakes (approved):
Nice!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
MiYanni: Updated.

PR #7442, opened 2021-07-29
Specify extensions for dialog shown for file properties

### LINKED ISSUE #7420: New Property Pages: Add ability to specify extension information to Browse dialog (subtype "file") properties

drewnoakes:
Fixes #7420

In https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/341091 we add support for properties to specify specific file extensions to be used as filters in the "open file dialog" we use when clicking "Browse" next to certain file path properties.

This PR adds the metadata to drive that feature, and documents its use for future.

image

image

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7442)

### DIFF
--- a/docs/repo/property-pages/property-specification.md
+++ b/docs/repo/property-pages/property-specification.md
@@ -348,16 +348,41 @@ internal sealed class MyCommandActionHandler : ILinkActionHandler
 }
 ```
 
-## File and Directory Properties
+## File Properties
 
-When a property's value represents a file or directory path, it should be modelled as a `StringProperty` with its `Subtype` attribute set to `file` or `directory` respectively. `folder` is an equivalent alternative to `directory`.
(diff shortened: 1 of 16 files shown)


ocallesp (approved):

jjmew (approved):
:shipit:

MiYanni (approved):

PR #7446, opened 2021-07-30
Remove unneeded UI thread dependency

davkean:
Fixes: .NET Project System's side of https://github.com/dotnet/project-system/issues/353.

Roslyn already immediately switches to the UI thread CreateProjectContextAsync, so this should be a no-op. However, this enables the above bug to be moved over to Roslyn and for them to remove the UI thread dependency in the future.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7446)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceProjectContextProviderTests.cs
```diff
@@ -220,11 +220,10 @@ private static WorkspaceProjectContextProvider CreateInstance(UnconfiguredProjec
 
             var projectFaultService = IProjectFaultHandlerServiceFactory.Create();
             project ??= UnconfiguredProjectFactory.Create();
-            threadingService ??= IProjectThreadingServiceFactory.Create();
```
drewnoakes, line 223: Also remove the `threadingService` parameter from this method as it is now unused.

drewnoakes (approved):

PR #7452, opened 2021-08-03
Enable Hot Reload during debugging

tmeschter:
Currently, we only enable Hot Reload in non-debugging (i.e., Ctrl+F5) scenarios, and equivalent functionality is provided by Edit and Continue (EnC) in debugging scenarios. Under the hood, however, EnC and Hot Reload are use mostly the same code paths and are really just two different names for the same feature.

And now that the feature is orthogonal to debugging, we would like to have one way to turn it on that is not tied to whether or not we are debugging the running process. Here we remove the check in `ProjectLaunchTargetsProvider` that limited Hot Reload to Ctrl+F5 scenarios.

However, there are some aspects of Hot Reload that need to understand whether or not the process is running under the debugger, and we are now also responsible for passing that information along. This is largely a matter of calling a different overload of `IHotReloadAgentManagerClient.AgentStartedAsync`, which in turn requires a number of changes to pipe the value through.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7452)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
@@ -20,7 +21,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
         /// <summary>
         /// Starts the Hot Reload Session.
         /// </summary>
+        /// <remarks>
+        /// TODO: remove when Web Tools is no longer calling this method.
+        /// </remarks>
+        [Obsolete("This should no longer be used; please use StartSessionAsync(bool, CancellationToken) instead.", false)]
         Task StartSessionAsync(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Starts the Hot Reload Session.
+        /// </summary>
+        /// <param name="runningUnderDebugger">
+        /// <see langword="true"/> if the process is being run under a debugger;
+        /// <see langword="false"/> otherwise.
+        /// </param>
+        /// <param name="cancellationToken">A token indicating if the operation has been cancelled.</param>
+        Task StartSessionAsync(bool runningUnderDebugger, CancellationToken cancellationToken);
 
         /// <summary>
         /// Stops the Hot Reload Session.
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
@@ -73,13 +73,20 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
             }
 
             return false;
+}
+
+        // TODO: remove when Web Tools is no longer calling this method.
+        public Task StartSessionAsync(CancellationToken cancellationToken)
+        {
+            return StartSessionAsync(runningUnderDebugger: false, cancellationToken);
         }
 
-        public async Task StartSessionAsync(CancellationToken cancellationToken)
+        public async Task StartSessionAsync(bool runningUnderDebugger, CancellationToken cancellationToken)
         {
             if (!_sessionActive)
             {
-                await _hotReloadAgentManagerClient.Value.AgentStartedAsync(this, cancellationToken);
+                HotReloadAgentFlags flags = runningUnderDebugger ? HotReloadAgentFlags.IsDebuggedProcess : HotReloadAgentFlags.None;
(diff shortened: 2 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
```diff
@@ -20,8 +21,22 @@ public interface IProjectHotReloadSession
         /// <summary>
         /// Starts the Hot Reload Session.
         /// </summary>
+        /// <remarks>
+        /// TODO: remove when Web Tools is no longer calling this method.
+        /// </remarks>
+        [Obsolete("This should no longer be used; please use StartSessionAsync(bool, CancellationToken) instead.", false)]
         Task StartSessionAsync(CancellationToken cancellationToken);
```
swesonga, line 28: >StartSessionAsync [](http://example.com/codeflow?start=13&length=17)

can this be removed since we're still in preview, or how will web tools know to not use this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
```diff
@@ -73,13 +73,20 @@ public async Task<bool> ApplyLaunchVariablesAsync(IDictionary<string, string> en
             }
 
             return false;
+}
```
drewnoakes, line 76: nit: indentation

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
```diff
@@ -73,13 +73,20 @@ public async Task<bool> ApplyLaunchVariablesAsync(IDictionary<string, string> en
...
+
+        // TODO: remove when Web Tools is no longer calling this method.
```
drewnoakes, line 78: Is there a work item tracking the WebTools side of this? Can we include the URL here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
```diff
@@ -73,13 +73,20 @@ public async Task<bool> ApplyLaunchVariablesAsync(IDictionary<string, string> en
...
+        public Task StartSessionAsync(CancellationToken cancellationToken)
+        {
+            return StartSessionAsync(runningUnderDebugger: false, cancellationToken);
         }
 
-        public async Task StartSessionAsync(CancellationToken cancellationToken)
+        public async Task StartSessionAsync(bool runningUnderDebugger, CancellationToken cancellationToken)
         {
             if (!_sessionActive)
```
drewnoakes, line 86: Is there a reason this no-ops if called when already started? What if I try to start with `runningUnderDebugger` false the first time, then true the second time? Unless there's a compelling reason not to, consider throwing an exception here if already active.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
```diff
@@ -73,13 +73,20 @@ public async Task<bool> ApplyLaunchVariablesAsync(IDictionary<string, string> en
...
             {
-                await _hotReloadAgentManagerClient.Value.AgentStartedAsync(this, cancellationToken);
+                HotReloadAgentFlags flags = runningUnderDebugger ? HotReloadAgentFlags.IsDebuggedProcess : HotReloadAgentFlags.None;
```
drewnoakes, line 88: Would it be better to have the caller pass these flags directly, rather than a boolean? Future flags wouldn't require an API change in that case.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSession.cs
tmeschter: I'm in communication with the Web Tools team. We can't remove this method without breaking their Hot Reload scenarios, so we're leaving it in until they have a chance to move off of it. Once we do I'll remove it entirely.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
tmeschter: No particular work item, just ongoing Hot Reload work.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
tmeschter: Part of the purpose of this type is to abstract away the details of the Hot Reload machinery; passing the flags themselves would defeat that.

drewnoakes (approved):

Issue #7453, opened 2021-08-04
IVsProjectRestoreInfoSource.WhenNominated returns immediately even when a nomination is pending


nkolev92:
**Visual Studio Version**:
17.0 preview 3

**Summary**:

IVsProjectRestoreInfoSource.WhenNominated returns immediately even when `HasPendingNomination` is true. 

This basically causes NuGet to have to invoke `WhenNominated` multiple times on the same project.

This basically leads to NuGet not being able to accurately reason about the effectiveness of the batching, as it pollutes the data with a high number of checks. 

NuGet still gets to batch the restores, but there are caveats. 

* What is currently happening is effectively busy spin. A threadpool is blocking doing a large amount of work.
* It's technically still more effective than what was happening before, but we're pay more than we need to.

To fully understand what's going on, I added the following logging: https://github.com/NuGet/NuGet.Client/compare/dev-nkolev92-logging?expand=1. 

I have added the logs at the bottom.

**Steps to Reproduce**:

1. Clone https://github.com/NuGet/NuGet.Client
1. Checkout `dev-nkolev92-testBranchSwitch` on the commandline.
1. Load NuGet-Src-Commandline.slnf and observe 1 restore.
1. Branch switch to `dev-nkolev92-testBranchSwitch2` in Visual Studio and observe 1 restore.
1. The branch differences can be observed at https://github.com/NuGet/NuGet.Client/compare/dev-nkolev92-testBranchSwitch2...dev-nkolev92-testBranchSwitch

You can observe the vs/nuget/restoreinformation telemetry and see that ProjectsReadyCheckCount is 100+ instead of the expected 2.

**Expected Behavior**:

WhenNominated returns only after the project design time build has completed.

**Actual Behavior**:

WhenNominated returns immediately. 

**User Impact**:

* Inability to reason about the success of the feature easily.
* Busy threadpool thread, slower batching.

cc @ocallesp @lifengl 
[RestoreCheckLogs.txt](https://github.com/dotnet/project-system/files/6934483/RestoreCheckLogs.txt)

ocallesp:
Pending: check _whenNominatedTask

nkolev92:
Closing in favor of https://github.com/NuGet/Home/issues/11132

PR #7459, opened 2021-08-04
Fix WhenNominated

ocallesp:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7459)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.ProjectRestoreInfoSource.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.ProjectRestoreInfoSource.cs
@@ -144,13 +144,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
                 {
                     _savedNominatedConfiguredVersion[configuredInput.ProjectConfiguration] = configuredInput.ConfiguredProjectVersion;
                 }
+            }
+        }
 
-                if (_whenNominatedTask is not null)
+        private void CompleteTaskWhenNominated()
+        {
+            if (_whenNominatedTask is not null)
+            {
+                if (_whenNominatedTask.TrySetResult(true))
                 {
-                    if (_whenNominatedTask.TrySetResult(true))
-                    {
-                        _whenNominatedTask = null;
-                    }
+                    _whenNominatedTask = null;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.ProjectRestoreInfoSource.cs
```diff
@@ -69,7 +69,8 @@ public Task WhenNominated(CancellationToken cancellationToken)
                             }
                             else
                             {
-                                _whenNominatedTask?.TrySetCanceled();
+                                _whenNominatedTask?.TrySetResult(true);
```
ocallesp, line 72: Finish task when design-time build finishes

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.ProjectRestoreInfoSource.cs
```diff
@@ -144,13 +144,16 @@ private void SaveNominatedConfiguredVersions(IReadOnlyCollection<PackageRestoreC
                 {
                     _savedNominatedConfiguredVersion[configuredInput.ProjectConfiguration] = configuredInput.ConfiguredProjectVersion;
                 }
+            }
+        }
 
-                if (_whenNominatedTask is not null)
+        private void CompleteTaskWhenNominated()
+        {
+            if (_whenNominatedTask is not null)
```
davkean, line 152: This used to be protected by a lock, now it is not. This has a race, by the time you've run TrySetResult this field could be null.

lifengl: yeah, I don't understand this is the fix. Just like David said, this must be in a lock, and also inside the same lock region when _saveNorminatedConfiguredVersion is updated.  There is no logic reason why this is a fix to the problem. Do we know the exact reason why nominated task is completed, because we return an expired task, or we complete it very quickly.

One possibility of this is that NuGet actually ask other projects when first project nominates, and then wait on the second project, which will nominate in a short time. Basically, it is chasing nominations in the system. Is that a possible explanation of this.  If that is the case, it might be expected?

nkolev92: NuGet iterates over the known project checking HasPendingNomination. 
If any project returns true, then we call WhenNominated. 
NuGet will wait on at most one project at any given time. 

https://github.com/dotnet/project-system/issues/7453 has some logs of all nuget calls to the project system.

Issue #7167, opened 2021-04-30
Program Icon property must copy .ico file to project (2)


sujitnayak:
**Visual Studio Version**: 16.10 P3 Main builds (31229.188.main on my machine)

**Summary**:
Adding program icon with new Project Properties UX has a different behavior than the previous UX as described in Actual Behavior section.

**Steps to Reproduce**:

1. Create WPF or Winforms project for .NET Core/.NET FX 5.0

2. Go to Project Properties -> Application -> Resources tab and add program icon.

3. 

**Expected Behavior**:
The .ico file gets copied to the project directory and ApplicationIcon property is written to project file with just the name of the .ico file

**Actual Behavior**:
The icon file does not get copied to project folder and the ApplicationIcon property written to project file is fully qualified path to the .ico file.

**User Impact**:
ClickOnce publish fails due to this breaking change as it expects the .ico file to be in the project directory.

melytc:
For reference, this is the previous experience looks like:
![image](https://user-images.githubusercontent.com/8518253/119036552-b5ed5f00-b965-11eb-8b64-2cd7af4bab2a.png)

And this is the new experience:
![image](https://user-images.githubusercontent.com/8518253/119036669-d61d1e00-b965-11eb-8f7a-75b61273c5aa.png)

drewnoakes:
See also #6457 for a very similar work item on a different property. I suspect the solution to each will look similar.

PR #7461, opened 2021-08-04
ApplicationIcon property fix

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7167

## Description
This allows the ApplicationIcon property to function similarly to how it did in the old property pages. How it worked was, if the file was within the project directory structure (project directory and below), it would use a relative path (from the project directory) to that file and set that as the ApplicationIcon's value. If the file was not within the project directory structure, it would copy it to the project directory and then simply have the filename as the path (which is effectively a relative path to the file). This implementation does that same interaction.

One minor difference here is, in the old pages, if the file already existed in the project directory when trying to copy it, it would show this error.
image
In this implementation, I simply override the file if it already exists in the project directory. Therefore, no error occurs.

The legacy property pages also used a completely different kind of text input control. It used a free-form dropdown. The dropdown itself contained .ico files that are already part of the project. Here, we don't have a control like that, so that implementation is not featured in the new pages. If we wanted something like that, we'd likely have a selector-style control. If I'm not mistaken, we might be implementing something like that for the VB property pages, so if that is desired here, we can make the switch. Otherwise, this works just as a text field with a Browse button, which is more than enough in a majority of scenarios.

NOTE: I believe the file filter to not be present in the Browse window since I haven't updated my VS version on this machine. I'm going to do that now, but I wanted to get the PR in first.

## GIF
ApplicationIconFix

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7461)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
@@ -51,8 +51,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
                     var destinationInfo = new FileInfo(Path.Combine(Path.GetDirectoryName(_unconfiguredProject.FullPath), fileName));
                     if (destinationInfo.Exists && destinationInfo.IsReadOnly)
                     {
-                        // https://stackoverflow.com/a/8081331/294804
-                        File.SetAttributes(destinationInfo.FullName, destinationInfo.Attributes & ~FileAttributes.ReadOnly);
+                        // The file cannot be copied over; return the previous value.
+                        return await defaultProperties.GetUnevaluatedPropertyValueAsync(propertyName);
                     }
                     _fileSystem.CopyFile(unevaluatedPropertyValue, destinationInfo.FullName, overwrite: true);
                     return fileName;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
@@ -8,24 +8,12 @@ using Microsoft.VisualStudio.IO;
 
 namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
 {
-    // Sets the PackageIcon property to the path from the root of the package, and creates the
-    // None item associated with the file on disk. The Include metadata of the None item is a relative
-    // path to the file on disk from the project's directory. The None item includes 2 metadata elements
-    // as Pack (set to True to be included in the package) and PackagePath (directory structure in the package
-    // for the file to be placed). The PackageIcon property will use the directory indicated in PackagePath,
-    // and the filename of the file in the Include filepath.
-    //
-    // Example:
-    // <PropertyGroup>
-    //   <PackageIcon>content\shell32_192.png</PackageIcon>
-    // </PropertyGroup>
-    //
-    // <ItemGroup>
-    //   <None Include="..\..\..\shell32_192.png">
-    //     <Pack>True</Pack>
-    //     <PackagePath>content</PackagePath>
-    //   </None>
-    // </ItemGroup>
+    // Sets the ApplicationIcon property to the path to the .ico file. This path is relative
+    // to the project file directory. If the selected file is not within the project directory
+    // tree, the file will be copied into the project file directory. Previously copied files
+    // are not deleted when a new file is selected. ApplicationIcon allows for both full paths
+    // and relative paths anywhere on disk, but we are not utilizing that in this implementation.
+    // The icon file does not need to be included in the project (such as a None/Content item).
     [ExportInterceptingPropertyValueProvider(ApplicationIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
     internal sealed class ApplicationIconValueProvider : InterceptingPropertyValueProviderBase
     {
@@ -41,8 +29,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
             _fileSystem = fileSystem;
         }
 
-        // TODO: Check if we need to delete previous file
-        // TODO: Check if file is already imported as a none/content item
         public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
         {
             // Allows the user to clear the application icon.
@@ -68,32 +54,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
             }
 
             return PathHelper.MakeRelative(_unconfiguredProject, unevaluatedPropertyValue);
-
-            //string relativePath = PathHelper.MakeRelative(_unconfiguredProject, unevaluatedPropertyValue);
-            //string existingPropertyValue = await defaultProperties.GetEvaluatedPropertyValueAsync(PackageIconPropertyName);
-            //IProjectItem? existingItem = await GetExistingNoneItemAsync(existingPropertyValue);
-            //string packagePath = string.Empty;
-            //if (existingItem != null)
-            //{
-            //    packagePath = await existingItem.Metadata.GetEvaluatedPropertyValueAsync(PackagePathMetadataName);
-            //    // The new filepath is the same as the current. No item changes are required.
-            //    if (relativePath.Equals(existingItem.EvaluatedInclude, StringComparisons.Paths))
-            //    {
-            //        return CreatePackageIcon(existingItem.EvaluatedInclude, packagePath);
-            //    }
-            //}
-
-            //// None items outside of the project file cannot be updated.
-            //if (existingItem?.PropertiesContext.IsProjectFile ?? false)
-            //{
-            //    await existingItem.SetUnevaluatedIncludeAsync(relativePath);
-            //}
-            //else
-            //{
-            //    await _sourceItemsProvider.AddAsync(None.SchemaName, relativePath, new Dictionary<string, string> { { PackMetadataName, bool.TrueString }, { PackagePathMetadataName, packagePath } });
-            //}
-
-            //return CreatePackageIcon(relativePath, packagePath);
         }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
@@ -1,0 +1,99 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.IO;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.Package
+{
+    // Sets the PackageIcon property to the path from the root of the package, and creates the
+    // None item associated with the file on disk. The Include metadata of the None item is a relative
+    // path to the file on disk from the project's directory. The None item includes 2 metadata elements
+    // as Pack (set to True to be included in the package) and PackagePath (directory structure in the package
+    // for the file to be placed). The PackageIcon property will use the directory indicated in PackagePath,
+    // and the filename of the file in the Include filepath.
+    //
+    // Example:
+    // <PropertyGroup>
+    //   <PackageIcon>content\shell32_192.png</PackageIcon>
+    // </PropertyGroup>
+    //
+    // <ItemGroup>
+    //   <None Include="..\..\..\shell32_192.png">
+    //     <Pack>True</Pack>
+    //     <PackagePath>content</PackagePath>
+    //   </None>
+    // </ItemGroup>
+    [ExportInterceptingPropertyValueProvider(ApplicationIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class ApplicationIconValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string ApplicationIconPropertyName = "ApplicationIcon";
+
+        private readonly UnconfiguredProject _unconfiguredProject;
+        private readonly IFileSystem _fileSystem;
+
+        [ImportingConstructor]
+        public ApplicationIconValueProvider(UnconfiguredProject unconfiguredProject, IFileSystem fileSystem)
+        {
+            _unconfiguredProject = unconfiguredProject;
+            _fileSystem = fileSystem;
+        }
+
+        // TODO: Check if we need to delete previous file
+        // TODO: Check if file is already imported as a none/content item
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+    // Sets the ApplicationIcon property to the path to the .ico file. This path is relative
+    // to the project file directory. If the selected file is not within the project directory
+    // tree, the file will be copied into the project file directory. Previously copied files
+    // are not deleted when a new file is selected. ApplicationIcon allows for both full paths
+    // and relative paths anywhere on disk, but we are not utilizing that in this implementation.
+    // The icon file does not need to be included in the project (such as a None/Content item).
+    [ExportInterceptingPropertyValueProvider(ApplicationIconPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class ApplicationIconValueProvider : InterceptingPropertyValueProviderBase
+    {
+        private const string ApplicationIconPropertyName = "ApplicationIcon";
```
drewnoakes, line 20: We have constant `ConfigurationGeneral.ApplicationIconProperty` which should be used in this type instead of defining a new constant.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+                return await defaultProperties.GetUnevaluatedPropertyValueAsync(propertyName);
+            }
+
+            // Copy the file into the project directory.
+            // This isn't a limitation of ApplicationIcon; it is to simply mimic the legacy property page interaction.
+            if (_unconfiguredProject.IsOutsideProjectDirectory(unevaluatedPropertyValue))
+            {
+                string fileName = Path.GetFileName(unevaluatedPropertyValue);
+                string destination = Path.Combine(Path.GetDirectoryName(_unconfiguredProject.FullPath), fileName);
+                _fileSystem.CopyFile(unevaluatedPropertyValue, destination, overwrite: true);
```
drewnoakes, line 52: What happens if the file already exists but is read only on the filesystem?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+                return fileName;
```
drewnoakes, line 53: Shouldn't this return `destination`? I don't understand why we copy the file to the project directory if we don't then use that path.

MiYanni: The path to the file (what is put into `ApplicationIcon`) is relative to the project directory. Therefore, the filename (example.ico) is the path to the file, which is why it is returned. This is how the legacy property pages worked too. `destination` is simply needed since this `IFileSystem` API needs 2 explicit paths to do the copy action. I tried just putting the directory as the destination at first, and it doesn't have the smarts to recognize that I wanted to copy a file into a folder.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
MiYanni: Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
MiYanni: This was interesting. Because the `Browse` button logic inserts the full-path, the exception that occurs causes the value provider to not return a new value. So, the UI just kinda doesn't make sense with what the actual value is (which remains the previous value). Therefore, I added logic to attempt to remove the read-only attribute if the file both already exists and is read-only. Then, I wrapped it in a try-catch since any number of other exceptions can happen. If the copy fails, it'll just return the previous value now.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -47,10 +45,23 @@ public ApplicationIconValueProvider(UnconfiguredProject unconfiguredProject, IFi
...
-                _fileSystem.CopyFile(unevaluatedPropertyValue, destination, overwrite: true);
-                return fileName;
+                try
+                {
+                    string fileName = Path.GetFileName(unevaluatedPropertyValue);
+                    var destinationInfo = new FileInfo(Path.Combine(Path.GetDirectoryName(_unconfiguredProject.FullPath), fileName));
+                    if (destinationInfo.Exists && destinationInfo.IsReadOnly)
+                    {
+                        // https://stackoverflow.com/a/8081331/294804
+                        File.SetAttributes(destinationInfo.FullName, destinationInfo.Attributes & ~FileAttributes.ReadOnly);
```
drewnoakes, line 55: I don't think this is safe to do. For example, if using TFS, the file may be marked as read-only because it is currently not checked out.

I'd suggest using your failure path if the target file exists and is read only. Ideally the user would see a message explaining why it failed, but we don't have a good way of surfacing such issues from interceptors.

MiYanni: Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -0,0 +1,70 @@
...
+                try
+                {
+                    string fileName = Path.GetFileName(unevaluatedPropertyValue);
+                    var destinationInfo = new FileInfo(Path.Combine(Path.GetDirectoryName(_unconfiguredProject.FullPath), fileName));
+                    if (destinationInfo.Exists && destinationInfo.IsReadOnly)
+                    {
+                        // The file cannot be copied over; return the previous value.
+                        return await defaultProperties.GetUnevaluatedPropertyValueAsync(propertyName);
+                    }
+                    _fileSystem.CopyFile(unevaluatedPropertyValue, destinationInfo.FullName, overwrite: true);
```
davkean, line 57: We should avoid this where possible; this is currently holding a write lock and is going to block the entire solution while this I/O occurs. It also doesn't add the file to the project nor go through source control APIs. I think you should just cut this scenario - the existing behavior is just weird and unexpected.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -0,0 +1,70 @@
...
+        public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+        {
+            // Allows the user to clear the application icon.
+            if (string.IsNullOrEmpty(unevaluatedPropertyValue))
+            {
+                return null;
+            }
+
+            // If selected path is invalid, ignore and return previous value.
+            if (!File.Exists(unevaluatedPropertyValue))
```
davkean, line 39: This is the unevaluated property value, this could be "$(FooBar)" you can't check to see if it exists.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
davkean: To see these problems; go do this:

1) `<EnableDefaultItems>false</EnableDefaultItems>`
2) Connect the project to TFS source control

davkean: Compare and contrast with legacy.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
MiYanni: This code only runs when setting this property thought the new property pages UI. It is reasonable to assume that people using the UI are not using MSBuild variables. The UI drives you toward using the `Browse` button to select a file. When used, it will insert the full filepath on disk to the page, and that will be passed into this method. So, the majority use-case is a full-path string.

Currently, there isn't an API to have a value provider set the value *after* evaluation has occurred. I'm assuming that could be added in the future. There's possibly a way to use the evaluated value Get method to do this but that's not the normal way to set property values in these new pages.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
MiYanni: Ah. I see. It adds it as a `Content` item. I can do that!
image

I talked with the team and it seems somewhat reasonable to block during the I/O since this only occurs when using the UI (prop pages) anyway. For now, it is a better solution than to have this property be broken.

I originally proposed we didn't do the file copy, and instead put a relative path to the file in as the property's value. I even tested it and everything and it worked just fine. The build .exe would pick up the new icon. But, for now, it seems the preferred fix is to keep the functionality the same as it was.

MiYanni: I've updated this to add the content item if a previous one didn't exist, or update the content item if it already exists.

jjmew (dismissed):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -73,8 +73,32 @@ internal sealed class ApplicationIconValueProvider : InterceptingPropertyValuePr
...
+            {
+                await _sourceItemsProvider.AddAsync(Content.SchemaName, propertyValue);
+            }
+
             return propertyValue;
         }
+
+        private async Task<IProjectItem?> GetExistingContentItemAsync(string existingPropertyValue)
+        {
+            foreach (IProjectItem item in await _sourceItemsProvider.GetItemsAsync(Content.SchemaName))
```
davkean, line 92: ISourceItemProvider has two methods that handle this for you; FindItemByNameAsync and GetItemAsync that takes an evaluated include.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
```diff
@@ -73,8 +73,32 @@ internal sealed class ApplicationIconValueProvider : InterceptingPropertyValuePr
                 propertyValue = PathHelper.MakeRelative(_unconfiguredProject, unevaluatedPropertyValue);
             }
 
-            await _sourceItemsProvider.AddAsync(Content.SchemaName, propertyValue);
+            string existingPropertyValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ConfigurationGeneral.ApplicationIconProperty);
+            IProjectItem? existingItem = await GetExistingContentItemAsync(existingPropertyValue);
```
davkean, line 77: I don't think ISourceItemsProvider.AddAsync will add a duplicate item for an item that already exists - have you confirmed that it does and you had to work around that?

davkean: Ah, I see you are handling the replacing part.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
MiYanni: This is a simplified version of something I did in a previous PR:
https://github.com/dotnet/project-system/blob/76e03d472b359876e5fbacb452edc3d6ca97d675/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/PackagePropertyPage/PackageIconValueProvider.cs#L60-L76

There is a `GetItemsAsync` that takes an item type and evaluated include, but not a `GetItemAsync`, so it will be very similar to what is there. And `FindItemByNameAsync` has no filter for item type, so I'd still have to check that anyway. I plan on making some shared code from my previous PRs that did similar things. But, not consolidating the code in this PR.

davkean: Can you test the following things to make sure this handles it correctly:

1) An item outside of the project file in an imports.
2) An item included as part of a glob (*.*)
3) An ApplicationIconProperty with a variable in it $(FooBar)
4) An ApplicationIconProperty with bad path characters "C:\$$$".

jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationIconValueProvider.cs
MiYanni: I've added `item.PropertiesContext.IsProjectFile` check to this so that scenarios 1 and 2 are not an issue. This exists in the code I linked above, but not in the method itself (as part of caller logic). For 3 and 4, it just reverts to the previous value (if-statement inside `OnSetPropertyValueAsync` code). I didn't support variables in the setter because it basically allows anything to be provided... which invalidates the purpose of having logic for the setter. Unless... there is a way to process an arbitrary value (like this unevaluated property) and evaluate it during the setter; even though it isn't written to the project yet?

melytc (approved):

PR #7462, opened 2021-08-05
Opt into Load context binds for package

davkean:
https://github.com/dotnet/project-system/pull/7447 opt'd Managed.VS into NGEN image.

Before this change, this dll used to be attempted to be loaded in two ways;

- Load context via the MEF cache
- LoadFrom via the [PackageRegistration] and [ProjectObject]

The latter would actually just respect the existing binary in the Load context resulting in a single load.

After NGEN, however, this changed - a Fusion limitation caused to actually load the IL image in the second step, resulting in two versions of the same assembly loaded.

This change forces the second step to use the Load context.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7462)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
@@ -7,6 +7,7 @@
     <!-- VSIX -->
     <GeneratePkgDefFile>true</GeneratePkgDefFile>
     <CreateVsixContainer>false</CreateVsixContainer>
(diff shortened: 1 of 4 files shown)


drewnoakes (approved):

swesonga (approved):

MiYanni (approved):

MiYanni:
I'm going to run a validation PR off of this PR.

tmeschter (approved):

davkean:
@MiYanni This isn't going to pass without https://devdiv.visualstudio.com/DevDiv/_git/VS/commit/acf6ad959ad16cd2cebc103d12cfad66c5759e2f?refName=refs%2Fheads%2Fdev%2Fvslsnap%2Finsertions%2Fmain.20210730230218.

PR #7465, opened 2021-08-06
Kill Hot Reload processes when requested

tmeschter:
There are going to be times when Hot Reload is in use and the process(es) need to be restarted. Maybe the user has made a code change that was applied to the process, but they don't have a good way of invoking the modified code. In that case we may provide a "Restart Hot Reload" button. Alternatively, they may make a "rude edit"--a code change that would compile, but is outside of the ability of Hot Reload to apply to a running process. In that case we show a dialog offering them the option of restarting. Either way, we need to shut down all the processes associated with a Hot Reload session and then execute a Ctrl+F5 to build and start the processes again. If we don't shut them down first, the build will fail.

The only part of this the project system is responsible for is shutting down the processes when requested. A higher-level component that is aware of all the active Hot Reload sessions and can track the status of the restart operation as a whole is responsible for the rest. Indeed, we already have a `StopProjectAsync` method, it just doesn't actually stop the process yet. Here we add that functionality.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7465)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
@@ -183,9 +183,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
 
             try
             {
-                await hotReloadState.Session.StopSessionAsync(default);
+                if (ImmutableInterlocked.TryRemove(ref _activeSessions, hotReloadState.Process.Id, out _))
+                {
+                    await hotReloadState.Session.StopSessionAsync(cancellationToken);
 
-                ImmutableInterlocked.TryRemove(ref _activeSessions, hotReloadState.Process.Id, out _);
+                    // First try to close the process nicely and if that doesn't work kill it.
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -183,9 +183,16 @@ private async Task<bool> StopProjectAsync(HotReloadState hotReloadState, Cancell
 
             try
             {
-                await hotReloadState.Session.StopSessionAsync(default);
+                if (ImmutableInterlocked.TryRemove(ref _activeSessions, hotReloadState.Process.Id, out _))
```
davkean, line 186: This performing a long running operation within a lock - should this be outside of the lock to avoid contention?

tmeschter: Which lock is that?

MiYanni (approved):

PR #7471, opened 2021-08-09
Revert "Opt into NGEN for Managed.dll and Managed.VS.dll"

drewnoakes:
This reverts commit f3351acbd822c1b852d901e52b657509cf4107b8.

While NGen is the right approach for these assemblies, it regresses performance as we don't currently have OptProf configured.

We will revisit this once we have OptProf configured for our assemblies.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7471)


swesonga (approved):
Is this the only change that needs to be reverted?

drewnoakes:
> Is this the only change that needs to be reverted?

As far as I know, yes. @davkean made another change on the VS side, but I don't believe that merged anywhere, and was only pushed to non-completed insertion PRs.

MiYanni (approved):

Issue #7318, opened 2021-06-14
Launch Profiles UI doesn't honor property visibility (2)


tmeschter:
**Steps to Reproduce**:

1. Add the following property to ProjectDebugPropertyPage.xaml:

  ``` xml
  <StringProperty Name="MyOtherProperty"
                  Visible="False" />
  ```

2. Ctrl+F5 to run VS in the experimental hive (the updated .xaml file will be picked up).

3. Create a new C# Console App targeting .NET 5 (or any .NET Core, really)

4. Open the Launch Profiles UI

**Expected Behavior**:

"MyOtherProperty" does not show in the UI.

**Actual Behavior**:

The property does in fact show up:

![image](https://user-images.githubusercontent.com/10506730/121965185-7fff8880-cd21-11eb-81e4-ab61850dfa2d.png)

**User Impact**:

Users see properties they shouldn't.

tmeschter:
Context: The WebTools team already extends this page to add several settings for their scenarios, but they also want to remove the "Use remote machine", "Remote machine name", and "Authentication mode" properties as these are not meaningful for their scenarios. This is supported, and can be accomplished by redefining the property with different settings (in this case, `Visible="False"`).

I've debugged through the back end and the Project Query API is returning the expected values, and in the front end I can see that `Property.IsVisible` returns "false" for these properties. I can't tell why the property still displayes.

drewnoakes:
This not just about "permanently" non-visible properties. The bug also relates to dynamic visibility, as seen when toggling the "Use remote debugging" check box.

drewnoakes:
Fixed in https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/340758.

PR #7478, opened 2021-08-11
Only show Enable Hot Reload setting when supported

### LINKED ISSUE #7474: Hide Hot Reload check box when Hot Reload is not supported
Currently we always show the Enable Hot Reload checkbox for launch profiles using the "Project" launch command:

image

We should only show this check box when the project supports Hot Reload; specifically, when it has the "SupportsHotReload" capability.

My current thought is to add a hidden "SupportsHotReload" bool property, and condition the visibility of the "Enable Hot Reload" property on that.

drewnoakes:
Fixes #7474.

The _Enable Hot Reload_ launch profile option will only be displayed for projects having the `SupportsHotReload` project capability defined.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7478)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
@@ -66,7 +66,13 @@
 
   <BoolProperty Name="HotReloadEnabled"
                 DisplayName="Enable Hot Reload"
-                Description="Apply code changes to the running application." />
+                Description="Apply code changes to the running application.">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(has-project-capability "SupportsHotReload")</NameValuePair.Value>
+      </NameValuePair>
+    </BoolProperty.Metadata>
+  </BoolProperty>
   
   <BoolProperty Name="NativeDebugging"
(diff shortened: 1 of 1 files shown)


drewnoakes:
I've tested that it works in the debugger, but we won't see this working in the UI until the fix for #7318 is merged.

Testing shows the option appears for both C# and VB SDK-style projects targeting .NET 6.0. Targeting earlier versions removes the capability, and hence the option from the UI.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
```diff
@@ -66,7 +66,13 @@
 
   <BoolProperty Name="HotReloadEnabled"
                 DisplayName="Enable Hot Reload"
-                Description="Apply code changes to the running application." />
+                Description="Apply code changes to the running application.">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(has-project-capability "SupportsHotReload")</NameValuePair.Value>
```
tmeschter, line 72: Oh yeah, I forgot about this one. Glad this was even more straightforward than expected.

tmeschter (approved):

drewnoakes:
> I've tested that it works in the debugger, but we won't see this working in the UI until the fix for #7318 is merged.

That has now been merged and will insert soon.

PR #7480, opened 2021-08-11
Consolidate code

tmeschter:
Consolidate most of the interesting code in `LaunchProfileByIdDataProducer`, `LaunchProfileFromProjectDataProducer` and `LaunchProfileDataProducer` into a new type, `ProjectLaunchProfileHandler`.

The first two types do a lot of similar operations to create entities for launch profiles, suggesting that their functionality should be consolidated into the same type. And the third type only exist because similar functionality is split between the other two.

Also, there were a lot of calls to the project's `ExportProvider` to retrieve `UnconfiguredProject`-level services, and any failure to obtain a service was silently ignored. The new `ProjectLaunchProfileHandler` type also exists within the `UnconfiguredProject` scope, allowing us to import them into the constructor and turning any failure to obtain the other services into a MEF error.

Mostly this change is just moving code around. However, there is a third location where we create new launch profile entities: `LaunchProfileActionBase`. And unfortunately it works differently than the other types. The goal is to eventually consolidate that path in here as well.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7480)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
@@ -2,15 +2,10 @@
 
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Build.Framework.XamlTypes;
-using Microsoft.VisualStudio.ProjectSystem.Debug;
-using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.ProjectSystem.Query;
 using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModel;
 using Microsoft.VisualStudio.ProjectSystem.Query.QueryExecution;
-using Microsoft.VisualStudio.ProjectSystem.VS.Utilities;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 {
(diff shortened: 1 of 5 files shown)


jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
```diff
@@ -25,64 +20,12 @@ public LaunchProfileFromProjectDataProducer(ILaunchProfilePropertiesAvailableSta
...
+                return launchProfileHandler.RetrieveAllLaunchProfileEntitiesAsync(queryExecutionContext, parent, _properties);
             }
-        }
-
-        private IEntityValue CreateLaunchProfileValue(IQueryExecutionContext queryExecutionContext, IEntityValue parent, QueryProjectPropertiesContext propertiesContext, Rule rule, int order, IProjectState projectState)
-        {
-            EntityIdentity identity = LaunchProfileDataProducer.CreateLaunchProfileId(parent, propertiesContext.ItemType!, propertiesContext.ItemName!);
 
-            return LaunchProfileDataProducer.CreateLaunchProfileValue(queryExecutionContext, identity, propertiesContext, rule, order, projectState, _properties);
+            return Task.FromResult(Enumerable.Empty<IEntityValue>());
```
drewnoakes, line 28: This allocation can be avoided via the change in PR #7482. @tmeschter if you want to make that change here while I'm offline, feel free to merge that PR.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/LaunchProfileFromProjectDataProducer.cs
tmeschter: I'll pick that up in my next change.

PR #7484, opened 2021-08-12
Use ConfiguredProjectDataSourceJoinBlock

### LINKED ISSUE #7483: Introduce ConfiguredProjectDataSourceJoinBlock
Currently we use `UnwrapCollectionChainedProjectValueDataSource` to aggregate data across project configurations. Its API is a bit hard to understand.

We should use the newer `ConfiguredProjectDataSourceJoinBlock` instead.

Added in CPS PR: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/329048

drewnoakes:
Fixes #7483

`ConfiguredProjectDataSourceJoinBlock` is a new Dataflow block provided by CPS that simplifies code that previously used `UnwrapCollectionChainedProjectValueDataSource`.

cc @lifengl, @davkean

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7484)


davkean:
I think this looks good, but can you get @lifengl to sign off? This is much nicer.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
```diff
@@ -38,37 +38,26 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreInfo> targe
             // (via ProjectRestoreUpdate) and combine it into a single IVsProjectRestoreInfo2 instance and publish that. When a change is 
             // made to a configuration, such as adding a PackageReference, we should react to it and push a new version of our output. If the 
             // active configuration changes, we should react to it, and publish data from the new set of implicitly active configurations.
```
MiYanni, line 40: Are the comments here and above still applicable due to the changes in which dataflow block is used? If so, it sounds like it should be above the method itself, not inside of it.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
```diff
@@ -38,37 +38,26 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreInfo> targe
...
 
-            // Join the source blocks, so if they need to switch to UI thread to complete 
-            // and someone is blocked on us on the same thread, the call proceeds
-            JoinUpstreamDataSources(restoreConfiguredInputSource, activeConfiguredProjectsSource);
+                // Link the active configured projects to our join block
+                _activeConfigurationGroupService.ActiveConfiguredProjectGroupSource.SourceBlock.LinkTo(joinBlock, DataflowOption.PropagateCompletion),
 
-            return disposables;
+                // Set the link up so that we publish changes to target block
+                mergeBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion)
```
lifengl, line 59: usually, we don't need dispose those links other than the joinBlock & the top link to the ProjectGroupSource, because it will release mergeBlock block from memory (it is only held by the its source block). of course, it doesn't hurt much by holding the link in the disposable bug.

lifengl (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
drewnoakes: Yes the comment is still valid. It's explaining what's happening, rather than how it happens. We just change the how here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
```diff
@@ -38,37 +38,26 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreInfo> targe
...
+            return new DisposableBag
+            {
+                joinBlock,
+                mergeBlock,
 
-            // Join the source blocks, so if they need to switch to UI thread to complete 
-            // and someone is blocked on us on the same thread, the call proceeds
-            JoinUpstreamDataSources(restoreConfiguredInputSource, activeConfiguredProjectsSource);
+                // Link the active configured projects to our join block
+                _activeConfigurationGroupService.ActiveConfiguredProjectGroupSource.SourceBlock.LinkTo(joinBlock, DataflowOption.PropagateCompletion),
```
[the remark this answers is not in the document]
drewnoakes: Thanks.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
drewnoakes: @lifengl I don't see why the merge block could be freed. Isn't it participating in the data flow?

PR #7486, opened 2021-08-13
Consolidate LaunchSettingsActionService

tmeschter:
Consolidate `LaunchSettingsActionService` into the new `ProjectLaunchProfileHandler`. There are two considerations here:

1. The `LaunchSettingsActionService` is one of the three places we create a new entity for a launch profile. The other two are already in `ProjectLaunchProfileHandler`, and this third one worked differently than the others for no particularly good reason.
2. There are just too many types to keep track of, both interfaces and implementations.

With this change, `ProjectLaunchProfileHandler` assumes responsibility for the most important aspects of manipulating launch profiles and the single point where Project Query API entities representing launch profiles are created. This also allows us to impose some consistency across the methods for adding, removing, and retrieving these entities (e.g. they all take an `IQueryExecutionContext` as the first parameter).

Another nice aspect of this change is that the implementations of `LaunchProfileActionBase` no longer deal with the `ILaunchProfile` type directly; rather, they now deal with entity IDs and values exclusively. When creating or removing a profile, the `ProjectLaunchProfileHandler` returns the _ID_ of the affected entity rather than the `ILaunchProfile` or entity itself; later we go back and use the ID to create the entity to return. This is also better for sitiuations where a query both creates a profile and manipulates it in some way as we will only create an entity for the final state of the profile.

There is a bit more to be done here. The entity we return for newly created launch profiles will contain the profile name, command name, and order, but _not_ the categories or properties. This is because categories and properties are themselves entities (as are property values). In a query that merely _retrieves_ entities the query engine handles invoking the appropriate data producers to fill in all the entities and child entities based on the properties requested in the query. However, an _update_ query currently provides no such help and we will have to perform this population ourselves. Based on how the add/remove/duplicate/rename operations are currently used by the UI it should be sufficient to provide the categories and properties (but not the property values).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7486)


jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/IProjectLaunchProfileHandler.cs
```diff
@@ -24,5 +25,115 @@ internal interface IProjectLaunchProfileHandler
...
+        /// <returns>
+        /// The renamed <see cref="ILaunchProfile"/>, or <see langword="null"/> if the
+        /// profile could not be renamed. This may occur if the profile specified by <paramref name="currentProfileName"/>
+        /// cannot be found.
+        /// </returns>
+        /// <remarks>
+        /// If <paramref name="newProfileName"/> is the name of an existing launch profile,
+        /// the existing profile will be replaced by the newly-renamed profile.
+        /// </remarks>
+        Task<(EntityIdentity, EntityIdentity)?> RenameLaunchProfileAsync(IQueryExecutionContext queryExecutionContext, IEntityValue parent, string currentProfileName, string newProfileName, CancellationToken cancellationToken = default);
```
drewnoakes, line 137: Using named elements on the returned tuple would be helpful.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/IProjectLaunchProfileHandler.cs
```diff
@@ -24,5 +25,115 @@ internal interface IProjectLaunchProfileHandler
...
+        /// The new name to give to the profile.
+        /// </param>
+        /// <param name="cancellationToken">
+        /// A token whose cancellation indicates that the caller no longer is interested in
+        /// the result.
+        /// </param>
+        /// <returns>
+        /// The renamed <see cref="ILaunchProfile"/>, or <see langword="null"/> if the
+        /// profile could not be renamed. This may occur if the profile specified by <paramref name="currentProfileName"/>
+        /// cannot be found.
```
drewnoakes, line 131: This comment doesn't explain why two values are returned.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/IProjectLaunchProfileHandler.cs
```diff
@@ -24,5 +25,115 @@ internal interface IProjectLaunchProfileHandler
...
+        /// The entity representing the project.
+        /// </param>
+        /// <param name="profileName">
+        /// The name of the profile to remove.
+        /// </param>
+        /// <param name="cancellationToken">'
+        /// A token whose cancellation indicates that the caller no longer is interested in
+        /// the result.
+        /// </param>
+        Task<EntityIdentity?> RemoveLaunchProfileAsync(IQueryExecutionContext queryExecutionContext, IEntityValue parent, string profileName, CancellationToken cancellationToken = default);
```
drewnoakes, line 107: Missing `<returns>`.

drewnoakes (approved):

Issue #7373, opened 2021-07-01
Settings and Resources editor discoverability (1+)


drewnoakes:
Historically the "Application Designer" included not just project properties but other kinds of configuration and data such as:

- Launch profiles
- Resources
- Settings

With the new Project Properties UI, these are no longer integrated into that experience.

Some customers will not know how to find these editors via other paths.

For resources/settings, we no longer have a "click here to create one" experience, and users must use the "New Item" experience instead. Can we do something about this?

For launch profiles, we have added a hyperlink to more information (see also #7337).

For settings and resources, similar guidance does not exist. In these two cases, we should also provide guidance for the case that the relevant files are not present in the project, ideally making it a single-click to add each of them them.

drewnoakes:
Notes for triage team: Adding textual information is straightforward, hence base estimate of 1. Adding hyperlinks gets more expensive, hence estimate of 1+. Please also decide if this is a `must-have` issue for 17.0.

drewnoakes:
@mckennabarlow can you think through how we want to surface this to the user, and what text we want to display? We need to get this text in before the localisation cut-off.

drewnoakes:
See also CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/344654

drewnoakes:
CPS PR has merged.

PR #7489, opened 2021-08-16
Add properties for adding resources/settings files

drewnoakes:
Contributes to https://github.com/dotnet/project-system/issues/7373

The legacy project properties pages embedded the assembly resources and settings files editors in the "application designer". The new UI takes a property-centric view of the project, and delegates these editors to their own standalone experience.

To help users in this transition, we want a link action in the UI that does two things:

1. Create the file if it does not currently exist
1. Opens the file's editor

CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/344654 adds link action handlers for resources and settings files. This change adds pages and properties to surface this to the user.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7489)


drewnoakes:
This should only be merged once the following have inserted:

- https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/344654
- https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/344681

jjmew (approved):
:shipit:

melytc (approved):

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ResourcesPropertyPage.xaml
```diff
@@ -0,0 +1,36 @@
...
+      xmlns="http://schemas.microsoft.com/build/2009/properties">
+
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFileWithInterception"
+                SourceOfDefaultValue="AfterContext"
+                HasConfigurationCondition="False" />
+  </Rule.DataSource>
+
+  <StringProperty Name="ResourcesPagePlaceholderDescription"
+                  DisplayName="Ignored"
```
MiYanni, line 17: Is this a special value to make it not use DisplayName? Interesting that something like empty string or excluding the attribute entirely wouldn't do the same thing.

MiYanni (approved):
So after this PR and the CPS PR, will there be a follow-up to add the link handlers? Right now, these are no-op handlers.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ResourcesPropertyPage.xaml
drewnoakes: No it's not special. Just capturing the fact that it's unused in any way. Without this `Name` is used in place of `DisplayName`. Omitting `DisplayName` altogether would avoid translating this string, but we have unit tests that verify all visible properties have a `DisplayName` value.

PR #7496, opened 2021-08-18
Fix a silly bug

tmeschter:
Fix a bug where I'm failing to await the result of an asynchronous
operation so the subsequent pattern match will always fail. The end
result is that the update query will not actually be updated with any
new launch profiles.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7496)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
@@ -40,7 +40,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
             s_requestedLaunchProfileProperties.Freeze();
         }
 
-        public Task OnRequestProcessFinishedAsync(IQueryProcessRequest request)
+        public async Task OnRequestProcessFinishedAsync(IQueryProcessRequest request)
         {
             foreach (IGrouping<IEntityValue, (IEntityValue projectEntity, EntityIdentity profileId)> group in AddedLaunchProfiles.GroupBy(item => item.projectEntity))
             {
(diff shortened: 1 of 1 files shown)


swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
```diff
@@ -56,7 +56,7 @@ public Task OnRequestProcessFinishedAsync(IQueryProcessRequest request)
 
                     foreach ((IEntityValue projectEntity, EntityIdentity addedProfileId) in group)
                     {
-                        if (handler.RetrieveLaunchProfileEntityAsync(request.QueryExecutionContext, addedProfileId, s_requestedLaunchProfileProperties) is IEntityValue launchProfileEntity)
+                        if (await handler.RetrieveLaunchProfileEntityAsync(request.QueryExecutionContext, addedProfileId, s_requestedLaunchProfileProperties) is IEntityValue launchProfileEntity)
```
sharwell, line 59: This is why `{ }` should be used in non-null check pattern matching.

```suggestion
                        if (await handler.RetrieveLaunchProfileEntityAsync(request.QueryExecutionContext, addedProfileId, s_requestedLaunchProfileProperties) is { } launchProfileEntity)
```

swesonga: @sharwell what is this operator called? (is it an operator)?

sharwell: I think it's just called the empty pattern. @jaredpar ?

MiYanni: Since when can empty braces be used as a type? How does it know to present itself as `IEntityValue`?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
sharwell: > Since when can empty braces be used as a type?

Property patterns were added in C# 8.

> How does it know to present itself as IEntityValue?

That's the result type for the `await` operation. The pattern syntax just adds a null check to it.

tmeschter: Changing the patterns we use has implications outside of this one specific example, so that's something for a different change.

PR #7497, opened 2021-08-18
Don't enable Hot Reload when remote debugging

tmeschter:
Currently we enable Hot Reload for both local and remote debugging. While it is not infeasible to make it work for remote debugging, our tooling does not currently support this. I doubt the Hot Reload transport layer currently supports moving the IL and metadata changes to a remote system, but more to the point for this bug we end up setting environment variables on the remote process with file paths that only make sense on the local machine. This causes the remote process to fail immediately as it isn't able to find the relevant file.

The fix here is to disable Hot Reload when remote debugging.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7497)


swesonga (approved):
Should all these conditions be in a separate method like ShouldHotReloadBeEnabled(resolvedProfile)?

tmeschter:
> Should all these conditions be in a separate method like ShouldHotReloadBeEnabled(resolvedProfile)?

Reasonable.

PR #7498, opened 2021-08-19
Hide the Hot Reload check box when Remote Debugging is checked

drewnoakes:
Follows #7497 where this same check is made in the project launch target provider.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7498)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
@@ -69,7 +69,10 @@
                 Description="Apply code changes to the running application.">
     <BoolProperty.Metadata>
       <NameValuePair Name="VisibilityCondition">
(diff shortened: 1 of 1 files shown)


tmeschter:
I think a better approach would be to conditionally disable the Hot Reload control rather than hide it, if we had that ability. That at least hints to the user that Hot Reload is _in general_ available but disabled for some other reason.

MiYanni:
@tmeschter Do we have that as a design concept for the new property pages (situationally disabled controls)? Maybe an alternative design is to show some text about Hot Reload when it cannot be used (different XAML element with opposite conditions), and then disable that text and enable the control when it can be used.

drewnoakes:
We have no support for disabling controls currently. Let's discuss this at our next meeting.

If we had and used that here, we would also need to uncheck hot reload when remote debugging was checked.

My vote is to merge this as is for now. It's consistent with how we handle other such features.

tmeschter (approved):

tmeschter:
Yeah, I don't want to hold this up--it just points to a general issue that we should tackle at some point.

MiYanni (approved):

PR #7500, opened 2021-08-19
Populate categories and properties

tmeschter:
When running a query to retrieve data, the query engine tells us which properties to include on the created entity, and also handles calling into the appropriate data producers to create the requested child entities (if any) _and_ hooks them up to them parent.

When running an update query that may create a new entity (such as when adding a launch profile), however, we don't know which properties and sub-entities to create. This is a limitation of the current Project Query API.

Currently when we create an entity for a new launch profile we just populate all the "simple" properties like name, command name, and order. Here we add support for adding all the related entities representing categories and properties, and the simple properties of both. Note this does not include any information about the editor, value, or supported values of a property.

In the absence of information about what was requested in the original property, our goal here is to provide the minimum set of data the UI will need after an update query has finished.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Collections.Immutable;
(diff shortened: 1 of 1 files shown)


tmeschter:
@drewnoakes If you need to start examining the updated query results (rather than waiting for them to come through the subscription) does this provide enough information? Or do you need more than the entities for the launch profile, its categories, and properties? E.g. do you need the property's editors or values?

drewnoakes:
We would need all the data. Consider adding a profile: the UI needs everything in one shot to populate the itself correctly.

I'll take a deeper look at this today or on my Monday.

tmeschter:
I see. I thought maybe you just needed some of the metadata, and could let the rest come in with the call to OnNext.

It is possible to provide everything--it's just messy as we need to recurse into all the child entities and populate them.

tmeschter:
Note that this is an intermediate step in getting to all the data so it can be reviewed and merged while I work on the rest.

tmeschter:
Alright, this is now populating all the data.

Issue #7502, opened 2021-08-20
Replace MinimumPlatformVersion  with SupportedOSPlatformVersion


drewnoakes:
Reported by @dplaisted in https://github.com/dotnet/project-system/pull/7132#issuecomment-902338688:

> It looks like the SDK no longer uses MinimumPlatformVersion version at all. So I think this UI needs to be updated to use SupportedOSPlatformVersion. @melytc

PR #7503, opened 2021-08-20
Change MinimumPlatformVersion to SupportedOSPlatformVersion.

melytc:
Addresses #7502.

The SDK now uses the `SupportedOSPlatformVersion` property. These changes reflect this in the project properties UI and saves it correctly in the project file.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -143,9 +143,9 @@
     </DynamicEnumProperty.Metadata>
   </DynamicEnumProperty>
 
-  <DynamicEnumProperty Name="MinimumPlatformVersion"
-                DisplayName="Minimum OS version"
-                Description="Specifies the minimum runtime version of the selected operating system."
+  <DynamicEnumProperty Name="SupportedOSPlatformVersion"
+                DisplayName="Supported OS version"
+                Description="Specifies the minimum version of the OS that an app or library supports running on. If it is not specified, it should default to the version equivalent to the Target OS version value."
                 Category="General"
                 EnumProvider="SdkSupportedTargetPlatformVersionEnumProvider">
     <DynamicEnumProperty.DataSource>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
@@ -27,7 +27,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         private const string TargetPlatformProperty = ConfigurationGeneral.TargetPlatformIdentifierProperty;
         private const string TargetPlatformVersionProperty = ConfigurationGeneral.TargetPlatformVersionProperty;
         private const string TargetFrameworkProperty = ConfigurationGeneral.TargetFrameworkProperty;
-        private const string MinimumPlatformVersionProperty = "MinimumPlatformVersion";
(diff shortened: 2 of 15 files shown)


jjmew (approved):
:shipit:

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -143,9 +143,9 @@
     </DynamicEnumProperty.Metadata>
   </DynamicEnumProperty>
 
-  <DynamicEnumProperty Name="MinimumPlatformVersion"
-                DisplayName="Minimum OS version"
-                Description="Specifies the minimum runtime version of the selected operating system."
+  <DynamicEnumProperty Name="SupportedOSPlatformVersion"
```
drewnoakes, line 146: I can see how this is a potentially confusing property. Do we have docs for this? The best I could find are:

https://docs.microsoft.com/dotnet/standard/frameworks#os-version-in-tfms

melytc: Agree, maybe @dsplaisted or @mhutch might know a docs to redirect users about its use.

dsplaisted: @terrajobst was also involved in this design.

We should have documentation for this, but I don't know if we do yet.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -143,9 +143,9 @@
...
+                DisplayName="Supported OS version"
+                Description="Specifies the minimum OS version that the project will run on. When unspecified, the Target OS version value is implied. Using an earlier version here requires code to add guards around later version APIs."
```
tmeschter, line 148: Is "Target" meant to be capitalized?

More importantly: good addition to the description. 👍🏻

tmeschter (approved):

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
jjmew, line 146: @dsplaisted  Does MinimumPlatformVersion exist in .NET 5?

dsplaisted: I don't think so.  I think that it should only be displayed or enabled in the UI when there is a target platform in the TargetFramework.  IE it shouldn't show up for `net6.0` either, but it should show up for `net6.0-windows10.0.17763`, `net6.0-android`, etc.

melytc: There's a condition visibility below that indicates the `SupportedOSPlatformVersion` should be shown when it's a .NET 5 project or greater and when there is a value in the `TargetPlatformIdentifier`, so I think we got that covered.

dsplaisted: > .NET 5 project or greater

I think it should probably be checking for .NET 6 or greater, though I think a .NET 5 project might not have a `TargetPlatformIdentifier` normally, so it may be moot.

mhutch: >I think a .NET 5 project might not have a TargetPlatformIdentifier normally, so it may be moot.

`net5.0-windows` projects would be able to set a property that didn't do anything? Not a big deal, but it would be safest to check for net6.0.

PR #7504, opened 2021-08-23
Use JoinableTaskContext in SequentialTaskExecutor

### LINKED ISSUE #3853: SequentialTaskExecutor violates threading rule 3, causing deadlocks.
[SequentialTaskExecutor.cs](https://github.com/dotnet/project-system/blob/f2118e1d747f006628729de22b901b908fa0722f/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs) uses `Task.ContinueWith` to control the execution flow of multiple tasks. This is a violation of [threading rule 3](https://github.com/Microsoft/vs-threading/blob/master/doc/threading_rules.md#rule-3-if-ever-awaiting-work-that-was-started-earlier-that-work-must-be-joined), as a posted task to `ExecuteAsync` can wait on the completion of a previous, unrelated call to `ExecuteAsync`. Docker tools has encountered this hang and supplied a dump, which I investigated and found this to be the root cause.

The fix is for `SequencialTaskExecutor` to have a private `JoinableTaskFactory` and `JoinableTaskCollection`. All work supplied to `ExecuteAsync` should be ran with the `JTF.RunAsync`, and when work must be posted inside the `ContinueWith`, the collection must be joined so that the previous enqueued tasks are joined. Additionally, reconsider the usage of `ContinueWith` chains, as if `ExecuteAsync` is called a lot, it may cause memory pressure.

drewnoakes:
Fixes #3853. See that issue for a description of the problem and its resolution.

This change will contribute towards [AB#1370097](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1370097)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7504)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs
@@ -20,10 +20,10 @@ namespace Microsoft.VisualStudio.Threading.Tasks
         private readonly JoinableTaskCollection _joinableCollection;
         private readonly JoinableTaskFactory _joinableFactory;
 
-        public SequentialTaskExecutor(JoinableTaskContextNode joinableTaskContext)
+        public SequentialTaskExecutor(JoinableTaskContextNode joinableTaskContext, string displayName)
         {
             _joinableCollection = joinableTaskContext.CreateCollection();
-            _joinableCollection.DisplayName = nameof(SequentialTaskExecutor);
+            _joinableCollection.DisplayName = $"{nameof(SequentialTaskExecutor)}_{displayName}";
             _joinableFactory = joinableTaskContext.CreateFactory(_joinableCollection);
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs
@@ -17,9 +17,18 @@ namespace Microsoft.VisualStudio.Threading.Tasks
         private Task _taskAdded = Task.CompletedTask;
         private readonly object _syncObject = new();
         private readonly CancellationTokenSource _disposedCancelTokenSource = new();
+        private readonly JoinableTaskCollection _joinableCollection;
+        private readonly JoinableTaskFactory _joinableFactory;
+
+        public SequentialTaskExecutor(JoinableTaskContextNode joinableTaskContext)
+        {
+            _joinableCollection = joinableTaskContext.CreateCollection();
+            _joinableCollection.DisplayName = nameof(SequentialTaskExecutor);
+            _joinableFactory = joinableTaskContext.CreateFactory(_joinableCollection);
+        }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs
```diff
@@ -17,9 +17,18 @@ internal sealed class SequentialTaskExecutor : IDisposable
         private Task _taskAdded = Task.CompletedTask;
         private readonly object _syncObject = new();
         private readonly CancellationTokenSource _disposedCancelTokenSource = new();
+        private readonly JoinableTaskCollection _joinableCollection;
+        private readonly JoinableTaskFactory _joinableFactory;
+
+        public SequentialTaskExecutor(JoinableTaskContextNode joinableTaskContext)
+        {
+            _joinableCollection = joinableTaskContext.CreateCollection();
+            _joinableCollection.DisplayName = nameof(SequentialTaskExecutor);
```
tmeschter, line 26: What about including the display name as a parameter to the constructor? If we're going to use `SequentialTaskExecutor` in multiple scenarios it may be helpful to distinguish one usage from another.

tmeschter (approved):

tmeschter:
I've approved this because I trust that you know what you're doing and I want to make sure other work is unblocked, but this seems like an advanced use of the Microsoft.VisualStudio.Threading library and I want to make sure I understand it fully as well.

Can you explain the role of each of the following elements?

1. The `JoinableTaskContextNode` passed to the `SequentialTaskExecutor` constructor
2. The `JoinableTaskCollection` created from the `JoinableTaskContextNode`
3. The `JoinableTaskFactory` created from the `JoinableTaskContextNode`
4. The call to `JoinableTaskCollection.Join` before awaiting the async `Func`

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/SequentialTaskExecutor.cs
drewnoakes: Good idea. Done.

drewnoakes:
> this seems like an advanced use of the Microsoft.VisualStudio.Threading library

Agreed. This pattern had been taken from CPS's [`TreePublishingBatchService`](https://devdiv.visualstudio.com/DevDiv/_git/CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.VS.Implementation%2FPackage%2FNodes%2FTreePublishingBatchService.cs&version=GBmain&line=77&lineEnd=123&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), which aims to achieve the same thing as we do here&mdash;to serialize a sequence of work scheduled asynchronously, preserving order, preventing overlap and supporting cancellation. The implementation also matches the description of the fix laid out in #3853.

@lifengl, I'd appreciate if you could lend an eye to this change, as you're the resident vs-threading expert. Perhaps we can go over it in tomorrow's CPS sync meeting.

PR #7506, opened 2021-08-24
Serialize language service workspace updates

drewnoakes:
Fixes AB#1390097

Includes #7504. Only review the last commit.

Roslyn internally uses blocking locks to protect access to its solution-level workspace data. When opening a solution with many projects, each project make requests to update workspaces. Those updates pass through this interface so that we can serialize them in order to reduce blocking lock contention on the thread pool from concurrent update requests.


jjmew (approved):
:shipit:

drewnoakes:
Converting to draft following discussion with @jasonmalinowski about potentially making `EndBatch` async. See also https://github.com/dotnet/roslyn/pull/53768.

PR #7509, opened 2021-08-25
Add 'Advanced' section to build property page

### LINKED ISSUE #7186: New prop pages - consider Advanced subgrouping for the truly advanced C# features
This one feels _somewhat_ like a "we're going back to what we had" thing, but the following C# properties are quite advanced and are unlikely to every be modified by people:

image

(Note that I checked the arithmetic overflow one myself).

Maybe they should just be moved under a subtitle called "Advanced". Thoughts?

drewnoakes:
Fixes #7186

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7509)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -26,6 +26,10 @@
     <Category Name="StrongNaming"
               Description="Configures strong name signing of build outputs."
               DisplayName="Strong naming" />
+    
+    <Category Name="Advanced"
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -149,75 +153,6 @@
                DisplayName="Embedded in DLL/EXE, portable across platforms" />
   </EnumProperty>
 
-  <StringProperty Name="LangVersion"
```
drewnoakes, line 152: This big diff is just the moving of these properties to later in the file and setting the category to 'Advanced'.

PR #7511, opened 2021-08-25
Show the language version

### LINKED ISSUE #7185: New prop pages - consider removing Language version
I think that we should just remove this section:

image

The policy is pretty well-understood now, and has been this way for two .NET releases. It's not actionable.

drewnoakes:
Fixes #7185

Previously, for language version, we showed a link to an explanation of why the language version could not be modified. #7185 tracks removal of this link, and suggests showing the language version instead.

This change moves the link to the help button, and displays the evaluated language version(s) based on target framework(s) in the UI. These remain read-only values.

For a .NET 6 project:

image

This change is particularly helpful for multi-targeting projects as it shows how the language level varies by target:

image


drewnoakes:
Requires CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/346712 to work correctly, though will not break if merged before that, as editor metadata can be safely ignored.

PR #7532, opened 2021-08-28
add code analysis property page

jmarolf:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7532)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/CodeAnalysisPropertyPage.xaml
```diff
@@ -0,0 +1,88 @@
...
+                HasConfigurationCondition="False" />
+  </Rule.DataSource>
+
+  <BoolProperty Name="RunAnalyzersDuringBuild"
+                DisplayName="Run on build"
+                Description="Always run code analysis on build."
+                HelpUrl="https://docs.microsoft.com/visualstudio/code-quality/disable-code-analysis"
+                Category="AllAnalyzers">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="SearchTerms" Value="analyzer;code analyzer;analysis;code analysis" />
```
drewnoakes, line 30: These search terms don't add anything. We already include the property name, display name, description, possible values and current values when computing search results.

Search terms only really make sense when there's a common alias for some property that does not appear anywhere in the text of the property. For example "pointer" for unsafe code.

The same is true for all other search terms here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/CodeAnalysisPropertyPage.xaml
```diff
@@ -0,0 +1,88 @@
...
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/code-styles-and-code-cleanup"
+                Category="NetAnalyzers">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="SearchTerms" Value="style;code style" />
+    </BoolProperty.Metadata>
+  </BoolProperty>
+
+  <BoolProperty Name="EnableNETAnalyzers"
+                DisplayName="Enable .NET analyzers"
+                Description="Run .NET analyzers to help with API useage."
```
drewnoakes, line 56:
```suggestion
                Description="Run .NET analyzers to help with API usage."
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/CodeAnalysisPropertyPage.xaml
```diff
@@ -0,0 +1,88 @@
...
+  </BoolProperty>
+
+  <EnumProperty Name="AnalysisLevel"
+                DisplayName="Analysis level"
+                Description="The set of analyzers that should be run in the project."
+                HelpUrl="https://docs.microsoft.com/dotnet/fundamentals/code-analysis/overview"
+                Category="NetAnalyzers">
+    <EnumProperty.DataSource>
+      <DataSource HasConfigurationCondition="False"
+                  Persistence="ProjectFileWithInterception" />
```
drewnoakes, line 71: I don't see an interceptor for this property.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/CodeAnalysisPropertyPage.xaml
```diff
@@ -0,0 +1,88 @@
...
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFile"
+                SourceOfDefaultValue="AfterContext"
+                HasConfigurationCondition="False" />
+  </Rule.DataSource>
+
+  <BoolProperty Name="RunAnalyzersDuringBuild"
+                DisplayName="Run on build"
+                Description="Always run code analysis on build."
+                HelpUrl="https://docs.microsoft.com/visualstudio/code-quality/disable-code-analysis"
```
drewnoakes, line 27: Tracking adding fwlinks for these `HelpUrl`s in https://github.com/dotnet/project-system/issues/7535.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/CodeAnalysisPropertyPage.xaml
```diff
@@ -0,0 +1,88 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
+<Rule Name="CodeAnalysisPropertyPage"
```
drewnoakes, line 3: Should the visibility of any of the properties on this page be conditional on the state of others? For example, a property might only be valid and therefore visible when a check box is checked.

drewnoakes:
Have you tested this? I see this is based off the `dev17.0x` branch, but that is very old and inactive. When I run that branch locally it fails to run. I think this should be based off `main`. I don't know if you have a good reason for targeting that branch, so rather than mess with your PR, I'll submit an alternative PR to this one.

Also, this doesn't take into account the `RunAnalyzers` property, which overrides both `RunAnalyzersDuringBuild` and `RunAnalyzersDuringLiveAnalysis`. If `RunAnalyzers` is set in the project (to either true or false) then attempting to check or uncheck other check boxes will have no effect.

Possible fixes for this:

- Intercept writes to `RunAnalyzersDuringBuild` and `RunAnalyzersDuringLiveAnalysis` and clear any value of `RunAnalyzers`.
- Hide the `RunAnalyzersDuringBuild` and `RunAnalyzersDuringLiveAnalysis` properties when `RunAnalyzers` has a non-empty value.
- Introduce a combo-box with options _Never run_, _On build only_, Live analysis only_, _Build and live analysis_.
- Possibly something else.

drewnoakes:
If the project targets .NET Framework (e.g. `net48`) then the _Analysis level_ combo box appears empty. Should it be defaulted to _None_? The docs on this are unclear to me:

https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysislevel

drewnoakes:
Should we add a property for `<AnalysisMode>` too? https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode

If it is only applicable to .NET 5 and later, we can hide it when targeting earlier versions.

tmeschter:
Bumping out to 17.1 on the expectation that we will not be able to get this localized in time for 17.0.

drewnoakes:
Merged in #7538.

PR #7539, opened 2021-08-30
Add runtime project properties section

drewnoakes:
This PR adds a new "Runtime" page containing properties that exist for .NET Core Applicaiton projects. Library projects, or those targeting .NET Framework would not see this page at all.

For more information on these properties, see https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#runtime-configuration-properties.

image

---

It seems that the SDK does not specify default values for several of these properties, hence I am leaving this as a draft for now to open a discussion around whether these would add value.


drewnoakes:
cc @BillWagner for input on whether these properties would add value, and whether we could rely on the SDK to provide default values to make the experience better here.

tmeschter:
What percentage of our users will ever need to change one of these settings? If it is very small, we shouldn't complicate the user experience by adding them.

MiYanni:
I see this as a good candidate of properties for the Advanced page. There would just be a Runtime section and none of the sub-sections. That's my suggestion.

drewnoakes:
Rather than an advanced page, I think we need an advanced mode. Lets discuss this in #7558.

MiYanni:
@drewnoakes Is this something still under active consideration? Might make sense to close it for the time being until we focus our attention again on property page changes.

PR #7543, opened 2021-08-31
Implement IProjectConfigurationDimensionProvider5

arkalyanms:
This implements the project configuration dimension acquisition from MsBuild.Project directly. This will be used immediately in the project evaluation cache invalidation scenario where we only have the MsBuild.Project in some code paths.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7543)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
@@ -6,6 +6,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
 using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.Build;
 
@@ -14,7 +15,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
     /// <summary>
     /// Base project configuration dimension provider
     /// </summary>
-    internal abstract class BaseProjectConfigurationDimensionProvider : IProjectConfigurationDimensionsProviderInternal
+    internal abstract class BaseProjectConfigurationDimensionProvider : IProjectConfigurationDimensionsProvider5, IProjectConfigurationDimensionsProviderInternal
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="BaseProjectConfigurationDimensionProvider"/> class.
@@ -72,6 +73,30 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
                     // Need evaluated property to get inherited properties defines in props or targets.
                     return evaluatedProject.GetProperty(PropertyName)?.EvaluatedValue;
                 });
+            }
+        }
+
+        /// <summary>
+        /// Gets the property values for the dimension.
+        /// </summary>
+        /// <param name="project"><see cref="Project"/>.</param>
+        /// <returns>Collection of values for the dimension.</returns>
+        /// <remarks>
+        /// From <see cref="IProjectConfigurationDimensionsProvider"/>.
+        /// </remarks>
+        private ImmutableArray<string> GetOrderedPropertyValues(Project project)
+        {
+            Requires.NotNull(project, nameof(project));
+
+            string? propertyValue = project.GetProperty(PropertyName)?.EvaluatedValue;
+
+            if (Strings.IsNullOrEmpty(propertyValue))
+            {
+                return ImmutableArray<string>.Empty;
+            }
+            else
+            {
+                return BuildUtilities.GetPropertyValues(propertyValue).ToImmutableArray();
             }
         }
 
@@ -203,6 +228,28 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
                     });
                 });
             }
+        }
+
+        public virtual async Task<IEnumerable<KeyValuePair<string, IEnumerable<string>>>> GetProjectConfigurationDimensionsAsync(Project project)
+        {
+            Requires.NotNull(project, nameof(project));
+
+            ImmutableArray<string> values = GetOrderedPropertyValues(project);
+            if (values.IsEmpty)
+            {
+                return ImmutableArray<KeyValuePair<string, IEnumerable<string>>>.Empty;
+            }
+            else
+            {
+                var dimensionValues = PooledArray<KeyValuePair<string, IEnumerable<string>>>.GetInstance();
+                dimensionValues.Add(new KeyValuePair<string, IEnumerable<string>>(DimensionName, values));
+                return dimensionValues.ToImmutableAndFree();
+            }
+        }
+
+        public async Task<IEnumerable<KeyValuePair<string, string>>> GetBestGuessDefaultValuesForDimensionsAsync(UnconfiguredProject project, string solutionConfiguration)
+        {
+            return await GetBestGuessDefaultValuesForDimensionsAsync(project);
         }
 
         private async Task<string?> FindDefaultValueFromDimensionPropertyAsync(UnconfiguredProject project)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Build.Construction;
 using Microsoft.Build.Evaluation;
-using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.Build;
 
 namespace Microsoft.VisualStudio.ProjectSystem.Configuration
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
@@ -15,7 +15,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
     /// <summary>
     /// Base project configuration dimension provider
     /// </summary>
-    internal abstract class BaseProjectConfigurationDimensionProvider : IProjectConfigurationDimensionsProvider5, IProjectConfigurationDimensionsProviderInternal
+    internal abstract class BaseProjectConfigurationDimensionProvider : IProjectConfigurationDimensionsProviderInternal
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="BaseProjectConfigurationDimensionProvider"/> class.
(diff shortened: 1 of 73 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -205,6 +230,28 @@ async Task UpdateUnderLockAsync(UnconfiguredProject project, Action<ProjectRootE
...
         }
 
+        public virtual async Task<IEnumerable<KeyValuePair<string, IEnumerable<string>>>> GetProjectConfigurationDimensionsAsync(Project project)
+        {
+            Requires.NotNull(project, nameof(project));
+
+            ImmutableArray<string> values = GetOrderedPropertyValues(project);
+            if (values.IsEmpty)
+            {
+                return ImmutableArray<KeyValuePair<string, IEnumerable<string>>>.Empty;
```
drewnoakes, line 240: Minor, but this and other similar code in this class could use `Enumerable.Empty` to reduce allocations a little.

```suggestion
                return Enumerable.Empty<KeyValuePair<string, IEnumerable<string>>>();
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -205,6 +230,28 @@ async Task UpdateUnderLockAsync(UnconfiguredProject project, Action<ProjectRootE
...
+            }
+            else
+            {
+                var dimensionValues = PooledArray<KeyValuePair<string, IEnumerable<string>>>.GetInstance();
+                dimensionValues.Add(new KeyValuePair<string, IEnumerable<string>>(DimensionName, values));
+                return dimensionValues.ToImmutableAndFree();
```
drewnoakes, line 246: Similarly this (and in other methods here) can just return an array, to simplify things and reduce allocations a little:

```suggestion
                return new[] { new KeyValuePair<string, IEnumerable<string>>(DimensionName, values) };
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -75,6 +76,30 @@ private async Task<ImmutableArray<string>> GetOrderedPropertyValuesAsync(Unconfi
...
 
+        /// <summary>
+        /// Gets the property values for the dimension.
+        /// </summary>
+        /// <param name="project"><see cref="Project"/>.</param>
+        /// <returns>Collection of values for the dimension.</returns>
+        /// <remarks>
+        /// From <see cref="IProjectConfigurationDimensionsProvider"/>.
+        /// </remarks>
+        private ImmutableArray<string> GetOrderedPropertyValues(Project project)
```
drewnoakes, line 87: Can `GetOrderedPropertyValuesAsync` be rewritten to use this method, to avoid the code duplication this introduces?

arkalyanms:
@drewnoakes There seems to be some new nullable reference warnings with probably the version upgrades such as _CS8600: Converting null literal or possible null value to non-nullable type_. Build with details [here](https://dev.azure.com/dnceng/public/_build/results?buildId=1332473&view=logs&j=908bce63-940e-59f8-7204-204b1f5ead0b&t=2c2fdffc-ec6b-5664-5bad-2d9431cb9fe3). 

How would you like to see that handled? Inline suppressions or actually fix each and thereby it's callers?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -50,7 +50,30 @@ private async Task<ImmutableArray<string>> GetOrderedPropertyValuesAsync(Unconfi
...
+            });
+        }
+
+        /// <summary>
+        /// Gets the property values for the dimension.
+        /// </summary>
+        /// <param name="project"><see cref="Project"/>.</param>
+        /// <returns>Collection of values for the dimension.</returns>
+        /// <remarks>
+        /// From <see cref="IProjectConfigurationDimensionsProvider5"/>.
```
drewnoakes, line 70: I don't understand what this remark means.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -92,14 +102,13 @@ private async Task<ImmutableArray<string>> GetOrderedPropertyValuesAsync(Unconfi
...
+                return Enumerable.Empty<KeyValuePair<string, string>>();
             }
             else
             {
                 // First value is the default one.
-                var defaultValues = PooledArray<KeyValuePair<string, string>>.GetInstance();
-                defaultValues.Add(new KeyValuePair<string, string>(DimensionName, values[0]));
-                return defaultValues.ToImmutableAndFree();
+                KeyValuePair<string, string>[]? dimensionValues = new[] { new KeyValuePair<string, string>(DimensionName, values[0]) };
+                return dimensionValues.AsEnumerable();
```
drewnoakes, line 111: Can this not just be the following?

```suggestion
                return new[] { new KeyValuePair<string, string>(DimensionName, values[0]) };
```

I don't see what `AsEnumerable` does here other than allocate something we don't need. I might be missing something in the browser though.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -120,13 +129,12 @@ private async Task<ImmutableArray<string>> GetOrderedPropertyValuesAsync(Unconfi
...
-                return ImmutableArray<KeyValuePair<string, IEnumerable<string>>>.Empty;
+                return Enumerable.Empty<KeyValuePair<string, IEnumerable<string>>>();
             }
             else
             {
-                var dimensionValues = PooledArray<KeyValuePair<string, IEnumerable<string>>>.GetInstance();
-                dimensionValues.Add(new KeyValuePair<string, IEnumerable<string>>(DimensionName, values));
-                return dimensionValues.ToImmutableAndFree();
+                KeyValuePair<string, IEnumerable<string>>[]? dimensionValues = new[] { new KeyValuePair<string, IEnumerable<string>>(DimensionName, values) };
+                return dimensionValues.AsEnumerable();
```
drewnoakes, line 137: Ditto:

```suggestion
                return new[] { new KeyValuePair<string, IEnumerable<string>>(DimensionName, values) };
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -205,6 +213,27 @@ async Task UpdateUnderLockAsync(UnconfiguredProject project, Action<ProjectRootE
...
         }
 
+        public virtual Task<IEnumerable<KeyValuePair<string, IEnumerable<string>>>> GetProjectConfigurationDimensionsAsync(Project project)
+        {
+            Requires.NotNull(project, nameof(project));
+
+            ImmutableArray<string> values = GetOrderedPropertyValues(project);
+            if (values.IsEmpty)
+            {
+                return Task.FromResult(Enumerable.Empty<KeyValuePair<string, IEnumerable<string>>>());
```
drewnoakes, line 223: We have a way to reuse cached empty enumerable tasks:

```suggestion
                return TaskResult.EmptyEnumerable<KeyValuePair<string, IEnumerable<string>>>();
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
```diff
@@ -205,6 +213,27 @@ async Task UpdateUnderLockAsync(UnconfiguredProject project, Action<ProjectRootE
...
+            }
+            else
+            {
+                KeyValuePair<string, IEnumerable<string>>[]? dimensionValues = new[] { new KeyValuePair<string, IEnumerable<string>>(DimensionName, values) };
+                return Task.FromResult(dimensionValues.AsEnumerable());
```
drewnoakes, line 228: Similar to the above, this can be done with compile-time hinting rather than run-time work.

```suggestion
                IEnumerable<KeyValuePair<string, IEnumerable<string>>> dimensionValues = new[] { new KeyValuePair<string, IEnumerable<string>>(DimensionName, values) };
                return Task.FromResult(dimensionValues);
```

or

```suggestion
                KeyValuePair<string, IEnumerable<string>>[] dimensionValues = new[] { new KeyValuePair<string, IEnumerable<string>>(DimensionName, values) };
                return Task.FromResult<IEnumerable<KeyValuePair<string, IEnumerable<string>>>>(dimensionValues);
```

drewnoakes (approved):
Looks good! Left some minor comments. Let me know when you would like this merged.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -429,21 +429,24 @@ protected async Task<LaunchSettingsData> ReadSettingsFileFromDiskAsync()
             // serialize their section. Unfortunately, this means the data is string to object which is messy to deal with
             var launchSettingsData = new LaunchSettingsData() { OtherSettings = new Dictionary<string, object>(StringComparers.LaunchProfileProperties) };
             var jsonObject = JObject.Parse(jsonString);
```
MiYanni, line 431: Design question to @drewnoakes and @tmeschter: Is there a reason we're doing this custom type conversion instead of making a JSON type converter? I'd recommend using System.Text.Json instead of Newtsonsoft Json for the performance benefits. A type converter for System.Text.Json is documented here: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#steps-to-follow-the-factory-pattern

System.Text.Json is in the non-framework versions of .NET but the package exists that works for 4.7.2 and up: https://www.nuget.org/packages/System.Text.Json
Docs: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-5-0

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
arkalyanms: It was meant to point to the usage of the api. Removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
arkalyanms: Done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
arkalyanms: Done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Configuration/BaseProjectConfigurationDimensionProvider.cs
arkalyanms: Done

arkalyanms:
> Looks good! Left some minor comments. Let me know when you would like this merged.

If you are ok with the current set of changes, please merge at your earliest convenience. I'll consume this back in CPS after Preview5 forks.

drewnoakes:
Merged. Thanks @arkalyanms!

PR #7544, opened 2021-08-31
Ensure there is a LocalizedControlType property in DataGridViewUITypeEditorColumn

melytc:
Fixes [AzDO#1324876](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1324876)

The Settings Designer has an accessibility bug where the last column, the one with the value of that row's setting, was missing a `LocalizedControlType`.

image

A little bit of context of `ControlType`s:
A control type defines what information a UI element contains and what it can do; control type is used by accessibility tools for users to know what operations can perform on these elements. More about them and [UI Automation Control Types.](https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-control-types-overview).

When looking at how we construct the Settings Designer, we can see that we use a custom `DataGridViewColumn` for the 'Value' column (look at [SettingsDesignerView.vb, line 163](https://github.com/dotnet/project-system/blob/124c733ab0e0966db02ba9fa61020c25d1fd17d4/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsDesignerView.vb#L163)). 

Following an entry in our Microsoft Docs on how to set the localized control type on a custom control (read more about it [here](https://docs.microsoft.com/en-us/accessibility-tools-docs/items/wpf/customcontrol_localizedcontroltype#code-sample-1-setting-the-controltype-of-a-custom-control-to-a-known-uia-controltype)), I was able to add a function in `DataGridViewUITypeEditorColumn` to call the `AutomationPeer` class that will set that specific `AutomationControlType` to be `AutomationControlType.Edit`.

This is what Accessibility Insights looks like with these changes with a WinForms' Setitngs Designer:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7544)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsDesigner/DataGridViewUITypeEditorCell.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsDesigner/DataGridViewUITypeEditorCell.vb
@@ -282,6 +282,23 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
             End Get
         End Property
 
-    End Class
+        Protected Overrides Function CreateAccessibilityInstance() As AccessibleObject
+            Return New DerivedClass(Me)
+        End Function
 
+        Protected Class DerivedClass
+            Inherits DataGridViewCellAccessibleObject
+
+            Public Sub New(owner As DataGridViewCell)
+                MyBase.New(owner)
+            End Sub
+
+            Public Overrides ReadOnly Property Role As AccessibleRole
+                Get
+                    Return AccessibleRole.Cell
+                End Get
+            End Property
+        End Class
+
+    End Class
 End Namespace


src/Microsoft.VisualStudio.Editors/SettingsDesigner/DataGridViewUITypeEditorColumnAccessible.vb
```diff
@@ -0,0 +1,5 @@
+﻿' Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+Public Class DataGridViewUITypeEditorColumnAccessible
```
drewnoakes, line 3: Was this file added intentionally? It looks like it could be deleted (which would fix the build failure):

```
RS0016: Symbol 'DataGridViewUITypeEditorColumnAccessible' is not part of the declared API
RS0016: Symbol 'implicit constructor for 'DataGridViewUITypeEditorColumnAccessible'' is not part of the declared API
```

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypeEditorDataGridViewColumn.vb
```diff
@@ -25,5 +28,23 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
                 MyBase.CellTemplate = Value
             End Set
         End Property
+
+        Protected Function OnCreateAutomationPeer() As AutomationPeer
+            Return New DataGridViewUITypeEditorColumnAutomationPeer(_userControl)
```
drewnoakes, line 33: `_userControl` is never initialised, so will always be `null`. Looking at the base classes (`UserControlAutomationPeer`, which derives from `FrameworkElementAutomationPeer`, which derives from `UIElementAutomationPeer`) I see it supposedly [throws an exception](https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/System/Windows/Automation/Peers/UIElementAutomationPeer.cs,27) if the `owner` argument is null, so I'm surprised this works. If you run this in a debugger and break on all exceptions, do you see an `ArgumentNullException` here?

`DataGridViewColumn` is a WinForms type. `AutomationPeer` is a WPF type. Again, I'm surprised this works!

drewnoakes (approved):
I commented on a few things here that I wasn't sure about, but given you've tested this and it works, I'm happy if you're happy.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/DataGridViewUITypeEditorColumnAccessible.vb
melytc: Ah this shouldn't be here, it was unintentional, thanks for catching it.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/TypeEditorDataGridViewColumn.vb
melytc: VS was playing with me 😕 I'm testing this again and I don't get the same results, but I'm sure it has to do with this. Let me investigate a little bit more.

PR #7546, opened 2021-09-01
Update PostAdaptiveCard to latest version

MiYanni:
Updated to the latest version. It was recommended by the .NET OSS Admins to use the commit SHA as the version number.

Edit: It worked because this PR was shown.
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7546)


.github/workflows/notify-teams-channel.yml
```diff
@@ -1,4 +1,4 @@
-name: Post PRs and issues to Teams
+name: notify-teams-channel
```
MiYanni, line 1: Updated the name that is displayed in the GitHub Checks area to match the naming convention we have for the other checks.

.github/workflows/notify-teams-channel.yml
```diff
@@ -7,12 +7,13 @@ on:
...
 jobs:
-  notify:
+  notify-teams-channel:
+    name: post
     runs-on: ubuntu-latest
     
     steps:
       - name: Notify
-        uses: davidwengier/PostAdaptiveCard@v1.0.4
+        uses: davidwengier/postadaptivecard@ed077f2bd4b80fdaaef21dae2e13a34e556d466d
```
swesonga, line 16: Where does this SHA come from?

MiYanni: It is the commit SHA that relates to that version of the action. I mentioned the lack of user-friendliness of accessing this in my email thread with the .NET OSS Admins.

tmeschter (approved):

jjmew (approved):
:shipit:

.github/workflows/notify-teams-channel.yml
```diff
@@ -1,18 +1,19 @@
-name: Post PRs and issues to Teams
+name: notify-teams-channel
 
 on:
   pull_request:
-    types: [opened, reopened]
+    types: [opened, reopened, ready_for_review]
```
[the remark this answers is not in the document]
MiYanni: Yeah. I noticed that draft PRs might remain for a couple weeks, depending on situations like a set of CPS changes that the PR is waiting on. So, I believe putting a new card in the channel when the PR becomes 'real' makes sense. Unfortunately, there isn't a way to have `opened` only represent non-draft PRs. That would be ideal so we'd only get 1 post..

PR #7548, opened 2021-09-01
Update CPS package versions

drewnoakes:
Also fix warnings after updating `Newtonsoft.Json`. Version 13 contains null annotations, which identified a few places where we could have had unexpected null values.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7548)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -59,8 +59,8 @@
     <PackageReference Update="Microsoft.VisualStudio.Shell.Design"                                    Version="17.0.0-previews-2-31421-281" />
     <PackageReference Update="Microsoft.VisualStudio.Shell.Framework"                                 Version="17.0.0-previews-2-31421-281" />
     <PackageReference Update="Microsoft.VisualStudio.TemplateWizardInterface"                         Version="17.0.0-preview-2-31223-026" />
-    <PackageReference Update="Microsoft.VisualStudio.Threading"                                       Version="17.0.26-alpha" />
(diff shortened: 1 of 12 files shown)


arkalyanms (approved):

swesonga (approved):

build/import/Packages.targets
```diff
@@ -99,7 +99,7 @@
     <PackageReference Update="Microsoft.VisualStudio.HotReload.Components"                            Version="6.0.0-preview.6.21380.4" />
 
     <!-- 3rd party -->
-    <PackageReference Update="Newtonsoft.Json"                                                        Version="12.0.2"/>
+    <PackageReference Update="Newtonsoft.Json"                                                        Version="13.0.1"/>
```
tmeschter, line 102: Is this the version we're going to ship in 17.0?

tmeschter: By "we" I mean VS.

arkalyanms: Not necessarily, if CPS and it's dependencies moves, this will move further along too.

MiYanni: This was the version it was updated to a few weeks back (in VS): https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/340261

MiYanni (approved):

Issue #7549, opened 2021-09-01
Review use of non-async IO via IFileSystem


drewnoakes:
This came up while working on [AB#1371899](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1371899)

We use `IFileSystem` as an abstraction over the filesystem, to make unit testing easier.

All file operations via that interface are currently synchronous. Several could be made to use async operations instead. We should review the use of this interface and update our code to reduce the blocking of threads.

drewnoakes:
#7551 handles the reading and writing of files. The only other operation on the interface I believe could be made async is `CopyFile`.

drewnoakes:
Looking further into the viability of making file copies async led me to:

- https://github.com/dotnet/runtime/issues/20697
- https://github.com/dotnet/runtime/issues/20695

The wisdom seems to be that `File.Copy` is generally faster than using `Stream.CopyToAsync`. Therefore unless we have a compelling reason to avoid blocking the calling thread on that I/O, we should leave the code as-is.

On that basis I'll close this out, as all other operations on our `IFileSystem` API have been made asynchronous.

PR #7551, opened 2021-09-01
Use asynchronous IO for file reads and writes

drewnoakes:
Contributes to #7549.

Our file system abstraction does not provide asynchronous methods for reading and writing files. We should avoid such synchronous operations and use async ones instead, as file IO tends to be "slow", and blocking threads during IO can prevent other useful work from occurring.

This PR adds new async methods for reading and writing files. It also removes the synchronous read/write methods, along with a bunch of other abstraction methods we do not currently use.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7551)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/RemoteCacheFile.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/RemoteCacheFile.cs
@@ -65,7 +65,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
 
                 // Make sure it is valid data before we write to the file (will throw on failure, we don't need the returned data)
                 VersionCompatibilityData.DeserializeVersionData(downLoadedVersionData);
-                _fileSystem.WriteAllText(_cacheFilePath, downLoadedVersionData);
+                await _fileSystem.WriteAllTextAsync(_cacheFilePath, downLoadedVersionData);
                 callBackOnSuccess?.Invoke();
             }
             catch
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/RemoteCacheFile.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/RemoteCacheFile.cs
@@ -37,19 +37,20 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Utilities
         /// <summary>
         /// If the cached file exists reads the data and returns it as a string
         /// </summary>
-        public string? ReadCacheFile()
+        public async Task<string?> ReadCacheFileAsync()
         {
             try
             {
                 // If the cached file exists read it
                 if (_fileSystem.FileExists(_cacheFilePath))
                 {
-                    return _fileSystem.ReadAllText(_cacheFilePath);
+                    return await _fileSystem.ReadAllTextAsync(_cacheFilePath);
                 }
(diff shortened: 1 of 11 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/Web/RemoteCacheFile.cs
```diff
@@ -29,22 +29,28 @@ public bool CacheFileIsStale()
...
         /// <summary>
         /// If the cached file exists reads the data and returns it as a string
         /// </summary>
-        public string? ReadCacheFile()
+        public async Task<string?> ReadCacheFileAsync()
         {
             try
             {
                 // If the cached file exists read it
                 if (_fileSystem.FileExists(_cacheFilePath))
```
swesonga, line 45: >_fileSystem.FileExists(_cacheFilePath) [](http://example.com/codeflow?start=20&length=38)

nit: can use CacheFileExists() here

swesonga (approved):

tmeschter (approved):

jjmew (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/Win32FileSystem.cs
```diff
@@ -46,24 +45,18 @@ public void CopyFile(string source, string destination, bool overwrite)
...
-
-        public void WriteAllText(string path, string content, Encoding encoding)
-        {
-            File.WriteAllText(path, content, encoding);
-        }
-
-        public void WriteAllBytes(string path, byte[] bytes)
-        {
-            File.WriteAllBytes(path, bytes);
+            using var stream = new FileStream(path, FileMode.Create, FileAccess.Write);
```
davkean, line 57: This is a synchronous open and will block the thread; you need to pass isAsync:true.

src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/Win32FileSystem.cs
```diff
@@ -46,24 +45,18 @@ public void CopyFile(string source, string destination, bool overwrite)
             File.Copy(source, destination, overwrite);
         }
 
-        public string ReadAllText(string path)
+        public async Task<string> ReadAllTextAsync(string path)
```
davkean, line 48: Unneeded state machine too.

src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/Win32FileSystem.cs
```diff
@@ -46,24 +45,18 @@ public void CopyFile(string source, string destination, bool overwrite)
...
         {
-            return File.ReadAllText(path);
+            using var stream = new FileStream(path, FileMode.Open, FileAccess.Read);
```
davkean, line 50: Ditto below.

src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/Win32FileSystem.cs
drewnoakes: Fixed in #7560.

src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/Win32FileSystem.cs
drewnoakes: This state machine is needed in order to ensure the disposal happens after the read completes. Ditto for `WriteAllTextAsync`.

PR #7552, opened 2021-09-01
Added new F# build properties

vzarytovskii:
### This adds 2 dropdowns for new F# target settings:
```xml
<PropertyGroup>
    <!-- Whether to use .NET Framework or .NET SDK (coreclr) version of compiler -->
    <FSharpPreferNetFrameworkTools>true</FSharpPreferNetFrameworkTools>
    <!-- Whether to use 64 or 32 bit version of compiler -->
    <FSharpPrefer64BitTools>true</FSharpPrefer64BitTools>
  </PropertyGroup>
```

(F# Pull Request: https://github.com/dotnet/fsharp/pull/11998)

@KevinRansom - Please, have a look and let me know if the following look OK (I've used wording you've sent in the e-mail):

### Here's an overview of new Build page:
buildpage-overview

### Here's how the `Framework tools` selector looks:
framework-tools-overview
framework-tools-expanded

### Generated XML, when `.NET Framework` is selected:

```xml
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <FSharpPreferNetFrameworkTools>true</FSharpPreferNetFrameworkTools>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
    <FSharpPreferNetFrameworkTools>true</FSharpPreferNetFrameworkTools>
  </PropertyGroup>
```

### And when `.NET SDK` is selected:
```xml
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <FSharpPreferNetFrameworkTools>false</FSharpPreferNetFrameworkTools>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
    <FSharpPreferNetFrameworkTools>false</FSharpPreferNetFrameworkTools>
  </PropertyGroup>
```

### Here's how the `Prefer 64 bit tools` selector looks:
prefer64bit-overview

prefer64bit-expanded

### Generated XML, when `32 bit compiler` is selected:
```xml
 <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <FSharpPrefer64BitTools>false</FSharpPrefer64BitTools>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
    <FSharpPrefer64BitTools>false</FSharpPrefer64BitTools>
  </PropertyGroup>
```

### And when `64 bit compiler` is selected:
```xml
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <FSharpPrefer64BitTools>true</FSharpPrefer64BitTools>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
    <FSharpPrefer64BitTools>true</FSharpPrefer64BitTools>
  </PropertyGroup>
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7552)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -35,20 +35,16 @@
                 HelpUrl="https://docs.microsoft.com/dotnet/fsharp/language-reference/compiler-options"
                 Category="General" />
 
-  <EnumProperty Name="FSharpPreferNetFrameworkTools"
+  <BoolProperty Name="FSharpPreferNetFrameworkTools"
               DisplayName="Framework tools"
-              Description="Specifies which compiler to use when building the project."
+              Description="Prefer .NET Framework version of the compiler when available."
               Category="General">
-    <EnumValue Name="True"
-             DisplayName=".NET Framework" />
-    <EnumValue Name="False"
-             DisplayName=".NET SDK" />
-    <EnumProperty.DataSource>
+    <BoolProperty.DataSource>
       <DataSource Persistence="ProjectFileWithInterception"
                  SourceOfDefaultValue="AfterContext"
                  HasConfigurationCondition="False" />
-    </EnumProperty.DataSource>
-  </EnumProperty>
+    </BoolProperty.DataSource>
+  </BoolProperty>
 
   <BoolProperty Name="FSharpPrefer64BitTools"
             DisplayName="Prefer 64 bit tools"
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -44,7 +44,7 @@
   </EnumProperty>
 
   <EnumProperty Name="FSharpPrefer64BitTools"
-            DisplayName="Perfer 64 bit tools"
+            DisplayName="Prefer 64 bit tools"
             Description="Use a 64 bit or 32 bit version of the compiler when available"
             Category="General">
     <EnumValue Name="false"
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -45,7 +45,7 @@
 
   <EnumProperty Name="FSharpPrefer64BitTools"
             DisplayName="Prefer 64 bit tools"
-            Description="Use a 64 bit or 32 bit version of the compiler when available"
+            Description="Use a 64-bit compiler on systems that support both 32-bit and 64-bit compilers."
             Category="General">
     <EnumValue Name="false"
            DisplayName="32 bit compiler" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -35,7 +35,7 @@
 
   <EnumProperty Name="FSharpPreferNetFrameworkTools"
               DisplayName="Framework tools"
-              Description="Compiler to use when building the project"
+              Description="Compiler to use when building the project."
               Category="General">
     <EnumValue Name="true"
              DisplayName=".NET Framework" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -37,7 +37,7 @@
 
   <EnumProperty Name="FSharpPreferNetFrameworkTools"
               DisplayName="Framework tools"
-              Description="Compiler to use when building the project."
+              Description="Specifies which compiler to use when building the project."
               Category="General">
     <EnumValue Name="True"
              DisplayName=".NET Framework" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -33,6 +33,26 @@
                 HelpUrl="https://docs.microsoft.com/dotnet/fsharp/language-reference/compiler-options"
                 Category="General" />
 
+  <EnumProperty Name="FSharpPreferNetFrameworkTools"
+              DisplayName="Framework tools"
+              Description="Compiler to use when building the project"
+              Category="General">
+    <EnumValue Name="true"
+             DisplayName=".NET Framework" />
+    <EnumValue Name="false"
+             DisplayName=".NET SDK" />
+  </EnumProperty>
+
+  <EnumProperty Name="FSharpPrefer64BitTools"
+            DisplayName="Perfer 64 bit tools"
+            Description="Use a 64 bit or 32 bit version of the compiler when available"
+            Category="General">
+    <EnumValue Name="false"
+           DisplayName="32 bit compiler" />
+    <EnumValue Name="true"
+           DisplayName="64 bit compiler" />
+  </EnumProperty>
+
   <EnumProperty Name="DebugType"
               DisplayName="Debug symbols"
               Description="Specifies the kind of debug symbols produced during build."
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
@@ -3,7 +3,9 @@
       OverrideMode="Extend"
       xmlns="http://schemas.microsoft.com/build/2009/properties">
 
-  <!-- TODO: We need fwlink urls generated for all the help pages -->
+  <!-- TODO (https://github.com/dotnet/fsharp/issues/12102):
+       Missing HelpUrl links need to be added.
+  -->
   <EnumProperty Name="Nullable" Category="General" Visible="False" />
   <BoolProperty Name="AllowUnsafeBlocks" Category="General" Visible="False" />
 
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
                 HelpUrl="https://docs.microsoft.com/dotnet/fsharp/language-reference/compiler-options"
                 Category="General" />
 
+  <EnumProperty Name="FSharpPreferNetFrameworkTools"
```
KevinRansom, line 36: How do we know what the default value is?
Is it the first listed, last or is there another property?

The default should be:
- FSharpPreferNetFrameworkTools=false
- FSharpPrefer64BitTools=true

vzarytovskii: By default, these properties are not written to the project file, if they aren't there.

vzarytovskii: I'm going to check it on the latest dev17

KevinRansom (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
tmeschter: At the moment we're largely dependent on the imported .props and .targets (e.g. from the SDK) providing a default value which is then what we will show in the UI. If  the value isn't specified *anywhere* in the project or imports the property will have no value.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
...
+              Description="Compiler to use when building the project"
+              Category="General">
+    <EnumValue Name="true"
+             DisplayName=".NET Framework" />
+    <EnumValue Name="false"
+             DisplayName=".NET SDK" />
+  </EnumProperty>
+
+  <EnumProperty Name="FSharpPrefer64BitTools"
+            DisplayName="Perfer 64 bit tools"
```
tmeschter, line 47: "Prefer" instead of "Perfer"

vzarytovskii:
```suggestion
            DisplayName="Prefer 64 bit tools"
```

tmeschter:
I expect you probably _don't_ want these properties to be configuration-dependent. Properties default to configuration-dependent which is why you're getting this written to the file:
```xml
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <FSharpPrefer64BitTools>true</FSharpPrefer64BitTools>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
    <FSharpPrefer64BitTools>true</FSharpPrefer64BitTools>
  </PropertyGroup>
```

when you probably want just this:

```xml
  <PropertyGroup>
    <FSharpPrefer64BitTools>true</FSharpPrefer64BitTools>
  </PropertyGroup>
```

This is controlled by the `HasConfigurationCondition` attribute on the `DataSource`. You can set a default for the entire `Rule` like so:

https://github.com/dotnet/project-system/blob/648af14095528d140903d64396711f66873be11d/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml#L35-L39

And for an individual property like so:

https://github.com/dotnet/project-system/blob/648af14095528d140903d64396711f66873be11d/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml#L83-L98

vzarytovskii:
> I expect you probably _don't_ want these properties to be configuration-dependent.

Hm, that's a good point. Let me think a bit about it.

> This is controlled by the `HasConfigurationCondition` attribute on the `DataSource`. You can set a default for the entire `Rule` like so:
> …
> And for an individual property like so:
> …

Thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
                 HelpUrl="https://docs.microsoft.com/dotnet/fsharp/language-reference/compiler-options"
                 Category="General" />
 
+  <EnumProperty Name="FSharpPreferNetFrameworkTools"
+              DisplayName="Framework tools"
+              Description="Compiler to use when building the project"
+              Category="General">
```
MiYanni, line 39: You're missing `HelpUrl`s for your new properties. If you can make them fwlinks, that is preferred.

vzarytovskii: We don't have any docs for these yet, unfortunately.

MiYanni: Could you make an issue for it so it isn't forgotten? Also, adding a comment above the properties would let us know that it wasn't *intentionally* forgotten. 😉

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
                 HelpUrl="https://docs.microsoft.com/dotnet/fsharp/language-reference/compiler-options"
                 Category="General" />
 
+  <EnumProperty Name="FSharpPreferNetFrameworkTools"
+              DisplayName="Framework tools"
+              Description="Compiler to use when building the project"
```
drewnoakes, line 38:
```suggestion
              Description="Compiler to use when building the project."
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
...
+              Category="General">
+    <EnumValue Name="true"
+             DisplayName=".NET Framework" />
+    <EnumValue Name="false"
+             DisplayName=".NET SDK" />
+  </EnumProperty>
+
+  <EnumProperty Name="FSharpPrefer64BitTools"
+            DisplayName="Prefer 64 bit tools"
+            Description="Use a 64 bit or 32 bit version of the compiler when available"
```
drewnoakes, line 48: I suggest making this text match the "Prefer 32-bit" property, and adding a trailing full stop.

```suggestion
            Description="Use a 64-bit compiler on systems that support both 32-bit and 64-bit compilers."
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
...
+              DisplayName="Framework tools"
+              Description="Compiler to use when building the project"
+              Category="General">
+    <EnumValue Name="true"
+             DisplayName=".NET Framework" />
+    <EnumValue Name="false"
+             DisplayName=".NET SDK" />
+  </EnumProperty>
+
+  <EnumProperty Name="FSharpPrefer64BitTools"
```
drewnoakes, line 46: This property feels like it should be a boolean (check box) to me. That would match what we have for ["Prefer 32-bit" property](https://github.com/dotnet/project-system/blob/e9f3bbd77e6e1e8d151e06580d469a1150745973/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml#L100-L123).

image

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +33,26 @@
...
+            DisplayName="Prefer 64 bit tools"
+            Description="Use a 64 bit or 32 bit version of the compiler when available"
+            Category="General">
+    <EnumValue Name="false"
+           DisplayName="32 bit compiler" />
+    <EnumValue Name="true"
+           DisplayName="64 bit compiler" />
```
drewnoakes, line 53: This label seems potentially confusing. There is no guarantee you'll end up with a 64-bit compiler. I think a check box would be clearer.

drewnoakes:
It looks like this will miss 17.0 at this point.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
vzarytovskii: Yeah, we discussed it internally, and will add a default value for it (as oppose to now it just having empty value, which is treated as default).

vzarytovskii:
> It looks like this will miss 17.0 at this point.

Yeah, that is fine, it's fine if we will make it to the update, and not initial release.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
vzarytovskii: I will probably make it a checkbox, however it may be a bit confusing at the moment, since we don't (yet) set the default value to 'true', but it is rather empty, see:

https://github.com/dotnet/fsharp/blob/release/dev17.0/src/fsharp/FSharp.Build/Microsoft.FSharp.Targets#L64-L69

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
vzarytovskii: Done, will add it to xaml file too
https://github.com/dotnet/fsharp/issues/12102

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
MiYanni: I changed our (C#) Prefer 32-bit property because we should be consistent with boolean values always being represented as checkboxes. I definitely recommend the same here.

vzarytovskii:
I have changed the FSharpPrefer64BitTools to be a boolean value (i.e. checkbox), now since we have default values merged.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.FSharp.xaml
```diff
@@ -33,6 +35,32 @@
                 HelpUrl="https://docs.microsoft.com/dotnet/fsharp/language-reference/compiler-options"
                 Category="General" />
 
+  <EnumProperty Name="FSharpPreferNetFrameworkTools"
+              DisplayName="Framework tools"
+              Description="Compiler to use when building the project."
```
drewnoakes, line 40: We use complete sentences for descriptions.

```suggestion
              Description="Specifies which compiler to use when building the project."
```

A question on this property. The name says "prefer" which makes it sound like it'll use the .NET Framework compiler if it's available. This metadata will make the property appear less like a preference and more like an absolute directive. What happens if I select ".NET Framework" in the UI, but the .NET Framework compiler is not available? Could the user be surprised here?

vzarytovskii: If the .NET Framework is not available, but selected, we will ignore the property, and chose the SDK compiler.

drewnoakes: In that case I think this should be a checkbox with text "Prefer the .NET Framework compiler when available."

Right now it seems like a guarantee that the specified compiler will be used.

vzarytovskii: @KevinRansom are you OK with having it as a boolean (checkbox)? Saying "Prefer .NET Framework version of the compiler when available".

vzarytovskii:
@KevinRansom I've made both properties boolean (checkbox), they will look like this in the UI:
image
Does it look OK to you, or you would prefer explicit checkbox(es)?

drewnoakes (approved):

PR #7553, opened 2021-09-02
Condition the CollectSuggestedWorkloads target on SDK projects with MissingWorkloadPacks

swesonga:
This avoids errors about the target GetSuggestedWorkloads not existing in non-SDK projects.

Note: the GetSuggestedWorkloads target is defined in https://github.com/dotnet/sdk/blob/release/6.0.1xx/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7553)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -377,7 +377,8 @@
   <!-- This target is used to collect the SuggestedWorkload items in the project.-->
   <Target Name="CollectSuggestedWorkloads"
           Returns="@(SuggestedWorkload)"
-          DependsOnTargets="GetSuggestedWorkloads" />
+          DependsOnTargets="GetSuggestedWorkloads"
+          Condition="'$(UsingMicrosoftNETSdk)' == 'true'" />
 
   <!-- Validates that the correct properties have been set for design-time compiles  -->
   <Target Name="_CheckCompileDesignTimePrerequisite">
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -378,7 +378,7 @@
   <Target Name="CollectSuggestedWorkloads"
           Returns="@(SuggestedWorkload)"
           DependsOnTargets="GetSuggestedWorkloads"
-          Condition="'$(UsingMicrosoftNETSdk)' == 'true'" />
+          Condition="'@(MissingWorkloadPack)' != '' And '$(UsingMicrosoftNETSdk)' == 'true'" />
 
   <!-- Validates that the correct properties have been set for design-time compiles  -->
   <Target Name="_CheckCompileDesignTimePrerequisite">


MiYanni (approved):

jjmew (approved):
:shipit:

davkean:
All managed projects see those targets.

davkean:
This is why these built-in targets need to be in Common.targets, even if its a dummy target.

swesonga:
> How would you see this warning today? That is, how would a non-SDK project end up being loaded with those design-time targets?
> 
> Or is the issue related to the version of the SDK maybe not having that target?

This was reported from opening a shared project (.shproj) but it also shows up for older SDKs as you've observed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
```diff
@@ -377,7 +377,8 @@
   <!-- This target is used to collect the SuggestedWorkload items in the project.-->
   <Target Name="CollectSuggestedWorkloads"
           Returns="@(SuggestedWorkload)"
-          DependsOnTargets="GetSuggestedWorkloads" />
+          DependsOnTargets="GetSuggestedWorkloads"
+          Condition="'$(UsingMicrosoftNETSdk)' == 'true'" />
```
AArnott, line 381: In [my repro](https://developercommunity.visualstudio.com/t/the-target-getsuggestedworkloads-does-not-exist-in/1520706?from=email), all the projects are Microsoft.NET.Sdk projects and yet the error remains. I don't see how this added condition will help in my case.

swesonga: Yes, this is incomplete as @davkean and @drewnoakes pointed out. I'm current looking for the proper location for this target.

swesonga:
@davkean @drewnoakes I'll merge the newly added condition as a workaround to get the product unblocked then investigate the move to Microsoft.Common.targets

rainersigwald:
> This is why these built-in targets need to be in Microsoft.Common.targets, even if its a dummy target.

There's no guarantee that the copy of `Microsoft.Common.targets` imported matches the project system, so this isn't sufficient.

AArnott:
MSBuild has APIs to allow you to request an msbuild target without breaking the build when it doesn't exist. The DependsOnTargets attribute value could be set to an msbuild property that is appended to near the dependency target so that the dependency only exists where the target does.

davkean:
> There's no guarantee that the copy of Microsoft.Common.targets imported matches the project system, so this isn't sufficient.

This has been the relationship of the project-system and MSBuild since literally the start of MSBuild. We all ship in the same box on the same schedule. The same cannot be said for the SDK.

swesonga:
> > There's no guarantee that the copy of Microsoft.Common.targets imported matches the project system, so this isn't sufficient.
> 
> This has been the relationship of the project-system and MSBuild since literally the start of MSBuild. We all ship in the same box on the same schedule. The same cannot be said for the SDK.

@davkean, wouldn't it be unexpected to have dummy design-time targets in Microsoft.Common.targets? I wonder if the better fix is for the Windows Application Packaging project system to import Microsoft.Managed.DesignTime.targets (or is there a separate design-time targets file for other project systems)?

davkean:
Note it's not unexpected, its been this way since the start. VS does not support running against versions of MSBuild it didn't ship with.

swesonga:
> No it's not unexpected, its been this way since the start. VS does not support running against versions of MSBuild it didn't ship with. The same cannot be said for the SDK.

That makes sense. @rainersigwald, is your point about not wanting such targets in MSBuild common targets more along the lines of a separation of concerns?

PR #7554, opened 2021-09-02
Explain alternative gestures to open the Launch Profile UI

### LINKED ISSUE #7337: Explain alternative gestures to open the Launch Profile UI (0.2)
Currently users are shown this text:

image

In fact there are better ways to get to the Launch Profile UI now:

image

image

The text we should should point users towards these alternative gestures so that it's clear you don't have to go an indirect route via the Project Properties UI.

drewnoakes:
Fixes #7337

This guides users towards additional ways to open the Launch Profiles UI, as going via the Project Properties UI is slower than the alternatives.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7554)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/DebugPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/DebugPropertyPage.xaml
@@ -15,7 +15,7 @@
 
   <StringProperty Name="DebugPagePlaceholderDescription"
                   DisplayName="Ignored"
-                  Description="The management of launch profiles has moved to a dedicated dialog. It may be accessed via the link below.">
+                  Description="The management of launch profiles has moved to a dedicated dialog. It may be accessed via the link below, via the Debug menu in the menu bar, or via the Debug Target command on the Standard tool bar.">
     <StringProperty.Metadata>
       <NameValuePair Name="VisibilityCondition">
         <NameValuePair.Value>(has-project-capability "LaunchProfiles")</NameValuePair.Value>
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/DebugPropertyPage.xaml
```diff
@@ -15,7 +15,7 @@
 
   <StringProperty Name="DebugPagePlaceholderDescription"
                   DisplayName="Ignored"
-                  Description="The management of launch profiles has moved to a dedicated dialog. It may be accessed via the link below.">
+                  Description="The management of launch profiles has moved to a dedicated dialog. It may be accessed via the link below, via the Debug menu in the menu bar, or via the Debug Target command on the Standard tool bar.">
```
MiYanni, line 18: nit: Is "toolbar" a single word at this point in the evolution of software? 🤓

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/DebugPropertyPage.xaml
drewnoakes: Not in WinForms/WPF, where it's `ToolBar` rather than `Toolbar`.

MiYanni: I didn't mean context-specific to the UI engine used. The user doesn't know/care about that (they don't see the names of the controls). I meant as a concept in software, such as:
- https://en.wikipedia.org/wiki/Toolbar
- https://www.merriam-webster.com/dictionary/toolbar
- https://docs.microsoft.com/en-us/windows/apps/design/controls/menus-and-context-menus (subsection called "Menus and toolbars")
- https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-toolbars (old guide)

PR #7559, opened 2021-09-03
Fix GitHub issue templates

drewnoakes:
#7476 added our a GitHub issue template for branch checklists. That PR used a newer method of specifying templates than the one we used previously.

Adding that template caused GitHub to show visitors the following when they try to create a new issue:

image

The indicated link would show our old template, but that is non-obvious.

This PR adds two new items to the list:

- Bug report
- Feature request

It also adds a link to the "Report a Problem" tool documentation.

## After

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7559)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,34 @@
+---
+name: Bug report
+about: Report an issue with the .NET Project System for Visual Studio
+title: 
+labels:
```
MiYanni, line 5: Please add `Bug` as an automatic label.

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,16 @@
+---
+name: Feature request
+about: Request a new feature of .NET Project System for Visual Studio
+title: 
+labels:
```
MiYanni, line 5: Please add `Feature Request` as an automatic label.

MiYanni (approved):
Good idea. I think we'll end up using this for a bit until the [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) feature is out of beta.

.github/ISSUE_TEMPLATE/bug-report.md
drewnoakes: I considered that but decided against it as I suspect we'll see a lot of stuff filed that isn't actually a bug. Let's do this in triage for things we actually believe are real bugs.

If that approach creates undue work I'm fine with adding the automatic label then.

.github/ISSUE_TEMPLATE/feature-request.md
drewnoakes: Ditto.

PR #7565, opened 2021-09-06
Fix NFE in the imports tree

### LINKED ISSUE #6675: NFE while constructing the project imports tree
In 16.8: the code below seems to add the same file to the tree, which may just need a simple check to deduplicate items?

System.ArgumentException - Microsoft.VisualStudio.ProjectSystem.Designers.ProjectTree.Add

> An item with the caption "Sdk.props"(ProjectTreeFlags: ProjectImportImplicit FileOnDisk ProjectImport FileSystemEntity) already exists under parent "Sdk.props" with the same sort order as the item being added.

```
["Microsoft.VisualStudio.ProjectSystem.Designers.ProjectTree.Add(UnattachedProjectTreeNode)
>>>Microsoft.VisualStudio.ProjectSystem.Designers.ProjectTree.Add(IProjectTree)
>>>Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__AddChild|22_9(IProjectTree2,<>c__DisplayClass22_1&)
>>>Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|22_8(IReadOnlyList`1,IProjectTree2)
>>>Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<set_ShowAllFiles>g__SyncNode|22_8(IReadOnlyList`1,IProjectTree2)
>>>Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports.ImportTreeProvider.<>c__DisplayClass22_0.<set_ShowAllFiles>b__7(IProjectVersionedValue`1,ConfiguredProjectExports,CancellationToken)
>>>Microsoft.VisualStudio.ProjectSystem.ProjectTreeProviderBase.TreeUpdateSubmission.<UpdateTree>d__19.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task)
>>>Microsoft.VisualStudio.ProjectSystem.CoreProjectTreeProviderBase.<>c__DisplayClass76_1.<<Initialize>b__2>d.MoveNext()
>>>System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)
>>>System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task)
>>>Microsoft.VisualStudio.ProjectSystem.CoreProjectTreeProviderBase.<>c__DisplayClass78_0.<<SubmitTreeUpdateCoreAsync>g__waitOnTask|0>d.MoveNext()
>>>System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>>>Microsoft.VisualStudio.ProjectSystem.CommonProjectSystemTools.Rethrow(Exception)
>>>Microsoft.VisualStudio.ProjectSystem.ProjectErrorReporting.<>c__DisplayClass6_0.<SubmitErrorReport>b__0()
>>>Microsoft.VisualStudio.ProjectSystem.ExceptionFilter.<>c__DisplayClass2_0.<Guard>g__action|0()
>>>GuardMethodClass.GuardMethod(Func`1,Func`2,Func`2)"]
```

drewnoakes:
Fixes #6675

This is a top NFE as reported by CPS.

As discussed in https://github.com/dotnet/project-system/issues/6675#issuecomment-913578589 the imports tree shows project file names as captions in the tree. While there are no duplicate paths, two different paths in a given level of the tree may have the same file name. That is supported on the initial population of the tree, as we also differentiate them by display order. However if these items are reordered, we attempt to update their display orders which can produce a transient state in which duplicates exist, which results in the _The node to add collides with a node already present_ error message we see.

The primary fix here is to look for duplicate file names in any given level of the tree and disambiguate them by including their path in parentheses in the caption.

A .NET 5 application will imports several `WorkloadManifest.targets` files at the same level of the tree. With this change they now appear as:

image

The full path should be used here (rather than numeric suffixes) as that remains stable in the face of reordering within the project that imports these.

Other fixes in this PR:

- Joining an upstream data source that was missing, which may help prevent dead locks.
- Avoid caching `IProjectTree` objects beyond their lifetimes by looking them up on each iteration.
- Optimise child node replacement.
- Enable skipping of dataflow items by our action block, as it is safe to do so.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
@@ -206,27 +206,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports
 
                                     IProjectTree updatedTree = SyncNode(
                                         imports: e.Value.Item1.Value,
-                                        tree: (IProjectTree2)currentTree.Value.Tree);
+                                        node: (IProjectTree2)currentTree.Value.Tree);
 
                                     return Task.FromResult(new TreeUpdateResult(updatedTree, e.DataSourceVersions));
                                 });
 
                             return;
 
-                            IProjectTree2 SyncNode(IReadOnlyList<IProjectImportSnapshot> imports, IProjectTree2 tree)
+                            IProjectTree2 SyncNode(IReadOnlyList<IProjectImportSnapshot> imports, IProjectTree2 node)
                             {
                                 var existingChildByPath = new Dictionary<string, IProjectTree2>(StringComparers.Paths);
 
-                                foreach (IProjectTree2 existingNode in tree.Children)
+                                foreach (IProjectTree2 existingNode in node.Children)
                                 {
                                     Assumes.NotNullOrEmpty(existingNode.FilePath);
 
                                     if (!imports.Any(import => StringComparers.Paths.Equals(import.ProjectPath, existingNode.FilePath)))
                                     {
                                         // Remove child that's no longer present
-                                        if (tree.TryFind(existingNode.Identity, out IProjectTree? child))
+                                        if (node.TryFind(existingNode.Identity, out IProjectTree? child))
                                         {
-                                            tree = (IProjectTree2)child.Remove();
+                                            node = (IProjectTree2)child.Remove();
                                         }
                                     }
                                     else
@@ -256,12 +256,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports
                                         // Recur down the tree
                                         newChild = SyncNode(import.Imports, newChild);
 
-                                        tree = AddChild(newChild);
+                                        node = AddChild(newChild);
                                     }
                                     else if (child.DisplayOrder != displayOrder)
                                     {
                                         // Child exists but with the wrong display order
-                                        tree = (IProjectTree2)child.SetDisplayOrder(displayOrder).Parent!;
+                                        node = (IProjectTree2)child.SetDisplayOrder(displayOrder).Parent!;
                                     }
                                     else
                                     {
@@ -270,15 +270,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports
 
                                         if (!ReferenceEquals(child, newChild))
                                         {
-                                            tree = ReplaceChild(child, newChild);
+                                            node = ReplaceChild(child, newChild);
                                         }
                                     }
                                 }
 
-                                return tree;
+                                return node;
 
-                                IProjectTree2 AddChild(IProjectTree2 child) => (IProjectTree2)tree.Add(child).Parent!;
-                                IProjectTree2 ReplaceChild(IProjectTree2 oldChild, IProjectTree2 newChild) => (IProjectTree2)tree.Remove(oldChild).Add(newChild).Parent!;
+                                IProjectTree2 AddChild(IProjectTree2 child) => (IProjectTree2)node.Add(child).Parent!;
+                                IProjectTree2 ReplaceChild(IProjectTree2 oldChild, IProjectTree2 newChild) => (IProjectTree2)node.Remove(oldChild).Add(newChild).Parent!;
                             }
                         }
                     }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
@@ -171,6 +171,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Tree.ProjectImports
                             DataflowBlockFactory.CreateActionBlock<IProjectVersionedValue<ValueTuple<IProjectImportTreeSnapshot, IProjectSubscriptionUpdate>>>(
                                 SyncTree,
                                 _project,
+                                skipIntermediateInputData: true, // We can skip versions without breaking the tree
                                 nameFormat: "Import Tree Action: {1}");
 
                         _subscriptions.Add(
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
```diff
@@ -270,15 +271,39 @@ IProjectTree2 SyncNode(IReadOnlyList<IProjectImportSnapshot> imports, IProjectTr
...
+                                        string fileName = result[import.ProjectPath];
+                                        
+                                        if (isDuplicateByFileName[fileName])
+                                        {
+                                            result[import.ProjectPath] = $"{fileName} ({Path.GetDirectoryName(import.ProjectPath)})";
+                                        }
+                                    }
+
+                                    return result;
+                                }
```
MiYanni, line 306: This implementation contains a lot of looping (3 full loops over `imports`, and many loops over `isDuplicateByFileName` via `ContainsKey`) and creates 2 dictionaries in the process. Instead, this works if you don't require the `StringComparers.Paths`:
```suggestion
                                Dictionary<string, string> GetCaptionByProjectPath()
                                {
                                    var result = new Dictionary<string, string>();

                                    foreach (IProjectImportSnapshot import in imports)
                                    {
                                        string fileName = Path.GetFileName(import.ProjectPath);
                                        
                                        result[import.ProjectPath] = result.ContainsValue(fileName) ?
                                          $"{fileName} ({Path.GetDirectoryName(import.ProjectPath)})" :
                                          fileName;
                                    }

                                    return result;
                                }
```

If you do require `StringComparers.Paths`, then the `ContainsValue` needs expanded as it has no way to provide a value comparer:
```suggestion
                                Dictionary<string, string> GetCaptionByProjectPath()
                                {
                                    var result = new Dictionary<string, string>();

                                    foreach (IProjectImportSnapshot import in imports)
                                    {
                                        string fileName = Path.GetFileName(import.ProjectPath);
                                        string caption = fileName;

                                        foreach (string value in result.Values)
                                        {
                                            if (value.Equals(fileName, StringComparers.Paths))
                                            {
                                                caption = $"{fileName} ({Path.GetDirectoryName(import.ProjectPath)})";
                                                break;
                                            }
                                        }

                                        result[import.ProjectPath] = caption;
                                    }

                                    return result;
                                }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
```diff
@@ -256,12 +257,12 @@ IProjectTree2 SyncNode(IReadOnlyList<IProjectImportSnapshot> imports, IProjectTr
...
                                         newChild = SyncNode(import.Imports, newChild);
 
-                                        tree = AddChild(newChild);
+                                        node = AddChild(newChild);
                                     }
                                     else if (child.DisplayOrder != displayOrder)
                                     {
                                         // Child exists but with the wrong display order
-                                        tree = ReplaceChild(child, child.SetDisplayOrder(displayOrder));
+                                        node = (IProjectTree2)child.SetDisplayOrder(displayOrder).Parent!;
```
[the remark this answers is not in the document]
drewnoakes: We know it is not null because we accessed it from its parent. An extension method could work here, but I don't think we need the runtime check.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
drewnoakes: This is not equivalent. If we have `A/B/C` and `X/Y/C` in that order, we need to display `C (A/B)` and `C (X/Y)`. The above code shows `C` and `C (X/Y)`.

> This implementation contains a lot of looping (3 full loops over imports, and many loops over isDuplicateByFileName via ContainsKey) and creates 2 dictionaries in the process. Instead, this works if you don't require the StringComparers.Paths:

Minor point: `ContainsKey` is O(1) not O(N). There's no looping hidden in there.

The tree tends to be fairly small at any given level. I am not massively concerned about the perf impact here.

I'll take another look with a fresh pair of eyes and see if I can make this allocate less. We may be able to reuse the temporary collections as we walk the tree, instead of throwing them away each time.

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/ProjectImports/ImportTreeProvider.cs
MiYanni: Ah, I understand. It wasn't obvious when reading the original code that both needed to be updated. In that case, you'd just have to update the original caption. The code gets messier so I understand not going this route:
```suggestion
                                Dictionary<string, string> GetCaptionByProjectPath()
                                {
                                    var hasAddedCaption = new List<string>();
                                    var result = new Dictionary<string, string>();

                                    foreach (IProjectImportSnapshot import in imports)
                                    {
                                        string fileName = Path.GetFileName(import.ProjectPath);
                                        string caption = filename;

                                        for (int i = 0; i < result.Keys.Count; ++i)
                                        {
                                            string resultProjectPath = result.Keys[i];
                                            if (result[resultProjectPath].Equals(fileName, StringComparers.Paths))
                                            {
                                                caption = CreateCaption(fileName, import.ProjectPath);
                                                hasAddedCaption.Add(import.ProjectPath);

                                                // Update the original caption since a duplicate has been found.
                                                if (!hasAddedCaption.Contains(resultProjectPath))
                                                {
                                                    result[resultProjectPath] = CreateCaption(result[resultProjectPath], resultProjectPath);
                                                    hasAddedCaption.Add(resultProjectPath);
                                                    break;
                                                }
                                            }
                                        }

                                        result[import.ProjectPath] = caption;
                                    }

                                    return result;

                                    static string CreateCaption(string fn, string pp) => $"{fn} ({Path.GetDirectoryName(pp)})";
                                }
```

drewnoakes: This now has nested loops, which gives me more concern regarding performance.

PR #7569, opened 2021-09-07
Make ConfiguredProjectImplicitActivationTrackingInstance contribute to operation progress

### LINKED ISSUE #7468: Intellisense ready state might be declared too earlier in warm load (when DT cache is ready.)
The intellisense stage almost finished immediately after loading the solution in some warm loading time, after that, I saw nuget restore messages in the output window (although mostly they didn’t do anything, but sometime it might have pending work.)

I think the problem happens because the NuGet restore dataflow endpoints is registered too late.  CPS calls all background works are done, when all registered dataflow points are finished. If one is registered after that, it will push the project to the processing state again, but the solution intellisense ready state only count the first time when a project indicates it’s ready.

To ensure that, we need to satisfy one condition below:
•	A critical dataflow affecting the intellisense stage must be registered before solution load is completed
•	If we need run background initialization code to initialize any critical dataflow a part of intellisense stage after solution load is completed (which may make sense, because we do want to maintain the initial loading time short). This background initialization task must be registered to the operation progress itself, as an umbrella task to cover the other registration work.  Generally, we prefer the 2nd solution, so we load less code during modal phase of the solution loading.)

I wonder that we don’t cover the NuGet restore part in this way. (The language service part was covered by the first condition, so I don’t think this too early happens in a cold load/DT cache not ready state .

drewnoakes:
Fixes #7468

During project load, we have several implicitly activated components (implementations of `IImplicitlyActiveConfigurationComponent`) who may themselves contribute to operation progress. It's important that we give them a chance to do so before operation progress completes.

Previously that was not guaranteed. Our `IProjectDynamicLoadComponent.LoadAsync` method in `ConfiguredProjectImplicitActivationTracking` would indirectly set up a dataflow before completing, but that dataflow would have to pump before the `IImplicitlyActiveConfigurationComponent` instances could be activated. Until those messages arrived, there existed a window during which operation progress could complete before the component (e.g. `PackageRestoreProgressTrackerInstance`) could register its data source with `IDataProgressTrackerService`.

This commit adds an additional `IDataProgressTrackerServiceRegistration` to cover that window of time. It is registered during `IProjectDynamicLoadComponent.LoadAsync`, and updated after each round of component activation/deactivation.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.Instance.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.Instance.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using Microsoft.VisualStudio.ProjectSystem.OperationProgress;
 
 namespace Microsoft.VisualStudio.ProjectSystem
 {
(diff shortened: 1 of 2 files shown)


drewnoakes:
cc @lifengl @adrianvmsft @davkean for eyeballs on this change.

swesonga:
> Previously that was not guaranteed. Our `IProjectDynamicLoadComponent.LoadAsync` method in `ConfiguredProjectImplicitActivationTracking` would indirectly set up a dataflow before completing, but that dataflow would have to pump before the `IImplicitlyActiveConfigurationComponent` instances could be activated.

Which code in LoadAsync indirectly sets up a dataflow?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.Instance.cs
```diff
@@ -41,6 +46,8 @@ public Task InitializeAsync()
 
             protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
             {
+                _dataProgressTrackerRegistration = _dataProgressTrackerService.RegisterForIntelliSense(this, _project, nameof(ConfiguredProjectImplicitActivationTrackingInstance));
```
swesonga, line 49: Should the name of the data source include additional information about the project and/or configuration? I wonder if that might be helpful when looking at a dump file with multiple configurations where it might not be clear which is still pending.

swesonga (approved):

MiYanni (approved):

drewnoakes:
>> Previously that was not guaranteed. Our `IProjectDynamicLoadComponent.LoadAsync` method in `ConfiguredProjectImplicitActivationTracking` would indirectly set up a dataflow before completing, but that dataflow would have to pump before the `IImplicitlyActiveConfigurationComponent` instances could be activated.
>
> Which code in LoadAsync indirectly sets up a dataflow?

There's some more detail on the full flow here https://github.com/dotnet/project-system/issues/7468#issuecomment-914239741

`ConfiguredProjectImplicitActivationTracking` implements `IProjectDynamicLoadComponent`, which creates an `ConfiguredProjectImplicitActivationTrackingInstance`, which creates the dataflow subscription I'm referring to:

https://github.com/dotnet/project-system/blob/5abac577b3240eb32264e1cfd3d8f230e6106d82/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ConfiguredProjectImplicitActivationTracking.Instance.cs#L42-L52

When that dataflow pumps, we create `PackageRestoreProgressTracker` which then creates `PackageRestoreProgressTrackerInstance`.

This PR ensures all of this is covered by operation progress.

PR #7570, opened 2021-09-08
Update build queue

tmeschter:
The VSEngSS-MicroBuild2019 pool is being phased out in favor of a new pool, VSEngSS-MicroBuild2019-1ES. This should be exactly the same setup, just managed by the 1ES team.

We have until the end of September 2021 to switch over, so if we experience problems with this new pool we can fall back to the old one while they are worked out.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7570)

### DIFF
--- a/build/ci/official.yml
+++ b/build/ci/official.yml
@@ -10,7 +10,7 @@ resources:
 - repo: self
   clean: true
 queue:
-  name: VSEngSS-MicroBuild2019
+  name: VSEngSS-MicroBuild2019-1ES
   demands: Cmd
   timeoutInMinutes: 90
 variables:


swesonga (approved):

build/ci/official.yml
```diff
@@ -10,7 +10,7 @@ resources:
 - repo: self
   clean: true
 queue:
-  name: VSEngSS-MicroBuild2019
+  name: VSEngSS-MicroBuild2019-1ES
```
MiYanni, line 13: We recently switched to the `VSEngSS-MicroBuild2019` across the board. This was advised to us. I didn't see anything about this `-1ES` variant. There is more work to do than switching just this file if we need to do this across the board. Unfortunately, some of our pipelines (such as release pipelines) do not exist as YAML files in our repo. They are only in DevOps. Let me know if we are doing this for everything, and we can walk through it together.

tmeschter: I can forward you the email. I was identified as the owner of the Dotnet-Project-System pipeline (🤷🏻‍♂️) which is why I got the notification.

tmeschter: But the short answer is that `VSEngSS-MicroBuild2019` is being retired so any uses will need to be replaced.

MiYanni: Got the email, thanks! Looks like it came from the same people that got us to do the original switch to scale sets. I understand. We'll update the file here and then do the other ones that need it too.

MiYanni (approved):

MiYanni:
I think we could go ahead with this change, right? I see no reason not to. We could jump on a call at some point and I'll show you the other places that need this change too.

tmeschter:
There's no reason to hold off; I just hadn't gotten around to clicking the button. :-)

PR #7573, opened 2021-09-09
Don't force load of VC projects

drewnoakes:
Fixes [AB#1388556](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1388556)

When a solution contains both VC and SDK-style projects, the `DotNetCoreProjectCompatibilityDetector` can unwittingly force VS to fully load projects rather than serve them from the cache.

It does this by retrieving the `UnconfiguredProject` from the `IVsHierarchy`, which internally QIs a `IVsBrowseObjectContext` interface, which loads the project.

We avoid that load by performing a capability query on the project instead.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetector.cs
@@ -15,6 +15,7 @@ using Microsoft.VisualStudio.ProjectSystem.VS.UI;
 using Microsoft.VisualStudio.ProjectSystem.VS.Utilities;
 using Microsoft.VisualStudio.Settings;
 using Microsoft.VisualStudio.Setup.Configuration;
+using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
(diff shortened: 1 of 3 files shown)


swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VersionCompatibility/DotNetCoreProjectCompatibilityDetector.cs
```diff
@@ -33,6 +34,8 @@ internal partial class DotNetCoreProjectCompatibilityDetector : IDotNetCoreProje
         private const string VersionDataFilename = "DotNetVersionCompatibility.json";
         private const int CacheFileValidHours = 24;
 
+        private const string RequiredProjectCapabilities = ProjectCapability.DotNet + " & " + ProjectCapability.PackageReferences;
```
tmeschter, line 37: I understand checking for `DotNet` but why `PackageReferences`?

tmeschter: Oh I see. This is just moving around existing code. OK.

tmeschter (approved):

MiYanni (approved):

PR #7574, opened 2021-09-09
Fix a race in starting Hot Reload sessions

tmeschter:
We have a bug in `HotReloadSessionManager.ActivateSessionAsync` such that it is possible for the process to exit _after_ we've hooked the `Process.Exited` event but before we've started the session. The event handler could end up running on another thread and calling `StopSessionAsync` before the session has started. Then `ActivateSessionAsync` ends up starting a session for the exited process--and because the `Exited` event has already fired we don't get another chance to stop it.

To fix this we need to prevent concurrent access to the `HotReloadSessionManager` state. We already protected the `ImmutableDictionary` of active sessions through the use of `ImmutableInterlocked` but this does nothing to protect the state of the objects _within_ the dictionary. In this change we introduce an `AsyncSemaphore` make use of it everywhere we touch the state.

Also, we were not actually `await`ing the session start, which means that even after introducing the semaphore it might still occur after the `Exited` handler ran. This was to avoid a deadlock that has already been mitigated, so we can re-introduce the `await`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7574)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
@@ -24,10 +24,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
         private readonly Lazy<IProjectHotReloadAgent> _projectHotReloadAgent;
         private readonly Lazy<IHotReloadDiagnosticOutputService> _hotReloadDiagnosticOutputService;
 
+        // Protect the state from concurrent access. For example, our Process.Exited event
+        // handler may run on one thread while we're still setting up the session on
+        // another. To ensure consistent and proper behavior we need to serialize access.
+        private readonly AsyncSemaphore _semaphore = new(initialCount: 1);
+
         private HotReloadState? _pendingSessionState = null;
-
         private ImmutableDictionary<int, HotReloadState> _activeSessions = ImmutableDictionary<int, HotReloadState>.Empty;
-
         private int _nextUniqueId = 1;
 
         [ImportingConstructor]
@@ -47,6 +50,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
 
         public async Task ActivateSessionAsync(int processId, bool runningUnderDebugger)
         {
+            using AsyncSemaphore.Releaser semaphoreReleaser = await _semaphore.EnterAsync();
+
             if (_pendingSessionState is not null)
             {
                 Assumes.NotNull(_pendingSessionState.Session);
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
Is the interlocked code still needed, given the semaphore?

tmeschter:
> Is the interlocked code still needed, given the semaphore?

No. I attempted to remove it; did I miss something?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -24,10 +23,13 @@ internal class ProjectHotReloadSessionManager : IProjectHotReloadSessionManager
         private readonly Lazy<IProjectHotReloadAgent> _projectHotReloadAgent;
         private readonly Lazy<IHotReloadDiagnosticOutputService> _hotReloadDiagnosticOutputService;
 
-        private HotReloadState? _pendingSessionState = null;
-
-        private ImmutableDictionary<int, HotReloadState> _activeSessions = ImmutableDictionary<int, HotReloadState>.Empty;
+        // Protect the state from concurrent access. For example, our Process.Exited event
+        // handler may run on one thread while we're still setting up the session on
+        // another. To ensure consistent and proper behavior we need to serialize access.
+        private readonly AsyncSemaphore _semaphore = new(initialCount: 1);
```
MiYanni, line 29: This is a great comment to give context. 😊

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -47,6 +49,8 @@ internal class ProjectHotReloadSessionManager : IProjectHotReloadSessionManager
 
         public async Task ActivateSessionAsync(int processId, bool runningUnderDebugger)
         {
+            using AsyncSemaphore.Releaser semaphoreReleaser = await _semaphore.EnterAsync();
```
MiYanni, line 52: This is cool! I hadn't seen something like this before where you use the lifetime of the variable (IDisposable) to be the unlocking mechanism. Pretty nifty.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
tmeschter: I wrote three drafts of that comment and I find this one to be vastly better than the first and yet still not as good as I'd like it. 😭🤣

drewnoakes:
>> Is the interlocked code still needed, given the semaphore?
>
> No. I attempted to remove it; did I miss something?

Sorry, my mistake.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
drewnoakes: Yeah it's nice. Beyond the convenience, it also ensures that any exception thrown within the scope will release the semaphore, so that it doesn't remain locked indefinitely after an error occurs.

PR #7577, opened 2021-09-10
Define a basic compliance pipeline

tmeschter:
Define a basic pipeline that runs PoliCheck.

The point here is just to get PoliCheck to run. Later commits will expand on this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7577)

### DIFF
--- a/build/ci/compliance.yml
+++ b/build/ci/compliance.yml
@@ -11,7 +11,7 @@ resources:
 - repo: self
   clean: true
 queue:
-  name: VSEngSS-MicroBuild2019
+  name: VSEngSS-MicroBuild2019-1ES
   demands: Cmd
   timeoutInMinutes: 90
 variables:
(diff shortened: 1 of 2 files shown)


build/ci/compliance.yml
```diff
@@ -0,0 +1,39 @@
...
+
+#
+# NOTE: triggers for this build are defined in the Web UI instead of here in the YAML file so they
+#       apply to all branches.
+
+resources:
+- repo: self
+  clean: true
+queue:
+  name: VSEngSS-MicroBuild2019
```
swesonga, line 14: Should this use the new 1ES pool?

swesonga (approved):

build/ci/compliance.yml
```diff
@@ -0,0 +1,47 @@
...
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+
+steps:
+- script: echo Hello, world!
+  displayName: 'Run a one-line script'
+
+- script: |
+    echo Add other tasks to build, test, and deploy your project.
+    echo See https://aka.ms/yaml
+  displayName: 'Run a multi-line script'
```
MiYanni, line 30: Looks like generic template text that doesn't need to be here.

MiYanni (commented):
I fully expect that we'll also need exception files (those that contain things to skip) for these checks. Every repo generally has a few things that need to be ignored.

build/ci/compliance.yml
tmeschter: Yeah, it will eventually.

build/ci/compliance.yml
tmeschter: Yeah, I left those there at first just to make sure something was running.

build/ci/compliance.yml
```diff
@@ -0,0 +1,50 @@
...
+
+- task: APIScan@2
+  inputs:
+    softwareFolder: '$(Build.SourcesDirectory)\artifacts'
+    softwareName: 'Dotnet-Project-System'
+    softwareVersionNum: '17.0'
+    softwareBuildNum: '$(Build.BuildId)'
+    symbolsFolder: 'SRV*http://symweb'
+
+- task: TSAUpload@2
```
MiYanni, line 47: What is TSAUpload? I'm having difficulty finding docs on it.

tmeschter: It picks up the output files from the previous stages and uses them to auto-generate bugs.

jjmew (changes requested):
:clock1:

jjmew (approved):
:shipit:

PR #7579, opened 2021-09-10
Replace 'white list' with 'safe list'

drewnoakes:
[AB#1398006](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1398006)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7579)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceEditorView.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceEditorView.vb
@@ -3268,7 +3268,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
                 If Not String.IsNullOrEmpty(fileExtension) Then
                     Dim isSafe As Boolean = False
 
-                    ' first of all, we check the white list for the resource type. If the extension matches one of them, we won't pop up a warning dialog.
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Editors/ResourceEditor/ResourceEditorView.vb
```diff
@@ -3268,7 +3268,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
                 If Not String.IsNullOrEmpty(fileExtension) Then
                     Dim isSafe As Boolean = False
 
-                    ' first of all, we check the white list for the resource type. If the extension matches one of them, we won't pop up a warning dialog.
+                    ' first of all, we check the safe list for the resource type. If the extension matches one of them, we won't pop up a warning dialog.
```
MiYanni, line 3271: What about 'allow list'? https://en.wikipedia.org/wiki/Whitelisting#Controversy

drewnoakes: I would ordinarily call it that, but it's referred to as a safe list in subsequent code so I feel that's a more consistent name here.

MiYanni (approved):

PR #7581, opened 2021-09-12
Overload WaitForFirstSnapshot for infinite timeout

drewnoakes:
All but one caller to `ILaunchSettingsProvider.WaitForFirstSnapshot` passes an infinite timeout. The contract for this method is that an infinite timeout implies the return value is non-null. However the C# nullable reference types feature cannot know this, and so each call included an assertion to appease the nullable analysis.

This change overloads this method with a form that does not accept a timeout, and which returns a non-nullable value. This simplifies calling code a little, and encourages future consumers of the interface to not pass a timeout. As we saw in #7542 ([AB#1370386](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1370386), [AB#1386457](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1386457)) timeouts can be fragile and should be avoided.

This overload also accepts a cancellation token, encouraging calling code to consider cancellation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7581)


MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/GlobalSuppressions.cs
```diff
@@ -21,4 +21,6 @@
...
 [assembly: SuppressMessage("Style", "VSTHRD200:Use \"Async\" suffix for async methods", Justification = "By design", Scope = "member", Target = "~M:Microsoft.VisualStudio.Threading.TaskResult.Null``1~System.Threading.Tasks.Task{``0}")]
 [assembly: SuppressMessage("Style", "VSTHRD200:Use \"Async\" suffix for async methods", Justification = "By design", Scope = "member", Target = "~M:Microsoft.VisualStudio.Threading.TaskResult.EmptyEnumerable``1~System.Threading.Tasks.Task{System.Collections.Generic.IEnumerable{``0}}")]
+[assembly: SuppressMessage("Style", "VSTHRD200:Use \"Async\" suffix for async methods", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.VisualStudio.ProjectSystem.Debug.LaunchSettingsProviderExtensions.WaitForFirstSnapshot(Microsoft.VisualStudio.ProjectSystem.Debug.ILaunchSettingsProvider,System.Threading.CancellationToken)~System.Threading.Tasks.Task{Microsoft.VisualStudio.ProjectSystem.Debug.ILaunchSettings}")]
```
swesonga, line 24: >WaitForFirstSnapshot [](http://example.com/codeflow?start=224&length=20)

Should this be a new public interface if we want to encourage the correct use of the API? Or couldn't this extension actually have the async suffix since it is only used internally?

drewnoakes: The goal was for this to be an overload that pops up in completion alongside the existing one.

Adding a suffix to the new method would make it look like there was both a blocking and non blocking version, which is not true.

Also there are `When` method in the BCL that return take without `Async` suffixes.

Versioning the interface would require casting or changing the declared type in most usages. We may have to do that in future.

swesonga (approved):

melytc (approved):

Issue #7587, opened 2021-09-13
WAPPROJ projects do not appear in startup project combo-box


srdjanjovcic:
## Visual Studio Version
16.11

## Summary
When project system enumerates projects which are startable, it excludes WAPPROJ projects which are clearly startable/debuggable.

## Steps to Reproduce
1. Create a solution with a WAPPROJ and two other executable projects.
2. Notice that start project drop-drown does not contain WAPPROJ name.

## Expected Behavior
Startup project drop-down should contain WAPPROJ alongside other executable projects.

## Actual Behavior
Startup project drop-down does not contain WAPPROJ.

## User Impact
User is not able to easily select WAPPROJ as a startup project.

## Investigation
I debugged it, and it appears to be regression from Project-System change [d10b546](https://github.com/dotnet/project-system/commit/d10b546): _Add handshake for startup project list._ WAP project system is based on CPS, and it does implement two `IDebugLaunchProvider`’s , but none of them implements `IStartupProjectProvider`, so check in `IsDebuggableAsync` function returns always false.

My providers are implemented in code outside of `Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll` (where interface resides), and interface in question is internal. They derive from `DebugLaunchProviderBase`, which is implemented in `Microsoft.VisualStudio.ProjectSystem.VS.dll`, which also has no access to that interface.

While simplest solution would be to make interface `IStartupProjectProvider` public, so it can be implemented by WAP's debug providers, it seems that ultimate cause is wrong assumption in regressing commit: `IsDebuggableAsync` implementation makes wrong assumption that only providers which implement that interface are debuggable. 

It may be prudent to explore a solution where old code (using `CanLaunchAsync`) is retained as a fallback if none of providers implement `IStartupProjectProvider`.

srdjanjovcic:
How is lifecycle of project-system managed? When this change will end up in Visual Studio?

drewnoakes:
@srdjanjovcic we generally insert to VS main every day.

@ocalles you didn't respond to the comments in #7588 before merging. What is the reasoning behind not including a more robust fix here?

Reopening this issue until we have clarity.

ocallesp:
The final fix was the suggested change.

drewnoakes:
My mistake, sorry.

PR #7588, opened 2021-09-14
Fix WAPPPROJ not in Startup Project

ocallesp:
This will allow external project like WAP to implement this interface and make them visible in the startup project drop-down.

Using `CanLaunchAsync `as a fallback if none of providers implement `IStartupProjectProvider`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7588)


drewnoakes:
@ocalles what do you think about @srdjanjovcic's suggestion in #7587 as an alternative to this PR:

> It may be prudent to explore a solution where old code (using CanLaunchAsync) is retained as a fallback if none of providers implement IStartupProjectProvider.

Can we find out how many providers exist that do not implement this interface? We should not regress the behaviour for those providers, and that suggestion seems a reasonable solution to that.

srdjanjovcic (approved):

melytc (approved):

PR #7605, opened 2021-09-16
Display message about Hot Reload being disabled

tmeschter:
Builds on #7604.

Display a message in the Hot Reload output pane when it is disabled because startup hooks are disabled:

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7605)


tmeschter:
Holding this one as a draft until we can get an exception for the new string, which will not be localized in time for 17.0.

tmeschter:
Looks like we can probably get a translation in time, actually.

drewnoakes (approved):

PR #7623, opened 2021-09-20
Update documentation

tmeschter:
Update the documentation on adding a new launch profile kind to reflect the fact that the use of the `ExportPropertyXamlRuleDefinitionAttribute` de facto requires the `OrderAttribute` as well since the metadata is pulled into an `OrderPrecedenceImportCollection`. Without the `OrderAttribute` the item isn't pulled into the collection at all.

I'm not entirely sure why `ExportPropertyXamlRuleDefinitionAttribute` data is being ordered to begin with. The only thing I can think of is to provide a mechanism for effectively adjusting the `AppliesTo` attribute.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7623)

### DIFF
--- a/docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
+++ b/docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md
@@ -156,7 +156,7 @@ namespace MyNamespace
             xamlResourceStreamName: "XamlRuleToCode:MyLaunchProfile.xaml",
             context: PropertyPageContexts.Project)]
         [AppliesTo("MyUniqueProjectCapability")]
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):

WilliamXieMSFT:
The static Order class is internal, consider making it a public class?

tmeschter:
> The static Order class is internal, consider making it a public class?

Good point. Let's fix that.

Note the `Order` class is only meant to be used internally; I don't want to expose something publicly if I don't need to, and it isn't really necessary here.

PR #7629, opened 2021-09-21
Add fwlinks

### LINKED ISSUE #7535: Set fwlinks for all HelpUrls
We need fwlinks for several property `HelpUrl`s in our XAML rules.

For cases where we use the same URL on multiple properties, we should create separate fwlinks for each in case we want to point them at different docs in future.

They should be owned by dotnettoolprojfwlink.

Not all of them have TODO comments. We need to search through all of them and add them.

This includes those in #7532 (unmerged at the time this issue was created).

tmeschter:
Fixes #7535.

Update the last few `HelpUrl`s that need fwlinks.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7629)


MiYanni (approved):
Looks good. We should make note somewhere that fwlinks should not contain the `/en-us` section of the URL. I double checked your links just to be sure it wasn't there.

swesonga (approved):

PR #7632, opened 2021-09-22
Annotate IInterceptingPropertyValueProvider MEF contract

drewnoakes:
All interfaces used for MEF exports should be annotated with details of their scope, their provider, and their cardinality.

I discovered this was not annotated while investigating something for the Xamarin folks. Will let CI run the unit tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7632)


swesonga (approved):

tmeschter:
Does CPS need an analyzer that will enforce this?

tmeschter (approved):

drewnoakes:
> Does CPS need an analyzer that will enforce this?

We have unit tests that generally catch this kind of issue. It seems that in this case they failed to do so as `IInterceptingPropertyValueProvider` is exported via `ExportInterceptingPropertyValueProviderAttribute` rather than `ExportAttribute`. We could improve our tests to cover that and other cases.

PR #7636, opened 2021-09-23
Stop Hot Reload on project unload

tmeschter:
Fixes [AB#1397443](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1397443).

Currently, if the user unloads a project or closes the solution while a Hot Reload session is in progress the session remains, but we now lack any way to interact with it. In some situations this "zombie" session can block changes to later sessions or cause other problems.

The fix here is to end a project's active sessions when the project is unloaded and the `ProjectHotReloadSessionManager` is disposed. Note that we don't wait for the call to `StopSessionAsync` to complete as we just need the session to be stopped sometime soon.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7636)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
@@ -15,7 +15,7 @@ using Microsoft.VisualStudio.Threading;
 namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
 {
     [Export(typeof(IProjectHotReloadSessionManager))]
-    internal class ProjectHotReloadSessionManager : IProjectHotReloadSessionManager
+    internal class ProjectHotReloadSessionManager : OnceInitializedOnceDisposedAsync, IProjectHotReloadSessionManager
     {
         private readonly UnconfiguredProject _project;
         private readonly IProjectFaultHandlerService _projectFaultHandlerService;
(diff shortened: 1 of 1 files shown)


swesonga (approved):
Out of curiosity, why did you choose to use a semaphore with capacity 1 instead C# lock statements in this implementation?

tmeschter:
`await` can't be used within a `lock` block--if control is returned to the caller (e.g. if the awaited `Task` is not complete) the lock will be released.

BillHiebert (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -127,6 +129,20 @@ public async Task<bool> TryCreatePendingSessionAsync(IDictionary<string, string>
...
 
+        private async Task OnUnconfiguredProjectUnloadingAsync(object? sender, EventArgs args)
+        {
+            _project.ProjectUnloading -= OnUnconfiguredProjectUnloadingAsync;
+
+            using AsyncSemaphore.Releaser semaphoreRelease = await _semaphore.EnterAsync();
+
+            foreach (HotReloadState sessionState in _activeSessions.Values)
+            {
+                _ = sessionState.Session?.StopSessionAsync(default);
```
drewnoakes, line 140: Would you want to have any exceptions here logged? If so, you could use the `RunAndForget` extension on `IProjectThreadingService`.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
davkean: You don't need to hook onto Unload, just perform this in your dispose method which will be called on shutdown.

davkean: Components who's lifetimes are tied to the project composition, shut themselves down in Dispose, implement OnceInitializedOnceDisposeAsync.

tmeschter: Even better. That's the piece I was missing.

drewnoakes (approved):

PR #7637, opened 2021-09-24
Avoid disposing DependenciesSnapshotProvider on the UI thread

### LINKED ISSUE #7635: DependenciesSnapshotProvider.DisposeCore blocks UI thread 110ms during closing Roslyn.sln
most time spent to dispose dataflow subscriptions, no need to block the UI thread.

It happens because OnProjectUnloadingAsync is called on the UI thread, and also blocks the unloading process. I guess it just need spin off the task to unsubscribe dataflow.

drewnoakes:
Fixes #7635

Previously we handled the `ProjectUnloading` event, which fired on the UI thread. As called out in #7635, this accounts for 111ms of UI delay when closing Roslyn.sln.

`DependenciesSnapshotProvider` is in the `UnconfiguredProject` scope and will be disposed along with the rest of the project's MEF parts without this special handling. That disposal happens async, on the thread pool.

By removing the `ProjectUnloading` hook, disposal happens a little later, and off the UI thread.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7637)


tmeschter:
I did not realize `ProjectUnloading` was fired on the main thread, which of course has implications for #7636.

swesonga (approved):

PR #7643, opened 2021-09-27
Remove "LanguageVersion" property from GeneralConfiguredBrowseObject

drewnoakes:
We have "LangVersion" here, which matches the actual MSBuild property. I've searched for potential consumers of this alias and haven't found any, so consider this unused.

My understanding is that this was needed prior to fc1f7f691851670909ed38fdfc3b4a4ac79c5e4b. That commit removed use of it from the legacy project property pages.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralConfiguredBrowseObject.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/GeneralConfiguredBrowseObject.xaml
@@ -96,16 +96,6 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <EnumProperty Name="LanguageVersion"
-                Visible="False">
-    <EnumProperty.DataSource>
-      <DataSource HasConfigurationCondition="False"
-                  PersistedName="LangVersion"
-                  Persistence="ProjectFile"
-                  SourceOfDefaultValue="AfterContext" />
-    </EnumProperty.DataSource>
-  </EnumProperty>
-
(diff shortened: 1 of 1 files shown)


swesonga (approved):

davkean:
I see Apex usage, look for ClientProjectProperties.

davkean:
And its part of https://docs.microsoft.com/en-us/dotnet/api/vslangproj80.csharpprojectconfigurationproperties3.languageversion?view=visualstudiosdk-2019#VSLangProj80_CSharpProjectConfigurationProperties3_LanguageVersion, which is the strongly typed version of these properties.

drewnoakes:
The `VSLangProj80.ProjectConfigurationProperties3` implementation looks ok to me.

@davkean what is `ClientProjectProperties`? I can't find any relevant mention of it on //index. Were you referring to the VSLangProj interface?

davkean:
Search using AzureDevOps, index doesn't show all the results.

davkean:
At the end of the day, this is an API that legacy project system provides, and you are removing that. In legacy, CSharpProjectConfigurationProperties represents the same values that can be found on the browse object.

drewnoakes:
> Search using AzureDevOps, index doesn't show all the results.

I can only find references to a type in the old internal Maui/Omni testing framework.

> At the end of the day, this is an API that legacy project system provides, and you are removing that. In legacy, CSharpProjectConfigurationProperties represents the same values that can be found on the browse object.

My goal was to avoid duplicating the property so we don't have both `LangVersion` and `LanguageVersion`. Given the implicit contract here maybe it's better to remove `LangVersion` instead.

PR #7652, opened 2021-10-01
Remove hard-coded top level VS workload IDs

### LINKED ISSUE #7564: Remove hard-coded VS Setup Component IDs
## Visual Studio Version
17.0

## Summary
https://github.com/dotnet/project-system/pull/7432 introduced the [IVisualStudioComponentIdTransformer](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/VisualStudioParentWorkloadTransformer.cs) interface to add top level Visual Studio workloads to the set of components required by MAUI projects. This mapping will be maintained in the SDK now that https://github.com/dotnet/sdk/pull/20200 has been merged. [IVisualStudioComponentIdTransformer ](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/VisualStudioParentWorkloadTransformer.cs) and all its implementations need to be removed in favor of the collection of VS setup component IDs generated by the SDK.

swesonga:
Fixes #7564.

The SDK includes the top level VS workloads with suggested workloads as necessary (as of https://github.com/dotnet/sdk/pull/20200). The SuggestedWorkload rule therefore needs to use that list instead of the mapping hard-coded into the project system. The old VisualStudioComponentId property is still in place in case a user is using an older .NET 6 preview SDK.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7652)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WorkloadDescriptor.cs
```diff
@@ -28,12 +34,12 @@ public WorkloadDescriptor(string workloadName, string visualStudioComponentId)
         /// <summary>
         ///     Gets the Visual Studio setup component ID corresponding to the .NET workload.
         /// </summary>
-        public string VisualStudioComponentId { get; }
+        public ISet<string> VisualStudioComponentIds { get; }
```
drewnoakes, line 37: If this doesn't have to be mutable, use `IReadOnlySet<string>`.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WorkloadDescriptor.cs
swesonga: That is a CPS-internal interface

PR #7654, opened 2021-10-01
Exclude Razor property pages explicitly

### LINKED ISSUE #7627: [Loc] New Property Pages: Razor Properties are not translated
Related: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1308594/

If you create a .NET Core WebApplication, the new property pages do not have translations for Razor Properties. Here, I was testing in Korean.
image

Tested after this commit: https://github.com/dotnet/project-system/commit/abb6acc50b01430f9c2e130d98cd0b3c853f216c

drewnoakes:
Fixes #7627

Some versions of the .NET SDK include a Razor property page that appears in the UI. This page is not intended for display.

We cannot remove this page from existing versions of the SDK, so have to explicitly exclude it from query results so that it doesn't appear in any UI.

## Before

image

## After

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7654)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
@@ -124,6 +124,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
                     if (projectCatalog.GetSchema(schemaName) is Rule rule
                         && !rule.PropertyPagesHidden)
                     {
+                        if (rule.Name == "RazorGeneral" || rule.Name == "RazorExtension")
+                        {
+                            // Some versions of the .NET SDK include a Razor property page that appears
+                            // in the UI. This page is not intended for display.
+                            //
+                            // We cannot remove this page from existing versions of the SDK, so have to
+                            // explicitly exclude it from query results so that it doesn't appear in any
+                            // UI.
+
+                            continue;
+                        }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
```diff
@@ -124,6 +124,18 @@ IEnumerable<IEntityValue> createPropertyPageValuesAsync()
                     if (projectCatalog.GetSchema(schemaName) is Rule rule
                         && !rule.PropertyPagesHidden)
                     {
+                        if (rule.Name == "RazorGeneral" || rule.Name == "RazorExtension")
```
swesonga, line 127: Should this use StringComparers.RuleNames?

swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageDataProducer.cs
MiYanni: Instead of a separate if-statement, it might make sense to have a static list of ignored rule names and have the first if-statement above this one do an `&& !ignoredRuleNames.Contains(rule.Name)` or something of the sort. Then, we're just filtering them out as part of the already existing workflow.

MiYanni (approved):

PR #7663, opened 2021-10-05
Fix FUTD corner-case when there are zero files.

ocallesp:
When a project contains only one file and it is deleted, FUTD will not marked as deleted in changes.Add().

With this change FUTD will take into consideration also if there are any changes and not only the count

FUTD
.


drewnoakes:
Can you please add a unit test for this case? We have good test coverage for this feature and it's easy to regress, so the tests add a lot of value here.

Issue #7642, opened 2021-09-27
Embedding package files through the project properties UI doesn't work when the package files are in the same directory as the project file


chgill-MSFT:
## Visual Studio Version

I am using Visual Studio 2022 preview 5, however I believe the bug repros for all VS 2022 previews.

## Summary

Embedding package files (README, package icon, license) through the package properties UI is broken when the files are in the same directory as the project file. 

Embedding package elements through the package properties UI works as expected when the files are outside the project file directory.

## Steps to Reproduce

1. Create a class library project in VS 2022.
2. Add a README.md file to the same directory as the project file.
3. Add the README.md file to the package through the project properties UI. This generates the following XML in the project file:

`<PackageReadmeFile>README.md</PackageReadmeFile>`

and

```xml
  <ItemGroup>
    <None Update="README.md">
      <Pack>True</Pack>
    </None>
  </ItemGroup>
```

4. Attempt to pack the package
5. Pack fails with the following error even though the file is there:

`NU5046: The readme file 'README.md' does not exist in the package.`

**Video from @timheuer:**

![2021-09-24_14-59-59](https://user-images.githubusercontent.com/4821/134744065-77d8dbee-eb9a-4961-b0a4-9f1bdac43b02.gif)

## Expected Behavior

The generated XML should be:

`<PackageReadmeFile>README.md</PackageReadmeFile>`

and

```xml
  <ItemGroup>
    <None Include="README.md">
      <Pack>True</Pack>
      <PackagePath></PackagePath>
    </None>
  </ItemGroup>
```

This is already the behavior when embedded files are outside the project file directory.

## Actual Behavior

When the package files are in the project file directory, the generated XML changes from `<None Include="README.md">` into `<None Update="README.md">` and omits the `<PackagePath>` node. This causes pack to fail.

`<PackageReadmeFile>README.md</PackageReadmeFile>`

and

```xml
  <ItemGroup>
    <None Update="README.md">
      <Pack>True</Pack>
    </None>
  </ItemGroup>
```

Another example using icons in the following image comparing VS 16.11 and VS 17.0 preview 5

<img src="https://user-images.githubusercontent.com/15097183/134979491-b35bab4b-974a-46e6-869c-88612146f5dc.png" width="500">

## User Impact

Embedding package files when they are in the same directory as the project file is broken in VS 2022. This is a very common scenario, especially for README files since each package in a repo will likely have a package-specific README in it's project file directory.

[...]

PR #7667, opened 2021-10-06
Fix missing PackagePath node

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7642

Changes the default PackagePath to `\` which will always explicitly add the `<PackagePath>` node, resolving `pack` command issues. Both empty PackagePath and `\` result in the root directory of the package being used.

More details can be found in the issue. In summary, when files are in the project directory or beneath it, they are automatically added via globbing. This causes the `None` item to change into an `Update` item. *I believe* this process (which is part of `AddAsync`) does some optimization, removing empty nodes. In this case for the `pack` operation, having an empty value is significant as it replaces the default of `content;contentFiles`. That means it would place the file into those directories, and it would fail to find it in the appropriate directory, since they're expecting the path to start from the root of the package. To resolve this, instead of an empty string (which indicates the root directory), we change to `\` which also indicates the root directory of the package. This won't optimize out the `<PackagePath>` node, and thus replace the default `content;contentFiles` value.


MiYanni:
For now, I've targeted this against `main`, but we can change it to `dev17.0.x` if we think this issue is significant enough since it is a breaking piece of functionality. The workaround for customers would be to add `<PackagePath><\PackagePath>` into the item.

drewnoakes:
I would vote for taking this change to 17.0. The fix is straighforward, and 17.0 is LTS.

tmeschter:
I also vote for 17.0.

melytc:
+1 to taking it to 17.0

PR #7668, opened 2021-10-06
Up-to-date check logs newest input

drewnoakes:
At its core, the fast up-to-date check finds the oldest output, and the newest input, and compares them.

Previously, when we consider the project up-to-date, we logged only the earliest output but not the newest input. It can be useful to see this, especially when debugging multi-project build logs.

This change adds such logging.


drewnoakes:
cc @rainersigwald, as this came up today while we were diagnosing an issue.

rainersigwald (approved):

PR #7670, opened 2021-10-06
Add ProjectHotReloadSessionManager tests

tmeschter:
Add some unit tests for the `ProjectHotReloadSessionManager`, specifically around creating pending sessions. Right now we largely check if the project has the "SupportsHotReload" capability, but there are a couple other things we should look at as well. Before I add those, however, I want to have some unit testing infrastructure in place.


tmeschter:
And just to emphasize the importance of unit tests, this turned up a minor bug--if the `TargetFrameworkVersion` property isn't defined we shouldn't create a session, but we will. In practice this isn't much of an issue because you're likely to have lots of other problems to deal with if that property isn't defined, but I'll fix it in a separate change.

PR #7678, opened 2021-10-11
Rename extensions to match type name

MiYanni:
The extension file/class should match the name of the class it is extending. Here, `IProperty` and `IProjectTreeProvider`, thus changing the name and the extension method class name. There are some .NET cultures that don't like this because the `I` is supposed to represent an interface, and this is not an interface. However, we follow the former convention for other interfaces, such as `IProjectTreeExtensions`, `IDependencyModelExtensions`, and `IVsHierarchyItemExtensions`.

At first, I thought the `PropertyExtensions.cs` file might have been a general 'properties' set of extension methods, but it was only specifically for the `IProperty` interface.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7678)


MiYanni:
It seems that a bunch of files in our repo follow this convention. So, there is a mix of 2 different conventions in our repo. Either I should change it to one way or change it to the other way.

PR #7681, opened 2021-10-12
Identify and report broken incremental builds

drewnoakes:
Relates to [AB#1419993](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1419993), [AB#1419997](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1419997), [AB#1419998](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1419998) and [AB#1419999](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1419999).

Some percentage of users have broken incremental builds. Some further percentage of those users will not know this is the case, or know how to address the problem. We would like to understand how widespread broken incremental builds are, and see whether flagging these issues to users helps them to self-diagnose such problems.

This PR adds the `IncrementalBuildFailureDetector` component which validates projects are up-to-date after building, as they should be. It does this by running the fast up-to-date check when the build completes.

When a project is found to have a broken incremental build, it can be reported in build output and/or via telemetry. We use feature flags to control these behaviours.

We could not reuse the existing up-to-date check API, as it tracks the time at which it was last called. Instead, this change introduces the `IBuildUpToDateCheckValidator` interface which may be safely called at any point. Implementing this interface required only a slight change to the up-to-date check code.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7681)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.cs
@@ -87,7 +87,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Build.Diagnostics
             Assumes.NotNull(_solutionBuildManager);
             Assumes.NotNull(_rdt);
 
-            if (_rdtCookie != 0)
+            if (_sbmCookie != 0)
             {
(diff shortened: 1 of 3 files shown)


drewnoakes:
Feature flags to control this added in https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/357249.

PR #7685, opened 2021-10-12
Add Rules from code

tmeschter:
Define a type that defines `Rule` instances in code (rather than loading them from XAML) and dynamically adds and removes them from the project based on capabilities.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7685)


tmeschter:
This approach has been superceded by `IRuleObjectProvider` (see #7922 for an example of that). Please use that approach instead of what is demonstrated here.

PR #7703, opened 2021-10-15
Up-to-date check ignores None and Content items

### LINKED ISSUE #7700: Adding/removing Content/None items can lead to overbuild
## Visual Studio Version

main

## Summary

The fast up-to-date check tracks changes to project items. When it sees one, it remembers when it happens and uses that information in future up-to-date decision making.

This item tracking should not apply to `None` or `Content` (and possibly others). When the set of these items change, there is no observable output changes. This can lead to a state where the project continues to build (overbuild) until it is changed or rebuilt.

## Steps to Reproduce

1. Create a console app
2. Add content item `json1.json`
3. Build
4. Build again

## Expected Behavior

Up-to-date on second build.

## Actual Behavior

> 1>FastUpToDate: The set of project items was changed more recently (15/10/2021 2:29:29 PM) than the earliest output 'C:\repos\ConsoleApp341\ConsoleApp341\obj\Debug\net6.0\ConsoleApp341.pdb' (15/10/2021 2:29:11 PM), not up to date.

## User Impact

MSBuild will be called for any project in this state until the project itself is changed in a way that triggers a change in outputs, or the project is rebuilt.

Users may not be aware of this state (although we are working on improving that, in a separate effort).

drewnoakes:
Fixes #7700 

This change fixes a potential overbuild issue where adding or removing a `None` or `Content` item in the project would mark the items as changed at that point in time, with the expectation that a subsequent build would produce newer outputs. This is not the case for `None` and `Content` items. If any such items do behave that way in a given project, they should be modelled as `UpToDateCheckInput` or `UpToDateCheckBuilt` items instead.

This change will also cause the up-to-date check to hold on to less in-memory data. For projects with a lot of `None` or `Content` items, this may have a positive effect.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7703)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/ProjectItemsSchema.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Items/ProjectItemsSchema.xaml
@@ -109,8 +109,8 @@
     DisplayName="Resource"
     ItemType="Resource" />
 
-  <ItemType Name="None" DisplayName="None"/>
-  <ItemType Name="Content" DisplayName="Content" />
+  <ItemType Name="None" DisplayName="None" UpToDateCheckInput="False" />
+  <ItemType Name="Content" DisplayName="Content" UpToDateCheckInput="False" />
   <ItemType Name="EmbeddedResource" DisplayName="Embedded resource" />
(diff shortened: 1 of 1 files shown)


drewnoakes:
Going a different route with this.

PR #7713, opened 2021-10-21
Use asynchronous batch disposal

drewnoakes:
Contributes to [AB#1370097](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1370097).
Replaces #7506.

The previous `EndBatch` method internally uses blocking locks to protect access to its solution-level workspace data. When opening a solution with many projects, each project make requests to update workspaces. Those updates pass through this interface so that we can serialize them in order to reduce blocking lock contention on the thread pool from concurrent update requests.

Using the `EndBatchAsync` allows calling threadpool threads to make progress on other useful work while waiting for this shared resource.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7713)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -45,7 +45,7 @@
     <PackageReference Update="Microsoft.VisualStudio.DataDesign.Common"                               Version="17.0.0-preview-2-31223-026"/>
     <PackageReference Update="Microsoft.VisualStudio.Data.Services"                                   Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VisualStudio.DataTools.Interop"                               Version="17.0.0-preview-2-31223-026" />
-    <PackageReference Update="Microsoft.VisualStudio.Debugger.Contracts"                              Version="17.2.0-beta.21378.1" />
+    <PackageReference Update="Microsoft.VisualStudio.Debugger.Contracts"                              Version="17.2.0" />
     <PackageReference Update="Microsoft.VisualStudio.Designer.Interfaces"                             Version="17.0.0-preview-2-31223-026" />
     <PackageReference Update="Microsoft.VisualStudio.ImageCatalog"                                    Version="17.0.0-previews-2-31421-281" />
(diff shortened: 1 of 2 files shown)


jasonmalinowski:
@drewnoakes Change has been merged on the Roslyn side; hopefully we'll have 100% fewer deadlocks which block insertion this time. :smile:

jasonmalinowski (approved):
:shipit:, as the Roslyn support has merged internally. That said you _may_ want to bump to newer packages just to make sure nothing strange happened -- otherwise you're technically still using bits from before the initial revert and then the revert of the revert.

jasonmalinowski:
@drewnoakes Not sure why an earlier version didn't have it, but the latest version 4.1.0-2.21558.8 does have it. And by "latest" I mean it was, as a actual complete coincidence (no joke!), uploaded four minutes prior to me checking. :smile:

drewnoakes:
Thanks @jasonmalinowski.

drewnoakes:
@jasonmalinowski as called out in #7748 the Roslyn side does not appear to have inserted as expected, leading to `MethodMissingException`. I've reverted this PR in #7748. Do we understand why the method would not be available in the latest VS main?

> Not sure why an earlier version didn't have it, but the latest version 4.1.0-2.21558.8 does have it.

I noticed locally that running a clean/restore/build with 4.1.0-2.21558.8 worked, but bumping that to 4.1.0-2.21558.*9* caused the compilation to fail. The feed appears to be going backward and forward. Are there two branches publishing to the same feed or inserting to VS main?

jasonmalinowski:
@dotnet/roslyn-infrastructure do we have a publishing configuration issue going on somewhere?

drewnoakes:
Updating to today's main, resetting my experimental hive, `git clean`ing, doing a full build, then testing, shows this works now.

I've reinstated the change in #7757.

PR #7739, opened 2021-11-03
Remove user input from telemetry event.

melytc:
Addresses [AB#1417218](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1417218).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7739)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
+++ b/src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
@@ -2453,7 +2453,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
                 If FixedName = "" Then
                     'ResGen wasn't able to create a valid identifier out of the ID (e.g. something like "$")
                     If CheckForFatallyInvalidIDs Then
-                        Exception = NewException(My.Resources.Microsoft_VisualStudio_Editors_Designer.GetString(My.Resources.Microsoft_VisualStudio_Editors_Designer.RSE_Err_BadIdentifier_2Arg, TrimmedName, FindInvalidCharactersInIdentifier(TrimmedName, CodeDomProvider)), HelpIDs.Err_InvalidName)
+                        Exception = NewException(My.Resources.Microsoft_VisualStudio_Editors_Designer.GetString(My.Resources.Microsoft_VisualStudio_Editors_Designer.RSE_Err_BadIdentifier_2Arg, FindInvalidCharactersInIdentifier(TrimmedName, CodeDomProvider)), HelpIDs.Err_InvalidName)
                         Return False
                     End If
                 Else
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editors/ResourceEditor/Resource.vb
```diff
@@ -2453,7 +2453,7 @@ Namespace Microsoft.VisualStudio.Editors.ResourceEditor
                 If FixedName = "" Then
                     'ResGen wasn't able to create a valid identifier out of the ID (e.g. something like "$")
                     If CheckForFatallyInvalidIDs Then
-                        Exception = NewException(My.Resources.Microsoft_VisualStudio_Editors_Designer.GetString(My.Resources.Microsoft_VisualStudio_Editors_Designer.RSE_Err_BadIdentifier_2Arg, TrimmedName, FindInvalidCharactersInIdentifier(TrimmedName, CodeDomProvider)), HelpIDs.Err_InvalidName)
+                        Exception = NewException(My.Resources.Microsoft_VisualStudio_Editors_Designer.GetString(My.Resources.Microsoft_VisualStudio_Editors_Designer.RSE_Err_BadIdentifier_2Arg, FindInvalidCharactersInIdentifier(TrimmedName, CodeDomProvider)), HelpIDs.Err_InvalidName)
```
drewnoakes, line 2456: I think this will throw a `FormatException`. The `RSE_Err_BadIdentifier_2Arg` resource string has a value resembling:

> The name of the resource '{0}' cannot be used as a valid identifier, because it contains one or more invalid characters: '{1}'.  Please remove or replace those characters and try again.

Notice the `_2Arg` suffix, indicating it needs two arguments.

The call to `Microsoft_VisualStudio_Editors_Designer.GetString` is essentially a call to `string.Format`. Because the resource string contains both `{0}` and `{1}`, it requires two arguments. By removing one, the `string.Format` call will fail.

You can try this out [here](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AECMA6XATgApcAGPAMQHtlIFiAiAbzIF8ACZnNxuDnAEpBAbiA==).

---

I think this fix is not sufficient to completely exclude customer data. Consider the case where the identifier was entirely non-ASCII (perhaps in Cyrillic, Arabic, CJK, ...). My understanding is that `FindInvalidCharactersInIdentifier` would return the original identifier unchanged in that case, and it would still appear in the exception message.

This exception message is likely also presented to the user to help them diagnose a problem. It might appear in the Error List, for example. Removing the identifier from such messages would be a set back.

---

Maybe the best solution here is to not include the exception's `Message` in the telemetry data at all. That way we avoid any kind of identifying information from this or other exceptions.

If we wish to classify exceptions, we should use hard coded identifiers (perhaps an enum) for this. I doubt that's necessary however.

jjmew: We don't look at this telemetry so I wonder if we even need to keep it.

melytc: I synced up with Drew and decided that's best not to send the telemetry event at all, but keep the exception as it is to ensure the user gets the full information. Thanks!

drewnoakes (approved):

drewnoakes:
Merging on @melytc's behalf due to OOF.

PR #7745, opened 2021-11-08
Flush Unshipped API contents into Shipped on build

MiYanni:
Added an after-build target to flush the Unshipped API contents into Shipped. This applies to projects in the `src` directory. This means we no longer need to manually flush these files.

Originally, I wanted to deleted the `PublicAPI.Unshipped.txt` file. Unfortunately, the analyzer (PublicApiAnalyzers) requires the file to be there:
```
1>------ Rebuild All started: Project: Microsoft.VisualStudio.ProjectSystem.Managed, Configuration: Debug Any CPU ------
1>CSC : warning AD0001: Analyzer 'Microsoft.CodeAnalysis.PublicApiAnalyzers.DeclarePublicApiAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
1>CSC : error CS2001: Source file 'C:\Code\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed\PublicAPI.Unshipped.txt' could not be found.
1>Done building project "Microsoft.VisualStudio.ProjectSystem.Managed.csproj" -- FAILED.
```
Instead, we just clear the contents of the file.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7745)

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -2,4 +2,13 @@
 <Project>
   <Import Project="..\Directory.Build.targets" />
   <Import Project="..\build\import\VisualStudio.XamlRules.targets"/>
+
+  <!--
+    Copies the contents of PublicAPI.Unshipped.txt into PublicAPI.Shipped.txt on build.
+    It assumes that the project directory contains these files (PublicApiAnalyzers also makes this assumption).
+    Uses AppendAllText instead of Add-Content since that cmdlet adds an empty line at the end of the file.
+  -->
+  <Target Name="FlushUnshippedAPI" AfterTargets="Build">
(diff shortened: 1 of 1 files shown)


ocallesp (approved):

MiYanni:
I wanted to double check that this works as expected. I ran through the instructions here: https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md
1. Added a new public class with a public method.
2. Saw the warning in VS stating the public class and public method were not declared in the txt document.
3. Clicked on the lightbulb for both the class and method, and used the 'Add [name] to public API' option.
4. Observed both the class and method were declared in the `PublicAPI.Unshipped.txt`.
5. Built the project.
6. Observed that the class and method were no longer in the `PublicAPI.Unshipped.txt` and now in the `PublicAPI.Shipped.txt`.

PR #7749, opened 2021-11-09
Handle absent IBuildUpToDateCheckValidator export

### LINKED ISSUE #7747: NullReferenceException when building vcxproj
## Visual Studio Version

Version 17.1.0 Preview 2.0 [31908.427.main]

## Summary

When building a mixed .NET / C++ solution, a `NullReferenceException` can occur.

## Steps to Reproduce

1. Open a solution having mixed C++ and .NET projects (`\\mlangfs1\public\v-masche\RealWorldSolutions\NETCoreCppCliTestWithNative.zip` is a good example)
2. Build

## Expected Behavior

No bang.

## Actual Behavior

Bang.

```
System.NullReferenceException: Object reference not set to an instance of an object.
at Microsoft.VisualStudio.ProjectSystem.VS.Build.Diagnostics.IncrementalBuildFailureDetector.ProjectChecker.<>c__DisplayClass7_0.<<OnProjectBuildCompleted>g__CheckAsync|1>d.MoveNext()
```

## User Impact

Yellow bar.

drewnoakes:
Fixes #7747

When a solution contains both .NET and C++ projects, some projects may not export an instance of `IBuildUpToDateCheckValidator`. This change handles such cases.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7749)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
@@ -24,7 +24,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Build.Diagnostics
         private sealed class ProjectChecker : IProjectChecker
         {
             private readonly UnconfiguredProject _project;
-            private readonly IActiveConfiguredValue<IBuildUpToDateCheckValidator> _upToDateCheckValidator;
+            private readonly IActiveConfiguredValue<IBuildUpToDateCheckValidator?> _upToDateCheckValidator;
             private readonly IProjectAsynchronousTasksService _projectAsynchronousTasksService;
 
             [ImportMany]
@@ -34,7 +34,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Build.Diagnostics
             public ProjectChecker(
                 UnconfiguredProject project,
                 IVsUIService<SVsFeatureFlags, IVsFeatureFlags> featureFlagsService,
-                IActiveConfiguredValue<IBuildUpToDateCheckValidator> upToDateCheckValidator,
+                IActiveConfiguredValue<IBuildUpToDateCheckValidator?> upToDateCheckValidator,
                 [Import(ExportContractNames.Scopes.UnconfiguredProject)] IProjectAsynchronousTasksService projectAsynchronousTasksService,
                 IVsUIService<SVsOutputWindow, IVsOutputWindow> outputWindow)
             {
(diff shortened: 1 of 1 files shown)


swesonga (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
```diff
@@ -24,7 +24,7 @@ internal sealed partial class IncrementalBuildFailureDetector
         private sealed class ProjectChecker : IProjectChecker
```
MiYanni, line 24: Couldn't there be some kind of `[AppliesTo()]` that would filter this to only apply to .NET projects? I'm assuming our incremental builds code shouldn't apply to `.vcxproj` anyway.

drewnoakes: This is a great question and I wondered the same but didn't manage to work out how that'd work while putting this PR together at the end of my day yesterday. This might be a good topic for tomorrow's CPS meeting. If there's a better way, I will file a follow up PR.

The relevant line of code is here: https://github.com/dotnet/project-system/blob/3ce139b4711f3eb61964749b2ceb1e23d6796019/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.cs#L121

MiYanni: I only asked this because I learned what `AppliesTo` meant from Tom's presentation this morning at the architectural meeting. 😝 You should check out the recording if you haven't already.

tmeschter: You want to tag any .NET-specific exports with: `[AppliesTo(ProjectCapability.DotNet)]`. If they are actually specific to C# and VB, then `[AppliesTo(ProjectCapability.CSharpOrVisualBasic)]`.

drewnoakes: We would just duplicate the capabilities from our IUpToDateCheckProvider, which also excludes shared projects.

PR #7757, opened 2021-11-11
Use asynchronous batch disposal

drewnoakes:
Contributes to [AB#1370097](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1370097).
Replaces #7713, which was reverted in #7748 over concerns about insertion synchronisation.

The previous `EndBatch` method internally uses blocking locks to protect access to its solution-level workspace data. When opening a solution with many projects, each project make requests to update workspaces. Those updates pass through this interface so that we can serialize them in order to reduce blocking lock contention on the thread pool from concurrent update requests.

Using the `EndBatchAsync` allows calling threadpool threads to make progress on other useful work while waiting for this shared resource.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7757)


jjmew (approved):
:shipit:

drewnoakes:
The Roslyn side may be experiencing thread pool starvation following this change (AB#1458498).

PR #7764, opened 2021-11-11
Remove unused VSPackage resources

### LINKED ISSUE #7761: Drop "Core" from template descriptions
https://github.com/dotnet/project-system/blob/9d276d96fd1fcda8e0db586c4c0c9042f74ab879/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSPackage.resx#L120-L149

Should be just ".NET".

drewnoakes:
Fixes #7761 

As pointed out by @Youssef1313, the strings in this resource file refer to ".NET Core", which should be updated to just ".NET". 

Looking into this, these resources do not appear to be used anywhere. I believe this entire file is safe to remove.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7764)


davkean:
There's two sort of resources here:

1) Templates: I don't know what the state of those were, I would search across the tree and GitHub for references to these.
2) The file filters - these are still used in referenced by the project types: https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs#L11.

drewnoakes:
The templates don't seem to use these strings anymore. However the file filters Dave pointed out are still used, so we cannot remove the RESX file completely here.

PR #7771, opened 2021-11-14
Use VS2022 for CI builds

drewnoakes:
We'll want to do this eventually. I am curious if this just works out of the box today.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7771)

### DIFF
--- a/build/ci/compliance.yml
+++ b/build/ci/compliance.yml
@@ -11,7 +11,7 @@ resources:
 - repo: self
   clean: true
 queue:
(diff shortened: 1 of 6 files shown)


drewnoakes:
End result: The first commit only impacts official internal builds, not GitHub PR builds, so cannot be tested here. Our GitHub PR builds override the image name, and changing from 2019 to 2022 doesn't work. I'm not going to investigate this further now.

PR #7779, opened 2021-11-19
Update README.md

timheuer:
Fix typo on VS version

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7779)

### DIFF
--- a/docs/repo/property-pages/README.md
+++ b/docs/repo/property-pages/README.md
@@ -1,6 +1,6 @@
 ﻿# Project Property Pages
 
-This documentation details the updated Project Properties UI and associated back end, added to Visual Studio in 2022.
+This documentation details the updated Project Properties UI and associated back end, first released in Visual Studio 2022.
 
 ## Goals
 
--- a/docs/repo/property-pages/README.md
+++ b/docs/repo/property-pages/README.md
@@ -1,6 +1,6 @@
 ﻿# Project Property Pages
 
-This documentation details the updated Project Properties UI and associated back end, added to Visual Studio in 2021.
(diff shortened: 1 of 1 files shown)


docs/repo/property-pages/README.md
```diff
@@ -1,6 +1,6 @@
 ﻿# Project Property Pages
 
-This documentation details the updated Project Properties UI and associated back end, added to Visual Studio in 2021.
```
jjmew, line 3: I think this comment is correct. This is not talking about the version for Visual Studio but the timeframe of when it was added

docs/repo/property-pages/README.md
```diff
@@ -1,6 +1,6 @@
...
+This documentation details the updated Project Properties UI and associated back end, added to Visual Studio in 2022.
```
tmeschter, line 3: Talking about the product version is probably more useful than the year but we want to be clear on which one we mean.

```suggestion
This documentation details the updated Project Properties UI and associated back end, first released in Visual Studio 2022.
```

docs/repo/property-pages/README.md
timheuer: It's confusing as it's read -- we never talk about timeframes we talk about when/what products they are introduced.

jjmew: The way the updated version reads makes sense

MiYanni (approved):

drewnoakes (approved):

jjmew (approved):
:shipit:

PR #7782, opened 2021-11-22
Fix data flows in LaunchSettingsProvider

tmeschter:
Related to [bug 1427295](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1427295).

The `ILaunchSettingsProvider` interface exposes `ILaunchSettings` to data flow via the `IReceivableSourceBlock<ILaunchSettings> SourceBlock { get; }`. The implementation (`LaunchSettingsProvider`) combines a couple other data flows (one for project capabilities and one for the active launch profile, stored in the .user file) plus changes to the individual launch profiles to create this data flow.

A problem here is that it doesn't "join" the other data flows it uses, and doesn't implement the necessary interfaces for its own data flow block to be "joinable". This means that the data flows cannot coordinate their asynchronous tasks, and you can easily end up in a situation where a component on the UI thread is waiting for a data flow to produce a value, that data flow is waiting on yet another data flow, and so on until you reach a data flow that is trying to schedule work to the UI thread to compute its value. And now you have a deadlock.

This change implements the fix in several parts. First, it updates `LaunchSettingsProvider` to derive from `ProjectValueDataSourceBase<T>`, a CPS utility type specifically for situations where you need to expose a new data source via data flow (note that if you want to create a new data flow be combining _existing_ data flows you want to use `ChainedProjectValueDataSourceBase<T>` instead). This provides the infrastructure needed to join the underlying data flows for project capabilities and the active launch profile, and allow anyone consuming the launch settings to join that data flow.

One wrinkle is that deriving from `ProjectValueDataSourceBase<ILaunchSettings>` means we are exposing a member:

```csharp
IReceivableSourceBlock<IProjectVersionedValue<ILaunchSettings>> SourceBlock { get; }
```

But the `ILaunchSettingsProvider` interfaces defines:

```csharp
IReceivableSourceBlock<ILaunchSettings> SourceBlock { get; }
```

We get around this mismatch by exposing the latter as an explicit interface member implementation, and exposing the former through a new interface `IVersionedLaunchSettingsProvider`. Any consumer that needs to join the data flow should switch to the new interface, but all existing consumers of `ILaunchSettingsProvider` are unaffected.

Second, it runs the logic to produce new `ILaunchSettings` within the data flow with a `JoinableTaskFactory.RunAsync` to ensure this work is coordinated with other tasks.

Third, it updates `DebugProfileDebugTargetGenerator` to use the new interface and properly join the data flow.

Also note that the interal implementation of `ILaunchSettings` already has a version number. We needed this to support the Project Query API, but we didn't implement it in the idiomatic way by wrapping the `ILaunchSettings` in an `IProjectVersionedValue<T>`. Now that we're doing that anyway, we should look at removing the non-idiomatic implementation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7782)


tmeschter:
Please pick this apart. If anything looks even a little off or you have any question, please let me know. I'm not 100% confident of this fix, and I suspect there are some important (though maybe expensive) refactorings we could do here to bring this all in line with best practices for handling data flows so I welcome all feedback.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider.cs
```diff
@@ -13,6 +13,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
     [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
     public interface ILaunchSettingsProvider
     {
+        /// <remarks>
+        /// If the <see cref="ILaunchSettings"/> provided by this block are going to feed
+        /// into another data flow block, strongly consider using <see cref="IVersionedLaunchSettingsProvider"/>
+        /// instead as that provides a joinable variant.
+        /// </remarks>
         IReceivableSourceBlock<ILaunchSettings> SourceBlock { get; }
```
MiYanni, line 21: Would it make sense to mark this as `Obsolete`? Can you use the new SourceBlock as a replacement for this one?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -164,17 +168,38 @@ protected set
             }
         }
 
+        public override NamedIdentity DataSourceKey { get; } = new NamedIdentity(nameof(LaunchSettingsProvider));
+
+        // _nextVersion represents the version we will use in the future, so we need to
+        // subtract 1 to get the current version.
+        public override IComparable DataSourceVersion => _nextVersion - 1;
```
MiYanni, line 175: Is it guaranteed that `_nextVersion` is always `currentVersion + 1`?

tmeschter:
Note that this was accidentally merged (and main subsequently reset to undo that) and has been recreated as https://github.com/dotnet/project-system/pull/7783.

PR #7783, opened 2021-11-22
Fix data flows in LaunchSettingsProvider

tmeschter:
Related to [bug 1427295](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1427295).

The `ILaunchSettingsProvider` interface exposes `ILaunchSettings` to data flow via the `IReceivableSourceBlock<ILaunchSettings> SourceBlock { get; }`. The implementation (`LaunchSettingsProvider`) combines a couple other data flows (one for project capabilities and one for the active launch profile, stored in the .user file) plus changes to the individual launch profiles to create this data flow.

A problem here is that it doesn't "join" the other data flows it uses, and doesn't implement the necessary interfaces for its own data flow block to be "joinable". This means that the data flows cannot coordinate their asynchronous tasks, and you can easily end up in a situation where a component on the UI thread is waiting for a data flow to produce a value, that data flow is waiting on yet another data flow, and so on until you reach a data flow that is trying to schedule work to the UI thread to compute its value. And now you have a deadlock.

This change implements the fix in several parts. First, it updates `LaunchSettingsProvider` to derive from `ProjectValueDataSourceBase<T>`, a CPS utility type specifically for situations where you need to expose a new data source via data flow (note that if you want to create a new data flow be combining _existing_ data flows you want to use `ChainedProjectValueDataSourceBase<T>` instead). This provides the infrastructure needed to join the underlying data flows for project capabilities and the active launch profile, and allow anyone consuming the launch settings to join that data flow.

One wrinkle is that deriving from `ProjectValueDataSourceBase<ILaunchSettings>` means we are exposing a member:

```csharp
IReceivableSourceBlock<IProjectVersionedValue<ILaunchSettings>> SourceBlock { get; }
```

But the `ILaunchSettingsProvider` interfaces defines:

```csharp
IReceivableSourceBlock<ILaunchSettings> SourceBlock { get; }
```

We get around this mismatch by exposing the latter as an explicit interface member implementation, and exposing the former through a new interface `IVersionedLaunchSettingsProvider`. Any consumer that needs to join the data flow should switch to the new interface, but all existing consumers of `ILaunchSettingsProvider` are unaffected.

Second, it runs the logic to produce new `ILaunchSettings` within the data flow with a `JoinableTaskFactory.RunAsync` to ensure this work is coordinated with other tasks.

Third, it updates `DebugProfileDebugTargetGenerator` to use the new interface and properly join the data flow.

Also note that the interal implementation of `ILaunchSettings` already has a version number. We needed this to support the Project Query API, but we didn't implement it in the idiomatic way by wrapping the `ILaunchSettings` in an `IProjectVersionedValue<T>`. Now that we're doing that anyway, we should look at removing the non-idiomatic implementation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7783)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider.cs
```diff
@@ -13,6 +13,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
     [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
     public interface ILaunchSettingsProvider
     {
+        /// <remarks>
+        /// If the <see cref="ILaunchSettings"/> provided by this block are going to feed
+        /// into another data flow block, strongly consider using <see cref="IVersionedLaunchSettingsProvider"/>
+        /// instead as that provides a joinable variant.
+        /// </remarks>
         IReceivableSourceBlock<ILaunchSettings> SourceBlock { get; }
```
MiYanni, line 21: Would it make sense to mark this as `Obsolete`? Can you use the new SourceBlock as a replacement for this one?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -164,17 +168,38 @@ protected set
             }
         }
 
+        public override NamedIdentity DataSourceKey { get; } = new NamedIdentity(nameof(LaunchSettingsProvider));
+
+        // _nextVersion represents the version we will use in the future, so we need to
+        // subtract 1 to get the current version.
+        public override IComparable DataSourceVersion => _nextVersion - 1;
```
MiYanni, line 175: Is it guaranteed that `_nextVersion` is always `currentVersion + 1`?

MiYanni (commented):
(Copied over my comments from the other PR)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/DebugProfileDebugTargetGenerator.cs
```diff
@@ -60,23 +60,23 @@ public override IReceivableSourceBlock<IProjectVersionedValue<IReadOnlyList<IEnu
...
+        private IVersionedLaunchSettingsProvider LaunchSettingProvider { get; }
         private IProjectThreadingService ProjectThreadingService { get; }
 
         /// <summary>
         /// This provides access to the class which creates the list of debugger values..
         /// </summary>
         public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>? options)
             => Task.FromResult<IDynamicEnumValuesGenerator>(
-                new DebugProfileEnumValuesGenerator(LaunchSettingProvider, ProjectThreadingService));
+                new DebugProfileEnumValuesGenerator((ILaunchSettingsProvider)LaunchSettingProvider, ProjectThreadingService));
```
drewnoakes, line 71: Did you consider having `IVersionedLaunchSettingsProvider` derive from `ILaunchSettingsProvider` to avoid this implicit contract between interfaces?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/DebugProfileDebugTargetGenerator.cs
```diff
@@ -60,23 +60,23 @@ public override IReceivableSourceBlock<IProjectVersionedValue<IReadOnlyList<IEnu
...
 
         protected override void Initialize()
         {
-            IPropagatorBlock<ILaunchSettings, IProjectVersionedValue<IReadOnlyList<IEnumValue>>> debugProfilesBlock = DataflowBlockSlim.CreateTransformBlock<ILaunchSettings, IProjectVersionedValue<IReadOnlyList<IEnumValue>>>(
+            IPropagatorBlock<IProjectVersionedValue<ILaunchSettings>, IProjectVersionedValue<IReadOnlyList<IEnumValue>>> debugProfilesBlock = DataflowBlockSlim.CreateTransformBlock<IProjectVersionedValue<ILaunchSettings>, IProjectVersionedValue<IReadOnlyList<IEnumValue>>>(
                 update =>
                 {
                     // Compute the new enum values from the profile provider
-                    var generatedResult = DebugProfileEnumValuesGenerator.GetEnumeratorEnumValues(update).ToImmutableList();
+                    var generatedResult = DebugProfileEnumValuesGenerator.GetEnumeratorEnumValues(update.Value).ToImmutableList();
```
drewnoakes, line 79: Unrelated to this PR, but `ToImmutableArray` might be better here. It has a compact (array-based) representation in memory, which is better if the consumer is not intending to create mutated forms. Given it's exposed as `IReadOnlyList` I think that's a fair assumption in this case.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IVersionedLaunchSettingsProvider.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    /// A variant of <see cref="ILaunchSettingsProvider.SourceBlock"/> that provides
```
drewnoakes, line 8: Did you mean:

```suggestion
    /// A variant of <see cref="ILaunchSettingsProvider"/> that provides
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -164,17 +168,38 @@ protected set
...
         /// The LaunchSettingsProvider sinks 2 sets of information:
         /// 1. Changes to the launchsettings.json file on disk
         /// 2. Changes to the ActiveDebugProfile property in the .user file
         /// </summary>
         protected override void Initialize()
         {
-            // Create our broadcast block for subscribers to get new ILaunchProfiles Information
+            base.Initialize();
+
+            // Create our broadcast block for subscribers to ILaunchSettingsProvider to get new ILaunchProfiles information
```
drewnoakes, line 195: The comment mentioned `ILaunchProfiles` but the code mentions `ILaunchSettings`. Is that right?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -164,17 +168,38 @@ protected set
...
             _broadcastBlock = DataflowBlockSlim.CreateBroadcastBlock<ILaunchSettings>();
             _changedSourceBlock = _broadcastBlock.SafePublicize();
 
+            // Create our broadcast block for subscribers to IVersionedLaunchSettingsProvider to get new ILaunchProfiles information
```
drewnoakes, line 199: Ditto.

drewnoakes:
LGTM but I would hold off until @lifengl gets a chance to review it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchSettingsProvider.cs
tmeschter: In general, the `SourceBlock` member on the new interface could serve as a replacement for this, yes.

The existing member is a part of a public API (`ILaunchSettingsProvider`) and is definitely consumed outside of our repo. Before we mark it as obsolete we need a public interface to replace it (`IVersionedLaunchSettingsProvider` is internal right now) and guidance about how to consume it properly (if you're piping the settings into another data flow block, you need to join the flows properly, etc.).

Tagging this method as obsolete puts a burden on the consumer, and I don't want to do that until we have those other pieces in place.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Yes.

Note that I would like to fix this up so that we're no longer embedding the version in the `ILaunchSettings` implementation, but getting it from the `IProjectVersionedValue` everywhere we need it. That is also beyond the scope of this change.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IVersionedLaunchSettingsProvider.cs
tmeschter: No, because right now it's really only a replacement for the one method, not the entire interface. If I update the new interface to derive from `ILaunchSettingsProvider` then I will update this accordingly.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Yeah that should be `ILaunchSettings`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/DebugProfileDebugTargetGenerator.cs
tmeschter: Yes, and I'm still thinking about it. :-)

At that point both `ILaunchSettingsProvider` and `IVersionedLaunchSettingsProvider` would have members with the same name, so the definition and implementation might get messy. But if we make `IVersionedLaunchSettingsProvider` public that would really be the right way to go.

tmeschter: Done. Note this just moved the cast elsewhere to clarify _which_ `SourceBlock` we wanted to link to.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/IVersionedLaunchSettingsProvider.cs
tmeschter: Updated as part of changing the derivation.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Fixed.

jjmew (approved):
:shipit:

drewnoakes (approved):

PR #7786, opened 2021-11-23
Use C# 10

drewnoakes:
- Upgrades the compiler we use, and increases the `LangVersion` property to 10.0
- Use C#10 constant string interpolation
- Use implicit usings, and remove redundant using directives

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7786)


tmeschter:
I'm in favor of all of these changes, but I'd like to see them in separate PRs.

melytc (approved):

drewnoakes:
The VS2022 build image does not include an SDK, hence the build failures. It is being looked at.

PR #7787, opened 2021-11-23
Speed up checkout for unit test CI builds

drewnoakes:
By only fetching a single branch to a depth of one, we can speed up the git checkout operation during unit test runs under continuous integration.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7787)

### DIFF
--- a/build/ci/unit-tests-template.yml
+++ b/build/ci/unit-tests-template.yml
@@ -4,6 +4,16 @@ jobs:
   pool: ${{ parameters.pool }}
   timeoutInMinutes: 20
   steps:
+    - checkout: none
+
+    - script: |
+        @echo on
+        @git version
+        git init
+        git fetch --no-auto-gc --no-recurse-submodules --no-show-forced-updates --no-tags --depth=1 "$(Build.Repository.Uri)" "$(Build.SourceVersion)"
+        git -c advice.detachedHead=false checkout --no-progress "$(Build.SourceVersion)"
+      displayName: Shallow Checkout
+
     - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /clearnugetcache /no-deploy /no-integration /no-ibc /configuration ${{ parameters.configuration }}
       displayName: Build ProjectSystem.sln
 


MiYanni (changes requested):
I'm a bit confused. Normal checkout only takes 12-18 seconds or so.
image

Checking the CI from this PR, this seems to be around the same amount of time (15s):
image

I think adding this extra complexity (script) to the yaml doesn't really buy us much than just using the normal checkout mechanism. You could also just set the fetch depth to 1 in the option for `checkout` instead of a custom script: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#checkout

drewnoakes:
This is an experiment to see if we can save some time on builds, and to teach me more about our builds.

There's a fair bit of variability in the checkout time, so to evaluate this properly you have to look at several builds. The debug build for this PR took 15 seconds. The release took 12, and the Spanish took 22. I can find checkouts from earlier builds that took longer than this which, in isolation, would suggest this PR does actually help such as this from 20211122.6:

image

I'm going to continue to tweak this a bit.

MiYanni:
I plan on working on the build process as a whole once I'm done the OptProf stuff. If you find anything interesting, let me know. I've been learning way more than I want to about our repo and our build processes than I signed up for. 😝

drewnoakes:
The built in `fetchDepth` feature is pulling down all tags, so will grab more than is needed:

https://dev.azure.com/dnceng/public/_build/results?buildId=1483838&view=logs&j=908bce63-940e-59f8-7204-204b1f5ead0b&t=a109f775-14ed-55ac-c674-ee2ba99df103

Here's what the built-in command does:

```bash
git version
git lfs version
git init "D:\a\_work\1\s"
git remote add origin https://github.com/dotnet/project-system
git config gc.auto 0
git config --get-all http.https://github.com/dotnet/project-system.extraheader
git config --get-all http.proxy
git config http.version HTTP/1.1
git -c http.extraheader="AUTHORIZATION: basic ***" fetch --force --tags --prune --prune-tags --progress --no-recurse-submodules origin --depth=1 +b42eb23f9958ca7f63b43357b3ab5406a72ac048:refs/remotes/origin/b42eb23f9958ca7f63b43357b3ab5406a72ac048
git -c http.extraheader="AUTHORIZATION: basic ***" fetch --force --tags --prune --prune-tags --progress --no-recurse-submodules origin --depth=1 +b42eb23f9958ca7f63b43357b3ab5406a72ac048
git checkout --progress --force refs/remotes/origin/b42eb23f9958ca7f63b43357b3ab5406a72ac048
```

Here's what we could have instead:

```bash
git version
git init
git fetch --no-auto-gc --no-recurse-submodules --no-show-forced-updates --no-tags --depth=1 "$(Build.Repository.Uri)" "$(Build.SourceVersion)"
git -c advice.detachedHead=false checkout --no-progress "$(Build.SourceVersion)"
```

- `--no-tags` avoids bringing down tags and their associated data.
- `--no-auto-gc` avoids running the maintenance than usually occurs after a fetch, as we know this is a freshly fetch repo so will be packed with no garbage to clean up (replacing the `git config` call of the built-in checkout task).
- `--no-recurse-submodules` avoids checking anything related to submodules, which we don't use.
- `--no-show-forced-updates` avoids checking for remote branches which had forced updates when the fetch completes.
- `-c advice.detachedHead=false` avoids logging that we're in a detached head, which is expected and therefore not noteworthy.
- `--no-progress` avoids filling the log with tens of messages during checkout.
- We also don't have `--prune` or `--prune-tags` which will avoid cleaning up remote branches, of which there are none (this is a fresh clone).

MiYanni:
While this is interesting in a technical sense, I don't know how useful it is for our repo or even this particular pipeline. For example, I just checked a previous repo I worked on and it had a 2.5 minute checkout time. This repo here will likely never be large enough to require a custom checkout script. I'd keep this script in your back pocket, but I don't think we need it here. I have some other ideas to speed up the build based on what I've learned so far and other techniques (like parallel jobs) that I've done in the past. 

For example, did this pipeline here, and total if ran synchronously, would be like 96 minutes. But, the pipeline only runs for 26 because of parallel jobs.
image

drewnoakes:
I'm trialling this change on Roslyn too (https://github.com/dotnet/roslyn/pull/57960). It seems to have dropped their checkout time from ~45 to ~30 seconds.

As I run this more on our repo, I do believe there is a reduction in build time. For example the most recent build showed checkout times of 11 and 10 seconds for debug and release, respectively.

In our repo this shallow check out reduces the size of the `.git` folder on disk from 47MB to 11MB. There'll be an equivalent reduction in bytes transferred between GitHub and AzDO.

Even if this doesn't shave much total time off our build, it still cuts down on resource usage. Over many builds, this adds up, both in terms of COGS and carbon footprint of our CI builds.

> I have some other ideas to speed up the build based on what I've learned so far and other techniques (like parallel jobs) that I've done in the past.

I love speed ups, but don't we already use parallel jobs?

image

MiYanni:
For unit-tests, it is using 3 jobs with the same template. The only difference between the 3 jobs is the configuration (Debug/Release) and the image used. They do run in parallel, but the main chunk of work is the same between all 3, the usage of `build.cmd`. Based on how the build in the repo works, I likely can separate out some of the actions. For example, I could build both debug and release in parallel, but have testing also be in parallel on the 2 different images (Spanish and non-Spanish). Just a top-of-the-head idea. Might not be faster, but worth a try. Additionally, I notice this template uses `/sign` for some reason... it really shouldn't. It seems to set up the actions to do signing, but then doesn't actually run it since these aren't MicroBuild machines. There are likely a lot of little things I can remove or simplify based on what I've seen so far.

build/ci/unit-tests-template.yml
```diff
@@ -4,6 +4,16 @@ jobs:
...
   steps:
+    - checkout: none
+
+    - script: |
+        @echo on
+        @git version
+        git init
+        git fetch --no-auto-gc --no-recurse-submodules --no-show-forced-updates --no-tags --depth=1 "$(Build.Repository.Uri)" "$(Build.SourceVersion)"
+        git -c advice.detachedHead=false checkout --no-progress "$(Build.SourceVersion)"
+      displayName: Shallow Checkout
```
MiYanni, line 15: If we're gonna use a script, might make sense to put it as an actual script file in this folder so it can be used by other yml pipelines too. You might need to pass in those 2 values as parameters. Then, you can also have some comments at the top (you can just copy/paste your comment in this PR about what the script does).

MiYanni (approved):

tmeschter:
This adds a fair bit of complexity to what is normally an operation we don't need to think about much; I'm going to want to see some real statistics about the time reduction. E.g., I like to see something like "over _x_ runs **without** this change, cloning took _y_ seconds with a standard deviation of _z_; over _i_ runs **with** the change, cloning took _j_ seconds with a standard deviation of _k_".

Complexity is a killer when it comes to infrastructure.

tmeschter:
> In our repo this shallow check out reduces the size of the .git folder on disk from 47MB to 11MB. There'll be an equivalent reduction in bytes transferred between GitHub and AzDO.

I'm assuming git is compressing the data for transfer so the size of the data on the network would be much smaller, no?

tmeschter:
> Additionally, I notice this template uses /sign for some reason... it really shouldn't.

If it isn't harming anything, let's just keep doing it. Our actual product binaries are signed, so we may as well exercise those code paths and possibly catch an issue early. Again, let's not have multiple ways of doing things without a good reason.

drewnoakes:
> > In our repo this shallow check out reduces the size of the .git folder on disk from 47MB to 11MB. There'll be an equivalent reduction in bytes transferred between GitHub and AzDO.
> 
> I'm assuming git is compressing the data for transfer so the size of the data on the network would be much smaller, no?

I wouldn't expect so. The .git folder is already compressed. My understanding is that during clone, the bytes a git client receives are written to disk unchanged (as a pack file).

So in this case the PR reduces IO by 36MB on the network, and in the disk (which on a VM may also be on the network).

drewnoakes:
> This adds a fair bit of complexity to what is normally an operation we don't need to think about much; I'm going to want to see some real statistics about the time reduction. E.g., I like to see something like "over _x_ runs **without** this change, cloning took _y_ seconds with a standard deviation of _z_; over _i_ runs **with** the change, cloning took _j_ seconds with a standard deviation of _k_".

The challenge here is getting stats with this change. We would need quite a few samples. One approach is to merge it, leave it run for a while, then review in future.

In terms of prior art, Roslyn implemented shallow clone a while back and saw benefits.

> Complexity is a killer when it comes to infrastructure.

Sure. I can imagine this breaking if we ever used submodules for example, though I find that very unlikely and the fix would be obvious.

For my taste, having explicit commands listed for the checkout makes it clear what actually happens. The built in task works, but is invisible and suboptimal.

drewnoakes:
I still think this is a good change, but I see little enthusiasm for it so don't want to force it in. Will close it out for now.

PR #7789, opened 2021-11-23
Use switch expressions/statements

drewnoakes:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7789)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringBuilderExtensions.cs
```diff
@@ -27,25 +27,23 @@ public static void AppendFormat(this StringBuilder builder, in StringFormat form
...
+                    break;
+                case 2:
+                    builder.AppendFormat(format.Format, format.Argument1, format.Argument2);
+                    break;
+                case 3:
+                    builder.AppendFormat(format.Format, format.Argument1, format.Argument2, format.Argument3);
+                    break;
+                default:
+                    builder.AppendFormat(format.Format, format.Arguments);
+                    break;
```
MiYanni, line 46: nit: Technically, you could use a string expression here since they all return `StringBuilder`. I'm actually surprised this extension doesn't return `StringBuilder`, like how the rest of the methods of the same name do. I'd have to look at how this is used, but it might make sense to do switch expression, update the return type to return StringBuilder, and see if the callers have any issues.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringBuilderExtensions.cs
drewnoakes: It looks like no one is using these extension methods. I'll delete them.

PR #7797, opened 2021-11-24
Use destructuring

drewnoakes:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7797)


davkean (approved):

tmeschter (approved):

MiYanni (approved):
This is actually deconstruction, not destructuring, which would be calling a destructor, not a deconstructor. *WoRdS aRe FuN* 🙃 https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct

melytc (approved):

drewnoakes:
> This is actually deconstruction, not destructuring

C# calls this deconstruction, but most other languages call this destructuring, or sometimes parallel assignment or unpacking. I've always referred to this as destructuring. Old habits.

> which would be calling a destructor

That would be destructing, which is something different again.

PR #7799, opened 2021-11-24
Add AppliesToProject attribute to DependenciesTreeSearchProvider

huguesv:
For 17.1 Preview 2, I've implemented the ability to delay instantiation of solution explorer search providers based on project capabilities or UI context.

Similar functionality already existed for attached collection source providers (which you already take advantage of), and it is now available for search providers.

So with this change here, if you load a solution that has no project with DependenciesTree capability, when you do a search in solution explorer we will not instantiate this provider, which means we avoid loading the managed project system assemblies.

Comparison of loaded assemblies after doing a solution explorer search of a solution with a python project:

Before this change:

image

After this change:

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7799)


davkean (approved):

drewnoakes (approved):
Thank you for adding that capability. This is a nice improvement.

Issue #6612, opened 2020-09-16
ProjectSystem VSIX contains unnecessary files (source and localised rule files)


davkean:
Our VSIX is shipping a bunch of source files:

![image](https://user-images.githubusercontent.com/1103906/93291552-b5c12080-f826-11ea-82c4-733ae68f2bbb.png)

Probably a regression from the work to embed the rules.

davkean:
This is worse, this also includes all the localized rule files as well. All in all, this is 3 MB of data that shouldn't be shipped.

davkean:
The GetCopyToOutputDirectoryItems target is producing them in the following (ie they are being treated as CopyToOutputDirectory items) but I'm not sure why:

```
    <ProjectReference Include="..\..\src\Microsoft.VisualStudio.ProjectSystem.Managed\Microsoft.VisualStudio.ProjectSystem.Managed.csproj">
      <Name>Microsoft.VisualStudio.ProjectSystem.Managed</Name>
      <IncludeOutputGroupsInVSIX>BuiltProjectOutputGroup;GetCopyToOutputDirectoryItems;SatelliteDllsProjectOutputGroup</IncludeOutputGroupsInVSIX>
      <IncludeOutputGroupsInVSIXLocalOnly>DebugSymbolsProjectOutputGroup</IncludeOutputGroupsInVSIXLocalOnly>
    </ProjectReference>
```

drewnoakes:
We should add some automation to prevent files accidentally being added or removed from the package, which runs in CI.

ocallesp:
@davkean 
The `CopyToOutputDirectory `for `XamlPropertyRule `should be **false** ?
https://github.com/dotnet/project-system/blob/master/build/import/VisualStudio.XamlRules.targets#L8

ocallesp:
I created a new work item for TSA so they can automate this
https://dev.azure.com/devdiv/DevDiv/_queries/edit/1220080

davkean:
TSA?

dmonroym:
@ocallesp have you been able to take a further look at this? With the work related to #6652 we'll be creating new packages so would like to avoid this happening

davkean:
This has increased to 454 additional files above our dlls and required files, I think this should be prioritized, just to reduce extract and copy time of these files.

drewnoakes:
Running a `git bisect` shows this regressed in commit e30c529bc40da262c9ff5c5ccaefce2684b713aa:

[...]

PR #7805, opened 2021-11-28
Add test to track changes to VSIX package contents

drewnoakes:
At some point in recent history, our ProjectSystem VSIX picked up a bunch of unexpected files (see #6612). It took us a while to notice this, at which point it was unclear which change caused this to regress.

This commit adds the current list of files as a text file that is under version control. There is one file per VSIX (we have three). Any addition or removal from a VSIX will cause a unit test to fail, which will draw our attention to the change immediately. If the change is intentional, we need only update a text file. If not, we can investigate immediately and block merging a regression.

This PR introduces a dependency on the `Verify.Xunit` library, which is the xUnit flavour of the Verify project:

https://github.com/VerifyTests/Verify

> Verify is a snapshot tool that simplifies the assertion of complex data models and documents.
>
> Verify is called on the test result during the assertion phase. It serializes that result and stores it in a file that matches the test name. On the next test execution, the result is again serialized and compared to the existing file. The test will fail if the two snapshots do not match: either the change is unexpected, or the reference snapshot needs to be updated to the new result.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7805)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Setup/PackageContentTests.ProjectSystem.verified.txt
```diff
@@ -0,0 +1,536 @@
...
+  ko/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  pl/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  pt-BR/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  ru/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  tr/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  zh-Hans/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  zh-Hant/Microsoft.VisualStudio.ProjectSystem.Managed.VS.resources.dll,
+  Microsoft.VisualStudio.ProjectSystem.Managed.VS.pkgdef,
+  Microsoft.VisualStudio.ProjectSystem.Managed.dll,
+  ProjectSystem/Rules/Dependencies/AnalyzerReference.cs.xaml,
```
drewnoakes, line 22: The `ProjectSystem` folder should not be present (#6612).

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Setup/PackageContentTests.ProjectSystem.verified.txt
```diff
@@ -0,0 +1,536 @@
...
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.it.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.ja.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.ko.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.pl.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.pt-BR.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.ru.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.tr.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.zh-Hans.xaml,
+  ProjectSystem/Rules/Items/ProjectItemsSchema.FSharp.zh-Hant.xaml,
+  AnalyzerReference.cs,
```
drewnoakes, line 490: None of these `.cs` files should be present (#6612).

drewnoakes:
The package `Verify.Xunit` is not available in any of our feeds. Investigating.

drewnoakes:
> ProjectSystem/Rules/Dependencies/AnalyzerReference.cs.xaml

A lower version is available. I've updated the PR to use that instead.

tmeschter (approved):
Could you add a description of what the Verify.Xunit package does to the PR description? I'm not familiar with it.

PR #7806, opened 2021-11-29
Remove localised rule file data from our VSIX

### LINKED ISSUE #6612: ProjectSystem VSIX contains unnecessary files (source and localised rule files)

drewnoakes:
Fixes #6612

Commit e30c529bc40da262c9ff5c5ccaefce2684b713aa added support for embedding rule files. In doing so, it set both `CopyToOutputDirectory` to `PreserveNewest` and `XlfOutputItem` to `EmbeddedResource` for these files.

This combination of metadata values caused 468 localised `.xaml` files to be included in our VSIX.

Removing them drops the VSIX size from 6972kB to 2626kB. A [sibling fix in CPS](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/366694) removes a further 30 `.cs` files, dropping the size to 2356kb.

This PR includes and builds upon #7805. In fcc7af49570aeca6528f89d9720b12bf10483aca you can see the removal of files from the snapshot test data. That test harness ensures we won't have surprises like this again.

While looking into this I took the chance to understand and tidy upon how we configure our rules in the `.csproj` and `.targets` files. This should help us to do the right thing with these in future.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7806)


drewnoakes:
I'll investigate the build failure tomorrow.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -28,154 +28,64 @@
...
+
+  <!--
+    NOTE ON XAML RULES
+
+    We use one of two item types for XAML rule files:
+
+    - "XamlPropertyRule" for rules that require a generated class. Useful when working with Dataflow project rule sources.
+    - "XamlPropertyRuleNoCodeBehind" for rules that do not require a generated class.
+
+    In addition to the above, there are two metadata that control how rules are treated during build:
```
tmeschter, line 46:
```suggestion
    In addition to the above, there are three metadata that control how rules are treated during build:
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Microsoft.VisualStudio.ProjectSystem.Managed.csproj
```diff
@@ -28,154 +28,64 @@
...
+
+    - "XlfInput" controls whether display strings in the rule file should be localised. Defaults to true.
+    - "IsEmbedded" indicates that the rule is embedded (see RuleExporter.cs).
+    - "GenerateProjectPropertiesClass" indicates that code generation should contribute to the partial "ProjectProperties" class.
```
tmeschter, line 50: It would be useful to list the defaults for "IsEmbedded" and "GenerateProjectPropertiesClass".

tmeschter (approved):
Why did the file names change (e.g. AppDesigner.xaml.cs -> AppDesigner.cs)?

Also, it would be good to verify that localized rules still work properly after this change.

drewnoakes:
> Why did the file names change (e.g. AppDesigner.xaml.cs -> AppDesigner.cs)?

So that they nest properly (`DependentUpon`) without requiring explicit updates in the project file, keeping the project file leaner.

> Also, it would be good to verify that localized rules still work properly after this change.

Will do before merging. The current build failure is an issue I need to address first.

MiYanni:
@drewnoakes Any update on this? I'd like to put the parent issue for this into the infrastructure project stuff for next sprint.

drewnoakes:
@MiYanni no update here. This needs more work.

drewnoakes:
@MiYanni fixed this elsewhere, so closing this. Thanks Yanni :)

PR #7811, opened 2021-12-06
Populate LaunchProfileValue.DisplayName

tmeschter:
Move to a newer version of CPS that includes the new `LaunchProfileValue.DisplayName` property and populate it. For the moment it will be the same as the `Name` property, but a future change will make `Name` a unique, persistent identifier for a launch profile.

Note this is the second step of a set of staged changes. CPS now defines the `LaunchProfileValue.DisplayName` property but can't take a dependency on it because the .NET Project System doesn't populate it at all. And since CPS can't depend on it, the project system can't be updated to make `Name` a persistent, unique value and handle renames as expected. This interim change allows CPS to move forward without a user-visible break or change in behavior.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7811)

### DIFF
--- a/build/import/Packages.targets
+++ b/build/import/Packages.targets
@@ -11,7 +11,7 @@
       https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json;
-      https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json
+      https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json;
     </RestoreSources>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


tmeschter:
This will remain a draft until we can pick up the real versions of the CPS packages with `LaunchProfileValue.DisplayName`.

build/import/Packages.targets
```diff
@@ -11,7 +11,7 @@
       https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json;
-      https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json;
+      https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json
```
MiYanni, line 14: Any reason to remove the semi-colon? Just makes it easier to add more sources in the future if it remains.

MiYanni (approved):

build/import/Packages.targets
tmeschter: No, that's just an artifact of changing the sources a couple times.

drewnoakes (approved):
Please squash this so that we don't have non-compiling commits in the history. I often use git bisect to identify which commit introduced a regression. Having non-compiling commits makes that process painful.

melytc (approved):

drewnoakes:
@tmeschter is this PR good to go in or are we waiting on something?

tmeschter:
Work on this stalled as I needed to focus on the single-target build for MAUI, but there is otherwise no reason to hold it.

PR #7812, opened 2021-12-07
Don't include .cs files in VSIX package

drewnoakes:
Contributes to #6612

Our VSIX file contained a bunch of source `.cs` files. This problem was addressed in the CPS PR https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/366694.

This PR uses a the updated version of CPS so that these files aren't included during build.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7812)


build/import/Packages.targets
```diff
@@ -12,6 +12,7 @@
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json;
+      https://pkgs.dev.azure.com/devdiv/_packaging/VSIDE-RealSigned-Release/nuget/v3/index.json
```
drewnoakes, line 15: CPS has moved its packages to a new feed, so that new feed has also been added.

@tmeschter this change may help with #7811.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Setup/PackageContentTests.ProjectSystem.verified.txt
```diff
@@ -487,36 +487,6 @@
...
-  DotNetCliToolReference.cs,
-  GeneralBrowseObject.cs,
-  ProjectDebugger.cs,
-  SupportedTargetFramework.cs,
-  SupportedNETCoreAppTargetFramework.cs,
-  SupportedNETStandardTargetFramework.cs,
-  SupportedNETFrameworkTargetFramework.cs,
-  SdkSupportedTargetPlatformIdentifier.cs,
-  SdkSupportedTargetPlatformVersion.cs,
-  VisualBasic.NamespaceImport.cs,
```
drewnoakes, line 519: The `PackageContentTests` unit test failed (as expected) after updating the CPS version, because these `.cs` files were not present in the VSIX. Removing them from this text file set a new baseline, and causes the test to pass.

MiYanni (approved):

build/import/Packages.targets
tmeschter: Is this feed public? We can only depend on publicly accessible feeds.

drewnoakes: @adrianvmsft suggested that the fact CPS packages stopped being published to the previous feed was a regression, so it's it doesn't seem like we should merge this change.

drewnoakes:
Converting to draft until we can get CPS packages from a suitable feed.

build/import/Packages.targets
drewnoakes: It turned out that a post-build stage on the pipeline responsible for uploading the packages blocked back in October and was stuck for 67 days. Cancelling that particular step allowed subsequent builds to complete. It took a few hours, but we now see newer CPS packages on the [feed](https://dev.azure.com/azure-public/vside/_packaging?_a=package&feed=vs-impl%40Local&view=versions&package=Microsoft.VisualStudio.ProjectSystem&version=17.0.596-pre&protocolType=NuGet).

I've updated this PR accordingly.

Issue #7816, opened 2021-10-19
Renaming file(with type) freezes UI


DavidKarlas:
**Version Used**: 
4.0.0-5.21479.6+91a7a90557a9bd8e8e120d1899f077077c829ce6

**Steps to Reproduce**:

1. Rename C# file that contains type with same name in huge .NET Framework solution
2. When dialog appears confirm that you want to rename type
3. IDE freezes for multiple seconds, without dialog to cancel it

**Expected Behavior**:
Wait dialog should appear which offers cancellation of rename.

**Actual Behavior**:
Visual Studio just freezes until rename is complete.

**Solution**:
Put `IVsThreadedWaitDialog` around [AbstractCodeModelService.cs#L497](https://github.com/dotnet/roslyn/blob/8751f99/src/VisualStudio/Core/Impl/CodeModel/AbstractCodeModelService.cs#L497) and pass `CancellationToken` to `Renamer.RenameSymbolAsync`.

PR #7817, opened 2021-12-08
Remove IWaitContext.AllowCancel

drewnoakes:
#7816 made me curious about how we handle threaded wait dialogs in the .NET Project System.

While looking at our code for that, I noticed buggy and unused code. This PR removes it, and tidies.

The `IWaitContext.AllowCancel` property is unused and will not work correctly. If a dialog was created that did not allow cancellation, then this property is later set to true, the backing cancellation token will not be updated. Further, it's unlikely that a consumer of this type would re-query the token after setting the property. It feels like a safer design, and a cleaner implementation, to remove this unused property and avoid potential future bugs.

Note that we don't use the `IWaitContext.Message` property either, but the implementation of that property appears to be correct so I left it in.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7817)


MiYanni (approved):
Looks good. Making it initialize-only make sense.

PR #7827, opened 2021-12-17
Localise fast up-to-date check logging

### LINKED ISSUE #7820: Incremental build logging is not localised
By default the fast up-to-date check does not log messages. Users can enable logging via _Tools | Options | Projects | SDK-Style_.

These log messages are currently not localised and will always appear in English.

In 17.1 we are trialling a feature that flags incremental build failures. Diagnosing these failures requires enabling this logging. That increases the urgency that this logging should be localised.

drewnoakes:
Fixes #7820

Moves all strings to resource dictionaries. A few strings were slightly modified to be more likely to succeed in other languages.

The lack of indentation for items within sets in verbose logging was also addressed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7827)


tmeschter (approved):

MiYanni (commented):
Nice! Interesting indent logic. 👍

drewnoakes:
> Nice! Interesting indent logic. 👍

I was concerned the leading space would be lost in translation (literally). It also addressed the issue with set messages not being nested.

PR #7836, opened 2022-01-06
Small changes

drewnoakes:
A set of small changes, pulled out from larger work to follow in a subsequent PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7836)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
@@ -86,5 +86,17 @@ namespace Microsoft.VisualStudio
 
             return builder.MoveToImmutable();
         }
+
+        public static TOut[] SelectArray<TIn, TOut>(this TIn[] array, Func<TIn, TOut> selector)
+        {
+            TOut[] output = new TOut[array.Length];
+
+            for (int i = 0; i < array.Length; i++)
+            {
+                output[i] = selector(array[i]);
+            }
+
+            return output;
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


tmeschter (approved):
Looks good with the usual caveat that we should hold this until we have a 17.2 branch.

src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
```diff
@@ -86,5 +86,17 @@ public static bool Any<T>(this T[] array, Func<T, bool> predicate)
 
             return builder.MoveToImmutable();
         }
+
+        public static TOut[] SelectArray<TIn, TOut>(this TIn[] array, Func<TIn, TOut> selector)
```
MiYanni, line 90: Is the performance difference with LINQ's `Select` because this version doesn't need to allocate a type from Linq's query engine behind-the-scenes?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/LinqExtensions.cs
drewnoakes: There savings are:
- No enumerator allocated for `array`
- The `output` array is pre-allocated at the correct size, whereas in the Linq case the `Select` produces an enumerable with no length information such that the chained `ToArray` call will have to buffer items, potentially allocating multiple arrays (doubling the length each time) until the input is exhausted, and then likely allocates another array of the correct length as part of trimming the buffer to size.

PR #7837, opened 2022-01-06
Enable external "related items" to use the interaction pattern

drewnoakes:
I am working to show `.props` and `.targets` files within NuGet package nodes in the Dependencies tree (#7838). It feels natural to want to double-click these to open them in the editor.

Currently that is not possible on the NuGet side as the explicit implementation of `IInteractionPatternProvider.GetPattern` could not be overridden. This PR changes that, however it will take a while for code to flow through to where I can consume it on the NuGet side. To avoid that delay, this PR also adds `IInvocationPattern` to the set of supported patterns. At runtime there is no behaviour change for types that do not actually implement IInvocationPattern. However those that do will now work correctly, meaning the NuGet scenario becomes unblocked.

The "interaction pattern" here means support for double-clicking nodes in the Dependencies tree.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7837)


tmeschter (approved):
Looks good to me, but we should hold this until our `main` branch is targeting 17.2.

MiYanni (approved):

PR #7843, opened 2022-01-10
Migrate main to 17.2

tmeschter:
Migrate the main branch to 17.2. This is largely a matter of updating Versions.props to use the proper major.minor version.

A dev17.1.x branch will be created in further work.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7843)

### DIFF
--- a/build/ci/integration-tests.yml
+++ b/build/ci/integration-tests.yml
@@ -10,7 +10,7 @@ trigger: none
 jobs:
 - job: Visual_Studio
   pool:
-    name: NetCore1ESPool-Public
+    name: NetCore1ESPool-Svc-Public
     demands: ImageOverride -equals Build.windows.10.amd64.vs2019.pre.open
   strategy:
     maxParallel: 2
--- a/build/ci/integration-tests.yml
+++ b/build/ci/integration-tests.yml
@@ -10,7 +10,7 @@ trigger: none
 jobs:
 - job: Visual_Studio
   pool:
-    name: NetCore1ESPool-Svc-Public
+    name: NetCore1ESPool-Public
     demands: ImageOverride -equals Build.windows.10.amd64.vs2019.pre.open
   strategy:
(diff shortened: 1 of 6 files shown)


MiYanni (changes requested):
With the new policies for CI pools, you'll need to do another PR prior to this one.
1. First PR should change all `NetCore1ESPool-Public` to `NetCore1ESPool-Svc-Public`
2. After that is merged, create the new `dev17.1.x` branch
3. Update this PR to change `NetCore1ESPool-Svc-Public` back to `NetCore1ESPool-Public`

This order is required or else the history on `main` won't contain the right information. Otherwise, trying to change to the `Svc` pools on the child branches will cause our auto-merge tool to try to apply those changes to `main`, which we don't want. `Svc` is only used on child branches.

Here was the original change to the pool names so you know what I'm talking about. It is just the CI folder's yaml files. https://github.com/dotnet/project-system/commit/fff9e58de2cdcf5a7209aa7f770e7f99fc6077e4

tmeschter:
Oh, that's interesting. Why the different pools for the servicing branches?

MiYanni:
@tmeschter From what I was told, it is for billing reasons. Meaning, different cost centers are charged for the different pools. So, we should only use non-Svc for `main`.

tmeschter:
@MiYanni What I've done is create three commits:
- switching to the servicing pools
- updating the version number
- switching the servicing pools back

Once this goes in, I'll update dev17.1.x to branch off the first. Sound good?

MiYanni (approved):
Just make sure to make dev17.1.x off of https://github.com/dotnet/project-system/pull/7843/commits/22c65721b9404e13f8d7078383a3105f2eead752. Thanks!

Issue #7847, opened 2022-01-17
Avoid redundant updates to the IWorkspaceProjectContext


drewnoakes:
During solution open, and any time a project is updated, we notify Roslyn of relevant project data via the `IWorkspaceProjectContext` interface. Each update occurs within a transaction which has the following phases:

1. Call `IWorkspaceProjectContext.BeginBatch`
2. Call whatever `IWorkspaceProjectContext` members need to be updated
3. Call `IWorkspaceProjectContext.EndBatchAsync`

While investigating #7541, I observed that we frequently perform steps 1 and 3, while skipping step 2. Such empty batches seem wasteful. I asked @jasonmalinowski about this via email and he wrote:

> Fewer batch operations should generally be helpful, yes. The main benefit would be that it produces fewer Solution snapshots that the rest of our system will start to analyze. Otherwise those “intermediate” states we’ll start parsing/producing data for, only for it to get thrown away.

There are currently three dataflow sources that can trigger us to update the `IWorkspaceProjectContext` in a batch:

1. Evaluation
2. Source items
3. Build

We are not handling version-only changes properly, and are pushing empty updates through to Roslyn.

Furthermore, source items derive directly from evaluation, so they should be bundled and treated as a single update.

When loading a new class library project, we currently fire nine updates. This should only be two: one for evaluation, and one for build.

jasonmalinowski:
@drewnoakes Hmm, I might have misunderstood the email conversation then. I was thinking you were doing multiple batches with stuff still in them -- I didn't quite realize this was empty batches. Those should be pretty harmless other than we're flooding the underlying semaphore slim with more work that'll release it. But sounds like your PR should be reducing other work at the same time too?

drewnoakes:
@jasonmalinowski there are two scenarios:

1. We send through empty batches (fixed in #7848)
2. We report evaluation changes twice, once for each of the `ProjectRule` and `SourceItemsRule` data sources

Our email conversation covered both of these cases. For a new console app I saw nine batches. Fixing point 1 gets this down to three batches. Fixing point 2 gets this down to two batches. So I guess it's the second case that's likely to be more impactful.

The fix for point 2 is on my local machine and I'll submit a PR once #7848 is merged in.

Do you have any thoughts on how to measure a performance delta here? If you have tooling for that and are interested, the code is on branch https://github.com/drewnoakes/project-system/tree/combine-evaluation-and-source-items.

jasonmalinowski:
[...]

PR #7848, opened 2022-01-17
Reduce number of workspace context batches in language service

drewnoakes:
Contributes to #7847.

As described in that issue, we can end up notifying Roslyn's `IWorkspaceProjectContext` of changes when there aren't any.

This PR cuts the number of such updates considerably by avoiding no-op batches.

With this change:

- For `OrchardCore.sln` we skip 1218 of 2177 batches (55% reduction)
- For `Roslyn.sln` we skip 892 of 2552 batches (35% reduction)

There are three parts to the change here, each in its own commit:

1. Introduce a snapshot for command line argument data (`CommandLineArgumentsSnapshot`) with its own data source (`CommandLineArgumentsProvider`). The snapshot includes a flag to indicate when it has changed.
2. Avoid redundant workspace context batches when no changes exist.
3. Make internal methods non-async.

The remaining portion of #7847 is to merge evaluation and source item updates into a single update. This will further reduce the number of batches sent to Roslyn.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7848)


tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextHost.WorkspaceProjectContextHostInstance.cs
```diff
@@ -189,12 +191,27 @@ private async Task ApplyProjectChangesUnderLockAsync(ProjectChange change, Works
                 // NOTE we cannot call CheckForInitialized here, as this method may be invoked during initialization
                 Assumes.NotNull(_contextAccessor);
 
-                IWorkspaceProjectContext context = _contextAccessor.Context;
                 bool isActiveEditorContext = _activeWorkspaceProjectContextTracker.IsActiveEditorContext(_contextAccessor.ContextId);
                 bool isActiveConfiguration = change.ActiveConfiguredProject == _project;
 
+                bool hasChange =
+                    change.Subscription.Value.ProjectChanges.Any(c => c.Value.Difference.AnyChanges) ||
+                    change.CommandLineArgumentsSnapshot?.IsChanged == true ||
```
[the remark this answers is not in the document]
drewnoakes: Yes. Of the three possible values (true, false, null) we are testing for only one.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/CommandLineArgumentsProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+            // Set the link up so that we publish changes to target block
+            transformBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+
+            // Join the source block, so if it needs to switch to UI thread to complete 
+            // and someone is blocked on us on the same thread, the call proceeds
+            JoinUpstreamDataSources(_projectBuildSnapshotService);
+
+            return transformBlock;
+
+            CommandLineArgumentsSnapshot Transform(IProjectBuildSnapshot buildSnapshot)
```
MiYanni, line 46: While I understand the usefulness of writing code like this as local functions, the clarity of scoping becomes messy after you start nesting local functions within local functions (like `IsUnchanged`) below. Maybe I'm just an old-head, but having logic like this as *minimally scoped as possible* seems like a good way to keep code intent clear. For example, if this code does not use class instance variable, it should be `static`. If all information can simply be passed into it instead of containing the scope of the outer method, it can just become its own `static method`. In this local function, the only value required is already being passed in as a parameter, `buildSnapshot`. The `lastSnapshot` variable is accessed from the method scope, but (as far as I can tell) the method itself doesn't use `lastSnapshot` at all. The only benefit is that `lastSnapshot` will exist for the duration of the method scope, but I can't tell if that is intended by the code available here. Similarly, `IsUnchanged` requires an `ImmutableArray<string>` (for snapshot arguments) and a target outputs (from the buildSnapshot) which could be passed into it, thus not needing any scopes above it, which makes it a `private static method` in my mind. This increases the ability (in my option) to easily understand that the code in that block only needs to be read for the complete picture; I don't need to look elsewhere. It is self-contained. The only reason I'm not a fan of this local functions pattern is I had past experience writing in ColdFusion, where the entire language has this scoping issue, but much, much worse. There were some bits of code I was refactoring that was 10 scopes deep, and you'd have to look through a few thousand line file to find where a certain variable was created and when it was actually set with useful information. Here, this is obviously not as complex as that, but I kinda see this as an anti-pattern in any language to unnecessarily scope code that can be totally independent. Here, the fact it is local is primarily used to associate that block with this method directly, which I think is a good thing, but not at the price of pulling in the method's scope. Additionally, this style causes more and more intending which I'm also not a fan of. Generally, the deeper the indentation, the more smelly the code is. That isn't always the case, but generally smaller, more encapsulated code more left justified. Lastly, the testability of these local functions is basically non-existent as you can't access them outside of the method. At least with static methods in the class, you can `InternalsVisisbleTo` so unit test projects can access them.  Personally, I think I'd only use local functions if that are very small and have multiple callsites in the method. Here, each of these local functions are only called once by their outer scope. That's just my 2-cents.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/CommandLineArgumentsProvider.cs
drewnoakes: There are trade-offs and style preferences. I like this kind of nesting as it limits the number of things I have to think about while reading the code. In general I find knowing where a bit of code can be called from to be the most useful thing. In this class, it's clear that all logic happens in response to a call to `LinkExternalInput`.

If each were moved into separate top-level functions it would become harder to work out how they relate to each other. This can become especially problematic in larger classes. Nesting functions keeps scopes tight and clear.

PR #7849, opened 2022-01-18
Support OptProf for project system assemblies and replace RepoToolset signing

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7191
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1409278
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1370981

**NOTE**: For this PR, it should use `Squash and merge` so that we don't add the incredible amount of commits to the repo.
**NOTE**: For this reviewing this PR, it is recommended to check `Hide whitespace` in the diff settings.
image
**NOTE**: I'd recommend using side-by-side view when reviewing the file: `official.yml`

# Primary Changes
- Enable use of OptProf for profile-guided optimization of our project system assemblies (`Managed.dll` and `Managed.VS.dll`)
- Remove use of RepoToolset custom signing in favor of using the MicroBuild signing plugin
  - The strict MicroBuild package versions required for the RepoToolset process conflicted with updated packages needed for using the current OptProf plugin
- Remove (currently not utilized) manual optimization merging for non-project system assemblies (`Editors.dll` and `AppDesigner.dll`)

### Pipeline URLs
- Our signed build and packaging pipeline: https://devdiv.visualstudio.com/DevDiv/_build?definitionId=9675
  - This is the pipeline that is driven by `official.yml`
- Our new OptProf profiling release pipeline: https://devdiv.visualstudio.com/DevDiv/_release?definitionId=3197
  - Release pipelines are not driven by a `.yml` in the repository. All information is saved into the pipeline itself.

# OptProf Implementation Details
## Overview
**OptProf** is the name of a workflow created by *MicroBuild* to do profile-guided optimization (PGO) for assemblies. At a high level, PGO rearranges the lookup tables in an assembly for calling methods/functions. This style of sorting needs metrics on how the assembly is used to know what to put at the top and conversely, what isn't used often. To do this, we need to feed OptProf a test to execute over our assemblies. OptProf will run this test and generate `.pgo` files to store the optimization data. The OptProf pipeline holds these `.pgo` files for us, so we never actually see them ourselves. The optimization data is merged into the assemblies we produce and that provides us with optimized `.dll` files.

### The Process
Unfortunately, to build an optimized `.dll`, we need the optimization information. To get the optimization information, we need to run the `.dll` against a test that exercises the functionality of the assembly. Do you see the issue? It is a chicken-and-the-egg scenario where we need to build the `.dll` first, and then get optimization data, and then build it again using the optimization data. Because of the pipeline design, this will cause our current `.dll` built by the pipeline to always use the previous build's optimization data. It isn't ideal, but that's how the design of OptProf is intended.

If you were do to the process mentioned above in the same pipeline, it would be *hours* long. The current pipeline to produce the optimization data takes about an *hour and a half* to run. A majority of this time is provisioning the VM to run the tests. There is *likely* a better way to do this, but this is the series of *task groups* they provide to do the optimization release pipeline. Again, I feel there is a better way to do this, but it isn't something I've looked into at this time.

### The Changes
To make this happen, we needed to change our current *official* build pipeline (the one that produces a signed build) and add a process as a release pipeline. The release pipeline could be integrated into our current release pipeline, but I didn't do that at this time. Instead, it is a separate pipeline that is kicked off when we produce a signed build. Release pipelines are not `.yaml` based like normal build pipelines, so I don't know if I can just copy/paste what I did from one into the other. Also, it has *a TON* of pipeline variables, so that would need to be moved over too.

For the official build pipeline, there are 2 core aspects: producing the metadata needed to perform optimization (in the release pipeline) and actually consuming the optimization data in the build of our assemblies. For the first run of these changes, I needed to disable the consumption of optimization data because I didn't have it yet. The plugin used is still required to be installed and configured, but you set the `ShouldSkipOptimize` property on the plugin to `true`. Once the pipeline properly produces and publishes the metadata, and the release pipeline uses the metadata and produces the optimization data, this flag can be set to `false` so the next build can utilize the optimization information.

### Why was this difficult?
While there is a decent amount of documentation, most of it is bare-bones in terms of explanation. It will generally say what is needed but not *why it is needed*. Our repo is not the target audience the documentation was geared toward. It is written as DevDiv documentation and (somewhat) assuming you are working in the VS repo (or at minimum, as a DevOps repo). We are neither in VS or using a DevOps repo. Additionally, there was a V1 version of this process and the (current?) V2 version. I based most of the work on the CPS implementation, which they originally had a V1 implementation and had to transition to a V2 implementation. From what I understand, V1 was more of a centralized process while V2 is a decentralized one.

The main piece of information I got from CPS's documentation of this process is this image:

OptProf Diagram

While this contains a good amount of information, it is difficult to reconcile on its own. Another piece of the process that I didn't implement is shown in this diagram, related to publishing the tests. Since I am reusing an already published test, I did not have to implement a process for that. The other primary source for documentation comes from the DevDiv wiki: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/700/OptProf. Even with the diagram in hand, because of the cyclical nature of it, it was difficult to find a place to start implementing this. I used the CPS `.yaml` as a guide and talked with Adrian to get some great introductory information.

One of the biggest challenges for our repo was trying to use the recommended tools/plugins with our current tools/plugins. We heavily rely on various tools in the [roslyn-tools](https://github.com/dotnet/roslyn-tools) repo. We use multiple tools from this repo and (from my understanding) many of these tools were the precursor for [arcade](https://github.com/dotnet/arcade). Even though these tools are still being used in some repos, they are not under active development/enhancement. Many of the tools have hard-coded dependencies on certain packages/plugin versions. These tools utilize the MicroBuild plugins, but don't directly run them.

OptProf is a MicroBuild plugin. MicroBuild uses a Core plugin package to include any other MicroBuild plugins installed into the machine. The signing process we were using is part of the `roslyn-tools`. The signing tool has a hard-coded dependency on [0.2.0](https://github.com/dotnet/roslyn-tools/blob/6a3845c885ae8747465e65cbadc267f00d98d1df/src/SignTool/SignTool/SignTool.SignToolBase.cs#L18) of MicroBuild.Core. As you can see [here](https://github.com/dotnet/roslyn-tools/blob/6a3845c885ae8747465e65cbadc267f00d98d1df/src/SignTool/SignTool/SignTool.SignToolBase.cs#L97) and [here](https://github.com/dotnet/roslyn-tools/blob/6a3845c885ae8747465e65cbadc267f00d98d1df/src/SignTool/SignTool/SignTool.SignToolBase.cs#L119), the path includes the folder `MicroBuild.Core`. Unfortunately, newer versions of Core were renamed (as were most of the MicroBuild plugins). This changed the folder path (and package name) to `Microsoft.VisualStudioEng.MicroBuild.Core`. So the `roslyn-tools` signing tool wouldn't work with modern Core without being changed to support both the old and the new pathing.

Based on this fact, I decided to replace the usage of the `roslyn-tools` signing tool with normal MicroBuild signing for our pipeline. There are only a handful of files we sign and I've implemented MicroBuild signing into other repos in the past, so I *thought* it would be fairly straightforward. Unfortunately, the signing tool did some custom logic to repack things and would also generate projects to include the items that need to be signed. In MicroBuild signing, you simply tell it which files need signing as part of your project's build information. These files are required to be in the project's output folder or sub-folders of the output folder. Unfortunately, we have files in a few different places that need signed.

To work around this, there are some file copy targets I created. The files we sign are our `.dll`, `.nupkg`, and `.vsix` files. The `.vsix` has another custom tool from the `roslyn-tools` repo that is used to create it. At this time, I was able to preserve the usage of that tool, but would advise we use the standard way to create `.vsix` files (as a future work item). I believe this uses the Swix plugin from MicroBuild. The other files (`.dll` and `.nupkg`) have a standard process for being signed by MicroBuild which made them fairly easy to get signed properly. Overall, with some custom targets, I was able to get all the necessary files signed, some of which required some tweaks to our setup projects.

Overall, there was a bunch of bits and pieces that needed modified to make OptProf fit into our current build processes and CI pipelines. There was a lot of learning/research involved and understanding of our current processes that we use. The OptProf side has a lot of plugins/tasks/variables involved, and the documentation for these is not descriptive enough to indicate exactly what was needed for our repo's unique situation (outside of VS repo and using GitHub). Based on all these learnings, I will be making a series of other issues to address problems and concerns I've found.

### OptProf.runsettings
The `.runsettings` file is used to provide metadata about the tests to use for generating optimization data. The current test we are using against our `Managed.dll` and `Managed.VS.dll` is **OpenAndCloseProjectTestSolution**. This test was originally used by CPS. They have since created a new version of the test to include their caching feature, which we are not interested in.

Also in CPS, they have a `.runsettingsproj` used to generate the runsettings file. This involved a process of generating CI variables that fed into the project. I was able to build the runsettingproj locally by hard-coding some of the variables required. The only difference I noticed between the runsettings file created locally and the one created in the CPS CI was the inclusion of 2 `TestStore` elements.

To simplify this process, I opted to have a static `.runsettings` file in our repo with an empty `TestStores` element. Everything else in this file is a direct copy/paste from a generated one in CPS, except targeting a different test. Then, I determined how those 2 `TestStore` elements were populated in the CPS CI and wrote a PowerShell script to do this. The script I created is called `UpdateRunSettings.ps1`.

### UpdateRunSettings.ps1
This script will take the `OptProf.runsettings` and update it to add 2 `TestStore` elements in the `TestStores` element of the document. This is an XML document, which means it can be easily loaded and modified using PowerShell. One of the `TestStore` elements is the location of the ProfilingInputs which are created by the build. We pass in a parameter to the script in CI set to `ProfilingInputs/$(System.TeamProject)/$(Build.Repository.Name)/$(Build.SourceBranchName)/$(Build.BuildId)`. The other `TestStore` element is the location of the build drop provided by the `BootStrapper.json` file that is generated during the CI pipeline. Since this file is `.json`, it is also easy to work with in PowerShell. The script pulls the necessary data out of the `BootStrapper.json` and makes it into the correct format for the `TestStore` element.

Each `TestStore` element has the value as a `Uri` attribute. For this value, it must be prefixed with `vstsdrop:` followed by the relative path to the locations. Here is an example of the `TestStore` elements that would get added to `OptProf.runsettings` by the script:
```xml
<TestStores>
  <TestStore Uri="vstsdrop:./XXXXXX/dotnet/project-system/OptProf3/XXXXXXX" />
  <TestStore Uri="vstsdrop:./XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX" />
</TestStores>
```

### official.yml
This file drives our *signed build pipeline* in Azure DevOps. This needed several changes to enable OpfProf to work properly and the updates necessary to make normal MicroBuild signing to work. I also ended up cleaning up and giving headers to sections of functionality. I plan on creating an issue for making this build pipeline composed of multiple jobs that can run in-parallel. Each of these 'sections' should simply be a separate `.yml` file, and `official.yml` should just be an orchestrator of these jobs. I've done this similar process before for other repos. I didn't end up doing all those changes as they weren't necessary for this PR.

The primary changes include:
- Added comments/documentation to what specific tasks are actually doing or why there are certain values
- Rearranged the order of a couple tasks to better align with groups of functionality
- Fixed a bug that caused the `PublishSymbols` to not function properly (`SearchPattern` was incorrectly formatted)
- Added a section of tasks for OptProf metadata publishing

### SignToolData.json
This file was from our previous signing mechanism using the `roslyn-tools` signing tool. It contained the information on which files needed signed. This generated project files that contained the MicroBuild information for signing. Then, the tool manually ran signing on those generated projects. This file is no longer needed as we simply added the `<FilesToSign/>` elements to our existing projects.

### Sign.proj
This file was a custom project to execute the `roslyn-tools` signing tool. This was no longer needed after migrating to MicroBuild signing.

### Files within `build/proj/internal`
These files had to do with manual IBC merge that was previously done on the `AppDesigner`/`Editors` projects. The use of this was disabled due to feed changes/build issues and had not been re-enabled since mid-2021. This was using optimization data from 2017 for these projects. At this point, we would properly implement OptProf for these assemblies to have them optimized. Thus, these files are no longer needed.

### _vsts.ps1
This was part of a custom DevOps variable process that is used in some repos for VS components. Adrian explained how it works in CPS, but for our CI, we don't use it at all. We would run it, but it didn't provide any information or variables for our build. This file was not needed.

Here are the files that are part of it in the CPS repo, for example. We did not have any files associated with this process other than this one, which is the bootstrapper for this process.
Screenshot 2021-11-03 172145

### *.verified.txt files
These files relate to the `PackageContentTests`. They are the hard-coded list of files that should be included within the `.nupkg` files. These lists had to be updated since the signing process was changed. The `roslyn-tools` signing tool would repack the `.nupkg` for each project and had their own custom way of doing that. This caused the file order to be specific to that process. I've changed the test so the file order is an alphabetical comparison. There were also minor tweaks as the signing process included some digital signatures that only appear in the package when done in CI with signing. A local build running this test should not be checking for those files in the packages.

### `<AddVsixForSigningDependsOn/>` property
This property was added because the `CommonFiles.csproj` project has a different VSIX creation process than the other 2 setup projects (`ProjectSystemSetup.csproj` and `VisualStudioEditorsSetup.csproj`). This changes which target that `AddVsixForSigning` uses for `DependsOnTargets`. Meaning, this is when the `<FilesToSign/>` elements are added so that signing can occur *after* the VSIX actually exists.

### `<IsManagedProjectSystemProject/>` property
This property was added to differentiate between two project groups: `Managed`/`Managed.VS` and `AppDesigner`/`Editors`. The former has this property set to `true`; the latter `false`. Currently, this allows OptProf to only function on projects where this property is `true`. There are a few other properties that use `IsManagedProjectSystemProject` to determine their values. There is one value intended for future use when OptProf is enabled on projects where this value is set to `false`. The other time that is currently in use relates to signing certificates.

## Misc Info
A collection of oddities, bugs, weird things, and general 'Huh?' about this entire series of processes.

### OptProf message when the tests are insufficient
This was trying to run OptProf on the `AppDesigner`/`Editors` projects. Our current test we use doesn't activate code from those DLLs. It will show this message in the logs:

image

image

### VS Bootstrapper error when the MicroBuild Signing plugin is not installed
This proves that the VS Bootstrapper plugin requires the MicroBuild Signing plugin to run. I have no idea why that is a requirement, but it is!

image

### OptProg diagram as I was trying to understand the paths
I ended up coloring the different parts to show start/end progression. This is when the bootstrapper process was confusing me.

OptProf-Outlines

### Changing the symbol artifact name will no longer match it with the existing artifacts
This means that you have to publish a new build where the artifact names are changed. Then, our release insertion pipeline will pick up the new names. If you don't do that, those release pipelines will no longer match with the preexisting artifact names. It is a really weird thing I ran into.

What it normally downloads as artifacts:

image

What it downloads after the name change in the `.yml`:

image

What is it missing? `symbols`! It is missing it because I renamed the artifact name from `symbols` to `Symbols` in the artifact publish step. It really seems like it shouldn't use the primary ID of those artifacts to be, essentially, the display name of the artifact. Really bizarre.

### Figure out the variables that were required
I had to document which variables map to which task groups. Right now, because I want to remove all the variables mapped to `Release`, I have to do the deletion of these variables *completely separately* from the deletion of the task. I don't know why deleting the task wouldn't remove the variables associated with it? It doesn't make a lot of sense.

image

image

This gets even complicated when you factor in variable groups, which this process involved several.

### VS Bootstrapper warning
I tried everything to make this warning go away. I tried several different variables you can add to the task. I found different URLs for them and tried each of those. I looked at several pipelines within DevDiv that use this task, and all of them contained that warning message in their run.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7849)


build/import/HostAgnostic.props
```diff
@@ -16,15 +16,17 @@
     <Reference Include="System.Xml.Linq" />
     <Reference Include="WindowsBase" />
 
+    <!-- Infrastructure -->
+    <PackageReference Include="MicroBuild.Core.Sentinel" />
+    <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" />
```
MiYanni, line 21: Created an `Infrastructure` block since these packages are no longer only related to the RepoToolset. As RepoToolset usage is reduced/removed, likely more packages would go into this section as they will likely run natively with MicroBuild plugins. Currently, most packages in the `Toolset` section use `ExcludeAssets="all"` which means they don't run via the plugins as the assets don't actually interact with the projects. Instead, the packages get restored but custom code in RepoToolset accesses them directly. Things in this `Infrastructure` section will properly use the package contents.

build/import/Versions.props
```diff
@@ -18,11 +18,8 @@
     <RoslynToolsRepoToolsetVersion>1.0.0-beta-62705-01</RoslynToolsRepoToolsetVersion>
     <RoslynToolsModifyVsixManifestVersion>1.0.0-beta-62327-02</RoslynToolsModifyVsixManifestVersion>
     <RoslynToolsSignToolVersion>1.1.0-beta3.21260.1</RoslynToolsSignToolVersion>
-    <RoslynDependenciesProjectSystemOptimizationDataVersion>2.6.0-beta1-62205-02</RoslynDependenciesProjectSystemOptimizationDataVersion>
-    <!--TODO: Merged PIAs - evaluate the need of this upgrade https://github.com/dotnet/project-system/issues/7191-->
-    <MicrosoftDotNetIBCMergeVersion>5.0.7-beta.20159.1</MicrosoftDotNetIBCMergeVersion>
```
MiYanni, line 23: These directly related with the optimization that was done for Editors/AppDesigner assemblies. The use of optimization was disabled sometime in 2021 because of changes in package feeds. Daniel had attempted to work around this but it seems there were still issues with the workaround as the `/no-ibc` flag was still set on the build.cmd call for CI. This means that as of sometime in 2021, the Editors/AppDesigner DLLs are no longer optimized. Additional information I learned is in the comments on https://github.com/dotnet/project-system/issues/7191. Basically, the optimization data for these DLLs is over 4 years old and we'd want to use OptProf to do optimization at this point anyway. Thus, I removed the custom use of IBCMerge since it wasn't even being ran anymore anyway. A separate issue will be created for doing OptProf for Editors/AppDesigner as there are a few different tasks that need done for that to be implemented.

build/proj/Build.proj
```diff
@@ -1,74 +1,49 @@
...
 
-  <Target
-    Name="ClearNuGetCache"
-    Condition="'$(ClearNuGetCache)' == 'true'">
-
-    <Message
-      Text="Clearing '$(NuGetPackageRoot)'..."
-      Importance="high" />
+  <Target Name="ClearNuGetCache" Condition="'$(ClearNuGetCache)' == 'true'">
+    <Message Text="Clearing '$(NuGetPackageRoot)'..." Importance="high" />
```
MiYanni, line 14: In these .proj files, there's a lot of horizontal flattening I did as carriage returns for every element made the file tall and skinny, which was more difficult for me to read. So, you'll see this collapsing throughout.

build/proj/CoreBuild.proj
```diff
@@ -1,21 +1,18 @@
...
-                                    
   Optional parameters:              
     CIBuild                         "true" when building on CI server
     Build                           "true" to build solution
     Rebuild                         "true" to rebuild solution
     Deploy                          "true" to deploy assets (e.g. VSIXes)
     Test                            "true" to run tests
     IntegrationTest                 "true" to run integration tests
-    Sign                            "true" to sign built binaries
-    SignType                        "real" to send binaries to signing service, "test" to only validate signing configuration.
```
MiYanni, line 16: The `Sign` target was removed entirely as MicroBuild signing is simply based on the plugin being present or not. In CI, the plugin gets installed which enables signing. It would be possible to put a toggle for signing as a pipeline variable, but it certainly has nothing to do with being a build flag at this point.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
```diff
@@ -1,24 +1,39 @@
...
+
+  <!-- Adds the appropriate vsix files for signing. -->
+  <Target Name="AddVsixForSigning" DependsOnTargets="SwixBuild" BeforeTargets="SignFiles">
+    <Copy SourceFiles="$(VisualStudioSetupInsertionPath)$(AssemblyName).vsix" DestinationFolder="$(OutDir)" />
+    <ItemGroup>
+      <FilesToSign Include="$(OutDir)$(AssemblyName).vsix">
+        <Authenticode>VsixSHA2</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
```
MiYanni, line 37: It will look like that this target is copied 3 times among the setup projects. However, this particular target has a different `DependsOnTargets` than the other two. The way this Vsix is generated is different than the other two as it relies on a process from RepoToolset. The other projects don't seem to follow this process. I didn't dig too deep here but that is what the build logs/binlogs show. This will need to get investigated again when transitioning off of RepoToolset.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Setup/PackageContentTests.cs
```diff
@@ -14,27 +15,29 @@ namespace Microsoft.VisualStudio.Setup
     [UsesVerify]
     public sealed class PackageContentTests
     {
+        private const string XmlSignature = "package/services/digital-signature/xml-signature";
```
MiYanni, line 18: With the use of MicroBuild signing, there are 3 new `digital-signature` files added to the NuGet packages. This one in specific, `xml-signature`, has a generated name at the time of signing. Thus, it cannot be hard-coded ahead of time as part of the test txt files. So, we filter out checking to see if that file is present.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Setup/PackageContentTests.cs
```diff
@@ -62,7 +65,7 @@ private static IEnumerable<string> GetPackageContents(string vsixName)
             
             using var archive = ZipFile.OpenRead(vsixPath);
             
-            return archive.Entries.Select(entry => entry.FullName);
+            return archive.Entries.Select(entry => entry.FullName).OrderBy(fn => fn);
```
MiYanni, line 68: The test was changed so that the file entries are now in alphabetical order! The order of the files before was based on the repacking process that RepoToolset did during the signing phase. When removing that signing process, this file order was no longer in a determined order. Thus, we explicitly order the entries here, and I've updated the txt files to be in alphabetical order to match.

NuGet.config
```diff
@@ -3,7 +3,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <!--TODO: Merged PIAs - evaluate the need of this once packages are in public feeds-->
+    <!-- TODO: Keep in-sync with build\import\Packages.targets -->
```
drewnoakes, line 6: nit:

```suggestion
    <!-- NOTE: Keep in-sync with build\import\Packages.targets -->
```

To me, TODO implies some pending task that can be completed, whereas this message is a caution for ongoing changes.

build/ci/official.yml
```diff
@@ -18,37 +21,61 @@ variables:
...
-- task: ms-vseng.MicroBuildTasks.30666190-6959-11e5-9f96-f56098202fef.MicroBuildSigningPlugin@1
+- task: MicroBuildSigningPlugin@4
   displayName: Install Signing Plugin
   inputs:
     signType: real
     esrpSigning: true
-    
-- task: ms-vseng.MicroBuildTasks.32f78468-e895-4f47-962c-58a699361df8.MicroBuildSwixPlugin@1
+
+# This is use as a PackageReference with ExcludeAssets="all", which means it is not executed.
```
drewnoakes, line 36:
```suggestion
# This is used as a PackageReference with ExcludeAssets="all", which means it is not executed.
```

build/ci/official.yml
```diff
@@ -18,37 +21,61 @@ variables:
...
+    AccessToken: '$(System.AccessToken)'
+    feedSource: 'https://devdiv.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json'
+    ShouldSkipOptimize: false
+    NumberCommitsToSearch: '100'
+
+###################################################################################################################################################################
+# BUILD SOLUTION
+###################################################################################################################################################################
 
-# TODO: Merged PIAs - reset to /ibc once private feeds are removed and Microsoft.DotNet.IBCMerge is reset
```
drewnoakes, line 40: Does IBC still play a role in our builds at all? Can it be removed from `build.cmd` and elsewhere?

build/ci/official.yml
```diff
@@ -61,117 +88,182 @@ steps:
...
       **/*.dll
-      **/*.exe'
-    publishSymbols: true
+      **/*.exe
+    # NOTE: This produces a warning about indexing (if enabled) but will still publish the symbols. See: https://github.com/microsoft/azure-pipelines-tasks/issues/15605
+    IndexSources: false
+    PublishSymbols: true
     SymbolServerType: TeamServices
-    SymbolsProduct: 'Roslyn Project System'
+    SymbolsProduct: 'DotNet Project System'
```
drewnoakes, line 120: Is this just a display string? If so the correct branding would be:

```suggestion
    SymbolsProduct: '.NET Project System'
```

build/ci/official.yml
```diff
@@ -61,117 +88,182 @@ steps:
...
+# https://docs.microsoft.com/azure/devops/pipelines/tasks/build/index-sources-publish-symbols
 - task: PublishSymbols@2
-  displayName: Archive symbols to VSTS
+  displayName: Publish Symbols to Azure Artifacts
   inputs:
-    SymbolsFolder: $(Build.ArtifactStagingDirectory)/symbols
-    SearchPattern: 
-      '**/*.pdb
+    SymbolsFolder: '$(Build.ArtifactStagingDirectory)/Symbols'
+    SearchPattern: |
```
drewnoakes, line 112: TIL about the pipe in YAML: https://stackoverflow.com/a/18708156/24874

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/Setup/PackageContentTests.cs
```diff
@@ -14,27 +15,34 @@ namespace Microsoft.VisualStudio.Setup
     [UsesVerify]
     public sealed class PackageContentTests
     {
+        // These files are only added as part of signing.
+        private const string DigitalSignature = "package/services/digital-signature";
+        private const string Rels = "_rels/.rels";
```
MiYanni, line 20: With the use of MicroBuild signing, there are 3 new `digital-signature` files added to the NuGet packages. The `.rels` file gets added only for ProjectSystem/Editors during signing. It exists in Common because that uses a different packing process. We simply ignore these extra files for the purpose of this test.

- _rels/.rels
- package/services/digital-signature/_rels/origin.psdor.rels
- package/services/digital-signature/origin.psdor
- package/services/digital-signature/xml-signature/[RANDOM].psdsxs

There is a special case for the Common test as it ends with these generated files. Thus, the ending `,` logic changes. So, for non-signed builds, we need to replace the non-comma line with a comma line.

build/ci/official.yml
```diff
@@ -61,117 +88,182 @@ steps:
...
+
+###################################################################################################################################################################
+# PUBLISH OPTPROF METADATA
+###################################################################################################################################################################
+
+# This is required to deploy the tests to devdiv.artifacts.visualstudio.com for the 'Deploy tests' step in the release pipeline.
+- task: artifactDropTask@0
+  displayName: 'Publish to Artifact Services: ProfilingInputs'
+  inputs:
+    dropServiceURI: 'https://devdiv.artifacts.visualstudio.com'
```
drewnoakes, line 145: Is the use of `visualstudio.com` correct here? I thought these URLs had been migrated to `azure.com`. That's certainly the case for AzDO boards and repos.

There are other usages of `visualstudio.com` elsewhere in this file too.

build/import/Packages.targets
```diff
@@ -2,6 +2,7 @@
 <Project>
 
   <PropertyGroup>
+    <!-- TODO: Keep in-sync with NuGet.config at repo root -->
```
drewnoakes, line 5: nit, following earlier:

```suggestion
    <!-- NOTE: Keep in-sync with NuGet.config at repo root -->
```

setup/Directory.Build.targets
```diff
@@ -10,4 +11,11 @@
   -->
   <Target Name="IncludeTransitiveProjectReferences" >
   </Target>
+
+  <!-- Copies the vsix files after they've been signed to the different output folders. -->
+  <Target Name="CopyVsixAfterSigning" DependsOnTargets="AddVsixForSigning" AfterTargets="SignFiles">
+    <Copy SourceFiles="$(OutDir)$(AssemblyName).vsix" DestinationFolder="$(VisualStudioSetupInsertionPath)" />
+    <Copy SourceFiles="$(OutDir)$(AssemblyName).vsix" DestinationFolder="$(VisualStudioSetupOutputPath)" />
+  </Target>
```
drewnoakes, line 19: Do these copies happen when building locally in VS? Ideally they wouldn't do, right?

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
drewnoakes: One common pattern that avoids this kind of duplication is to specify the `DependsOnTargets` as a property, such as `$(AddVsixForSigningDependsOnTargets)`. Perhaps you can use that here too.

src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
```diff
@@ -1,17 +1,21 @@
...
 <Project Sdk="Microsoft.NET.Sdk">
+
   <Import Project="..\..\build\import\VisualStudioDesigner.props" />
+
   <PropertyGroup>
     <!-- TODO: Function doesn't return a value on all code paths (https://github.com/dotnet/project-system/issues/2592) -->
     <NoWarn>$(NoWarn);42353;NU5125</NoWarn>
     <StrongNameKeyId>Microsoft</StrongNameKeyId>
     <TargetFramework>net472</TargetFramework>
+    <IsProjectSystem>false</IsProjectSystem>
```
drewnoakes, line 11: This `IsProjectSystem` property caught my attention. Is there a clearer name?

drewnoakes (commented):
This is clearly a heroic amount of work and at this point you are the expert in this area, so I defer to your judgement on merging it in. We're early in 17.2, so the timing is good.

I left a few questions, and look forward to the docs you mentioned. Some additional questions:

- How will we verify the performance impact of these changes?
- Do we need to review the scenario used to gather profile data?

build/ci/official.yml
MiYanni: IBC is simply the tool used to merge optimization data into an assembly. The flags in `build.cmd` are quite out-of-date but I didn't want to go through and change the calls to `build.cmd` that assume those flags are there. Instead, I've noted that I want to clean up and simplify the build hierarchy, which would include removing unused flags, as a separate issue I'll make. I've removed IBC usage/references from everywhere else except `build.cmd` in this PR. Note: OptProf itself uses IBCMerge.exe but that is part of the plugin's responsibilities. So we don't need to do anything manually with it.

NuGet.config
MiYanni: Fixed.

build/ci/official.yml
MiYanni: Fixed.

build/ci/official.yml
MiYanni: I actually hunted for where this information was used and couldn't find an answer. However, it is the only place we still seem to use the term "Roslyn Project System" in the repo, so I figured updating it made sense. I changed it to use `.NET` now so we should be consistent with the rest of the repo/branding.

build/import/Packages.targets
MiYanni: Fixed.

src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
MiYanni: The intent is to differentiate between the project system projects (`Managed.dll` and `Managed.VS.dll`) versus the designer projects (`Editors.dll` and `AppDesigner.dll`). Any ideas for a better name? Right now, the OptProf test used only works for the project system projects. So, we can't have that running for the others.

setup/Directory.Build.targets
MiYanni: This target is for the setup projects (the ones that build the NuGet packages). Is it normal to run those locally? I think that might happen when running `build.cmd`. I can create a binlog and take a look.

build/ci/official.yml
MiYanni: I think you're referring that these go to the same place:
- https://devdiv.visualstudio.com/
- https://dev.azure.com/devdiv/

They just have a redirect service running for this stuff. I haven't seen a newer version of this endpoint, since it is the drop service endpoint and not a webpage endpoint. Just took a quick search in //index and the usage of this task, `artifactDropTask`, use this same endpoint. If there is an alternative, no one is using it. Got it from the docs here: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/1877/OptProf?anchor=build-definition-tasks

build/ci/official.yml
MiYanni: This was an interesting find. The current search pattern was broken. If you look at any of our current builds, this task will say "0 files found". For example: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=5636220&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=8e355ed5-37f3-5c2a-bb8d-5efc727482d6&l=9

I don't know how long long it hasn't been working, but this change fixes in. In my test builds, you can see it actually publishing the symbols: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=5637304&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=8e355ed5-37f3-5c2a-bb8d-5efc727482d6&l=9

So, this is basically a bug fix in our pipeline.

setup/Directory.Build.targets
MiYanni: Looking into it, the nupkg files do get built when you run `build.cmd`. I looked at the binlog and these new targets do not get ran. However, the reason that they don't run is interesting. The `SignFiles` target only exists when the MicroBuild signing plugin is installed. Therefore, the binlog shows this:
> The target "SignFiles" listed in a BeforeTargets attribute at "C:\Code\project-system\src\Directory.Build.targets (18,42)" does not exist in the project, and will be ignored.

So, the reason these targets don't run because `SignFiles` itself doesn't exist locally. So, that's some good knowledge to move forward with. I think keeping this implementation now is fine. However, one of the goals of cleaning up the build process would be reducing the amount of file copying (which is extensive at the moment). I'll put this information into an issue when I make one for replacing the RepoToolset Vsix process. This copying only exists because it needs to feed into that.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
MiYanni: I ended up doing this and set that property in the corresponding projects. Now, this target is in the `setup/Directory.Build.targets`.

src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
MiYanni: I have changed this to `IsManagedProjectSystemProject`. That shouldn't clash with anything that might exist and is more specific.

drewnoakes: Thanks, that's clearer somehow.

build/ci/official.yml
```diff
@@ -18,37 +21,61 @@ variables:
...
+# BUILD SOLUTION
+###################################################################################################################################################################
 
-# TODO: Merged PIAs - reset to /ibc once private feeds are removed and Microsoft.DotNet.IBCMerge is reset
 - script: $(Build.SourcesDirectory)\build.cmd /build /test /ci /sign /diagnostic /no-deploy /no-integration /no-ibc /no-clearnugetcache /configuration $(BuildConfiguration)
   displayName: Build ProjectSystem.sln
 
-- task: Ref12.ref12-analyze-task.ref12-analyze-task.Ref12Analyze@0
-  displayName: Produce Codex index for http://index
+# Note: Does not verify that .vsix files are signed.
```
tmeschter, line 63: Is that a limitation of `MicroBuildCodesignVerify` or a choice on our part?

build/proj/CoreBuild.proj
tmeschter: Is "test" signing an option?

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
```diff
@@ -1,24 +1,32 @@
 ﻿<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project Sdk="Microsoft.NET.Sdk">
+
   <Import Project="..\..\build\import\VisualStudio.props" />
-  
+
   <PropertyGroup>
+    <!-- VSIX -->
+    <VsixSigningDependencyTarget>SwixBuild</VsixSigningDependencyTarget>
```
tmeschter, line 8: What's this?

tmeschter (approved):

build/ci/official.yml
MiYanni: I don't actually know for sure, but I feel like it is a limitation. When you go to add this task, here are the options:

image

If you click the `About this task` link, it only goes to our DevDiv wiki. The information for it is quite... malnourished: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/653/Codesign-Verification

The only setting I could look into is which folders we target. Maybe if I target a different folder, I can have it verify the `.vsix` file. But there don't seem to be settings on which file extensions the tool will verify. 🤔

build/proj/CoreBuild.proj
MiYanni: Ah, yes! Thank you for reminding me. Sorry, a lot of stuff to keep track of.

What I can do is add a pipeline variable that sets the sign type. This is what CPS does in their pipeline:
image

So, we can change that value that way. As it stands, I didn't modify our pipeline, but I can add an extra variable for this with no impact to our current builds.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
MiYanni: In my description in the PR, I describe this in the `MSBuild Properties` section.

build/import/HostAgnostic.props
drewnoakes: My understanding of these is that anyone who references our packages will now inherit a reference to these packages too. Should these be `PrivateAssets="all"`?

build/import/Packages.targets
```diff
@@ -2,6 +2,7 @@
 <Project>
 
   <PropertyGroup>
+    <!-- NOTE: Keep in-sync with NuGet.config at repo root -->
```
drewnoakes, line 5: Just an idea, but we could add a unit test to verify these are in sync.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
drewnoakes: A more idomatic name would be `AddVsixForSigningDependsOn`. See the MSBuild common targets for other examples.

drewnoakes (approved):

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
MiYanni: This has been renamed to `AddVsixForSigningDependsOn`. Thanks!

build/import/HostAgnostic.props
MiYanni: Good find! https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

That's something that wouldn't have been obvious to me from simply building everything through the infrastructure. I wonder if there is a way we could test for that... For now, I've added this to those elements. Thanks!

build/proj/CoreBuild.proj
MiYanni: Ah ha! Did some research and it seems the preferred method for something like this is [runtime parameters](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops&tabs=script) in the yaml file. I've done this and here's what it looks like:

image

The default is `real` so we should be good to go. If you want to do a manual run of `test`, select it and it'll change the value when it installs the MicroBuild signing plugin.

build/ci/official.yml
MiYanni: Created an issue for this here: https://github.com/dotnet/project-system/issues/7868

setup/Directory.Build.targets
MiYanni: Created an issue for this here: https://github.com/dotnet/project-system/issues/7869

build/ci/official.yml
MiYanni: I added the folder in which we output the .vsix and now it is checking those files! It even checks that the files in the .vsix are signed properly. Run log here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5686095&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=0881e0b9-84b3-5ca0-e860-5a43d1ab541c

build/import/Packages.targets
MiYanni: I think there is a way to eliminate having 2 of them. I'd rather opt for that strategy. Added an issue here: https://github.com/dotnet/project-system/issues/7870

MiYanni:
@drewnoakes 
> How will we verify the performance impact of these changes?
I will be making a tool we can use to test the performance based on the code you wrote. I'd hope that we can have it in the repo here so we could use it from time to time. I'll keep the commit IDs for before and after this PR, so I can directly compare using the tool.

> Do we need to review the scenario used to gather profile data?
That would be great. This is using this test: `Microsoft.VisualStudio.ProjectSystem.DDRIT.OptProfOpenCloseTest.OpenAndCloseProjectTestSolution`. This is in the VS repo. CPS used to use this test but has migrated to using a "WithCache" version of this test. If we wanted to do this efficiently, we'd craft a specific test to exercise the paths we want to be efficient. For now, this PR adds the infrastructure for us to run tests. We can always add new tests if we think we need to make it better.

MiYanni:
Here is a successful run of the OptProf release pipeline: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?releaseId=1253466&_a=release-pipeline-progress

drewnoakes:
> I will be making a tool we can use to test the performance based on the code you wrote. I'd hope that we can have it in the repo here so we could use it from time to time.

Let's evolve it in the repo it's in for now. I suspect we can build a nice toolkit and share it more broadly outside of this repo. It's not really specific to the project system. We need to extend it for the incremental build performance work too, and I'd rather not fork the tool and dilute the effort.

The validation scenario you construct should be identical to the one used to build the profile data. If it's possible to re-use parts of the test code in the VS repo, that would be idea as others could build on that too.

build/import/Packages.targets
drewnoakes: That'd be even better.

MiYanni:
For testing performance purposes:
- Commit before: https://github.com/dotnet/project-system/commit/3f45eb308715c2bccf510bd863134579ab73b74f
- Commit after: https://github.com/dotnet/project-system/commit/4641c4bff09b447f282a2ef4caffbe9136641fd7

PR #7857, opened 2022-01-25
Make analyzer package assets private and remove some package references

drewnoakes:
Follows from an internal email chain.

This makes our analyzer package references use private assets, so that consumers of those packages do not transitively take on dependencies on those analyzers too.

This PR also removes the `Microsoft.VisualStudio.ProjectSystem*` package references from the App Designer projects, where they are not needed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7857)

### DIFF
--- a/build/import/HostAgnostic.props
+++ b/build/import/HostAgnostic.props
@@ -37,7 +37,7 @@
     <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Query" />
     <PackageReference Include="Microsoft.VisualStudio.Validation" />
     <PackageReference Include="Microsoft.VisualStudio.Threading" />
-    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Analyzers" />
+    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Analyzers" PrivateAssets="all" />
     <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.SDK.Tools" />
     <PackageReference Include="System.Threading.Tasks.Dataflow" />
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
```diff
@@ -12,6 +12,9 @@
     <Summary>AppDesigner designer host implementation for Visual Studio project property pages</Summary>
     <PackageTags>Roslyn Project AppDesigner VisualStudio</PackageTags>
   </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Remove="Microsoft.VisualStudio.ProjectSystem*" />
+  </ItemGroup>
```
MiYanni, line 17: One of the issues I'll be making is about cleaning up our props/targets structure. While this works well enough, it is more like a bandaid on the props that this project includes. I'm just mentioning this here as an FYI.

MiYanni (approved):

build/import/HostAgnostic.props
```diff
@@ -35,7 +35,7 @@
     <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Query" />
     <PackageReference Include="Microsoft.VisualStudio.Validation" />
     <PackageReference Include="Microsoft.VisualStudio.Threading" />
-    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Analyzers" />
+    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Analyzers" PrivateAssets="all" />
```
MiYanni, line 38: Might make sense to add a comment on why these are private. Same with the other analyzer package. Otherwise, that reason will be lost with time and in an internal email thread. :-P

drewnoakes: Done.

PR #7864, opened 2022-01-27
Fix broken "Copy Local" browse object for reference nodes

drewnoakes:
Fixes issue reported in https://developercommunity.visualstudio.com/t/Included-assemblies-set-Copy-Local-FALSE/1640813

For assembly, COM and project references displayed in the Dependencies tree, the "Copy Local" row in the "Properties" window was incorrectly modifying `CopyLocal` metadata rather than `Private`. The `CopyLocal` metadatum has no effect over these references and appears to be ignored.

This change restore the correct behaviour of this property in the UI.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7864)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
@@ -17,13 +17,12 @@
                       DisplayName="Aliases"
                       Separator="," />
 
-  <BoolProperty Name="CopyLocal"
+  <BoolProperty Name="Private"
                 Description="Indicates whether the reference will be copied to the output directory."
                 DisplayName="Copy Local">
     <BoolProperty.DataSource>
       <DataSource HasConfigurationCondition="False"
                   ItemType="Reference"
-                  PersistedName="Private"
                   Persistence="ProjectFile"
                   SourceOfDefaultValue="AfterContext" />
     </BoolProperty.DataSource>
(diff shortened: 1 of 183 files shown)


tmeschter:
Do we know when this regressed?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
```diff
@@ -17,13 +17,12 @@
...
                       Separator="," />
 
-  <BoolProperty Name="CopyLocal"
+  <BoolProperty Name="Private"
                 Description="Indicates whether the reference will be copied to the output directory."
                 DisplayName="Copy Local">
     <BoolProperty.DataSource>
       <DataSource HasConfigurationCondition="False"
                   ItemType="Reference"
-                  PersistedName="Private"
```
tmeschter, line 26: The `PersistedName` is supposed to control what actually gets written to the project file. So while this change effectively works around the problem, I suspect there is a deeper bug here.

tmeschter (commented):
This will work, but I think there is some deeper bug with `DataSource.PersistedName` not being honored.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/AssemblyReference.xaml
drewnoakes: Agreed. We have many other usages of `PersistedName`, however all on properties that are either non-visible or read only. As far as I can tell, they are all working correctly. I do not know whether this property ever worked correctly.

@tmeschter do you have a sense of which component might be at fault here?

drewnoakes:
I filed https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1468948 to track the underlying bug.

My vote would still be to merge this as it's a safe change with no negative consequences and addresses the regression. @tmeschter are you on board with that?

tmeschter (approved):

tmeschter:
@drewnoakes Yeah, so long as we're tracking the underlying issue somewhere I'm fine with taking this as well.

PR #7865, opened 2022-01-27
Don't require the UI thread for settings

tmeschter:
Visual Studio settings are exposed through the `Microsoft.VisualStudio.Settings.ISettingsManager`. We hide this behind our `IProjectSystemOptions` interface, partly so we can access these options in the Microsoft.VisualStudio.ProjectSystem.Managed project (which doesn't depend on VS-specific interfaces). Currently, our implementation of `IProjectSystemOptions` treats `ISettingsManager` as being bound to the UI thread, which means it explicitly transitions to the UI thread before retrieving the service or calling methods on it.

However, this is unnecessarily restrictive, as the `ISettingsManager` documentation explicitly states that its methods may be called on any thread. Here we remove the transitions to the main thread, and access `ISettingsManager` through an `IVsService<T>` instead of `IVsUIService<T>`.

The immediate practical effect of this change is the Fast Up-to-Date Check will no longer need to transition to the main thread once (or more) for every project being built. This will also allow me to simplify the `ImplicitlyTriggeredDebugBuildManager`--it caches a certain setting value and listens for events to reset the value, presumably to avoid a transition to the main thread every time the value is needed. After this change I think we can get rid of all of that.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7865)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
@@ -59,20 +59,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Build
             ErrorHandler.ThrowOnFailure(_solutionBuildManager.AdviseUpdateSolutionEvents3(this, out _cookie3));
 
             _skipAnalyzersForImplicitlyTriggeredBuild = await _options.GetSkipAnalyzersForImplicitlyTriggeredBuildAsync(cancellationToken);
-            _options.RegisterOptionChangedEventHandler(OnOptionChangedAsync);
+            await _options.RegisterOptionChangedEventHandlerAsync(OnOptionChangedAsync);
         }
 
-        protected override Task DisposeCoreAsync(bool initialized)
+        protected override async Task DisposeCoreAsync(bool initialized)
         {
             if (initialized)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
```diff
@@ -59,20 +59,18 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
...
-        protected override Task DisposeCoreAsync(bool initialized)
+        protected override async Task DisposeCoreAsync(bool initialized)
         {
             if (initialized)
             {
                 (_solutionBuildManager as IVsSolutionBuildManager2)?.UnadviseUpdateSolutionEvents(_cookie);
                 _solutionBuildManager!.UnadviseUpdateSolutionEvents3(_cookie3);
 
-                _options.UnregisterOptionChangedEventHandler(OnOptionChangedAsync);
+                await _options.UnregisterOptionChangedEventHandlerAsync(OnOptionChangedAsync);
```
drewnoakes, line 72: This could return the task rather than await it, to avoid a state machine.

drewnoakes (approved):
Nice!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
tmeschter: Done.

PR #7875, opened 2022-01-31
Localized file check-in by OneLocBuild Task: Build definition ID 981: Build ID 1582980

MiYanni:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7875)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/xlf/AssemblyReference.xaml.zh-Hant.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/xlf/AssemblyReference.xaml.zh-Hant.xlf
@@ -34,12 +34,12 @@
       </trans-unit>
       <trans-unit id="BoolProperty|Private|DisplayName">
         <source>Copy Local</source>
-        <target state="translated">複製到本機</target>
+        <target state="new">Copy Local</target>
         <note />
       </trans-unit>
       <trans-unit id="BoolProperty|Private|Description">
         <source>Indicates whether the reference will be copied to the output directory.</source>
-        <target state="translated">表示是否將參考複製到輸出目錄。</target>
+        <target state="new">Indicates whether the reference will be copied to the output directory.</target>
         <note />
       </trans-unit>
       <trans-unit id="BoolProperty|EmbedInteropTypes|DisplayName">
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/xlf/AssemblyReference.xaml.zh-Hant.xlf
```diff
@@ -1,97 +1,97 @@
...
+        <note />
+      </trans-unit>
+      <trans-unit id="StringListProperty|Aliases|Description">
+        <source>A comma-delimited list of aliases to this reference.</source>
+        <target state="translated">這個參考的逗號分隔別名清單。</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="BoolProperty|Private|DisplayName">
+        <source>Copy Local</source>
+        <target state="new">Copy Local</target>
```
drewnoakes, line 37: This looks wrong. If you disable whitespace changes in the diff you can see that this PR just undoes a bunch of translations. It's unclear why it's doing that.

drewnoakes (commented):
Not sure about this one. @MiYanni does this seem right to you?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/xlf/AssemblyReference.xaml.zh-Hant.xlf
drewnoakes: image

drewnoakes:
Actually, I think this follows from #7864 in which I tried to preserve the translations here. Seems like the upstream tooling didn't optimize the change I made, likely having its own source of truth for this data.

MiYanni:
@drewnoakes Yes. I tried the same thing in the past too. You can't force translations to persist if the items themselves change or if you manually change the translations. Since their 'source of truth' doesn't match what we give them, they assume it is invalid and mark it as `new`. As it stands, we shouldn't need to review these PRs. I did have them add that description to the PRs that are generated. It used to just say `null` in the past, such as this: https://github.com/dotnet/project-system/pull/7527#issue-981622816

I'm also working on getting my creds to no longer create these PRs. I figured out how they do it themselves in Arcade. Asking people to see if I can get more info on it.

PR #7878, opened 2022-02-01
Simplify the `ImplicitlyTriggeredDebugBuildManager`

tmeschter:
Ostensibly, the `ImplicitlyTriggeredDebugBuildManager` just tracks whether or not a particular build is triggered by F5/Ctrl+F5 (thus an "implicit" build as it is one part of a larger action) and updates the `IImplicitlyTriggeredBuildManager` accordingly. Components that need to know if an implicit build is starting can then query the `IImplicitlyTriggeredBuildManager` implementation.

However, the implementation ties the implicit build tracking in general to the more-specific "Skip analyzers for implicitly triggered builds" feature. That is, we currently only consider a build implicit if it is due to F5/Ctrl+F5 _and_ the feature is turned on.

We would now like to make other features--unrelated to "Skip analyzers for implicitly triggered builds"--also consider whether or not a build is implicit, and so we need to fully separate the implicit build tracking from the work of skipping analyzers.

Here we update `ImplicitlyTriggeredDebugBuildManager` so it has no knowledge of skipping analyzers and always sets the flag on `IImplicitlyTriggeredBuildManager` appropriately. The check for the setting has been moved down in the build properties provider that sets the actual MSBuild properties used by the build to turn off the analyzers.


MiYanni (commented):
`IImplicitlyTriggeredBuildManager` is marked as obsolete in CPS with the message:
> Intended only for use between internal Microsoft components, and likely to be removed in future. Do not take a dependency upon this interface.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
```diff
@@ -13,14 +12,13 @@
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Build
 {
     /// <summary>
-    /// Tracks implicitly builds triggered as part of F5/Shift+F5 debugging commands and
-    /// skips analyzer execution for these builds by invoking into <see cref="IImplicitlyTriggeredBuildManager"/>.
+    /// Tracks implicitly builds triggered as part of F5/Ctrl+F5 debug/launch commands and
+    /// updates the <see cref="IImplicitlyTriggeredBuildManager"/> appropriately.
```
mavasani, line 16: Note that IImplicitlyTriggeredBuildManager interface is also used by builds triggered as part of Run Tests/Debug Tests commands. Unit test library themselves track invoking the APIs on this interface, it doesn’t rely on this type to detect an implicit build as part of those commands.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
```diff
@@ -13,14 +12,13 @@
...
     /// </summary>
     [Export(ExportContractNames.Scopes.UnconfiguredProject, typeof(IProjectDynamicLoadComponent))]
     [AppliesTo(ProjectCapability.DotNet)]
     internal class ImplicitlyTriggeredDebugBuildManager : OnceInitializedOnceDisposedAsync, IProjectDynamicLoadComponent, IVsUpdateSolutionEvents2, IVsUpdateSolutionEvents3
```
mavasani, line 20: Do we want to rename this type to remove “Debug” in the name, given you plan to use this type for other kinds of implicit builds in the near future?

mavasani (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
tmeschter: Yeah, I tracked down those other uses of `IImplicitlyTriggeredBuildManager` while I was trying to understand why the interface was defined in CPS. This works out well for me, because I will need to distinguish between implicit builds started as part of F5/Ctrl+F5, and those started as part of other operations.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/ImplicitlyTriggeredDebugBuildManager.cs
tmeschter: The type will continue to be specific to F5/Ctrl+F5 scenarios, though I'm going to update it to retrieve other information related to those scenarios such as the set of startup projects.

I did think about renaming it as we often use "Debug" when we're really referring to launching a project, with or without the debugger. If I do that I'll want to update a bunch of type names.

tmeschter:
@MiYanni Yeah, the `IImplicitlyTriggeredBuildManager` interface doesn't really belong in CPS (it doesn't use it) but it needs to be available even when the .NET Project System isn't installed. When we eventually figure out where it should live we don't want to break a bunch of code we aren't aware of, hence the warning.

PR #7884, opened 2022-02-02
Add a global setting

tmeschter:
Part of [AB#1438932](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1438932).

Add a VS-wide setting to turn single-target builds on or off, and expose it in our Tools | Options | Project and Solutions | SDK-Style Projects page. Note that it is off by default for the time being. It is in good enough shape for testing, but we need further updates to the Fast Up-to-Date Check to make it fully complete.

Here's what the options page looks like:

image

Feel free to pick apart the wording or UI design. :-) I'd prefer if the explanatory text ("When starting a project that targets...") was aligned with the checkbox _text_ rather than the box itself, but it's not immediately clear to me if there is a good way to do that.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7884)


davkean:
Can someone point me to a PR/doc how this works?

davkean:
I've read https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1438932 and I cannot picture how this is going to work given that MSBuild picks the target. Imagine this:

Project B (net45) -> Project A (net45, netcoreapp5.0)
Project C (netcoreapp5.0) -> Project A (net45, netcoreapp5.0)

Can you explain how this will work?

tmeschter:
It's worth starting with the motivating scenario, which involves a MAUI app targeting multiple platforms. That is, the app project includes something like:

``` xml
<TargetFrameworks>net6.0-iOS;net6.0-android;net6.0-windows</TargetFrameworks>
```

Currently, building this app project will cause us to build it for each of the listed target frameworks. The builds of some of these targets will be expensive; for example, the iOS build may need to connect to a build system on another computer. In an F5 or Ctrl+F5 scenario you don't really care about building _all_ of the target frameworks since you're only going to launch the app for one of them; the build of the others is overhead and only serves to slow down the inner loop.

So we're making a very limited change to address that particular scenario.

**IF**:

- There is only one startup project
- and it targets multiple frameworks
- and we are running the build as part of F5 or Ctrl+F5

**THEN**:

- Limit the build of that project (and that project _only_) to a single framework by explicitly setting the `TargetFramework` in the global build properties to the active debug framework.

In your example above, I assume you mean that Project B has a `<ProjectReference>` to Project A, as does Project C. In that case, if either B or C is the startup project then we won't make any changes to how any of the projects build because B and C only target a single framework as it is, and A is not the startup project.

If A is the startup project then we won't be building B or C, and we can adjust the `TargetFramework` for A.

davkean:
That setting description/name gives the impression that the feature is more than that.

tmeschter:
Yes, I see what you mean. I'll work on the wording.

tmeschter:
I've updated the wording.

src/Microsoft.VisualStudio.Editors/OptionPages/GeneralOptionPageResources.resx
```diff
@@ -144,4 +144,13 @@
...
     <value>File Nesting Settings</value>
   </data>
+  <data name="General_MultiTargetingSettings_Title" xml:space="preserve">
+    <value>Multi-Targeting Settings</value>
+  </data>
+  <data name="General_PreferSingleTargetBuildsOnLaunch" xml:space="preserve">
+    <value>Prefer single-target builds when starting a project.</value>
+  </data>
+  <data name="General_PreferSingleTargetBuildsOnLaunchDescription" xml:space="preserve">
+    <value>When starting a project that targets multiple .NET frameworks the build time can be reduced by limiting the build to a single framework.</value>
```
MiYanni, line 154: My English sense might be wrong here, but should this have a comma? I think the rule of thumb is that there should be a comma if you can take the first part of the sentence and move it to the end. Ah, found it. It is called an [introductory clause (2nd one)](https://www.iue.edu/student-success/coursework/commas.html):
```suggestion
    <value>When starting a project that targets multiple .NET frameworks, the build time can be reduced by limiting the build to a single framework.</value>
```

MiYanni (approved):

PR #7887, opened 2022-02-03
Limit startup projects to a single framework

tmeschter:
This change introduces a new global build property provider, `StartupProjectSingleTargetGlobalBuildPropertyProvider`, that modifies the build such that a multi-targeting startup project will only build the one relevant target in F5 or Ctrl+F5 scenarios.

There are multiple conditions that we check:

- The project must be multi-targeting (sometimes called "cross targeting"). If it isn't multi-targeting then there is no point in messing with the build properties.
- The build must be implicitly triggered. This means that the build is occurring as an incidental part of some other operation, such as F5, Ctrl+F5, running tests, debugging tests, etc. If the user does an explicit Build or Rebuild we want to continue to build everything.
- There must be a single startup project specified. If there are _no_ start up projects specified, then we don't know which project to limit to a single framework. If there are _multiple_ start up projects then they may depend on one another in ways that make limiting the build of one or the other invalid.
- The current project must be that single startup project.
- The project must have the "SingleTargetBuildForStartupProjects" capability. This allows us to limit this functionality to certain tested project types for the time being.

We will also want to support a VS-wide setting for this feature so the user can turn it off entirely. The setting, and use of it in `StartupProjectSingleTargetGlobalBuildPropertyProvider`, will be added in a future change.

To support these checks, we need to know the set of designated startup projects. We now acquire their full paths when an implicit build starts, and store that in the `ImplicitlyTriggeredBuildManager` for later use.

Related to [AB#1438932](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1438932).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7887)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/IImplicitlyTriggeredBuildState.cs
```diff
@@ -15,6 +16,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.Managed.Build
     [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
     internal interface IImplicitlyTriggeredBuildState
     {
+        /// <summary>
+        /// Indicates if the current build (if any) is implicitly triggered.
+        /// </summary>
         bool IsImplicitlyTriggeredBuild { get; }
```
drewnoakes, line 22: Should this be `bool?` where `null` indicates no active build?

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/IImplicitlyTriggeredBuildState.cs
tmeschter: We're not tracking builds in general, just the implicit ones; so we can't really distinguish explicit build/implicit build/no build.

PR #7889, opened 2022-02-04
Fix broken source item reporting

### LINKED ISSUE #7860: Review ContextState in language service
> Remove it, I'm pretty sure Roslyn handles this correctly on the switch. This is just whether Roslyn goes looking for open documents the first time the file appears.

_Originally posted by @davkean in https://github.com/dotnet/project-system/pull/7855#discussion_r792197596_

drewnoakes:
Fixes [AB#1468419](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1468419)
Fixes #7860

https://github.com/dotnet/project-system/pull/7855 introduced a bug that stopped dynamic files (such as `.razor` files) being passed to Roslyn. The change incorrectly passed the evaluation project update to the source item handlers, rather than the source items update. There was no unit test for this scenario.

This fixes the bug by passing the right data to the right handler, and adding a unit test to ensure coverage of this behaviour.

This PR also re-instates #7862 which we previously reverted.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7889)


davidwengier (approved):
Well now that you have figured out, yep, its really obvious in the other PR :)

I wonder if its worth a roslyn issue.. its not quite "unused parameter", but there might be an argument for "unused tuple value you bothered to provide a nice name for"

PR #7900, opened 2022-02-07
Update branch checklist

drewnoakes:
Some changes to the checklist that I would have found helpful while creating the `17.2.x` branch today. I've also updated the OneNote separately.

@tmeschter please take a look as the most recent person to use this check list.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7900)

### DIFF
--- a/.github/ISSUE_TEMPLATE/branch-checklist.md
+++ b/.github/ISSUE_TEMPLATE/branch-checklist.md
@@ -9,13 +9,23 @@ assignees: ''
 
 _Descriptions of these steps can be found in the team OneNote._
 
+<!-- Replace all ? characters as you work through this. -->
+
 - [ ] Identify base commit for new branch
-- [ ] Create branch on GitHub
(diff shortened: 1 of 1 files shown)


tmeschter (approved):

MiYanni (approved):
I was actually going to make changes to this myself too. However, I realized what I wanted to add didn't align with the process of going through this checklist. But it does need done in the same cadence, though. Any suggestions? I wrote this at first and realized it wasn't technically accurate.

>We need to be making tags for new versions of our codebase. I retroactively added some tags in August 2021, but we currently don't have tags for:
> - 16.11.x
> - 17.0.x
> - 17.1.x
>
> This PR adds the step in our branch creation template to make sure we make tags of our current version when we make branches for new version. Basically, when 17.2 was made, we should have created the 17.1.x tag.

PR #7903, opened 2022-02-08
Wrap IVsRunningDocumentTable and IVsSolutionBuildManager

drewnoakes:
The `IVsRunningDocumentTable` and `IVsSolutionBuildManager` interfaces do not expose idiomatic .NET patterns, making it easy to hit bugs like #7891.

This PR creates new internal MEF components for both RDT and SBM events within VS, which wraps their COM-based APIs with more ergonomic patterns.

It also adds async APIs so we can call from any thread.

This change will simplify the fix for https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/1471474 which must also listen for these events.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7903)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/AsyncDisposable.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System;
+using System.Threading.Tasks;
+using IAsyncDisposable = System.IAsyncDisposable;
+
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    internal sealed class AsyncDisposable : IAsyncDisposable
```
MiYanni, line 9: This seems like a good utility that should be outside of the ProjectSystem scope, since it isn't specific to it.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/SolutionBuildManager.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            _vsSolutionBuildManagerService = solutionBuildManagerService;
+        }
+
+        protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
+        {
+            await JoinableFactory.SwitchToMainThreadAsync(cancellationToken);
+
+            _vsSolutionBuildManager2 = await _vsSolutionBuildManagerService.GetValueAsync(cancellationToken);
+            _vsSolutionBuildManager3 = (IVsSolutionBuildManager3)_vsSolutionBuildManager2;
```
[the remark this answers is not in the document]
drewnoakes: Good question. Yes, `SVsSolutionBuildManager` is known to implement that interface.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Utilities/AsyncDisposable.cs
drewnoakes: This would be available to pretty much all code in the project system.

I'm not sure there's enough value here to make it more generally available. You might also expect it to be implemented in terms of `Task` rather than `ValueTask`.

PR #7907, opened 2022-02-09
Don't indent case contents when block

drewnoakes:
This makes VS format case blocks the way we seem to do them in this repo.

## Before

```c#
switch (foo)
{
    case Bar:
        {
            // block content
            break;
        }
}
```

## After

```c#
switch (foo)
{
    case Bar:
    {
        // block content
        break;
    }
}
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7907)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -170,7 +170,7 @@ csharp_new_line_between_query_expression_clauses = true
 
 # Identation options
 csharp_indent_case_contents = true
-csharp_indent_case_contents_when_block = true
+csharp_indent_case_contents_when_block = false
 csharp_indent_switch_labels = true
 csharp_indent_labels = no_change
 csharp_indent_block_contents = true


MiYanni (changes requested):
I thought something looked odd. Case blocks don't need curly braces. They are only creating inner scopes which aren't necessary in most... *cases* 🤣 (terrible code humor). So, we should likely just strip the curly braces.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-switch-statement

## Before

```c#
switch (foo)
{
    case Bar:
        {
            // block content
            break;
        }
}
```

## After

```c#
switch (foo)
{
    case Bar:
        // block content
        break;
}
```

tmeschter:
I don't think the two are mutually exclusive. I'd rather not have the curly braces if we don't need them, but when we do I'd like them to be aligned with the `case` rather than indented.

tmeschter (approved):

.editorconfig
```diff
@@ -170,7 +170,7 @@ csharp_new_line_between_query_expression_clauses = true
 
 # Identation options
 csharp_indent_case_contents = true
-csharp_indent_case_contents_when_block = true
+csharp_indent_case_contents_when_block = false
```
MiYanni, line 173: If we're changing this, I assume it'll produce warnings on the build. This PR should resolve those warnings. My previous comment meant that if you change this and it produces those warnings, then a majority of these cases could just be fixed to no longer contain `{ }`. I'm not saying that this cannot be set to `false`. What I am saying is that setting this to `false` in this PR should contain the changes to remove the extra curly braces in this PR.

MiYanni (changes requested):

drewnoakes:
Braces can be needed to constrain the scope of variables within the blocks.

```c#
switch (i)
{
    case 1:
        int j = 1;
        break;
    case 2:
        int j = 2; // error
        break;
}

switch (i)
{
    case 1:
    {
        int j = 1;
        break;
    }
    case 2:
    {
        int j = 2; // no error
        break;
    }
}
```

MiYanni (approved):
As I mentioned in my original comment about the fact that curly braces can be used for scoping, I decided to check the repo if there was even a situation where this occurs. There was only a single usage of braces within the cases in a switch expression in the file `FlagsStringMatcher.cs` and those cases were not scoping any variables.

I went ahead and picked a set of criteria, and looked through the 18 switch statements in C# in the repo. I created a PR here that outlines the criteria I used to update those switch statements.  https://github.com/dotnet/project-system/pull/7910

Since those changes are independent of the `csharp_indent_case_contents_when_block` editorconfig value, when the change in this PR is combined with the changes in my PR, there will no longer be a warning given for the switch expression in `FlagsStringMatcher.cs`.

PR #7908, opened 2022-02-09
Run operation directly off the UI thread

drewnoakes:
Previously we would use `RunAndForget` on the UI thread, then have that continuation switch to the thread pool.

By using `ForkOptions.StartOnThreadPool` we can invoke directly on the thread pool.

This also allows removing a state machine.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7908)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
@@ -11,7 +11,6 @@ using Microsoft.VisualStudio.Composition;
 using Microsoft.VisualStudio.ProjectSystem.Build;
 using Microsoft.VisualStudio.ProjectSystem.UpToDate;
 using Microsoft.VisualStudio.Shell.Interop;
-using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Build.Diagnostics
 {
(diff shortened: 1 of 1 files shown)


MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
```diff
@@ -49,13 +48,9 @@ private sealed class ProjectChecker : IProjectChecker
...
-                    async () =>
-                    {
-                        await TaskScheduler.Default;
-
-                        await CheckAsync(_projectAsynchronousTasksService.UnloadCancellationToken);
-                    },
-                    unconfiguredProject: _project);
+                    () => CheckAsync(_projectAsynchronousTasksService.UnloadCancellationToken),
+                    unconfiguredProject: _project,
+                    options: ForkOptions.StartOnThreadPool | ForkOptions.CancelOnUnload | ForkOptions.NoAssistanceMask);
```
tmeschter, line 53: What does this one do?

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.ProjectChecker.cs
drewnoakes: Which one do you mean?

- `StartOnMainThread` Ensures that the fork begins execution on the main thread.
- `CancelOnUnload` Allows not executing the delegate if the project context is being disposed of before it would otherwise start execution.
- `NoAssistanceMask` = `SuppressRelevance` | `HideLocks`
- `SuppressRelevance` Any `Microsoft.VisualStudio.Threading.JoinableTask` relevance inherited from the caller is suppressed in the fork.
- `HideLocks` Any project locks not propagated to the fork.

PR #7909, opened 2022-02-09
Don't check configurations we aren't building

tmeschter:
Related to [AB#1438932](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1438932).

When a project targets multiple versions of the .NET framework (i.e. it has a `TargetFrameworks` property) CPS represents this as multiple `ConfiguredProject`s--one per framework. When the user builds (either through an explicit build or rebuild, or implicitly as part of F5, Ctrl+F5, running unit tests, etc.) CPS asks us to build _one_ of those specific configurations. This is generally not what we want, and so we force MSBuild to build for all the frameworks (this is accomplished by resetting the `TargetFramework` property to the empty string in the `TargetFrameworkGlobalBuildPropertyProvider` type).

Meanwhile, the Fast Up-to-Date Check feature supports this by examining _all_ relevant configurations--again, regardless of which one CPS requests we build. The inputs and outputs of each individual configuration are compared, and if any input is newer than any output the project as a whole is not considered up-to-date and we need to invoke MSBuild.

However, with the changes in #7884 we _don't_ always build for all the target frameworks. When we know which specific target framework the user is going to launch with F5/Ctrl+F5 we can often limit the build to just that single target framework (please see that PR for details as there are a bunch of conditions that must be met for this to work properly). This can speed up the code/build/debug development cycle by skipping the builds of the irrelevant frameworks.

The problem is that the Fast Up-to-Date Check is still checking all the configurations. After one of these "single target framework" builds, the Fast Up-to-Date Check will always report failure because the other configurations are out of date. So we'll only build one configuration, but we'll always try to build it even when we don't need to.

The fix here is to make the Fast Up-to-Date Check aware of these single target builds and restrict its checks accordingly. If a `TargetFramework` is specified in the global properties for the build, we need only consider the inputs and outputs for the corresponding configuration. Otherwise, we can skip the checks.

Here is some sample build output from a project that targets `net5.0` and `net6.0` but is only building for `net5.0`:

```
Build started...
1>FastUpToDate: Ignoring up-to-date check items with Kind="ImplicitBuild" (UpToDateTest)
1>FastUpToDate: Checking configuration Debug|AnyCPU|net6.0: (UpToDateTest)
1>FastUpToDate:     Skipping up-to-date check: not building this configuration. (UpToDateTest)
1>FastUpToDate: Checking configuration Debug|AnyCPU|net5.0: (UpToDateTest)
1>FastUpToDate: Input 'C:\Users\tomescht\source\repos\UpToDateTest\UpToDateTest\UpToDateTest.csproj.user' is newer (2/9/2022 10:56:10 AM) than earliest output 'C:\Users\tomescht\source\repos\UpToDateTest\UpToDateTest\bin\Debug\net5.0\UpToDateTest.pdb' (2/9/2022 10:55:05 AM), not up-to-date. (UpToDateTest)
1>------ Build started: Project: UpToDateTest, Configuration: Debug Any CPU ------
1>You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
1>Skipping analyzers to speed up the build. You can execute 'Build' or 'Rebuild' command to run analyzers.
1>UpToDateTest -> C:\Users\tomescht\source\repos\UpToDateTest\UpToDateTest\bin\Debug\net5.0\UpToDateTest.dll
========== Build: 1 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========
```
Note there are other ways of implementing this that might be, in some sense, cleaner than this approach. I can imagine updating CPS itself to be aware that we may want to build multiple configurations, and having some interface that allows it to run the Fast Up-to-Date Check individually for each of those configurations rather than assuming we want to build the single "active" configuration. However, that would be a lot of change for something that is straightforward to implement in our existing system.

Also, we have this overarching issue (the mismatch between the CPS concept of a single "active" configuration, and our need to work across multiple equally-important configurations) in other places (for example, the Dependencies node in Solution Explorer). If we are going to solve that problem, I would like us to solve it across the board and not just for one feature.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7909)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
@@ -58,7 +60,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             /// in that compilation. We use this property as a proxy for compilation start time, whereas
             /// the outputs represent compilation end time.
             /// </remarks>
-            private DateTime _lastCheckedAtUtc = DateTime.MinValue;
+            //private DateTime _lastCheckedAtUtc = DateTime.MinValue;
+            private ImmutableDictionary<ProjectConfiguration, DateTime> _lastCheckedAtUtc = ImmutableDictionary<ProjectConfiguration, DateTime>.Empty;
 
             /// <summary>
             /// Lazily constructed Dataflow subscription. Set back to <see langword="null"/> in <see cref="Dispose"/>.
@@ -81,7 +84,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             }
 
             public async Task<bool> RunAsync(
-                Func<UpToDateCheckConfiguredInput, DateTime, CancellationToken, Task<bool>> func,
+                Func<UpToDateCheckConfiguredInput, IReadOnlyDictionary<ProjectConfiguration, DateTime>, CancellationToken, Task<(bool UpToDate, ImmutableArray<ProjectConfiguration> CheckedConfigurations)>> func,
                 bool updateLastCheckedAt,
                 CancellationToken cancellationToken)
             {
@@ -124,19 +127,29 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                         cancellationToken: token);
                 }
 
-                bool result = await func(state.Value, _lastCheckedAtUtc, token);
+                (bool upToDate, ImmutableArray<ProjectConfiguration> checkedConfigurations) = await func(state.Value, _lastCheckedAtUtc, token);
 
                 if (updateLastCheckedAt)
                 {
-                    UpdateLastCheckedAtUtc();
+                    UpdateLastCheckedAtUtc(checkedConfigurations);
                 }
 
-                return result;
+                return upToDate;
             }
 
             public void UpdateLastCheckedAtUtc()
             {
-                _lastCheckedAtUtc = DateTime.UtcNow;
+                UpdateLastCheckedAtUtc(_lastCheckedAtUtc.Keys);
+            }
+
+            public void UpdateLastCheckedAtUtc(IEnumerable<ProjectConfiguration> checkedProjectConfigurations)
+            {
+                DateTime utcNow = DateTime.UtcNow;
+
+                foreach (ProjectConfiguration configuration in checkedProjectConfigurations)
+                {
+                    _lastCheckedAtUtc = _lastCheckedAtUtc.SetItem(configuration, utcNow);
+                }
             }
 
             public void EnsureInitialized()
@@ -192,8 +205,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
 
             void UpdateLastCheckedAtUtc();
 
+            void UpdateLastCheckedAtUtc(IEnumerable<ProjectConfiguration> checkedProjectConfigurations);
+
             Task<bool> RunAsync(
-                Func<UpToDateCheckConfiguredInput, DateTime, CancellationToken, Task<bool>> func,
+                Func<UpToDateCheckConfiguredInput, IReadOnlyDictionary<ProjectConfiguration, DateTime>, CancellationToken, Task<(bool UpToDate, ImmutableArray<ProjectConfiguration> CheckedConfigurations)>> func,
                 bool updateLastCheckedAt,
                 CancellationToken cancellationToken);
         }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
@@ -60,7 +60,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             /// in that compilation. We use this property as a proxy for compilation start time, whereas
             /// the outputs represent compilation end time.
             /// </remarks>
-            private readonly Dictionary<ProjectConfiguration, DateTime> _lastCheckedAtUtc = new();
+            private readonly Dictionary<ProjectConfiguration, DateTime> _lastCheckedAtUtcByConfiguration = new();
 
             /// <summary>
             /// Lazily constructed Dataflow subscription. Set back to <see langword="null"/> in <see cref="Dispose"/>.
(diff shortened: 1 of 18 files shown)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1336,14 +1336,49 @@ static int HashItems(params (string itemType, string[] paths)[] items)
...
+
+            var sourceSnapshot = new Dictionary<string, IProjectRuleSnapshotModel>
+            {
+                [Compile.SchemaName] = SimpleItems("ItemPath1")
+            };
+
+            var itemChangedTime = DateTime.UtcNow.AddMinutes(-4);
+            var outputTime = DateTime.UtcNow.AddMinutes(-3);
+            var inputTime = DateTime.UtcNow.AddMinutes(-2);
+            var lastCheckTime = DateTime.UtcNow.AddMinutes(-1);
```
drewnoakes, line 1355: The up-to-date check currently updates this value each time it's called (on the subscription object IIRC).

We currently track that value across all implicitly active configurations. I think you'll need to change the code to track it per configuration to handle all cases correctly. Take a look at the tests that play with this value to see the kind of trap if helps avoid, and I think you'll be able to construct a test that will fail unless this is handled per configuration.

drewnoakes: The last check time is currently modelled here: https://github.com/dotnet/project-system/blob/c8137374c0f7e6458397627b530ddb14099cb6b9/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs#L49-L61

tmeschter: I see. It sounds like a possible scenario is:
1. The user explicitly builds a project that multi-targets frameworks A and B.
2. During the build a file that is only included in a build of A is updated before the output for A is produced.
   - At this point `lastCheck` < `inputs(A)` < `outputs(A)`. If we were to check *all* configurations now, A would be (correctly) seen as out-of-date.
3. The user then runs the output of framework B (with, say, Ctrl+F5). The up-to-date check skips checking A (since we're not building it) and reports B is up-to-date because `inputs(B)` < `lastCheck` < `outputs(B)`. We update `lastCheck`.
4. The user explicitly builds again. B is, of course, up-to-date. A is *not* up-to-date, but we report that it is because `inputs(A)` < `outputs(A)` < `lastCheck`.
   - We then erroneously skip building A because we thought it was up-to-date, even though it wasn't.

Does that sound right?

tmeschter: @drewnoakes Alright, I've updated the code to address this concern.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
```diff
@@ -124,19 +127,29 @@ public Subscription(IUpToDateCheckConfiguredInputDataSource inputDataSource, Con
...
+                UpdateLastCheckedAtUtc(_lastCheckedAtUtc.Keys);
+            }
+
+            public void UpdateLastCheckedAtUtc(IEnumerable<ProjectConfiguration> checkedProjectConfigurations)
+            {
+                DateTime utcNow = DateTime.UtcNow;
+
+                foreach (ProjectConfiguration configuration in checkedProjectConfigurations)
+                {
+                    _lastCheckedAtUtc = _lastCheckedAtUtc.SetItem(configuration, utcNow);
```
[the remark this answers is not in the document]
tmeschter: Ah. Yeah I don't think I need an `ImmutableDictionary`; I was going to fix that and got sidetracked by a different problem.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -823,11 +823,11 @@ async Task<(bool, ImmutableArray<ProjectConfiguration>)> CheckAsync(UpToDateChec
...
-                        if (!CheckGlobalConditions(logger, configurationLastCheckedAtUtc, validateFirstRun: !isValidationRun, implicitState) ||
-                            !CheckInputsAndOutputs(logger, configurationLastCheckedAtUtc, timestampCache, implicitState, ignoreKinds, token) ||
-                            !CheckMarkers(logger, timestampCache, implicitState) ||
-                            !CheckCopyToOutputDirectoryFiles(logger, timestampCache, implicitState, token) ||
-                            !CheckCopiedOutputFiles(logger, timestampCache, implicitState, token))
+                        if (!CheckGlobalConditions(logger, configurationLastCheckedAtUtc, validateFirstRun: !isValidationRun, implicitState)
+                            || !CheckInputsAndOutputs(logger, configurationLastCheckedAtUtc, timestampCache, implicitState, ignoreKinds, token)
+                            || !CheckMarkers(logger, timestampCache, implicitState)
+                            || !CheckCopyToOutputDirectoryFiles(logger, timestampCache, implicitState, token)
+                            || !CheckCopiedOutputFiles(logger, timestampCache, implicitState, token))
```
MiYanni, line 830: You didn't have to change this! 😝 It was just a comment about it. If we want this style to be standard, we should just add a rule/analyzer in our editorconfig about it. I'm sure there is one that would match this style.

tmeschter:
@drewnoakes Did you want to take another look at this before I merged it in?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
```diff
@@ -58,7 +60,7 @@ private sealed class Subscription : ISubscription
             /// in that compilation. We use this property as a proxy for compilation start time, whereas
             /// the outputs represent compilation end time.
             /// </remarks>
-            private DateTime _lastCheckedAtUtc = DateTime.MinValue;
+            private readonly Dictionary<ProjectConfiguration, DateTime> _lastCheckedAtUtc = new();
```
drewnoakes, line 63: The name `_lastCheckedAtUtcByConfiguration` would be more consistent with other dictionary names in this area of the code. 

This convention makes the variable sound like a dictionary. The current name sounds like a `DateTime`.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
drewnoakes: Apologies, I missed your comment. Yes this looks like a valid scenario, and the unit test looks good.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
MiYanni: Sounds like we should decide if we want that to be the standard then.

PR #7910, opened 2022-02-10
Simplify or refine existing switch statements

MiYanni:
Relates to: https://github.com/dotnet/project-system/pull/7907

I went through all the switch statements (not switch expressions) in the C# code in our repo.
- 18 switch statements were present prior to this PR; 7 remain after this PR
- 1 switch statement in `FlagsStringMatcher.cs` had redundant `{ }` blocks for the cases; removed them
- Many switch statements could be replaced with switch expressions
- Some switch statements could be replaced with pattern matching
- Added a comment for some switch statements in which the value was discarded after conversion to switch expression

Criteria used:
- If all cases returned a value and were a single statement within the case (the exception being the `default` case), then I converted them to switch expressions
- If there were more than 3 cases when converting to a switch expression, I justified (aligned) the RHS values (except if `_` case was a throw statement)
  - The exception was `AbstractExtenderCATIDProvider.cs` as it had custom commenting
- If the switch statement cases had ambiguous scoping (`{ }`), the scoping was removed
- If the switch statement could be simplified using pattern matching into a single if statement, it was converted to an if statement
  - If a pattern is using more than 1 `or`, the values are vertically aligned to match the visual style of the prior switch statement's cases
- If none of the above criteria applied, then I left the switch expression alone

### Note
It is suggested to use the *Split* view to review this PR:

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7910)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
@@ -23,19 +23,37 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Automation
 
             return extenderCATIDType switch
             {
-                ExtenderCATIDType.HierarchyExtensionObject or ExtenderCATIDType.AutomationProject => GetExtenderCATID(ExtendeeObject.Project),               // IVsHierarchy.GetProperty(VSITEMID.Root, VSHPROPID_ExtObjectCATID)
-                                                                                                                                                             // DTE.Project
-                ExtenderCATIDType.HierarchyBrowseObject or ExtenderCATIDType.ProjectBrowseObject => GetExtenderCATID(ExtendeeObject.ProjectBrowseObject),    // IVsHierarchy.GetProperty(VSHPROPID_BrowseObjectCATID)
-                                                                                                                                                             // EnvDTE.Project.Properties
-                ExtenderCATIDType.ConfigurationBrowseObject => GetExtenderCATID(ExtendeeObject.Configuration),                                               // IVsCfgProvider2.GetCfgProviderProperty(VSCFGPROPID_IntrinsicExtenderCATID)/DTE.Configuration
-                ExtenderCATIDType.HierarchyConfigurationBrowseObject or ExtenderCATIDType.ProjectConfigurationBrowseObject => GetExtenderCATID(ExtendeeObject.ConfigurationBrowseObject), // IVsHierarchy.GetProperty(VSHPROPID_CfgBrowseObjectCATID)
-                                                                                                                                                             // EnvDTE.Configuration.Properties
-                ExtenderCATIDType.AutomationProjectItem => GetExtenderCATID(ExtendeeObject.ProjectItem),                                                     // EnvDTE.ProjectItem
-                ExtenderCATIDType.AutomationReference or ExtenderCATIDType.ReferenceBrowseObject => GetExtenderCATID(ExtendeeObject.ReferenceBrowseObject),  // VSLangProject.Reference
-                                                                                                                                                             // EnvDTE.ProjectItem.Properties (when reference)
-                ExtenderCATIDType.FileBrowseObject => GetExtenderCATID(ExtendeeObject.FileBrowseObject),                                                     // EnvDTE.ProjectItem.Properties (when file)
-                ExtenderCATIDType.AutomationFolderProperties or ExtenderCATIDType.FolderBrowseObject => GetExtenderCATID(ExtendeeObject.FolderBrowseObject), // FolderProperties or EnvDTE.ProjectItem.Properties (when folder)
-                ExtenderCATIDType.Unknown or _ => UnknownOrDefault()                                                                                         // EnvDTE.ProjectItem.Properties (when not file, folder, reference)
+                ExtenderCATIDType.HierarchyExtensionObject or                       // IVsHierarchy.GetProperty(VSITEMID.Root, VSHPROPID_ExtObjectCATID)
+                ExtenderCATIDType.AutomationProject =>                              // DTE.Project
+                    GetExtenderCATID(ExtendeeObject.Project),
+
+                ExtenderCATIDType.HierarchyBrowseObject or                          // IVsHierarchy.GetProperty(VSHPROPID_BrowseObjectCATID)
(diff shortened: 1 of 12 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
```diff
@@ -21,41 +21,27 @@ internal abstract class AbstractExtenderCATIDProvider : IExtenderCATIDProvider
             // The latter issue we can do nothing about, however, to address the former, map these types to a truer form it makes 
             // it easier on implementors and maintainers of this to understand the objects we're talking about.
 
-            switch (extenderCATIDType)
+            return extenderCATIDType switch
```
MiYanni, line 24: This is the only conversion from switch statement to switch expression that was a little odd. I tried a few different alignments and settled on sticking with the spirit of the original formatting, which aligned the right-handed comments. If there is another alignment that makes sense, suggest it. If we want to revert to the previous switch statement, I will do so.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Input/Commands/Ordering/OrderingHelper.cs
```diff
@@ -33,20 +33,14 @@ public static Task MoveAsync(ConfiguredProject configuredProject, IProjectAccess
                 project.ReevaluateIfNecessary();
                 ImmutableArray<ProjectItemElement> addedElements = GetAddedItemElements(previousIncludes, project);
 
-                switch (action)
+                // TODO: Should the result (success or failure) be ignored?
+                bool _ = action switch
```
drewnoakes, line 37: Discards like this don't need a type declaration.

```suggestion
                _ = action switch
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
```diff
@@ -21,41 +21,27 @@ internal abstract class AbstractExtenderCATIDProvider : IExtenderCATIDProvider
...
+                ExtenderCATIDType.AutomationReference or ExtenderCATIDType.ReferenceBrowseObject => GetExtenderCATID(ExtendeeObject.ReferenceBrowseObject),  // VSLangProject.Reference
+                                                                                                                                                             // EnvDTE.ProjectItem.Properties (when reference)
+                ExtenderCATIDType.FileBrowseObject => GetExtenderCATID(ExtendeeObject.FileBrowseObject),                                                     // EnvDTE.ProjectItem.Properties (when file)
+                ExtenderCATIDType.AutomationFolderProperties or ExtenderCATIDType.FolderBrowseObject => GetExtenderCATID(ExtendeeObject.FolderBrowseObject), // FolderProperties or EnvDTE.ProjectItem.Properties (when folder)
+                ExtenderCATIDType.Unknown or _ => UnknownOrDefault()                                                                                         // EnvDTE.ProjectItem.Properties (when not file, folder, reference)
```
drewnoakes, line 38: I'm not sure this is clearer. I'm generally a fan of switch expressions, but I think there's a risk here of skipping one of the `or`'d values when reading this code. The previous code was easier to skim read, and when debugging we spend a lot of time skimming quickly through code to understand how all the layers of the system come together. Readability is key for doing that correctly and effectively.

The RHS comments are also less clear in this form.

I expect the compiler to emit basically the same code in both cases. Maybe slightly worse if the local function is not inlined.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Automation/AbstractExtenderCATIDProvider.cs
```diff
@@ -21,41 +21,45 @@ internal abstract class AbstractExtenderCATIDProvider : IExtenderCATIDProvider
...
             // it easier on implementors and maintainers of this to understand the objects we're talking about.
 
-            switch (extenderCATIDType)
+            return extenderCATIDType switch
             {
-                case ExtenderCATIDType.HierarchyExtensionObject:                                // IVsHierarchy.GetProperty(VSITEMID.Root, VSHPROPID_ExtObjectCATID)
-                case ExtenderCATIDType.AutomationProject:                                       // DTE.Project
-                    return GetExtenderCATID(ExtendeeObject.Project);
+                ExtenderCATIDType.HierarchyExtensionObject or                       // IVsHierarchy.GetProperty(VSITEMID.Root, VSHPROPID_ExtObjectCATID)
+                ExtenderCATIDType.AutomationProject =>                              // DTE.Project
```
tmeschter, line 27: _This is not something you need to change for this PR, I just see an opportunity for a minor but interesting conversation._

When splitting an expression across multiple lines, I tend to put operators at the beginning of the next line rather than the end of the previous line. For example, instead of:

``` csharp
if (alpha ||
    beta)
{
    ...
}
```

I would write:

``` csharp
if (alpha
    || beta)
{
    ...
}
```

Of course, this example is so short that in practice I would put the whole expression on one line. 😀 But the advantage of this--when the expression has multiple subexpressions or is long--is that all the operators are lined up on the left and you can immediately see how all the subexpressions compose as you read, rather than waiting for your eye to reach the end of the line.

MiYanni: For now, I won't change the style on this PR, but I agree that figuring out a standard style would be good. Given this example,
```C#
if (extension is
    ".fs" or
    ".fsi" or
    ".fsx" or
    ".fsscript" or
    ".ml" or
    ".mli")
```

Would we have to put the `is` on the next line? Like this?
```C#
if (extension
    is ".fs"
    or ".fsi"
    or ".fsx"
    or ".fsscript"
    or ".ml"
    or ".mli")
```

Or do you keep the `is` on the first line and just have the one, lonely value?
```C#
if (extension is
    ".fs"
    or ".fsi"
    or ".fsx"
    or ".fsscript"
    or ".ml"
    or ".mli")
```

PR #7919, opened 2022-02-14
Allow newlines in command line arguments prop

### LINKED ISSUE #7392: Launch Profiles window should allow line breaks in command lines
**Visual Studio Version**: 17.0 preview2 31430.40.main

When preparing a long/complex command line for debugging a project, it hugely boosts readability to allow newlines in the command line. This allows you to put each argument on its own line, for example.

I did try just updating launchSettings.json to put `\n` in the command line but I found that it introduced args which consisted of just `\n`, for example. I believe some of the underlying handling would need to be updated for this to work.

adamint:
Fixes #7392  by making the CommandLineArguments StringProperty multiline, and then stripping out the newlines in ProjectLaunchTargetsProvider

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7919)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
@@ -267,6 +267,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
                 }
             }
 
+            var commandLineArgs = resolvedProfile.CommandLineArgs != null ? StripCommandLineArgs(resolvedProfile.CommandLineArgs) : null;
+
             // Is this profile just running the project? If so we ignore the exe
             if (IsRunProjectCommand(resolvedProfile))
             {
@@ -284,15 +286,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
                     defaultWorkingDir = workingDirectory;
                 }
 
-                if (!string.IsNullOrWhiteSpace(resolvedProfile.CommandLineArgs))
+                if (!string.IsNullOrWhiteSpace(commandLineArgs))
                 {
-                    arguments = arguments + " " + resolvedProfile.CommandLineArgs;
+                    arguments = arguments + " " + commandLineArgs;
                 }
             }
             else
             {
                 executable = resolvedProfile.ExecutablePath;
-                arguments = resolvedProfile.CommandLineArgs;
+                arguments = commandLineArgs;
             }
 
             string workingDir;
@@ -604,6 +606,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             }
 
             return null;
+        }
+
+        internal static string StripCommandLineArgs(string commandLineArgs)
+        {
+            return commandLineArgs.Replace(Environment.NewLine, " ");
         }
 
         /// <summary>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
@@ -267,7 +267,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
                 }
             }
 
-            string? commandLineArgs = resolvedProfile.CommandLineArgs?.Replace("\r\n", " ").Replace("\r", " ").Replace("\n", " ");
+            string? commandLineArgs = resolvedProfile.CommandLineArgs?
+                .Replace("\r\n", " ")
+                .Replace("\r", " ")
+                .Replace("\n", " ")
+                .Replace("\\r\\n", "\r\n")
+                .Replace("\\n", "\n");
 
             // Is this profile just running the project? If so we ignore the exe
             if (IsRunProjectCommand(resolvedProfile))
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.Debugger.UI.Interfaces.HotReload;
@@ -267,7 +268,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
                 }
             }
 
-            var commandLineArgs = resolvedProfile.CommandLineArgs != null ? StripCommandLineArgs(resolvedProfile.CommandLineArgs) : null;
+            var commandLineArgs = resolvedProfile.CommandLineArgs != null ? Regex.Replace(resolvedProfile.CommandLineArgs, "[\r\n]+", " ") : null;
 
             // Is this profile just running the project? If so we ignore the exe
             if (IsRunProjectCommand(resolvedProfile))
@@ -606,11 +607,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
             }
 
             return null;
-        }
-
-        internal static string StripCommandLineArgs(string commandLineArgs)
-        {
-            return commandLineArgs.Replace(Environment.NewLine, " ");
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=7919) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=7919) before we can accept your contribution.<br/><br/>:x: adamint [sign now](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=7919)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/project-system?pullRequest=7919) it.</sub>

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -267,6 +267,8 @@ private async Task<bool> IsIntegratedConsoleEnabledAsync()
                 }
             }
 
+            var commandLineArgs = resolvedProfile.CommandLineArgs != null ? StripCommandLineArgs(resolvedProfile.CommandLineArgs) : null;
```
drewnoakes, line 270: This can be simplified a little using null propagation `?.`:

```suggestion
            var commandLineArgs = resolvedProfile.CommandLineArgs?.Replace(Environment.NewLine, " ");
```

The `StripCommandLineArgs` method could then be removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -606,6 +608,11 @@ private static async Task<Guid> GetDebuggingEngineAsync(ConfiguredProject config
             return null;
         }
 
+        internal static string StripCommandLineArgs(string commandLineArgs)
+        {
+            return commandLineArgs.Replace(Environment.NewLine, " ");
```
drewnoakes, line 613: Are we sure that incoming new lines will always be `\r\n` (which would be the value of `Environment.NewLine` on Windows)? If `launchSettings.json` is saved with different line endings, I'm not sure they'd be normalised before arriving here.

One option would be:

```c#
return Regex.Replace(commandLineArgs, "[\r\n]+", " ")
```

Or in English: Replace any number of consecutive `\r` and/or `\n` characters with a single space.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
```diff
@@ -26,7 +26,15 @@
...
-                  Description="Command line arguments to pass to the executable." />
+                  Description="Command line arguments to pass to the executable.">
+    <StringProperty.ValueEditors>
+     <ValueEditor EditorType="MultiLineString">
+      <ValueEditor.Metadata>
+        <NameValuePair Name="UseMonospaceFont" Value="True" />
+      </ValueEditor.Metadata>
+    </ValueEditor>
+    </StringProperty.ValueEditors>
+  </StringProperty>
```
drewnoakes, line 37: super minor nit: the indentation is a bit inconsistent here.

drewnoakes (commented):
This looks good. You'll want to do a lot of testing here. As mentioned in comments, what happens if `launchSettings.json` is converted to other line endings (as might happen on a git checkout, for example)? What the value contains multiple newlines between values.

Consider writing a small console application that prints out its arguments and validates that it behaves correctly both with and without that change.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
adamint: Is there any use of \r in modern Windows besides as part of a newline? If so, a safer regex would be (\n|\r\n)+

adamint: > This looks good. You'll want to do a lot of testing here. As mentioned in comments, what happens if `launchSettings.json` is converted to other line endings (as might happen on a git checkout, for example)? What the value contains multiple newlines between values.
> 
> Consider writing a small console application that prints out its arguments and validates that it behaves correctly both with and without that change.

I have done this, but just on Windows. I hadn't thought about your point about mixing different newlines across environments, thank you! I'll change that

adamint: @drewnoakes these are the args I used to test: -arg1\r\n-arg2\narg3\narg4\r\narg5

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
adamint: Fixed the indentation. Sorry!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -267,6 +268,8 @@ private async Task<bool> IsIntegratedConsoleEnabledAsync()
                 }
             }
 
+            var commandLineArgs = resolvedProfile.CommandLineArgs != null ? Regex.Replace(resolvedProfile.CommandLineArgs, "[\r\n]+", " ") : null;
```
tmeschter, line 271: Generally we prefer explicit types over `var`. The exception is when the type is explicit in the initializer.

tmeschter, line 271: You can simplify this to:

``` csharp
string? commandLineArgs = resolvedProfile?.Replace("\r\n", " ").Replace("\r", " ").Replace("\n", " ");
```

Which avoids the need for the explicit null check and the use of a regex. A regex is a fairly heavyweight tool for a pattern of this complexity.

tmeschter (dismissed):
Please add tests covering this behavior to the `ProjectLaunchTargetsProviderTests` type. I've left some other feedback on specific lines of code but the addition of a test is the most important change.

MiYanni (approved):
Same comment as Tom about using string replace instead of regex.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
adamint: Gotcha, noted.

drewnoakes: These are not equivalent in the face of multiple newlines, but I'm not sure it matters. The chained replaces will allocate more memory, but this is not on a hot path.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
```diff
@@ -26,7 +26,15 @@
 
   <StringProperty Name="CommandLineArguments"
                   DisplayName="Command line arguments"
-                  Description="Command line arguments to pass to the executable." />
+                  Description="Command line arguments to pass to the executable.">
```
drewnoakes, line 29: Maybe explain that multi-line is acceptable here:

```suggestion
                  Description="Command line arguments to pass to the executable. You may break arguments into multiple lines.">
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -5,6 +5,7 @@
 using System.ComponentModel.Composition;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Text.RegularExpressions;
```
drewnoakes, line 8: This can be removed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
adamint: Changed.

drewnoakes:
The CI error is easy to fix. You just need to build the code locally on your PC to have the translation files be updated, then push them up too.

adamint:
Gotcha, built and committed.

drewnoakes:
I don't know that it was possible before, but most shells support having arguments that themselves contain newlines. For example, when you commit from the command line with `git commit -m "My message` you can insert newlines that are part of the message argument, up until the closing `"` is met. Any newlines between the `"` are kept.

I don't know if we supported this in the legacy property pages. It is something to test for.

drewnoakes (approved):

PR #7922, opened 2022-02-16
Recreate ApplicationPropertyPage in code.

tmeschter:
This is a *partial* example of recreating ApplicationPropertyPage.xaml in code. Not all properties are present, and the original file still exists. However, this is enough to understand what a real-world property page would look like when defined in code.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.cs
```diff
@@ -0,0 +1,257 @@
...
+                }
+            };
+
+            DynamicEnumProperty targetPlatformIdentifierProperty = new()
+            {
+                Name = "TargetPlatformIdentifier",
+                DisplayName = PPR.ApplicationPage_TargetPlatformIdentifierProperty_DisplayName,
+                Description = PPR.ApplicationPage_TargetPlatformIdentifierProperty_Description,
+                Category = GeneralCategoryName,
+                EnumProvider = "SdkSupportedTargetPlatformIdentifierEnumProvider",
```
drewnoakes, line 170: These could be constants on the provider classes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.cs
```diff
@@ -0,0 +1,257 @@
...
+                    new()
+                    {
+                        // TODO: EditorType could come from a constant
+                        EditorType = "LinkAction",
+                        Metadata =
+                        {
+                            new() { Name = "Action", Value = "URL" },
+                            new() { Name = "URL", Value = "http://go.microsoft.com/fwlink/?LinkID=287120" }
+                        }
+                    }
```
drewnoakes, line 160: Some helper methods for this kind of thing could be helpful.

This could be written as:

```suggestion
                    HyperlinkActionEditor("http://go.microsoft.com/fwlink/?LinkID=287120")
```

MiYanni:
@tmeschter Is this still under active consideration? Might make sense to close this until we decide to revisit the property pages. I also wanted to ask, does this have any added advantages for unit testing scenarios? Like, make them simpler or easier, or make mocking easier. Also, the localization cost for this would take time as we'd have to convert the strings into resources.

PR #7923, opened 2022-02-17
Don't block thread pool threads during file IO

drewnoakes:
Fixes [AB#1484422](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1484422)

The file operations can take a while to complete. The previous code protected these operations and data structures using a monitor lock, which blocks threads. This could lead to thread pool threads being blocked

This change uses an async semaphore, which allows threads that are not granted access to the critical section to yield and do other useful work until they are able to gain access.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7923)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
@@ -68,12 +68,12 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.UpToDate
             }
         }
 
-        public async Task<(int ItemHash, DateTime ItemsChangedAtUtc)?> RestoreStateAsync(string projectPath, IImmutableDictionary<string, string> configurationDimensions)
+        public async Task<(int ItemHash, DateTime ItemsChangedAtUtc)?> RestoreStateAsync(string projectPath, IImmutableDictionary<string, string> configurationDimensions, CancellationToken cancellationToken)
         {
-            await InitializeAsync();
-            await InitializeDataAsync();
+            await InitializeAsync(cancellationToken);
+            await InitializeDataAsync(cancellationToken);
 
-            using (await _lock.EnterAsync())
+            using (await _lock.EnterAsync(cancellationToken))
             {
                 Assumes.NotNull(_dataByConfiguredProject);
 
@@ -83,16 +83,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.UpToDate
                 return null;
             }
 
-            async Task InitializeDataAsync()
+            async Task InitializeDataAsync(CancellationToken cancellationToken)
             {
                 if (_cacheFilePath is null || _dataByConfiguredProject is null)
                 {
-                    string filePath = await GetCacheFilePathAsync(CancellationToken.None);
+                    string filePath = await GetCacheFilePathAsync(cancellationToken);
 
                     // Switch to a background thread before doing file I/O
                     await TaskScheduler.Default;
 
-                    using (await _lock.EnterAsync())
+                    using (await _lock.EnterAsync(cancellationToken))
                     {
                         if (_cacheFilePath is null || _dataByConfiguredProject is null)
                         {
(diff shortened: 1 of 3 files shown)


drewnoakes:
I am going to investigate an additional potential bug, where these files are read more than they need to be.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
```diff
@@ -223,16 +223,19 @@ private static Dictionary<(string ProjectPath, IImmutableDictionary<string, stri
 
         public int OnAfterCloseSolution(object pUnkReserved)
```
davkean, line 224: I question if close actually needs to actually block on saving the cache.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
```diff
@@ -125,9 +125,9 @@ async Task<string> GetCacheFilePathAsync(CancellationToken cancellationToken)
             }
         }
 
-        public void StoreState(string projectPath, IImmutableDictionary<string, string> configurationDimensions, int itemHash, DateTime itemsChangedAtUtc)
+        public async Task StoreStateAsync(string projectPath, IImmutableDictionary<string, string> configurationDimensions, int itemHash, DateTime itemsChangedAtUtc)
         {
```
davkean, line 129: It feels like Cancellation representing project unload should be plumbed through here in a future PR, quick opens and closes are going to block on this completing.

davkean (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
drewnoakes: I wondered the same, but was concerned that VS could exit midway through writing the file. Is there some way I can queue the task to hold back shutdown?

davkean: JoinableTaskFactory that comes from your package block until all pending tasks in its collection have completed on Package.Close (and hence VS shutdown).

But it does not look like CPS's does this for its collection.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
drewnoakes: Fixed in 27e25428330438da6929295b101725d9ad98c0da.

drewnoakes:
> I am going to investigate an additional potential bug, where these files are read more than they need to be.

There was no additional bug here. It was just multiple threads blocked on the lazy load of this data. They will now all yield on the semaphore correctly.

I am waiting to hear from @lifengl how to not block solution close, but also not fail to complete our async work before the process exits. Then I think this is good to go.

davkean:
@drewnoakes I think just merge in regardless - you haven't made this worse in regards to blocking Close.

davkean (approved):

drewnoakes:
Ok, I'll file a follow up PR later if needed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
lifengl: Generally, our pattern is that you start to store the information during PrioritySolutionClosing event and start saving work in the background after capturing all states requiring UI thread.  So the work can be done in the background during other solution closing work.  When you receive solution close event, waits the background task. In that case, almost all work will be done before solution close, instead of blocking the close work.

drewnoakes: Addressed in https://github.com/dotnet/project-system/pull/7930

PR #7928, opened 2022-02-19
VB property pages: Application Framework settings

MiYanni:
This PR is just for visibility at the moment.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7928)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -379,8 +379,8 @@
   <!-- TODO: Shortlink for help URL -->
   <!--Description="Enable modern visual styling for control rendering using the latest Control Library (COMCTL)."-->
   <BoolProperty Name="EnableVisualStyles"
-                DisplayName="Visual styles"
-                Description="Enable the use the Windows XP styles for windows, controls, and so on."
+                DisplayName="Windows XP visual style"
+                Description="Enable the use the Windows XP styles for the application."
                 HelpUrl="https://docs.microsoft.com/dotnet/api/microsoft.visualbasic.applicationservices.windowsformsapplicationbase.enablevisualstyles"
                 Category="ApplicationFramework">
     <BoolProperty.Metadata>
@@ -412,8 +412,8 @@
 
   <!-- TODO: Shortlink for help URL -->
   <BoolProperty Name="SaveMySettingsOnExit"
-                DisplayName="Save My.Settings on exit"
-                Description="Allow the application to save the user settings on exit."
+                DisplayName="Save user settings on exit"
+                Description="Allow the application to save the user settings (My.Settings) on exit."
                 HelpUrl="https://docs.microsoft.com/dotnet/api/microsoft.visualbasic.applicationservices.windowsformsapplicationbase.savemysettingsonexit"
                 Category="ApplicationFramework">
     <BoolProperty.Metadata>
@@ -440,16 +440,22 @@
         </NameValuePair.Value>
       </NameValuePair>
     </EnumProperty.Metadata>
+    
+    <!-- DisplayName="100% scale factor (no DPI detection)" -->
     <EnumValue Name="DpiUnaware"
-               DisplayName="100% scale factor (no DPI detection)" />
+               DisplayName="DPI unaware" />
+    <!-- DisplayName="100% scale factor with improved GDI/GDI+ content quality (no DPI detection)" -->
     <EnumValue Name="DpiUnawareGdiScaled"
-               DisplayName="100% scale factor with improved GDI/GDI+ content quality (no DPI detection)" />
+               DisplayName="DPI unaware, GDI scaled" />
+    <!-- DisplayName="Primary monitor's scale factor is used for all monitors (DPI detected on-launch)" -->
     <EnumValue Name="SystemAware"
-               DisplayName="Primary monitor's scale factor is used for all monitors (DPI detected on-launch)" />
+               DisplayName="System aware" />
+    <!-- DisplayName="Dynamic scale factor (DPI detected on-launch and on-change)" -->
     <EnumValue Name="PerMonitor"
-               DisplayName="Dynamic scale factor (DPI detected on-launch and on-change)" />
+               DisplayName="Per monitor" />
+    <!-- DisplayName="Dynamic scale factor with improved comctl32 and dialog scaling (DPI detected on-launch and on-change including child windows)" -->
     <EnumValue Name="PerMonitorV2"
-               DisplayName="Dynamic scale factor with improved comctl32 and dialog scaling (DPI detected on-launch and on-change including child windows)" />
+               DisplayName="Per monitor V2" />
   </EnumProperty>
 
   <!-- TODO: Shortlink for help URL -->
@@ -467,12 +473,12 @@
       </NameValuePair>
     </EnumProperty.Metadata>
     <!-- TODO: Value provider logic needed to convert string enums to integer values -->
-    <!-- 0 -->
+    <!-- 0 - DisplayName="Windows authentication via My.User object" -->
     <EnumValue Name="Windows"
-               DisplayName="Windows authentication via My.User object" />
-    <!-- 1 -->
+               DisplayName="Windows" />
+    <!-- 1 - DisplayName="Custom authentication implemented within the application" -->
     <EnumValue Name="ApplicationDefined"
-               DisplayName="Custom authentication implemented within the application" />
+               DisplayName="Application-defined" />
   </EnumProperty>
 
   <!-- TODO: Shortlink for help URL -->
@@ -490,12 +496,12 @@
       </NameValuePair>
     </EnumProperty.Metadata>
     <!-- TODO: Value provider logic needed to convert string enums to integer values -->
-    <!-- 0 -->
+    <!-- 0 - DisplayName="Shut down when the main form closes" -->
     <EnumValue Name="AfterMainFormCloses"
-               DisplayName="Shut down when the main form closes" />
-    <!-- 1 -->
+               DisplayName="After main form closes" />
+    <!-- 1 - DisplayName="Shut down only after the last form closes" -->
     <EnumValue Name="AfterAllFormsClose"
-               DisplayName="Shut down only after the last form closes" />
+               DisplayName="After last form closes" />
   </EnumProperty>
 
   <!-- TODO: Shortlink for help URL -->
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
@@ -14,6 +14,9 @@
     <Category Name="Resources"
               DisplayName="Win32 Resources"
               Description="Resource settings for the application." />
+    <Category Name="ApplicationFramework"
+              DisplayName="Application Framework"
+              Description="Application Framework settings for the application." />
     <Category Name="Packaging"
               DisplayName="Packaging"
               Description="Packaging settings for the application." />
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -176,6 +179,24 @@
...
+                Description="Enable the Visual Basic Application Framework for this project."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/application-page-project-designer-visual-basic#enable-application-framework"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::OutputType" />
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>
+          (and
+            (has-evaluated-value "Application" "OutputType" "WinExe")
+            (not (has-csharp-lang-version-or-greater "1.0")))
```
MiYanni, line 194: So, this is fun. We have logic to check the version of the C# language used. So, to make this for 'VB', I made it show up for when the C# language version is **NOT** greater than 1.0. This is just for development purposes. We'll need to add a real flag, specifically for VB.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -355,6 +376,156 @@
...
+               DisplayName="After last form closes" />
+  </EnumProperty>
+
+  <!-- TODO: Shortlink for help URL -->
+  <DynamicEnumProperty Name="SplashScreen"
+                       DisplayName="Splash screen"
+                       Description="Sets the form to be used as a splash screen for the application."
+                       HelpUrl="https://docs.microsoft.com/dotnet/api/microsoft.visualbasic.applicationservices.windowsformsapplicationbase.splashscreen"
+                       Category="ApplicationFramework"
+                       EnumProvider="StartupObjectsEnumProvider">
```
MiYanni, line 513: This will need a custom provider to reflect on the forms available.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/381213 adds `has-vb-lang-version-or-greater` which should help here.

MiYanni: @drewnoakes Thank you! <3

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -176,6 +179,24 @@
     </DynamicEnumProperty.Metadata>
   </DynamicEnumProperty>
 
+  <!-- TODO: Shortlink for help URL -->
+  <BoolProperty Name="UseApplicationFramework"
```
melytc, line 183: I know wording it's confusing, but this property actually translates into `MyType` in the project file.

melytc (approved):
Looking good!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
MiYanni: I see that now. Thanks! Looks like:
- `true` = `<MyType>WindowsForms</MyType>`
- `false` = `<MyType>WindowsFormsWithCustomSubMain</MyType>`

melytc: Right! We probably will need to intercept that boolean value to convert it into the actual string.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
melytc: While talking with Drew, he had a good point: we should consider adding `has-vb-language` or similar in cases where we're only concerned with the language (not the version).

> So, this is fun. We have logic to check the version of the C# language used. So, to make this for 'VB', I made it show up for when the C# language version is **NOT** greater than 1.0. This is just for development purposes. We'll need to add a real flag, specifically for VB.

As a side note, the project properties UI is also enabled for F# so showing a property by `not (has-csharp-language)` would result in it being shown in VB and F#. More to the point in having a language-specific visibility condition. 😄

MiYanni: Or:
- `has-language "VB"`
- `has-language "CSharp"`
- `has-lang-version "VB" "16.0" "<"`
- `has-lang-version "CSharp" "7.0" ">"`
- `has-lang-version "CSharp" "6.0" "="`

Just make every variant (something which has multiple potential values) a parameter. This allows selection of language, version, and scope (<, >, =). You could add (<=, >=) if needed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
MiYanni: Looking into it, it seems that `MyType` is more involved. It feels similar to `Output Type`, but specifically relates to the VB `My` concept. It feels like we should actually have `MyType` as its own property entirely. It being changed related to this concept of 'application framework' is just an artifact of that concept, not the `MyType` property itself. https://docs.microsoft.com/en-us/dotnet/visual-basic/developing-apps/customizing-extending-my/customizing-which-objects-are-available-in-my#predefined-_mytype-values

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: I've added functions `is-csharp` and `is-vb` in that PR as well.

The distinction between `is-` and `has-` comes down to configurations. You can configure different language versions in different configurations (for example) but you can't realistically configure different languages in different configurations.

We also shied away from requiring the user to pass variables such as "VB" or "CSharp" to make it harder to pass the wrong thing. For example, the above would not work. The `Language` property has value `"C#"` not `"CSharp"`.

drewnoakes: Note also that `(not (has-csharp-lang-version-or-greater "1.0"))` would match an F# project.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
drewnoakes: @melytc is on the case with https://github.com/dotnet/project-system/pull/7936

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -176,6 +179,24 @@
...
+                DisplayName="Application Framework"
+                Description="Enable the Visual Basic Application Framework for this project."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/application-page-project-designer-visual-basic#enable-application-framework"
+                Category="General">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::OutputType" />
```
drewnoakes, line 189: I don't think this line is necessary.

```suggestion
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -176,6 +179,24 @@
     </DynamicEnumProperty.Metadata>
   </DynamicEnumProperty>
 
+  <!-- TODO: Shortlink for help URL -->
+  <BoolProperty Name="UseApplicationFramework"
+                DisplayName="Application Framework"
+                Description="Enable the Visual Basic Application Framework for this project."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/application-page-project-designer-visual-basic#enable-application-framework"
+                Category="General">
```
drewnoakes, line 187: I would stick this in the `ApplicationFramework` category directly, to be more consistent with how properties behave elsewhere.

As it stands, checking this box toggles a bunch of properties which are likely not visible on screen at that moment. Generally when you check a box that enables other properties, they appear immediately after that checkbox (e.g. generate documentation, ...).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -355,6 +376,156 @@
...
+  <BoolProperty Name="EnableVisualStyles"
+                DisplayName="Windows XP visual style"
+                Description="Enable the use the Windows XP styles for the application."
+                HelpUrl="https://docs.microsoft.com/dotnet/api/microsoft.visualbasic.applicationservices.windowsformsapplicationbase.enablevisualstyles"
+                Category="ApplicationFramework">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::UseApplicationFramework" />
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>
+          (has-evaluated-value "Application" "UseApplicationFramework" true)
```
drewnoakes, line 390: Please test the following:

1. Set output type to WinExe
2. Check "Use application framework"
3. Verify this property is visible
4. Change from WinExe to Library
5. Verify this property is hidden

I suspect this property may remain visible in the last of the above steps. I don't think we set `UseApplicationFramework` to `false` when making the `OutputType` something other than `WinExe`.

It would be possible to do that with interception, but then there'd be a delay between changing that drop-down and the properties being hidden.

Instead it might be better to `and` this (and other VB properties) with `(has-evaluated-value "Application" "OutputType" "WinExe")`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
```diff
@@ -355,6 +376,156 @@
...
 
+  <!-- TODO: Shortlink for help URL -->
+  <!--Description="Enable modern visual styling for control rendering using the latest Control Library (COMCTL)."-->
+  <BoolProperty Name="EnableVisualStyles"
+                DisplayName="Windows XP visual style"
+                Description="Enable the use the Windows XP styles for the application."
+                HelpUrl="https://docs.microsoft.com/dotnet/api/microsoft.visualbasic.applicationservices.windowsformsapplicationbase.enablevisualstyles"
+                Category="ApplicationFramework">
+    <BoolProperty.Metadata>
+      <NameValuePair Name="DependsOn" Value="Application::UseApplicationFramework" />
```
drewnoakes, line 387: Pretty sure you can remove all these `DependsOn` metadata.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
melytc: Splash Screen wasn't included in @KlausLoeffelmann spec, do we want to bring that in, too? I suspect it wouldn't be using the same EnumProvider as StartupObjects.

MiYanni: Yes, we'd want splash screen. Based on what Klaus said at the start of the meeting (only having two days to write this up), I assumed he wouldn't catch everything. I'd need to do a similar investigation for the other pages (other than Application) to make sure we have property parity. **Additionally**, there are actually more properties that are related to this stuff than we expose, but I have to assume the target for now is parity and not additions.

MiYanni:
These changes ended up getting handled during the rest of the VB property pages stubbing work that is currently on-going.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.xaml
KlausLoeffelmann: Good catch!
I absolutely missed that - yes, we need the SplashScreen.
I take a look tomorrow to see if there are any important things to look out for!

PR #7929, opened 2022-02-21
Improve solution close performance

drewnoakes:
Previously we would block the main thread while waiting for serialization of the fast up-to-date check's persistent state to occur.

This change has two parts:

1. Return immediately from the `OnAfterSolutionClose` callback.
2. Schedule the actual serialization on the thread pool.

To guarantee that the async serialization completes before the VS process exits, we register the task with the package's JTF. Tasks registered with that factory must all complete before the package can be unloaded and VS can shut down.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7929)


drewnoakes:
Seeing [@lifengl's comment](https://github.com/dotnet/project-system/pull/7923#discussion_r809416906), and disussing with @davkean, I'm going to go a different route here.

Dave pointed out one possible issue with this approach. During branch switches where the .sln file changes, if we don't join on the writing of state as part of solution close, the reloading projects might not see the persisted state yet, or might hit file access exceptions while restoring state if writing is still happening.

I'll open a new PR.

PR #7936, opened 2022-02-23
Add StartupObject property experience for VB

melytc:
The `StartupObject` property has a different behavior compared to the C# experience. 
- This setting doesn't get serialized exclusively to the project file; in certain scenarios, it also needs to be serialized into the myapp file.

**Scenarios**
- If the `ApplicationForm` is disabled, this settings gets serialized into the project file.

We have been showing a correct set of values in the dropdown. The expected startup objects should be:
For C#, this is being done by calling FindEntryPoints (see  https://github.com/dotnet/roslyn/issues/59696).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7936)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
@@ -42,7 +42,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
                 && bool.TryParse(pair.Value, out bool optionValue)
                 && optionValue) ?? false;
 
-            return Task.FromResult<IDynamicEnumValuesGenerator>(new StartupObjectsEnumGenerator(_workspace, _unconfiguredProject, includeEmptyValue));
+            bool searchForEntryPointsInFormsOnly = false;
+
+            if (_unconfiguredProject.Capabilities.Contains(ProjectCapability.VisualBasic))
+            {
+                searchForEntryPointsInFormsOnly = true;
+                includeEmptyValue = false;
+            }
+
+            return Task.FromResult<IDynamicEnumValuesGenerator>(new StartupObjectsEnumGenerator(_workspace, _unconfiguredProject, includeEmptyValue, searchForEntryPointsInFormsOnly));
         }
     }
 
@@ -52,18 +60,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
         private readonly Workspace _workspace;
         private readonly UnconfiguredProject _unconfiguredProject;
         private readonly bool _includeEmptyValue;
-
-        /// <summary>
-        /// When we implement WinForms support, we need to set this for VB WinForms projects
-        /// </summary>
-        private static bool SearchForEntryPointsInFormsOnly => false;
+        private readonly bool _searchForEntryPointsInFormsOnly;
 
         [ImportingConstructor]
-        public StartupObjectsEnumGenerator(Workspace workspace, UnconfiguredProject project, bool includeEmptyValue)
+        public StartupObjectsEnumGenerator(Workspace workspace, UnconfiguredProject project, bool includeEmptyValue, bool searchForEntryPointsInFormsOnly)
         {
             _workspace = workspace;
             _unconfiguredProject = project;
             _includeEmptyValue = includeEmptyValue;
+            _searchForEntryPointsInFormsOnly = searchForEntryPointsInFormsOnly;
+
         }
 
         public async Task<ICollection<IEnumValue>> GetListedValuesAsync()
@@ -84,7 +90,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
             }
 
             IEntryPointFinderService? entryPointFinderService = project.LanguageServices.GetService<IEntryPointFinderService>();
-            IEnumerable<INamedTypeSymbol>? entryPoints = entryPointFinderService?.FindEntryPoints(compilation?.GlobalNamespace, SearchForEntryPointsInFormsOnly);
+            IEnumerable<INamedTypeSymbol>? entryPoints = entryPointFinderService?.FindEntryPoints(compilation?.GlobalNamespace, _searchForEntryPointsInFormsOnly);
             if (entryPoints is not null)
             {
                 enumValues.AddRange(entryPoints.Select(ep =>


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
```diff
@@ -42,7 +42,15 @@ public Task<IDynamicEnumValuesGenerator> GetProviderAsync(IList<NameValuePair>?
                 && bool.TryParse(pair.Value, out bool optionValue)
                 && optionValue) ?? false;
 
-            return Task.FromResult<IDynamicEnumValuesGenerator>(new StartupObjectsEnumGenerator(_workspace, _unconfiguredProject, includeEmptyValue));
+            bool searchForEntryPointsInFormsOnly = false;
+
+            if (_unconfiguredProject.Capabilities.Contains(ProjectCapability.VisualBasic))
+            {
+                searchForEntryPointsInFormsOnly = true;
+                includeEmptyValue = false;
```
drewnoakes, line 50: Super minor: this `if` branch overwrites any value set above, so it is possible to move that work into an `else` branch here and not do that work for VB projects.

Also, why does this vary by language? It'd be nice to call that out in a comment, as the rationale for the behaviour here isn't self-evident.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
```diff
@@ -52,18 +60,16 @@ internal class StartupObjectsEnumGenerator : IDynamicEnumValuesGenerator
...
 
         [ImportingConstructor]
-        public StartupObjectsEnumGenerator(Workspace workspace, UnconfiguredProject project, bool includeEmptyValue)
+        public StartupObjectsEnumGenerator(Workspace workspace, UnconfiguredProject project, bool includeEmptyValue, bool searchForEntryPointsInFormsOnly)
         {
             _workspace = workspace;
             _unconfiguredProject = project;
             _includeEmptyValue = includeEmptyValue;
+            _searchForEntryPointsInFormsOnly = searchForEntryPointsInFormsOnly;
+
```
drewnoakes, line 72: nit: blank line

```suggestion
```

drewnoakes (approved):
LGTM

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
```diff
@@ -84,7 +90,7 @@ public async Task<ICollection<IEnumValue>> GetListedValuesAsync()
             }
 
             IEntryPointFinderService? entryPointFinderService = project.LanguageServices.GetService<IEntryPointFinderService>();
-            IEnumerable<INamedTypeSymbol>? entryPoints = entryPointFinderService?.FindEntryPoints(compilation?.GlobalNamespace, SearchForEntryPointsInFormsOnly);
+            IEnumerable<INamedTypeSymbol>? entryPoints = entryPointFinderService?.FindEntryPoints(compilation?.GlobalNamespace, _searchForEntryPointsInFormsOnly);
```
tmeschter, line 93: Do the old property pages use the `IEntryPointFinderService`? If so, do they ever pass "true" for the second parameter?

If not, I suggest some manual testing to make sure everything is working as expected when `_searchForEntryPointsInFormsOnly` is "true".

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
tmeschter: If we're going to check project capabilities and set these values in code, a more idiomatic approach would be to split this type in two: one implementation for C# (with `[AppliesTo(ProjectCapability.CSharp)]`) and one for VB (with `[AppliesTo(ProjectCapability.VisualBasic)]`). They would both be tagged with `[ExportDynamicEnumValuesProvider("StartupObjectsEnumProvider")]`.

I think an even better alternative, though, is to let this be set by an option in the `Rule`, like how we handle `includeEmptyValue`. This gives us more control over the behavior--we can have some default options in the base ApplicationPropertyPage.xaml and then extend/override those through project-type-specific .xaml files.

For example, the behavior with this PR will be to only ever show WinForms entry points in VB projects--even for VB console and WPF apps. We could instead have a "VB WinForms"-specific Application property page that extends the default one and changes the search behavior for only those.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
melytc: I was testing this yesterday, and we used a different method: Interop.IVBEntryPointProvider in [ApplicationPropPageVBWinForms.vb](https://github.com/dotnet/project-system/blob/9b25af877a11bb4011f38b21c067e2e0475c79c4/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageVBWinForms.vb#L708). Since we already use the `IEntryPointFinderService` for the new properties pages (at least C# with that flag turned off), I tested what happened when I called it with a true flag; I found out that it does have a bug: https://github.com/dotnet/roslyn/issues/59696.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
melytc: > If we're going to check project capabilities and set these values in code, a more idiomatic approach would be to split this type in two: one implementation for C# (with [AppliesTo(ProjectCapability.CSharp)]) and one for VB (with [AppliesTo(ProjectCapability.VisualBasic)]). They would both be tagged with [ExportDynamicEnumValuesProvider("StartupObjectsEnumProvider")].

Totally, these changes were intended to quickly test the EntryPointFinderService 😃

melytc: > I think an even better alternative, though, is to let this be set by an option in the Rule, like how we handle includeEmptyValue. This gives us more control over the behavior--we can have some default options in the base ApplicationPropertyPage.xaml and then extend/override those through project-type-specific .xaml files.

That's a good point I haven't considered.

PR #7938, opened 2022-02-23
Remove restore notification to Cps - being replaced by events from NuGet

adrianvmsft:
Removing the nuget restore notification to CPS. This functionality is being replaced by the new events added by NuGet that CPS listens to.

The CPS change is tracked by [CPS PR 381604](https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/381604)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7938)


tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/PackageRestore/PackageRestoreDataSourceTests.cs
```diff
@@ -94,7 +94,6 @@ private static PackageRestoreDataSource CreateInstance(UnconfiguredProject? proj
             solutionRestoreService ??= IVsSolutionRestoreServiceFactory.Create();
             IProjectDiagnosticOutputService logger = IProjectDiagnosticOutputServiceFactory.Create();
             IFileSystem fileSystem = IFileSystemFactory.Create();
-            var projectDependentFileChangeNotificationService = IProjectDependentFileChangeNotificationServiceFactory.Create();
```
drewnoakes, line 97: Does this leave any other usages of `IProjectDependentFileChangeNotificationServiceFactory`? If not, please remove it.

tmeschter:
The only change here is that we no longer tell CPS when the project.assets.json file has changed, correct?

adrianvmsft:
> The only change here is that we no longer tell CPS when the project.assets.json file has changed, correct?

Yes, that is correct. CPS is subscribing to NuGet restore events - which will use instead

MiYanni (approved):

drewnoakes (approved):

PR #7950, opened 2022-03-03
Add windows platform to TargetFramework value for .NET 7 when needed.

melytc:
Fixes [AB#1473348](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1473348).

The platform value is needed to be added to the `TargetFramework` value for WPF and Windows Forms projects to be able to build properly. We have been adding this value for .NET 5 and 6; this change extends this behavior for .NET 7.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7950)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
@@ -208,7 +208,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
 
         private static bool IsNetCore5OrHigher(string targetFrameworkAlias)
         {
-            return targetFrameworkAlias.Contains("5.0") || targetFrameworkAlias.Contains("6.0");
+            return targetFrameworkAlias.Contains("5.0") || targetFrameworkAlias.Contains("6.0") || targetFrameworkAlias.Contains("7.0");
         }
 
         private async Task<bool> IsWindowsPlatformNeededAsync()


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
```diff
@@ -208,7 +208,7 @@ private async Task<string> ComputeValueAsync(ComplexTargetFramework complexTarge
 
         private static bool IsNetCore5OrHigher(string targetFrameworkAlias)
         {
-            return targetFrameworkAlias.Contains("5.0") || targetFrameworkAlias.Contains("6.0");
+            return targetFrameworkAlias.Contains("5.0") || targetFrameworkAlias.Contains("6.0") || targetFrameworkAlias.Contains("7.0");
```
drewnoakes, line 211: This problem is going to surface again with .NET 8.

We have access to the TargetFrameworkIdentifier and TargetFrameworkVersion here. Can we use those instead?

For example,

```c#
return TargetFrameworkIdentifier == ".NETCoreApp" && ParseTargetFrameworkVersion(TargetFrameworkVersion) > 6.0m;
```

The comparisons should use the correct `StringComparer`. `TargetFrameworkVersion` has form `v6.0`, so `ParseTargetFrameworkVersion` would need to remove the leading `v` and parse the rest using `decimal.TryParse`. If it's possible that the version has form `0.0.0` then we should use `System.Version` instead.

drewnoakes (approved):
We need a proper fix here. I'm approving this in case you don't have bandwidth to fix this and need it in preview 2. If you do merge it, please either file a follow up PR or create an issue.

I think adding a more robust fix here should be straightforward though.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/CompoundTargetFrameworkValueProvider.cs
melytc: Ideally, yes. There's a bigger problem around that is the property querying is for the previous TargetFramework, not the one that is actually being changed to. But I agree, I think it's a good time to rethink how we're handling it. I'll try to come up with something today.

MiYanni (approved):

melytc:
I'm going to merge this, while I work on an overall improvement in this class here: https://github.com/dotnet/project-system/issues/7958

Issue #7888, opened 2022-02-04
Add SBOM plugin for our CI pipeline


MiYanni:
SBOM = Software Bill of Materials. This will collect a list of files we produce that are inserted into VS.

This is required to be done (by 2/11) for any components that insert into VS. Documentation here: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/26284/Enabling-SBOM-For-Your-Component-Insertion-into-VS

Edit: The due date has changed to 2/25.

PR #7955, opened 2022-03-05
Add SBOM to our build pipeline

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7888

### What is it?
SBOM is Software Bill of Materials. It is intended to be a way to show what 'materials' we rely on to make our software. In this case, the actual SBOM contains our *dependency tree* but as a flat list.

### Why do we need this now?
There was a U.S. presidental order to 'strengthen our software supply chain'. Creating an SBOM allows visibility for what it takes to create our software and help prevent bad actors within the chain. An SBOM is a requirement for us to do Visual Studio insertions (the insertion PR check will be added soon).

### What do we need to provide?
Our build needs to produce the SBOM in the [SPDX (Software Package Data Exchange)](https://spdx.dev/) format. This is stored as a JSON file containing the list of packages we use and the relationships of those packages, along with bits of other metadata. From my testing, our file is only ~3400 lines long, but some groups within Microsoft are having issues because the file they produce is too large. We're a small codebase so we shouldn't expect this to be an issue. The file produced is named `manifest.spdx.json`.

### How is the SPDX file created?
Fortunately, MicroBuild created a tool to run for this purpose. The tool is implemented as a DevOps pipeline task, either named [ADO SBOM Generator](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/ado-sbom-generator) or `ManifestGeneratorTask` in the YAML file. This file is ran on our assemblies we produce, which include:
- `Microsoft.VisualStudio.AppDesigner.dll`
- `Microsoft.VisualStudio.Editors.dll`
- `Microsoft.VisualStudio.ProjectSystem.Managed.dll`
- `Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll`
- And all `*.resources.dll` (localized) versions of the above DLLs

### What do we do with the SPDX file?
After this file is created, it needs to be part of our Visual Studio insertion process. || *Until this point, I was vaguely aware of how this works. I still don't have all the details, but I was able to gain a small bit of clarity.* || As it stands, we rely on the [RoslynInsertionTool](https://github.com/dotnet/roslyn-tools/tree/main/src/RoslynInsertionTool) to perform this work. That tool relies on the [MicroBuild SWiX (Simplified Windows Installer XML) plugin](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/21892/Setup-authoring) which creates our [VSIX (Visual Studio Integration Extension)](https://docs.microsoft.com/en-us/visualstudio/extensibility/anatomy-of-a-vsix-package) package for the insertion PR. The SWiX plugin takes in a JSON manifest file and produces a VSMAN file. The plugin will also publish the packages listed in the JSON manifest to a VS drop location and include those links in the VSMAN file. The only requirement for us is to *include the SPDX file using the name `*_sbom.json`, where `*` is the name of the VSIX* within the JSON manifest.

Here is an example of the `packages` section of the JSON manifest:

image

Here is an example of the `packages` section of the VSMAN file produced by the SWiX plugin based on the JSON manifest above:

image

### How was this accomplished?
Our build currently does *all actions* as a **single** MSBuild pipeline. This would not allow us to run the SBOM Generator task during our build, as this task only runs as an independent tool within the DevOps pipeline and does not trigger off of MSBuild data; for example, how MicroBuild signing works. Instead, you provide the path to the folder where your assemblies are placed and it spits out the SPDX file. This means that you are required to have the DLLs **before** to the SBOM Generator task and be able to inject the SPDX into the manfiest JSON **after** the task completes. || *Based on this requirement alone, my first attempts involved simply splitting the build into multiple parts: Build, Package, and Test. This proved to be difficult as many parts of our build relies on the prior actions taking place (such as the right MSBuild properties to be set or MSBuild target to be ran). Only after learning the actual end-to-end process from discussing with the documentation dev did I change direction and make the solution as simple as possible.* || To do this, I've created a 'quick source build' (QuickSrcBuild). This takes the projects located in the `src` folder and **only** builds them. No packaging, no signing, no testing, etc. This satisfies the minimum requirements necessary to run the SBOM generator task.

After the SPDX is generated, we needed a way to include it as part of the JSON manifest for the RoslynInsertionTool to create our VSMAN files. The RoslynInsertionTool generates our JSON manifest, so I needed to inject a process to modify the JSON manifest immediately after it is created. The RoslynInsertionTool generates both the VSIX and the JSON manifest at the same time. Fortunately, I already determined when this occurs in the `AddVsixForSigning` target created in the OptProf changes (via the `AddVsixForSigningDependsOn` property). This new process simply runs immediately after that target occurs. The new target, `UpdateManifestJsonForSbom`, runs the PowerShell script `UpdateManifestJsonForSbom.ps1` which will:
- Read the name of the VSIX package from the JSON manifest for the particular assembly
- Update the JSON manifest for the particular assembly to include the filename and size of the SPDX
- Copy the SPDX to the same location as the JSON manifest and rename it to `{VSIX package name}_sbom.json`

With the JSON manifest updated to include the SPDX, the RoslynInsertionTool properly produces the VSMAN file. The same SPDX is used for all packages, which I'm told is fine to do. It is not necessary to make separate SPDX per package that only contains SBOM information for the assemblies in that package. Over-reporting is better than under-reporting. 😉

### Was this the recommended process for SBOM?
This was **not** the recommended process. The reason the recommended process could not be used is:
- Our build and package processes being within the same MSBuild pipeline
- Our reliance on the RoslynInsertionTool to run the SWiX plugin

These two factors contributed to the primary issues with implementing SBOM. For the [recommended process](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/26284/Enabling-SBOM-For-Your-Component-Insertion-into-VS), we would rely on an updated version of the SWiX plugin to read metadata in 'our' SWIXPROJ and VSMANPROJ files. Unfortunately, those files are not in our repo and are part of the RoslynInsertionTool. Originally, I had a script to update these project files but that effort was useless without being able to update the version of the SWiX plugin being used. To detatch from the RoslynInsertionTool process, we would need to make our own SWIXPROJ and VSMANPROJ files and have the SWiX plugin run on these appropriately. As for splitting the build and packaging processes, we would need to reconfigure our build to be able to do these processes independently. I timeboxed the effort I put into that and decided it can be done at a later date.

### How much time did this add to our build?
It added a little over 3 minutes. The SBOM generation takes between 1.5 to 2 minutes and the QuickSrcBuild takes about 1.5 minutes. There is an opportunity for improvement after we can split the build process into multiple MSBuild phases.

image

### How do the VSMAN files relate to the VSIX packages?
There are 3 VSIX packages we produce:
- `Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.vsix`
- `ProjectSystem.vsix`
- `VisualStudioEditorsSetup.vsix`

There are 2 VSMAN files we produce:
- `Microsoft.VisualStudio.Editors.vsman`
- `Microsoft.VisualStudio.ProjectSystem.Managed.vsman`

The `Microsoft.VisualStudio.Editors.vsman` file contains the `VisualStudioEditorsSetup.vsix` package. The `Microsoft.VisualStudio.ProjectSystem.Managed.vsman` file contains both the `Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.vsix` and `ProjectSystem.vsix` packages.

#### Successful build
- https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5848220
- https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5861709
- https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5866657

#### In-progress
- ~~SBOM seems to be correctly inserted into .vsman but contains the wrong information in the SBOM itself~~
  - This was an incorrect statement. The SBOM correctly contains the dependency tree of our DLLs.
- ~~Signing is still broken since splitting the build~~
  - Fixed signing by making pre-SBOM build not do signing and combining packaging and testing together, doing all signing.
- ~~Test results are not found while publishing~~
  - Was missing Test property for build. Fixed.
- ~~Do a validation insertion and make sure it is working properly~~
  - ~~Awaiting results: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/384512~~
  - ~~Build still failing: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/385044~~
    - ~~https://cloudbuild.microsoft.com/build?id=7e0fbc28-f016-9e05-0898-4bde0b60ca39~~
  - ~~Awaiting insertion test results: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/385256~~
  - Successful insertion! https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/385256

#### Note
For this PR, I will use `Squash and merge` so that we don't add excessive commits to the repo.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7955)


setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
```diff
@@ -8,7 +8,8 @@
     <AddVsixForSigningDependsOn>SwixBuild</AddVsixForSigningDependsOn>
 
     <!-- VS Insertion -->
-    <VisualStudioInsertionComponent>Microsoft.VisualStudio.ProjectSystem.Managed</VisualStudioInsertionComponent>
+    <!-- NOTE: Having VisualStudioInsertionComponent defined is the condition for VisualStudio.SetupPackage.vsmanproj to be built (via FinalizeInsertionVsixFiles target in VisualStudio.Insertion.proj) in RepoToolset. -->
+    <VisualStudioInsertionComponent>Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles</VisualStudioInsertionComponent>
```
MiYanni, line 12: I believe this to be a copy/paste bug from 2017. AFAIK, this should be the name of the insertion metadata/vsman we produce.
image

build/proj/CoreBuild.proj
```diff
@@ -29,15 +29,38 @@
...
+    <CommonBuildProperty Include="TestProjects=$(TestProjects)" />
   </ItemGroup>
 
   <ItemGroup>
     <SolutionRestoreAndBuildProperty Include="@(CommonBuildProperty)"/>
     <SolutionRestoreAndBuildProperty Include="__DeployProjectOutput=$(Deploy)" Condition="'$(Deploy)' != ''" />
   </ItemGroup>
 
-  <Target Name="LogInitialMessage">
-    <Message Text="Building $(SolutionFileName) [$(Configuration)]...%0a" Importance="high" />
```
MiYanni, line 40: I removed this target as it only logged a message. I added a message at the top of each target in this file instead.

build/optprof/UpdateRunSettings.ps1
```diff
@@ -35,5 +37,5 @@ $testStores.AppendChild($buildDropStore)
 
 $runsettingsXml.Save($runsettingsPath)
 Write-Host 'Saved Output:'
-Write-Host "profilingInputsStore: $profilingInputsStore"
-Write-Host "buildDropStore: $buildDropStore"
\ No newline at end of file
+Write-Host "profilingInputsStore: $($profilingInputsStore.Uri)"
+Write-Host "buildDropStore: $($buildDropStore.Uri)"
```
MiYanni, line 41: I had flubbed this output message from my OptProf PR. Fixing it here. No impact on functionality.

build/optprof/UpdateRunSettings.ps1
```diff
@@ -14,11 +14,13 @@ $runsettingsXml = [Xml.XmlDocument](Get-Content $runsettingsPath)
 # https://stackoverflow.com/questions/33813700/empty-xml-node-rendered-as-string-in-powershell
 $testStores = $runsettingsXml.RunSettings.TestConfiguration.SelectSingleNode('TestStores')
 
-$profilingInputsStore = $runsettingsXml.CreateElement('TestStore')
+# https://stackoverflow.com/a/59090765/294804
+$profilingInputsStore = $runsettingsXml.CreateElement('TestStore', $runsettingsXml.DocumentElement.NamespaceURI)
```
MiYanni, line 18: This simply removes the `xmlns=""` from the element produced.

build/proj/CoreBuild.proj
```diff
@@ -48,16 +64,18 @@
...
 
   <Target Name="Build">
+    <Message Text="Building $(SolutionFileName) [$(Configuration)]...%0a" Importance="high" />
     <ItemGroup>
       <SolutionBuildTarget Include="Rebuild" Condition="'$(Rebuild)' == 'true'" />
       <SolutionBuildTarget Include="Build" Condition="'$(Build)' == 'true' and '$(Rebuild)' != 'true'" />
       <SolutionBuildTarget Include="Test" Condition="'$(Test)' == 'true'" />
       <SolutionBuildTarget Include="IntegrationTest" Condition="'$(IntegrationTest)' == 'true'" />
-      <SolutionBuildTarget Include="Pack" Condition="'$(BuildOrRebuild)' == 'true'" />
+      <SolutionBuildTarget Include="Pack" Condition="'$(BuildOrRebuild)' == 'true' AND '$(QuickSrcBuild)' != 'true'" />
```
MiYanni, line 78: Strangely enough, we don't use the `Pack` property to actually indicate if the Pack target is used. The quick Src build needs to not Pack.

setup/Directory.Build.targets
```diff
@@ -22,6 +22,11 @@
     </ItemGroup>
   </Target>
 
+  <!-- Updates insertion manifests (json) for creating the vsman files for insertion to include the SPDX SBOM json. -->
+  <Target Name="UpdateManifestJsonForSbom" AfterTargets="AddVsixForSigning">
+    <Exec ContinueOnError="false" Command="powershell -NonInteractive -NoLogo -NoProfile -ExecutionPolicy Unrestricted -Command &quot;. $(RepoRoot)build\script\UpdateManifestJsonForSbom.ps1 -manifestJsonPath '$(VisualStudioSetupInsertionPath)$(VisualStudioInsertionComponent).json' -sbomMetadataPath '$(ArtifactsBinDir)spdx_2.2\manifest.spdx.json'&quot;" />
+  </Target>
```
MiYanni, line 28: Technically, this should be `AfterTargets="$(AddVsixForSigningDependsOn)"` as that value (target) indicates when VSIX packaging has completed. Doing it as `AfterTargets="AddVsixForSigning"` unnecessarily binds the SBOM process to signing. But there is not currently a scenario where we wouldn't sign and do SBOM.

src/Directory.Build.targets
```diff
@@ -10,12 +10,12 @@
     It assumes that the project directory contains these files (PublicApiAnalyzers also makes this assumption).
     Uses AppendAllText instead of Add-Content since that cmdlet adds an empty line at the end of the file.
   -->
-  <Target Name="FlushUnshippedAPI" AfterTargets="Build">
-    <Exec Command="powershell.exe -NonInteractive -NoLogo -NoProfile -Command &quot;$unshipped = '$(MSBuildProjectDirectory)\PublicAPI.Unshipped.txt'; $content = Get-Content $unshipped -Raw; if(-Not [String]::IsNullOrWhiteSpace($content)) { [IO.File]::AppendAllText('$(MSBuildProjectDirectory)\PublicAPI.Shipped.txt', [Environment]::NewLine + $content); Clear-Content $unshipped }&quot;" />
+  <Target Name="FlushUnshippedAPI" AfterTargets="Build" Condition="'$(CIBuild)' != 'true'">
+    <Exec Command="powershell -NonInteractive -NoLogo -NoProfile -ExecutionPolicy Unrestricted -Command &quot;$unshipped = '$(MSBuildProjectDirectory)\PublicAPI.Unshipped.txt'; $content = Get-Content $unshipped -Raw; if(-Not [String]::IsNullOrWhiteSpace($content)) { [IO.File]::AppendAllText('$(MSBuildProjectDirectory)\PublicAPI.Shipped.txt', [Environment]::NewLine + $content); Clear-Content $unshipped }&quot;" />
```
MiYanni, line 14: This target doesn't need to run in CI.

src/Directory.Build.targets
```diff
@@ -10,12 +10,12 @@
...
   </Target>
 
   <!-- Adds the appropriate assemblies for signing. -->
-  <Target Name="AddAssembliesForSigning" BeforeTargets="SignFiles">
+  <Target Name="AddAssembliesForSigning" BeforeTargets="SignFiles" Condition="'$(QuickSrcBuild)' != 'true'">
```
MiYanni, line 18: We don't do signing for the quick Src build.

build/optprof/UpdateRunSettings.ps1
drewnoakes: Perhaps add that info in a comment in the script directly.

MiYanni: I figured the link comment was more informative.
image

build/ci/official.yml
```diff
@@ -69,8 +71,27 @@ steps:
...
+    BuildDropPath: '$(Build.SourcesDirectory)\artifacts\$(BuildConfiguration)\bin\Dlls'
+    ManifestDirPath: '$(Build.SourcesDirectory)\artifacts\$(BuildConfiguration)\bin'
+
+- task: MSBuild@1
   displayName: Build ProjectSystem.sln
+  inputs:
+    solution: $(Build.SourcesDirectory)\build\proj\Build.proj
+    msbuildArchitecture: 'x86'
+    configuration: $(BuildConfiguration)
+    msbuildArguments: '/m /warnaserror /nologo /clp:Summary /p:Build=true /p:Test=true /p:CIBuild=true /v:normal /bl:$(Build.SourcesDirectory)\artifacts\$(BuildConfiguration)\log\Build.binlog'
```
tmeschter, line 94: Will this cause (non-integration) tests to be run, like build.cmd used to?

build/optprof/UpdateRunSettings.ps1
```diff
@@ -14,11 +14,13 @@ $runsettingsXml = [Xml.XmlDocument](Get-Content $runsettingsPath)
...
 $profilingInputsStore.SetAttribute('Uri', "vstsdrop:$profilingInputsPath")
-$testStores.AppendChild($profilingInputsStore)
+$null = $testStores.AppendChild($profilingInputsStore)
```
tmeschter, line 20: Is this just to emphasize that we're ignoring the result of `AppendChild`?

tmeschter: Oh, this prevents the object returned from `AppendChild` from being output to the log?

build/proj/CoreBuild.proj
```diff
@@ -29,15 +29,31 @@
...
+  <Target Name="GenerateSolutionForQuickSrcBuild" BeforeTargets="Restore" Condition="'$(QuickSrcBuild)' == 'true'">
+    <PropertyGroup>
+      <SolutionName>Src</SolutionName>
+      <SolutionFolder>$(RepoRoot)src/</SolutionFolder>
+      <SolutionFileName>$(SolutionName).sln</SolutionFileName>
+      <SolutionPath>$(SolutionFolder)$(SolutionFileName)</SolutionPath>
+    </PropertyGroup>
+    <Message Text="Generating Solution file [$(SolutionFileName)]..." Importance="high" />
+
+    <Exec Command="dotnet --version" />
```
tmeschter, line 49: Is this needed, or a leftover from the development process?

build/proj/CoreBuild.proj
```diff
@@ -29,15 +29,31 @@
...
+    <Exec Command="dotnet new sln -n $(SolutionName) -o $(SolutionFolder) --force" />
+
+    <ItemGroup>
+      <ProjFiles Include="$(SolutionFolder)**/*.*proj" />
+    </ItemGroup>
+    <!-- https://github.com/dotnet/cli/issues/6295#issuecomment-346973582 -->
+    <Exec Command="dotnet sln $(SolutionPath) add &quot;%(ProjFiles.FullPath)&quot;" />
```
tmeschter, line 56: So here we're add all the project files (under src) to Src.sln, but where do we actually build Src.sln?

tmeschter (commented):
Please update the PR message to explain what you're doing at a high level, and more importantly _why_ you're doing it. Someone is going to look at this in a year and go "what the heck is SBOM?"--and having an explanatory message in source control will help them a ton.

build/ci/official.yml
MiYanni: Yes. `/p:Test=true` causes tests to be run.

build/optprof/UpdateRunSettings.ps1
MiYanni: Correct (to the second question, not the first). It prevents putting the object returned into the output stream. One of the many things I learned about PowerShell from a previous team. 😝

build/proj/CoreBuild.proj
MiYanni: It isn't needed, no. In previous projects, I've had this for simply logging purposes when I first use the dotnet CLI. It was more applicable for hosted DevOps images as they update and change versions of things. I can remove it here.

build/proj/CoreBuild.proj
MiYanni: The build is still done in the `Build` target. It runs `Src.sln` because we've changed the `SolutionFileName` and `SolutionPath` properties. You can see they default to `ProjectSystem.sln` at the top of the file. So, I'm just replacing them. No changes to the build itself were necessary.

build/proj/CoreBuild.proj
MiYanni: Removed.

build/ci/official.yml
drewnoakes: So long as it doesn't run our integration tests. They are not likely to work in CI (though if they do work, that's great).

build/proj/CoreBuild.proj
```diff
@@ -29,15 +29,30 @@
...
+  <Target Name="GenerateSolutionForQuickSrcBuild" BeforeTargets="Restore" Condition="'$(QuickSrcBuild)' == 'true'">
+    <PropertyGroup>
+      <SolutionName>Src</SolutionName>
+      <SolutionFolder>$(RepoRoot)src/</SolutionFolder>
+      <SolutionFileName>$(SolutionName).sln</SolutionFileName>
+      <SolutionPath>$(SolutionFolder)$(SolutionFileName)</SolutionPath>
+    </PropertyGroup>
+    <Message Text="Generating Solution file [$(SolutionFileName)]..." Importance="high" />
+
+    <Exec Command="dotnet new sln -n $(SolutionName) -o $(SolutionFolder) --force" />
```
drewnoakes, line 49: Why `--force`? Might be worth a comment.

setup/Directory.Build.targets
drewnoakes: Your description makes sense. It'd be good as a comment in the file.

drewnoakes (approved):

MiYanni:
This PR is still blocked as there are failures in insertions for it that need investigated.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.csproj
MiYanni: This was an incorrect assumption. It seems we produce 3 VSIX files and 2 VSMAN files, where the `Managed.vsman` contains both the `CommonFiles` and `ProjectSystem` VSIX packages. From a build prior to these changes, you can see the VSMAN files underlined in red and the VSIX packages underlined in yellow.

Screenshot 2022-03-10 172643

PR #7960, opened 2022-03-08
Strip out spaces in VB settings class/member .settings properties

adamint:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1289213/ by replacing spaces in the namespace/class name of the VB settings file with underscores.

Test as follows:
- Create new WPF VB project
- Create new folder (My Project)
- Create new settings file (Settings.settings)
- Click show all files 
- Observe no errors in the generated Settings VB file

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7960)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsSingleFileGeneratorBase.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsSingleFileGeneratorBase.vb
@@ -8,6 +8,7 @@ Imports System.Reflection
 Imports System.Runtime.InteropServices
 
 Imports Microsoft.VisualStudio.Designer.Interfaces
+Imports Microsoft.VisualStudio.Editors.DesignerFramework
 Imports Microsoft.VisualStudio.Editors.Interop
 Imports Microsoft.VisualStudio.OLE.Interop
 Imports Microsoft.VisualStudio.Shell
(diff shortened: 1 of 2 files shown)


adamint:
I have removed the class name escaping, as it leads to a mismatch between the name of the file and the name of the class.

src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsSingleFileGeneratorBase.vb
```diff
@@ -171,7 +170,7 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
                     ' If this is the "default" settings file, we add the "My" module as well...
                     '
                     If shouldGenerateMyStuff Then
-                        AddMyModule(CompileUnit, projectRootNamespace, wszDefaultNamespace)
+                        AddMyModule(CompileUnit, projectRootNamespace, wszDefaultNamespace.Replace(" ", "_"))
```
tmeschter, line 173: I believe there are other places in Microsoft.VisualStudio.Editors and/or Microsoft.VisualStudio.AppDesigner where we need to munge a file name into a valid VB or C# identifier. It's probably worth spending a few minutes looking around to see how that is done, and whether it can be re-used here. If so, that would take care of other potentially problematic characters beyond the space.

That being said, I don't want the perfect to be the enemy of the good, so don't spend too long on that.

tmeschter (approved):

src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsSingleFileGeneratorBase.vb
adamint: It turns out there are already existing methods to generate valid identifiers. These are reused now for both class and namespace

MiYanni (approved):

drewnoakes (approved):

PR #7963, opened 2022-03-09
Optimize up-to-date check for CopyAlways items

### LINKED ISSUE #7961: Optimize fast up-to-date check's handling of CopyAlways items
## Summary

When an item is to be copied to the output directly, there are two options:

1. `CopyToOutputDirectory="Always"` always copies source over destination
2. `CopyToOutputDirectory="PreserveNewest"` copies the file is the timestamp of the source is newer than the destination

Today, having a single `Always` item in a project causes the fast up-to-date check to immediately flag the project as out of date, leading to a build.

It is very easy for users to select the `Always` option and unwittingly break incremental build performance.

~In practice, there does not seem to be a valid use case for `Always` beyond the behaviour of `PreserveNewest`.~

An example of when you might want the behaviour of `Always` is in https://github.com/dotnet/project-system/issues/7961#issuecomment-1063474242.

## Proposal

- ~Change the fast up-to-date check to treat `Always` as it currently treats `PreserveNewest` by default.~
- Change the fast up-to-date check to check timestamps and file sizes for `Always` items, and not fail if they are the same.
- Add an MSBuild project property that restores the old behaviour if needed.

## User Impact

`Always` items are currently responsible for:

- Around 20% of all scheduled builds in VS.
- Around 17% of all incremental build failures in VS.

With the proposed change, the failures should disappear and builds will only be scheduled when the item to be copied is actually out of date with respect to its target file.

drewnoakes:
Fixes #7961

Today, having a single `CopyToOutputDirectory` `Always` item in a project causes the fast up-to-date check to immediately flag the project as out of date, leading to a build.

It is very easy for users to select the `Always` option and unwittingly break incremental build performance.

This PR changes the default behaviour such that having an `Always` item doesn't fail the up-to-date check. Instead, we only fail if the files differ in size or timestamp.

Users who prefer the old behaviour can opt out of this optimization by setting the `DisableFastUpToDateCopyAlwaysOptimization` project property to true.

There were quite a few changes to the fast up-to-date check's log output to better explain behaviour in light of this change, such as including the `CopyToOutputDirectory` value for each item, along with item types.

Unit tests were expanded to cover these new behaviours.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7963)


docs/up-to-date-check.md
```diff
@@ -195,6 +195,23 @@ The logs captured by that tool are usually adequate to diagnose build problems.
 
 ---
 
+## CopyToOutputDirectory Always vs. PreserveNewest
```
tmeschter, line 198: Bonus points for updating documentation. 🥳

docs/up-to-date-check.md
```diff
@@ -195,6 +195,23 @@ The logs captured by that tool are usually adequate to diagnose build problems.
...
+1. `CopyToOutputDirectory="Always"` always copies source over destination.
+2. `CopyToOutputDirectory="PreserveNewest"` copies the file is the timestamp of the source is newer than the destination.
+
+In reality, the use of `Always` is almost never what you actually want, yet it seems like a reasonable choice at first blush.
+
+If `Always` implies the file must be copied, even if it appears to be identical, then the presence of an `Always` item in the must cause the fast up-to-date check to immediately flag the project as out of date, leading to a build that can perform the copy.
+
+It is very easy for users to select the `Always` option and unwittingly break incremental build performance.
+
+In practice, there does not seem to be a valid use case for `Always`. The fast up-to-date check in VS therefore treats it as identical to `PreserveNewest` and only schedules a build if the timestamps or file sizes differ.
```
tmeschter, line 211: Since we're comparing timestamps, this still covers the case where the file is edited by the application at run time, and the user wants it to be replaced with a "fresh" copy when building, correct?

Also, you may want to document this as the one valid case for `Always` that we are aware of.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -654,17 +660,23 @@ private bool CheckCopyToOutputDirectoryFiles(Log log, in TimestampCache timestam
...
+                        // Ignore items which are never copied. Only process Always and PreserveNewest items.
+                        //
+                        // Note that if we see Always items, then state.IsTreatCopyAlwaysAsPreserveNewestDisabled must
+                        // be false, as when it is true and any Always item exists, the check returns before this point.
+
                         continue;
                     }
 
-                    token.ThrowIfCancellationRequested();
+                    System.Diagnostics.Debug.Assert(item.CopyType != CopyType.Always || !state.IsTreatCopyAlwaysAsPreserveNewestDisabled);
```
tmeschter, line 679: Did you mean to leave this in?

tmeschter (approved):
I do suggest updating the PR comment to include the one valid case we know of for `Always`: replacing a file that has been updated by the running application.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: Yes, it's a valid assertion. Though I don't want to throw at runtime for the user should it ever not hold.

docs/up-to-date-check.md
drewnoakes: If the application modifies the file in the output directory at run time, then that output directory item will be newer.

Now that I think of it, I believe we can make this work correctly for Always too:

- For PreserveNewest, schedule a build if the source is different or newer than the destination
- For Always, schedule a build if the source is different or newer or older than the destination

I'll take another pass at this PR and get that fixed, and update the documentation as per your suggestions. Thanks.

drewnoakes:
This PR has been updated in response to the idea mentioned in https://github.com/dotnet/project-system/issues/7961#issuecomment-1065577334.

The first iteration of this PR treated `Always` items as though they were `PreserveNewest`.

`PreserveNewest` items are only copied when the source item is modified.

For `Always` items, we also need to copy if the output is modified.

The new default behaviour has been changed so that `Always` items only fail the build if the timestamps or file sizes differ.

docs/up-to-date-check.md
drewnoakes: > I'll take another pass at this PR and get that fixed, and update the documentation as per your suggestions.

Done and done.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -654,20 +660,26 @@ private bool CheckCopyToOutputDirectoryFiles(Log log, in TimestampCache timestam
...
+                        // Ignore items which are never copied. Only process Always and PreserveNewest items.
+                        //
+                        // Note that if we see Always items, then state.IsTreatCopyAlwaysAsPreserveNewestDisabled must
+                        // be false, as when it is true and any Always item exists, the check returns before this point.
+
                         continue;
                     }
 
-                    token.ThrowIfCancellationRequested();
+                    System.Diagnostics.Debug.Assert(item.CopyType != CopyType.Always || !state.IsCopyAlwaysOptimizationDisabled);
```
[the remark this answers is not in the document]
drewnoakes: Yes, we have a symbol somewhere called `Debug`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckInputItem.cs
```diff
@@ -90,20 +90,20 @@ public UpToDateCheckInputItem(string path, string itemType, IImmutableDictionary
...
-                    if (string.Equals(value, Compile.CopyToOutputDirectoryValues.PreserveNewest, StringComparisons.PropertyLiteralValues))
+                    if (string.Equals(value, None.CopyToOutputDirectoryValues.PreserveNewest, StringComparisons.PropertyLiteralValues))
                     {
-                        return BuildUpToDateCheck.CopyType.CopyIfNewer;
+                        return BuildUpToDateCheck.CopyType.PreserveNewest;
                     }
                 }
 
-                return BuildUpToDateCheck.CopyType.CopyNever;
+                return BuildUpToDateCheck.CopyType.Never;
```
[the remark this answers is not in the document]
drewnoakes: Switch expressions aren't a great fit when you need to specific a comparer unfortunately.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -676,39 +688,58 @@ private bool CheckCopyToOutputDirectoryFiles(Log log, in TimestampCache timestam
...
                     else
                     {
-                        return log.Fail("CopyToOutputDirectoryDestinationNotFound", nameof(Resources.FUTD_CheckingPreserveNewestFileDestinationNotFound_1), destination);
+                        return log.Fail("CopyToOutputDirectoryDestinationNotFound", nameof(Resources.FUTD_CheckingCopyToOutputDirectoryItemDestinationNotFound_1), destinationPath);
                     }
 
-                    if (destinationTime < itemTime)
+                    if (item.CopyType == CopyType.PreserveNewest)
+                    {
+                        if (destinationTime < sourceTime)
```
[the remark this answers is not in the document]
drewnoakes: The non-obvious fact here is that when `log.Fail` is called, the indent value is reset to zero. The only way to exit this block is by normal iteration (which increments/decrements correctly) or by `return log.Fail` which sets it back to zero so that we always show failures in column zero.

Issue #7056, opened 2021-03-26
Improve cancellation support in LaunchSettingsProvider


tmeschter:
The `LaunchSettingsProvider` handles concurrent modifications by queuing them up in a `SequentialTaskExecutor` where they are processed one at a time. Once a `Task` is queued up there is no way of cancelling it, with the exception that disposing the `LaunchSettingsProvider` will cancel all pending `Task`s.

We should looking into updating `LaunchSettingsProvider` (and `ILaunchSettingsProvider`) to accept and respect `CancellationToken`s.

drewnoakes:
It should probably accept `CancellationToken` _instead of_ the current timeout values. It's possible to model the latter with the former.

drewnoakes:
> It should probably accept `CancellationToken` _instead of_ the current timeout values. It's possible to model the latter with the former.

#7581 addresses this for internal code via an extension method. Unfortunately the interface is public so cannot be changed. We may like to consider making the extension method public. We may also need to version the API so that the cancellation token can be passed to internals (which is not currently done in the extension method).

drewnoakes:
We should consider replacing the use of `SequentialTaskExecutor` with a dataflow block. We are pulling items off a datasource and queuing them in the executor. It'd likely be simpler to just use dataflow throughout.

That said, it's unclear to me what would trigger a cancellation here. `ILaunchSettingsProvider` seems like a place we could have accepted cancellation tokens, but it's a public API and hard to change now.

PR #7966, opened 2022-03-09
Wait as long as necessary for an ILaunchSettings snapshot

tmeschter:
Fixes AB#1488489.

Various methods in `LaunchSettingsProvider` call `GetSnapshotThrowIfErrorsAsync` to get the latest `ILaunchSettings` snapshot prior to adding, removing, or updating the launch settings. `GetSnapshotThrowIfErrorsAsync`, in turn, calls `WaitForFirstSnapshot` to ensure we've loaded the initial set of launch settings from the launchSettings.json, and then throws an exception if there was a problem getting the settings.

There are multiple issues with `GetSnapshotThrowIfErrorsAsync` that we address in this commit.

## 1. The call to `WaitForFirstSnapshot` is passed a five second timeout.

If it takes longer than five seconds to read the launchSettings.json and produce the `ILaunchSettings` snapshot we'll give up and throw an exception. This is problematic for a couple of reasons.

First, the methods that call `GetSnapshotThrowIfErrorsAsync` are all trying to modify the launch settings; throwing an exception just means the operation is aborted without any recovery. Even if the callers _were_ set up to recover from the exception, they would just end up trying again. It's better to just wait as long as necessary to obtain the snapshot.

Indeed, some exceptions I've seen indicate that we're trying to read a launchSettings.json file from a folder in OneDrive in order to add a default "in memory" launch profile. If OneDrive needs to download that file on-demand then it makes sense that we've exceeded the five second timeout. In this case the default profile won't be added, and the user will get a gold bar with a cryptic exception message. Nor will the default profile be added until the project is reloaded. Had we just waited until the file was available this would have been avoided.

The second issue is that we don't distinguish between the case where we couldn't load the file before the timeout expired, and the case where the launchSettings.json contained invalid JSON and we were unable to load it. In both cases we throw an exception claiming that the launchSettings.json is faulty, even though in many cases that won't be true.

The fix here is to wait indefinitely for the snapshot to become available, and throw different exceptions to distinguish between a `null` return and an invalid file. Note the added `Assumes.NotNull` should never throw; in the current implementation a different `Assumes.NotNull` in `WaitForFirstSnapshot` would throw first. It is included here to document the expected behavior and to guard against future changes to `WaitForFirstSnapshot`.

## 2. Dropping the error message.

If the launchSettings.json really is malformed the exception we throw will include a message to that effect. However, it does not include any details about the error. It turns out that the "Error" profile we create has these details stashed away in the `OtherSettings` collection.

Here we extract that message and include in the exception we throw.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -1052,6 +1035,33 @@ internal async Task<string> GetLaunchSettingsFilePathNoCacheAsync()
...
         }
 
+        /// <summary>
+        /// Helper retrieves the current snapshot and if there were errors in the launchsettings.json file
+        /// or there isn't a snapshot, it throws an error. There should always be a snapshot of some kind returned.
+        /// </summary>
+        private async Task<ILaunchSettings> GetSnapshotThrowIfErrorsAsync()
+        {
+            ILaunchSettings? currentSettings = await WaitForFirstSnapshot(Timeout.Infinite);
+            Assumes.NotNull(currentSettings);
```
drewnoakes, line 1045: This requires the caller to know that the value is non-null when an infinite timeout is passed. I think we have this in a few places. It'd be nice to replace this with a call to a new overload that doesn't accept a timeout and returns a non-null value.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
tmeschter: Agreed. I think that work would be related to #7056.

davkean:
Do we ever wait on the UI thread for this? If so, then this could hang the product as JTF context is not passed up the chain due to this: http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Managed&leftSymbol=gubvha3royio&file=ProjectSystem%5CDebug%5CLaunchSettingsProvider.cs&line=166. If not, just be aware never to block UI thread on it.

tmeschter:
@davkean That's a good point. I don't think we can assume it will always be called off the UI thread. What's the right way to harden this against potential problems?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsProvider.cs
```diff
@@ -1052,6 +1035,33 @@ internal async Task<string> GetLaunchSettingsFilePathNoCacheAsync()
...
+            if (currentSettings.Profiles.Count == 1 && string.Equals(currentSettings.Profiles[0].CommandName, ErrorProfileCommandName, StringComparisons.LaunchProfileCommandNames))
+            {
+                string fileName = await GetLaunchSettingsFilePathAsync();
+
+                if (currentSettings.Profiles[0].OtherSettings is { } otherSettings
+                    && otherSettings.TryGetValue(ErrorProfileErrorMessageSettingsKey, out object? errorMessageObject))
+                {
+                    throw new Exception(string.Format(Resources.JsonErrorsNeedToBeCorrected_WithErrorMessage_2, fileName, errorMessageObject));
+                }
+                else
```
MiYanni, line 1056: nit: Unnecessary `else`

MiYanni (approved):

melytc (approved):

PR #7969, opened 2022-03-11
Opt in to support solution reload reduction in branch switch scenarios

arkalyanms:
This PR is dependent on https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/385286

Currently for branch switches all changes to the solution file will trigger solution reloads. I have improvements targeted in the BFO and solution code space to reduce the solution reloads and is gated behind a feature flag that is disabled at the moment. 

I would also like to start targeting these changes for CPS projects first before opening it up for legacy and VC. This PR introduces the needed attributes that I can then consume in the project-system side to claim default support. 

Note: This will still be gated behind the solution reload optimization feature flag that is default disabled for now. It will enabled after A/B testing the VS side changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7969)


davkean (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ProjectTypeRegistration.cs
```diff
@@ -16,7 +16,8 @@
     DisableAsynchronousProjectTreeLoad = true,
     PossibleProjectExtensions = "vbproj",
     NewProjectRequireNewFolderVsTemplate = true,
-    SupportsCodespaces = true)]
+    SupportsCodespaces = true,
+    SupportsReloadFreeBranchSwitch = true)]
```
drewnoakes, line 20: As mentioned on the CPS PR, could the property name be `SupportsSolutionChangeWithoutReload`?

```suggestion
    SupportsSolutionChangeWithoutReload = true)]
```

davkean: Which is a very good point, because you'd do this regardless of whether it was a branch switch or not.

drewnoakes:
@arkalyanms is this for 17.2 or 17.3?

arkalyanms:
> > This is for 17.3.
> 
> `main` is still targeting 17.2.

Makes sense. This can wait.

arkalyanms:
@drewnoakes I just realized the typo in my above comment. This is for 17.2 Preview 3.0. The feature will be go in default disabled for 17.2 though and is going in just to get some preview coverage when features activate the feature flag or via an experiment.

drewnoakes:
@arkalyanms no problem. Can you get the build to pass?

arkalyanms:
> @arkalyanms no problem. Can you get the build to pass?

Sure, will bump up the CPS versions.

drewnoakes:
Thanks @arkalyanms. This should be inserted overnight, though we have seen some delays in insertions over the last week.

Issue #7977, opened 2022-03-14
ArgumentNullException when setting same value across all configuration dimensions


drewnoakes:
## Visual Studio Version

17.2.0 Preview 3.0 [32308.589.main]

## Summary

It's possible to trigger a yellow bar while using project properties.

## Steps to Reproduce

1. Create a new .NET Console App (targeting .NET 6)
2. Open the Project Properties UI
3. Navigate to Build | Errors and warnings | Warning level
4. Click the gear to vary by configuration
5. Click the gear again to use the same value across all configurations

This produces a yellow bar for exception:

```
=====================
15/03/2022 10:50:55
Recoverable
System.AggregateException: One or more errors occurred. ---> System.ArgumentNullException: Value cannot be null.
Parameter name: unevaluatedPropertyValue
   at Microsoft.Requires.NotNull[T](T value, String parameterName)
   at Microsoft.VisualStudio.ProjectSystem.Properties.ProjectPropertiesBase.Microsoft.VisualStudio.ProjectSystem.Properties.IContextSpecificProjectProperties.SetPropertyValueInContextAsync(String propertyName, String groupLabel, String unevaluatedPropertyValue, IReadOnlyDictionary`2 dimensionalConditions)
   at Microsoft.VisualStudio.ProjectSystem.PropertyPages.PageProperty.<SetValueAtDataSourceAsync>d__43.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.ProjectSystem.PropertyPages.PageEnumPropertyBase.<SetValueAsync>d__8.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.VisualStudio.ProjectSystem.VS.Query.ProjectSetUIPropertyValueActionCore.<ExecuteAsync>d__7.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #7982, opened 2022-03-15
Update project properties UI for warning levels

### LINKED ISSUE #7978: Show better descriptions for warning levels
Today the "Warning level" property shows:

image

It's not clear what these numbers mean.

From https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#warninglevel we see:

| Level | Description |
|-|-|
| 0 | Turns off emission of all warning messages. |
| 1 | Displays severe warning messages. |
| 2 | Displays level 1 warnings plus certain, less-severe warnings, such as warnings about hiding class members. |
| 3 | Displays level 2 warnings plus certain, less-severe warnings, such as warnings about expressions that always evaluate to true or false. |
| 4 | Displays all level 3 warnings plus informational warnings. |
| 5 | Displays level 4 warnings plus [additional warnings](https://github.com/dotnet/roslyn/blob/a6013f3213c902c0973b2d371c3007217d610533/docs/compilers/CSharp/Warnversion%20Warning%20Waves.md) from the compiler shipped with C# 9.0. |

We should add some of these details to the display text for these enum values in our rules files.

Additionally:

- As level 0 turns off all warnings, when selected we should hide "Suppress specific warnings".
- As level 5 is only available for C# 9 and later, we would ideally hide this value (however we do not support per-value visibility conditions)

drewnoakes:
Fixes #7978

- Update description on "Warning level" proprety
- Add display values to "Warning level" enum members
- Hide other warning related properties when "Warning level" set to zero

warning-level-zero


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -175,37 +175,48 @@
...
+               DisplayName="2 - Less severe warnings" />
     <EnumValue Name="3"
-               DisplayName="3" />
+               DisplayName="3 - Less severe warnings" />
     <EnumValue Name="4"
-               DisplayName="4" />
+               DisplayName="4 - Informational warnings" />
     <EnumValue Name="5"
-               DisplayName="5" />
+               DisplayName="5 - Additional warnings from C# 9" />
```
tmeschter, line 192: Any reason we shouldn't use the text in #7978 for each of these?

drewnoakes: Some of those sentences were very long, which I assumed wouldn't work well. But you're right that I should at least try it.

drewnoakes: Long strings work in the UI better than I expected. I've updated the PR now to:

image

image

image

Any concerns?

MiYanni: I avoided doing this kind of thing in the property pages for VB PR. You can see those original descriptions here: https://github.com/dotnet/project-system/pull/7928/commits/8b1a5fa2a7d602eb6fedaf6a9d6b36ac995255f5#diff-ae4eb5fe552b47b5719cb7301a0831ca640f3ac659e1a68136bfac7147280f69R444-R458

I originally had descriptions similar to this, but decided against it for a few reasons:
- No clear pattern to follow
  - Here, you used `{value} - {description}` but that is just freehand pattern, not something built into the control
- Without testing, how does the dropdown perform on a small window?
  - Does the item name wrap? Go off the edge of the application? Truncate?
- How easy is it to read/understand from a user perspective?
  - Is it accessible this way?
  - Does it work well with translations/localization?

Instead, I opted to only have the value and relied on the ❓ button to allow the user to read what the values do if they need that information. The linked documentation has better descriptions than we can easily convey in the UI. I also believe we should have a designer weigh in on this concept as these few instances won't be the only dropdowns with long descriptions. We need to have solid guidance to give other teams when implementing their own property pages. Adding @mckennabarlow for input.

melytc: I like to have a more descriptive value, so I'm in favor of having these texts to let the user know what those numbers translate to. My only concern is, as Yanni mentioned, how it behaves when the window resizes, specially with long strings. I do think it's easier to understand but I'll leave that to be answered by McKenna :).

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
drewnoakes: There's a balance here. We want to keep things concise, without removing useless information that could be kept.

The help button is great for learning context around the property's purposes and impact, but is not great as a reference to decode the values we show in the UI.

I also checked and we don't use the `{number} - {description}` pattern elsewhere. But I think it is a natural and obvious pattern. We also need the numbers there to support the description explaining that higher levels include lower levels.

I tested truncation of long values. It currently crops on the RHS, but that's true of all our current controls. We have complete control over this in WPF though, and can wrap long strings if necessary.

tmeschter:
@drewnoakes As I was investigating #7977, I discovered that `<WarningLevel>` can have values greater than 5, and in some cases the SDK forces it to higher values. For example, a new .NET 6 console app will have to set to "6"; "9999" is also a possible value. In many cases the UI won't show a value--not because it doesn't have a value, but because we can't map it to one of the supported values.

I'd suggest holding off on any changes here until we can rationalize the overall behavior of `WarningLevel`.

drewnoakes:
@tmeschter I wonder if we can use an enum provider to always include the project's current value. That should address the issue. If so, that's orthogonal to the change here.

drewnoakes:
I can see value in having a combo box for some enum properties. That would also allow MSBuild expressions to be added for enum properties, in case that was useful.

tmeschter:
@drewnoakes Per conversation with the Compiler team, can you add the following values while you're at it?

- 6 - "Warnings from C# 10"
- 7 - "Warnings from C# 11"
- 9999 - "All warnings"

Those are additional values that are currently provided by the SDK, so we should really have them as admissible values.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
```diff
@@ -175,37 +175,48 @@
...
 
   <StringProperty Name="NoWarn"
                   DisplayName="Suppress specific warnings"
                   Description="Blocks the compiler from generating the specified warnings. Separate multiple warning numbers with a comma (',') or semicolon (';')."
                   HelpUrl="https://go.microsoft.com/fwlink/?linkid=2147300"
-                  Category="ErrorsAndWarnings" />
+                  Category="ErrorsAndWarnings">
+    <StringProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(not (has-evaluated-value "Build" "WarningLevel" "0"))</NameValuePair.Value>
```
tmeschter, line 202: Now that I think about it, I don't think we can condition `NoWarn`, `TreatWarningsAsErrors`, `WarningsAsErrors`, or `WarningsNotAsErrors` on the value of `WarningLevel`. Analyzers may report warnings regardless of the `WarningLevel`, and these other properties can control which of those are promoted to errors or suppressed.

tmeschter (changes requested):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
drewnoakes: Good point. Fixed. Maybe we should explain this in the description of the WarningLevel property somehow. Can you think of a nice addition to the current text:

> Specifies the level to display for compiler warnings. Higher levels produce more warnings, and include all warnings from lower levels.

drewnoakes:
> can you add the following values

Done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
tmeschter: In talking with the compiler team about #7977, the expectation is that, moving forward, the `WarningLevel` will primarily be about compiler warnings _and_ warnings from analyzers built into the SDK. 3rd party analyzers are not generally expected to use `WarningLevel`. So I think I would write something like:

> Specifies the level to display for compiler warnings and .NET SDK analyzer warnings. Higher levels produce more warnings, and include all warnings from lower levels.

tmeschter (approved):

abelbraaksma:
I went over this change and the comments here and I see one mention of VB, and none of F#. What is the expectation how this behaves in terms of shown values for VB projects and/or F# projects? 

Currently, in the latter case, it shows the C# options (which are invalid for F#) and I believe for VB, it also shows the C# descriptions. Not sure if this was an intentional side effect, but I assume that any text/description etc should be either language agnostic, or language specific. Reported separately in #9469.

PR #7985, opened 2022-03-16
Passing VersionOverride metadata as part of nomination

jeffkl:
In https://github.com/NuGet/NuGet.Client/pull/4426 I added a new attribute to `<PackageReference />` to allow the ability to override a centrally defined package version.  This change adds the property to nomination in project system

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7985)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/CollectedPackageReference.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/CollectedPackageReference.xaml
@@ -56,6 +56,11 @@
                   ReadOnly="True"
                   Visible="false" />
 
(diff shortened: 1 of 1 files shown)


drewnoakes (changes requested):
This change is in the wrong file. This file is used to populate package nodes in the dependencies tree.

The one used by nomination is `src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\Rules\CollectedPackageReference.xaml`.

jeffkl:
Sorry about that, I was referencing a pretty old PR that did something similar but didn't realize things had changed.  I've pushed a new commit.

drewnoakes (approved):

PR #7989, opened 2022-03-21
Remove redundant casts (solution-wide codefix)

drewnoakes:
Roslyn flagged these as redundant, and they were removed via a codefix.

Three categories of casts were removed:

1. C# 10 does a better job of inferring the type for ternary conditional expressions, based on the target type
2. Casts that were previously required for nullable reference types
3. Casts of zero to enum types are not required

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7989)


MiYanni (approved):
I'm very happy for this language enhancement. It always felt quite sloppy to have to add these casts to the code. Much cleaner. 🙂

PR #7990, opened 2022-03-21
Idempotent change won't mark dependency unresolved

### LINKED ISSUE #2510: Idempotent Project Changes can leave Dependency Node Unresolved
**Repro Steps**

Create a .NET Standard project and add the following package references:

```xml
  <ItemGroup>
    <PackageReference Include="Newtonsoft.Json" Version="10.0.0" />
    <PackageReference Include="StyleCop.Analyzers" Version="1.0.2" />
  </ItemGroup>
```

Observe that the Newtonsoft.Json reference is flagged with warning NU1603 Package not found, approximate best match was resolved.

image

Add NoWarn="NU1603" to the StyleCop.Analyzers reference (i.e. to the wrong node)

**Expected**

No changes to dependencies node.

**Actual**

StyleCop.Analyzers becomes and remains unresolved, anticipating a new resolved node. This is likely because the change did not trigger a dataflow with a "Difference".

Workaround is to make some other change that triggers a dataflow (e.g. change version to 1.0.1)

### LINKED ISSUE #5356: Renaming a project marks package references as unresolved
**Visual Studio Version**: 16.2.2

**Steps to Reproduce**:

1. Create new .NET Core Console App
2. Add a package reference and wait for it to resolve
3. Rename the project
4. Notice package reference is unresolved

Making a change to the set of packages will trigger it to refresh and appear correctly.

drewnoakes:
Fixes #2510 
Fixes #5356

Previously, any evaluation of the project that changed a dependency would mark it as unresolved until the build data arrives.

Some kinds of changes trigger a change in evaluation that is not accompanied by a corresponding change in build data. The evaluation update would set the dependency to unresolved, expecting the build to resolve it. Yet the build delta would exist and so the dependency would be left unresolved.

The changes in this PR cause the tree to remember the set of resolved references. When an evaluation occurs, we do not mark a known-resolved dependency as unresolved. If the build does not trigger a change, then the dependency remains in the correct state.

There are some cleanups in separate commits. The main change is in 9859b6fa64e06f4e7df073193aab1fa7b8dba2b6.

## Before

fix-2510-before gif

## After

fix-2510-after gif

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7990)


drewnoakes:
This also fixes #5356, where renaming the project would force dependencies to become unresolved.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
```diff
@@ -181,6 +166,10 @@ void HandleChangesForRule(bool resolved, IProjectChangeDescription evaluationPro
 
             bool isImplicit = IsImplicit(projectFullPath, evaluationProperties);
 
+            // When we only have evaluation item, mark the dependency as resolved if we currently have a resolved item improve
+            // this, so that if a resolved item is known, we don't revert back to unresolved state.
```
MiYanni, line 170: nit: Run-on sentence

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
```diff
@@ -181,6 +166,10 @@ void HandleChangesForRule(bool resolved, IProjectChangeDescription evaluationPro
...
+            isResolved = isResolved || changesBuilder.HasResolvedItem(targetFramework, ProviderType, originalItemSpec);
```
MiYanni, line 171: nit: Is short-circuiting required here? If not, you can do bitwise OR instead.
```suggestion
            isResolved |= changesBuilder.HasResolvedItem(targetFramework, ProviderType, originalItemSpec);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/RuleHandlers/PackageRuleHandler.cs
```diff
@@ -225,6 +185,21 @@ public PackageRuleHandler(ITargetFrameworkProvider targetFrameworkProvider)
...
+                // This package is present in build results, but not in evaluation.
+                // We disallow packages which are not present in evaluation.
+                dependencyModel = null;
+                return false;
+            }
+
+            bool isImplicit = IsImplicit(projectFullPath, evaluationProperties);
+
+            // When we only have evaluation data, mark the dependency as resolved if we currently have a corresponding resolved item
+            bool isResolved = isResolvedItem || changesBuilder.HasResolvedItem(targetFramework, ProviderType, originalItemSpec);
```
MiYanni, line 201: nit: Same comment as above.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
drewnoakes: Thanks, fixed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/RuleHandlers/DependenciesRuleHandlerBase.cs
drewnoakes: Not "required", but I prefer to avoid unneeded work where possible.

PR #7993, opened 2022-03-21
Avoid race condition in JointRuleSource

drewnoakes:
Follows @lifengl's comment here: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/387180

CPS provides the `JointRuleSource` Dataflow source block. When linking to it, you pass the names of rules you want to receive data for.

Historically, we've passed a single string array containing names from both evaluation and design-time builds. The implementation would then split these rule names into those two buckets based on the corresponding rule's data source details.

In order to look up a rule by name, the latest version of `IProjectCatalogSnapshot` (from `ProjectCatalogSnapshotService`) is retrieved. This can create a race condition during system initialization, where the subscription for rules arrives before the catalog snapshot.

This change splits out the evaluation from the build rules using the new `JointRuleDataflowLinkOptions` type, anywhere we subscribe to `JointRuleSource`. This avoids needing the catalog snapshot, and hence avoids the race condition.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/7993)


drewnoakes:
@lifengl is there a good way to test that this is working? I couldn't use the new `LinkRuleSubscriptionTo` method, and I didn't manage to trace the `DataflowLinkOptions` value through all the layers of abstraction to prove it will be picked up, but I did have the sense it should work.

lifengl (approved):
looks good to me

MiYanni (approved):

PR #8013, opened 2022-03-26
Re-enable OptProf

MiYanni:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1497655

OptProf was disabled because it required the VS Bootstrapper task. This task was silently updated during SBOM changes and caused our build to start failing. To allow us to continue checking in code, I disabled OptProf which uses the VS Bootstrapper task. The VS Bootstrapper task required an updated MicroBuild Swix plugin, which at the time, I believed could not be updated because an old version is required by RepoToolset.

I did some investigation and found that the version of the Swix plugin we were installing via the DevOps task was `1.1.87`. However, if you look at the build logs, the version used by RepoToolset is `1.1.37`. This means the task we have in our pipeline was **only** being used by VS Bootstrapper, and RepoToolset acquires its own package to do VSIX creation. Thus, I could update the plugin to the latest version (which installs 1.1.2XX) and I put this task in the OptProf section since it is not needed prior to doing the build itself.

For this PR to work (meaning, the build succeeds), we need an updated set of optimization data. Currently, I hooked up our OptProf release pipeline to the branch for this PR. This **should** give us optimization data to use when the normal build runs. If this doesn't work, after this PR, I'd need to run a build manually and have (the use of) optimization disabled so that it can produce new optimization data for the following build. Either way, I'll have to update our release pipeline here to trigger against `main` at the time of merging this PR: https://devdiv.visualstudio.com/DevDiv/_release?view=all&_a=releases&definitionId=3197

To the issue side of things, I've been talking to Jackson and it is likely that OptProf reduces the NGEN'ed binary size. The issue created by the perf tests seems to suggest that disabling OptProf increased the size of the NGEN'ed binaries. We don't produce these binaries but they are created as part of the setup authoring process (a process that is, to me, a black box at the moment). So, the failure is on a set of information I've never seen before, and we don't directly produce during a build. If we see the size of the `.ni.dll` binaries in the perf tests go back down after this PR, that proves this assumption.

Additionally, this PR will add a warning to our build. I couldn't figure out how to resolve it in a timely manner so it will remain for the time being.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8013)

### DIFF
--- a/build/ci/official.yml
+++ b/build/ci/official.yml
@@ -51,6 +51,9 @@ steps:
 - task: MicroBuildSwixPlugin@1
   displayName: Install Swix Plugin
 
+- task: MicroBuildSwixPlugin@4
+  displayName: Install Swix Plugin
+
 # This requires the OptimizationInputs via the 'Publish OptimizationInputs drop' step in the release pipeline.
 # To generate ProfilingInputs for the first time, set this input: ShouldSkipOptimize: true
 # See documentation here: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/2602/Build-With-Expired-(or-Without-Prior)-Optimization-Profiling-Data
@@ -66,7 +69,6 @@ steps:
     feedSource: 'https://devdiv.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json'
     ShouldSkipOptimize: ${{ parameters.SkipOptimize }}
     NumberCommitsToSearch: '100'
-  enabled: false
 
 ###################################################################################################################################################################
 # BUILD SOLUTION
@@ -186,7 +188,6 @@ steps:
     usePat: true
     AccessToken: '$(System.AccessToken)'
   condition: succeeded()
-  enabled: false
 
 # MicroBuildBuildVSBootstrapper requires MicroBuildSigningPlugin for signjson.exe to run.
 - task: MicroBuildBuildVSBootstrapper@2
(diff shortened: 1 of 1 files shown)


MiYanni:
Successful build here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5938729&view=results

MiYanni:
Ran into an issue with the OptProf release pipeline where I used `feature/ReenableOptProf` as the branch but this is what the source branch name came out as:
```
[RELEASE_ARTIFACTS_PROJECTSYSTEM-OPTPROF_SOURCEBRANCHNAME] --> [ReenableOptProf]
```
Something is dropping the `feature/` out of the branch name. I'm not sure if that is the cause of the problem, but I've created a new branch, `ReenableOptProf2`, without the `feature/` in it. Trying that branch now.

MiYanni:
New successful build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5939314&view=results
Currently running OptProf pipeline: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=1296631

drewnoakes (approved):

MiYanni:
First build (no optimization): https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5953028&view=results
OptProf pipeline: TBA
Second build (with optimization): TBA

PR #8021, opened 2022-03-30
Enforce correct MEF types and add implicit usings

drewnoakes:
Adds two new implicit usings:

- `System.Collections.Immutable` because we use it a lot
- `System.ComponentModel.Composition` because we use it a lot, and it's easy to confuse these types with those in `System.Composition` which should not be used.

These implicit usings introduced one ambiguity, which is fixed in 621716109beb0ec8e6cb7242581264254441b815.

The removal of redundant using directives is in be53ffd582c33a92fde83d343013e2a7463b8834.

As part of making the correct namespace for MEF attributes available implicitly, we also ban the use of incorrect MEF attributes in a05fb152f004bab826ff2621a8d71ff5292e2538. With that change, attempting to use the wrong types will be flagged during compilation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8021)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -40,6 +40,14 @@
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
   </PropertyGroup>
 
+  <!-- Modify the set of implicit usings -->
+  <ItemGroup>
+    <!-- We use immutable collections extensively and expect no naming conflicts -->
+    <Using Include="System.Collections.Immutable" />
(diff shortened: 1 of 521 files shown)


MiYanni (approved):

melytc (approved):
Nice!

PR #8023, opened 2022-03-30
Remove intermediate Dataflow block

drewnoakes:
Previously, an intermediate Dataflow block was created for the sole purpose of specifying rule names. It's more to configure this directly during the `SyncLink` operation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8023)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/Subscriptions/DependencyRulesSubscriberBase.cs
```diff
@@ -81,14 +75,7 @@ public void ReleaseSubscriptions()
 
             _subscriptions ??= new DisposableBag();
```
lifengl, line 76: maybe we no longer need a DisposableBag here as we have only one link to dispose.

lifengl (approved):
changes look good to me.

melytc (approved):

PR #8040, opened 2022-04-05
Fix encoding errors in doc comments

drewnoakes:
���, ���� � ���.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8040)


MiYanni (approved):
I always wondered what editor ends up causing this. The character was likely `right single quotation mark` (aka `’` or U+2019), but something malforms it into `replacement character` (aka `�` or U+FFFD). Not sure why that happens, though.

PR #8044, opened 2022-04-07
Consume new IEntryPointFinderService api

jmarolf:
Depends on https://github.com/dotnet/roslyn/pull/60614. 

That PR will need to be merged first and then consumed by this Repo.

Both Roslyn and Project-System will need to be inserted into VS in the same PR to avoid things breaking.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8044)


MiYanni:
@jmarolf Howdy. It looks like the PR this is dependent on hasn't been updated in half a year. I'd recommend we close this for the time being until the dependent changes are introduced. Thanks!

PR #8045, opened 2022-04-07
[17.2] Fix broken analyzer child nodes in Solution Explorer

drewnoakes:
In 17.2, CPS made a change to how paths are used for non-file items in the tree. These changes improve the performance of searching the dependency tree by path, as the pseudo-id used for items not having paths allows a fast lookup.

That change broke the ability for Roslyn to determine the correct file path for analyzer nodes (via `CanonicalName`), which then broke the ability to attach child nodes to the analyzer in Solution Explorer (for example, to show source generators and the files they create).

In this commit, we add the `FileSystemEntity` flag to analyzer nodes. This causes CPS to preserve the full path provided on the `IProjectTree`, which restores the `CanonicalName` for Roslyn.

## Before

image

## After

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8045)


drewnoakes:
Merging here without review in order to meet 17.2p3 deadlines. We will take review on the insertion PR instead and back this out here if there are concerns.

lifengl (commented):
looks good to me

PR #8051, opened 2022-04-08
Fix up-to-date tracking for App.config files

### LINKED ISSUE #8015: updated app.config not output - SDK-style project file in Visual Studio
## Visual Studio Version
Visual Studio 2022 17.1.2

## Summary
In an SDK-style project when the app.config file is updated in Visual Studio, the changes aren't reflected in the output .exe.config.

Seems to match https://github.com/dotnet/project-system/issues/4569 but I think even rebuild doesn't help.

## Steps to Reproduce

1. Have a .NETFramework SDK-style project, eg.:
``` xml
	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<TargetFramework>net472</TargetFramework>
	</PropertyGroup>

	<ItemGroup>
		<Reference Include="System.Configuration" />
	</ItemGroup>
</Project>
```

2. Run/debug the app.
3. Update the App.config
4. Run/debug the app again (without restarting). 

NB: Adding [`<AppConfig>App.config</AppConfig>`](https://stackoverflow.com/a/50373936) didn't seem to help.

## Expected Behavior
When the app runs, the config file in the bin directory (eg. appname.exe.config) has the changes.

## Actual Behavior
The config file in the bin directory is unchanged.

## Workaround

1. Restart visual studio each time the config should be updated
OR
2. Set the "Copy to Output Directory" property to "Always"

``` xml
	  <None Update="App.config">
	    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
	  </None>
```

## User Impact
Lots of wasted time debugging until you release, and then again each time your forget ;)

drewnoakes:
Fixes #8015

There was an attempt two years ago to fix this, in #5994. I don't see how that fix could have ever worked.

The MSBuild item that modelled the `App.config` source/destination paths had an invalid condition. It was verifying that the an `AppConfig` _item_ existed, while it should have been verifying an `AppConfig` _property_ existed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8051)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -440,7 +440,7 @@
            https://github.com/dotnet/msbuild/blob/d2f9dbccd913c5612fd3a3cb78b2524fbcb023da/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1152-L1165
            We skip this check if AppConfig is empty, which occurs for .NET Framework console apps. See https://github.com/dotnet/project-system/issues/6758.
       -->
-      <UpToDateCheckBuilt Condition=" '@(AppConfigWithTargetPath)' != '' and '@(AppConfig)' != '' " Include="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Original="$(AppConfig)"/>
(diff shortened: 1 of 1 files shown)


MiYanni (commented):
A couple questions:
- It seems using the item template "Application configuration file" doesn't add the `<AppConfig/>` property to the project file. How will this fix help without manually editing the project file?
- Would a unit test be useful for this scenario or is it something that only an integration/functional test could handle?

drewnoakes:
There's no need to add an AppConfig property to the project file. Simply having an App.config file on disk is enough to make things work.

To follow the thread on this, see the comment right above the change. It references a target. You'll see that target's implementation in MSBuild's common targets. It calls a task that populates the AppConfig property, along with the AppConfigWithTargetPath item we are also using.

As for testing, because the logic requires a design time build, we would need some kind of integration test (or an alternative way to run DTBs in a test).

Those targets populate an UpToDateCheckBuilt item, which is covered by our existing unit tests.

MiYanni (approved):

PR #8067, opened 2022-04-12
Add the PackageOutputPath property to the Package page

### LINKED ISSUE #2709: VS2017 .NET Standard - Pack option - generate Nuget Packages in a unified directory for all projects in the solution / customize package creation path
_From @mss37 on August 14, 2017 10:35_

Please read the following information before posting the issue.
## Details about Problem

NuGet product used (VS UI | Package Manager Console | dotnet.exe):

NuGet version (Package Manager Console Host Version 4.1.0.2427):

dotnet.exe --version (2.0.0-preview2-006497):

VS version (VS2017 26403.7):

OS version (win10):

Worked before? If so, with which NuGet version: No

## Detailed repro steps so we can see the same problem

1. Create a .NET Standard project (class library)

2. Configure it to create a Nuget package (from project properties, Package tab).

3. Build, it will build in (bin/debug), but there is no way to configure it to create the package in a relative path, where that path can be set as a local package repo, to reference from it later.

## Other suggested things
Suggest to add this option (to customize the path where Nuget packages are created).

_Copied from original issue: NuGet/Home#5753_

drewnoakes:
Fixes #2709

This property allows specifying the path on disk where package files are dropped when the project is packed.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8067)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
@@ -174,6 +174,11 @@
     </StringProperty.Metadata>
   </StringProperty>
 
+  <StringProperty Name="PackageOutputPath"
+                  DisplayName="Package Output Path"
+                  Description="Determines the output path in which the package will be dropped."
+                  Category="General" />
+
   <DynamicEnumProperty Name="NeutralLanguage"
                        DisplayName="Assembly neutral language"
                        EnumProvider="NeutralLanguageEnumProvider"
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/PackagePropertyPage.xaml.cs.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/PackagePropertyPage.xaml.cs.xlf
@@ -172,6 +172,16 @@
         <target state="translated">Soubor s licencí</target>
         <note />
       </trans-unit>
+      <trans-unit id="StringProperty|PackageOutputPath|Description">
+        <source>Determines the output path in which the package will be dropped.</source>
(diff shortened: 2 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -157,6 +157,11 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <StringProperty Name="PackageOutputPath"
+                  DisplayName="Package Output Path"
+                  Description="Determines the output path in which the packed package will be dropped."
+                  Category="General" />
```
MiYanni, line 163: No help url available?

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
drewnoakes: I couldn't find a sensible URL for this property. There's a passing reference to it in a table of MSBuild properties that impact pack operations, but that table only mirrors the text we have here so doesn't add anything.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
```diff
@@ -157,6 +157,11 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <StringProperty Name="PackageOutputPath"
+                  DisplayName="Package Output Path"
+                  Description="Determines the output path in which the packed package will be dropped."
```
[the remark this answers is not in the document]
drewnoakes: Seems reasonable. I copied the text from NuGet's documentation. I'll have to deal with conflicts here anyway, so will change it.

Issue #8066, opened 2022-04-12
Condition package properties on IsPackable property


drewnoakes:
A project can only be packed if the `IsPackable` property is true.

For example:

1. Create a new ASP.NET Core Web Application
2. Right click the project and select Pack
3. See warning:
   > C:\Program Files\dotnet\sdk\6.0.300-preview.22178.8\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(203,6): warning : **This project cannot be packaged because packaging has been disabled. Add <IsPackable>true</IsPackable> to the project file to enable producing a package from this project.**

Note that the project can be made packable by manually setting the `IsPackable` property to `true`.

Even though this web app is not packable, we still show all the package properties in the UI.

I can see two possible ways to improve this:

1. Model `IsPackable` as a non-visible property in `ConfigurationGeneralPage.xaml`
2. Model `IsPackable` as a visible and editable property in `PackagePropertyPage.xaml`

In either case, we can use visibility conditions to hide pack-related properties from the UI. Note that some properties on the "Packaging" page, such as versions, have utility outside of pack actions. We would need to do a proper review here.

PR #8070, opened 2022-04-12
Add properties to control symbol packages

drewnoakes:
When packing a project, the tooling supports producing an additional package containing debug symbols.

This change adds a boolean property to opt in to this feature, and a drop down to select the format of this package.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8070)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/PackagePropertyPage.xaml
@@ -174,6 +174,11 @@
     </StringProperty.Metadata>
   </StringProperty>
 
+  <StringProperty Name="PackageOutputPath"
+                  DisplayName="Package Output Path"
+                  Description="Determines the output path in which the package will be dropped."
+                  Category="General" />
+
   <DynamicEnumProperty Name="NeutralLanguage"
                        DisplayName="Assembly neutral language"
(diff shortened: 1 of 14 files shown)


MiYanni (approved):
I'm noticing a pattern where we have these booleans (checkboxes) that basically enable sub-features. It isn't obvious which booleans do this without checking/unchecking the box. Not a problem to solve for this PR but something to think about; maybe there is a better way to expose this pattern to the user.

melytc (approved):

drewnoakes:
> It isn't obvious which booleans do this without checking/unchecking the box.

This is an interesting point to think about.

In the legacy UI, it was more common for UI to be greyed out, but I think that was more because doing dynamic layout in WinForms is difficult.

We now have several use cases where we expect hiding to occur. For example, hiding entire pages of properties for MAUI projects when they don't target particular platforms. Another example is in #8066. If we wanted to show non-editable properties, we'd have to bear those cases in mind.

Also, if search returns a non-editable property, it might not be clear exactly why that property was non-editable, or how to make it editable.

One option would be to indicate when toggling one property would impact other property's visibilities. I see two open questions for that:

1. How do you present that to the user in a way that is more helpful than confusing?
2. How do we even know that changing a property would impact another property's visibility, when the visibility condition is on other those properties?

drewnoakes:
Rebased on top of #8067 (which I will merge first) to address merge conflicts. That made the diff here look larger, so just review d30643b890850ffc8e8291d172e9d25fa47bf072.

Issue #7330, opened 2021-06-18
Re-create environment variable data grid


tmeschter:
We need to re-create the environment variable data grid we had in the old property pages:

![image](https://user-images.githubusercontent.com/10506730/122601761-a45fab80-d026-11eb-96ac-7da51b158134.png)

Right now the environment variables are exposed as a specially-formatted text string, which is not as good an experience.

Issue #167, opened 2016-05-19
Add a checkbox to skip loading NGEN'd assemblies when debugging


davkean:
Failing (or in addition) to: https://github.com/dotnet/roslyn-project-system/issues/166, it would be great to be have a check box that sets `COMPLUS_ZAPDISABLE=1` in the target process.

feasibletrash0:
would be cool if it was just a checkbox

davkean:
https://docs.microsoft.com/en-us/visualstudio/debugger/jit-optimization-and-debugging?view=vs-2019#limitations-of-the-suppress-jit-optimization-option

drewnoakes:
@tmeschter do you have a sense of what would be required to add this with the new launch profiles system?

tmeschter:
Adding a checkbox to the UI is easy. The question is how to represent those settings in the launchSettings.json.

If we continue to store them in the launch profile as environment variables then we need to filter them out for the control that handles environment variables in general.

If we store the setting separate from the environment variables then the code path that actually launches the process needs to understand the setting and update the environment variables appropriately.

Neither is hard to implement, but either approach may have implications for other project types that reuse either the property page or the process launch path.

drewnoakes:
The first approach sounds similar to how we treated conditional compilation symbols in the legacy UI.

For the second approach, we could version the public `ILaunchProfile` interface to include these properties. Doing that could also help with preserving the order of our dictionary items.

tmeschter:
I definitely favor the second approach.

The properties could also go in `OtherSettings`, which is always a safe choice.

Ideally whatever we do here would also be respected in command-line scenarios that use launchSettings.json. We should bring this up with the SDK/CLI team before making any decisions about how to represent this setting.

PR #8072, opened 2022-04-13
Preserve environment variable order in launch profiles

### LINKED ISSUE #8071: Launch profiles should preserve environment variable order
## Visual Studio Version

17.0+

## Summary

Launch profiles may specify custom environment variables to be set on the launched process.

In `launchSettings.json` this might resemble:

```json
{
  "profiles": {
    "ConsoleApp399": {
      "commandName": "Project",
      "environmentVariables": {
        "donut": "4",
        "carrot": "3",
        "banana": "2",
        "apple": "1"
      }
    }
  }
}
```

However once we load this data into a `LaunchProfile` that original order is not maintained. This happens because `ILaunchProfile.EnvironmentVariables` is defined as an `ImmutableDictionary<,>` which does not preserve the order of the items within it.

We received feedback about this in https://developercommunity.visualstudio.com/t/Dont-Reorder-my-launchSettingsjson-env/1701070 ([AB#1505693](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1505693)).

## 1. UI doesn't reflect JSON

Environment variables from `launchSettings.json` are reordered alphabetically by name before display, as seen here:

image

## 2. Edits made by the user jump around during editing

zygote

This is confusing. It is also blocking progress on #7330 (provide a better UI editor for environment variables). We can't have the data reordering beneath us while the user makes edits.

drewnoakes:
Fixes #8071

`ILaunchProfile.EnvironmentVariables` is defined as an `ImmutableDictionary<,>`, which does not preserve the order of the items within it. This means we would reorder a user's environment variables between the `launchSettings.json` file and the UI.

Unstable ordering has two problems:

1. It has blocked us from providing a better UI editor for environment variables. We can't have the data reordering beneath us while the user makes edits.
2. User confusion/frustration. We received feedback about this in https://developercommunity.visualstudio.com/t/Dont-Reorder-my-launchSettingsjson-env/1701070

Unfortunately `ILaunchProfile` is a public API, so we cannot modify the collection type used to store environment variables. Therefore, the approach taken to provide stable ordering in this commit can be summarised as:

- Add new state on `LaunchProfile` that tracks the order of keys in the environment variables map.
- Add extension methods on `ILaunchProfile` that produce ordered data when the underlying type is `LaunchProfile`.
- Create static class `ActiveLaunchProfileEnvironmentVariableValueProvider` to own the formatting and parsing of the string property value representation of environment variable name/value pairs.
- Make `LaunchProfile` immutable to provide clearer guarantees of consistency between the collections and the key order array. This required changing quite a bit of test code, and changing how the token replacement works. Ordinarily I'd like to pull a change like this out as a separate commit, but I really wanted those guarantees as part of this change, and I was already too far invested when I realised to pull it out as a separate commit in any reasonable timeframe.

## After

zygote-after

See how the order in the UI matches that in the JSON file. Edits made in the text box are stable when committed, and the order is reflected correctly in JSON, meaning this round trips correctly.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8072)


drewnoakes:
@tmeschter we could build on this to introduce:

```c#
public interface ILaunchProfile2
{
    ImmutableArray<KeyValuePair<string, string>> EnvironmentVariables { get; }
    ImmutableArray<KeyValuePair<string, object>> OtherSettings { get; }
}
```

If a launch profile wants stable ordering of these values, they add this interface and we would only invoke these versions of the properties.

In 17.3 we may also wish to further extend `ILaunchProfile2` to support the feature request from #167.

```c#
public interface ILaunchProfile2
{
    ImmutableArray<KeyValuePair<string, string>> EnvironmentVariables { get; }
    ImmutableArray<KeyValuePair<string, object>> OtherSettings { get; }
    bool SkipLoadingReadyToRunImagesDuringDebugging { get; }
}
```

What do you think of this idea? Are you aware of anything else we may like to add to such an interface while we are at it?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfile.cs
```diff
@@ -1,14 +1,138 @@
...
+            {
+                // Preserve the order of keys
+                environmentVariablesKeyOrder = launchProfile.EnvironmentVariablesKeyOrder;
+                otherSettingsKeyOrder = launchProfile.OtherSettingsKeyOrder;
+            }
+            else
+            {
+                // We don't know the order here, so just use dictionary order (which will be randomised by hash codes)
+                environmentVariablesKeyOrder = environmentVariables?.Keys.ToImmutableArray() ?? ImmutableArray<string>.Empty;
+                otherSettingsKeyOrder = otherSettings?.Keys.ToImmutableArray() ?? ImmutableArray<string>.Empty;
```
drewnoakes, line 58: Looking again, these should probably sorted alphabetically for consistency with elsewhere.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/ActiveLaunchProfileEnvironmentVariableValueProvider.cs
```diff
@@ -47,92 +47,9 @@ public override bool SetPropertyValue(string propertyName, string value, IWritab
...
-            }
-
-            return string.Join(",", value.OrderBy(kvp => kvp.Key, StringComparer.Ordinal).Select(kvp => $"{encode(kvp.Key)}={encode(kvp.Value)}"));
-
-            static string encode(string value)
-            {
-                return value.Replace("/", "//").Replace(",", "/,").Replace("=", "/=");
-            }
-        }
-        private static void ParseStringIntoDictionary(string value, Dictionary<string, string> dictionary)
```
drewnoakes, line 69: This code appeared twice. Moved a single copy to a new class.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/LaunchProfiles/LaunchProfileProjectProperties.cs
```diff
@@ -378,90 +373,6 @@ public void SetRuleContext(Rule rule)
...
-
-            return string.Join(",", value.OrderBy(kvp => kvp.Key, StringComparer.Ordinal).Select(kvp => $"{encode(kvp.Key)}={encode(kvp.Value)}"));
-
-            static string encode(string value)
-            {
-                return value.Replace("/", "//").Replace(",", "/,").Replace("=", "/=");
-            }
-        }
-
-        private static void ParseStringIntoDictionary(string value, Dictionary<string, string> dictionary)
```
drewnoakes, line 396: Here's the other copy.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchProfileExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+            };
+        }
+
+        private static Dictionary<string, T> ToDictionary<T>(ImmutableDictionary<string, T> dic, ImmutableArray<string> keyOrder, IEqualityComparer<string> comparer)
+        {
+            var result = new Dictionary<string, T>(comparer);
+
+            foreach (string key in keyOrder)
+            {
+                result.Add(key, dic[key]);
```
tmeschter, line 74: Is your expectation that the dictionary's enumeration order will match the insertion order here? The documentation specifically calls out that the enumeration order is undefined.

tmeschter (changes requested):
In general I like the change, except for the dependency on `Dictionary<string, string>` preserving item order. That's an implementation detail and I am uneasy taking a dependency on that. Could we make use of [OrderedDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary) instead?

tmeschter:
@drewnoakes I like the idea of an `ILaunchProfile2` that preserves the ordering of environment variables and other settings. However, I'm not sure about combining that with a solution for #167 as I have further thoughts on it, which I'll raise on that issue.

tmeschter:
@drewnoakes Oh, and I think the `ActiveLaunchProfile*` types are probably all dead code. They date back to when we were still showing the _active_ launch profile in the new property pages. They shouldn't be used at all when using the Launch Profile UI. I need to make sure the WebTools team doesn't have any lingering dependencies on them and then they can be deleted.

drewnoakes:
> In general I like the change, except for the dependency on `Dictionary<string, string>` preserving item order. That's an implementation detail and I am uneasy taking a dependency on that. Could we make use of [OrderedDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary) instead?

OrderedDictionary will not help. It reorders items based on the key.

While the fact that Dictionary preserves insertion order is something that could theoretically change, there is no way that would ever happen. It would break far too many programs.

tmeschter:
> OrderedDictionary will not help. It reorders items based on the key.

That's [SortedDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sorteddictionary-2?view=net-6.0). From the [OrderedDictionary ](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?view=net-6.0) docs:

> The elements of an [OrderedDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?view=net-6.0) are not sorted by the key, unlike the elements of a [SortedDictionary<TKey,TValue>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sorteddictionary-2?view=net-6.0) class. You can access elements either by the key or by the index.

Also, it is not at all obvious that a `Dictionary` preserves order, or that we are using it specifically for its order-preserving properties. That's going to lead to maintenance problems down the line. Even if `Dictionary` works in practice, I'd rather a type that is explicit about the behavior we need.

drewnoakes:
My mistake. I shouldn't jump on the computer until I've woken up a little :)

I'll take a look to see whether using OrderedDictionary helps. There is a public API that uses Dictionary that may make that challenging.

It might be that the `ILaunchProfile2` approach is clearer overall.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/ILaunchProfileExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// it returns an <see cref="ImmutableDictionary{TKey, TValue}"/> which does not preserve order.
+        /// This extension method looks for internal state on <paramref name="profile"/> that allows
+        /// enumeration of environment variables in the same order they are specified in the
+        /// <c>launchSettings.json</c> file.
+        /// </remarks>
+        /// <param name="profile">The profile to read from.</param>
+        /// <returns>An ordered enumeration of environment variable name/value pairs.</returns>
+        public static IEnumerable<KeyValuePair<string, string>> EnumerateEnvironmentVariables(this ILaunchProfile profile)
+        {
+            if (profile.EnvironmentVariables is null)
```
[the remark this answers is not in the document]
drewnoakes: When I converted the others, this had a different implementation. It changed since then. I've made it a switch expression too, thanks.

tmeschter:
Yeah, I think the `ILaunchProfile2` approach might be best. That will also allow WebTools and others to opt into the new system.

drewnoakes:
I started on it yesterday and it feels like a good approach. I'll have an update for this early next week.

drewnoakes:
I spent some time on this and ended up reducing the number of concepts in this part of the code quite a bit. There's an `ILaunchProfile2` type now that preserves order.

There are fewer types now, and fewer constructors on those types. More stuff is immutable, and less copying occurs when we can determine that copies are not needed.

The diff is quite heavy. It may be easier to review commit by commit. Or perhaps pull it down, test it, and step through some operations. I've spent quite a bit of time testing it and am happy with it as a foundation for work towards #7330. I'll dust off my old branch. Now that the underlying data is stable, it should be easier to get that branch ready for review.

drewnoakes:
I'd also point out that it wasn't possible to completely rid ourselves of a dependency on the ordering behaviour of `Dictionary<,>`. Doing so would require versioning `IWritableLaunchSettings` and `IWritableLaunchProfile`.

drewnoakes:
I forgot to mention that I looked into `OrderedDictionary` and it's not a generic type. I don't think the loss of type safety is worth it in this case.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchProfileExtensions.cs
```diff
@@ -17,13 +19,30 @@ public static bool IsInMemoryObject(this object persistObject)
             return persistObject is IPersistOption profile2 && profile2.DoNotPersist;
         }
 
+        public static bool TryGetSetting(this ILaunchProfile profile, string key, [NotNullWhen(returnValue: true)] out object? value)
+        {
+            if (profile is ILaunchProfile2 lp2)
```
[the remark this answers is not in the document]
drewnoakes: Yes, thank you. I'll add a unit test for this, and also review other places where we special-case `ILaunchProfile2` to make sure nothing else was missed.

Issue #8074, opened 2022-04-13
Remove duplicate PackageReference items


tmeschter:
We have a number of duplicate PackageReference items, including:

- XliffTasks
- Microsoft.VisualStudio.Composition
- Microsoft.VSSDK.BuildTools

Recently NuGet has started warning about these duplicates, and the command line build (via build.cmd) elevates them to errors.

For the moment we'll suppress the warning (NU1504) until we can fix the duplicates.

MiYanni:
This may be fixed with the Infrastructure Simplification efforts. I've added it to that project list for review during that endeavor.

PR #8075, opened 2022-04-13
Suppress NU1504

tmeschter:
Related to #8074.

Warnings about duplicate PackageReference items are being promoted to errors in build.cmd and thus breaking our build. Here we suppress the warning until we can remove the duplicates.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8075)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -91,4 +91,10 @@
     <NoWarn>$(NoWarn);NU5125</NoWarn>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!--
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
Is it actually difficult to fix the duplicates?

tmeschter:
@drewnoakes Yes, I tried. One problem is that projects pull in different sets of .props files, so the PackageReference items are duplicated in some project but not others; removing one of the duplicates fixes the NU1504 error for one project but then causes other (sometimes silent) problems in a different project.

Issue #6920, opened 2021-02-03
Add a Project Query API action to explicitly delete a property


tmeschter:
The new project property pages allow you to set a property value to anything you want, and those changes will be persisted into the project file as an MSBuild property. However, we have no way to remove a property from the project file entirely.

This would be used by the new property pages to provide a gesture (a button or menu item) allowing the user to reset the value of a property to a default value by deleting it from the project file. Currently the best we can do is to set the property value to the empty string, but this won't work if the default value comes from a previous imported .props file--the empty string will "shadow" the default. And even where it does work, we're left with an empty property definition; this is aesthetically displeasing and runs counter to the lean design of SDK-style projects.

drewnoakes:
It may be possible to achieve this by passing `null` rather than requiring a new API.

adamint:
Null indicates already for an unevaluated value that it "should be removed entirely and/or reset to a default value" - because of this, I do not think it makes sense to have a sentinel value for evaluated and not for unevaluated and think that in the context of setting an evaluated property, it should mirror the unevaluated prop.

tmeschter:
We need to be careful about what level of the system we're talking about, as `null` means different things at different levels.

At the MSBuild level, reading a property that isn't defined will return `null` for the unevaluated value, and the empty string for the evaluated value. MSBuild will throw if you try to set the value to `null`. The MSBuild-specific implementation of IProjectProperties (the ProjectProperties type) reflects this. Note MSBuild does not make a distinction between setting evaluated vs unevaluated values.

Higher up at the `IProperty` level things get messier. `IProperty.GetValueAsync` can return `null` if the underlying data source returns `null` **or** the string value returned by the underlying data source cannot be converted to the expected type (bool, int, enum value, etc.). `IProperty.SetValueAsync` can accept a `null` value but what happens from there is undefined, and not consistent from one type of property to another. For bools and ints, `null` is effectively converted to the empty string. For strings and enum values, `null` will end up getting passed straight through for the data source to deal with. If that underlying data source is MSBuild, we'll get exceptions.

I think the current shape of the Project Query API (where we let you set the evaluated value to `null` but not the unevaluated) is a case of us exposing the semantics of the underlying layer without property thinking them through. We should do one of the following:

1. Have an explicit "delete" operation alongside "set evaluated value" and "set unevaluated value" that calls `IProperty.DeleteAsync`, and disallow `null` for either of the set operations.
[...]

Issue #6921, opened 2021-02-03
Add a UI gesture to reset a property to its default value


tmeschter:
This is the UX counterpart to #6920.

The new project property pages allow you to set a property value to anything you want, and those changes will be persisted into the project file as an MSBuild property. However, we have no way to reset a property to its default value. We should have a button or menu item to do so.

adamint:
Addressed in https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/393810

PR #8083, opened 2022-04-18
Allow null set on unevaluated property value

adamint:
Similar to evaluated property values, this allows null-set unevaluated property values. This does not preclude an empty unevaluated value from being set.
This is for #6920 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8083)

### DIFF
--- a/docs/repo/property-pages/back-end-architecture.md
+++ b/docs/repo/property-pages/back-end-architecture.md
@@ -34,6 +34,8 @@ The Project Query API treats its use of brokered services as an implementation d
 
 Internally, the Project Query API handles serializing the query, sending it across the network, deserializing it, and coordinating the actions of an extensible set of data providers. These data providers are responsible for filling in the requested properties on various entities, creating child entities from their parents, and handling various "actions" that update state in some way. For example, one provider may be responsible for creating project entities representing all the projects in the solution, while a different provider is responsible for filling in the project's name, path, GUID, language, etc. There may be multiple providers that handle the same relationship from a parent to its children or the same action. Among other things, the Project Query API implementation glues together the results of all of these providers to create the final query result.
 
+Please note that internally, setting either a property's evaluated or unevaluated value to null will call `IProject.DeleteAsync` to delete the property.
+
 ### 5. Standard query data providers
 
 CPS provides a number of built-in/standard query data providers. These generally expose data about the solution, projects, project items, references, project configuration, etc. We build off the data model that is populated by some of these providers but do not interact with them directly; no more will be said about them here.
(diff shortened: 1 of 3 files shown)


tmeschter (changes requested):
This is going to cause problems; see https://github.com/dotnet/project-system/issues/6920#issuecomment-1101650588.

adamint:
@tmeschter is this the way it should be done?

docs/repo/property-pages/back-end-architecture.md
```diff
@@ -34,6 +34,8 @@ The Project Query API treats its use of brokered services as an implementation d
 
 Internally, the Project Query API handles serializing the query, sending it across the network, deserializing it, and coordinating the actions of an extensible set of data providers. These data providers are responsible for filling in the requested properties on various entities, creating child entities from their parents, and handling various "actions" that update state in some way. For example, one provider may be responsible for creating project entities representing all the projects in the solution, while a different provider is responsible for filling in the project's name, path, GUID, language, etc. There may be multiple providers that handle the same relationship from a parent to its children or the same action. Among other things, the Project Query API implementation glues together the results of all of these providers to create the final query result.
 
+Please note that internally, setting either a property's evaluated or unevaluated value to null will call `IProject.DeleteAsync` to delete the property.
```
tmeschter, line 37: I think this information should go in the comments for the relevant types in the CPS system. This is, fundamentally, a change in the semantics of the Project Query API itself (though we currently have the only implementation of that part of the API).

tmeschter (approved):
Yeah, that's more what I had in mind.

docs/repo/property-pages/back-end-architecture.md
adamint: I will make sure to document this. Taking on #6921

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/ProjectSetUnevaluatedUIPropertyValueAction.cs
```diff
@@ -25,9 +25,9 @@ public ProjectSetUnevaluatedUIPropertyValueAction(SetUnevaluatedUIPropertyValue
 
         protected override Task SetValueAsync(IProperty property)
         {
-            if (property is IEvaluatedProperty evaluatedProperty)
+            if (_parameter.Value == null)
             {
-                return evaluatedProperty.SetUnevaluatedValueAsync(_parameter.Value ?? string.Empty);
+                return property.DeleteAsync();
```
drewnoakes, line 30: Why don't we need to call `SetUnevaluatedValueAsync` here anymore?

drewnoakes: Ah, I assume this is https://github.com/dotnet/project-system/issues/6920#issuecomment-1101569461

PR #8085, opened 2022-04-20
Pass the language service an initial approximation of command line options

jasonmalinowski:
This allows the language service to start parsing files correctly from the start, and also able to read caches sooner since we'll know the right inputs. This is the project system side of https://github.com/dotnet/roslyn/issues/60014

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8085)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
```diff
@@ -79,10 +79,15 @@ public async Task ReleaseProjectContextAsync(IWorkspaceProjectContextAccessor ac
...
                 try
                 {
-                    // Update LastDesignTimeBuildSucceeded within a batch to avoid thread pool starvation.
+                    // Update additional properties within a batch to avoid thread pool starvation.
                     // https://github.com/dotnet/project-system/issues/8027
 
                     context.LastDesignTimeBuildSucceeded = false;  // By default, turn off diagnostics until the first design time build succeeds for this project.
+
+                    // Pass along any early approximation we have of the command line options
+#pragma warning disable CS0618 // This was obsoleted in favor of the one that takes an array, but here just the string is easier; we'll un-Obsolete this API
```
jasonmalinowski, line 88: Because I'm using a property here rather than an item group, this overload that we obsoleted in favor of the one that takes an item group is easier to use. It wasn't clear to me how to grab an item group from evaluation in the rule and if that's easy to change I'm happy to do so; otherwise we'll just use this overload.

The reason this was obsoleted was because this string can be _very_ large in the case of the fully evaluated one (since it contains references and source file paths), and was ending up in the large object heap. But in this case, this string isn't nearly as large.

jasonmalinowski, line 88: Because I'm using a property here rather than an item group, this overload that we obsoleted in favor of the one that takes an item group is easier to use. It wasn't clear to me how to grab an item group from evaluation in the rule and if that's easy to change I'm happy to do so; otherwise we'll just use this overload.

The reason this was obsoleted was because this string can be _very_ large in the case of the fully evaluated one (since it contains references and source file paths), and was ending up in the large object heap. But in this case, this string isn't nearly as large.

jasonmalinowski:
@drewnoakes: I think I'd like to get this merged in now, even as we still sort out further changes in the SDK. Even if we aren't getting full parse options, at least getting the language version saves us a reparse on files that don't have #if directives, which is usually most of them. That'll be faster because of the optimizations I put in for branch switch.

drewnoakes:
@jasonmalinowski that sounds good to me. Please add test coverage for this new property, then we'll get this merged in.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/WorkspaceProjectContextProvider.cs
tmeschter: What does the string look like in this case?

jasonmalinowski: @tmeschter: good question! The Roslyn side of the change is https://github.com/dotnet/roslyn/pull/60855, so you are basically getting `-langver:... -define:...`

jasonmalinowski:
@drewnoakes: where would tests go in this case?

drewnoakes:
[`ProjectFilePathAndDisplayNameEvaluationHandlerTests`](https://github.com/dotnet/project-system/blob/debe514a304a3e158e33b7a3aaa27c40c2721c69/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/Handlers/ProjectFilePathAndDisplayNameEvaluationHandlerTests.cs) is an example of such a test, for the `MSBuildProjectFullPath` property.

jasonmalinowski:
@drewnoakes: hmm, are those tests different though? Unless I'm misreading it, those are for testing that we correctly propagate changes. In this case, any changes to this property isn't propagated at all -- it's a one time shot only for the _creation_ of the context in the first place.

drewnoakes:
@jasonmalinowski I pushed a commit to your branch that adds unit tests.

drewnoakes (approved):

PR #8096, opened 2022-04-21
Reduce allocations during global property construction

drewnoakes:
Constructs and reuses immutable collection objects across repeated calls.

Avoids async state machines in some cases. Caches tasks where possible.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8096)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/StartupProjectSingleTargetGlobalBuildPropertyProvider.cs
```diff
@@ -43,10 +46,8 @@ internal class StartupProjectSingleTargetGlobalBuildPropertyProvider : StaticGlo
             _projectSystemOptions = projectSystemOptions;
         }
 
-        public override async Task<IImmutableDictionary<string, string>> GetGlobalPropertiesAsync(CancellationToken cancellationToken)
+        public override Task<IImmutableDictionary<string, string>> GetGlobalPropertiesAsync(CancellationToken cancellationToken)
```
melytc, line 49: Does it still need the `Async` suffix?

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Build/StartupProjectSingleTargetGlobalBuildPropertyProvider.cs
drewnoakes: Good question.

The `Async` suffix advertises to the caller that they should await the return value. We still return a `Task<>` here, so the `Async` suffix still applies.

The `async` modifier controls whether the method's implementation needs to compiler to produce a state machine. This method was changed to avoid any `await`s, meaning we don't need the `async` modifier.

melytc: Oh, I see, thanks for the clarification!

PR #8097, opened 2022-04-21
Prevent a .csproj.user file being created during build

drewnoakes:
We missed this item update in our .csproj file. As a consequence, for a clean clone of our repo, CPS creates a .user file that contains this metadata for the project. This means the project becomes immediately out-of-date, and we report an incremental build failure.

Adding this into the .csproj file fixes this issue.

The `.csproj.user` file has the following content:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup />
  <ItemGroup>
    <Compile Update="PropertyPages\PropertyPageControl.cs">
    </Compile>
  </ItemGroup>
</Project>
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8097)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
@@ -40,6 +40,9 @@
     <Compile Update="ProjectSystem\VS\LanguageServices\VisualBasic\VisualBasicCodeDomProvider.cs">
       <SubType>Component</SubType>
     </Compile>
+    <Compile Update="ProjectSystem\VS\PropertyPages\DebugPropertyPage.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
     <Compile Update="ProjectSystem\VS\PropertyPages\DebugPageControl.xaml.cs">
       <DependentUpon>DebugPageControl.xaml</DependentUpon>
     </Compile>


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
```diff
@@ -40,6 +40,9 @@
     <Compile Update="ProjectSystem\VS\LanguageServices\VisualBasic\VisualBasicCodeDomProvider.cs">
       <SubType>Component</SubType>
     </Compile>
+    <Compile Update="ProjectSystem\VS\PropertyPages\DebugPropertyPage.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
```
MiYanni, line 45: If I understand this correctly, the file itself would be `Included` by globbing. Here, you're doing `Update` as to add the `SubType` to it. Why is that needed?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
drewnoakes: That's right. The `Include` comes from globs. The `Update` here is generated by the IDE once it learns that this source file is actually a WinForms `UserControl` subtype.

I believe the metadata is needed within the IDE to support designers. I assume it's written into the project file for performance, though I don't know why it only happens during build. I'll schedule this as a question for next weeks sync with the CPS team.

MiYanni: I saw the details you put in the CPS meeting channel. <3 This feels like a mechanic that made sense in non-SDK-style projects, but since *most* compile items are included via globbing now, it is adding these compile items back into the .csproj as `Update`, which is kind of an anti-pattern in SDK-style projects. If we get information on `SubType`, maybe we can make some kind of simplification here on the CPS side of things.

drewnoakes: For sure. I'll be interested to discuss this next week. Thanks for calling it out.

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
drewnoakes: Closing the loop here with insights from Lifeng. SubType is indeed used to support designers. The data is not being added during build -- rather it's being added during the language service's code analysis. Code analysis looks for attributes on a type's ancestors to determine the subtype.

There is a capability to control whether `SubType` is stored in the `.csproj` or `.csproj.user` file. The .NET Project System opts for the `.csproj.user` file to avoid bloating the `.csproj`.

Issue #8036, opened 2022-04-04
Create dev17.2.x branch checklist


drewnoakes:
_Descriptions of these steps can be found in the team OneNote._

- [x] Identify base commit for new branch
  - [x] According to the [schedule](https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/10097/Dev17-Release), `main` snapped to `rel/d17.2` on 1 April 2022 at 4PM PT
  - [x] Last [completed Project System insertion PR](https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequests?_a=completed&createdBy=9f64bc2f-479b-429f-a665-fec80e130b1f&assignedTo=6e89082d-fdd2-4442-a310-051df5bdc73c): https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/390040
  - [x] Last inserted GitHub PR: https://github.com/dotnet/project-system/pull/8021
  - [x] Merge commit of that PR: 2d8184bf78a11568ae502597ed5ba6d24bdd736c
- [x] Create branch on GitHub at that commit
  - [x] https://github.com/dotnet/project-system/tree/dev17.2.x
- [ ] ~For branches not matching `dev*` or `feature/*` (usually we skip these)~
  - [ ] ~Update the [YAML file](https://github.com/dotnet/project-system/blob/main/build/ci/unit-tests.yml) to support CI/PR builds~
  - [ ] ~Update the [signed build definition](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=9675) to build the new branch~
- [x] Update Roslyn Tools [config.xml](https://github.com/dotnet/roslyn-tools/blob/main/src/GitHubCreateMergePRs/config.xml) file to flow branch changes to the latest dev branch
  - [x] dotnet/roslyn-tools PR: https://github.com/dotnet/roslyn-tools/pull/1169
- [x] Update `Versions.props` so `<ProjectSystemVersion>` matches the version of VS, if needed
    - [x] In new branch: https://github.com/dotnet/project-system/blob/dev17.2.x/build/import/Versions.props
    - [x] In `main`: https://github.com/dotnet/project-system/blob/main/build/import/Versions.props
- [ ] Clone existing release definition to insert this branch into VS (see OneNote)
- [x] Update [README.md](https://github.com/dotnet/project-system/blob/main/README.md) (in `main`) if we need new badges
[...]

PR #8100, opened 2022-04-22
[17.2] Don't load VC++ projects during build

drewnoakes:
Fixes [AB#1527166](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1527166)

Two components in the .NET Project System listen to build events. When notified about a project's build progress, they are passed `IVsHierarchy` objects.

Previously we used `AsUnconfiguredProject` to obtain the project from a hierarchy, however when this method is invoked on a VC++ project it forces that project to fully load. This is a performance regression for mixed .NET/VC++ solutions.

With this change, we use `IProjectService2.GetLoadedProject` to obtain the project. CPS won't make any operation against an external project system in this case, so there's no chance of inadvertently loading any unloaded components.

We also validate that the project has the required capabilities for the calling component.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8100)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectServiceExtensions.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectServiceExtensions.cs
@@ -1,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Shell.Interop;
+using static Microsoft.VisualStudio.VSConstants;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS
+{
+    /// <summary>
+    /// Extension methods for <see cref="IProjectService"/>.
+    /// </summary>
+    internal static class IProjectServiceExtensions
+    {
+        /// <summary>
+        /// Obtains the <see cref="UnconfiguredProject"/> for <paramref name="vsHierarchy"/>,
+        /// optionally testing whether it has a capability match to <paramref name="appliesToExpression"/>.
+        /// </summary>
+        /// <param name="projectService">The project service to invoke this method on.</param>
+        /// <param name="vsHierarchy">The VS hierarchy object, which must represent a CPS project.</param>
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectServiceExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            // We need IProjectService2.GetLoadedProject.
+            if (projectService is IProjectService2 projectService2)
+            {
+                vsHierarchy.GetCanonicalName((uint)VSITEMID.Root, out string? projectFilePath);
+
+                // Ensure we have a path.
+                if (!Strings.IsNullOrEmpty(projectFilePath))
+                {
+                    UnconfiguredProject? unconfiguredProject = projectService2.GetLoadedProject(projectFilePath);
```
tmeschter, line 34: This is the critical part, right? This will only return the `UnconfiguredProject` if it is _already_ loaded?

tmeschter (approved):

jacdavis (approved):
:shipit:

MiYanni (commented):
We may need to update the dev17.2.x branch to be in the state of which we last made changes in VS's 17.2 branch. I'd definitely align those prior to merging this.

drewnoakes:
> We may need to update the dev17.2.x branch to be in the state of which we last made changes in VS's 17.2 branch. I'd definitely align those prior to merging this.

This was done when VS snapped for 17.2 in #8036.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/IProjectServiceExtensions.cs
drewnoakes: Yes. And importantly it happens without invoking any code on the hierarchy other than querying for the canonical name. This makes it very unlikely we inadvertently trigger any code paths that force loads.

drewnoakes:
Insertion PR for 17.2 is here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/395039

PR #8103, opened 2022-04-22
Localized file check-in by OneLocBuild Task: Build definition ID 981: Build ID 1733672

MiYanni:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8103)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/PackagePropertyPage.xaml.it.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/PackagePropertyPage.xaml.it.xlf
@@ -14,12 +14,12 @@
       </trans-unit>
       <trans-unit id="BoolProperty|IncludeSymbols|Description">
         <source>Create an additional symbol package when the project is packed.</source>
-        <target state="new">Create an additional symbol package when the project is packed.</target>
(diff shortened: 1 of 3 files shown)


drewnoakes:
@MiYanni this appears to be an exact dupe of #8102.

MiYanni:
@drewnoakes When you see something like this, look at the build ID. This one here is 1733672. The one you linked is 1733665.
- Build definition ID 981 refers to: [https://dnceng.visualstudio.com/public/_build?**definitionId=981**](https://dnceng.visualstudio.com/public/_build?definitionId=981)
- Build ID 1733672 refers to: [https://dnceng.visualstudio.com/public/_build/results?**buildId=1733672**](https://dnceng.visualstudio.com/public/_build/results?buildId=1733672)
- Build ID 1733665 refers to: [https://dnceng.visualstudio.com/public/_build/results?**buildId=1733665**](https://dnceng.visualstudio.com/public/_build/results?buildId=1733665)
- Looks like both builds were kicked off at the same time: image

I believe the system is working as intended as there's no real way for it to know that both builds would produce the same results. Them being kicked off at the same time could only result in creating 2 PRs. The question I have, if this PR is a dupe of that one, why are there any changes in this PR at all? Meaning, "files changed" should be zero since https://github.com/dotnet/project-system/pull/8102 is already merged.

tmeschter (approved):

PR #8112, opened 2022-04-26
Only use design-time build data for source files in MAUI projects

tmeschter:
Most data we pass to the Roslyn language service (i.e. references, additional files, etc.) comes from design-time builds and only design-time builds. This is to ensure the data we send to the LS matches the data sent to the compiler during a normal build.

However, for source files specifically we use both evaluation and design-time build data. This is because the evaluation data is available sooner than the build data, and this allows us to populate Roslyn sooner during a project load or when a file is added to the project.

Evaluation and design-time data can differ on the set of source files, so we do extra work to reconcile the differences. This mostly works, except for the case where a design-time build removes a file present in evaluation. To properly reconcile that particular difference we would need to compare the results of a design-time build against the corresponding evaluation; however, we are not guaranteed to see the results of every evaluation or every design-time build. We could easily end up in a situation where we see a file "added" by evaluation, but never see a design-time build that "removes" it, and we will tell Roslyn about files that aren't actually part of the project. This will lead to problems like spurious errors in the IDE that can't be fixed, or a failure to report errors that are really there.

MAUI projects have a bunch of files that are platform-specific. Evaluation includes the files for all target frameworks, and then MAUI-specific targets remove the ones that _don't_ apply to a particular framework. This falls right into the case we don't support well, and leads to multiple problems:

1. Platform-specific files incorrectly show all target frameworks in the Nav Bar.
2. Spurious errors in the Error List.
3. Squiggles in the editor, unless they change the Nav Bar to the "correct" target framework.
4. Possibly they will not see actual errors in their code (e.g. if you use a Windows-only API in an Android-specific file, but the Nav Bar is set to Windows, you won't get an squiggle about that).
5. The spurious errors block Hot Reload as the language service thinks the project is very broken.

To avoid this problem, this change effectively opts MAUI out of evaluation data for `Compile` items. In MAUI Single Project scenarios the existing `CompileItemHandler` will be replaced by a new `MauiSingleProjectCompileItemHandler`. This handler does not receive any information about evaluations, and only uses design-time build data (as we already do for additional files, references, etc.). As such there is no need to reconcile the two sources of information.

The downside to this is that a newly-added file, or a file that opens automatically when the project loads, may be associated with the "Miscellaneous" project for a short time instead of the correct project. Most language service features (proper colorization, Intellisense, refactorings) will be limited or unavailable. This will correct itself automatically once the design-time builds completes and is processed.

The biggest part of this change involves making `IWorkspaceContextHandler` implementations and consumers capability-aware so we can limit `MauiSingleProjectCompileItemHandler` to projects with the MauiSingleProject capability. Currently the `ApplyChangesToWorkspaceContext` types pulls in all handlers regardless of capability. Fixing this requires adding an `AppliesTo` attribute to all implementations, and updating `ApplyChangesToWorkspaceContext`  to filter the implementations based on the project's current capabilities.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8112)


tmeschter:
Closing in favor of #8112.

PR #8115, opened 2022-04-29
Ignore certain items from evaluation

tmeschter:
Related to [dotnet/maui #6681](https://github.com/dotnet/maui/pull/6681).
Related to [AB#1525249](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1525249).

This is an alternative to the approach in #8112.

Currently MAUI includes platform-specific files are for every platform in the MSBuild evaluation, and then at build a target removes _all_ these items and then adds back the ones for the current platform.

This works for an actual build but screws up VS, as the Project System doesn't properly handle a `Compile` item that is present in evaluation but removed during a build. The end result is that the platform-specific files are not actually treated as platform-specific, leading to a number of user-visible issues:

1. Platform-specific files incorrectly show all target frameworks in the Nav Bar
2. Spurious errors in the Error List
3. Squiggles in the editor, unless the user changes the Nav Bar to the "correct" target framework
4. Possibly they will not see actual errors in their code (e.g. if you use a Windows-only API in an Android-specific file, but the Nav Bar is set to Windows, you won't get an squiggle about that)
5. The spurious errors block Hot Reload as the Language Service thinks the project is very broken and will not attempt to apply changes

With this change the Project System will now check items for special metadata indicating that they should be ignored. If an item is marked with `ExcludeFromCurrentConfiguration=true` the Project System will pretend it does not exist in the evaluation data for the current build configuration. And as the file no longer exists in the evaluation data, it no longer matters that the Project System does not properly handle the removal in the target.

See [dotnet/maui #6681](https://github.com/dotnet/maui/pull/6681) for the related MAUI SDK changes that add the metadata to the appropriate items.

Note that this does not currently handle the case where the `ExcludeFromCurrentConfiguration` metadata changes on an item. That is a change we can make in the future, if need be.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8115)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.ProjectSystem.VS;
 
 namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
@@ -159,8 +160,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
         private bool IsItemInCurrentConfiguration(string includePath, IImmutableDictionary<string, IImmutableDictionary<string, string>> metadata)
         {
             if (metadata.TryGetValue(includePath, out IImmutableDictionary<string, string> itemMetadata)
-                && itemMetadata.TryGetValue("ExcludeFromCurrentConfiguration", out string excludeFromCurrentConfiguration)
-                && StringComparers.PropertyValues.Equals(excludeFromCurrentConfiguration, "true"))
+                && itemMetadata.GetBoolProperty(Compile.ExcludeFromCurrentConfigurationProperty) is true)
             {
                 return false;
             }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
@@ -156,6 +156,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
 
         protected abstract void UpdateInContext(string fullPath, IImmutableDictionary<string, string> previousMetadata, IImmutableDictionary<string, string> currentMetadata, bool isActiveContext, IProjectDiagnosticOutputService logger);
 
+        private bool ItemInCurrentConfiguration(string includePath, IImmutableDictionary<string, IImmutableDictionary<string, string>> metadata)
+        {
+            if (metadata.TryGetValue(includePath, out IImmutableDictionary<string, string> itemMetadata)
+                && itemMetadata.TryGetValue("ExcludeFromCurrentConfiguration", out string excludeFromCurrentConfiguration)
+                && StringComparers.PropertyValues.Equals(excludeFromCurrentConfiguration, "true"))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
         private void ApplyChangesToContext(IProjectChangeDiff difference, IImmutableDictionary<string, IImmutableDictionary<string, string>> previousMetadata, IImmutableDictionary<string, IImmutableDictionary<string, string>> currentMetadata, bool isActiveContext, IProjectDiagnosticOutputService logger, bool evaluation)
         {
             foreach (string includePath in difference.RemovedItems)
@@ -165,6 +177,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices.Handlers
 
             foreach (string includePath in difference.AddedItems)
             {
+                if (evaluation && !ItemInCurrentConfiguration(includePath, currentMetadata))
+                {
+                    // The item is present in evaluation but contains metadata indicating it should be
+                    // ignored.
+                    continue;
+                }
+
                 AddToContextIfNotPresent(includePath, currentMetadata, isActiveContext, logger);
             }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
```diff
@@ -156,6 +156,18 @@ public void ApplyProjectBuild(IComparable version, IProjectChangeDiff difference
 
         protected abstract void UpdateInContext(string fullPath, IImmutableDictionary<string, string> previousMetadata, IImmutableDictionary<string, string> currentMetadata, bool isActiveContext, IProjectDiagnosticOutputService logger);
 
+        private bool ItemInCurrentConfiguration(string includePath, IImmutableDictionary<string, IImmutableDictionary<string, string>> metadata)
+        {
+            if (metadata.TryGetValue(includePath, out IImmutableDictionary<string, string> itemMetadata)
+                && itemMetadata.TryGetValue("ExcludeFromCurrentConfiguration", out string excludeFromCurrentConfiguration)
+                && StringComparers.PropertyValues.Equals(excludeFromCurrentConfiguration, "true"))
```
drewnoakes, line 163: I think we can use `CompileSchema.ExcludeFromCurrentConfigurationProperty` here to avoid the literal string here. There's also a `GetBoolProperty` helper method, from memory.

Something like:

```suggestion
            if (metadata.TryGetValue(includePath, out IImmutableDictionary<string, string> itemMetadata)
                && itemMetadata.GetBoolProperty(CompileSchema.ExcludeFromCurrentConfigurationProperty) is true)
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandlerTests.cs
```diff
@@ -217,6 +217,20 @@ public void AddEvaluationChanges_CanAddItemWithMetadata()
...
         }
 
+        [Fact]
+        public void AddEvaluationChanges_ItemsWithExclusionMetadataAreIgnored()
+        {
+            var handler = CreateInstance(@"C:\Project\Project.csproj");
+
+            var difference = IProjectChangeDiffFactory.WithAddedItems("A.cs;B.cs");
+            var metadata = MetadataFactory.Create("A.cs", ("ExcludeFromCurrentConfiguration", "true"))
+                                          .Add("B,cs", ("ExcludeFromCurrentConfiguration", "false"));
```
drewnoakes, line 227: Is the comma here a typo? If so, is this test working as expected?

```suggestion
                                          .Add("B.cs", ("ExcludeFromCurrentConfiguration", "false"));
```

drewnoakes: Maybe add a third item `C.cs` with no metadata to explicitly validate behaviour when no metadata is present.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
```diff
@@ -176,6 +195,9 @@ private void ApplyChangesToContext(IProjectChangeDiff difference, IImmutableDict
                 foreach (string includePath in difference.ChangedItems)
                 {
                     UpdateInContextIfPresent(includePath, previousMetadata, currentMetadata, isActiveContext, logger);
+
+                    // TODO: Check for changes in the metadata indicating if we should ignore the file
+                    // in the current configuration.
```
MiYanni, line 200: FINALLY! I understand. I've asked before if there is some kind of metric used for doing the line breaks on these comments. It seems y'all have a setting that does line breaks at 80 characters. Where is that setting and is there a reason to use 80? Many lines of code are much wider than 80 characters. This method definition is 313 characters wide. Seems like 80 is some kind of archaic value prior to the advent of widescreen monitors.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
```diff
@@ -156,6 +156,18 @@ public void ApplyProjectBuild(IComparable version, IProjectChangeDiff difference
 
         protected abstract void UpdateInContext(string fullPath, IImmutableDictionary<string, string> previousMetadata, IImmutableDictionary<string, string> currentMetadata, bool isActiveContext, IProjectDiagnosticOutputService logger);
 
+        private bool ItemInCurrentConfiguration(string includePath, IImmutableDictionary<string, IImmutableDictionary<string, string>> metadata)
```
MiYanni, line 159: nit: use `IsItemInCurrentConfiguration`

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Handlers/AbstractEvaluationCommandLineHandler.cs
tmeschter: I just manually wrap them at 80. Or sometimes 100. Depends on how I'm feeling that day. :-) I find comment widths in that range to be easier to read.

But we're not really consistent about comment widths.

MiYanni: Oh, interesting. I tend to use English language logic for comments. Meaning, if I have several sentences of comments, I'll line break to separate concepts, similar to how paragraphs are used in writing. Otherwise, it looks disorienting to have code lines hundreds of characters wide with skinny comment blocks, IMO. I like line breaks to signify significance, similar to what they do in code (in many cases).

jacdavis: I do the samething as yanni. I line break whenever the concept changes or when it starts to look too wide. I don't use any specific width of line breaks anymore than I do for code.

tmeschter:
Retargeted to dev17.3-preview1.

melytc (approved):

PR #8084, opened 2022-04-19
Restore the editing experience for environment variables in launch profiles

drewnoakes:
Fixes #7330

When we moved launch profiles to the new UI, we regressed the editing experience for environment variables. Previously we had an editable grid of values:

image

With the update to the new UI, this was collapsed into a single string with specific rules around encoding and escaping:

image

CPS PR https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/394046 adds a new editor type for grid-based editing of properties.

This change configures the launch profiles UI to use that new editor for environment variable properties.

Here's how the whole feature works, end to end:

environment-variable-grid

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8084)

PR #8126, opened 2022-05-03
Document use of NameValueList property editor

drewnoakes:
We recently added support for name/value pair editing (https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/394046). We started using this for launch profiles in https://github.com/dotnet/project-system/pull/8084. This commit documents the new capability so that others can more easily discover and use it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8126)

### DIFF
--- a/docs/repo/property-pages/property-specification.md
+++ b/docs/repo/property-pages/property-specification.md
@@ -195,6 +195,24 @@ We use this on the `LangVersion` property, for example, as this value is intenti
 </StringProperty>
 ```
 
+### Name/Value List
+
+When a property contains a variable number of name/value pairs, you can use the `NameValueList` editor on `StringProperty` to display a two-column grid in the UI that allows users to edit values and add/remove rows.
+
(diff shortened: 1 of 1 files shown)


docs/repo/property-pages/property-specification.md
```diff
@@ -195,6 +195,24 @@ We use this on the `LangVersion` property, for example, as this value is intenti
 </StringProperty>
 ```
 
+### Name/Value List
+
+When a property contains a variable number of name/value pairs, you can use the `NameValueList` editor on `StringProperty` to display a two-column grid in the UI that allows users to edit values and add/remove rows.
+
+The property's string value should be encoded with format resembling `A=1,B=2`, using `/` as an escape character if needed. See `LaunchProfileEnvironmentVariableEncoding` in this repo for further details.
```
MiYanni, line 202: Forward slash (not backslash) is the escape character?

MiYanni (approved):

docs/repo/property-pages/property-specification.md
drewnoakes: That's right, in this case.

PR #8127, opened 2022-05-03
Add lock to protect s_listOfMissingRuntimes.

ocallesp:
Fixes [AB#1523369](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1523369) threadpool starvation due to incorrect synchronization.

The issue here is that s_listOfMissingRuntimes was not thread-safe.

This PR:
- Renames `s_listOfMissingRuntimes` to `_missingRuntimesRegistered `
- Changes the type to use a thread-safe collection.
- Moves `_missingRuntimesRegistered`  to `MissingSetupComponentRegistrationService` that fits the lifetime of this global variable.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8127)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -92,17 +92,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             UnregisterProjectConfiguration(projectGuid, project);
         }
 
-        public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string runtimeComponentId)
+        public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string? runtimeComponentId)
         {
-            // If a runtime is detected as missing, it will remain in that state until the user install it using the VS Installer.
-            // Due to this, during a solution load all the projects might register to install the same runtime version.
-            // This optimization can avoid executing displayMissingComponentsTask unnecesarily by registering a missing runtime once in the solution.
-            if (!_missingRuntimesRegistered.Add(runtimeComponentId))
+            if (runtimeComponentId is not null)
             {
-                return;
-            }
+                _missingRuntimesRegistered.Add(runtimeComponentId);
 
-            _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, runtimeComponentId);
+                _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, runtimeComponentId);
+            }
 
             UnregisterProjectConfiguration(projectGuid, project);
         }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -21,6 +21,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         private static readonly ImmutableHashSet<string> s_supportedReleaseChannelWorkloads = ImmutableHashSet.Create(StringComparers.WorkloadNames, WasmToolsWorkloadName);
 
+        private readonly ConcurrentHashSet<string> _missingRuntimesRegistered = new(StringComparers.WorkloadNames);
         private readonly ConcurrentDictionary<Guid, IConcurrentHashSet<WorkloadDescriptor>> _projectGuidToWorkloadDescriptorsMap;
         private readonly ConcurrentDictionary<Guid, string> _projectGuidToRuntimeDescriptorMap;
         private readonly ConcurrentDictionary<Guid, IConcurrentHashSet<ProjectConfiguration>> _projectGuidToProjectConfigurationsMap;
@@ -73,6 +74,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         private void ClearMissingWorkloadMetadata()
         {
+            _missingRuntimesRegistered.Clear();
             _projectGuidToRuntimeDescriptorMap.Clear();
             _projectGuidToWorkloadDescriptorsMap.Clear();
             _projectGuidToProjectConfigurationsMap.Clear();
@@ -92,6 +94,14 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string runtimeComponentId)
         {
+            // If a runtime is detected as missing, it will remain in that state until the user install it using the VS Installer.
+            // Due to this, during a solution load all the projects might register to install the same runtime version.
+            // This optimization can avoid executing displayMissingComponentsTask unnecesarily by registering a missing runtime once in the solution.
+            if (!_missingRuntimesRegistered.Add(runtimeComponentId))
+            {
+                return;
+            }
+
             _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, runtimeComponentId);
 
             UnregisterProjectConfiguration(projectGuid, project);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -92,7 +92,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             UnregisterProjectConfiguration(projectGuid, project);
         }
 
-        public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string? runtimeComponentId)
+        public void RegisterSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string? runtimeComponentId)
         {
             if (runtimeComponentId is not null)
             {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
@@ -17,7 +17,6 @@ namespace Microsoft.VisualStudio.ProjectSystem
 
         private Guid _projectGuid;
         private bool _enabled;
-        private static readonly HashSet<string> s_listOfMissingRuntimes = new(StringComparer.OrdinalIgnoreCase);
 
         private readonly ConfiguredProject _project;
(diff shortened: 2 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
```diff
@@ -49,7 +50,11 @@ public Task UnloadAsync()
 
         protected override Task DisposeCoreAsync(bool initialized)
         {
-            s_listOfMissingRuntimes.Clear();
+            lock (s_listOfMissingRuntimesLock)
```
davkean, line 53: This component lives per project, yet clears a solution-wide collection?

ocallesp: It checks all projects, but if all projects are missing Runtime 5.0 it will only register once in the solution.

davkean: As in, let's pretend this solution has 500 projects. This lock will get hit 500 times to attempt to clear a collection that is already empty after the first call. Why do we need the other 499 calls?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
```diff
@@ -17,6 +17,7 @@ internal class MissingSdkRuntimeDetector : OnceInitializedOnceDisposedAsync, IPr
 
         private Guid _projectGuid;
         private bool _enabled;
+        private static readonly object s_listOfMissingRuntimesLock = new();
```
drewnoakes, line 20: We don't need a new object here to lock on. `s_listOfMissingRuntimes` doesn't change its value, so you can lock on that directly.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
drewnoakes: Yes this is broken. This was likely trying to clean up on solution close, but it'll do the wrong thing when a project is unloaded or reloaded, for example.

It looks like this component is trying to de-duplicate data for `IMissingSetupComponentRegistrationService`. That component should do its own de-duplication, and all state in this `MissingSdkRuntimeDetector` should be tied to its own lifetime.

drewnoakes:
This fixes the corruption, but it has also highlighted a correctness issue that should be addressed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
ocallesp: @drewnoakes I did the suggested changes. 
The de-duplication was moved to `MissingSdkRuntimeDetector`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
ocallesp: HashSet was replaced with ConcurrentDictionary

ocallesp: HashSet was replaced with ConcurrentDictionary

ocallesp:
Will do a git squash when ready to merge.

ocallesp:
Will do a git squash when ready to merge.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -92,6 +95,13 @@ public void RegisterMissingWorkloads(Guid projectGuid, ConfiguredProject project
 
         public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string runtimeComponentId)
         {
+            if (_missingRuntimesRegistered.ContainsKey(runtimeComponentId))
+            {
+                return;
+            }
+
+            _missingRuntimesRegistered.TryAdd(runtimeComponentId, 0);
```
drewnoakes, line 103:
```suggestion
            if (!_missingRuntimesRegistered.TryAdd(runtimeComponentId, 0))
            {
                return;
            }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
drewnoakes: What was the reason for changing it?

ocallesp: It is thread-safe

drewnoakes: So is a locked dictionary. But I see now that the class you've moved this logic too also uses `ConcurrentDictionary<,>` so it makes sense to keep it consistent.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -19,6 +19,8 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
     {
         private const string WasmToolsWorkloadName = "wasm-tools";
 
+        private readonly ConcurrentDictionary<string, int> _missingRuntimesRegistered = new();
```
drewnoakes, line 22: Use `ConcurrentHashSet<T>` instead. You don't need to store values here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -92,6 +95,11 @@ public void RegisterMissingWorkloads(Guid projectGuid, ConfiguredProject project
 
         public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string runtimeComponentId)
         {
+            if (!_missingRuntimesRegistered.TryAdd(runtimeComponentId, 0))
+            {
+                return;
+            }
+
```
drewnoakes, line 102: Do we even need to track this? What is the downside to calling this method more than once.

As it stands, this early return means we only clean up the first project to call for a given `runtimeComponentId`.

Is it important to store all project GUIDs here? `UnregisterProjectConfiguration` might also need to be called for more than just the first caller.

drewnoakes (commented):
This PR moves the tracking of missing runtimes from `MissingSdkRuntimeDetect` to `MissingSetupComponentRegistrationService`. The latter component is in the global scope, so can store this state in non-static fields. It clears out all its state on solution close.

The ownership of this state has moved between components with different scopes. The original was ConfiguredProject scope, the new is Global scope. However the original component stored that state in static fields, so it was effectively global before too.

I do wonder whether we need this early return at all. It might actually be masking a bug. It's hard to say by looking at the code. If it is important, we should document why above the return, so that it's clear in future. This code is not easy to comprehend quickly when you're unfamiliar with it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -19,6 +19,8 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
     {
         private const string WasmToolsWorkloadName = "wasm-tools";
 
+        private readonly ConcurrentDictionary<string, int> _missingRuntimesRegistered = new();
+        private readonly ConcurrentHashSet<string> n = new();
```
drewnoakes, line 23: I think this was pushed by mistake.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
ocallesp: We want to avoid calling the Setup apis.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
```diff
@@ -73,14 +70,22 @@ private async Task RegisterSdkRuntimeNeededInProjectAsync(ConfiguredProject proj
 
                 string? TargetFrameworkVersion = await configuration.TargetFrameworkVersion.GetDisplayValueAsync();
```
MiYanni, line 71: nit: The two variables above start with a capital letter, but are method variables. That should start with a lowercase.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
```diff
@@ -73,14 +70,22 @@ private async Task RegisterSdkRuntimeNeededInProjectAsync(ConfiguredProject proj
...
 
+                string? componentId = GetComponentId(TargetFrameworkIdentifier, TargetFrameworkVersion);
+                
+                _missingSetupComponentRegistrationService.RegisterSdkRuntimeComponentId(_projectGuid, project, componentId);
+            }
+
+            static string? GetComponentId(string TargetFrameworkIdentifier, string TargetFrameworkVersion)
```
MiYanni, line 78: 2 things:
- Why are the parameters changing from nullable to non-nullable (variables above the `GetComponentId` call are nullable but the parameters are non-nullable)?
- nit: Function parameters should start with a lowercase letter.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
ocallesp: `RegisterSdkRuntimeNeededInProjectAsync()` used to skip the execution of `RegisterSdkRuntimeComponentId()` when there is no valid runtime. Now we always call `RegisterSdkRuntimeComponentId()` and use null to say that there is no valid runtime is found.

PR #8142, opened 2022-05-06
Add documentation on property value interception

drewnoakes:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8142)

### DIFF
--- a/docs/repo/property-pages/property-value-interception.md
+++ b/docs/repo/property-pages/property-value-interception.md
@@ -2,6 +2,11 @@
 
 Sometimes it is necessary to intercept reads and writes of property values, and modify those values while they are in flight. Property value interception allows this.
 
+Examples of why you may wish to do this:
+
+- You wish to [provide a default value](#providing-default-values) for the property in the UI, when no value is provided in props/targets.
+- The property is a [pseudo property](#pseudo-properties) and should not be written to the project. Its values may be ignored, or may actually be derived from and modify other properties instead.
+
 ## Mechanism
 
 The property must opt in to interception. Its `DataSource` must specify the `Persistence` property as one of:
@@ -44,6 +49,8 @@ internal sealed class MyPropertyValueProvider : InterceptingPropertyValueProvide
 
 The name of the class is not important, but you might like to follow the convention offered here.
 
+A single interceptor class may be used to intercept multiple properties. To do that, add one `ExportInterceptingPropertyValueProvider` attribute per property you wish to intercept, changing the name on each accordingly.
+
 ## Base classes
 
 While you can implement `IInterceptingPropertyValueProvider` directly, you may like to use one of these base types.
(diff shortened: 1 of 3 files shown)


docs/repo/property-pages/property-value-interception.md
```diff
@@ -0,0 +1,73 @@
...
+  ...
+</StringProperty>
+```
+
+> See [Property Specification](property-specification.md) for more information on specifying properties.
+
+There must be a corresponding export of `IInterceptingPropertyValueProvider` for the property. For example:
+
+```c#
+[ExportInterceptingPropertyValueProvider("MyProperty", ExportInterceptingPropertyValueProviderFile.ProjectFile)]
```
MariaSolOs, line 33: Maybe also mention here how multiple attributes could be used to intercept several properties or that a string array could be passed as a parameter.

drewnoakes: Thank you, that's definitely something we should cover. I think we could make more use of this feature in this repo, actually.

docs/repo/property-pages/property-value-interception.md
```diff
@@ -0,0 +1,160 @@
...
+        return Task.FromResult("❤️"); // This property's value is always ❤️
+    }
+}
+```
+
+⚠️ The property name passed in the `ExportInterceptingPropertyValueProvider` attribute **must** match the property being intercepted.
+
+The name of the class is not important, but you might like to follow the convention offered here.
+
+A single interceptor class may be used to intercept multiple properties. To do that, add one `ExportInterceptingPropertyValueProvider` attribute per property you wish to intercept, changing the name on each accordingly.
```
tmeschter, line 52: You can also pass multiple string values in an array to a single `ExportInterceptingPropertyValueProvider` attribute.

docs/repo/property-pages/property-value-interception.md
```diff
@@ -0,0 +1,160 @@
...
+## Base classes
+
+While you can implement `IInterceptingPropertyValueProvider` directly, you may like to use one of these base types.
+
+- `InterceptingPropertyValueProviderBase` has virtual methods you can overload, with base implementations that just forward values unchanged.
+- `NoOpInterceptingPropertyValueProvider` fixes the value as null or the empty string. Useful when a property should never be written to the project file. See [Fixed content properties](#fixed-content-properties) for more.
+
+## Providing default values
+
+Consider a boolean (check box) property that should default to `true` when no value is specified. You can achieve this with something like the following:
```
tmeschter, line 63: There's already a [BaseProperty.Default] property for specifying a default value, but we don't make use of it. I wonder if we could bake that into the system somehow so that you didn't need to do this for simple things.

tmeschter: (Not directly related to this change, though.)

tmeschter (approved):

docs/repo/property-pages/property-value-interception.md
tmeschter: Oh! The definition of `ExportInterceptingPropertyValueProviderAttribute` already lets you specify the "underlying" property provider, with three values currently supported:
- ProjectFile
- UserFile
- UserFileWithXamlDefaults

If we added a `ProjectFileWithXamlDefaults` to CPS we could update `ExportInterceptingPropertyValueProviderAttribute` to surface that as a fourth option. Or the property could just specify it as the `DataSource` if it doesn't need interception.

docs/repo/property-pages/property-value-interception.md
drewnoakes: Thanks, fixed. Actually the attribute has `AllowMultiple = false`, so my previous revision of this paragraph was outright wrong.

docs/repo/property-pages/property-value-interception.md
drewnoakes: That's great. I've filed https://github.com/dotnet/project-system/issues/8152 to track this.

Issue #7237, opened 2021-05-18
Re-create VB Compile property page (2)


tmeschter:
We need to re-create the VB Compile property page in the new property page system.

For reference, this is the Compile property page as it currently appears for VB WinForms apps targeting .NET 5:

![image](https://user-images.githubusercontent.com/10506730/118734105-8407bb80-b7f2-11eb-8707-f642cb709384.png)

Note that this is very different from the C# Compile property page. It may warrant a complete separate implementation, rather than having a "base" page with the common properties plus per-language "extension" pages.

This work should also include the options exposed in the "Advanced Compile Options" dialog.

_This work was identified as part of #6935._

PR #8146, opened 2022-05-09
Define properties for the VB Compile (Build) page

tmeschter:
Related to #7237.

Define the properties of the VB "Compile" page (called "Build" in C#/F#) and arrange them into categories. A couple of things to note:

- The goal is to get everything stubbed out so we have controls to interact with in future changes and so we can start to discuss the look and feel of the page. Most of these properties do not yet read and write values correctly.
- Items are likely to move around in future commits. This change puts _most_ items in the same order as the old Compile property page, with the exception of "Generate XML documentation file" and "Register for COM interop". The old page visually located these under "Warning configurations" even though they have nothing to do with warnings. Here we put them in the "General" category instead.
- I found VB to be sufficiently different from C#/F# that this page replaces the base "Build" page entirely, rather than extracting shared functionality to the base page and then extending it to add VB-specific functionality. As such, there is some amount of duplicated functionality, but not as much as you might expect. Even where the behavior is the same VB organizes the settings into different categories or uses different display names and description text. This largely precludes reusing a shared property in the base page. I think this is fine, given that I expect the VB pages to change little in the future, if at all.

Screen shot:
VBCompilePage

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8146)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
@@ -1,6 +1,375 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Rule Name="Build"
-      OverrideMode="Extend"
+      OverrideMode="Replace"
+      Description="Specifies properties that control how the project builds."
+      DisplayName="Compile"
+      PageTemplate="generic"
+      Order="200"
       xmlns="http://schemas.microsoft.com/build/2009/properties">
 
+  <Rule.Categories>
+    <Category Name="General"
+              DisplayName="General" />
+
+    <Category Name="CompileOptions"
+              DisplayName="Compile Options" />
+
+    <Category Name="WarningConfigurations"
+              DisplayName="Warning configurations" />
+
+    <Category Name="Events"
+              Description="Configures custom events that run before and after build."
+              DisplayName="Events" />
+
+    <Category Name="Advanced"
+              DisplayName="Advanced"
+              Description="Advanced settings for the application." />
+  </Rule.Categories>
+
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFileWithInterception"
+                HasConfigurationCondition="True" />
+  </Rule.DataSource>
+
+  <StringProperty Name="Output path"
+                  DisplayName="Build output path"
+                  Category="General"
+                  Subtype="directory" />
+
+  <!-- TODO: Update this with correct behavior for SDK-style projects -->
+  <BoolProperty Name="GenerateDocumentationFile"
+                DisplayName="Generate XML documentation file"
+                Description="Specifies whether to generate documentation information."
+                Category="General" />
+
+  <!-- TODO: Condition the visibility on this being a class library -->
+  <BoolProperty Name="RegisterForComInterop"
+                DisplayName="Register for COM interop"
+                Description="Specifies whether your managed application will expose a COM object (a COM-callable wrapper) that enables a COM object to interact with the application."
+                Category="General" />
+
+  <!-- TODO: Replace HelpUrl with a fwlink -->
+  <EnumProperty Name="OptionExplicit"
+                DisplayName="Option explicit"
+                Description="Specifies whether to require explicit declaration of variables."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/compile-page-project-designer-visual-basic#compiler-configuration-options"
+                Category="CompileOptions">
+    <EnumValue Name="On" DisplayName="On" />
+    <EnumValue Name="Off" DisplayName="Off" />
+  </EnumProperty>
+
+  <!-- TODO: Replace HelpUrl with a fwlink -->
+  <!-- TODO: Enum values need to to come from a dynamic enum provider -->
+  <EnumProperty Name="OptionStrict"
+                DisplayName="Option strict"
+                Description="Specifies whether to enforce strict type semantics."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/compile-page-project-designer-visual-basic#compiler-configuration-options"
+                Category="CompileOptions">
+    <EnumValue Name="On" DisplayName="On" />
+    <EnumValue Name="Off" DisplayName="Off" />
+    <EnumValue Name="Custom" DisplayName="(custom)" />
+  </EnumProperty>
+
+  <!-- TODO: Replace HelpUrl with a fwlink -->
+  <EnumProperty Name="OptionCompare"
+                DisplayName="Option compare"
+                Description="Specifies the type of string comparison to use."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/compile-page-project-designer-visual-basic#compiler-configuration-options"
+                Category="CompileOptions">
+    <EnumValue Name="Binary" DisplayName="Binary" />
+    <EnumValue Name="Text" DisplayName="Text" />
+  </EnumProperty>
+
+  <!-- TODO: Replace HelpUrl with a fwlink -->
+  <EnumProperty Name="OptionInfer"
+                DisplayName="Option Infer"
+                Description="Specifies whether to allow local type inference in variable declarations."
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/compile-page-project-designer-visual-basic#compiler-configuration-options"
+                Category="CompileOptions">
+    <EnumValue Name="On" DisplayName="On" />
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -1,6 +1,375 @@
...
+              DisplayName="Advanced"
+              Description="Advanced settings for the application." />
+  </Rule.Categories>
+
+  <Rule.DataSource>
+    <DataSource Persistence="ProjectFileWithInterception"
+                HasConfigurationCondition="True" />
+  </Rule.DataSource>
+
+  <StringProperty Name="Output path"
```
melytc, line 35: Should we also bring the `Browse` button for this property?
image

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -1,6 +1,375 @@
...
+      PageTemplate="generic"
+      Order="200"
       xmlns="http://schemas.microsoft.com/build/2009/properties">
 
+  <Rule.Categories>
+    <Category Name="General"
+              DisplayName="General" />
+
+    <Category Name="CompileOptions"
+              DisplayName="Compile Options" />
```
drewnoakes, line 16: "Options" might be a better fit here. It's sitting right under "Compile" so I don't think we need to repeat that word. It'll also prevent the category's name breaking across two lines.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -1,6 +1,375 @@
...
+
+    <Category Name="WarningConfigurations"
+              DisplayName="Warning configurations" />
```
drewnoakes, line 19: Similarly, the label "Warnings" would suffice here IMO and prevent wrapping in the navigation tree.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
drewnoakes: I would have expected `Subtype="directory"` to do that. It seems to work for other properties (i.e. base output path).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -1,6 +1,375 @@
...
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/compile-page-project-designer-visual-basic#compiler-configuration-options"
+                Category="WarningConfigurations">
+    <EnumValue Name="None" DisplayName="None" />
+    <EnumValue Name="Warning" DisplayName="Warning" />
+    <EnumValue Name="Error" DisplayName="Error" />
+  </EnumProperty>
+
+  <!-- TODO: Replace HelpUrl with a fwlink -->
+  <EnumProperty Name="LateBinding"
+                DisplayName="Late binding; call could fail at run time"
```
drewnoakes, line 141: Could the text after the semicolon be moved to a description? I think that'd be more consistent with our other properties, and I'm also not sure how semicolons translate to other languages' grammars.

Ditto below for `ImplicitType`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -1,6 +1,375 @@
...
+                Category="Advanced">
+    <EnumProperty.Metadata>
+      <NameValuePair Name="SearchTerms" Value="debug type" />
+    </EnumProperty.Metadata>
+    <EnumValue Name="none"
+               DisplayName="No symbols are emitted" />
+    <!--
+    Note that 'pdbonly' is the same as 'full'.
+    <EnumValue Name="pdbonly"
+               DisplayName="PDB Only" />
```
drewnoakes, line 336: I assume our current interceptors will also apply here, to map `pdbonly` to `full` in the UI.

drewnoakes: On that note, I wonder if we can filter interceptors by project capability in cases where we need different behaviour per language. I'm guessing not, seeing as we don't have any `AppliesTo` on the exports today.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -1,6 +1,375 @@
...
+                DisplayName="Degine TRACE constant"
+                HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/advanced-compiler-settings-dialog-box-visual-basic#compilation-constants"
+                Category="Advanced"/>
+
+  <!-- TODO: Replace HelpUrl with a fwlink -->
+  <StringProperty Name="DefineConstants"
+                  DisplayName="Custom constants"
+                  HelpUrl="https://docs.microsoft.com/visualstudio/ide/reference/advanced-compiler-settings-dialog-box-visual-basic#compilation-constants"
+                  Category="Advanced">
+    <StringProperty.Description>Enter any custom constants for your application in this text box. Entries should be delimited by commas, using this form: Name1="Value1",Name2="Value2",Name3="Value3".</StringProperty.Description>
```
drewnoakes, line 363: Could we re-use the `NameValueList` editor here? The format looks a little different, so perhaps we'd need to add the encoding extension point that we've discussed previously.

drewnoakes (approved):
Very nice!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
tmeschter: We could make them capability-aware, but that would require tagging all the interceptors with an `AppliesTo` attribute or they just stop showing up. That's fine for us since we could do it all at once, but poses a problem for other teams implementing interceptors. We would need to get them to add the attributes _before_ we made the change.

We should probably do it anyway; right now we risk pulling in interceptors for things like MAUI when we don't need/want them.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
tmeschter: I would certainly like to, but as this has _always_ been a simple string that depended on proper user formatting it's a lower priority.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
tmeschter: Yes, I have `Subtype="directory"` and the "Browse" button is present.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
drewnoakes: Is the encoding actually different? If not, it should be very cheap to change the editor here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
drewnoakes: Filed https://github.com/dotnet/project-system/issues/8170 to track this.

PR #8163, opened 2022-05-11
Move implementation of ConcurrentHashSet<T> to Microsoft.VisualStudio.ProjectSystem.Managed

ocallesp:
`Microsoft.VisualStudio.ProjectSystem.Managed.VS` already has a dependency on `Microsoft.VisualStudio.ProjectSystem.Managed`,
so moving `ConcurrentHashSet` to `Microsoft.VisualStudio.ProjectSystem.Managed.VS` will allow code in this assembly to get access to this collection.

-----
Note: `ConcurrentHashSet` will be used in the `MissingSdkRuntimeDetector`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8163)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Collections/IConcurrentHashSet.cs
```diff
@@ -47,5 +47,10 @@ internal interface IConcurrentHashSet<T> : IEnumerable<T>
         /// otherwise, <c>false</c>
         /// </returns>
         public bool AddRange(IEnumerable<T> elements);
+
+        /// <summary>
+        /// Removes all elements from System.Collections.Concurrent.IConcurrentHashSet`1
+        /// </summary>
+        public void Clear();
```
drewnoakes, line 54: This is an interface, so we should not have visibility modifiers. I know it's not related to this PR, but could you also remove them from this and the other members in this interface?

```suggestion
        void Clear();
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/Collections/IConcurrentHashSet.cs
```diff
@@ -47,5 +47,10 @@ internal interface IConcurrentHashSet<T> : IEnumerable<T>
         /// otherwise, <c>false</c>
         /// </returns>
         public bool AddRange(IEnumerable<T> elements);
```
drewnoakes, line 49:
```suggestion
        bool AddRange(IEnumerable<T> elements);
```

drewnoakes (approved):

jacdavis:
>         private readonly ConcurrentDictionary<T, object> _map;

What's the point of this class over just using the ConcurrentDictionary directly? The interface to allow for testing?

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed/Collections/ConcurrentHashSet.cs:11 in 52e2fcc. [](commit_id = 52e2fcc65171f4739a29dd35221525a9528ea656, deletion_comment = False)

jacdavis:
>         IEnumerator IEnumerable.GetEnumerator()

Be careful with enumerators on concurrent dictionary. This is a threadsafe implementation but modifications are reflected immediately in the enumerator. That is, it is NOT a snapshot of the underlying data. If another thread is modifying the collection, the new items may (or may not) actually be reflected in the enumerator depending on where it is added. Not sure if this matters for your scenario. 

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed/Collections/ConcurrentHashSet.cs:30 in 52e2fcc. [](commit_id = 52e2fcc65171f4739a29dd35221525a9528ea656, deletion_comment = False)

jacdavis:
>             return _map.TryAdd(item, s_hashSetObject);

Why are you adding this static member instead of the value? Is the expectation that this is simply used to test for existence of a key? If so, is there a better name than ConcurrentHashSet? HashSet implies key->value mapping where duplicates overwrite, not key existence. The name currently implies a generic HashSet

---
Refers to: src/Microsoft.VisualStudio.ProjectSystem.Managed/Collections/ConcurrentHashSet.cs:37 in 52e2fcc. [](commit_id = 52e2fcc65171f4739a29dd35221525a9528ea656, deletion_comment = False)

MiYanni (approved):
I looked at Jackson's comments on the type itself, but I understand Osvaldo is just trying to use our existing code and doesn't know the details on why this class was implemented. He just wanted to use it in the Non-`.VS` project since the type didn't exist there. You'd have to look at the history of this file to see why it was implemented over using standard .NET ConcurrentDictionary, but that's not directly related to the work Osvaldo is doing.

jacdavis:
@MiYanni, so, this is all existing code? Codeflow as showing it as move + changes.

ocallesp:
We removed the visibility modifier in the interface.

MiYanni:
@jacdavis I don't use CodeFlow. GitHub simply showed the file being 'renamed' (moved).
image

PR #8171, opened 2022-05-12
Detect runtimes outside vs

ocallesp:
Scenario to fix [AB#1460328](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1460328) : A user installs a standalone package downloaded from the msft website, then after that the user installs VS and creates a project using a specific runtime version. VS will not be able to detect if a runtime is already installed, so VS might prompt the gold bar in the solution explorer suggesting to install the missing runtime even thought it might be already installed.

One way to detect the packages installed in a machine is with deep detection, but it is complex and costly in terms of the size of the implementation and performance. Even if we cached the results, we might see a regression on Perf DDRITs.

A workaround is to rely on `HKLM\SOFTWARE\WOW6432Node\dotnet\Setup\InstalledVersions\x64\sharedfx\Microsoft.NETCore.App` and enumerate the value names. This path applies for x86, x64 and arm.

This pr implements the workaround. It reads the Registry Key to get the list of packages installed and cache it.
For every runtime detected in a project we always check if the runtime version is in the list and if not then we check the installed packages in the VS Setup before trying to prompt the gold bar.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8171)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Concurrent;
 using Microsoft.ServiceHub.Framework;
+using Microsoft.VisualStudio.ProjectSystem.Runtimes;
 using Microsoft.VisualStudio.ProjectSystem.Utilities;
 using Microsoft.VisualStudio.ProjectSystem.Workloads;
 using Microsoft.VisualStudio.Shell.Interop;
@@ -19,6 +20,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
     {
         private const string WasmToolsWorkloadName = "wasm-tools";
 
+        private static readonly ImmutableDictionary<string, string> s_packageVersionToComponentId = ImmutableDictionary.Create<string, string>(StringComparer.OrdinalIgnoreCase)
+            .Add("v2.0", "Microsoft.Net.Core.Component.SDK.2.1")
+            .Add("v2.1", "Microsoft.Net.Core.Component.SDK.2.1")
+            .Add("v2.2", "Microsoft.Net.Core.Component.SDK.2.1")
+            .Add("v3.0", "Microsoft.NetCore.Component.Runtime.3.1")
+            .Add("v3.1", "Microsoft.NetCore.Component.Runtime.3.1")
+            .Add("v5.0", "Microsoft.NetCore.Component.Runtime.5.0")
+            .Add("v6.0", "Microsoft.NetCore.Component.Runtime.6.0");
+
+        private static readonly string s_netCoreRegistryKeyPath = "SOFTWARE\\WOW6432Node\\dotnet\\Setup\\InstalledVersions\\x64\\sharedfx\\";
+        private static readonly string s_netCoreRegistryKeyName = "Microsoft.NETCore.App";
         private static readonly ImmutableHashSet<string> s_supportedReleaseChannelWorkloads = ImmutableHashSet.Create(StringComparers.WorkloadNames, WasmToolsWorkloadName);
 
         private readonly ConcurrentHashSet<string> _missingRuntimesRegistered = new(StringComparers.WorkloadNames);
@@ -37,6 +49,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         private uint _solutionCookie = VSConstants.VSCOOKIE_NIL;
         private IVsSolution? _vsSolution;
         private bool? _isVSFromPreviewChannel;
+
+        private readonly object _lock = new();
+        private static HashSet<string>? _netCoreRegistryKeyValues = null;
 
         [ImportingConstructor]
         public MissingSetupComponentRegistrationService(
@@ -72,6 +87,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             }
         }
 
+        private HashSet<string> NetCoreRegistryKeyValues
+        {
+            get
+            {
+                if (_netCoreRegistryKeyValues is null)
+                {
+                    lock (_lock)
+                    {
+                        if (_netCoreRegistryKeyValues is null)
+                        {
+                            var registryKeyReader = new RegistryReaderRuntimesVersions();
+
+                            _netCoreRegistryKeyValues = (HashSet<string>?)registryKeyReader.ReadValueForCurrentUser(s_netCoreRegistryKeyPath, s_netCoreRegistryKeyName);
+                        }
+                    }
+                }
+
+                return _netCoreRegistryKeyValues!;
+            }
+        }
+
         private void ClearMissingWorkloadMetadata()
         {
             _missingRuntimesRegistered.Clear();
@@ -92,13 +128,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             UnregisterProjectConfiguration(projectGuid, project);
         }
 
-        public void RegisterSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string? runtimeComponentId)
+        public void RegisterPossibleMissingSdkRuntimeVersion(Guid projectGuid, ConfiguredProject project, string runtimeVersion)
         {
-            if (runtimeComponentId is not null)
+            // Workaround to fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1460328
+            // VS has no information about the packages installed outside VS, and deep detection is not suggested for performance reasons.
+            // This workaround reads the Registry Key HKLM\SOFTWARE\WOW6432Node\dotnet\Setup\InstalledVersions\x64\sharedfx\Microsoft.NETCore.App
+            // and get the installed runtime versions from the value names.
+            if (!string.IsNullOrEmpty(runtimeVersion) &&
+                !NetCoreRegistryKeyValues.Contains(runtimeVersion) &&
+                s_packageVersionToComponentId.TryGetValue(runtimeVersion, value: out string? componentId))
             {
-                _missingRuntimesRegistered.Add(runtimeComponentId);
-
-                _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, runtimeComponentId);
+                _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, componentId);
             }
 
             UnregisterProjectConfiguration(projectGuid, project);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/IRegistry.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/IRegistry.cs
@@ -1,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Microsoft.VisualStudio.IO
+{
+    /// <summary>
+    ///     Provides methods for reading values from the registry.
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    internal interface IRegistry
+    {
+        /// <summary>
+        /// Opens a registry key denoted by <paramref name="keyPath"/> for the current user,
+        /// and returns the value (if any) associated with <paramref name="name"/>. Returns
+        /// <see langword="null"/> if the registry key or value do not exist, or if an error
+        /// occurs while reading the value.
+        /// </summary>
+        object? ReadValueForCurrentUser(string keyPath, string name);
+    }
+}
+
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
@@ -10,11 +10,6 @@ namespace Microsoft.VisualStudio.ProjectSystem
     {
         private static readonly string s_netCoreTargetFrameworkIdentifier = ".NETCoreApp";
 
-        private static readonly ImmutableDictionary<string, string> s_packageVersionToComponentId = ImmutableDictionary.Create<string, string>(StringComparer.OrdinalIgnoreCase)
-            .Add("v2.1", "Microsoft.Net.Core.Component.SDK.2.1")
-            .Add("v3.1", "Microsoft.NetCore.Component.Runtime.3.1")
-            .Add("v5.0", "Microsoft.NetCore.Component.Runtime.5.0");
-
         private Guid _projectGuid;
         private bool _enabled;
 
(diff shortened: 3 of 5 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -37,6 +45,37 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
...
+        {
+            get
+            {
+                if (_netCoreRegistryKeyValues is null)
+                {
+                    lock (_lock)
+                    {
+                        if (_netCoreRegistryKeyValues is null)
+                        {
+                            var regkey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry64);
```
MiYanni, line 61: Again... only works on 64-bit Windows?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -37,6 +45,37 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
...
+                            var subKey = regkey.OpenSubKey(s_netCoreRegistryKeyPath);
+
+                            _netCoreRegistryKeyValues = new(StringComparer.OrdinalIgnoreCase);
+
+                            foreach (string valueName in subKey.GetValueNames())
+                            {
+                                string versionNumber = valueName.Substring(0, valueName.LastIndexOf('.'));
```
MiYanni, line 68: Is this guaranteed to always be in this format?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -90,9 +130,17 @@ public void RegisterMissingWorkloads(Guid projectGuid, ConfiguredProject project
             UnregisterProjectConfiguration(projectGuid, project);
         }
 
-        public void RegisterMissingSdkRuntimeComponentId(Guid projectGuid, ConfiguredProject project, string runtimeComponentId)
+        public void RegisterPossibleMissingSdkRuntimeVersion(Guid projectGuid, ConfiguredProject project, string runtimeVersion)
         {
-            _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, runtimeComponentId);
+            // Workaround to fix #AB1460328.
```
MiYanni, line 135: Please provide full link to the issue since that issue number format is specific to which Azure DevOps org it is part of. Meaning, someone in the future (or anyone else looking at the code) might not know how to use this value.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -90,9 +130,17 @@ public void RegisterMissingWorkloads(Guid projectGuid, ConfiguredProject project
...
+            // VS has no information about the packages installed outside VS, and deep detection is not suggested for performance reasons.
+            // This workaround reads the Registry Key HKLM\SOFTWARE\WOW6432Node\dotnet\Setup\InstalledVersions\x64\sharedfx\Microsoft.NETCore.App
+            // and get the installed runtime versions from the value names.
+            if (!string.IsNullOrEmpty(runtimeVersion) && !NetCoreRegistryKeyValues.Contains(runtimeVersion))
+            {
+                s_packageVersionToComponentId.TryGetValue(runtimeVersion, out string? componentId);
+                _projectGuidToRuntimeDescriptorMap.GetOrAdd(projectGuid, componentId!);
```
MiYanni, line 142: Is `componentId` guaranteed to not be null here (I'm looking at that exclamation point)? `TryGetValue` should be part of the if-statement because if you don't get a value, what is the purpose of running `GetOrAdd`?

MiYanni (commented):
This solution seems to have potential of breaking over time. Meaning, it 'ages'. I feel like there could be some point where it becomes a bug and no one knows why. I'd use a solution like this in a tool I'm writing on my own, but in a 25 year old product, this just seems like a bug waiting to happen. Someone can correct me if I'm wrong, but I don't think we read registry values elsewhere in our codebase (but likely in CPS). May be a better way to implement this solution to make it more resilient?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -19,8 +20,15 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
     {
         private const string WasmToolsWorkloadName = "wasm-tools";
 
+        private static readonly ImmutableDictionary<string, string> s_packageVersionToComponentId = ImmutableDictionary.Create<string, string>(StringComparer.OrdinalIgnoreCase)
+            .Add("v2.1", "Microsoft.Net.Core.Component.SDK.2.1")
+            .Add("v3.1", "Microsoft.NetCore.Component.Runtime.3.1")
+            .Add("v5.0", "Microsoft.NetCore.Component.Runtime.5.0");
```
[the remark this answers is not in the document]
ocallesp: To remove the hardcoded map, Sdk first needs to implement something on their side to pass this information using design-time.

That improvement can go in another pr.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
```diff
@@ -73,14 +65,13 @@ private async Task RegisterSdkRuntimeNeededInProjectAsync(ConfiguredProject proj
 
                 string? TargetFrameworkVersion = await configuration.TargetFrameworkVersion.GetDisplayValueAsync();
```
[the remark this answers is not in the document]
ocallesp: Done. I did a rebase on main which already changed this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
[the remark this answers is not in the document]
kvenkatrajan: Osvaldo, has this been brought up with the SDK team & a backlog item been added? If so please share the work item/issue # prior to resolving this comment. We want to avoid keeping hard coded values for mapping SDK versions.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
MiYanni: What I meant that is there is no 64-bit registry on a 32-bit Windows (or at least that's how it used to be). There are explicitly 2 different registries for 32-bit versus 64-bit.
Read this: https://serverfault.com/questions/7973/is-there-a-distinction-between-64-bit-and-32-bit-regedit

drewnoakes: VS has its own registry hive, so I don't think this is a concern.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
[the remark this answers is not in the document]
ocallesp: I created this https://github.com/dotnet/project-system/issues/8186 to keep track of this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -19,8 +20,15 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
...
+
+        private static readonly string s_netCoreRegistryKeyPath = "SOFTWARE\\WOW6432Node\\dotnet\\Setup\\InstalledVersions\\x64\\sharedfx\\Microsoft.NETCore.App";
```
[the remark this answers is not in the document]
ocallesp: I haven't try this on a 32-bit machine, but my understanding is that this registry key path also applies to 32-bit.

It seems that Windows-11 only supports 64-bit.

ocallesp:
> This solution seems to have potential of breaking over time. Meaning, it 'ages'. I feel like there could be some point where it becomes a bug and no one knows why. I'd use a solution like this in a tool I'm writing on my own, but in a 25 year old product, this just seems like a bug waiting to happen. Someone can correct me if I'm wrong, but I don't think we read registry values elsewhere in our codebase (but likely in CPS). May be a better way to implement this solution to make it more resilient?

I agree with you.
After a meeting with the VS Setup and Sdk, this was the recommended suggestion.
This information should be provided by VS Setup through `IVsSetupCompositionService`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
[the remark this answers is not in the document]
tmeschter: I suggest creating a bug in the dotnet/sdk repo about this.

One idea: They could expose this information as MSBuild items, as they do for think like the list of .NET versions supported by the SDK.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
tmeschter: This is looking at a machine-wide part of the registry, which is the point. If this information were available in the VS-instance-specific part of the registry, we probably wouldn't need to look at the registry at all.

But as VS is 64-bit (and it explicitly requires a 64-bit CPU and 64-bit Windows) I don't think this is much of a concern.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -38,6 +49,38 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
         private IVsSolution? _vsSolution;
         private bool? _isVSFromPreviewChannel;
 
+        private readonly object _lock = new();
+        private static HashSet<string>? _netCoreRegistryKeyValues = null;
+
+        private HashSet<string> NetCoreRegistryKeyValues
```
tmeschter, line 55: I would move all the registry access to another type that implements an `IRegistry` interface. I _removed_ a service like that in https://github.com/dotnet/project-system/pull/7880 because it was no longer needed, but you can bring it back.

That separates one concern (accessing the registry) from the work of this type (checking if the necessary runtime is installed) and will make this easier to understand and test with unit tests.

tmeschter (approved):
I'm not particularly bothered by the hard-coded values. The set will expand (slowly) over time, but I think this is a reasonable workaround until we can get the information from the SDK or some other VS component.

davkean:
We already ran into stale data between .NET Core 1.1 -> 2.0 timeframe looking at something similar. This will break with a simple insertion from MSBuild and you will never find out.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -38,6 +50,9 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
         private IVsSolution? _vsSolution;
         private bool? _isVSFromPreviewChannel;
 
+        private readonly object _lock = new();
```
drewnoakes, line 53: This needs to be static in order to protect `_netCoreRegistryKeyValues` properly, as that field is static.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -72,6 +87,27 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
...
+                        if (_netCoreRegistryKeyValues is null)
+                        {
+                            var registryKeyReader = new RegistryReaderRuntimesVersions();
+
+                            _netCoreRegistryKeyValues = (HashSet<string>?)registryKeyReader.ReadValueForCurrentUser(s_netCoreRegistryKeyPath, s_netCoreRegistryKeyName);
+                        }
+                    }
+                }
+
+                return _netCoreRegistryKeyValues!;
```
drewnoakes, line 107: Why suppress this? It can return null. The caller needs to handle that.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/RegistryReaderRuntimesVersions.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.Runtimes
+{
+    internal class RegistryReaderRuntimesVersions : IRegistry
+    {
+        public object? ReadValueForCurrentUser(string keyPath, string name)
+        {
+            var regkey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry64);
+            var subKey = regkey.OpenSubKey(keyPath + name);
+
+            HashSet<string>? netCoreRegistryKeyValues = new(StringComparer.OrdinalIgnoreCase);
```
drewnoakes, line 15: This local will never be null.

src/Microsoft.VisualStudio.ProjectSystem.Managed/IO/IRegistry.cs
```diff
@@ -0,0 +1,23 @@
...
+    ///     Provides methods for reading values from the registry.
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
+    internal interface IRegistry
+    {
+        /// <summary>
+        /// Opens a registry key denoted by <paramref name="keyPath"/> for the current user,
+        /// and returns the value (if any) associated with <paramref name="name"/>. Returns
+        /// <see langword="null"/> if the registry key or value do not exist, or if an error
+        /// occurs while reading the value.
```
drewnoakes, line 18: This documentation does not match the implementation.

PR #8175, opened 2022-05-13
Reduce allocations in ActiveEditorContextTracker

drewnoakes:
This class tracks a set of context IDs. It enforced the set behaviour, yet used an ImmutableList. Using an ImmutableHashSet simplifies the code.

The two optimistic update functions have been changed to not allocate a closure object per call.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8175)


MiYanni (approved):

drewnoakes:
`ActiveIntellisenseProjectContext` can return the first context ID. `ImmutableHashSet<>` does not preserve order, so this code introduces a behaviour change. Closing for now.

PR #8177, opened 2022-05-16
Make VB warning sets work

tmeschter:
The VB "Compile" property page doesn't allow you to disable individual warnings or promote individual warnings to errors. Instead, it groups related warnings into sets and lets the user adjust the severity of the whole set at once. This commit adds the `VBWarningsValueProvider` interceptor, which converts back and forth from the "None", "Warning", and "Error" values for the various warning sets displayed on the VB compile page, and the values of the "NoWarn" and "WarningsAsErrors" properties (among others) persisted to the project file.

This also adds the `VBDisableAllWarningsValueProvider`, a simpler interceptor that maps between the "Disable all warnings" checkbox on the property page, and the "WarningLevel" property in the project file.

Related to #7237.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8177)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
@@ -11,7 +11,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties;
 /// values persisted to the project file.
 /// </para>
 /// <para>
-/// Note this types stores IDs in instances of <see cref="ImmutableSortedSet{T}"/>,
+/// Note this type stores IDs in instances of <see cref="ImmutableSortedSet{T}"/>,
 /// rather a simple array or list. Two reasons:
 /// <list type="number">
 /// <item>We need to perform actual set operations (union, difference, etc.)</item>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
@@ -1,0 +1,292 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+/// <remarks>
+/// <para>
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
```diff
@@ -0,0 +1,292 @@
...
+/// <remarks>
+/// <para>
+/// The VB "Compile" property page doesn't allow you to disable individual warnings
+/// or promote individual warnings to errors. Instead, it groups related warnings
+/// into sets and lets the user adjust the severity of the whole set at once. This
+/// type is responsible for mapping between those sets and the actual property
+/// values persisted to the project file.
+/// </para>
+/// <para>
+/// Note this types stores IDs in instances of <see cref="ImmutableSortedSet{T}"/>,
```
drewnoakes, line 14:
```suggestion
/// Note this type stores IDs in instances of <see cref="ImmutableSortedSet{T}"/>,
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
```diff
@@ -0,0 +1,292 @@
...
+        UnusedLocalVariablePropertyName,
+        InstanceVariableAccessesSharedMemberPropertyName,
+        RecursiveOperatorOrPropertyAccessPropertyName,
+        DuplicateOrOverlappingCatchBlocksPropertyName
+    },
+    ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+internal sealed class VBWarningsValueProvider : InterceptingPropertyValueProviderBase
+{
+    // The names of the property "sets" as known to the property page.
+    internal const string ImplicitConversionPropertyName = "ImplicitConversion";
```
drewnoakes, line 40: Just to confirm: none of these literal values can be sourced from generated rule file constants? Using such constants helps make explicit the otherwise implicit relationship between them, causing compile errors if we remove the properties from the rule, for example.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/VBWarningsValueProvider.cs
tmeschter: I'll double check. Note that the system doesn't use `Rule`s at this level* (i.e. we don't need a `Rule` to define which properties we can access), so I don't want to introduce a false dependency, either.

*That's a lie, but true enough for the purposes of this discussion. :-)

PR #8188, opened 2022-05-19
Only escape \r\n in command line args

### LINKED ISSUE #8167: Command line arguments in Debug Profiles UI are not correctly escaping escape sequences
## Visual Studio Version
Version 17.3.0 Preview 2.0 [32503.442.main]

## Summary
The command line arguments requires the user to understand the underlying format that it is used to correctly escape their own strings. This should be handled by whatever writes the file. 

I cannot figure out a workaround for this issue, other than renaming the argument to not include "\n"

## Steps to Reproduce

1.  File -> New -> Console App with program.cs that contains:

``` C#
Console.WriteLine(args[0]);
```

2.  Right-click on project and choose Properties -> Debug -> Open debug launch profiles UI
3. In command-line enter "C:\Users\davkean\AppData\Local\Temp\VsEngDropExplorer\93d5d61d38386807d9f4663c7c74839f\TestExecutionOutputs\OrchardCoreWithCache.TestSolutionOpenClose\Warmup-1-20220511-193634_dtl-qe1vpwty\NGenLogs\Framework64\ngen.log"
4. F5

## Expected Behavior
```
C:\Users\davkean\AppData\Local\Temp\VsEngDropExplorer\93d5d61d38386807d9f4663c7c74839f\TestExecutionOutputs\OrchardCoreWithCache.TestSolutionOpenClose\Warmup-1-20220511-193634_dtl-qe1vpwty\NGenLogs\Framework64\ngen.log
```

## Actual Behavior
```
C:\Users\davkean\AppData\Local\Temp\VsEngDropExplorer\93d5d61d38386807d9f4663c7c74839f\TestExecutionOutputs\OrchardCoreWithCache.TestSolutionOpenClose\Warmup-1-20220511-193634_dtl-qe1vpwty\NGenLogs\Framework64
gen.log
```

adamint:
This fixes #8167. VS editor escapes user-input `\r\n` as `\\r\\n`, whereas newlines in the editor are `\r\n`. Since VS Windows is Windows-only, we do not need to be OS-agnostic about newlines and can assume we just need to escape `\r\n`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8188)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
 using Microsoft.VisualStudio.Buffers.PooledObjects;
 using Microsoft.VisualStudio.Debugger.UI.Interfaces.HotReload;
 using Microsoft.VisualStudio.IO;
@@ -261,7 +262,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Debug
                 }
             }
 
-            string? commandLineArgs = resolvedProfile.CommandLineArgs?.Replace("\r\n", " ");
+            string? commandLineArgs = resolvedProfile.CommandLineArgs is null
+                ? null
+                : Regex.Replace(resolvedProfile.CommandLineArgs, "[\r\n]+", " ");
 
             // Is this profile just running the project? If so we ignore the exe
             if (IsRunProjectCommand(resolvedProfile))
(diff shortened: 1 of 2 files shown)


tmeschter (approved):

davkean:
> Since VS Windows is Windows-only, we do not need to be OS-agnostic about newlines and can assume we just need to escape \r\n

This isn't a good assumption, especially when Git is setup to normalize line endings to `\n`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Debug/ProjectLaunchTargetsProvider.cs
```diff
@@ -261,12 +261,7 @@ private async Task<bool> IsIntegratedConsoleEnabledAsync()
                 }
             }
 
-            string? commandLineArgs = resolvedProfile.CommandLineArgs?
-                .Replace("\r\n", " ")
-                .Replace("\r", " ")
-                .Replace("\n", " ")
-                .Replace("\\r\\n", "\r\n")
-                .Replace("\\n", "\n");
+            string? commandLineArgs = resolvedProfile.CommandLineArgs?.Replace("\r\n", " ");
```
drewnoakes, line 264: As suggested in https://github.com/dotnet/project-system/pull/7919#discussion_r806293499 I'd consider using this instead:

```suggestion
            string? commandLineArgs = resolvedProfile.CommandLineArgs is null
                ? null
                : Regex.Replace(resolvedProfile.CommandLineArgs, "[\r\n]+", " ");
```

There is no guarantee that the file doesn't contain other kinds of line endings. Git can meddle with line endings, for example.

The bug in 17.2 was caused by the addition of the last two lines. That was escaping the literal `@"\n"` (i.e. two characters) rather than the newline character.

adamint:
@drewnoakes

drewnoakes (approved):

PR #8190, opened 2022-05-19
Add a web workload detector IPA

ocallesp:
There is a requirement from the Wpf team to suggest a user to install Web workload through the gold bar in the Solution Explorer when a project uses any of the following high level SDKs:

*	`<Project Sdk="Microsoft.NET.Sdk.Web">`
*	`<Project Sdk="Microsoft.NET.Sdk.Razor">`
*	`<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">`

With this implementation we should be able to detect the required capabilities during solution load and when
NuGet packages are using this SDKs, or when a user make changes the project file to one of this types.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8190)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -138,7 +138,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             var workloadDescriptorSet = _projectGuidToWorkloadDescriptorsMap.GetOrAdd(projectGuid, guid => new ConcurrentHashSet<WorkloadDescriptor>());
             workloadDescriptorSet.AddRange(workloadDescriptors);
 
-            UnregisterProjectConfiguration(projectGuid, project);
+            DisplayYellowBarInSolutionExplorer(project);
 
             bool AreNewComponentIdsToRegister(ISet<WorkloadDescriptor> workloadDescriptors)
             {
@@ -212,14 +212,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         public void UnregisterProjectConfiguration(Guid projectGuid, ConfiguredProject project)
         {
             RemoveConfiguration(projectGuid, project);
-
-            bool displayMissingComponentsPrompt = ShouldDisplayMissingComponentsPrompt();
-            if (displayMissingComponentsPrompt)
-            {
-                var displayMissingComponentsTask = DisplayMissingComponentsPromptAsync();
-
-                _projectFaultHandlerService.Forget(displayMissingComponentsTask, project: project.UnconfiguredProject, ProjectFaultSeverity.Recoverable);
-            }
+            DisplayYellowBarInSolutionExplorer(project);
 
             void RemoveConfiguration(Guid projectGuid, ConfiguredProject project)
             {
@@ -233,8 +226,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
                 {
                     _projectGuidToProjectConfigurationsMap.TryGetValue(projectGuid, out projectConfigurationSet);
                 }
-                
+
                 projectConfigurationSet?.Remove(project.ProjectConfiguration);
+            }
+        }
+
+        private void DisplayYellowBarInSolutionExplorer(ConfiguredProject project)
+        {
+            bool displayMissingComponentsPrompt = ShouldDisplayMissingComponentsPrompt();
+            if (displayMissingComponentsPrompt)
+            {
+                var displayMissingComponentsTask = DisplayMissingComponentsPromptAsync();
+
+                _projectFaultHandlerService.Forget(displayMissingComponentsTask, project: project.UnconfiguredProject, ProjectFaultSeverity.Recoverable);
             }
         }
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -138,7 +138,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             var workloadDescriptorSet = _projectGuidToWorkloadDescriptorsMap.GetOrAdd(projectGuid, guid => new ConcurrentHashSet<WorkloadDescriptor>());
             workloadDescriptorSet.AddRange(workloadDescriptors);
 
-            DisplayYellowBarInSolutionExplorer(project);
+            DisplayMissingComponentsPromptIfNeeded(project);
 
             bool AreNewComponentIdsToRegister(ISet<WorkloadDescriptor> workloadDescriptors)
             {
@@ -212,7 +212,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
         public void UnregisterProjectConfiguration(Guid projectGuid, ConfiguredProject project)
         {
             RemoveConfiguration(projectGuid, project);
-            DisplayYellowBarInSolutionExplorer(project);
+            DisplayMissingComponentsPromptIfNeeded(project);
 
             void RemoveConfiguration(Guid projectGuid, ConfiguredProject project)
             {
@@ -231,7 +231,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
             }
         }
 
-        private void DisplayYellowBarInSolutionExplorer(ConfiguredProject project)
+        private void DisplayMissingComponentsPromptIfNeeded(ConfiguredProject project)
         {
             bool displayMissingComponentsPrompt = ShouldDisplayMissingComponentsPrompt();
             if (displayMissingComponentsPrompt)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
@@ -31,8 +31,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS
 
         private static readonly ImmutableHashSet<string> s_supportedReleaseChannelWorkloads = ImmutableHashSet.Create(StringComparers.WorkloadNames, WasmToolsWorkloadName);
 
-        private readonly object _wpfComponentLock = new();
-        private readonly ConcurrentHashSet<string> _wpfComponentIdsDetected;
+        private readonly object _webComponentIdsDetectedLock = new();
+        private readonly ConcurrentHashSet<string> _webComponentIdsDetected;
         private readonly ConcurrentHashSet<string> _missingRuntimesRegistered = new(StringComparers.WorkloadNames);
         private readonly ConcurrentDictionary<Guid, IConcurrentHashSet<WorkloadDescriptor>> _projectGuidToWorkloadDescriptorsMap;
         private readonly ConcurrentDictionary<Guid, string> _projectGuidToRuntimeDescriptorMap;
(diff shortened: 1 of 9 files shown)


davkean:
I don't understand the relationship between WPF and Web?

One of the early rules we had around project-system was avoiding the hardcoding of "in-flight" data like this. In 6 months, this data will be out of date as the product changes, the folks that own that those components should own this.

vijayrkn:
> I don't understand the relationship between WPF and Web?

This is for supporting the Blazor Hybrid scenarios. Users can start with just the desktop workload and a WPF application but convert them to a Blazor Hybrid application by changing the project's SDK to Microsoft.NET.Sdk.Razor. 

Documentation: 
https://docs.microsoft.com/en-us/aspnet/core/blazor/hybrid/tutorials/wpf?view=aspnetcore-6.0

Once this is done, the user will lose the ability to work with Razor based SDK projects since the required web components will be missing. 

We can't add this logic to the web layer because our components will not be initialized until the web component is installed. This logic needs to be in a layer that will be available when the WPF project is loaded.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+
+                _componentIdsDetected.Add(s_webComponentId);
+            }
+
+            return Workloads;
+        }
+
+        #region Scenarios to detect
+
+        private static bool WpfDetected(bool dotnetCoreRazor, bool windowsForm, bool wpf) => dotnetCoreRazor && (windowsForm || wpf);
```
ocallesp, line 58: More scenarios to detect will be added here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+        private ISet<WorkloadDescriptor> CreateWorkloadDescriptor(IProjectCapabilitiesSnapshot projectCapabilitiesSnapshot)
+        {
+            var Workloads = ImmutableHashSet<WorkloadDescriptor>.Empty;
+
+            bool dotnetCoreRazor = projectCapabilitiesSnapshot.IsProjectCapabilityPresent(ProjectCapability.DotNetRazor);
+            bool windowsForm = projectCapabilitiesSnapshot.IsProjectCapabilityPresent(ProjectCapability.WindowsForms);
+            bool wpf = projectCapabilitiesSnapshot.IsProjectCapabilityPresent(ProjectCapability.WPF);
+
+            // Detect all possible scenarios and add the corresponding needed component ids.
+            if (!_componentIdsDetected.Contains(s_webComponentId) && WpfDetected(dotnetCoreRazor, windowsForm, wpf))
```
ocallesp, line 46: Because capabilities change very often, we want to detect an specific scenario only once.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -195,11 +226,22 @@ void RemoveConfiguration(Guid projectGuid, ConfiguredProject project)
                 {
                     _projectGuidToProjectConfigurationsMap.TryGetValue(projectGuid, out projectConfigurationSet);
                 }
-                
+
                 projectConfigurationSet?.Remove(project.ProjectConfiguration);
             }
         }
 
+        private void DisplayYellowBarInSolutionExplorer(ConfiguredProject project)
```
tmeschter, line 234: Suggestion: `DisplayMissingComponentsPromptIfNeeded`.

This type doesn't know that the information will end up in a gold bar in the Solution Explorer; that's up to the `IMissingComponentRegistrationService`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/IWpfWorkloadDescriptorDataSource.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Workloads
+{
+    [ProjectSystemContract(ProjectSystemContractScope.ConfiguredProject, ProjectSystemContractProvider.Private, Cardinality = Composition.ImportCardinality.ExactlyOne)]
+    internal interface IWpfWorkloadDescriptorDataSource : IProjectValueDataSource<ISet<WorkloadDescriptor>>
```
tmeschter, line 6: Please add a doc comment explaining the purpose of this type.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Collections.Concurrent;
+using System.Threading.Tasks.Dataflow;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Workloads
+{
+    [Export(typeof(IWpfWorkloadDescriptorDataSource))]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class WpfWorkloadDescriptorDataSource : ChainedProjectValueDataSourceBase<ISet<WorkloadDescriptor>>, IWpfWorkloadDescriptorDataSource
```
tmeschter, line 10: This name is misleading. The type doesn't detect WPF workloads; it detects if the project uses DotNetCoreRazor and either WPF or WinForms. The same goes for the `IWpfWorkloadDescriptorDataSource` interface, the `WpfMissingWorkloadDetector` type, and the fields added to the `MissingSetupComponentRegistrationService`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+            return transformBlock;
+        }
+
+        private ISet<WorkloadDescriptor> CreateWorkloadDescriptor(IProjectCapabilitiesSnapshot projectCapabilitiesSnapshot)
+        {
+            var Workloads = ImmutableHashSet<WorkloadDescriptor>.Empty;
+
+            bool dotnetCoreRazor = projectCapabilitiesSnapshot.IsProjectCapabilityPresent(ProjectCapability.DotNetRazor);
+            bool windowsForm = projectCapabilitiesSnapshot.IsProjectCapabilityPresent(ProjectCapability.WindowsForms);
+            bool wpf = projectCapabilitiesSnapshot.IsProjectCapabilityPresent(ProjectCapability.WPF);
```
tmeschter, line 43: We only need to check for the Microsoft.VisualStudio.Component.Web component when the project has the DotNetCoreRazor capability and either the WinForms or WPF capabilities, correct? Could checking for these capabilities be moved to the `AppliesTo` attribute on `WpfMissingWorkloadDetector` so that it only load when the correct capabilities are present?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    [Export(typeof(IWpfWorkloadDescriptorDataSource))]
+    [AppliesTo(ProjectCapability.DotNet)]
+    internal class WpfWorkloadDescriptorDataSource : ChainedProjectValueDataSourceBase<ISet<WorkloadDescriptor>>, IWpfWorkloadDescriptorDataSource
+    {
+        private static readonly string s_webWorkloadName = "Web";
+        private static readonly string s_webComponentId = "Microsoft.VisualStudio.Component.Web";
+        private static readonly WorkloadDescriptor s_webWorkload = new(s_webWorkloadName, s_webComponentId);
+
+        private readonly ConcurrentHashSet<string> _componentIdsDetected = new();
```
tmeschter, line 16: Would `_componentsIdsRequired` be a better name? If looks like we're checking if the web component is _needed_, not if it is _already_ present.

tmeschter (commented):
I agree with @davkean that we should find a solution where other teams can tell us _what_ component are needed and _when_ without us needing to manage that logic in our repo. For example, (as best I can tell from this change) the web component is needed when a project has both the DotNetCoreRazor capability and either the WinForms or WPF capability.

As it is, the `MissingSetupComponentRegistrationService` and the related types are filling up with logic we're not really qualified to own and maintain.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
ocallesp: Perhaps _IWebWorkloadDescriptorDataSource_

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/IWpfWorkloadDescriptorDataSource.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
ocallesp: yes. This was changed in the code cleanup commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Workloads/WpfWorkloadDescriptorDataSource.cs
ocallesp: changed

ocallesp:
Merge conflict fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -31,6 +31,8 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
 
         private static readonly ImmutableHashSet<string> s_supportedReleaseChannelWorkloads = ImmutableHashSet.Create(StringComparers.WorkloadNames, WasmToolsWorkloadName);
 
+        private readonly object _wpfComponentLock = new();
+        private readonly ConcurrentHashSet<string> _wpfComponentIdsDetected;
```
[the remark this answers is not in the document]
ocallesp: Renamed it.

tmeschter (approved):
I'm going to approve this, but I *do* think we need a more general approach. Could you please create an issue about designing a system where other teams can tell us what components to check for and when we should check for them, so we don't need to have this logic in our own repo?

ocallesp:
> I'm going to approve this, but I _do_ think we need a more general approach. Could you please create an issue about designing a system where other teams can tell us what components to check for and when we should check for them, so we don't need to have this logic in our own repo?

Next week Vijay and me are going to discuss about how to move this logic  out of dotnet project system.

vijayrkn:
> > I'm going to approve this, but I _do_ think we need a more general approach. Could you please create an issue about designing a system where other teams can tell us what components to check for and when we should check for them, so we don't need to have this logic in our own repo?
> 
> Next week @vijayrkn and me are going to discuss about how to move this logic out of dotnet project system.
> 
> This is the issue created to improve the design: #8371

I understand the concerns raised here where you don't want the project system to know about the various workload requirements. I am happy to discuss alternative options. In this case since the detection is based on the capabilities provided by the Razor/Web SDK, maybe they can also provide the VS required component list.

PR #8196, opened 2022-05-23
Finish up work on the diagnostics part of VB Compile/Build page

### LINKED ISSUE #7237: Re-create VB Compile property page (2)

tmeschter:
Fixes #7237.

1. Create read-only versions of all the diagnostic sets. Sometimes a different setting forces a diagnostic set to a particular value. For example, when "Option strict" is on, several of the diagnostic sets will be treated as errors regardless of any other setting, and when all warnings have been disabled, the diagnostic sets will effectively all be "None" (disabled). In this case there is no reason to let the user choose a different value in the drop down; even if it makes a change to the project file, it won't change how the related diagnostics are handled in the end. The user would just end up in the situation where they make a change and the UI just switches back to the previous value.
2. Move to `DynamicEnumProperty`s so we can show the proper set of items. The diagnostic sets can generally have one of three values: "None", indicating the set is disabled; "Warning", indicating the diagnostics will be warnings (generally the default), and "Error", indicating the diagnostic has been promoted to an error. However, if `TreatWarningsAsErrors` is turned on, there are only two meaningful states: "None", and "Error"--the diagnostic can be disabled entirely, but if it is not it will be promoted to an error. In this case we don't want the UI to show "Warning" in the drop down menu, because it would just switch back to error. To accomplish this we make all of these `DynamicEnumProperty`s instead of `EnumProperty`s, and introduce the `VBDiagnosticSeverityEnumProvider`. This type uses a `Rule` (from `TreatWarningsAsErrorsRuleProvider`) to extract the value of the `TreatWarningsAsErrors` property in the project, and chooses the set of enum items accordingly.
3. Replace hard-coded links with fwlinks.

Note the second point involves moving the `SupportedValuesProvider` and `SingleRuleSupportedValuesProvider ` down to the Microsoft.VisualStudio.ProjectSystem.Managed project. This reflects that these are generally useful base types rather than being specific to target framework drop downs.

VBWarningDropDowns2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8196)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/TreatWarningsAsErrorsRuleProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.Rules;
+
+/// <summary>
+/// Defines a <see cref="Rule"/> to extract information to support the <see cref="VBDiagnosticSeverityEnumProvider"/>.
+/// </summary>
+[ExportRuleObjectProvider(nameof(TreatWarningsAsErrorsRuleProvider), PropertyPageContexts.Project)]
+[AppliesTo(ProjectCapability.VisualBasic)]
+[Order(0)]
+internal sealed class TreatWarningsAsErrorsRuleProvider : IRuleObjectProvider
```
tmeschter, line 14: I know I previously said I didn't want to introduce yet another way for us to ship `Rule` objects in our repo, but the truth is that defining this `Rule` in code was way easier than creating a new .xaml file and making sure it got plumbed through the build and setup authoring properly. I can still pull it out to a .xaml file if there are strong objections to this.

melytc (approved):
Looks good!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/TreatWarningsAsErrorsRuleProvider.cs
MiYanni: I'm currently working on those build and setup authoring processes. I can confirm they are not as easy as just writing code.

drewnoakes (approved):
This leaves me wondering if we should re-use the visibility conditions for editability too. There'd be a lot less duplication here that way. I don't think it'd be too hard to implement that feature in the UI. The data model wouldn't need to change.

drewnoakes:
> Presumably we would want to expose a "ReadonlyCondition" counterpart to the "VisibilityCondition" in the data model?

Ah yes, I forgot we had a dedicated CLR property for that. We _could_ stick a new condition in the metadata dictionary, but adding it to the project model would be more consistent.

drewnoakes:
I filed https://github.com/dotnet/project-system/issues/8204 to track the idea.

Issue #7379, opened 2021-07-01
Re-create VB WPF Application property page


drewnoakes:
We need to recreate the VB Application property page for WPF projects in the new project properties system.

---

This is the WPF Application property page for a legacy project (.NET Framework):

![image](https://user-images.githubusercontent.com/350947/124107919-0d283880-daa9-11eb-9408-acb965d4e6b7.png)

Highlighted areas are features the new experience is currently missing.

---

This is the WPF Application property page for a .NET Core (SDK-style) project:

![image](https://user-images.githubusercontent.com/350947/124108307-727c2980-daa9-11eb-99e9-7da8ef59d837.png)

It has fewer features than the .NET Framework equivalent.

tmeschter:
Really there are three WPF-specific items here:

- "Startup URI" drop down
- "Shutdown mode" drop down
- "Edit XAML" button

The last opens the Application.xaml file in the text editor (as opposed to the design mode). That's pretty straight forward.

The other two end up changing values in the Application.xaml file, which can also be done by editing the .xaml itself. In that case you will get XAML Intellisense support.

The "View Windows Settings" and "View Application Events" buttons seem to function the same as they do in WinForms apps.

We may be able to get away without this, or at least it isn't the highest priority.

Issue #7236, opened 2021-05-18
Re-create VB WinForms Application property page (4)


tmeschter:
We need to recreate the VB Application property page for WinForms projects in the new project properties system.

For reference, this is the Application property page as it currently appears for VB WinForms apps targeting .NET 5:

![image](https://user-images.githubusercontent.com/10506730/118733539-5110f800-b7f1-11eb-8b85-cd02796b6ea0.png)

Note this differs from the C# Application property page in some key ways:

- Different handling of resources
- The "Assembly Information..." button that opens a separate dialog for changing assembly information. It is not clear if this applies to .NET Core projects or only makes sense for .NET Framework projects.
- The "View Windows Settings" button to open the app.config file.
- "Enable application framework" and all the related settings. These are interesting as they may change settings in code or the app.config file rather than in the .vbproj file.

There are a couple of options for implementing this. C# and VB could share a "base" Application page with all the common settings, and then each have their own "extending" page that adds the language-specific items. Or, they could each have their own totally separate "Application" page, with a few properties duplicated. I think it really only matters if we expect there to be additional common settings that we wish to add in the future.

_This work was identified as part of #6935._

drewnoakes:
Several of these values are stored in a `.myapp` file.

RussKie:
The values are stored in myapp file, and AFAIK, changes at this dialog invoke an "old-style" source generator, which emits necessary code into Application.Desiger.vb.

@KlausLoeffelmann @KathleenDollard we could consider moving to the Roslyn source generators (similar to https://github.com/dotnet/winforms/issues/4909), as well as moving away from myapp file in favour of vbproj MSBuild properties. 
Then maintaining the VS propoerty page experience wouldn't be as daunting experience as it is today.
Thoughts?

paul1956:
As long as we can do a 1 time upgrade it is a fantastic idea, if we can't do it automatically some instructions on what a developer needs to do is an OK option. Right now in Core there is come confusion on how some of the properties should even be set like StartupObject, setting it wrong loses the StartUp screen for example and default Icon does nothing on this screen. What does it even mean in Core to turn off Enable XP visual styles. All the old properties that were in "Assembly Information..." are used in SplashScreen form and About Form are still hard to discover, I believe some are on package screen and stored in vbproj file but they don't populate "My" and some don't appear anywhere, for example.
```
<AssemblyVersion>5.0.3.4</AssemblyVersion>
<FileVersion>5.0.3.4</FileVersion>
[...]

PR #8197, opened 2022-05-23
ApplicationFramework property interceptor.

melytc:
Related to https://github.com/dotnet/project-system/issues/7379 and https://github.com/dotnet/project-system/issues/7236.

The `UseApplicationFramework` property gets saved to the project file as `<MyType>`; when enabled, the value `WindowsForms` is saved; when disabled, the value `WindowsFormsWithCustomSubMain` is saved.

To test it, enable the `ProjectPropertiesEditor` ProjectCapability in the Ms.VisualBasic.DesignTime.targets file.

Note: changes to this property also ipact the myapp file; I'll bring this up in an upcoming PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8197)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
@@ -30,25 +30,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
 
         public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
         {
-            var value = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkMSBuildProperty);
+            string? value = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkMSBuildProperty);
 
-            if (value == EnabledValue)
+            return value switch
             {
-                return "true";
-            }
-            else if (value == DisabledValue)
-            {
-                return "false";
-            }
-            else
-            {
-                return string.Empty;
-            }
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+                return "true";
+            }
+            else if (value == DisabledValue)
+            {
+                return "false";
+            }
+            else
+            {
+                return string.Empty;
+            }
```
drewnoakes, line 64: You can use a switch expression here if you like. Ditto above.

```suggestion
            return value switch
            {
                EnabledValue => "true",
                DisabledValue => "false",
                _ => string.Empty
            };
```

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }
+            return null;
+        }
+
+        public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
+        {
+            var value = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkMSBuildProperty);
+
+            if (value == EnabledValue)
```
ocallesp, line 35: The switch-case statement in `OnGetUnevaluatedPropertyValueAsyn` is doing the same logic as in this IF statement. Please use the same implementation.

tmeschter (approved):

PR #8198, opened 2022-05-23
Expose a mechanism to apply custom hot reload updates

BillHiebert:
- In order to support static asset file updates in Winforms and WPF projects, I added a new interface which provides a generic mechanism to apply those updates to all running hot reload sessions, without exposing internal details of the project system.
- The IProjectHotReloadUpdateApplier interface is exported from IProjectHotReloadSessionManager since this is the object which track hot reload sessions. It has a boolean method which returns true if sessions are active, and a second method to apply updates.
- The apply update function takes an async delegate which will be called once for each active session.  The delegate is passed the active IDeltaApplier for that session, and the assumption is the delegate can apply its update via that object.
- Doing it this way means the manage project system is shielded from these more specific updates by providing access to its connection to the running application.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8198)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.HotReload.Components.DeltaApplier;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
+{
+    /// <summary>
+    /// Internal interface which Hot Reload sessions implement to provide access to the IDeltaApplier
+    /// </summary>
+    internal interface IProjectHotReloadSessionInternal
+    {
+        /// <summary>
+        /// Returns the delta applier for this session
+        /// </summary>
+        IDeltaApplier? DeltApplier { get; }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
@@ -12,6 +12,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
         /// <summary>
         /// Returns the delta applier for this session
         /// </summary>
-        IDeltaApplier? DeltApplier { get; }
+        IDeltaApplier? DeltaApplier { get; }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
@@ -1,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.HotReload.Components.DeltaApplier;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
+{
+    /// <summary>
+    /// Provides a mechanism to apply custom updates to the running hot reload sessions. Assumes all updates can be applied via the 
+    /// IDeltaApplier created for this session
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.System, Cardinality = Composition.ImportCardinality.ExactlyOne)]
+    public interface IProjectHotReloadUpdateApplier
+    {
+        /// <summary>
+        /// Returns true if there is at least one hot reload session active in the project
+        /// </summary>
+        bool HasActiveHotReloadSessions { get; }
+
+        /// <summary>
+        /// Called to apply a custom update via the IDeltaApplier. The applyFunction will be called once for every active session, and passed the IDeltaApplier
+        /// for that session. In the case of multiple sessions, updates will stop being applied on the first call to an applyFunction that throws an exception, and 
+        /// that exception propagates back to the caller.
+        /// </summary>
+        Task ApplyHotReloadUpdateAsync(Func<IDeltaApplier, CancellationToken, Task> applyFunction, CancellationToken cancelToken);
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
@@ -5,7 +5,7 @@ using Microsoft.VisualStudio.HotReload.Components.DeltaApplier;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
 {
-    internal class ProjectHotReloadSession : IManagedHotReloadAgent, IProjectHotReloadSession
+    internal class ProjectHotReloadSession : IManagedHotReloadAgent, IProjectHotReloadSession, IProjectHotReloadSessionInternal
     {
         private readonly string _variant;
         private readonly string _runtimeVersion;
@@ -38,6 +38,16 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
         // IProjectHotReloadSession
 
         public string Name { get; }
+
+        /// <inheritdoc />
+        public IDeltaApplier? DeltApplier
+        {
+            get
+            {
+                EnsureDeltaApplierforSession();
+                return _deltaApplier;
+            }
+        }
 
         public async Task ApplyChangesAsync(CancellationToken cancellationToken)
         {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
@@ -40,14 +40,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
         public string Name { get; }
 
         /// <inheritdoc />
-        public IDeltaApplier? DeltApplier
-        {
-            get
-            {
-                EnsureDeltaApplierforSession();
-                return _deltaApplier;
-            }
-        }
+        public IDeltaApplier? DeltaApplier => _deltaApplier;
 
         public async Task ApplyChangesAsync(CancellationToken cancellationToken)
         {
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
@@ -39,7 +39,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
 
         public string Name { get; }
 
-        /// <inheritdoc />
         public IDeltaApplier? DeltaApplier => _deltaApplier;
 
         public async Task ApplyChangesAsync(CancellationToken cancellationToken)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
@@ -381,7 +381,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.HotReload
         /// <inheritdoc />
         public async Task ApplyHotReloadUpdateAsync(Func<IDeltaApplier, CancellationToken, Task> applyFunction, CancellationToken cancelToken)
         {
-            using AsyncSemaphore.Releaser semaphoreRelease = await _semaphore.EnterAsync();
+            using AsyncSemaphore.Releaser semaphoreRelease = await _semaphore.EnterAsync(cancelToken);
 
             // Run the updates in parallel
             List<Task> updateTasks = new List<Task>();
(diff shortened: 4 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
```diff
@@ -39,6 +39,16 @@ internal class ProjectHotReloadSession : IManagedHotReloadAgent, IProjectHotRelo
 
         public string Name { get; }
 
+        /// <inheritdoc />
+        public IDeltaApplier? DeltApplier
+        {
+            get
+            {
+                EnsureDeltaApplierforSession();
```
jacdavis, line 47: >EnsureDeltaApplierforSession [](http://example.com/codeflow?start=16&length=28)

.Netfx design guidelines expect property getters to be side-effect free. Is this truly necessary here? What is the impact of EnsureDeltaApplierforSession? Would a method to do this be a better approach?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -27,6 +28,8 @@ internal class ProjectHotReloadSessionManager : OnceInitializedOnceDisposedAsync
         private HotReloadState? _pendingSessionState = null;
         private int _nextUniqueId = 1;
 
+        public bool HasActiveHotReloadSessions => _activeSessions.Count != 0;
```
jacdavis, line 31: >Count [](http://example.com/codeflow?start=66&length=5)

Since this isn't a concurrent dictionary, count isn't thread-safe. Does this matter in this case? Is this collection modified from multiple background tasks simultaneously?

jacdavis: Looking below, you managed the modification with the _semaphore semaphore. Do you care that there is potentially races on this count?

jacdavis (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
BillHiebert: I looked at the source for Dictionary.Count and in the way I am using it I think it is fine since I am just checking against 0.  This dictionary is modified when a hot reload session starts and ends.  I could wait on the lock to ensure it is not modified, but I am not sure it matters in this scenario since updates to the dictionary should be relatively infrequent.

jacdavis: Sounds good. Thanks for checking.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
BillHiebert: I'll remove this since the places where a session is started already call this function.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Internal interface which Hot Reload sessions implement to provide access to the IDeltaApplier
+    /// </summary>
+    internal interface IProjectHotReloadSessionInternal
+    {
+        /// <summary>
+        /// Returns the delta applier for this session
+        /// </summary>
+        IDeltaApplier? DeltApplier { get; }
```
kvenkatrajan, line 15: Sorry to nitpick, the propertyName is missing a character: Delt"a"Applier

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManagerTests.cs
```diff
@@ -145,6 +145,45 @@ public async Task WhenDebugSymbolsIsFalse_APendingSessionIsNotCreated()
...
+                // so if the property isn't set that's OK.
+            };
+
+            var activeConfiguredProject = CreateConfiguredProject(capabilities, propertyNamesAndValues);
+            var manager = CreateHotReloadSessionManager(activeConfiguredProject);
+
+            var environmentVariables = new Dictionary<string, string>();
+            await manager.TryCreatePendingSessionAsync(environmentVariables);
+
+            Assert.False(manager.HasActiveHotReloadSessions);
```
kvenkatrajan, line 184: Shouldn't this be Assert.True for : WhenActiveSession_HasSessionsReturnsTrue()

BillHiebert: Yes, I meant to remove that test. That function just creates a pending session but the actual session isn't created until the process starts and ActivateSessionAsync is called. Looking at the  code there is no good way to mock that since there are direct calls to Process.GetProcessId which throws if the process doesn't exist.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// Provides a mechanism to apply custom updates to the running hot reload sessions. Assumes all updates can be applied via the 
+    /// IDeltaApplier created for this session
+    /// </summary>
+    [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.System, Cardinality = Composition.ImportCardinality.ExactlyOne)]
+    public interface IProjectHotReloadUpdateApplier
+    {
+        /// <summary>
+        /// Returns true if there is at least one hot reload session active in the project
+        /// </summary>
+        bool HasActiveHotReloadSessions { get; }
```
drewnoakes, line 17: Is the idea that a caller would test this property before calling `ApplyHotReloadUpdateAsync`?

Would a single `Try*` method be more consistent with `IProjectHotReloadSessionManager.TryCreatePendingSessionAsync`?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Called to apply a custom update via the IDeltaApplier. The applyFunction will be called once for every active session, and passed the IDeltaApplier
+        /// for that session. In the case of multiple sessions, updates will stop being applied on the first call to an applyFunction that throws an exception, and 
+        /// that exception propagates back to the caller.
+        /// </summary>
+        Task ApplyHotReloadUpdateAsync(Func<IDeltaApplier, CancellationToken, Task> applyFunction, CancellationToken cancelToken);
```
drewnoakes, line 24: Would the caller want to know if this actually applied a change? This could return `false` if no session exists, for example.

drewnoakes (approved):

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManagerTests.cs
BillHiebert: We only care about active ones since they are the only ones that we can apply changes to

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -375,6 +378,26 @@ private async Task<bool> StopProjectAsync(HotReloadState hotReloadState, Cancell
             return true;
         }
 
+        /// <inheritdoc />
+        public async Task ApplyHotReloadUpdateAsync(Func<IDeltaApplier, CancellationToken, Task> applyFunction, CancellationToken cancelToken)
+        {
+            using AsyncSemaphore.Releaser semaphoreRelease = await _semaphore.EnterAsync();
+
+            foreach (HotReloadState sessionState in _activeSessions.Values)
```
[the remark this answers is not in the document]
BillHiebert: I was thinking about that as well and since they are independent connections they can be run in parallel.  I'll make that change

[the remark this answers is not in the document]
BillHiebert: pushed this change

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
tmeschter: This check could be useful if the caller needs to do expensive work before calling `ApplyHotReloadUpdateAsync`.

But do we really need this at all?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
tmeschter: We could have the function return an array of results, where the result type is user-defined. Something like:

`Task<T[]> ApplyHotReloadUpdateAsync<T>(Func<IDeltaApplier, CancellationToken, Task<T>> applyFunction, CancellationToken cancellationToken)`

The resulting array would be empty if there were no active sessions, and if there were issues applying the change you could get back some information about them. Would that be useful?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
```diff
@@ -39,6 +39,9 @@ internal class ProjectHotReloadSession : IManagedHotReloadAgent, IProjectHotRelo
 
         public string Name { get; }
 
+        /// <inheritdoc />
```
tmeschter, line 42: No need for this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Internal interface which Hot Reload sessions implement to provide access to the IDeltaApplier
+    /// </summary>
+    internal interface IProjectHotReloadSessionInternal
+    {
+        /// <summary>
+        /// Returns the delta applier for this session
+        /// </summary>
+        IDeltaApplier? DeltaApplier { get; }
```
tmeschter, line 15: Could we just add this to `IProjectHotReloadSession` rather than creating a new interface?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
BillHiebert: We do. It is the only mechanism to know whether there are active hot reload sessions and in the case of scoped CSS  files whether a design time build is needed to generate an updated css file. And even in the general case of tracking say static css files. We only want to process and send updates if the application is actually running

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
BillHiebert: See my previous comment. It is important to know ahead of time whether processing needs to occur

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadSessionInternal.cs
BillHiebert: I'm fine doing that. I was just avoiding changing existing interfaces

BillHiebert: Actually, that is a public interface which is why didn't change it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/IProjectHotReloadUpdateApplier.cs
drewnoakes: It's fine as-is, so long as the caller doesn't need to know 100% whether the update was applied. There is a race condition here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSessionManager.cs
```diff
@@ -375,6 +378,33 @@ private async Task<bool> StopProjectAsync(HotReloadState hotReloadState, Cancell
             return true;
         }
 
+        /// <inheritdoc />
+        public async Task ApplyHotReloadUpdateAsync(Func<IDeltaApplier, CancellationToken, Task> applyFunction, CancellationToken cancelToken)
+        {
+            using AsyncSemaphore.Releaser semaphoreRelease = await _semaphore.EnterAsync();
```
drewnoakes, line 384:
```suggestion
            using AsyncSemaphore.Releaser semaphoreRelease = await _semaphore.EnterAsync(cancelToken);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/HotReload/ProjectHotReloadSession.cs
BillHiebert: Removed

drewnoakes:
Looks good to me. @BillHiebert are you happy for this to be merged now?

BillHiebert:
@drewnoakes  I am fine with it. Thanks

PR #8199, opened 2022-05-23
Use VS platform's IVsService

### LINKED ISSUE #7680: Consume IVsService from VS
IVsService is being moved to VS in this PR:

https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/357502

Once that implementation is available to us, we should use it instead of our own.

drewnoakes:
Fixes #7680.

This was previously attempted in #7808 and later reverted in #7818 due to a bug on the platform side that has since been fixed.

---

The pattern we use for accessing VS services via MEF imports has been ported to VS platform.

This commit updates our package references and removes our implementation in favour of the platform one.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8199)


jacdavis (approved):
:shipit:

PR #8203, opened 2022-05-24
Port AppDotXamlDocument

tmeschter:
Port AppDotXamlDocument.vb in Microsoft.VisualStudio.Editors to Microsoft.VisualStudio.ProjectSystem.Managed.VS. In the old property pages for WPF projects, this type supports getting and setting the "Startup URI" and "Shutdown Mode" properties by reading and writing to the Application.xaml via an `IVsTextLines` instance. We will need the same functionality in the new property pages, but also the option to make changes that are not backwards compatible with the old pages. So rather than trying to understand and re-create the behavior in new code I've opted for a straight port of the existing code to C#.

I've opted to move it to C# with only minor changes so we can get the WPF properties working end-to-end as soon as possible, and then we have the option of iterating on the implemenation. So it's not pretty, and not how we would write it now, but we can refactor it later.

As yet nothing consumes this code.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8203)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
@@ -925,9 +925,11 @@ internal class AppDotXamlDocument : AppDotXamlDocument.IDebugLockCheck, AppDotXa
             // Replace the empty tag in the buffer with a start/end element tag
             // and the new value
             string newText =
-$@"<{_fullyQualifiedPropertyName}>
-{EscapeXmlString(value)}
-</{_fullyQualifiedPropertyName}>";
+                $"""
+                <{_fullyQualifiedPropertyName}>
+                {EscapeXmlString(value)}
+                </{_fullyQualifiedPropertyName}>
+                """;
 
             // We know where to replace, so go ahead and do it.
             replaceTextInstance.ReplaceText(DefinitionStart, DefinitionEndPlusOne, newText);
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
@@ -6,6 +6,7 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Xml;
 using Microsoft.VisualStudio.TextManager.Interop;
+using DiagnosticsDebug = System.Diagnostics.Debug;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.WPF;
 
@@ -33,6 +34,11 @@ internal class AppDotXamlDocument : AppDotXamlDocument.IDebugLockCheck, AppDotXa
     private const string ApplicationElementName = "Application";
     private const string StartupUriPropertyName = "StartupUri";
     private const string ShutdownModePropertyName = "ShutdownMode";
+
+    private const char SingleQuote = '\'';
+    private const char DoubleQuote = '"';
+
+    private static readonly char[] ClosingAngleBracketHelperCharacters = new[] { SingleQuote, DoubleQuote, '/', '>' };
 
     private readonly IVsTextLines _vsTextLines;
 
@@ -153,7 +159,7 @@ internal class AppDotXamlDocument : AppDotXamlDocument.IDebugLockCheck, AppDotXa
     /// </summary>
     private XmlTextReader CreateXmlTextReader()
     {
-        System.Diagnostics.Debug.Assert(_debugBufferLockCount > 0, "Should be using BufferLock!");
+        DiagnosticsDebug.Assert(_debugBufferLockCount > 0, "Should be using BufferLock!");
         StringReader stringReader = new(GetAllText());
         // Required by Fxcop rule CA3054 - DoNotAllowDTDXmlTextReader
         XmlTextReader xmlTextReader = new(stringReader) { DtdProcessing = DtdProcessing.Prohibit };
@@ -189,12 +195,10 @@ internal class AppDotXamlDocument : AppDotXamlDocument.IDebugLockCheck, AppDotXa
     {
         int index = 0;
 
-        const char singleQuote = '\'';
-        const char doubleQuote = '"';
         while (index < line.Length)
         {
             // Find the next character of interest
-            index = line.IndexOfAny(new[] { singleQuote, doubleQuote, '/', '>' }, index);
+            index = line.IndexOfAny(ClosingAngleBracketHelperCharacters, index);
             if (index < 0)
             {
                 return -1;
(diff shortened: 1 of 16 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
...
+
+    /// <summary>
+    /// Represents a property that was found in property element syntax with an empty tag,
+    /// e.g. &lt;Application.StartupUri/>
+    /// </summary>
+    private class XamlPropertyInPropertyElementSyntaxWithEmptyTag : XamlPropertyInPropertyElementSyntax
+    {
+        /// <summary>
+        /// 
+        /// </summary>
```
drewnoakes, line 906: Delete?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
...
+                // The property value is not changing. Leave things alone.
+                return;
+            }
+
+            // Replace the empty tag in the buffer with a start/end element tag
+            // and the new value
+            string newText =
+$@"<{_fullyQualifiedPropertyName}>
+{EscapeXmlString(value)}
+</{_fullyQualifiedPropertyName}>";
```
drewnoakes, line 930: C# 10's raw strings can be used to keep things indented.

```suggestion
            string newText =
                $"""
                <{_fullyQualifiedPropertyName}>
                {EscapeXmlString(value)}
                </{_fullyQualifiedPropertyName}>
                """;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Xml;
+using Microsoft.VisualStudio.TextManager.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WPF;
+
+/// <summary>
+/// When reading and writing the Application.xaml file, we have a requirement that
```
drewnoakes, line 13:
```suggestion
/// When reading and writing the App.xaml file, we have a requirement that
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
...
+    private static int FindClosingAngleBracketHelper(string line)
+    {
+        int index = 0;
+
+        const char singleQuote = '\'';
+        const char doubleQuote = '"';
+        while (index < line.Length)
+        {
+            // Find the next character of interest
+            index = line.IndexOfAny(new[] { singleQuote, doubleQuote, '/', '>' }, index);
```
drewnoakes, line 197: This array is allocated potentially many times per line of the file. Perhaps pull it out to a static field.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
...
+                    else
+                    {
+                        index = closingQuote + 1;
+                    }
+                    break;
+                case '>':
+                    // Found '>'
+                    return index;
+                case '/':
+                    if (line.Substring(index).StartsWith("/>"))
```
drewnoakes, line 223: Can avoid the substring allocation here.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Xml;
+using Microsoft.VisualStudio.TextManager.Interop;
```
drewnoakes, line 8: There are 11 qualified references to `System.Diagnostics.Debug` in this file. Perhaps:

```suggestion
using Microsoft.VisualStudio.TextManager.Interop;
using Debug = System.Diagnostics.Debug;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
```diff
@@ -0,0 +1,944 @@
...
+    void IDebugLockCheck.OnBufferUnlock()
+    {
+        _debugBufferLockCount--;
+        System.Diagnostics.Debug.Assert(_debugBufferLockCount >= 0, "Extra buffer unlock");
+    }
+
+    /// <summary>
+    /// Represents a position in a text buffer.
+    /// </summary>
+    private class Location
```
drewnoakes, line 697: Seems like this could be a `readonly struct` to reduce allocations.

drewnoakes (approved):
Looks good. I called out a few easy changes that could improve the implementation without deviating much from the VB implementation.

MiYanni (approved):
As the intent of the PR is to "get things going ASAP", I'm fine with it being a straight port. For the future, I'd highly advise against using by-hand reading/manipulation of the XML DOM in this way. It is something that isn't generally done anymore as libraries/tooling have advanced quite significantly. If you want, once this code is 'hooked up and running', I can take a stab at refactoring it. Might be nice to have some unit tests at that point too so that I know I'm not breaking things.

tmeschter:
> As the intent of the PR is to "get things going ASAP", I'm fine with it being a straight port. For the future, I'd highly advise against using by-hand reading/manipulation of the XML DOM in this way. It is something that isn't generally done anymore as libraries/tooling have advanced quite significantly. If you want, once this code is 'hooked up and running', I can take a stab at refactoring it. Might be nice to have some unit tests at that point too so that I know I'm not breaking things.

This definitely isn't how I would write it today, no. :-)

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/AppDotXamlDocument.cs
tmeschter: The file name is actually Application.xaml, not App.xaml.

drewnoakes: Is it not this file?

image

PR #8205, opened 2022-05-25
Code cleanup

ocallesp:
- Remove IRegistry
- Improve the description of RegistryReaderRuntimesVersions
- Remove incorrectly assigned null types
- Rename property names to better describe the purpose.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8205)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/RegistryReaderNetCoreRuntimesVersions.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.Win32;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Runtimes
+{
+    internal sealed class RegistryReaderNetCoreRuntimesVersions
```
MiYanni, line 7: nit: Use adjectives before the noun of the class name. Also, the double plural is a bit confusing. Suggested: `NetCoreRuntimeVersionsRegistryReader`
Some other places in the repo with long names you'll see an `Of` convention too, which would be like `RegistryReaderOfNetCoreRuntimeVersions`, but ending a class name with an plural (`s`) is also confusing, so I don't suggest this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -50,7 +48,7 @@ internal class MissingSetupComponentRegistrationService : IMissingSetupComponent
         private IVsSolution? _vsSolution;
         private bool? _isVSFromPreviewChannel;
 
-        private readonly object _lock = new();
+        private static readonly object _lock = new();
```
MiYanni, line 51: I don't think this should be static as you don't want to share a lock between instances of the same class. Each should have their own lock.

MiYanni (approved):
Approving as I won't be at work tomorrow. But definitely look at that `static` on the lock as I don't know if that is intended.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
ocallesp: removing statis since this is a service.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/RegistryReaderNetCoreRuntimesVersions.cs
ocallesp: changed.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/NetCoreRuntimeVersionsRegistryReader.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.Win32;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Runtimes
+{
+    internal sealed class NetCoreRuntimeVersionsRegistryReader
```
drewnoakes, line 7: This should be a `static` class.

PR #8206, opened 2022-05-25
Require valuedefinedincontext property for launch profile page

adamint:
This is to fix an issue where the property is not obtained in the launch profile page.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8206)


melytc:
Could you describe more about the context of the issue?

adamint:
> Could you describe more about the context of the issue?

ValueDefinedInContext is a required property for launch profile window, so trying to create a new profile leads to this:
54c694f9-eb47-4118-8d7a-84eec763a37c

We just need to add ValueDefinedInContext to the list of required properties for this window

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionBase.cs
```diff
@@ -79,6 +79,7 @@ static LaunchProfileActionBase()
             s_requestedValueProperties.RequireProperty(UIPropertyValueType.EvaluatedValuePropertyName);
             s_requestedValueProperties.RequireProperty(UIPropertyValueType.SupportedValuesPropertyName);
             s_requestedValueProperties.RequireProperty(UIPropertyValueType.UnevaluatedValuePropertyName);
+            s_requestedValueProperties.RequireProperty(UIPropertyValueType.ValueDefinedInContextPropertyName);
```
drewnoakes, line 82: If I understand correctly, this is the primary change here.

Other changes are consequences of updating the CPS package and consuming its updated nullable annotations.

drewnoakes (approved):

PR #8208, opened 2022-05-27
Code coverage adjustment

MiYanni:
Replaced the logic to determine if we want to publish CodeCov during the build to use a boolean variable.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8208)


kvenkatrajan (commented):
If UsingCodeCoverage is a boolean value, do we not need to compare against true/false instead of empty string?

drewnoakes (approved):

PR #8210, opened 2022-05-28
WIP: intercept application framework properties and save them in the myapp file.

melytc:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8210)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
@@ -1,32 +1,36 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System.Xml.Linq;
+using Microsoft.VisualStudio.Shell.Design.Serialization;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms;
 
 internal class MyAppDocument
 {
-    private readonly XDocument doc;
-    private readonly XElement root;
+    private XDocument _doc;
+    private XElement _root;
+    private readonly DocDataTextReader _reader;
 
-    private readonly string name;
+    private readonly string _fileName;
 
-    public MyAppDocument(string fileName, string filePath)
+    public MyAppDocument(string fileName, DocDataTextReader textReader)
     {
-        var xmlString = File.ReadAllText(filePath);
-        doc = XDocument.Parse(xmlString);
-        root = doc.Root;
-        name = fileName;
+        _reader = textReader;
+        _doc = XDocument.Load(_reader);
+        _root = _doc.Root;
+        _fileName = fileName;
     }
 
-    public string GetProperty(string propertyName)
+    public string GetProperty(string filePath, string propertyName)
     {
-        return (string)root.Element(propertyName);
+        _doc = XDocument.Load(_reader);
+        _root = _doc.Root;
+        return (string)_root.Element(propertyName);
     }
 
     public void SetProperty(string propertyName, string propertyValue)
     {
-        root.Element(propertyName).Value = propertyValue;
-        doc.Save(name);
+        _root.Element(propertyName).Value = propertyValue;
+        _doc.Save(_fileName);
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Xml.Linq;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms;
+
+internal class MyAppDocument
+{
+    private readonly XDocument doc;
+    private readonly XElement root;
+
+    private readonly string name;
+
+    public MyAppDocument(string fileName, string filePath)
+    {
+        var xmlString = File.ReadAllText(filePath);
+        doc = XDocument.Parse(xmlString);
+        root = doc.Root;
+        name = fileName;
+    }
+
+    public string GetProperty(string propertyName)
+    {
+        return (string)root.Element(propertyName);
+    }
+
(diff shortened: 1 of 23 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms;
+
+internal class MyAppDocument
+{
+    private readonly XDocument doc;
+    private readonly XElement root;
+
+    private readonly string name;
+
+    public MyAppDocument(string fileName, string filePath)
```
drewnoakes, line 14: I think you'll need to parse the file each time you get/set a property, rather than once at startup. You cannot assume that external systems won't modify the file over time, and because you keep an instance of `​MyAppDocumen`t​ alive in ​`MyAppFileProcessor`​ (which has the lifetime of the project) any changes made after it's parsed won't be reflected in the UI, for example.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        var xmlString = File.ReadAllText(filePath);
+        doc = XDocument.Parse(xmlString);
```
drewnoakes, line 17:
```suggestion
        doc = XDocument.Load(filePath);
```

This allows `XDocument` to parse the file as it streams from disk, rather than allocating its entire contents in a string before parsing.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -0,0 +1,181 @@
...
+        _project = project;
+        _serviceProvider = serviceProvider;
+        _runningDocumentTable = runningDocumnetTable;
+        _threadingService = threadingService;
+
+        _myappDocument = new MyAppDocument("Application.myapp", "filePath"); //TODO: get actual file name and file path.
+    }
+    public async Task<bool> GetMySubMainAsync()
+    {
+        await _threadingService.SwitchToUIThread();
```
drewnoakes, line 45: Do you need to switch to the UI thread in these methods?

If you do parse the file's contents on every read, you'll want to do that on the thread pool.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/WinForms/IMyAppXamlFileAccessor.cs
```diff
@@ -0,0 +1,109 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms;
+
+/// <summary>
+/// Provides access to properties in the myapp file.
+/// </summary>
+internal interface IMyAppXamlFileAccessor
```
drewnoakes, line 8: As it turns out this isn't a XAML file, could this just be `IMyAppFileAccessor`?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkPropertiesValueProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+        return propertyName switch
+        {
+            ApplicationFrameworkProperty => _myAppXamlFileAccessor.SetMySubMainAsync(unevaluatedPropertyValue),
+            SingleInstanceProperty => _myAppXamlFileAccessor.SetSingleInstanceAsync(Convert.ToBoolean(unevaluatedPropertyValue)),
+            SaveMySettingsOnExitProperty => _myAppXamlFileAccessor.SetSaveMySettingsOnExitAsync(Convert.ToBoolean(unevaluatedPropertyValue)),
+            HighDpiModeProperty => _myAppXamlFileAccessor.SetHighDpiModeAsync(Convert.ToInt16(unevaluatedPropertyValue)),
+            AuthenticationModeProperty => _myAppXamlFileAccessor.SetAuthenticationModeAsync(Convert.ToInt16(unevaluatedPropertyValue)),
+            ShutdownModeProperty => _myAppXamlFileAccessor.SetShutdownModeAsync(Convert.ToInt16(unevaluatedPropertyValue)),
+            SplashScreenProperty => _myAppXamlFileAccessor.SetSplashScreenAsync(unevaluatedPropertyValue),
+            MinimumSplashScreenDisplayTimeProperty => _myAppXamlFileAccessor.SetMinimumSplashScreenDisplayTimeAsync(Convert.ToInt16(unevaluatedPropertyValue)),
```
drewnoakes, line 59: These `Convert.To*` methods will throw if the input string isn't convertible. Is there any chance a non-conforming string would be received here, and if so what should we do about it?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppDocument.cs
melytc: Yeah, that's a good point. I added that change in #8252

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
melytc: Removed it in #8252 , thanks Drew!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkPropertiesValueProvider.cs
melytc: I see that happening on the Get method, when a property is not defined in the myapp file yet; but for the Set method, I don't see that happening (but maybe I'm wrong).

PR #8215, opened 2022-06-01
Fix overbuild on project load

### LINKED ISSUE #8212: Overbuild on project load in some cases
The fast up-to-date check can cause overbuild in the following scenario:

- Open and build project
- Add item to project
- Build project again
- Close / reopen project
- Build again

That last build should be up-to-date but is not.

> The set of project items was changed more recently (25/05/2022 22:10:07) than the last successful build start time (01/01/0001 11:00:00), not up-to-date.

This leads to a single instance of overbuild for the project. From that point on, incremental build works correctly.

drewnoakes:
Fixes #8212.

Changes in 17.2 intended to fix an overbuild loop inadvertently introduced a single occurrence of overbuild for the first build of a project, in some cases.

Those changes allow the fast up-to-date check to track the time at which the last successful build started, and use that timestamp in comparisons for other kinds of changes to know whether those changes should trigger subsequent builds.

The issue arose because the "last successful build start time" value is not persisted between sessions. On reloading a project, the value is `DateTime.MinValue`, which is always before the last set of item changes, and therefore triggers a build. That build then updates the in-memory "last successful build start time", and from that point on the project behaves as desired.

Changes in this commit:

- Persist the "last successful build time" value along with the other state we keep.
- Bump version of `.futdcache` file, as we now store additional information.
- Make `DateTime` values nullable to it's clearer that we need to handle the missing-value case, and remove use of `DateTime.MinValue`.
- Remove redundant `WasStateRestored` property from our snapshot. It's enough to test for a null timestamp in the place that property was read.
- Move handling of build events off the main thread, on to the thread pool.
- Update and extend documentation in code.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8215)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckBuildEventNotifier.cs
```diff
@@ -113,7 +136,7 @@ private IEnumerable<IBuildUpToDateCheckProviderInternal> FindActiveConfiguredPro
                 }
             }
 
-            return Enumerable.Empty<IBuildUpToDateCheckProviderInternal>();
+            return null;
```
drewnoakes, line 139: This was changed to return `null` so that a consumer could avoid a JTF.RunAsync call when no providers exist.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
```diff
@@ -21,9 +21,9 @@ internal sealed partial class UpToDateCheckStatePersistence : OnceInitializedOnc
         // The name of the file within the .vs folder that we use to store data for the up-to-date check.
         // Note the suffix that indicates the file format's version. If a change is made to the format,
         // this number must be bumped.
-        private const string ProjectItemCacheFileName = ".futdcache.v1";
+        private const string ProjectItemCacheFileName = ".futdcache.v2";
```
drewnoakes, line 24: Version number of the persistence file bumped.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
```diff
@@ -63,64 +63,75 @@ protected override async Task DisposeCoreUnderLockAsync(bool initialized)
...
+                    {
+                        return (storedData.ItemHash, storedData.ItemsChangedAtUtc);
+                    }
 
                     return null;
                 },
                 cancellationToken);
+        }
 
-            async Task InitializeDataAsync(CancellationToken cancellationToken)
```
drewnoakes, line 84: This local function was promoted to class level. Makes the diff a bit messy here, as another pair of Store/Restore methods were added in this location.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Subscription.cs
```diff
@@ -33,26 +37,15 @@ private sealed class Subscription : ISubscription
...
-            /// </summary>
-            /// <remarks>
-            /// This value is required in order to protect against a race condition described in
-            /// https://github.com/dotnet/project-system/issues/4014. Specifically, if source files are
-            /// modified during a compilation, but before that compilation's outputs are produced, then
-            /// the changed input file's timestamp will be earlier than the compilation output, making
-            /// it seem as though the compilation is up to date when in fact the input was not included
-            /// in that compilation. Comparing against compilation start time fixes that issue.
-            /// </remarks>
-            private readonly Dictionary<ProjectConfiguration, DateTime> _lastSuccessfulBuildStartTimeUtcByConfiguration = new();
```
drewnoakes, line 54: This state was moved to `IUpToDateCheckStatePersistence`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -27,6 +27,9 @@ internal sealed partial class BuildUpToDateCheck
         internal const string FastUpToDateCheckIgnoresKindsGlobalPropertyName = "FastUpToDateCheckIgnoresKinds";
         internal const string TargetFrameworkGlobalPropertyName = "TargetFramework";
 
+        // This analyzer fires for comparisons against the following constants. Disable it in this file.
+        #pragma warning disable CA1820 // Test for empty strings using string length
```
drewnoakes, line 31: This suppression is unrelated to this PR but cleans up some diagnostics in the IDE.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -116,10 +122,15 @@ private bool CheckGlobalConditions(Log log, DateTime lastSuccessfulBuildStartTim
                 return log.Fail("Disabled", nameof(Resources.FUTD_DisableFastUpToDateCheckTrue));
             }
 
-            if (validateFirstRun && !state.WasStateRestored && lastSuccessfulBuildStartTimeUtc == DateTime.MinValue)
+            if (validateFirstRun && lastSuccessfulBuildStartTimeUtc is null)
```
drewnoakes, line 125: `WasStateRestored` has been removed. It's enough to check whether `lastSuccessfulBuildStartTimeUtc` is `null` or not.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
```diff
@@ -143,16 +153,54 @@ public Task StoreStateAsync(string projectPath, IImmutableDictionary<string, str
...
+                await TaskScheduler.Default;
+
+                if (_cacheFilePath is null || _dataByConfiguredProject is null)
+                {
+                    _cacheFilePath = filePath;
+                    _dataByConfiguredProject = Deserialize(_cacheFilePath);
+                }
+            }
+
+            return;
```
MiYanni, line 172: This shouldn't be needed, correct?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -266,10 +280,12 @@ bool CheckInputsAndOutputs(IEnumerable<(string Path, string? ItemType, bool IsRe
...
                     }
 
-                    if (inputTime > lastSuccessfulBuildStartTimeUtc && lastSuccessfulBuildStartTimeUtc != DateTime.MinValue)
+                    if (inputTime > lastSuccessfulBuildStartTimeUtc)
                     {
                         // Bypass this test if no check has yet been performed. We handle that in CheckGlobalConditions.
-                        return log.Fail("InputModifiedSinceLastSuccessfulBuildStart", itemType is null ? nameof(Resources.FUTD_InputModifiedSinceLastSuccessfulBuildStart_3) : nameof(Resources.FUTD_TypedInputModifiedSinceLastSuccessfulBuildStart_4), input, inputTime.Value, lastSuccessfulBuildStartTimeUtc, itemType ?? "");
+                        Assumes.NotNull(inputTime);
+                        Assumes.NotNull(lastSuccessfulBuildStartTimeUtc);
+                        return log.Fail("InputModifiedSinceLastSuccessfulBuildStart", itemType is null ? nameof(Resources.FUTD_InputModifiedSinceLastSuccessfulBuildStart_3) : nameof(Resources.FUTD_TypedInputModifiedSinceLastSuccessfulBuildStart_4), input, inputTime, lastSuccessfulBuildStartTimeUtc, itemType ?? "");
```
MiYanni, line 288: nit: Consider making parameters multi-line.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckBuildEventNotifier.cs
drewnoakes: I don't like to use `Any` before enumerating. It's unclear whether you are double enumerating the source or not in such cases. For example, can you tell by looking at the code whether it is wasteful to enumerate the return value of `GetExportedValues` more than once?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UpToDate/UpToDateCheckStatePersistence.cs
drewnoakes: Correct. It's a convention used in several places to separate the method's statements from any local functions. The idea is it makes it easier to see that there are no statements hidden later on in the method, as it would be otherwise valid to have statements after local functions.

PR #8216, opened 2022-06-01
Extend fast up-to-date check telemetry

### LINKED ISSUE #8213: Extend fast up-to-date telemetry
Investigation into incremental build performance of 17.2 identified some extra data that would be helpful in telemetry:

- Include FUTD check number for the project since load. This would be 1 for the first build, and increase thereafter. Unloading the project would reset this back to 1 for the next check. Having this data would help flag issues specific to the first build after solution load, or potential fluctuations in the performance or behaviour of the check over lifetime of the project.
- Include (hashed) project identifier. This would allow comparing behaviour of the check between versions of VS.

Tasks

- [x] Code change (PR #8216)
- [x] GDPR categorisation of new values
- [x] Update DataChef (PR https://devdiv.visualstudio.com/DevDiv/_git/DataChef/pullrequest/403465)

drewnoakes:
Contributes to #8213 

Adds:
- Project identifier
- Check number

See linked issue for context.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8216)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -44,6 +44,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
         private readonly IProjectAsynchronousTasksService _tasksService;
         private readonly ITelemetryService _telemetryService;
         private readonly IFileSystem _fileSystem;
+        private readonly ISafeProjectGuidService _guidService;
         private readonly IUpToDateCheckHost _upToDateCheckHost;
 
         private IImmutableDictionary<string, string> _lastGlobalProperties = ImmutableStringDictionary<string>.EmptyOrdinal;
@@ -61,6 +62,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             [Import(ExportContractNames.Scopes.ConfiguredProject)] IProjectAsynchronousTasksService tasksService,
             ITelemetryService telemetryService,
             IFileSystem fileSystem,
+            ISafeProjectGuidService guidService,
             IUpToDateCheckHost upToDateCheckHost)
         {
             _inputDataSource = inputDataSource;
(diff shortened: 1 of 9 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -903,6 +920,13 @@ async Task<(bool, ImmutableArray<ProjectConfiguration>)> CheckAsync(UpToDateChec
                     _lastFailureReason = logger.FailureReason ?? "";
                 }
             }
+
+            static async Task<(T1, T2)> WhenAll<T1, T2>(Task<T1> t1, Task<T2> t2)
+            {
+                await Task.WhenAll(t1, t2);
+
+                return (await t1, await t2);
```
MiYanni, line 928: Since the tasks have already been completed, is the use of `await` here for each task to simply get the result value of each task?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: That's right. Could also write `.GetAwaiter().GetResult()` but I don't like seeing that in code.

I would like to pull this out into something a bit more generic and reusable, to make it easier to run multiple async operations concurrently. Just need to find the right API for that. I suspect it'd take a custom awaiter implementation.

MiYanni: Discussion about this idea: https://github.com/dotnet/csharplang/discussions/380
And someone made a library to do something similar: https://github.com/buvinghausen/TaskTupleAwaiter/
Package on NuGet: https://www.nuget.org/packages/TaskTupleAwaiter/1.0.1

drewnoakes: Thanks. I've filed https://github.com/dotnet/project-system/pull/8223 to make this more generally reusable.

Issue #8170, opened 2022-05-12
Support AppliesTo filtering for property interceptors


drewnoakes:
The reading and writing of property values can be intercepted by exports of `IInterceptingPropertyValueProvider` as discussed in [this document](https://github.com/dotnet/project-system/blob/main/docs/repo/property-pages/property-value-interception.md).

Today we match the property name against the property definition in order to find the correct export for a given property.

This may be problematic in cases where we have properties with the same name, but in different project types.

For example, perhaps C# and VB projects have the same property, but treat them differently.

The classical way of achieving this is via `[AppliesTo("<expression>")]` metadata on the export, where `<expression>` describes the _project capabilities_ to which the export applies.

This issue is tracking the idea of allowing (or possibly even requiring) this metadata on exports in order to match them to properties.

---

The challenge with doing this today is that no current exports have `AppliesTo` metadata. The current capability filtering treats the absence of `AppliesTo` as meaning no match can happen.

There are two ways we could allow this metadata to be added:

1. Update _every_ existing export of `IInterceptingPropertyValueProvider` to have this metadata. This includes all those in this repository, JavaScript/TypeScript projects, MAUI projects, Xamarin projects and possibly others.
2. Create a new mechanism for capability filtering that treats the absence of `AppliesTo` as automatically _including_ rather than excluding the export. This seems like something that could be generally useful in future too.

adamint:
relevant to #8387

PR #8217, opened 2022-06-02
Add project capability for the property system

drewnoakes:
Non-.NET projects want to be able to use the property infrastructure, such as .esproj.

Today, they're unable to use property interception because they don't define the .NET project capability.

This change introduces a new capability for these features, and adds that capability to managed projects. This allows other (non-.NET) project types to add this capability to their own design-time targets.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8217)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets
@@ -92,6 +92,7 @@
     <ProjectCapability Include="ProjectImportsTree" />
     <ProjectCapability Include="LaunchProfiles" />
     <ProjectCapability Include="NoGeneralDependentFileIcon"/>
+    <ProjectCapability Include="ProjectPropertyInterception" />
     <ProjectCapability Include="PackageReferences" Condition="'$(RestoreProjectStyle)' == 'PackageReference'"/>
     
     <!--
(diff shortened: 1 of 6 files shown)


tmeschter (commented):
Most of the "properties" subsystem comes from CPS; what we're really adding is the ability to intercept and alter values on their way to one of the standard property providers. So if we add a new capability I would suggest something more specific that "PropertySystem", like "InterceptedProjectProperties".

We may also be able to use a broader, existing capability. For example, most of the `IProjectPropertyProvider`s implemented by CPS have `[AppliesTo(ProjectCapabilities.AlwaysApplicable)]` meaning they are present for _all_ projects. We could probably get away with the same thing as I don't think a particular `IProjectPropertyProvider` implementation is instantiated until it is actually used. However, attaching this functionality to a particular capability is a reasonable and conservative choice.

tmeschter:
We would need to deal with #8170 at the same time; otherwise we risk non-.NET projects picking up .NET-specific interceptors.

MariaSolOs:
I love this idea given that it represents a great opportunity for extensibility in CPS and reduce code duplication. Thank you for starting this discussion!

drewnoakes:
I went with `ProjectPropertyInterception` for the capability name.

> most of the IProjectPropertyProviders implemented by CPS have [AppliesTo(ProjectCapabilities.AlwaysApplicable)] meaning they are present for all projects

My concern with using this is that we'd inadvertently cause the .NET Project System assemblies to load in cases where it shouldn't, for example in a C++ solution. Having a capability prevents that problem.

> We would need to deal with https://github.com/dotnet/project-system/issues/8170 at the same time; otherwise we risk non-.NET projects picking up .NET-specific interceptors.

I don't see how the timing of these two changes would be coupled. Could you elaborate? It seems that this PR doesn't materially change anything around #8170.

I'm marking this as ready for review. It won't make 17.3p2 at this point.

drewnoakes:
> Only in that the intercepting property providers were effectively limited to .NET projects

I see, thanks. My sense is that components have to opt-in to interception specifically on a per-property basis, so this shouldn't be a problem in practice. I do want us to address #8170, and will discuss this with CPS in the next sync meeting.

tmeschter (approved):

PR #8220, opened 2022-06-02
Add WPF-related properties

tmeschter:
Related to #7379.

1. Update the `AppDotXamlDocument` type to support reading and writing the "Startup URI" and "Shutdown Mode" properties. This also involves a change to return `null` when one of these properties is not set in the file, rather than the empty string. This lets the caller distinguish between the value not being set, and being set to an actual empty value.
2. Introduce an `IApplicationXamlFileAccessor` interface and related implementation to manage access to a project's Application.xaml file. This type handles loading the Application.xaml file into a `DocData`, managing the lifetime of the `DocData`, and creating the `AppDotXamlDocument` as needed.
3. Add the `WPFValueProvider` interceptor that uses the `IApplicationXamlFileAccessor` to expose "Startup URI" and "Shutdown Mode" as properties.
4. Update the VB Application property page to include the WPF-related properties, but only show them for projects that produce executables, use WPF, and use the Application Framework.
5. Update the `ApplicationFrameworkValueProvider` to support WPF projects. Checking or unchecking the "Use Application Framework" check box has a much different effect in WPF than other projects; we need to toggle the "StartupObject" property between "Sub Main" (when the application framework is off) and undefined (when it is), and toggle whether or not the Application.xaml file is handled as an "ApplicationDefinition" item type or not. In this commit we basically create two totally different code paths for reading/writing: one for WPF projects, and one for everything else. We may want to revisit this to clean it up.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8220)

### DIFF
--- a/build/import/VisualStudio.props
+++ b/build/import/VisualStudio.props
@@ -25,6 +25,7 @@
     <PackageReference Include="Microsoft.VisualStudio.RpcContracts" />
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" />    
     <PackageReference Include="Microsoft.VisualStudio.Shell.15.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Design" />
     <PackageReference Include="Microsoft.VisualStudio.Shell.Framework" />
     <PackageReference Include="Microsoft.VisualStudio.Telemetry" />
     <PackageReference Include="Microsoft.VisualStudio.Utilities" />
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/ApplicationXamlFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/ApplicationXamlFileAccessor.cs
@@ -81,11 +81,6 @@ internal class ApplicationXamlFileAccessor : IApplicationXamlFileAccessor
         }
     }
 
-    public async Task<bool> ApplicationXamlFileExistsAsync()
-    {
-        return !string.IsNullOrEmpty(await GetFilePathAsync(create: false));
-    }
-
     private async Task<AppDotXamlDocument?> TryGetApplicationXamlFileAsync(bool create)
     {
         if (_applicationDotXamlDocument is null)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/ApplicationXamlFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/ApplicationXamlFileAccessor.cs
@@ -1,7 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using System.Runtime.InteropServices;
-using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.WPF;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Design.Serialization;
 using Microsoft.VisualStudio.Shell.Interop;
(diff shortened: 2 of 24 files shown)


build/import/VisualStudio.props
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="Microsoft.VisualStudio.RpcContracts" />
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" />    
     <PackageReference Include="Microsoft.VisualStudio.Shell.15.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Design" />
```
tmeschter, line 28: This reference was now needed in the Microsoft.VisualStudio.ProjectSystem.Managed.VS project, not just Microsoft.VisualStudio.Editors and Microsoft.VisualStudio.AppDesigner.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/ApplicationXamlFileAccessor.cs
```diff
@@ -0,0 +1,190 @@
...
+    {
+        _project = project;
+        _serviceProvider = serviceProvider;
+        _runningDocumentTable = runningDocumentTable;
+        _threadingService = threadingService;
+    }
+
+    public async Task<string?> GetStartupUriAsync()
+    {
+        await _threadingService.SwitchToUIThread();
```
tmeschter, line 39: We serialize access to the `_applicationDotXamlDocument` and related data structures by always switching to the UI thread. The `DocData` needs to be accessed from that thread, anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WPF/ApplicationXamlFileAccessor.cs
```diff
@@ -0,0 +1,190 @@
...
+                {
+                    _applicationDotXamlDocument = new AppDotXamlDocument(vsTextLines);
+                }
+            }
+        }
+
+        return _applicationDotXamlDocument;
+    }
+
+    private void SaveDocDataIfOnlyEditor()
```
tmeschter, line 107: This attempts to re-create some behavior in the old property pages that may not be needed. If we're the only ones with the Application.xaml file open for editing we'll save any pending changes; but it the file is held open by others (e.g., it is open in an editor) we won't.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
+    {
+        return await GetPropertyValueAsync(propertyName);
+    }
+
+    public override async Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties)
+    {
+        return await GetPropertyValueAsync(propertyName);
+    }
```
drewnoakes, line 35: Ditto, state machines.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+                await _applicationXamlFileAccessor.SetStartupUriAsync(unevaluatedPropertyValue);
+                break;
+
+            case ShutdownModePropertyName:
+                await _applicationXamlFileAccessor.SetShutdownModeAsync(unevaluatedPropertyValue);
+                break;
+
+            default:
+                throw new InvalidOperationException($"The {nameof(WPFValueProvider)} does not support the '{propertyName}' property.");
+        }
```
drewnoakes, line 51: Could also use a switch expression here, to match the below.

```suggestion
        await propertyName switch
        {
            StartupURIPropertyName => _applicationXamlFileAccessor.SetStartupUriAsync(unevaluatedPropertyValue),
            ShutdownModePropertyName => _applicationXamlFileAccessor.SetShutdownModeAsync(unevaluatedPropertyValue),
            _ => throw new InvalidOperationException($"The {nameof(WPFValueProvider)} does not support the '{propertyName}' property.")
        };
```

drewnoakes (approved):

MiYanni (approved):

PR #8223, opened 2022-06-03
Add extension for awaiting tuples of tasks

drewnoakes:
Following discussion: https://github.com/dotnet/project-system/pull/8216#pullrequestreview-992880814

When two async tasks can be run concurrently, we should run them that way to reduce overall time. The usual way of doing this is not very ergonomic (see below). Allowing a tuple to be awaited is quite nice. Examples follow.

### Serially

This is slower than it needs to be as the second task cannot start until the first completes.

```c#
var result1 = await DoSomethingAsync();
var result2 = await DoAnotherTaskAsync();
```

### Concurrent (BCL only)

This runs both tasks concurrently, but is harder to read than before.

```c#
var task1 = DoSomethingAsync();
var task2 = DoAnotherTaskAsync();
await Task.WhenAll(task1, task2);
var result1 = await task1;
var result2 = await task2;
```

### Concurrent (With tuples of tasks)

This commit enables awaiting a tuple of tasks. The tasks run run concurrently, and the result of those tasks is a tuple.

```c#
var (result1, result2) = await (DoSomethingAsync(), DoAnotherTaskAsync());
```

---

In future, we may extend this, for example to support more than two items in the tuple, or to support cases where not all tasks return a value.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8223)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -852,7 +852,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
 
                 globalProperties.TryGetValue(FastUpToDateCheckIgnoresKindsGlobalPropertyName, out string? ignoreKindsString);
 
-                (LogLevel requestedLogLevel, Guid projectGuid) = await(
+                (LogLevel requestedLogLevel, Guid projectGuid) = await (
                     _projectSystemOptions.GetFastUpToDateLoggingLevelAsync(token),
                     _guidService.GetProjectGuidAsync(token));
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -852,7 +852,7 @@ async Task<(bool, ImmutableArray<ProjectConfiguration>)> CheckAsync(UpToDateChec
 
                 globalProperties.TryGetValue(FastUpToDateCheckIgnoresKindsGlobalPropertyName, out string? ignoreKindsString);
 
-                (LogLevel requestedLogLevel, Guid projectGuid) = await WhenAll(
+                (LogLevel requestedLogLevel, Guid projectGuid) = await(
```
tmeschter, line 855: Nit: You may want to adjust the spacing between `await` and `(` to emphasize that this is awaiting a tuple, rather than calling a method named await.

tmeschter (approved):

ocallesp (approved):

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes:
```suggestion
                (LogLevel requestedLogLevel, Guid projectGuid) = await (
```

PR #8227, opened 2022-06-06
Add fabricbot.json

tmeschter:
FabricBot is moving away from a designer to "config-as-code". This change takes our existing configuration and copies it as-is to the blessed location in the repo; FabricBot will automatically pick it up from there. See https://1esdocs.azurewebsites.net/datainsights/merlinbot/extensions/Fabricbot_Overview.html for more information.

Many of the tasks are disabled, and we should clean them up at some future point.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8227)

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,1652 @@
+{
+  "version": "1.0",
+  "tasks": [
+    {
+      "taskType": "trigger",
+      "capabilityId": "AutoMerge",
+      "subCapability": "AutoMerge",
+      "version": "1.0",
+      "config": {
+        "taskName": "Automatically merge PR once CI passes",
+        "label": "auto-merge",
+        "minMinutesOpen": "12",
+        "mergeType": "merge",
+        "deleteBranches": true,
+        "requireAllStatuses": true,
+        "removeLabelOnPush": true,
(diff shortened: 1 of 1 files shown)


melytc (approved):

MiYanni (commented):
I'm a bit confused here. Did we already have a bot performing these tasks? I haven't seen anything like that. Good example is this one: "Close stale issues". Even after looking at the docs, I haven't heard of either of these bots before. Where did this config file come from?

drewnoakes (approved):

tmeschter:
@MiYanni This configuration was previously all internal, and configurable through https://portal.fabricbot.ms/bot/. One thing it did was automatically tag issues and PRs with the correct milestone depending on their target branch. If you've ever seen "msftbot" making updates to an issue/PR, that was probably controlled by this configuration. Many of the tasks in the .json are disabled, though, which is probably why they don't seem familiar.

I may have been the only one on the team with permission to update the bot. 🤷🏻‍♂️ However, they are moving to a configuration-as-code model, with the .json stored in the repo itself. In future the designer will be able to create a PR with updates.

PR #8232, opened 2022-06-07
Track project configurations during rebuilds

### LINKED ISSUE #8231: Overbuild if first ever build is a rebuild
From [AB#1550661](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1550661)

To reproduce:

1. Create a new project (or freshly clone a repository)
2. Rebuild
3. Build

Currently, the final build fails the project saying that the fast up-to-date check has never run. While that's true, we have actually completed a build, and so we should be able to use the start time of the rebuild in step 2 when deciding whether we need to build in step 3.

drewnoakes:
Fixes #8231
Fixes [AB#1550661](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1550661)

The fast up-to-date check tracks the set of configurations that are built, in order to remember the time of the last successful build in each configuration.

This needs to be tracked per-configuration because we support single-target builds, for MAUI for example.

Previously, we weren't tracking the last set of configurations in the case of rebuilds. This change corrects that.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8232)


AraHaan:
In that case should Rebuild nuke the previous restore and re-run it or assume it’s already up to date on restore and skip it?

drewnoakes:
Restore is a separate process and it is kicked off automatically when needed. Any build should be waiting for that restore to complete, again if needed. If you see a case where that doesn't happen, it's a bug.

MiYanni (approved):

PR #8234, opened 2022-06-08
Add Rule file for PackageVersion items

tmeschter:
Related to [AB#1544123](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1544123).

When CPS adds or modifies PackageVersion items, we would like it to persist the "Version" metadata as an attibute rather than as a child element. This can be controlled by providing a `Rule` for PackageVersion items that specifies `PersistenceStyle="Attribute"` for the relevant properties. If the `Rule` is placed in the "File" or "BrowseObject" context then the CPS types responsible for writing item metadata will pick it up and use the appropriate setting.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8234)


MiYanni (approved):
It looks good, but the `BrowseObjectRulesShouldMatchNone` test is failing from adding this file. So, that would need to be fixed.

davkean:
@tmeschter BrowseObject is for the properties window, this should be in the File context.

Issue #7238, opened 2021-05-18
Re-create VB References property page (10)


tmeschter:
We need to re-create the VB References property page in the new property page system.

For reference, this is the References property page as it currently appears for VB WinForms apps targeting .NET 5:

![image](https://user-images.githubusercontent.com/10506730/118734414-350e5600-b7f3-11eb-965e-b76181bef962.png)

I don't think this needs to be an exact copy of the existing page. For example, references can be managed via the Dependencies node, so I don't think we need to recreate that control here. However, I believe this is the only way to manage the "Reference Paths" and the "Imported namespaces".

_This work was identified as part of #6935._

drewnoakes:
Triage note: Need to validate which parts of this experience need to be brought forward.

cc: @KathleenDollard @cartermp

PR #8235, opened 2022-06-10
Implement MultiStringSelectorEditor to add imported namespaces feature to VB property pages

adamint:
Implements the new MultiStringSelectorEditor within the context of namespace imports. Allows the dynamic addition and removal of VB imports (C# usings will use the same MultiStringSelectorEditor), as well as user-defined imports.

To test, the associated CPS PR needs to also be checked out and built (https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/404223)

Page is shown below:
image


adamint:
I also need to add usage documentation for this editor

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+                        if (importProjectItem.IsImported)
+                        {
+                            _specialImports.Add(import);
+                        }
+                        
+                        project.RemoveItem(importProjectItem);
+                    });
+
+                }
+                catch (Exception)
```
drewnoakes, line 110: Catching all exceptions can mask serious problems. CPS has an extension for filtering out critical exceptions in cases like this. If we don't have access to it, we should probably port it and use it. Or change this to catch more specific exception type(s).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+        foreach (string importToAdd in importsToAdd)
+        {
+            if (importToAdd.Length > 0)
+            {
+                await _projectAccessor.OpenProjectXmlForWriteAsync(_configuredProject.UnconfiguredProject, project =>
+                {
+                    project.AddItem("Import", importToAdd);
+                });
+            }
+        }
```
drewnoakes, line 128: Getting a write lock is "expensive" so shouldn't be done in a loop. Something like this might help: 

```suggestion
        // Verify we have at least one valid import to add before acquiring a write lock.
        if (importsToAdd.Any(importToAdd => importToAdd.Length > 0))
        {
            await _projectAccessor.OpenProjectXmlForWriteAsync(_configuredProject.UnconfiguredProject, project =>
            {
                foreach (string importToAdd in importsToAdd)
                {
                    if (importToAdd.Length > 0)
                    {
                        project.AddItem("Import", importToAdd);
                    }
                }
            });
        }
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ReferencesPage.xaml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
+<Rule Name="References"
```
drewnoakes, line 3: Why is this rule empty?

drewnoakes (approved):
Really nice bit of work here! Left some minor comments to consider before merging, mostly around reducing allocations.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/DotNetImportsEnumProvider.cs
```diff
@@ -0,0 +1,111 @@
...
+        private readonly UnconfiguredProject _unconfiguredProject;
+        private readonly Workspace _workspace;
+        
+        public DotNetImportsEnumGenerator(UnconfiguredProject unconfiguredProject, Workspace workspace)
+        {
+            _unconfiguredProject = unconfiguredProject;
+            _workspace = workspace;
+        }
+
+        private async Task<List<string>> GetNamespacesAsync()
```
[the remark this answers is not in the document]
adamint: what do you mean by turning the final foreach into a generator, since bool isn't an iterator interface type?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ReferencesPage.xaml
adamint: I meant to add usings here :) but the code didn't get changed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
adamint: Good point. Only catching ArgumentException/InvalidOperationException now

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/DotNetImportsEnumProvider.cs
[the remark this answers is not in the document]
drewnoakes: Iterators cannot be async. But an async method can return the result of a local non-async function that is a generator. Then you don't need to allocate any list here.

melytc (approved):

melytc:
Related to #7238

tmeschter:
FYI @adamint , I updated the affected .xlf files and pushed my changes to your branch in an attempt to get the checks passing. You'll need to be sure to merge them into your local clone before you make any further changes.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ReferencesPage.VisualBasic.xaml
```diff
@@ -0,0 +1,42 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
+<Rule Name="References"
```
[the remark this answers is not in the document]
adamint: I don't think so, because that's how the build page is done

PR #8237, opened 2022-06-10
Only access App.xaml in WPF apps

tmeschter:
Fixes AB#1553442.

In #8220 we added VB property page support for WPF-related properties, like "Startup URI" and "Shutdown Mode". Ultimately these properties are read from the Application.xaml file--which generally only exists in WPF projects. The properties themselves are always present in the .xaml file, which means we end up trying to retrieve their values whenever the .xaml file is in use--we just hide the controls unless the project is a WPF application. Unfortunately the current interceptor for this property always tries to access the Application.xaml file, which will fail with an exception if the file isn't present.

To fix this, the interceptor first checks if this is a WPF application before trying to get the values from the Application.xaml. Specifically, we check if "UseWPF" is true, and that the output type of the project is "WinExe".

This also adds a number of unit tests to enforce that the Application.xaml is only accessed at appropriate times.

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
@@ -13,8 +13,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties;
     ExportInterceptingPropertyValueProviderFile.ProjectFile)]
 internal class WPFValueProvider : InterceptingPropertyValueProviderBase
 {
-    private const string StartupURIPropertyName = "StartupURI";
-    private const string ShutdownModePropertyName = "ShutdownMode_WPF";
+    internal const string StartupURIPropertyName = "StartupURI";
+    internal const string ShutdownModePropertyName = "ShutdownMode_WPF";
+    internal const string UseWPFPropertyName = "UseWPF";
+    internal const string OutputTypePropertyName = "OutputType";
+    internal const string WinExeOutputTypeValue = "WinExe";
 
     private readonly IApplicationXamlFileAccessor _applicationXamlFileAccessor;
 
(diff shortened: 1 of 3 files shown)


MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
```diff
@@ -26,35 +29,53 @@ public WPFValueProvider(IApplicationXamlFileAccessor applicationXamlFileAccessor
...
+            };
+        }
+
+        return string.Empty;
+    }
+
+    private static async Task<bool> IsWPFApplicationAsync(IProjectProperties defaultProperties)
+    {
+        string useWPFString = await defaultProperties.GetEvaluatedPropertyValueAsync(UseWPFPropertyName);
+        string outputTypeString = await defaultProperties.GetEvaluatedPropertyValueAsync(OutputTypePropertyName);
```
drewnoakes, line 75: Could use the new valuetuple awaiter to run these awaits concurrently.

drewnoakes (approved):

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
```diff
@@ -26,35 +29,53 @@ public WPFValueProvider(IApplicationXamlFileAccessor applicationXamlFileAccessor
...
+                _ => throw new InvalidOperationException($"The {nameof(WPFValueProvider)} does not support the '{propertyName}' property.")
+            };
+        }
+
+        return string.Empty;
+    }
+
+    private static async Task<bool> IsWPFApplicationAsync(IProjectProperties defaultProperties)
+    {
+        string useWPFString = await defaultProperties.GetEvaluatedPropertyValueAsync(UseWPFPropertyName);
```
davkean, line 74: This use be looking at the WPF capability to determine this - which can probably be provided to the class itself as an AppliesTo.

PR #8239, opened 2022-06-11
Avoid NuGet restore cycles

### LINKED ISSUE #8326: VS does not stop restoring
Creating this bug to track the work needed in Dotnet Project System.
This should be a p0

https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1539268

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1540772

## Summary

Dotnet Project System should detect cycles when doing nuget restores.

## Steps to Reproduce

1.  Create a C# ConsoleApp 
2.  Install the nuget package Microsoft.AppCenter version 4.4.0 

image

## Expected Behavior
VS should not hang or stuck in NuGet restore.

## Actual Behavior
VS gets into a cycle and stuck in NuGet restore.

## User Impact
 There is already a pr in review.
https://github.com/dotnet/project-system/pull/8239

ocallesp:
Fixes #8326

Avoid infinite NuGet restores.
Scenario:
Projects might have a package that sets `AssetTargetsFallback` value that can cause a certain `PackageReference` to be included. This package contains a props file that changed the `AssetTargetFallback` property value, and the `PackageReference` Condition was no longer true, causing it to be removed. Once removed, the `AssetTargetFallback` value from the `csproj` was used again, causing an infinite restore -> design time build -> restore loop.

This change:
Since DotNet ProjectSystem can detect changes in NuGet restores by calculating a hash, we can use this hash to detect changes in `AssetTargetFallBack` and check when this value gets repeated.

This pr makes use of the `NuGetRestoreCycleDetector` to keep the last N hashes, and once it detects a repeated value (cycle) Dotnet ProjectSystem can stop NuGet restores.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8239)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
@@ -1,0 +1,106 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
+{
+    // This algorithm is to detect the pattern A -> B -> A -> B -> A in the most recent N values.
+    // To keep track of the most N recent values we are using a queue of fixed size, where:
+    //   The most recent value is inserted at the front in O(1) time, and the oldest value is removed at the 
+    // back in O(1) time.
+    //   The counter will keep track of how many times values have appeared in the queue consecutively.
+    // i.e.
+    // A -> B -> C -> A -> B -> D -> X -> Y -> X -> Y -> X -> Y -> X
+    //                                        |---cycle detected----|
+    internal sealed class NuGetRestoreCycleDetector
+    {
+        // The fixed size of the queue
+        private readonly int _size = 5;
+
+        private readonly object _lock;
+        private readonly Queue<byte[]> _values;
+        private readonly Dictionary<byte[], int> _lookupTable;
+        private int _counter;
+
+        /// <summary>
+        ///     Fixed size of the numbers of values to store.
+        /// </summary>
+        /// <remarks>
+        ///     This represents how deep to search for hash cycle.
+        /// </remarks>
+        public int Size { get; private set; }
+
+        public NuGetRestoreCycleDetector()
+        {
+            _lock = new object();
+            _values = new Queue<byte[]>();
+            _lookupTable = new Dictionary<byte[], int>();
+            Size = _size;
+        }
+
+        /// <summary>
+        ///     Validate if hash1 cycle exist. 
+        ///     If no cycle exist, then the hashValue is stored.
+        /// </summary>
+        /// <param name="hashValue">This is hashValue that is used to verify if it exists in previous restores</param>
+        /// <returns>True if it contains hash1 cycle, otherwise false</returns>
+        public bool ComputeCycleDetection(byte[] hashValue)
+        {
+            lock (_lock)
+            {
+                if (QueueContainsValue(hashValue))
+                {
+                    _counter++;
+                    
+                    // Verify that a hashValue has repeated in almost all cases
+                    if (_counter > Size)
+                    {
+                        Clear();
+                        return true;
+                    }
+                }
+                else
+                {
+                    _counter = 0;
+                }
+
+                Add(hashValue);
+            }
+
+            return false;
+
+            bool QueueContainsValue(byte[] hashValue)
+            {
+                if (_lookupTable.TryGetValue(hashValue, out int hashCounter) && hashCounter > 0)
+                {
+                    return true;
+                }
+                return false;
+            }
+        }
+
+        private void Add(byte[] newHashValue)
+        {
+            if (_values.Count >= Size)
+            {
+                var oldestHashValue = _values.Dequeue();
+                _lookupTable[oldestHashValue]--;
+            }
+
+            _values.Enqueue(newHashValue);
+            if (_lookupTable.ContainsKey(newHashValue))
+            {
+                _lookupTable[newHashValue]++;
+            }
+            else
+            {
+                _lookupTable.Add(newHashValue, 1);
+            }
+        }
+
+        public void Clear()
+        {
+            _counter = 0;
+            _values.Clear();
+            _lookupTable.Clear();
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
@@ -2,9 +2,13 @@
 
 using System.Threading.Tasks.Dataflow;
 using Microsoft.Internal.Performance;
+using Microsoft.VisualStudio.Imaging;
 using Microsoft.VisualStudio.IO;
 using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.ProjectSystem.VS.UI.InfoBarService;
+using Microsoft.VisualStudio.Telemetry;
 using NuGet.SolutionRestoreManager;
+using Microsoft.Internal.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
 {
@@ -49,6 +53,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
         // |_________________________________________|    |_________________________________________|
         //
 
+        private readonly NuGetRestoreCycleDetector _cycleDetector = new();
+        private readonly IVsUIService<SVsFeatureFlags, IVsFeatureFlags> _featureFlagsService;
+        private readonly ITelemetryService _telemetryService;
+        private readonly IInfoBarService _infoBarService;
         private readonly UnconfiguredProject _project;
         private readonly IPackageRestoreUnconfiguredInputDataSource _dataSource;
         private readonly IProjectAsynchronousTasksService _projectAsynchronousTasksService;
@@ -61,6 +69,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
 
         [ImportingConstructor]
         public PackageRestoreDataSource(
+            IVsUIService<SVsFeatureFlags, IVsFeatureFlags> featureFlagsService,
+            ITelemetryService telemetryService,
+            IInfoBarService infoBarService,
             UnconfiguredProject project,
             IPackageRestoreUnconfiguredInputDataSource dataSource,
             [Import(ExportContractNames.Scopes.UnconfiguredProject)] IProjectAsynchronousTasksService projectAsynchronousTasksService,
@@ -71,6 +82,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
             PackageRestoreSharedJoinableTaskCollection sharedJoinableTaskCollection)
             : base(project, sharedJoinableTaskCollection, synchronousDisposal: true, registerDataSource: false)
         {
+            _featureFlagsService = featureFlagsService;
+            _telemetryService = telemetryService;
+            _infoBarService = infoBarService;
             _project = project;
             _dataSource = dataSource;
             _projectAsynchronousTasksService = projectAsynchronousTasksService;
@@ -129,11 +143,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
 
                 if (_latestHash != null && hash.AsSpan().SequenceEqual(_latestHash))
                 {
+                    _cycleDetector.Clear();
                     SaveNominatedConfiguredVersions(value.ConfiguredInputs);
                     return true;
                 }
 
                 _latestHash = hash;
+
+                if (await CycleDetectedAsync(hash, _projectAsynchronousTasksService.UnloadCancellationToken))
+                {
+                    return false;
+                }
 
                 _restoreStarted = true;
                 JoinableTask<bool> joinableTask = JoinableFactory.RunAsync(() =>
@@ -161,6 +181,42 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
             }
 
             return success;
+        }
+
+        private async Task<bool> CycleDetectedAsync(byte[] hash, CancellationToken cancellationToken)
+        {
+            bool enabled = await IsNuGetRestoreCycleDetectionEnabledAsync(cancellationToken);
+
+            if (!enabled)
+            {
+                return false;
+            }
+
+            bool cycleDetected = _cycleDetector.ComputeCycleDetection(hash);
+
+            if (!cycleDetected)
+            {
+                return false;
+            }
+
+            _telemetryService.PostEvent(TelemetryEventName.NuGetRestoreCycleDetected);
+
+            await _infoBarService.ShowInfoBarAsync(
+                VSResources.InfoBarMessageNuGetCycleDetected,
+                KnownMonikers.StatusError,
+                cancellationToken
+            );
+
+            return true;
+        }
+
+        private async Task<bool> IsNuGetRestoreCycleDetectionEnabledAsync(CancellationToken cancellationToken)
+        {
+            await _project.Services.ThreadingPolicy.SwitchToUIThread(cancellationToken);
+
+            IVsFeatureFlags featureFlagsService = _featureFlagsService.Value;
+
+            return featureFlagsService.IsFeatureEnabled(FeatureFlags.EnableNuGetRestoreCycleDetection, defaultValue: false);
         }
 
         private async Task<bool> NominateForRestoreAsync(ProjectRestoreInfo restoreInfo, CancellationToken cancellationToken)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.InfoBarEntry.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.InfoBarEntry.cs
@@ -1,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.UI.InfoBarService;
+
+internal partial class VsInfoBarService
+{
+    private class InfoBarEntry : IVsInfoBarUIEvents
+    {
+        private readonly IVsInfoBarUIElement _element;
+        private readonly InfoBarUI[] _items;
+        private readonly Action<InfoBarEntry> _onClose;
+        private readonly uint _cookie;
+
+        public InfoBarEntry(string message, IVsInfoBarUIElement element, InfoBarUI[] items, Action<InfoBarEntry> onClose)
+        {
+            Assumes.NotNull(message);
+            Assumes.NotNull(element);
+            Assumes.NotNull(items);
+            Assumes.NotNull(onClose);
+
+            Message = message;
+            _element = element;
+            _items = items;
+            _onClose = onClose;
+            element.Advise(this, out _cookie);
+        }
+
+        public string Message { get; }
+
+        public void Close()
+        {
+            _element.Close();
+        }
+
+        public void OnActionItemClicked(IVsInfoBarUIElement element, IVsInfoBarActionItem actionItem)
+        {
+            Requires.NotNull(element, nameof(element));
+            Requires.NotNull(actionItem, nameof(actionItem));
+
+            InfoBarUI item = _items.First(i => i.Title == actionItem.Text);
+            item.Action();
+
+            if (item.CloseAfterAction)
+            {
+                Close();
+            }
+        }
+
+        public void OnClosed(IVsInfoBarUIElement element)
+        {
+            Requires.NotNull(element, nameof(element));
+
+            _element.Unadvise(_cookie);
+            _onClose(this);
+        }
+    }
+}
+
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
@@ -1,0 +1,131 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Imaging.Interop;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.UI.InfoBarService;
+
+/// <summary>
+/// Implementation of <see cref="IInfoBarService"/> that pushes messages to the info bar attached to Visual Studio's main window.
+/// </summary>
+[Export(typeof(IInfoBarService))]
+internal partial class VsInfoBarService : IInfoBarService
+{
+    private readonly IProjectThreadingService _threadingService;
+    private readonly IVsShellServices _vsShellServices;
+    private readonly IVsUIService<SVsShell, IVsShell> _vsShell;
+    private readonly IVsUIService<SVsInfoBarUIFactory, IVsInfoBarUIFactory> _vsInfoBarFactory;
+    private readonly List<InfoBarEntry> _entries = new();
+
+    [ImportingConstructor]
+    public VsInfoBarService(
+        IProjectThreadingService threadingService,
+        IVsShellServices vsShellServices,
+        IVsUIService<SVsShell, IVsShell> vsShell,
+        IVsUIService<SVsInfoBarUIFactory, IVsInfoBarUIFactory> vsInfoBarFactory)
+    {
+        _threadingService = threadingService;
+        _vsShellServices = vsShellServices;
+        _vsShell = vsShell;
+        _vsInfoBarFactory = vsInfoBarFactory;
+    }
+
+    public async Task ShowInfoBarAsync(string message, ImageMoniker image, CancellationToken cancellationToken, params InfoBarUI[] items)
+    {
+        Requires.NotNullOrEmpty(message, nameof(message));
+        Requires.NotNull(items, nameof(items));
+
+        if (!_vsShellServices.IsInServerMode)
+        {
+            await _threadingService.SwitchToUIThread(cancellationToken);
+
+            IVsInfoBarHost? host = FindMainWindowInfoBarHost();
+            if (host == null)
+            {
+                return;
+            }
+
+            // We want to avoid posting the same message over and over again, so we remove any existing info bar
+            // with the same message, and add the new one which will cause it to float to the bottom of the host.
+            RemoveInfoBar(message);
+            AddInfoBar(host, message, image, items);
+        }
+    }
+
+    private IVsInfoBarHost? FindMainWindowInfoBarHost()
+    {
+        if (_vsShell.Value is null)
+        {
+            return null;
+        }
+
+        if (ErrorHandler.Failed(_vsShell.Value.GetProperty((int)__VSSPROPID7.VSSPROPID_MainWindowInfoBarHost, out object mainWindowInfoBarHost)))
+        {
+            return null;
+        }
+
+        return mainWindowInfoBarHost as IVsInfoBarHost;
+    }
+
+    private IVsInfoBarUIElement? CreateInfoBarUIElement(string message, ImageMoniker image, InfoBarUI[] items)
+    {
+        if (_vsInfoBarFactory.Value is null)
+        {
+            return null;
+        }
+
+        var actionItems = new List<IVsInfoBarActionItem>(items.Length);
+
+        foreach (InfoBarUI item in items)
+        {
+            switch (item.Kind)
+            {
+                default:
+                case InfoBarUIKind.Button:
+                    Assumes.True(item.Kind == InfoBarUIKind.Button);
+                    actionItems.Add(new InfoBarButton(item.Title));
+                    break;
+                case InfoBarUIKind.Hyperlink:
+                    actionItems.Add(new InfoBarHyperlink(item.Title));
+                    break;
+            }
+        }
+
+        var infoBarModel = new InfoBarModel(
+            message,
+            actionItems.ToArray(),
+            image,
+            isCloseButtonVisible: true);
+
+        return _vsInfoBarFactory.Value.CreateInfoBar(infoBarModel);
+    }
+
+    private void AddInfoBar(IVsInfoBarHost host, string message, ImageMoniker image, InfoBarUI[] items)
+    {
+        IVsInfoBarUIElement? element = CreateInfoBarUIElement(message, image, items);
+        if (element != null)
+        {
+            var entry = new InfoBarEntry(message, element, items, OnClosed);
+            _entries.Add(entry);
+            host.AddInfoBar(element);
+        }
+    }
+
+    private void RemoveInfoBar(string message)
+    {
+        // Assumption is that message is "good" enough to uniquely identify problems
+        InfoBarEntry entry = _entries.Find(e => e.Message == message);
+        entry?.Close();
+    }
+
+    private void OnClosed(InfoBarEntry entry)
+    {
+        Requires.NotNull(entry, nameof(entry));
+
+        _threadingService.VerifyOnUIThread();
+
+        _entries.Remove(entry);
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/VSShellServices.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/VSShellServices.cs
@@ -1,0 +1,64 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using Microsoft.VisualStudio.ProjectSystem;
+using Microsoft.VisualStudio.ProjectSystem.VS;
+using Microsoft.VisualStudio.Shell.Interop;
+
+#nullable disable
+
+namespace Microsoft.VisualStudio.Shell;
+
+/// <summary>
+/// See IVSShellServices.
+/// </summary>
+[Export(typeof(IVsShellServices))]
+[AppliesTo(ProjectCapabilities.AlwaysApplicable)]
+internal class VSShellServices : OnceInitializedOnceDisposed, IVsShellServices
+{
+    /// <summary>
+    /// The service provider providing access to the VS services.
+    /// </summary>
+    [Import]
+    private IVsUIService<SVsShell, IVsShell> ServiceProvider { get; set; }
+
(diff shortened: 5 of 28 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            lock (_lock)
+            {
+                foreach (var b in _values)
+                {
+                    if (AreEqual(b, hashValue))
+                    {
+                        return true;
+                    }
+                }
```
drewnoakes, line 41: This loop detects whether the current hash has been seen within the last N hashes.

You could trigger this by adding a package reference, then removing it. Once removed, the project is back in the prior state and restore should complete.

With this change, I believe we would not restore after removing the recently added package reference.

As an alternative, consider something like:

- store the last N hashes
- each time a new hash is produced
  - append it to the list
  - discard any hashes added longer than some threshold ago
  - walk back through the list, looking for cycles of repeating patterns of length 2, 3, 4, ...
  - if a repeating pattern is found, there is a cycle

I'm not 100% sure about that approach, but I think it would be more robust than the approach here.

nkolev92: > I am wondering if NuGet can easily detect a bad package

What would be a bad package in this case?

If you are referring to a package with props/targets that change the inputs, the MSBuild side is hidden from NuGet, so NuGet has no means of figuring it out, so this logic is technically handling that.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -134,6 +135,10 @@ private async Task<bool> RestoreCoreAsync(PackageRestoreUnconfiguredInput value)
                 }
 
                 _latestHash = hash;
+                if (_cycleDetector.ComputeCycleDetection(hash))
+                {
+                    return true;
```
drewnoakes, line 140: We should surface the fact that a loop is suspected to the user, probably via a gold bar.

Something like "An infinite NuGet restore loop is suspected in project _MyProject_. This can happen when a package changes the project's imports, which requires a further restore. The most recent restore was skipped in order to break the cycle."

This serves two purposes:

1. Highlights the issue to the user in case they wish to fix it. It would be great if we could identify the package causing the issue to further help the user.
2. Shows the user that a restore was cancelled, which may be important if they do not actually have a cycle. Perhaps they were just switching between two branches several times, for example.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -49,6 +49,7 @@ internal partial class PackageRestoreDataSource : ChainedProjectValueDataSourceB
         // |_________________________________________|    |_________________________________________|
         //
 
+        private readonly NuGetRestoreCycleDetector _cycleDetector = new(5);
```
drewnoakes, line 52: Do we have a sense of how long it would take to restore five times? If it's not very long, I'd vote to increase this threshold.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
```diff
@@ -0,0 +1,90 @@
...
+                    if (counter > Size)
+                    {
+                        counter = 0;
+                        ClearQueue();
+                        return true;
+                    }
+                }
+                else
+                {
+                    Add(hashValue);
```
drewnoakes, line 50: If the queue does not contain this value, I think the counter should be reset to zero.

Also the length of the queue should probably be capped. Otherwise the probability of finding the hash in the queue increases over time.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
drewnoakes: > I modified the logic a little to detect the cycle only when counter == Size.

I'm concerned that this is only matching values and not looking for patterns.

For example (where each letter represents a hash):

- A B A B A B A B A B A B
- A B C A B C A B C A B C
- A B A C A D E A F A G A

The first two sequences look more like infinite cycles to me than the third one, yet the third would trip the check here.

nkolev92:
Any updates on this change @ocallesp?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
nkolev92: You can maybe use https://docs.microsoft.com/en-us/nuget/visual-studio-extensibility/nuget-api-in-visual-studio#ivsnugetprojectupdateevents-interface to detect the timing of restore start/end. 

We can also add other events if they'd make life easier for you to do detect these loops.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
ocallesp: @nkolev92 do you think these patterns are possible to appear ?

nkolev92: In theory, you can come up with packages/authoring that has a pattern of any length. 

In my experience, I've only seen  A B A B patterns.  

The most common occurrences of continuous restores is the dynamically generate version one. So maybe we try to detect situations where say a bunch of consecutive design time builds lead to a restore nomination. If we want to special case that, we can probably use the NuGet start/stop restore markers for help.

ocallesp: @nkolev92  if we are in a loop, does it mean that we already completed NuGet restore ?

nkolev92: yep. 

The loop is effectively, design time build which nominates, restore, design time build which nominates, restore.

ocallesp: @nkolev92 at the project system level we don't know what package is causing the loop. We can show a yellow bar to notify the user that one of the nuget packages is invalid. Is there a way to identify the package from the NuGet side ? so we can get the name of the package from NuGet and use it in the yellow bar ?

nkolev92: > @nkolev92 at the project system level we don't know what package is causing the loop. We can show a yellow bar to notify the user that one of the nuget packages is invalid and is causing a nuget restore cycle.

Yep, that's in line with what I'd expect based on how MSBuild works :) 

>  Is there a way to identify the package from the NuGet side ? so we can get the name of the package from NuGet and use it in the yellow bar ?

NuGet doesn't really anything MSBuild in VS, so there's pretty much no way of figuring it out.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
drewnoakes: > Shows the user that a restore was cancelled, which may be important if they do not actually have a cycle. Perhaps they were just switching between two branches several times, for example.

Perhaps we add a button to the infobar to allow forcing a restore. Similar to how CPS's project fault infobar has a button to reload the project.

MiYanni:
@ocalles You're getting build failures with these changes.
```
(CoreCompile target) -> 
         D:\a\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\PackageRestore\PackageRestoreDataSource.cs(7,44): error CS0234: The type or namespace name 'Implementation' does not exist in the namespace 'Microsoft.VisualStudio.ProjectSystem' (are you missing an assembly reference?)
         D:\a\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\PackageRestore\PackageRestoreDataSource.cs(55,26): error CS0246: The type or namespace name 'IInfoBarService' could not be found (are you missing a using directive or an assembly reference?)
         D:\a\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\PackageRestore\PackageRestoreDataSource.cs(68,13): error CS0246: The type or namespace name 'IInfoBarService' could not be found (are you missing a using directive or an assembly reference?)
```

ocallesp:
> @ocalles You're getting build failures with these changes.
> 
> ```
> (CoreCompile target) -> 
>          D:\a\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\PackageRestore\PackageRestoreDataSource.cs(7,44): error CS0234: The type or namespace name 'Implementation' does not exist in the namespace 'Microsoft.VisualStudio.ProjectSystem' (are you missing an assembly reference?)
>          D:\a\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\PackageRestore\PackageRestoreDataSource.cs(55,26): error CS0246: The type or namespace name 'IInfoBarService' could not be found (are you missing a using directive or an assembly reference?)
>          D:\a\_work\1\s\src\Microsoft.VisualStudio.ProjectSystem.Managed.VS\ProjectSystem\VS\PackageRestore\PackageRestoreDataSource.cs(68,13): error CS0246: The type or namespace name 'IInfoBarService' could not be found (are you missing a using directive or an assembly reference?)
> ```

Fixed.
I added the implementation of the InfoBar service as part of this pr.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
ocallesp: Done.
An InfoBar will be displayed in the main window with the message "NuGet restore detected a cycle. NuGet Error NU1108"

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
ocallesp: InfoBar added.

ocallesp:
Tests:
Open a solution having an invalid NuGet package. Cycle was detected.
Open roslyn and orchardcore. No cycles detected.

ocallesp:
> Any updates on this change @ocallesp?

Today I add a new commit with the implementation of InfoBar. This pr is complete. Waiting for reviewers and approval

kvenkatrajan:
@ocallesp - not required for the initial merge but anyways we can add telemetry around this for when we detect a cyclic restore? Will help determine how often our users run into this scenario, we detect it & display the infobar

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
ocallesp: In the next commit the threshold will be increased to 10

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryEventName.cs
```diff
@@ -71,5 +71,10 @@ internal static class TelemetryEventName
         ///     Indicates that the user was notified of the suspected incremental build failure.
         /// </summary>
         public const string IncrementalBuildValidationFailureDisplayed = Prefix + "/incrementalbuild/validationfailure/displayed";
+
+        /// <summary>
+        ///     Indicates that the NuGet restore detected a cycle.
+        /// </summary>
+        public const string NuGetRestoreCycleDetected = Prefix + "/nugetrestore/cycledetected";
```
jacdavis, line 78: >cycledetected [](http://example.com/codeflow?start=80&length=13)

Don't forget to get this blessed once it starts coming through.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/VSShellServices.cs
```diff
@@ -0,0 +1,67 @@
...
+    public VSShellServices()
+        : base(synchronousDisposal: true) // disposal requires UI thread unadvise, so async defeats that.
+    {
+    }
+
+    protected override void Initialize()
+    {
+        vsShell = ServiceProvider.Value;
+
+        if (ErrorHandler.Succeeded(vsShell.GetProperty((int)__VSSPROPID.VSSPROPID_IsInCommandLineMode, out object result)) && (bool)result)
```
jacdavis, line 44: >if (ErrorHandler.Succeeded(vsShell.GetProperty((int)__VSSPROPID.VSSPROPID_IsInCommandLineMode, out object result)) && (bool)result) [](http://example.com/codeflow?start=8&length=131)

Are we sure this always executes on the ui thread? IIRC, IVsShell  requires the ui thread and doesn't have a proxy.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/VSShellServices.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            IsInServerMode = CheckIsInServerMode();
+        }
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private bool CheckIsInServerMode()
+    {
+        if (ErrorHandler.Succeeded(vsShell.GetProperty((int)__VSSPROPID11.VSSPROPID_ShellMode, out object value))
```
jacdavis, line 53: >GetProperty [](http://example.com/codeflow?start=43&length=11)

Same as above. What thread does this execute on? I would feel better about a JTF call to ensure we're on the ui thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.InfoBarEntry.cs
```diff
@@ -0,0 +1,61 @@
...
+        private readonly uint cookie;
+
+        public InfoBarEntry(string message, IVsInfoBarUIElement element, InfoBarUI[] items, Action<InfoBarEntry> onClose)
+        {
+            Assumes.NotNull(message);
+            Assumes.NotNull(element);
+            Assumes.NotNull(items);
+            Assumes.NotNull(onClose);
+
+            Message = message;
```
jacdavis, line 24: >Message [](http://example.com/codeflow?start=12&length=7)

normally, C# static analysis will throw a warning if fields are accessed without "this.". ISt his disabled in this codebase?

jacdavis: note that I'm having a hrad time understanding this code without the this. prefixes. For instance, I assumed cookie was a local in this function.

jacdavis: It looks like there is existing code that doesn't put in the "this." for property invocation, but fields all use the _ prefix.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
```diff
@@ -0,0 +1,132 @@
...
+        _vsShell = vsShell;
+        _vsInfoBarFactory = vsInfoBarFactory;
+    }
+
+    public async Task ShowInfoBarAsync(string message, ImageMoniker image, CancellationToken cancellationToken, params InfoBarUI[] items)
+    {
+        Requires.NotNullOrEmpty(message, nameof(message));
+        Requires.NotNull(items, nameof(items));
+
+        if (!_vsShellServices.IsInServerMode)
```
jacdavis, line 41: >if (!_vsShellServices.IsInServerMode) [](http://example.com/codeflow?start=8&length=37)

What is the right behavior if this is in server mode? Right now, it silently succeeds.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            var entry = new InfoBarEntry(message, element, items, OnClosed);
+            entries.Add(entry);
+            host.AddInfoBar(element);
+        }
+    }
+
+    private void RemoveInfoBar(string message)
+    {
+        // Assumption is that message is "good" enough to uniquely identify problems
```
jacdavis, line 119: > // Assumption is that message is "good" enough to uniquely identify problems [](http://example.com/codeflow?start=7&length=77)

Are we sure this is the case? Seems like a bug waiting to happen

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -49,6 +52,9 @@ internal partial class PackageRestoreDataSource : ChainedProjectValueDataSourceB
         // |_________________________________________|    |_________________________________________|
         //
 
+        private readonly NuGetRestoreCycleDetector _cycleDetector = new(10);
```
jacdavis, line 55: >10 [](http://example.com/codeflow?start=72&length=2)

What is this magic number? Why 10 here?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <returns>True if it contains hash1 cycle, otherwise false</returns>
+        public bool ComputeCycleDetection(byte[] hashValue)
+        {
+            lock (_lock)
+            {
+                if (QueueContainsValue(hashValue))
+                {
+                    counter++;
+                    
+                    // Verify that a hashValue has repeted in almost all cases
```
jacdavis, line 40: >repeted [](http://example.com/codeflow?start=51&length=7)

spelling

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
```diff
@@ -0,0 +1,90 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
+{
+    internal sealed class NuGetRestoreCycleDetector
+    {
+        private readonly object _lock;
+        private readonly Queue<byte[]> _values;
```
jacdavis, line 8: >Queue [](http://example.com/codeflow?start=25&length=5)

Why are we using a queue here and not some data structure with O(1) lookup? I expected this to be a hash table of hashes. A queue doesn't make a lot of sense in my mind. Am I missing some info?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
jacdavis: Are you sure IsInServerMode works off the ui thread? This is invoked on a arbitrary thread in cycle detector.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.InfoBarEntry.cs
tmeschter: Fields should use the _ prefix.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryEventName.cs
kvenkatrajan: think it would be helpful to add additional properties in addition to cycledetected. So have the event name as : 
/nugetrestore/cycledetection and properties like vs.projectsystem.managed.nugetrestore.cycledetection.cycledetected, vs.projectsystem.managed.nugetrestore.cycledetection.durationmillis, vs.projectsystem.managed.nugetrestore.cycledetection.successfulrestorecount,
vs.projectsystem.managed.nugetrestore.cycledetection.projectid or vs.projectsystem.managed.nugetrestore.cycledetection.solutionid perhaps. Any additional items you think may be useful to help troubleshoot when we detect a cyclical restore.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
kvenkatrajan: IsInServerMode -> is this to account for Nexus based scenarios? Do we need to explicitly call out a NotSupportedException when IsInServerMode = true?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.InfoBarEntry.cs
ocallesp: prefix added. Fixed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
ocallesp: We just want to search if there is an existing InfoBar with the same message (not taking into account buttons/links). We can change this implementation if that is really needed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
ocallesp: 5 is enough to detect this scenario
X -> Y -> X -> Y -> X

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
ocallesp: I added the hash _lookupTable to improve the search to O(1)

src/Microsoft.VisualStudio.ProjectSystem.Managed/Telemetry/TelemetryEventName.cs
jacdavis: I *believe* project id and solutionid are sent automatically. Duration and succesffulrestorecount would be nice. I think succesffulrestorecount  will be unreliable as I believe we must stop as soon as a cycle is detected.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
jacdavis: I'm still not convinced a queue is the correct data structure here. Can you please call me to discuss as requested?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
jacdavis: Please don't use a magic number. Put it in a named constant that explains the reasoning behind it, or at a minimum, add a comment about why this number was chosen.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -135,6 +145,17 @@ private async Task<bool> RestoreCoreAsync(PackageRestoreUnconfiguredInput value)
 
                 _latestHash = hash;
 
+                if (_cycleDetector.ComputeCycleDetection(hash))
```
nkolev92, line 148: Even if we hit this here, still nominate on line 160 right?

That should probably not happen correct?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -139,6 +139,7 @@ private async Task<bool> RestoreCoreAsync(PackageRestoreUnconfiguredInput value)
 
                 if (_latestHash != null && hash.AsSpan().SequenceEqual(_latestHash))
                 {
+                    _cycleDetector.Clear();
```
ocallesp, line 142: @nkolev92 with this change we avoid showing the infobar in the scenario where
a user adds a nuget package, then delete the nuget package, then add the same package, then delete the same package.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
ocallesp: I think we should return succeed = false.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
```diff
@@ -2,9 +2,12 @@
 
 using System.Threading.Tasks.Dataflow;
 using Microsoft.Internal.Performance;
+using Microsoft.VisualStudio.Imaging;
 using Microsoft.VisualStudio.IO;
 using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.ProjectSystem.VS.UI.InfoBarService;
 using NuGet.SolutionRestoreManager;
+using Microsoft.VisualStudio.Telemetry;
```
tmeschter, line 10: Nit: sort the using statement.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
tmeschter: This was originally to account for Nexus scenarios, yes, and as we don't really have a server SKU under active development it probably doesn't matter much. I think it's reasonable to throw an exception if we are in server mode, and also agree with @jacdavis that we need to verify `IsInServerMode` is ok to call from an arbitrary thread, and/or transition to the UI thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
```diff
@@ -0,0 +1,132 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Imaging.Interop;
+using Microsoft.VisualStudio.ProjectSystem.VS.UI.InfoBarService;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore.InfoBarService;
```
tmeschter, line 9: The InfoBarService isn't really specific to package restore scenarios; I suggest moving this to a different namespace and folder.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/VSShellServices.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        return false;
+    }
+
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+        }
+    }
```
tmeschter, line 66: Is this method dead code?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -295,4 +295,7 @@ In order to debug this project, add an executable project to this solution which
     <value>The .xaml file was in an unexpected format, near line {0} column {1}.</value>
     <comment>{0} is the line number of the problematic XAML, and {1} is the column.</comment>
   </data>
+  <data name="InfoBarMessageNuGetCycleDetected" xml:space="preserve">
+    <value>NuGet restore detected a cycle. NuGet Error NU1108</value>
```
tmeschter, line 299: Can we provide the user more details, or explain the impact of a cycle? We should help the user understand why a cycle in NuGet Restore is bad.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
ocallesp: fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
ocallesp: fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
ocallesp: Umm.... I thought this was updated by the NamespaceUpdate.

I'll fix this in the next commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/NuGetRestoreCycleDetector.cs
ocallesp: done.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Shell/VSShellServices.cs
ocallesp: deleted in last commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/UI/InfoBarService/VsInfoBarService.cs
ocallesp: fixed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreDataSource.cs
ocallesp: sorted.
Fixed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
ocallesp: I changed this to something more verbose.

_A circular dependency A -> B -> A was detected during NuGet Restore which caused NuGet to stop restoring.
Because the NuGet restore didn't finish correctly, the project might be in a bad state.

Projects might have a package that sets AssetTargetFallback, not supported for this scenario, which caused certain PackageReference to be included.

NuGet Error NU1108_

ocallesp:
Some of the pending code review suggestions will be fixed as part of https://github.com/dotnet/project-system/issues/8389

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -295,4 +295,12 @@ In order to debug this project, add an executable project to this solution which
     <value>The .xaml file was in an unexpected format, near line {0} column {1}.</value>
     <comment>{0} is the line number of the problematic XAML, and {1} is the column.</comment>
   </data>
+  <data name="InfoBarMessageNuGetCycleDetected" xml:space="preserve">
+    <value>A circular dependency A -&gt; B -&gt; A was detected during NuGet Restore which caused NuGet to stop restoring.
+Because the NuGet restore didn't finish correctly, the project might be in a bad state.
+
+This might be caused because one or more projects have a package that sets AssetTargetFallback, not supported for this scenario, which caused certain PackageReference to be included.
+
+NuGet Error NU1108</value>
```
nkolev92, line 304: I'd recommend removing this. 

NU1108 talks about a restore cycle within a restore graph. This merely exposes a limitation in what VS can do wrt to DT builds. 

The NU1108 text doesn't really talk about any of these issues anyways.

drewnoakes: @ocallesp were you able to apply @nkolev92's feedback here?

drewnoakes: This was addressed in ff84ebd58605f6878decb2619c841f4a73e96e7c.

PR #8241, opened 2022-06-14
Move apply resources out of ApplicationPropPage constructor

ocallesp:
Handle [AB#1257559](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1257559) to reduce UI delays.

This bug was reported by PerfWatson UIDelay using the tool _VS Delay View_.
The report is pointing `Microsoft.VisualStudio.Editors.PropertyPages.ApplicationPropPage.InitializeComponent()` called by the constructor.

`InitializeComponent()` is heavy and we want to move code not needed in the constructor to another place.
The code that is moved out is  `ApplyResources()` and it is moved to `OnCreateControl()`.

Based on  [msft doc](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.createcontrol?view=windowsdesktop-6.0#system-windows-forms-control-createcontrol) it seems that  `OnCreateControl()` is a good place since `CreateControl()` is used when we need add handles not created in the constructor.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8241)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.Designer.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPage.Designer.vb
@@ -35,7 +35,6 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
         Friend WithEvents iconTableLayoutPanel As System.Windows.Forms.TableLayoutPanel
 
         <System.Diagnostics.DebuggerStepThrough()> Private Sub InitializeComponent()
-            Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(ApplicationPropPage))
             Me.TopHalfLayoutPanel = New System.Windows.Forms.TableLayoutPanel()
             Me.AssemblyNameLabel = New System.Windows.Forms.Label()
             Me.AssemblyName = New System.Windows.Forms.TextBox()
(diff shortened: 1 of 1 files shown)


adamint (approved):

ocallesp:
> So, the only file in the PR is a generated designer file. Meaning, it is likely there is a file missing from the commit. I'm assuming you changed `ApplicationPropPage.vb` to cause these changes on build, correct? Please commit the changed source file.

Yes, this is an auto-generated file from `ApplicationPropPage.vb` and even if we open this file in VS it will overwrite the changes in `ApplicationPropPage.Designer.vb`.

My assumption is that this is the old property pages, and we don't usually make changes to it.
Unless there is a p0/p1 in the UI like in this case.

I will close this pr, and if we get more hits, I will retake this change even though it is in the auto-generated file.

MiYanni:
@ocalles Every run of `build.cmd` (or rebuilding the solution itself) rebuilds the `Microsoft.VisualStudio.Editors` project. So, you cannot make manual changes in a generated file as they are designed to be overridden on-build. The only way to make this change is by changing `ApplicationPropPage.vb`. Technically, we don't need to commit `.Designer.vb` files into the repo, but we do because we want to see what changes occur.

drewnoakes:
@MiYanni I don't believe that's true for WinForms controls. The WinForms designer file is the source of truth for the layout of the UI. It would only be regenerated if someone opened the file in the designer in VS, made a change then saved it.

MiYanni:
@drewnoakes Oh, interesting. Yeah, I just tested it myself right now. I cannot delete a `.Designer.vb` file and have the build succeed. That just shows how long it has been since I've done any major WinForms dev. 😅

PR #8252, opened 2022-06-17
Intercept WinForms properties

melytc:
WinForms project have specific Applcation Framework properties that need to be read and saved in the corresponding Application.myapp file. For now, we will intercept those properties via the `ApplicationFrameworkPropertiesValueProvider`, but ideally we would like to implement a `IProjectPropertiesProvider` to access the myapp file.

Addresses #7236.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8252)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
@@ -30,226 +30,16 @@ internal class MyAppFileAccessor : IMyAppFileAccessor
         _fileName = fileName;
     }
 
-    public async Task<bool?> GetMySubMainAsync()
+    private async Task<MyAppDocument?> TryGetMyAppFileAsync()
     {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(false);
-        string? value = myAppDocument?.GetProperty(_filePath, MySubMainProperty);
+        string filePath = Path.GetFullPath(_fileName);
 
-        if (value is null)
-            return null;
-        else
-            return bool.Parse(value);
-    }
-    
-    public async Task SetMySubMainAsync(string mySubMain)
-    {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(true);
-
-        if (myAppDocument is not null)
-        {
-            myAppDocument.SetProperty(MySubMainProperty, mySubMain);
-        }
-    }
-
-    public async Task<string?> GetMainFormAsync()
-    {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(false);
-        string? value = myAppDocument?.GetProperty(_filePath, MainFormProperty);
-
-        if (value is null)
-            return null;
-        else
-            return value;
-    }
-    
-    public async Task SetMainFormAsync(string mainForm)
-    {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(true);
-
-        if (myAppDocument is not null)
-        {
-            myAppDocument.SetProperty(MySubMainProperty, mainForm);
-        }
-    }
-
-    public async Task<bool?> GetSingleInstanceAsync()
-    {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(false);
-        string? value = myAppDocument?.GetProperty(_filePath, SingleInstanceProperty);
-
-        if (value is null)
-            return null;
-        else
-            return bool.Parse(value);
-    }
-
-    public async Task SetSingleInstanceAsync(bool singleInstance)
-    {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(true);
-
-        if (myAppDocument is not null)
-        {
-            myAppDocument.SetProperty(SingleInstanceProperty, singleInstance.ToString());
-        }
-    }
-    
-    public async Task<int?> GetShutdownModeAsync()
-    {
-        MyAppDocument? myAppDocument = await TryGetMyAppFileAsync(false);
-        string? value = myAppDocument?.GetProperty(_filePath, ShutdownModeProperty);
-
-        if (value is null)
-            return null;
-        else
-            return int.Parse(value);
-    }
-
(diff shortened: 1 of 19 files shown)


MiYanni (commented):
I talked with Melissa about using an XmlSerializer instead and simply having an object model that represents the MyApp file. It'll do the job simpler and easier. It should also make it easier if we want to switch to the workflow that @tmeschter mentioned previously of using an in-memory representation of a document editor to read/write the file (instead of direct serialization to disk). It isn't necessary for the first implementation, though.

melytc:
My initial thought about the two approaches is that it depends on what we value more: I see how saving/modifying properties as needed (per property) can save us memory but be more time consuming; in the other hand, having a model with the expected structure (and maybe save a dictionary with everything else that could exist there) with an approach of saving everything in this model will cost more memory but be time efficient.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkPropertiesValueProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            //ApplicationFrameworkProperty => (await _myAppXamlFileAccessor.GetMySubMainAsync()).ToString() ?? string.Empty,
+            EnableVisualStylesProperty => (await _myAppXamlFileAccessor.GetEnableVisualStylesAsync()).ToString() ?? string.Empty,
+            SingleInstanceProperty => (await _myAppXamlFileAccessor.GetSingleInstanceAsync()).ToString() ?? string.Empty,
+            SaveMySettingsOnExitProperty => (await _myAppXamlFileAccessor.GetSaveMySettingsOnExitAsync()).ToString() ?? string.Empty,
+            HighDpiModeProperty => (await _myAppXamlFileAccessor.GetHighDpiModeAsync()).ToString() ?? string.Empty,
+            AuthenticationModeProperty => (await _myAppXamlFileAccessor.GetAuthenticationModeAsync()).ToString() ?? string.Empty,
+            ShutdownModeProperty => (await _myAppXamlFileAccessor.GetShutdownModeAsync()).ToString() ?? string.Empty,
+            SplashScreenProperty => await _myAppXamlFileAccessor.GetSplashScreenAsync() ?? string.Empty,
+            MinimumSplashScreenDisplayTimeProperty => (await _myAppXamlFileAccessor.GetMinimumSplashScreenDisplayTimeAsync()).ToString() ?? string.Empty,
```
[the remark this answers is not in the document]
melytc: We discussed this in our 1:1 and addressed it in the last commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkPropertiesValueProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        if (propertyName == AuthenticationModeProperty)
+        {
+            value = value switch
+            {
+                "0" => "Windows",
+                "1" => "ApplicationDefined",
+                "" => "",
+
+                _ => throw new InvalidDataException($"Invalid value '{value}' for '{propertyName}' property.")
```
drewnoakes, line 128: `InvalidOperationException` might be more appropriate than `InvalidDataException`. The latter is intended for invalid data in streams. Not a big deal though.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        if (myAppDocument is null)
+        {
+            myAppDocument = new MyAppDocument(_docData);
+        }
+
+        return myAppDocument;
+
+        //return null; // Consider when we will reach here: when file doesn't exist.
```
drewnoakes, line 67: You should decide one way or another about this before merging.

drewnoakes (approved):
Nice!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkPropertiesValueProvider.cs
melytc: Good point, addressed in the next commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
melytc: Decided to do a try/catch when creating the DocData, and return null when we can't create it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkPropertiesValueProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+        });
+
+        return null;
+    }
+
+    private async Task<string> GetPropertyValueAsync(string propertyName)
+    {
+        string? value = propertyName switch
+        {
+            //ApplicationFrameworkProperty => (await _myAppXamlFileAccessor.GetMySubMainAsync()).ToString() ?? string.Empty,
```
MiYanni, line 107: nit: Commented code.

MiYanni (approved):

PR #8255, opened 2022-06-20
Handle null pointer in PropertyPages when getting StartupObjecs for VB

ocallesp:
Fixes [AB#1556918](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1556918)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8255)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
@@ -69,17 +69,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
                 return Array.Empty<IEnumValue>();
             }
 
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
My understanding of the bug and fix:

- `IEntryPointFinderService.FindEntryPoints` is not null annotated, and does not accept a null `symbol` argument (the first agument).
- The fix here is to prevent passing a `null` value when the compilation is `null`

@ocallesp could you also remove the `?` from the code after your change (line 86?) as it is no longer needed and might create confusion in future.

ocallesp:
> My understanding of the bug and fix:
> 
> * `IEntryPointFinderService.FindEntryPoints` is not null annotated, and does not accept a null `symbol` argument (the first agument).
> * The fix here is to prevent passing a `null` value when the compilation is `null`
> 
> @ocallesp could you also remove the `?` from the code after your change (line 86?) as it is no longer needed and might create confusion in future.

Done

PR #8282, opened 2022-06-28
Intercept UseApplicationFramework property.

### LINKED ISSUE #8257: Intercept `UseApplicationFramework` property in `ApplicationFrameworkPropertiesValueProvider`.
We need to intercept the `UseApplicationFramework` property in two interceptors:
1. `ApplicationFrameworkValueProvider`, where we set the `MyType` property in the project file.
2. `ApplicationFrameworkPropertiesValueProvider`, where we set the `MySubMain` property in the Application.myapp file.

We already have case 1 in place, which was done as the initial port for the VB pages.

This property is a special case because it is used by both the WPF and WinForms projects, and because it affects a property in two files, the project file and the myapp file. That means that we need two interceptors (which I tried to in my PR #8252 but got errors since we don't support having two interceptors for one property) or we use the interceptor we already have and make it aware that changes in the myapp file need to happen too.

Related to #8252.

melytc:
Fixes #8257.
I ended up merging the two interceptor classes in order to achieve this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8282)


MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -19,38 +33,72 @@ internal sealed class ApplicationFrameworkValueProvider : InterceptingPropertyVa
...
         private readonly UnconfiguredProject _project;
         private readonly IProjectItemProvider _sourceItemsProvider;
+        private IMyAppFileAccessor _myAppXamlFileAccessor;
 
         [ImportingConstructor]
         public ApplicationFrameworkValueProvider(
             UnconfiguredProject project,
-            [Import(ExportContractNames.ProjectItemProviders.SourceFiles)] IProjectItemProvider sourceItemsProvider)
+            [Import(ExportContractNames.ProjectItemProviders.SourceFiles)] IProjectItemProvider sourceItemsProvider,
+            IMyAppFileAccessor MyAppXamlFileAccessor)
```
drewnoakes, line 54: nit: lower case first letter

```suggestion
            IMyAppFileAccessor myAppXamlFileAccessor)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -19,38 +33,72 @@ internal sealed class ApplicationFrameworkValueProvider : InterceptingPropertyVa
...
+        internal const string SaveMySettingsOnExitProperty = "SaveMySettingsOnExit";
+        internal const string HighDpiModeProperty = "HighDpiMode";
+        internal const string AuthenticationModeProperty = "VBAuthenticationMode";
+        internal const string ShutdownModeProperty = "ShutdownMode";
+        internal const string SplashScreenProperty = "SplashScreen";
+        internal const string MinimumSplashScreenDisplayTimeProperty = "MinimumSplashScreenDisplayTime";
+
         private readonly UnconfiguredProject _project;
         private readonly IProjectItemProvider _sourceItemsProvider;
+        private IMyAppFileAccessor _myAppXamlFileAccessor;
```
drewnoakes, line 48: nit: readonly field

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -190,5 +240,114 @@ private static async Task<bool> IsWPFApplicationAsync(IProjectProperties default
...
+            {
+                unevaluatedPropertyValue = unevaluatedPropertyValue switch
+                {
+                    "AfterMainFormCloses" => "0",
+                    "AfterAllFormsClose" => "1",
+                    _ => unevaluatedPropertyValue
+                };
+            }
+
+            await(propertyName switch
```
drewnoakes, line 335: nit: this looks a bit like an invocation because of the `(`. Could be `await propertyName switch ...` instead.

drewnoakes (approved):
Just some nits.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
tmeschter, line 48: Did you mean `_myAppXmlFileAccessor` rather than `_myAppXamlFileAccessor`?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -190,5 +240,114 @@ private static async Task<bool> IsWPFApplicationAsync(IProjectProperties default
 
             return null;
         }
+
+        private async Task<string> GetPropertyValueAsync(string propertyName)
+        {
+            string? value = propertyName switch
```
tmeschter, line 246: `value` can't be null here; at worst it will be the empty string.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -190,5 +240,114 @@ private static async Task<bool> IsWPFApplicationAsync(IProjectProperties default
...
+            {
+                ApplicationFrameworkProperty => (await _myAppXamlFileAccessor.GetMySubMainAsync()).ToString() ?? string.Empty,
+                EnableVisualStylesProperty => (await _myAppXamlFileAccessor.GetEnableVisualStylesAsync()).ToString() ?? string.Empty,
+                SingleInstanceProperty => (await _myAppXamlFileAccessor.GetSingleInstanceAsync()).ToString() ?? string.Empty,
+                SaveMySettingsOnExitProperty => (await _myAppXamlFileAccessor.GetSaveMySettingsOnExitAsync()).ToString() ?? string.Empty,
+                HighDpiModeProperty => (await _myAppXamlFileAccessor.GetHighDpiModeAsync()).ToString() ?? string.Empty,
+                AuthenticationModeProperty => (await _myAppXamlFileAccessor.GetAuthenticationModeAsync()).ToString() ?? string.Empty,
+                ShutdownModeProperty => (await _myAppXamlFileAccessor.GetShutdownModeAsync()).ToString() ?? string.Empty,
+                SplashScreenProperty => await _myAppXamlFileAccessor.GetSplashScreenAsync() ?? string.Empty,
+                MinimumSplashScreenDisplayTimeProperty => (await _myAppXamlFileAccessor.GetMinimumSplashScreenDisplayTimeAsync()).ToString() ?? string.Empty,
```
tmeschter, line 256: Nit: you could probably extract out all the occurrences of `?? string.Empty`, though I'm not entirely sure that would be an improvement.

tmeschter (approved):
I would like to see unit tests for this, but maybe get this in and come back for those in a later change.

PR #8283, opened 2022-06-28
Hide new properties from Application Framework category.

melytc:
We can hold from showing these 3 new properties from the UI until we have the correct logic in place in the code generator for the Application.Designer.vb that will come in 17.4.

This is how it would look like:
image

Note: we agreed that it was okay to hide these 3 properties as they weren't surfaced in the old experience, and we want to ensure the end-to-end scenario works before introducing them.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8283)


MiYanni (approved):

drewnoakes (approved):
Is unhiding these tracked in a work item somewhere?

melytc:
I linked an issue to track the remaining work that needs to be done for these properties; thanks for the reminder, @drewnoakes!

Issue #8219, opened 2022-06-02
Migrate to Codecov GitHub action


MiYanni:
Currently, we use a [deprecated package](https://www.nuget.org/packages/Codecov/) to upload [Codecov](https://about.codecov.io/) information. A better solution is to use the [Codecov GitHub action](https://github.com/marketplace/actions/codecov). This came about after [this PR](https://github.com/dotnet/project-system/pull/8208) as we no longer needed a token to do this process.

PR #8285, opened 2022-06-29
Remove the use of the RoslynTools within our repo

MiYanni:
## TODO
- Documentation (including this description)
- `unit-tests` is failing since main rebase (will investigate next)
- Rebase again with `main` before undrafting


build/import/IntegrationTests.targets
```diff
@@ -1,38 +1,29 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This file was entirely whitespace changes. Integration tests do not function.

build/import/UnitTests.targets
```diff
@@ -1,66 +1,35 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This file was only whitespace changes to match the rest of the `.props`/`.targets`/`.*proj` files in the repo.

build/import/VisualStudio.XamlRules.targets
```diff
@@ -1,5 +1,6 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This file was entirely whitespace changes.

build/import/Workarounds.props
```diff
@@ -1,17 +1,21 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This file is also entirely whitespace changes.

setup/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles/Microsoft.VisualStudio.ProjectSystem.Managed.CommonFiles.swixproj
```diff
@@ -0,0 +1,34 @@
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+
+  <!-- Note: SwixBuild automatically adds the .vsix to FilesToSign via the AddTargetPathToFilesToSign target. -->
+
+  <Import Project="..\Directory.Build.props" />
+  <!-- This currently loads all SwixBuild packages, and multiple packages existing in your NuGet cache breaks the build. -->
+  <Import Project="$(MicroBuildCoreDir)Microsoft.VisualStudioEng.MicroBuild.Core.props" />
+  <!-- <Import Project="$(SwixBuildDir)Microsoft.VisualStudioEng.MicroBuild.Plugins.SwixBuild.props" /> -->
```
MiYanni, line 9: You'll see this commented Import (for props/targets) in both the `.swixproj` and `.vsmanproj`. I haven't decided to use this instead of the `MicroBuild.Core` as the Core-based import is how it is *supposed* to work and is more resilient to breaking long-term. However, it will cause **local builds** to be broken that contain multiple *SwixBuild* versions in your NuGet package cache. So, as it currently stands, our team will have issues building locally when this PR is merged until they remove those packages (old versions) from the cache. But long term, the Core-based solution is the right one as it imports *any* MicroBuild plugins that may be required to build these project types. I'm aware that the *SwixBuild* does build them but I'd have to verify if only importing that has all the information (properties/targets) required. Also, other repos I've looked at also do this Core-based import. I also started a thread with the MicroBuild team about this issue with multi-version *SwixBuild* imports using *MicroBuild.Core*, but haven't made much traction there.

setup/ProjectSystemSetup/ProjectSystemSetup.csproj
```diff
@@ -51,4 +62,18 @@
     </Compile>
   </ItemGroup>
 
+  <Target Name="VSManifestBuild" AfterTargets="Build">
```
MiYanni, line 65: TODO: I will add comments to targets like these.

setup/ProjectSystemSetup/source.extension.vsixmanifest
```diff
@@ -1,8 +1,7 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
+﻿<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="EDDCB5BE-3C18-444E-8E50-AE06CD345872" Version="|%CurrentProject%;GetBuildVersion|" Language="en-US" Publisher="Microsoft" />
+    <Identity Id="EDDCB5BE-3C18-444E-8E50-AE06CD345872" Version="|%CurrentProject%;GetVsixVersion|" Language="en-US" Publisher="Microsoft" />
```
MiYanni, line 4: TODO: Add comment on use of GetVsixVersion.

src/Microsoft.VisualStudio.AppDesigner/Resources/Designer.Designer.vb
```diff
@@ -14,56 +14,62 @@ Option Explicit On
...
         Friend Shared ReadOnly Property ResourceManager() As Global.System.Resources.ResourceManager
             Get
                 If Object.ReferenceEquals(resourceMan, Nothing) Then
+                    ' NOTICE: If this file is regenerated, the value below will change to "Designer". You need to manually change it back to "Microsoft.VisualStudio.AppDesigner.Designer" before pushing the changes.
+                    ' You should also add this comment back to the file so the next person is aware of the issue.
+                    ' Related links:
+                    ' - https://github.com/dotnet/project-system/issues/1058
+                    ' - https://github.com/dotnet/project-system/pull/6202
+                    ' - https://github.com/dotnet/project-system/pull/3741
                     Dim temp As Global.System.Resources.ResourceManager = New Global.System.Resources.ResourceManager("Microsoft.VisualStudio.AppDesigner.Designer", GetType(Designer).Assembly)
```
MiYanni, line 52: This file was regenerated and I added this big comment linking to the related issues for the problem that happens here when the file is regenerated.

src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
```diff
@@ -14,6 +14,8 @@
     <Description>This package implements the AppDesigner, which is the designer host used for project property pages among other things in Visual Studio</Description>
     <Summary>AppDesigner designer host implementation for Visual Studio project property pages</Summary>
     <PackageTags>Roslyn Project AppDesigner VisualStudio</PackageTags>
+    <!-- By default the SDK sets RootNamespace to the name of the project. Clear it here to avoid prefixing our namespaces. -->
+    <RootNamespace />
```
MiYanni, line 18: TODO: Consider putting this in the Languages.props.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/ExportRuleAttribute.cs
```diff
@@ -20,7 +24,7 @@ internal class ExportRuleAttribute : ExportPropertyXamlRuleDefinitionAttribute
         ///     One or more of <see cref="PropertyPageContexts"/>.
         /// </param>
         public ExportRuleAttribute(string ruleName, params string[] contexts)
-            : base(ThisAssembly.FullName, $"XamlRuleToCode:{ruleName}.xaml", string.Join(";", contexts))
+            : base(Assembly.GetExecutingAssembly().FullName, $"XamlRuleToCode:{ruleName}.xaml", string.Join(";", contexts))
```
MiYanni, line 27: In migrating to Nerdbank.GitVersioning, there is a generation mechanism that conflicted with our own, which created a `.ThisAssembly.cs` file. The only value that was not represented in the NBGV version was this value, `FullName`. There was 2 ways to approach replacing this (as FullName was only referenced one, here in this file). The simplest is to use this reflection call to `GetExecutingAssembly()`. If that isn't performant enough, then I have commented code at the top of the class to create a const string, using the exact same logic we used to generate in our `.ThisAssembly.cs` file.

version.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
+  "version": "17.3",
+  "assemblyVersion": "17.3",
```
MiYanni, line 4: TODO: This needs to be updated along with Versions.props. Determine if there is a simple way to do that. Otherwise, comment at these locations and update the New Branch Checklist items.

build/import/Packages.targets
```diff
@@ -2,41 +2,57 @@
...
+    <!-- https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/672/Archive-Symbols-with-Symweb?anchor=portable-pdbs -->
+    <PackageReference Update="Microsoft.DiaSymReader.Pdb2Pdb"                                         Version="1.1.0-beta2-22320-02" />
     <PackageReference Update="Microsoft.Net.Compilers.Toolset"                                        Version="4.2.0-1.final" />
-    <PackageReference Update="OpenCover"                                                              Version="4.7.1138-rc" />
-    <PackageReference Update="Codecov"                                                                Version="$(CodecovVersion)" />
-    <PackageReference Update="MicroBuild.Plugins.SwixBuild"                                           Version="$(MicroBuildPluginsSwixBuildVersion)" />
-    <PackageReference Update="Microsoft.DevDiv.Validation.MediaRecorder"                              Version="15.0.199" />
+    <!-- This package is deprecated. CodecovUploader is now the recommended package. -->
+    <PackageReference Update="Codecov"                                                                Version="1.13.0" />
```
MiYanni, line 52: This wasn't changed as this issue outlines be the better fix: https://github.com/dotnet/project-system/issues/8219

build/import/Versions.props
```diff
@@ -1,28 +1,17 @@
...
-    <ProjectSystemVersion>17.3.0</ProjectSystemVersion> <!-- This should always match with the VS release we're targeting -->
-    <VersionBase>$(ProjectSystemVersion)</VersionBase>
-    <PreReleaseVersionLabel>preview3</PreReleaseVersionLabel>
-    <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
-
-    <!-- Opt-in and out of repo features -->
-    <UsingToolVSSDK>true</UsingToolVSSDK>
-    <UsingToolIbcOptimization>true</UsingToolIbcOptimization>
-    <UsingToolXUnit>false</UsingToolXUnit>
-    <UsingToolMicrosoftNetCompilers>false</UsingToolMicrosoftNetCompilers>
```
MiYanni, line 15: The concept of "repo features" was specific to RepoToolset.

build/import/Versions.props
```diff
@@ -1,28 +1,17 @@
...
-
-    <!-- Toolset -->
-    <RoslynToolsRepoToolsetVersion>1.0.0-beta-62705-01</RoslynToolsRepoToolsetVersion>
-    <RoslynToolsModifyVsixManifestVersion>1.0.0-beta-62327-02</RoslynToolsModifyVsixManifestVersion>
-    <RoslynToolsSignToolVersion>1.1.0-beta3.21260.1</RoslynToolsSignToolVersion>
-    <MicrosoftDiaSymReaderPdb2PdbVersion>1.1.0-beta1-62624-01</MicrosoftDiaSymReaderPdb2PdbVersion>
-    <MicroBuildPluginsSwixBuildVersion>1.1.37</MicroBuildPluginsSwixBuildVersion>
-    <CodecovVersion>1.12.4</CodecovVersion>
```
MiYanni, line 23: Specifying package versions in this file was only necessary for very specific scenarios. Most of these packages are no longer used. The ones that are still used are now generating a path property, which is the only reason the version number was needed outside of a `PackageReference` node.

src/Microsoft.VisualStudio.AppDesigner/Microsoft.VisualStudio.AppDesigner.vbproj
MiYanni: Attempted this but it is set too early. The SDK ends up overriding it.
image

setup/ProjectSystemSetup/ProjectSystemSetup.csproj
MiYanni: Comment added.

setup/ProjectSystemSetup/source.extension.vsixmanifest
MiYanni: Comment added.

version.json
MiYanni: Updated branch-checklist.md and added comment to Versions.props.

build/import/GenerateThisAssemblySourceFile.targets
```diff
@@ -1,44 +0,0 @@
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This file is no longer necessary as all the properties (other than `FullName`) are generated by Nerdbank.GitVersioning. See ExportRuleAttribute.cs for the FullName workaround.

build/import/NuGet.props
```diff
@@ -1,10 +0,0 @@
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: We don't use this functionality (change the packages based on environment variable). `NuGetPackageRoot` already exists as part of the build normally.

build/import/OverrideRepoToolsetVersions.props
```diff
@@ -1,94 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
MiYanni, line 1: This was replaced by Nerdbank.GitVersioning.

build/import/ProjectLayout.props
```diff
@@ -1,16 +0,0 @@
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This was merged into the main Directory.Build.props.

build/proj/CodeCov.proj
```diff
@@ -1,32 +0,0 @@
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: This was moved to tests/Directory.Build.targets.

build/proj/GenerateDependentAssemblyVersionFile.proj
```diff
@@ -1,46 +0,0 @@
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
```
MiYanni, line 1: Replaced this with CreateDependentAssemblyVersionsCsv in setup/Directory.Build.targets

build/script/UpdateManifestJsonForSbom.ps1
```diff
@@ -1,29 +0,0 @@
-# Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
```
MiYanni, line 1: This workaround was no longer needed since we generate the SBOM element via SwixBuild now.

tests/Microsoft.VisualStudio.Editors.UnitTests/ObjectSerializerTests.vb
```diff
@@ -85,7 +85,7 @@ Namespace Microsoft.VisualStudio.Editors.UnitTests
...
...1BL3pObUFQOHptUUQvTTh3QS96UC9BUDltQUFEL1pqTUF6R1ptQVA5bW1RRC9ac3dBekdiL0FQK1pBQUQvbVRNQS81bG1BUCtabVFEL21jd0EvNW4vQVAvTUFBRC96RE1BLzh4bUFQL01tUUQvek13QS84ei9BUC8vTXdETS8yWUEvLytaQVAvL3pBQm1adjhBWnY5bUFHYi8vd0QvWm1ZQS8yYi9BUC8vWmdBaEFLVUFYMTlmQUhkM2R3Q0dob1lBbHBhV0FNdkx5d0N5c3JJQTE5ZlhBTjNkM1FEajQrTUE2dXJxQVBIeDhRRDQrUGdBOFB2L0FLU2dvQUNBZ0lBQUFBRC9BQUQvQUFBQS8vOEEvd0FBQVA4QS93RC8vd0FBLy8vL0FBQUFBQUFBQUFBVEV4TUFBQUFBQUFBQUFBQUFBQk1Udkx6c0V4TVJBQUFBQUFBQUV4Tzh2THk4dkJNVEFBQUFBQUFBN0x5OEd3SWJ2THh0RXhNQUFBQUFBT3diR3dJQ0FuSzg3RzBUQUFBQUFBRHNHMDhIQWhzQ3ZPeHRiUk1BQUFEc0d3Y0NCeHNiR3h1ODYyMXRBQUFBN0J0UEFnSUhBZ0lidk94dGJRQVRBT3diR3dJSEcrd0NCN3pzN0cxdEUrd2JHeHNiVDArODlmWDE5ZXpzYlJQc0d4dTh2THoxOVFjSEJ3ZjE3T3dUN0x5ODlmWHNCd2ZzR3h2c0cvWHNFK3oxOWV6c0J4dnM5ZlViN096c0FBRHM3QWNIRy9YczlmWHM3T3dBQUFBQUFBRHM3T3pzOWV6czdBQUFBQUFBQUFBQUFBRHM3T3pzQUFBQUFBQUFBQUQrRHdDUytBY2JHK0FIY2dMQUF5ZjF3QVByYmNBQkVoT0FBUUFBZ0FBQUFJQUFBSklBQUJzYkFBQUNBZ0FBQWljQUF4eHRBQThTRThBL0FBRHcvd0FBPC9JY29uRGF0YT48SWNvblNpemUgaTp0eXBlPSJhOlNpemUiIHhtbG5zPSIiPjxhOmhlaWdodD4wPC9hOmhlaWdodD48YTp3aWR0aD4wPC9hOndpZHRoPjwvSWNvblNpemU+PC9JY29uPjxOYW1lPlRyYXNoIENhbiAoRnVsbCk8L05hbWU+PFZhbHVlPjEzMzcuOTg8L1ZhbHVlPjwvQ29tcGxleFRlc3RDbGFzcz4="}
```
MiYanni, line 88: Since the build changed, it changed how this type serialized itself.

adamint:
After reading through the rationale of why we're using MicroBuild components directly instead of using Arcade, I still wanted to hear what you think the maintenance and rampup costs of maintaining our own infra layers will be, compared to the simplicity of using Arcade

MiYanni:
> After reading through the rationale of why we're using MicroBuild components directly instead of using Arcade, I still wanted to hear a little more on what you think the maintenance and rampup costs of owning our own infra layers will be, compared to the simplicity of using Arcade

There is part of this story that wasn't presented. It seems that other teams are having issues maintaining their engineering infrastructure because it has grown nearly 'beyond control'. What this causes is a social 'fear' or stigma around cleaning or fixing the infrastructure. It just becomes this behemoth that you do as little as possible to resolve issues, and then don't touch it otherwise.

Part of the goal here is to represent the minimal-viable-infrastructure required for our repo to do everything it needs to do. Removing the use of *roslyn-tools* is a good chunk of that goal. Then, add supporting documentation around how it works so no one feels intimidated by it. It will grow and change over time, just like a codebase does, but I'd like to be part of the solution to infrastructure management instead of just doing the same thing everyone else does and just letting another team handle it.

While *arcade* certainly works, the parallels between *arcade* and *roslyn-tools* are strong since they share history. I saw the issues with *roslyn-tools* where, because something 'works', it can fall into disrepair or be ignored. I'd rather set a higher standard than that and achieve a simplistic infrastructure. Something that other repos can ask us, "Hey, how did you make this so simple?" I've gotten so many questions when I did the simple PowerShell script to workaround SBOM issues. It was like people were perplexed that the problem didn't need a complicated solution. I feel our repo is fundamentally simple enough to not need the added complexity of *arcade*.

To answer your specific question, the simpler you keep the infrastructure, the less it costs to maintain and ramp-up. I know this team could 'offload' that cost to *arcade* but I feel like that is "using a sledgehammer to crack a nut." It is engineered to be hyper-generic and satisfy so many conditions and situations. Our concern and scope is so small that we shouldn't have issues maintaining it ourselves. That's what I believe, at least, after thinking about this for a year and a half now.

drewnoakes (approved):
This is a monumental amount of work! 🎉

@MiYanni walked the team through this today.

PR #8290, opened 2022-06-29
Make sure Prefer 32 Bit property is appearing when it should

### LINKED ISSUE #8016: Prefer 32-bit property not displaying when it should
## Summary

The "Build" page's "Prefer 32-bit" property should be visible in some cases, as defined here:

https://github.com/dotnet/project-system/blob/791fc3fe6d18391a5ab3b9cafbed3a9a04a3488b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml#L109-L120

## Steps to Reproduce

1. Create this project:

```xml

  <PropertyGroup>
    <TargetFramework>net48</TargetFramework>
    <OutputType>WinExe</OutputType>
  </PropertyGroup>

</Project>
```

2. Open the Project Properties UI 
3. "Prefer 32-bit" is not visible

## Expected Behavior

Should be visible.

## Actual Behavior

Is not visible.

adamint:
Currently, the property only shows up if the following condition is satisfied: `(has-evaluated-value "Build" "PlatformTarget" "AnyCPU")`

However, even though the value is saved to the csproj as AnyCPU, the actual unevaluated/evaluated value of AnyCPU is _Any CPU_. This change simply renames AnyCPU to Any CPU in the visibility condition.

Fixes #8016 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8290)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
@@ -122,7 +122,7 @@
         <NameValuePair.Value>
           (and
             (has-net-framework)
-            (has-evaluated-value "Build" "PlatformTarget" "AnyCPU")
+            (has-evaluated-value "Build" "PlatformTarget" "Any CPU")
             (or
               (has-evaluated-value "Application" "OutputType" "Exe")
               (has-evaluated-value "Application" "OutputType" "WinExe")
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.xaml
@@ -110,7 +110,7 @@
         <NameValuePair.Value>
           (and
             (has-net-framework)
-            (has-evaluated-value "Build" "PlatformTarget" "AnyCPU")
+            (has-evaluated-value "Build" "PlatformTarget" "Any CPU")
             (or
               (has-evaluated-value "Application" "OutputType" "Exe")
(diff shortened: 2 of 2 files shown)


MiYanni (approved):
Oh wow. Looks like this was a bug I made here: https://github.com/dotnet/project-system/commit/d3324600ecb0c4cfb67054d01b17938edad460dc

Not sure how I was supposed to know that it would need to evaluate "Any CPU" versus "AnyCPU". 🙃 I could have sworn it was based on the value for the property itself, which looking around online is definitely:
```xml
<PlatformTarget>AnyCPU</PlatformTarget>
```

But if you look up the same concept when passing it to MSBuild, it used "Any CPU" as the value:
```powershell
/p:Platform="Any CPU"
```

The rabbit hole of this problem seems deeper than what is on the surface. There might be a something with how we're collecting the data for these visibility evaluations. However, if this fixes the problem for now, so be it.

Note: I'm aware that `PlatformTarget` and `Platform` are different properties. But setting one seems to set the other, at least when I looked a binlogs testing this.

melytc (approved):

adamint:
> Oh wow. Looks like this was a bug I made here: [d332460](https://github.com/dotnet/project-system/commit/d3324600ecb0c4cfb67054d01b17938edad460dc)
> 
> Not sure how I was supposed to know that it would need to evaluate "Any CPU" versus "AnyCPU". 🙃 I could have sworn it was based on the value for the property itself, which looking around online is definitely:
> 
> ```
> <PlatformTarget>AnyCPU</PlatformTarget>
> ```
> 
> But if you look up the same concept when passing it to MSBuild, it used "Any CPU" as the value:
> 
> ```powershell
> /p:Platform="Any CPU"
> ```
> 
> The rabbit hole of this problem seems deeper than what is on the surface. There might be a something with how we're collecting the data for these visibility evaluations. However, if this fixes the problem for now, so be it.
> 
> Note: I'm aware that `PlatformTarget` and `Platform` are different properties. But setting one seems to set the other, at least when I looked a binlogs testing this.

It seems to _not_ come from our property interceptor for this property, which is strange...

PR #8291, opened 2022-06-29
Use editability conditions in build property pages

### LINKED ISSUE #8204: Allow properties to define an expression to control when they are editable
Today we have an expression that controls property visibility.

We could add a similar expression that controls when a property is marked as read only, conditional upon other property values.

This facility would simplify the approach taken in #8196 for example. In that PR, we add a duplicate of every property that might potentially be disabled, using mutually exclusive visibility conditions to hide the editable version and show the readonly one, based on other property values.

Here's an example of how this might work. The `EditabilityCondition` allows the UI to toggle the state of the UI element dynamically at runtime.

```xml
  <DynamicEnumProperty Name="SomeProperty"
                       ...>
    <DynamicEnumProperty.Metadata>
      <NameValuePair Name="EditabilityCondition">
        <NameValuePair.Value>
          (not (has-evaluated-value "Build" "WarningSeverity" "DisableAll"))
        </NameValuePair.Value>
      </NameValuePair>
  </DynamicEnumProperty>
```

The majority of properties would not define a `EditabilityCondition`, and there would be no overhead for those properties.

adamint:
Editability conditions replace the need for readonly properties, so we can remove the existing readonly properties in build property pages. This is a draft PR while the impl of the EditabilityCondition is done.

Resolves #8204 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8291)


drewnoakes (approved):

adamint:
This can be merged once https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/410209 is in VS main. 

Additionally, https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1566265/ should be set to resolved after this PR is merged.

MiYanni (approved):

docs/repo/property-pages/property-specification.md
```diff
@@ -212,7 +212,37 @@ The property's string value should be encoded with format resembling `A=1,B=2`,
...
+            <ValueEditor.Metadata>
+                <NameValuePair Name="TypeDescriptorText" Value="Imported Namespaces" />
+                <NameValuePair Name="AllowsCustomStrings" Value="True" />
+            </ValueEditor.Metadata>
+        </ValueEditor>
+    </DynamicEnumProperty.ValueEditors>
+</DynamicEnumProperty>
+```
+
+This example is taken from the _Imported Namespaces_ property on the Visual Basic project properties references page.
```
tmeschter, line 245: You could add a link to that .xaml file in source.

docs/repo/property-pages/visibility-conditions.md
```diff
@@ -10,8 +10,9 @@ Visibility conditions are not interpreted by the back-end, which treats them as
 
 ## Specification
 
-There are three kinds of visibility conditions:
+There are three four of visibility conditions:
```
tmeschter, line 13: Err, "four kinds" not "three four".

docs/repo/property-pages/visibility-conditions.md
```diff
@@ -10,8 +10,9 @@ Visibility conditions are not interpreted by the back-end, which treats them as
...
 - VisibilityCondition is a visibility condition on a property (whether the property should be shown).
+- EditabilityCondition is a visibility condition on a property that determines whether the property should editable (if true or not specified) or readonly (if evaluated to false).
```
tmeschter, line 15: I wouldn't call this a visibility condition per se, as it doesn't affect the visibility of anything. :-)

tmeschter (approved):

docs/repo/property-pages/visibility-conditions.md
drewnoakes: Also singular _condition_.

docs/repo/property-pages/visibility-conditions.md
drewnoakes: ...should _be_ editable...

docs/repo/property-pages/visibility-conditions.md
adamint: as seen by the sentences I wrote, I can english very well @tmeschter 😄 
@drewnoakes plural conditions is correct in american english

adamint:
Note: I updated the docs to refer to the broader category of visibility+isreadonly conditions as "property conditions." If there's something I should be calling them instead, please let me know @drewnoakes

docs/repo/property-pages/visibility-conditions.md
drewnoakes: Thanks. TIL there is a difference of opinion on this! https://english.stackexchange.com/questions/5539/types-of-things-vs-types-of-thing

I'd write:
- Four visibility conditions
- Four kinds of visibility condition

Apparently Edge would write:

image

😆

drewnoakes (approved):

PR #8301, opened 2022-07-01
Localized file check-in by OneLocBuild Task: Build definition ID 981: Build ID 1857583

MiYanni:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8301)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/xlf/VSResources.ja.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/xlf/VSResources.ja.xlf
@@ -112,7 +112,7 @@
       </trans-unit>
       <trans-unit id="IncrementalBuildFailureWarningMessage_2">
         <source>WARNING: Potential incremental build failure in '{0}'. {1}. See https://aka.ms/incremental-build-failure.</source>
-        <target state="new">WARNING: Potential incremental build failure in '{0}'. {1}. See https://aka.ms/incremental-build-failure.</target>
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/xlf/VSResources.ja.xlf
```diff
@@ -112,7 +112,7 @@
       </trans-unit>
       <trans-unit id="IncrementalBuildFailureWarningMessage_2">
         <source>WARNING: Potential incremental build failure in '{0}'. {1}. See https://aka.ms/incremental-build-failure.</source>
-        <target state="new">WARNING: Potential incremental build failure in '{0}'. {1}. See https://aka.ms/incremental-build-failure.</target>
+        <target state="translated">警告: '{0}' での潜在的なインクリメンタル ビルド エラー。参照: https://aka.ms/incremental-build-failure</target>
```
MiYanni, line 115: We've technically already accepted this translation here: https://github.com/dotnet/project-system/blob/b6ca713b222980af880a7d2e812957fafe55d0f5/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/xlf/ja/VSResources.xlf.lcl#L269

Clearly, this specific translation happened in-between this string being changed. Thus, confused the loc system. I'll rebuild this branch and if we continue getting translation issues with this specific string, we need to report it.

PR #8320, opened 2022-07-12
Allow multi string selector to define readonly boolean

### LINKED ISSUE #8311: Disable namespaces that are unable to be unselected
### Summary

As can be seen in the below image, namespaces that are unable to be unselected are still enabled. There should be an extension to the MultiStringSelector editor/control to allow for a list of shown but disabled strings.

 image

### User impact

There is low user impact, as no functionality is actually affected.

adamint:
This sets the property string to follow the same pattern as environment variable strings - having a name-value pair where the value is whether the string is readonly.

Fixes #8311 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8320)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyValuePairListEncoding.cs
```diff
@@ -0,0 +1,94 @@
...
+                        escaped = !escaped;
+                    }
+                    else
+                    {
+                        escaped = false;
+                    }
+                }
+
+                return (string.Empty, string.Empty);
+            }
```
MiYanni, line 74: I spent some time looking at these since the logic is written as (in English):
> If `[character]` is proceeded by `/`, ignore it while splitting by `[character]`.

There's a very non-obtuse way to do that, `Regex.Split`. It is like `String.Split`, but does it based on a regular expression instead of just a character. Here's an example for the first function, where you want to split on `,`, but not on `/,`.

```C#
// See https://aka.ms/new-console-template for more information
using System.Text.RegularExpressions;

var string1 = "fdsafdsa,fdsafdsa,fdasfdsa/,dsafdsa/,dasfdsa,fdsa,fdsa,";
var string2 = "fgdg;g/g/fds/gfd;,,;;,;,gfds;.g;fd.s;,;fd.s;///fd;s,gf;ds.g;fd/,/gfd,/,d/sg,f/ds";

var regex = new Regex(@"(?<!\/),");
Console.WriteLine("1:");
regex.Split(string1).ToList().ForEach(Console.WriteLine);
Console.WriteLine("2:");
regex.Split(string2).ToList().ForEach(Console.WriteLine);
```

The expression used is `(?<!\/),` which says, "Match `,` that isn't preceded by `/`". The result is:
image

This works the same as your code. You can do the same for `=` in the second function. Here's some simpler ways I wrote the function before using `Regex.Split`.

This is the same as yours but removes the boolean because it is completely redundant.
```C#
static IEnumerable<string> ReadEntries(string rawText)
{
    int entryStart = 0;
    for (int i = 0; i < rawText.Length; i++)
    {
        // The character prior to ',' cannot be '/'.
        if (rawText[i] == ',' && !(i > 0 && rawText[i - 1] == '/'))
        {
            yield return rawText.Substring(entryStart, i - entryStart);
            entryStart = i + 1;
        }
    }

    yield return rawText.Substring(entryStart);
}
```

Second one here splits them based on `,` but then combines those that end in `/`. It is not efficient but allowed me to realize I simply wanted a conditional `String.Split` (which is what `Regex.Split` is).
```C#
static IEnumerable<string> ReadEntries2(string rawText)
{
    var entries = rawText.Split(',');
    var current = string.Empty;
    foreach(var entry in entries)
    {
        // If the entry ends with a "/", we want to combine it with subsequent entries.
        if (entry.EndsWith("/"))
        {
            current += entry;
            continue;
        }

        yield return current + entry;
        current = string.Empty;
    }
}
```

MiYanni (changes requested):
The logic for `KeyValuePairListEncoding` should have unit tests so that it is guaranteed to work as intended.

adamint:
@MiYanni the KeyValuePairListEncoding is copied from CPS with no changes, so @drewnoakes could you address the comments on this? (it was just renamed, no logical change)

MiYanni:
> @MiYanni the KeyValuePairListEncoding is copied from CPS with no changes, so @drewnoakes could you address the comments on this? (it was just renamed, no logical change)

Oh... 🙃

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
```diff
@@ -22,82 +22,86 @@ public ImportedNamespacesValueProvider(ConfiguredProject configuredProject, IPro
...
+                .Select(item => (Value: item.EvaluatedInclude, IsImported: item.IsImported))
+                .Where(import => !string.IsNullOrEmpty(import.Value))
                 .ToImmutableArray();
         });
 
         return existingImports;
     }
     
-    private async Task<string> GetSelectedImportStringAsync()
+    private async Task<IEnumerable<(string Name, string Value)>> GetSelectedImportStringAsync()
```
drewnoakes, line 42: Minor: this method's name says it returns a string, but no longer does.

Rather than rename it to `GetSelectedImportNameValuePairsAsync`, you could keep it returning a string and add the `_encoding.Format` call here instead. That'd remove duplicate calls to that method below too.

Also, you can remove the collection allocation and stream the values using Linq, which the encoding function handles nicely. Something like:

```c#
private async Task<string> GetSelectedImportStringAsync()
{
    ImmutableArray<(string Value, bool IsImported)> existingImports = await GetSelectedImportsAsync();

    return _encoding.Format(
        existingImports
            .Where(pair => !string.IsNullOrEmpty(pair.Value))
            .Select(pair => (Key: pair.Value, Value: pair.IsImported.ToString())));
}
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyValuePairListEncoding.cs
```diff
@@ -0,0 +1,94 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug;
+
+ using System.Collections.Generic;
+ using System.Linq;
+
+    internal sealed class KeyValuePairListEncoding
```
tmeschter, line 8: Nit: the class is indented.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
```diff
@@ -40,64 +41,57 @@ private async Task<ImmutableArray<string>> GetSelectedImportsAsync()
...
-                sb.Append(import).Append(';');
-            }
-        }
+        ImmutableArray<(string Value, bool IsImported)> existingImports = await GetSelectedImportsAsync();
+
+        var selectedImports = existingImports
+                .Where(pair => !string.IsNullOrEmpty(pair.Value))
+                .Select(pair => (Key: pair.Value, Value: pair.IsImported.ToString())).ToList();
 
         if (!containsProjectName)
```
tmeschter, line 53: Looks like `containsProjectName` will always be false as it is only assigned at the definition.

adamint: thank you for the catch

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
```diff
@@ -40,64 +41,57 @@ private async Task<ImmutableArray<string>> GetSelectedImportsAsync()
...
     {
-        return GetSelectedImportStringAsync();
+        return await GetSelectedImportStringAsync();
     }
 
-    public override Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
+    public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
     {
-        return GetSelectedImportStringAsync();
+        return await GetSelectedImportStringAsync();
```
drewnoakes, line 68: There are pros and cons to this change, but it does add an extra `Task<>` allocation for each call. The pro is that this method would appear in any async stacks, whereas if you just returned the task directly (rather than being async and awaiting it) you'd see this method in the continuation chain.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
```diff
@@ -114,19 +108,19 @@ public override Task<string> OnGetEvaluatedPropertyValueAsync(string propertyNam
...
         }
 
         // Verify we have at least one valid import to add before acquiring a write lock.
-        if (importsToAdd.Any(importToAdd => importToAdd.Length > 0))
+        if (importsToAdd.Any(importToAdd => importToAdd.Key.Length > 0))
         {
             await _projectAccessor.OpenProjectXmlForWriteAsync(_configuredProject.UnconfiguredProject, project =>
             {
-                foreach (string importToAdd in importsToAdd)
+                foreach (KeyValuePair<string, bool> importToAdd in importsToAdd)
```
drewnoakes, line 119: You can use deconstruction here too if you like.

```suggestion
                foreach ((string key, _) in importsToAdd)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ReferencesPage/ImportedNamespacesValueProvider.cs
adamint: What would you recommend?

drewnoakes: We generally favour avoiding the allocation, but it's not a huge deal in this case as it's not a tight loop. Feel free to leave it.

drewnoakes (approved):

PR #8330, opened 2022-07-18
Exit import scanning foreach when first match found

drewnoakes:
For an SDK-style project there may be hundreds of these imports. As soon as we find a match, we can process it an exit the loop.


melytc (approved):

MiYanni (approved):

tmeschter (commented):
The expectation is that there is only one matching import element, so once we find it we don't need to look for others?

PR #8333, opened 2022-07-19
Create Application.myapp if none exists during property access/set

### LINKED ISSUE #8258: Create Application.myapp file when one is not found in project directory.
I need to validate what the correct experience should be when the `Application.myapp` does not exist in a VB WinForms project and the user tries to set the Application Framework properties: my assumption is that we need to create one.

Related #8252.

adamint:
This PR modifies the MyAppFileAccessor to check if the Application.myapp file exists in the project when obtaining a MyAppDocument instance (when reading/setting a property). If none exists, one is created with the contents being the default winforms template.

Fixes #8258

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
@@ -15,8 +15,8 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
     private readonly IProjectThreadingService _threadingService;
     private readonly IPhysicalProjectTreeStorage _storage;
     private DocData? _docData;
-    private MyAppDocument? myAppDocument;
-    private string _absolutePath;
+    private MyAppDocument? _myAppDocument;
+    private readonly string _absolutePath;
 
     private const string MySubMainProperty = "MySubMain";
     private const string MainFormProperty = "MainForm";
@@ -63,7 +63,7 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
 
     private void DocData_Modifying(object sender, EventArgs e)
     {
-        myAppDocument = null;
+        _myAppDocument = null;
     }
 
     private async Task<MyAppDocument?> TryGetMyAppFileAsync()
@@ -88,7 +88,7 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
         {
             _docData = new DocData(_serviceProvider, _absolutePath);
             _docData.Modifying += DocData_Modifying;
-            myAppDocument = new MyAppDocument(_docData);
+            _myAppDocument = new MyAppDocument(_docData);
         }
         catch (NullReferenceException)
         {
@@ -98,7 +98,7 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
 
         await TaskScheduler.Default;
 
-        return myAppDocument;
+        return _myAppDocument;
     }
 
     private async Task SetPropertyAsync(string propertyName, string value)
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
@@ -90,7 +90,7 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
             _docData.Modifying += DocData_Modifying;
             myAppDocument = new MyAppDocument(_docData);
         }
-        catch (Exception)
+        catch (NullReferenceException)
         {
             // If we've reached here, the file write succeeded but the file may have been deleted by the time UI thread is switched to
             return null;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
@@ -70,7 +70,6 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
     {
         if (_docData?.Data is null)
         {
-            await TaskScheduler.Default;
             // Create My Project directory if it doesn't exist. If it does, nothing happens
             await _storage.CreateFolderAsync("My Project");
 
@@ -94,6 +93,7 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
         catch (Exception)
         {
             // If we've reached here, the file write succeeded but the file may have been deleted by the time UI thread is switched to
+            return null;
         }
 
         await TaskScheduler.Default;
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
@@ -29,16 +29,19 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
     private const string SplashScreenProperty = "SplashScreen";
     private const string MinimumSplashScreenDisplayTimeProperty = "MinimumSplashScreenDisplayTime";
 
-    private const string DefaultMyappFileContents = "<?xml version=\"1.0\" encoding=\"utf-8\"?>" +
-                                                    "<MyApplicationData xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\">" +
-                                                    "  <MySubMain>true</MySubMain>" +
-                                                    "  <MainForm>Form1</MainForm>" +
-                                                    "  <SingleInstance>false</SingleInstance>" +
-                                                    "  <ShutdownMode>0</ShutdownMode>" +
-                                                    "  <EnableVisualStyles>true</EnableVisualStyles>" +
-                                                    "  <AuthenticationMode>0</AuthenticationMode>" +
-                                                    "  <SaveMySettingsOnExit>true</SaveMySettingsOnExit>" +
-                                                    "</MyApplicationData>";
+    private const string DefaultMyappFileContents =
+        """
+        <?xml version="1.0" encoding="utf-8"?>
+        <MyApplicationData xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+          <MySubMain>true</MySubMain>
+          <MainForm>Form1</MainForm>
+          <SingleInstance>false</SingleInstance>
+          <ShutdownMode>0</ShutdownMode>
+          <EnableVisualStyles>true</EnableVisualStyles>
+          <AuthenticationMode>0</AuthenticationMode>
+          <SaveMySettingsOnExit>true</SaveMySettingsOnExit>
+        </MyApplicationData>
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -47,20 +65,34 @@ private void DocData_Modifying(object sender, EventArgs e)
...
+                await writer.WriteAsync(DefaultMyappFileContents);
+                await writer.FlushAsync();
+                writer.Close();
+                
+                // Now that we've created the file, we can instantiate _docData and myAppDocument.
+                _docData = new DocData(_serviceProvider, _absolutePath);
+                _docData.Modifying += DocData_Modifying;
+
+                myAppDocument = new MyAppDocument(_docData);
+                return myAppDocument;
```
melytc, line 95: I wonder if we could extract this, so we don't have to use the main thread to create the location folder, create the file, write the file, flush the writer, and close it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -47,20 +65,34 @@ private void DocData_Modifying(object sender, EventArgs e)
 
     private async Task<MyAppDocument?> TryGetMyAppFileAsync()
     {
-        if (_docData is null)
+        if (_docData?.Data is null || !File.Exists(_absolutePath))
```
melytc, line 68: Checking if the file path exists sounds like a double validation here: we already know that if the DocData is null it means that there is no file, thus the file path doesn't exist neither.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -27,12 +29,28 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
...
+    public MyAppFileAccessor([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
+        UnconfiguredProject project,
+        IProjectThreadingService threadingService,
+        IPhysicalProjectTree projectTree)
     {
         _serviceProvider = serviceProvider;
         _project = project;
         _threadingService = threadingService;
+        _storage = projectTree.TreeStorage;
+        _absolutePath = _project.MakeRooted("My Project\\Application.myapp");
```
MiYanni, line 53: nit: Use string literal
```suggestion
        _absolutePath = _project.MakeRooted(@"My Project\Application.myapp");
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -47,20 +65,34 @@ private void DocData_Modifying(object sender, EventArgs e)
...
             {
-                _docData = new DocData(_serviceProvider, absolutePath);
+                _docData = new DocData(_serviceProvider, _absolutePath);
                 _docData.Modifying += DocData_Modifying;
             } 
             catch (Exception)
             {
-                // If we have reached here, it means that the file doesn't exist in that location.
-                return null;
+                // If we have reached here, it means that the file doesn't exist in that location, so try to create it
```
MiYanni, line 78: Generally, doing actual logic in a `catch` block is not a good design, especially writing files and other I/O. I'd do some adjustment in how we determine the myapp file needs created. This `catch` should only be for exceptional situations.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
adamint: Done, thanks!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -27,12 +29,28 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
...
+    private const string DefaultMyappFileContents = "<?xml version=\"1.0\" encoding=\"utf-8\"?>" +
+                                                    "<MyApplicationData xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\">" +
+                                                    "  <MySubMain>true</MySubMain>" +
+                                                    "  <MainForm>Form1</MainForm>" +
+                                                    "  <SingleInstance>false</SingleInstance>" +
+                                                    "  <ShutdownMode>0</ShutdownMode>" +
+                                                    "  <EnableVisualStyles>true</EnableVisualStyles>" +
+                                                    "  <AuthenticationMode>0</AuthenticationMode>" +
+                                                    "  <SaveMySettingsOnExit>true</SaveMySettingsOnExit>" +
+                                                    "</MyApplicationData>";
```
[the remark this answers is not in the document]
adamint: I wasn't aware we were allowed to use c# 11 features!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
adamint: Refactored!

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -65,43 +68,35 @@ private void DocData_Modifying(object sender, EventArgs e)
...
+            _docData = new DocData(_serviceProvider, _absolutePath);
+            _docData.Modifying += DocData_Modifying;
             myAppDocument = new MyAppDocument(_docData);
         }
+        catch (Exception)
+        {
+            // If we've reached here, the file write succeeded but the file may have been deleted by the time UI thread is switched to
+        }
+
+        await TaskScheduler.Default;
```
drewnoakes, line 99: I don't think switching back here does anything useful. What happens in the caller if you remove this?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -65,43 +68,35 @@ private void DocData_Modifying(object sender, EventArgs e)
...
-        if (myAppDocument is null)
+        
+        await _threadingService.SwitchToUIThread();
+        try
         {
+            _docData = new DocData(_serviceProvider, _absolutePath);
+            _docData.Modifying += DocData_Modifying;
             myAppDocument = new MyAppDocument(_docData);
         }
+        catch (Exception)
```
drewnoakes, line 94: Catching all exceptions like this is generally a bad idea, especially when the catch block is empty.

What are you actually preventing against here, and is there an exception type for that specifically?

adamint: I meant to keep the existing logic, which catches all exceptions and returns null
https://github.com/dotnet/project-system/blob/5136a986dd9b987ab8a0824934b271998605fe8d/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs#L60

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
adamint: I thought this was your prior recommendation? All seems good if I remove this

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
MiYanni: I just checked (git blame is your friend) and that logic comes from the comment (and PR) here: https://github.com/dotnet/project-system/pull/8252#discussion_r903592523

What Drew said is correct, it is generally poor design to 'catch everything'; for the reason is it can hide issues or even cause them. In this case, it was done to avoid when the file doesn't exist. You can ask @melytc about it, and figure out the appropriate exception for that situation.

melytc: A `NullReferenceException` may be adequate here: we want to catch the exception raised when trying to create the `DocData`, which can happen if the `absolutePath` is not found given there's no file.

adamint: Gotcha, thanks to the both of you 😄 I'll change to NullReferenceException

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
drewnoakes: > I thought this was your prior recommendation?

My suggestion was to use a local function to enclose all the stuff that needs the UI thread. Something like this:

```c#
// on background thread
await LocalFunctionAsync();
// still on background thread

async Task LocalFunctionAsync()
{
    await SwitchToMainThreadAsync();
    // on main thread
}
```

In your case, you don't do any more real work after the main thread code, so it's pretty much equivalent.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -13,8 +13,10 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
     private readonly IServiceProvider _serviceProvider;
     private readonly UnconfiguredProject _project;
     private readonly IProjectThreadingService _threadingService;
+    private readonly IPhysicalProjectTreeStorage _storage;
     private DocData? _docData;
     private MyAppDocument? myAppDocument;
+    private string _absolutePath;
```
drewnoakes, line 19: This field can be readonly.

That said, I don't think we need to retain a reference to this path unless `TryGetMyAppFileAsync` is called very often for a given project.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/WindowsForms/MyAppFileAccessor.cs
```diff
@@ -13,8 +13,10 @@ internal class MyAppFileAccessor : IMyAppFileAccessor, IDisposable
     private readonly IServiceProvider _serviceProvider;
     private readonly UnconfiguredProject _project;
     private readonly IProjectThreadingService _threadingService;
+    private readonly IPhysicalProjectTreeStorage _storage;
     private DocData? _docData;
     private MyAppDocument? myAppDocument;
```
drewnoakes, line 18: I know it's not this PR, but if you're editing this file could you add an underscore to this field name so that it's consistent with our naming style?

drewnoakes (approved):
Nice!

PR #8336, opened 2022-07-21
Ensure incremental build failure checks use correct configuration

drewnoakes:
Previously, we assumed that all builds we were checking for incremental build failures were building the active configuration.

This is not always the case. For example, when publishing an application, you can configure the publish to occur in "Release" configuration, while VS's active configuration is "Debug".

The prior discrepancy could potentially cause us to incorrectly identify builds that occur for publish to have incremental build failures.

This commit fixes that by obtaining the specific `ConfiguredProject` being built via the `IVsCfg` object passed via the Solution Build Manager to our build event handler.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8336)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.cs
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.ProjectSystem.UpToDate;
 using Microsoft.VisualStudio.ProjectSystem.VS.Editor;
 using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Build/Diagnostics/IncrementalBuildFailureDetector.cs
```diff
@@ -114,11 +112,9 @@ int IVsUpdateSolutionEvents2.UpdateProjectCfg_Done(IVsHierarchy pHierProj, IVsCf
                     return HResult.OK;
                 }
 
-                if (IsRelevantBuild(dwAction))
+                if (IsRelevantBuild(dwAction) && pCfgProj is IVsBrowseObjectContext { ConfiguredProject: { } configuredProject })
                 {
-                    UnconfiguredProject? unconfiguredProject = _projectServiceAccessor.GetProjectService().GetUnconfiguredProject(pHierProj, appliesToExpression: BuildUpToDateCheck.AppliesToExpression);
-
-                    IProjectChecker? checker = unconfiguredProject?.Services.ExportProvider.GetExportedValueOrDefault<IProjectChecker>();
+                    IProjectChecker? checker = configuredProject.Services.ExportProvider.GetExportedValueOrDefault<IProjectChecker>();
```
drewnoakes, line 117: This is the core change. Here we are using the `pCfgProj` to obtain the relevant `IProjectChecker`. Before we were using the `pHierProj`, which only gave us `UnconfiguredProject`. With the change we can find the `ConfiguredProject` that was built.

The rest of this PR is moving `IProjectChecker` from `UnconfiguredProject` scope to `ConfiguredProject` scope.

davkean (approved):

MiYanni (approved):

tmeschter (approved):

PR #8337, opened 2022-07-21
Use ordered rule snapshot data

drewnoakes:
CPS recently added the ability to acquire data (items/properties) from project data subscription snapshots, by casting those snapshots to `IDataWithOriginalSource<T>`.

Using this type in the language service host means we no longer need to build our own ordered snapshot derived from `IProjectBuildSnapshot.TargetOutputs["CompileDesignTime"]`. We can use the rule data instead.

Previously we were using both sources of data. Now we only need one, so can remove quite a bit of code.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8337)


drewnoakes:
I need to extend the tests to provide `IDataWithOriginalSource<>` as an alternative implementation of our immutable dictionaries in test snapshot data.

drewnoakes:
> I need to extend the tests to provide `IDataWithOriginalSource<>` as an alternative implementation of our immutable dictionaries in test snapshot data.

This has been done.

MiYanni (approved):

tmeschter (approved):

ocallesp (approved):

ocallesp:
Unit-test `BuildUpdate_InvokesCommandLineHandlerWhenChangesExist `failed

MiYanni:
@drewnoakes Looks like the `BuildUpdate_InvokesCommandLineHandlerWhenChangesExist` test is still failing for these changes.

drewnoakes:
Yes I need to review the expectations set on a mock to get it passing. I'll get to it in turn.

drewnoakes:
Includes #8479, so merge after that.

PR #8321, opened 2022-07-14
Optimise language service during configuration switches

drewnoakes:
Addresses half of #7541 (the switching half, not the reload half).
Fixes #7615

This change alters how the language service constructs and manages Roslyn workspace objects.

Previously, a workspace would be created per `ConfiguredProject`. With this change, one is created per "slice".

The easiest way to explain slices is with an example. Consider the following configurations:

- `Debug|AnyCPU|net48`
- `Debug|AnyCPU|net60`
- `Release|AnyCPU|net48`
- `Release|AnyCPU|net60`

If `Debug|AnyCPU` is active in the IDE, then the first two configurations are active. If the user switches to `Release`, those two are deactivated, and the last two become active.

When considering slices, the configuration (eg. `Debug`/`Release`) and platform (eg. `AnyCPU`) are ignored. Only additional (implicit) dimensions are considered (eg. `TargetFramework`).

CPS provides a new facility (via `IActiveConfigurationGroupSubscriptionService`) that allows subscribing to data for (in this example) `net48` and `net60` slices of the project. If the user switches configuration, the same subscription is used, however any difference in the data between configurations is received (via Dataflow) as a delta.

Note that for single-targeting projects (i.e. a single target framework) there is only one slice for the project, and it still updates as described.

Given we now receive a delta when the configuration changes, we are able to pass that delta on to Roslyn. Roslyn will be able to optimise this process to avoid discarding all work done for one configuration, and starting fresh with another configuration when a switch occurs.

We don't have MEF scopes for slices, and slices do not belong to a specific configuration, so MEF components that work with slices needed to be moved from `ConfiguredProject` scope to `UnconfiguredProject` scope.

This change also reduces the number of types/interfaces/concepts involved in creating, updating and generally working with workspaces.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8321)

PR #8340, opened 2022-07-25
Fix scheduling of evaluation/build data application

drewnoakes:
Follows from #8321

Previously the code intended to make build updates wait for the application of initial evaluation data had two problems:

1. The `await` occurred within the `ApplyProjectBuild` callback provided to `OnProjectChangedAsync`, however that method attempts to dereference the `ContextId` which is invalid until the first evaluation data has been applied. This commit moves the `await` earlier to prevent this.

2. The `ApplyProjectBuild` callback was `async void` which is an anti-pattern in general. In this particular case, it broke the expectation that a lock is held throughout the operation, as the first yield within the async method would return control to the caller, which would then release the lock while the remainder of the `async void` method continued unprotected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8340)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
@@ -374,6 +374,12 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
         Assumes.True(_state is WorkspaceState.Initialized);
         Assumes.NotNull(_buildProgressRegistration);
 
+        // The Roslyn workspace context is created when the first evaluation data arrives.
+        // It's possible that build data arrives before evaluation data, in which case
+        // the Roslyn context would be null. To prevent problems, we wait for evaluation
+        // data to have been processed at least once before continuing.
+        await _hasEvaluationData.Task;
+
         await OnProjectChangedAsync(
             _buildProgressRegistration,
             update,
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -385,17 +391,11 @@ private async Task OnBuildUpdateAsync(IProjectVersionedValue<BuildUpdate> update
 
         return;
 
-        async void ApplyProjectBuild(
+        void ApplyProjectBuild(
```
drewnoakes, line 394: The `async void` in the previous code was an incomplete refactor on my part, fixed here.

We have an analyzer that should have flagged this, but it doesn't work for local functions apparently. I've filed https://github.com/microsoft/vs-threading/issues/1074 to track that.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -374,6 +374,12 @@ private async Task OnBuildUpdateAsync(IProjectVersionedValue<BuildUpdate> update
         Assumes.True(_state is WorkspaceState.Initialized);
         Assumes.NotNull(_buildProgressRegistration);
 
+        // The Roslyn workspace context is created when the first evaluation data arrives.
+        // It's possible that build data arrives before evaluation data, in which case
+        // the Roslyn context would be null. To prevent problems, we wait for evaluation
+        // data to have been processed at least once before continuing.
+        await _hasEvaluationData.Task;
```
davkean, line 381: It looks like this might fault the block and report a error? Previously, we'd not fire up the language service for projects that had bad language data, (NoTargets for example would fall into this I think).

davkean (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
drewnoakes: I just ran some tests with `NoTargets` and didn't see anything problematic from the user's perspective, but I agree this should be investigated. I would have expected it to fail and either fault the project or silently report an NFE.

Filed https://github.com/dotnet/project-system/issues/8343 to track this.

davkean: Set one of the properties its reading during init to empty.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
drewnoakes: Submitted https://github.com/microsoft/vs-threading/pull/1075 to extend the analyzer to cover local functions and async lambdas.

PR #8341, opened 2022-07-25
Fix shared project after rename from "build" to "eng"

drewnoakes:
### Before

image

(Nothing to expand.)

### After

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8341)

### DIFF
--- a/eng/eng.shproj
+++ b/eng/eng.shproj
@@ -8,6 +8,6 @@
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
   <PropertyGroup />
-  <Import Project="build.projitems" Label="Shared" />
+  <Import Project="eng.projitems" Label="Shared" />
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
 </Project>


MiYanni (approved):

MiYanni:
My bad. 😅

melytc (approved):

eng/eng.shproj
```diff
@@ -8,6 +8,6 @@
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
   <PropertyGroup />
-  <Import Project="build.projitems" Label="Shared" />
+  <Import Project="eng.projitems" Label="Shared" />
```
tmeschter, line 11: This didn't raise an error somewhere?

tmeschter (approved):

eng/eng.shproj
MiYanni: See, that's what I would have assumed too... when renaming stuff, I try to be thorough and hope that other things will error or cause some failure if I miss it. The shared projects, I know for a fact, don't actually affect anything with the build. But I'd assume an import to a non-existing file would make MSBuild angry or something. 🤔

drewnoakes: If there was an error, it was swallowed. The MSBuild docs on `<Import>` don't explain the expected behaviour if the file doesn't exist, so I filed https://github.com/MicrosoftDocs/visualstudio-docs/issues/8316.

ghogen: @rainersigwald @drewnoakes  If there's no error from MSBuild when the imported file can't be found, that would be a MSBuild product bug, I suggest. I get MSB4019 in a simple test case, so a repro should be found that captures this case.

drewnoakes: Does the Label="Shared" attribute make the difference? (Not at a computer right now to check for myself.)

ghogen: No, but we need to update the import element article because it doesn't mention Label="Shared".  I don't know what it means. https://docs.microsoft.com/en-us/visualstudio/msbuild/import-element-msbuild?view=vs-2022

rainersigwald: `Label` doesn't mean much of anything but can be applied to most MSBuild XML elements. It's basically a comment, but accessible through the MSBuild "document model".

PR #8344, opened 2022-07-25
Document potential discrepancies between FUTDC logging and build summary

drewnoakes:
A customer understandably reported this as a point of confusion. I wrote this explanation to share with them, and as a reference for future people who might hit this quirky behaviour.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8344)


drewnoakes:
Looks like GitHub is returning an [HTTP 406](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) on https://github.com/dotnet/roslyn-tools/blob/main/src/GitHubCreateMergePRs/config.xml in our branch checklist. I don't think there is any issue with the change in this PR.

MiYanni (approved):

melytc (approved):

tmeschter (approved):

PR #8345, opened 2022-07-25
Add analyzer to enforce use of `is null` for null checks

drewnoakes:
Seems like we prefer `is null` over `== null` during PR reviews.

Here's an analyzer that does that for us.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8345)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -216,6 +216,10 @@ csharp_using_directive_placement = outside_namespace:error
 csharp_prefer_static_local_function = true:warning
 csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:warning
 
+# CSInNull analyzer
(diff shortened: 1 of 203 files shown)


MiYanni (approved):
I think it is a good change and gets people in the habit of using the pattern matching notation.

melytc (approved):
My understanding is that if we don't expect to have an operator overload for `==`, then using the `is` operator should be fine. I also don't have a strong preference, but this looks nice. 🙂

tmeschter:
Of course, it's always possible for an `==` operator to be added later, potentially changing the meaning of our code without us realizing. I prefer the `is null` approach myself, so I'm all for this. :-D

tmeschter (approved):

drewnoakes:
Let's hold off on merging this until p2, in case we want to revert anything prior to that.

Issue #7868, opened 2022-01-29
Clean up `build.cmd` and build hierarchy


MiYanni:
From my work on OptProf, I ended up removing some build options, as they no longer applied to the build system, but instead the CI system (such as signing type) or were deprecated (such as IBC). These options were left in the `build.cmd` file as I'd need to clean up places where `build.cmd` is called to no longer use those options. Looking into it, it seems that our build hierarchy itself could use some refinement and I see this as an opportunity to improve and simplify our build setup.

#### Currently:
- `build.cmd` calls `msbuild` on `Build.proj`
- `Build.proj` calls the [`<MSBuild/>` task](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-task?view=vs-2022) on `CoreBuild.proj`
- `CoreBuild.proj` calls the [`<MSBuild/>` task](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-task?view=vs-2022) on `ProjectSystem.sln` with a specified list of targets

There are some aspects in `CoreBuild.proj` that have to do with the `roslyn-tools` that we utilize, such as how we make our Insertion PRs. I plan on working toward phasing out our dependence on those tools, so that should all (hopefully) go away. I ended up doing work similar to this in [Azure PowerShell's `build.proj`](https://github.com/Azure/azure-powershell/blob/main/build.proj). Despite it being quite large, it used to be much larger AND call many other projects, targets, etc. We decided to simplify it by putting all of it in once place. I'm not sure I'd do *exactly* that here, but I believe we can keep things fairly lightweight and simple. Our repo doesn't have a lot of projects to build, so the build infra **should** be simple.

Issue #8352, opened 2022-07-26
Cannot use 'dotnet build' to build the repo


MiYanni:
Normally, any SDK-style project that `msbuild` can build, the dotnet CLI using `dotnet build` can also build. This is because under-the-hood, the dotnet CLI is calling `msbuild`. The dotnet CLI also exposes this directly with the [dotnet msbuild](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-msbuild) command. However, these commands are calling a version of msbuild called **msbuild core**. There are some slight differences between msbuild and msbuild core. The difference that concerns us is a set of types in the msbuild SDK that msbuild core doesn't have. These are the 'XamlTypes' via `Microsoft.Build.Framework.XamlTypes`. An [existing issue against MSBuild](https://github.com/dotnet/msbuild/issues/6045) details this difference. Once msbuild core contains XamlTypes, we should verify we can properly build the repo using `dotnet build`.

Additionally, once the repo can build using `dotnet build`, our [build.cmd](https://github.com/dotnet/project-system/blob/main/build.cmd) can be simplified to call `dotnet build` instead of requiring a separate mechanism to locate `msbuild`. The dotnet CLI is always added to the system PATH when the dotnet SDK is installed, so we don't need any additional lookup to use this.

robertmclaws:
Would be great if we could get this fixed

Issue #8351, opened 2022-07-26
Repo fails to build normally/run tests when built with Build Tools


MiYanni:
Related: https://github.com/dotnet/project-system/issues/8319

There are a couple situations with this problem. Instead of installing Visual Studio, you can install [Build Tools for Visual Studio](https://visualstudio.microsoft.com/downloads/?q=build+tools). Technically, our build should succeed with these tools. However, we have a process that deploys the VS Extensions we build to the `Exp` hive. We should properly recognize when you are building with Build Tools (the MSBuild.exe's path will be within BuildTools hierarchy). We also need to fix any issues related to building/testing with Build Tools.

## Situation One
To build with Build Tools, you need to set `/p:DeployExtension=false`. This will cause it to no longer deploy to the `Exp` hive. However, everything will build and VSIXs are created successfully. Without setting this property to false, you will see this error, "Unable to locate the extensions directory."
![Screenshot 2022-07-20 181254](https://user-images.githubusercontent.com/17788297/180895328-c7c29966-fe5e-416f-bf6e-d742a9e9cb0d.png)

## Situation Two
Even after setting `/p:DeployExtension=false`, the full build will fail. The tests that are ran fail with various errors. We should fix these errors and simply not run certain tests that apply to deployment when `/p:DeployExtension=false`.
![Screenshot 2022-07-20 183328](https://user-images.githubusercontent.com/17788297/180895700-f6fe746d-7de2-4b3c-a19c-db55886f3ca1.png)

## Installing Build Tools and configuration
Here are a couple screenshots showing what to install with Build Tools, and how hitting the Launch button opens the Developer Command Prompt.

### Workloads
![Screenshot 2022-07-20 180156](https://user-images.githubusercontent.com/17788297/180897688-a359164f-72dc-4614-8553-39e2c8465ee5.png)

### Launching Developer Command Prompt
![Screenshot 2022-07-20 180600](https://user-images.githubusercontent.com/17788297/180897690-8b96d3a3-f21b-4298-b20e-e377ee4e722b.png)

MiYanni:
As per [this issue](https://github.com/dotnet/project-system/issues/8319), we decided our repo would not support building with Build Tools for Visual Studio. I updated the documentation to reflect this in [this PR](https://github.com/dotnet/project-system/pull/8473).

Issue #8350, opened 2022-07-25
Repo fails to build without /m MSBuild flag


MiYanni:
Looking at the [MSBuild CLI doc](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2022), the `m` flag sets the number of concurrent processes to use. Without a number, it uses the number of processors on the PC. Without the flag itself, it'll be set to 1. It seems that our build will fail if ran completely synchronously.

![image](https://user-images.githubusercontent.com/17788297/180891791-0a01040f-11bb-48b5-a37d-49995f37fbaa.png)

It seems some kind of dependency (build ordering) isn't specified properly, as the VSIX container should exist after the `CreateVsixContainer` target executes. Investigating a binlog of this should be able to determine the problem.

MiYanni:
I finally hit this in CI, even with the `/m` flag. So, it is definitely possible to hit randomly while the build runs on multiple processes.
![image](https://user-images.githubusercontent.com/17788297/184429481-27c2b4f7-4e2f-4231-b820-d12b2ccbebca.png)

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6549829&view=results

MiYanni:
Another failures here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6966047&view=results

PR #8353, opened 2022-07-26
Reduce build usage/activation complexity

MiYanni:
Resolves: https://github.com/dotnet/project-system/issues/7868 \
Related: https://github.com/dotnet/project-system/issues/8319

### Description
This PR does some cleanup to our build activation process. Unlike [this PR](https://github.com/dotnet/project-system/pull/8285) which changed how our build internally functioned, this PR changes how you start the build. It also does some cleanup and reduction in complexity of things that either no longer function or were leftovers from other infrastructure changes. I also performed some testing/investigation on other aspects of building that were *assumed* to be true, but I didn't see evidence of why.

### Changes Outline
- Refactored our *build.cmd*
  - No longer has custom parameters beyond `/?` (pass-through of MSBuild arguments)
  - Removed mechanisms that are no longer hooked up (IBC/Integration Tests/etc.)
  - Replaced *SetVSEnvironment.cmd* with *GetMSBuildPath.ps1*
  - Made MSBuild not reliant on being within the PATH (.exe is called directly)
- Refactored *Build.proj*
  - Removed *CoreBuild.proj* since it served no purpose being a separate file
  - Removed separate target for Restore as it was done as a workaround (no longer seems necessary)
- Allowed all pipelines to call *build.cmd* instead of manually using MSBuild task on *Build.proj*
- Consolidated versioning to only exist in the *version.json* (from Nerdbank.GitVersioning)
  - Previously still needed *Versions.props*
- Moved inline PowerShell-based Targets into separate PowerShell files (.ps1)
  - Allows for commenting and better readability
- Removed CodeCov publishing since CodeCov information is not being produced currently
  - Re-adding CodeCov is covered by: https://github.com/dotnet/project-system/issues/8219
- Removed integration test activation since they won't work currently
  - Details on making these work in the future outlined here: https://github.com/dotnet/project-system/issues/8358
- Cleaned up some information in our primary *README.md*
  - Added information on building the repo
- A smattering of whitespace and casing cleanup

### Details
The *build.cmd* of other repos (like those using Arcade) have a [very small set of logic](https://github.com/dotnet/msbuild/blob/main/build.cmd) that ends up calling another script/process to drive the whole build. In that same mindset, I wanted to *build.cmd* to be fairly equivalent to calling `msbuild eng/Build.proj`. To accomplish that, you need to run MSBuild and provide the basic MSBuild parameters for calling it.

I attempted to use `dotnet build` to do this work instead of having to locate MSBuild, but that has its own issues; which I created some of them listed below. So, we cannot build our repo via `dotnet build`. Based on how our *build.cmd* worked, it looked like our build needed Developer Command Prompt. I tested this and not installing Visual Studio but installing Build Tools for VS. From this, I determined that the Developer Command Prompt is not required for our repo to build and opted to call MSBuild directly. I also found issues using Build Tools for VS and noted those in our README.md.

Many of the other changes were tangentially related as our old build process tied into them, specifically CodeCov and integration tests. Removing the mechanisms to activate these on build means parts of them were not needed. The integration tests themselves still remain in the repo for the time being.

Since I was creating another PowerShell script, I figured it was time to migrate all of the one-line scripts in various .targets files into their own files. This just makes it easier for others to look at, read, and understand. Plus, ability to add comments on certain lines is very useful. Any other cleanup were just things I came across along the way.

### Created Issues
These are issues that I found as part of investigation/changes made for this PR.

- https://github.com/dotnet/project-system/issues/8350
- https://github.com/dotnet/project-system/issues/8351
- https://github.com/dotnet/project-system/issues/8352

### Successful Official Build
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6479990&view=results

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8353)


Directory.Build.props
```diff
@@ -47,13 +49,19 @@
...
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <!-- TODO: Might be related to: https://github.com/dotnet/project-system/issues/5619 -->
     <EnableSourceLink Condition="'$(CIBuild)' == 'true'">true</EnableSourceLink>
     <NoWarn>$(NoWarn);NU5125</NoWarn>
+
+    <!-- Test project properties -->
+    <IsUnitTestProject Condition="'$(IsUnitTestProject)' == ''">false</IsUnitTestProject>
+    <IsIntegrationTestProject Condition="'$(IsIntegrationTestProject)' == ''">false</IsIntegrationTestProject>
+    <IsTestProject Condition="'$(IsTestProject)' == ''">false</IsTestProject>
```
MiYanni, line 64: These did not need to be above the `Import` statements.

README.md
```diff
@@ -1,49 +1,79 @@
...
+| Release             | Unit Tests (Debug)                      | Unit Tests (Release)                      | Localization
+|---------------------|:---------------------------------------:|:-----------------------------------------:|:------------:
+| [16.11][1611Branch] | [![Build Status][1611Debug]][1611Build] | [![Build Status][1611Release]][1611Build] | [![Build Status][1611Spanish]][1611Build]
+| [17.0][170Branch]   | [![Build Status][170Debug]][170Build]   | [![Build Status][170Release]][170Build]   | [![Build Status][170Spanish]][170Build]
+| [17.1][171Branch]   | [![Build Status][171Debug]][171Build]   | [![Build Status][171Release]][171Build]   | [![Build Status][171Spanish]][171Build]
+| [17.2][172Branch]   | [![Build Status][172Debug]][172Build]   | [![Build Status][172Release]][172Build]   | [![Build Status][172Spanish]][172Build]
+| [17.3][173Branch]   | [![Build Status][173Debug]][173Build]   | [![Build Status][173Release]][173Build]   | [![Build Status][173Spanish]][173Build]
+| [main][MainBranch]  | [![Build Status][MainDebug]][MainBuild] | [![Build Status][MainRelease]][MainBuild] | [![Build Status][MainSpanish]][MainBuild]
```
MiYanni, line 10: Align them tables! 🦾

build.cmd
```diff
@@ -1,89 +1,57 @@
...
-if /I "%1" == "/?"                                                                                  call :Usage && exit /b 1
-if /I "%1" == "/build"                set "OptBuild=true"  && set "OptRebuild=false"                && shift && goto :ParseArguments
-if /I "%1" == "/no-build"             set "OptBuild=false" && set "OptRebuild=false"                && shift && goto :ParseArguments
-if /I "%1" == "/rebuild"              set "OptBuild=false" && set "OptRebuild=true"                 && shift && goto :ParseArguments
-if /I "%1" == "/test"                 set "OptTest=true"                                            && shift && goto :ParseArguments
-if /I "%1" == "/no-test"              set "OptTest=false"                                           && shift && goto :ParseArguments
-if /I "%1" == "/integration"          set "OptIntegrationTest=true"                                 && shift && goto :ParseArguments
-if /I "%1" == "/no-integration"       set "OptIntegrationTest=false"                                && shift && goto :ParseArguments
-if /I "%1" == "/deploy"               set "OptDeploy=true"                                          && shift && goto :ParseArguments
```
MiYanni, line 31: I do not believe this worked after the roslyn-tools removal. For now, `/p:DeployExtension=false` will do this. That is noted in the README.md for those using Build Tools for VS.

build.cmd
```diff
@@ -1,89 +1,57 @@
...
-if /I "%1" == "/no-deploy"            set "OptDeploy=false"                                         && shift && goto :ParseArguments
-if /I "%1" == "/diagnostic"           set "OptDiagnostic=true"                                      && shift && goto :ParseArguments
-if /I "%1" == "/no-diagnostic"        set "OptDiagnostic=false"                                     && shift && goto :ParseArguments
-if /I "%1" == "/sign"                 set "OptSign=true"                                            && shift && goto :ParseArguments
-if /I "%1" == "/no-sign"              set "OptSign=false"                                           && shift && goto :ParseArguments
-if /I "%1" == "/ci"                   set "OptCI=true"     && set "OptNodeReuse=false"              && shift && goto :ParseArguments
-if /I "%1" == "/no-ci"                set "OptCI=false"    && set "OptNodeReuse=true"               && shift && goto :ParseArguments
-if /I "%1" == "/ibc"                  set "OptIbc=true"                                             && shift && goto :ParseArguments
```
MiYanni, line 39: This hasn't done anything for a while. It was OptProf data for AppDesigner/Editors using very old optimization information.

build.cmd
```diff
@@ -1,89 +1,57 @@
...
-if /I "%1" == "/no-ibc"               set "OptIbc=false"                                            && shift && goto :ParseArguments
-if /I "%1" == "/clearnugetcache"      set "OptClearNuGetCache=true"                                 && shift && goto :ParseArguments
```
MiYanni, line 41: This was removed since it wasn't directly related to our build. It is just deleting the contents of `%USERPROFILE%\.nuget\packages\`. If we want that, I can add it back.

eng/imports/RepoLayout.props
```diff
@@ -8,7 +8,6 @@
     <ArtifactsConfigurationDir>$(ArtifactsDir)$(Configuration)\</ArtifactsConfigurationDir>
     <ArtifactsBinDir>$(ArtifactsConfigurationDir)bin\</ArtifactsBinDir>
     <ArtifactsObjDir>$(ArtifactsConfigurationDir)obj\</ArtifactsObjDir>
-    <ArtifactsLogDir>$(ArtifactsConfigurationDir)log\</ArtifactsLogDir>
```
MiYanni, line 11: Other than the directory being created, this property was not used by our build.

build.cmd
```diff
@@ -1,89 +1,57 @@
...
-if /I "%1" == "/no-test"              set "OptTest=false"                                           && shift && goto :ParseArguments
-if /I "%1" == "/integration"          set "OptIntegrationTest=true"                                 && shift && goto :ParseArguments
-if /I "%1" == "/no-integration"       set "OptIntegrationTest=false"                                && shift && goto :ParseArguments
-if /I "%1" == "/deploy"               set "OptDeploy=true"                                          && shift && goto :ParseArguments
-if /I "%1" == "/no-deploy"            set "OptDeploy=false"                                         && shift && goto :ParseArguments
-if /I "%1" == "/diagnostic"           set "OptDiagnostic=true"                                      && shift && goto :ParseArguments
-if /I "%1" == "/no-diagnostic"        set "OptDiagnostic=false"                                     && shift && goto :ParseArguments
-if /I "%1" == "/sign"                 set "OptSign=true"                                            && shift && goto :ParseArguments
```
MiYanni, line 35: This hasn't worked since the OptProf implementation. Signing is directly controlled by having the MicroBuild Signing plugin installed or not.

eng/pipelines/official.yml
```diff
@@ -36,20 +36,37 @@ parameters:
   default: false
 
 ###################################################################################################################################################################
-# INSTALLATION
+# SBOM GENERATION
+# This occurs above plugin installation so that the Src build does not perform signing.
```
MiYanni, line 40: This comment here explains why this content was moved above installing plugins.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.CSharp.DesignTime.targets
```diff
@@ -42,25 +43,26 @@
     <PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)ApplicationPropertyPage.CSharp.xaml">
       <Context>Project</Context>
     </PropertyPageSchema>
-    
+
     <PropertyPageSchema Include="$(ManagedXamlResourcesDirectory)BuildPropertyPage.CSharp.xaml">
       <Context>Project</Context>
     </PropertyPageSchema>
 
+    <!-- See: https://github.com/dotnet/project-system/pull/8235 -->
```
MiYanni, line 51: I added this since I saw the commented code here and wondered why it was in the repo. We should always try to add information on why commented code exists in the code itself.

tests/Directory.Build.targets
```diff
@@ -2,34 +2,7 @@
...
-
   <Import Project="..\eng\imports\UnitTests.targets" Condition="'$(IsUnitTestProject)' == 'true'" />
   <Import Project="..\eng\imports\IntegrationTests.targets" Condition="'$(IsIntegrationTestProject)' == 'true'" />
 
-  <PropertyGroup>
-    <_CoverageReport>$(ArtifactsTestResultsDir)$(AssemblyName).dll.coverage</_CoverageReport>
-  </PropertyGroup>
-
-  <!-- NOTICE: This relies on the unit test projects producing .coverage files. Currently, unit test projects do not produce .coverage files. -->
-  <Target Name="UploadCodecov" AfterTargets="Test" Condition="'$(UsingCodeCoverage)' == 'true' AND '$(Configuration)' == 'Debug' AND '$(IsUnitTestProject)' == 'true' AND Exists('$(_CoverageReport)')">
```
MiYanni, line 14: As the comment said for this target, this target was running in our unit test pipeline, but it wasn't publishing any code coverage information.

eng/Build.proj
```diff
@@ -1,37 +1,55 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
-<Project DefaultTargets="ClearNuGetCache;PrepareMachine;CoreBuild">
```
davkean, line 2: ClearNuGetCache was deliberate to avoid taking a dependency on packages that had been deleted, or when packages sources changed.

MiYanni: That target only ran if you provided the `/clearnugetcache` as a parameter to *build.cmd* or `/p:ClearNuGetCache=false` to *Build.proj*. It wasn't being used in CI and there wasn't a *clear* need for it as part of our build process. As I mentioned [here](https://github.com/dotnet/project-system/pull/8353/files#r931696942), it was simply deleting all the packages our of your Windows user cache. If this repo had a local cache, such as a `packages` folder that is created at the repo root on restore, I could see it being nice to have a flag to clear that and maybe even tying that into Rebuild.

As mentioned, if there is a scenario where it is necessary, I haven't ran into it yet, but I can add back in an equivalent. Generally, clearing out your *entire* cache shouldn't be necessary; it is more in exceptional situations for resolving specific issues. For the scenario you mentioned, we don't take on many new dependencies at this point, and I did a full sweep of package source feeds recently, so I don't feel a strong need for this flag/target.

MiYanni: Here's a good example of a current PR where, maybe your cache on your machine had the right packages. However, CI run will always fail if the package dependencies are not defined properly. Drew can test it locally by clearing his cache, but this should not be seen as a normal scenario. https://github.com/dotnet/project-system/pull/8346

drewnoakes (approved):
Thanks for the comments. Seems good to me. Let's wait for p2 to merge it though.

MiYanni:
p1 branch was created this morning, so `main` is for p2 at this point. Merging.

PR #8355, opened 2022-07-26
[17.3] Intercept StartupObject property for VB WinForms projects

melytc:
The `StartupObject` property has a different behavior compared to the C# experience. This setting doesn't get serialized exclusively to the project file; in certain scenarios, it also needs to be serialized into the myapp file.

**Scenarios**
- If the `ApplicationFramework` is disabled, this settings gets serialized into the project file.

We have been showing a correct set of values in the dropdown. The expected startup objects for VB WinForms projects should be:

For C#, this is being done by calling `FindEntryPoints`. We will use it for VB too; the flag `searchForEntryPointsInFormsOnly` in the rule file will help us differentiate the expected values. 

Note: this is for **17.3 Preview 6.**

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8355)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
@@ -37,14 +37,14 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
                 if (unevaluatedPropertyValue.Contains("Form")) // Is there a better way to identify a form?
                 {
                     // If the user selects a Form, the value should be serialized to the myapp file.
-                    await _myAppXmlFileAccessor.SetStartupObjectAsync(unevaluatedPropertyValue);
+                    await _myAppXmlFileAccessor.SetMainFormAsync(unevaluatedPropertyValue);
                     await defaultProperties.DeletePropertyAsync(StartupObjectProperty);
                     return null;
                 }
 
                 // If the ApplicationFramework is enabled, the value Sub Main should always be serialized to the project file.
                 await defaultProperties.SetPropertyValueAsync(StartupObjectProperty, "Sub Main");
-                await _myAppXmlFileAccessor.SetStartupObjectAsync(string.Empty);
+                await _myAppXmlFileAccessor.SetMainFormAsync(string.Empty);
                 return null;
             }
         }
@@ -66,7 +66,7 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
         string valueInProjectFile = await base.OnGetUnevaluatedPropertyValueAsync(propertyName, evaluatedPropertyValue, defaultProperties);
 
         if (string.IsNullOrEmpty(valueInProjectFile))
-            return await _myAppXmlFileAccessor.GetStartupObjectAsync() ?? string.Empty;
+            return await _myAppXmlFileAccessor.GetMainFormAsync() ?? string.Empty;
 
         return valueInProjectFile;
     }
@@ -82,7 +82,7 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
         string valueInProjectFile = await base.OnGetUnevaluatedPropertyValueAsync(propertyName, unevaluatedPropertyValue, defaultProperties);
 
         if (string.IsNullOrEmpty(valueInProjectFile))
-            return await _myAppXmlFileAccessor.GetStartupObjectAsync() ?? string.Empty;
+            return await _myAppXmlFileAccessor.GetMainFormAsync() ?? string.Empty;
 
         return valueInProjectFile;
     }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
@@ -5,6 +5,7 @@ using Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties;
 
+// The AppliesTo metadata has no effect given the limitations described in https://github.com/dotnet/project-system/issues/8170.
 [ExportInterceptingPropertyValueProvider(StartupObjectProperty, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
 [AppliesTo(ProjectCapability.VisualBasic)]
 internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBase
@@ -34,25 +35,14 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
 
             if (Equals(applicationFrameworkValue, EnabledValue))
             {
-                if (unevaluatedPropertyValue.Contains("Form")) // Is there a better way to identify a form?
-                {
-                    // If the user selects a Form, the value should be serialized to the myapp file.
-                    await _myAppXmlFileAccessor.SetMainFormAsync(unevaluatedPropertyValue);
-                    await defaultProperties.DeletePropertyAsync(StartupObjectProperty);
-                    return null;
-                }
-
-                // If the ApplicationFramework is enabled, the value Sub Main should always be serialized to the project file.
-                await defaultProperties.SetPropertyValueAsync(StartupObjectProperty, "Sub Main");
-                await _myAppXmlFileAccessor.SetMainFormAsync(string.Empty);
-                return null;
+                // TO-DO: Validate scenario where the selected value is a Form.
+                await _myAppXmlFileAccessor.SetMainFormAsync(unevaluatedPropertyValue);
             }
         }
 
         // Else, if it's other than a Windows Forms project, save the StartupObject property in the project file as usual.
         // Or if the ApplicationFramework property is disabled, save the StartupObject property in the project file as usual.
-        await defaultProperties.SetPropertyValueAsync(propertyName, unevaluatedPropertyValue);
-        return null;
+        return unevaluatedPropertyValue;
     }
 
     public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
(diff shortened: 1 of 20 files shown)


melytc:
Took some suggestions from https://github.com/dotnet/project-system/pull/7936

ocallesp (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+    public StartupObjectValueProvider(IMyAppFileAccessor myAppXmlFileAccessor)
+    {
+        _myAppXmlFileAccessor = myAppXmlFileAccessor;
+    }
+
+    public override async Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+    {
+        string isWindowsFormsProject = await defaultProperties.GetEvaluatedPropertyValueAsync(UseWinFormsProperty);
+
+        if (isWindowsFormsProject == "true")
```
drewnoakes, line 31: What about other capitalisations?

Perhaps:

```suggestion
        if (bool.TryParse(isWindowsFormsProject, out bool b) && b)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            string applicationFrameworkValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkProperty);
+
+            if (applicationFrameworkValue == EnabledValue)
```
drewnoakes, line 35: Consider replacing `==` with an explicit comparer via `StringComparers`.

Ditto the two other comparisons with `DisabledValue` below.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    public override async Task<string> OnGetEvaluatedPropertyValueAsync(string propertyName, string evaluatedPropertyValue, IProjectProperties defaultProperties)
+    {
+        // StartupObject can come from the project file or the myapp file.
+        string applicationFrameworkValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkProperty);
+        
+        if (applicationFrameworkValue == DisabledValue)
+            return await base.OnGetUnevaluatedPropertyValueAsync(propertyName, evaluatedPropertyValue, defaultProperties);
+
+        string valueInProjectFile = await base.OnGetUnevaluatedPropertyValueAsync(propertyName, evaluatedPropertyValue, defaultProperties);
```
drewnoakes, line 66: Would this be easier to read?

```suggestion
        string valueInProjectFile = await base.OnGetUnevaluatedPropertyValueAsync(propertyName, evaluatedPropertyValue, defaultProperties);

        if (applicationFrameworkValue == DisabledValue)
            return valueInProjectFile;
```

That diff's a bit of a mess. Basically, the local variable `valueInProjectFile` can be moved up, and its value re-used in the if block to reduce duplication. I found myself checking that both of the old calls were the same. Also "return valueInProjectFile" reads quite naturally.

Ditto below in the next method.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.VisualBasic.xaml
```diff
@@ -12,12 +12,40 @@
     <Category Name="WPF"
               DisplayName="WPF"
               Description="Settings specific to WPF applications." />
+
+    <Category Name="General"
+              DisplayName="General"
+              Description="General settings for the application." />
```
drewnoakes, line 18: Should the General category go first?

drewnoakes (approved):
Thanks for documenting your code so thoroughly. It made reviewing this much easier.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        string isWindowsFormsProject = await defaultProperties.GetEvaluatedPropertyValueAsync(UseWinFormsProperty);
+
+        if (isWindowsFormsProject == "true")
+        {
+            string applicationFrameworkValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkProperty);
+
+            if (applicationFrameworkValue == EnabledValue)
+            {
+                if (unevaluatedPropertyValue.Contains("Form")) // Is there a better way to identify a form?
```
KlausLoeffelmann, line 37: I am a bit uneasy with this, but I am also not familiar with the project system specifics in the context.

What does `unevaluatedPropertyValue` contain here? If it is the Form's name, then this isn't going to work, since there is no guarantee that the string "Form" is part of the name. If it is the type, is this the fully qualified name?

If it's the name, do we have the means to do a type resolution based on the name without hopping to the WinForms Designer's server process to resolve the type?

KlausLoeffelmann (commented):
I have a question regarding identifying the type of the start object, which might be problematic. But then, I am also not very familiar with the code base! 😄

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/StartupObjectsEnumProvider.cs
```diff
@@ -47,18 +52,15 @@ internal class StartupObjectsEnumGenerator : IDynamicEnumValuesGenerator
...
         private readonly UnconfiguredProject _unconfiguredProject;
         private readonly bool _includeEmptyValue;
-
-        /// <summary>
-        /// When we implement WinForms support, we need to set this for VB WinForms projects
-        /// </summary>
-        private static bool SearchForEntryPointsInFormsOnly => false;
+        private readonly bool _searchForEntryPointsInFormsOnly;
 
         [ImportingConstructor]
```
tmeschter, line 57: Not directly related to your changes but I don't think this `[ImportingConstuctor]` is needed as this type isn't MEF-exported.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.VisualBasic.xaml
tmeschter: You shouldn't need this if the base page defines a "General" category (which it should).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ApplicationPropertyPage.VisualBasic.xaml
```diff
@@ -12,12 +12,40 @@
...
   <!-- TODO: Add a hyperlink/button to open the app.manifest. Previously, we had a View Windows Settings button. -->
 
   <!-- TODO: Add a hyperlink/button to open the ApplicationEvents.vb. Previously, we had a View Application Events button. -->
-  
+
+  <!-- The StartupObject property has a different set of appropriate values for VB and for C#.
+       For VB projects, we expect to see every Form in the assembly directly or indirectly inherited from Form.
+       We specify this by setting the searchForEntryPointsInFormsOnly to true.
+       We also want to ensure that the property always has a value. -->
+  <DynamicEnumProperty Name="StartupObject"
```
tmeschter, line 29: Note that since this page extends an existing page (rather than replacing it) this definition of the StartupObject property will completely replace any definition in the base page.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
KlausLoeffelmann: It depends. In Framework, the caption of the Startup _object_ changes to Startup _form_ if the AppFramework is enabled. And then only Forms are shown, rest is filtered out. 

My concern is that there was additional filtering code involved when getting back potential startup objects with the AppFramework enabled, and that those filters are missing here.

Here is a typical scenario in Framework without AppFramework enabled:

image

And here, some project _with_ AppFramework enabled:

image

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
melytc: Good point, added on last commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -0,0 +1,89 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.ProjectSystem.VS.WindowsForms;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties;
+
+[ExportInterceptingPropertyValueProvider(StartupObjectProperty, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+[AppliesTo(ProjectCapability.VisualBasic)]
```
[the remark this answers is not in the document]
MiYanni: Maybe put that as a comment above it in the code?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
melytc: After discussing it offline, we've decided to remove the form validation and save the value in both the project file and the myapp file; I will revisit this soon after this.

MiYanni:
The in-progress optional checks can be ignored.

MiYanni:
@drewnoakes Please don't try to re-run it. The required checks have already passed. The PR is ready to merge.

melytc:
Closing in favor of https://github.com/dotnet/project-system/pull/8370.

PR #8359, opened 2022-07-28
Implement a deterministic string hash

### LINKED ISSUE #7801: Replace use of string.GetHashCode
We currently use `string.GetHashCode` as part of a hash persisted by the fast up-to-date check.

https://github.com/dotnet/project-system/blob/aa645eb69baee28b9c3083bb8aba018040020798/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.ItemHashing.cs#L24-L33

If and when VS moves to .NET 5+, we will not be able to rely on the output of this method being stable across process runs. We will need to replace this with a stable hash.

As the current hash value is persisted in a file in the `.vs` folder, changing this algorithm may regress performance. For people using versions from both before and after such a change on the same machine would see regular performance regressions. For that reason, we should aim to recreate the .NET Framework string hash function in our code, and use that.

ocallesp:
fixes #7801 

Problem to solve:
Docs for `string.GetHashCode()` says that hash codes for identical strings can differ across .NET implementations, across .NET versions, and across .NET platforms (such as 32-bit and 64-bit) for a single version of .NET. In some cases, they can even differ by application domain.

The fix:
We want to use calculate the same hash value even if there are changes in .NET version. To accomplish this
we have to always use the same hash algorithm.
This PR implements a hash algorithm for string to replace `string.GetHashCode()`.

The implementation in here was taken from `string.GetHashCode()` with some minor changes. 
https://github.com/microsoft/referencesource/blob/master/mscorlib/system/string.cs

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8359)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
```diff
@@ -0,0 +1,59 @@
...
+        public static int GetDeterministicHashCode(this string str)
+        {
+            unsafe
+            {
+                fixed (char* src = str)
+                {
+                    Contract.Assert(src[str.Length] == '\0', "src[this.Length] == '\\0'");
+                    Contract.Assert(((int)src) % 4 == 0, "Managed string should start at 4 bytes boundary");
+
+#if WIN32
```
drewnoakes, line 18: You can remove the `WIN32` blocks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Diagnostics.Contracts;
+
+namespace Microsoft.VisualStudio.Text
```
drewnoakes, line 5: For new files, consider using file-scoped namespaces.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    internal static class StableStringHash
+    {
+        public static int GetDeterministicHashCode(this string str)
```
drewnoakes, line 9: I think "Stable" would be a better description than "Deterministic".

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Diagnostics.Contracts;
+
+namespace Microsoft.VisualStudio.Text
+{
+    internal static class StableStringHash
```
drewnoakes, line 7: Can you add a comment here explaining where you pulled this implementation from? I know it's in the PR description, but having it here in the code would be helpful.

Also explain that it should not be changed, as doing so would break asset compatibility.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
drewnoakes, line 9: I don't think we would want to use this hash function outside of the FUTDC. Perhaps move the method to `BuildUpToDateCheck.ItemHashing.cs`.

I'd also not make it an extension method, as otherwise it'll appear in IntelliSense in a bunch of places we don't want it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public static int GetDeterministicHashCode(this string str)
+        {
+            unsafe
+            {
+                fixed (char* src = str)
+                {
+                    Contract.Assert(src[str.Length] == '\0', "src[this.Length] == '\\0'");
+                    Contract.Assert(((int)src) % 4 == 0, "Managed string should start at 4 bytes boundary");
```
drewnoakes, line 16: The code contracts can be removed as we don't use them.

drewnoakes (commented):
Please add some unit tests. Make sure this function produces identical values to .NET Framework x64, so that we maintain asset compatibility. This Sharplab might be useful:

https://sharplab.io/#v2:EYLgHgbALANALiATgVwHYB8ACAmAjAWACgjNcBOACgCIAJAUwBsGB7GAAgHVnEGATAQioA6AOJ04NAIYBnABYBhZrzoUAlKoDcQA

KirillOsenkov:
In case it's useful, I've looked into string hashing before:
https://github.com/KirillOsenkov/MSBuildStructuredLog/wiki/String-Hashing

For MSBuild BinaryLogger we decided to go with this one, which is very good and super simple:
https://github.com/dotnet/msbuild/blob/e1de0c476fac967bd1f1d4c17e5c8a70513cfdee/src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs#L1218-L1246

Microsoft.NET.StringTools.dll (part of MSBuild) chose djb2:
https://github.com/dotnet/msbuild/blob/e1de0c476fac967bd1f1d4c17e5c8a70513cfdee/src/StringTools/InternableString.cs#L310-L397

drewnoakes:
@KirillOsenkov one constraint here is that we use the output of the .NET Framework x64 hash in the `.vs` folder today, and we want to preserve that function in code, as if we move to .NET 5+ in future the hash will vary between process lifetimes. By using the same logic, users moving between versions before/after this change will still be able to keep their caches.

KirillOsenkov:
Makes sense!

drewnoakes:
Your [wiki article](https://github.com/KirillOsenkov/MSBuildStructuredLog/wiki/String-Hashing) is awesome btw.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
ocallesp: The code was moved to BuildUpToDateCheck.ItemHashing.cs.

done.

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
ocallesp: removed

done

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
ocallesp: comment added.

done

src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/StringHash.cs
ocallesp: removed.

done

ocallesp:
> .NET Framework x64

If we want the same output as the .NET Framework x64, then we should use the same algorithm.

In this implementation it gives a different value.

ocallesp:
Yes. This implementation returns the same hash 562640209  for "Hello, World!"  .Net Framework x64

adamint (changes requested):
Looks good, but yes please add some unit tests

drewnoakes:
Some unit tests with known hash values values from netfx would be good to see. The test will ensure we don't accidentally change values in future and break the cache.

Include the empty string, and two strings that differ only by case.

ocallesp:
Unit-tests added in last commit.
They are using precomputed hash values for net framework x64

ocallesp:
> Some unit tests with known hash values values from netfx

The unit-tests are already using known (precomputed) hash values. If this is not the case, can you explain more about the proposed test ?

drewnoakes:
I think you pushed them as I wrote my comment. They look good. The extra cases I'd consider adding are the empty string and two strings that differ only by case.

drewnoakes (approved):

adamint (approved):

PR #8361, opened 2022-07-28
Revert language service changes for 17.4p1

drewnoakes:
Upon attempting to insert these changes into VS, we discovered an issue in Roslyn that led to a hang. Therefore we are holding this change back until 17.4p2 rather than p1.

Roslyn has fixed that issue in https://github.com/dotnet/roslyn/pull/62951, but we don't have enough time to coordinate our insertion and testing for p1.

We have other changes we want for p1, so we are reverting this work for now, and will re-instate this once the branch is open for p2.

This PR reverts both #8321 and #8340. When re-instating this change, we should revert this reversion PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8361)


MiYanni (approved):

melytc (approved):

MiYanni:
@drewnoakes Here's the build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6485163&view=results

PR #8368, opened 2022-07-29
Update unit-tests.yml

tmeschter:
Update unit-tests.yml so it only runs on builds/PRs targeting dev17.3.x.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8368)


adamint:
Would it be useful to instead to dev17.* ?

MiYanni:
This is not a good idea to do as this change gets moved forward into `main` based on the bot we use from roslyn-tools.

PR #8370, opened 2022-07-29
[17.4] Intercept StartupObject property for VB WinForms projects

melytc:
Note: this PR has the same commits as #8355, but targets dev17.4-preview1.

The `StartupObject` property has a different behavior compared to the C# experience. This setting doesn't get serialized exclusively to the project file; in certain scenarios, it also needs to be serialized into the myapp file.

**Scenarios**
- If the `ApplicationFramework` is disabled, these settings get serialized into the project file.

We have been showing a correct set of values in the dropdown. The expected startup objects for VB WinForms projects should be:

For C#, this is being done by calling `FindEntryPoints`. We will use it for VB too; the flag `searchForEntryPointsInFormsOnly` in the rule file will help us differentiate the expected values. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8370)


drewnoakes (approved):

jacdavis (approved):
:shipit:

melytc:
This PR would benefit from having this API updated: https://github.com/dotnet/project-system/pull/8044. I'll hold merge until we have it ready.

ocallesp:
@melytc
It seems that there has not been changes in #8044  since April. 

Is #8044 blocking this pr ? 
Can we merge this pr and then later make the corresponding change after #8044 ?

melytc:
Closing as I will be bringing this to main directly.

Issue #8363, opened 2022-07-29
VS Insertion PR descriptions are missing


MiYanni:
After the changes made related to versioning in the [roslyn-tools removal PR](https://github.com/dotnet/project-system/pull/8285), it broke our description details for the VS insertion PRs. This is one of the remaining roslyn-tools that we still use. However, it isn't part of the repo itself. This needs to be fixed in the short term, prior to replacing it within [this issue](https://github.com/dotnet/project-system/issues/7911).

#### Error Message
> Unable to find details for previous build ([BUILDNUMBER])

MiYanni:
Bad news. The design of Roslyn Insertion Tool's build number is only 2 digits. This might not be able to be fixed until later, but still investigating.
![image](https://user-images.githubusercontent.com/17788297/181846459-4813a373-68c2-469c-aad2-1204ca68517e.png)

Issue #7911, opened 2022-02-11
Remove use of Roslyn Insertion Tool


MiYanni:
In an effort to remove our dependence on [roslyn-tools](https://github.com/dotnet/roslyn-tools), it would be smart if we no longer used RIT (Roslyn Insertion Tool). This tool is activated by a PowerShell script called `OneOffInsertion.ps1` which we run in our release pipeline: https://devdiv.visualstudio.com/DevDiv/_release?view=all&_a=releases&definitionId=1242

The entire purpose of that pipeline is to install the package for RIT and run that PowerShell script. We'd need to determine exactly what that tool is doing and how to do that functionality on our own. The rationale is that roslyn-tools is not actively developed and many components are becoming more and more outdated as time goes on. We want full control of our infrastructure to keep things simple and up-to-date. This allows us to adjust to policy changes from the Visual Studio org.

Some additional information about package publishing done by the tool can be seen in what I put into our `official.yml`:
https://github.com/dotnet/project-system/blob/c8137374c0f7e6458397627b530ddb14099cb6b9/build/ci/official.yml#L266-L276

This process should be replaced with [Automated VS Insertions](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/629/Automated-VS-Insertion).

PR #8373, opened 2022-07-30
Fix insertion pipeline issue and VS insertion description

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/8363

The first fix that actually fixes the build was a bug I introduced in https://github.com/dotnet/project-system/pull/8353. I changed the CreateDependentAssemblyVersions to use `AssemblyVersion`, which is 4 digits, when before it was using `ProjectSystemVersion` which was 3 digits. That value is gone now, but the process to create that file was tacking on an extra `.0`, so it became 5 digits which broke the VS insertion release pipeline. I removed the extra `.0` here.

The VS insertion PR descriptions was a bit more involved. Had to debug how the Roslyn Insertion Tool (which we currently use, until I implement https://github.com/dotnet/project-system/issues/7911) to find out that it only allows 2 digit version numbers. Azure Pipelines uses 2 digits, the first being the date and the second being the number of pipelines ran on that particular day. So, pipeline runs can never collide. However, the weird part about all this is the version number is pulled **FROM THE URI OF THE VSMAN FILE**: https://github.com/dotnet/roslyn-tools/blob/07fbfa5a646338471b499a1e93f70170378437f6/src/RoslynInsertionTool/RoslynInsertionTool/Components.cs#L29-L41

So, even if I change the URI to use our new numbers (17.4.X), it won't work. I need to use the 2 digit numbers for the build for this to work. In the test build I ran, it still won't show the right result since the *previous build merged* doesn't use 2 digit numbers. However, after this build, subsequent builds will show the description correctly.

Test insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/414366

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8373)


version.json
```diff
@@ -1,14 +1,4 @@
...
-  "assemblyVersion": "17.4",
-  "buildNumberOffset": 0,
-  "cloudBuild": {
-    "buildNumber": {
-      "enabled": true,
-      "includeCommitId": {
-        "when": "never"
-      }
-    }
-  }
```
MiYanni, line 13: So, not using the cloudBuild mechanism outline here anymore: https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/cloudbuild.md#automatically-match-cloud-build-numbers-to-to-your-git-version
Also, `assemblyVersion` wasn't needed since it isn't different than `version`. And `buildNumberOffset` is deprecated, and wasn't needed anyway.

eng/pipelines/official.yml
```diff
@@ -228,10 +228,15 @@ steps:
...
 
+# This creates the drop location for the .vsman files for VS insertion.
 - task: MicroBuildUploadVstsDropFolder@1
   displayName: Publish VS Setup
   inputs:
     DropFolder: 'artifacts\$(BuildConfiguration)\VSSetup\Insertion'
+    # TODO: Change this to use $(GitBuildVersion) when working on: https://github.com/dotnet/project-system/issues/7911
+    # https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/cloudbuild.md#automatically-match-cloud-build-numbers-to-to-your-git-version
+    # 'Products/[DevDiv]/[dotnet/project-system]/[main]/[x.x]'
+    DropName: 'Products/$(System.TeamProject)/$(Build.Repository.Name)/$(Build.SourceBranchName)/$(Build.BuildNumber)'
```
MiYanni, line 239: This is how we'd modify the URI for the .vsman publishing. However, it doesn't fix the issue. This is the default value, but I leave it here with comments so that we know we can change it or investigate it if needed when working on https://github.com/dotnet/project-system/issues/7911.

PR #8379, opened 2022-08-02
Removing the unsafe code from the stable hash algorithm.

ocallesp:
Current implementation of `GetStableHashCode()` is broken for Linux or Mac (depends on UTF-8 strings) while the CLR may have chosen to stick with Utf16.

Removing the unsafe codes removes the dependency upon the internal representation of the string in memory.

This is to make it work correctly on any platform.

-----------------------------
Benchmark

|                   Method |     Mean |     Error |    StdDev |
|------------------------- |---------:|----------:|----------:|
| GetStableHashCodeUnsafe | 7.068 ns | 0.1463 ns | 0.1796 ns |
| GetStableHashCodeSafe | 9.247 ns | 0.1851 ns | 0.1641 ns |
| CalculateStringGetHashCode | 6.655 ns | 0.1581 ns | 0.2110 ns |
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8379)


MiYanni (approved):

drewnoakes (approved):
Thanks for including benchmark results too.

davkean:
System.String is UTF16 on every OS and this code doesn't run on Mac/Linux, so I don't follow the Mac/Linux comment?

PR #8384, opened 2022-08-05
Handle null pointer

ocallesp:
Fixes [AB#1587569](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1587569)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8384)


drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Acquisition/MissingSetupComponentRegistrationService.cs
```diff
@@ -277,16 +277,20 @@ private async Task DisplayMissingComponentsPromptAsync()
             }
 
             var setupCompositionService = await _vsSetupCompositionService.GetValueAsync();
```
drewnoakes, line 279: If this can return null, it should be annotated as such.

PR #8391, opened 2022-08-10
add "My" namespace to generated settings type

adamint:
To fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1589004?src=WorkItemMention&src-action=artifact_link, we need to add "My" namespace to the MySettings class. 
image

This works for .net framework 4.8 VB, as well as .net core VB and .net core C#, but before approving it, another pair of eyes on the generated file would be nice

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8391)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsSingleFileGeneratorBase.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsSingleFileGeneratorBase.vb
@@ -144,7 +144,10 @@ Namespace Microsoft.VisualStudio.Editors.SettingsDesigner
                 '   appropriate code.
                 '
                 Dim projectRootNamespace As String = String.Empty
-
+                If isVB Then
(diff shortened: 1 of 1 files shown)


melytc:
Was this always broken? I'm trying to think why this is breaking now if there haven't been significant changes in the last months/years to this code generator.

adamint:
this may have been related to my most recent designer changes

tmeschter (approved):
Make sure you test:

1. C# projects using both the old and new project systems.
2. VB projects using both the old and new project systems.
3. VB projects where the root and/or default namespaces have been changed from default values.

MiYanni (approved):
+1 to Tom's comment.

adamint:
works!
caveat that System.Configuration.ConfigurationManager is a required package for .net core projects

PR #8395, opened 2022-08-10
Use nuget.config instead of RestoreSources

MiYanni:
### Description
This is part of an effort for [Central Feed Services (CFS)](https://eng.ms/docs/cloud-ai-platform/developer-services/one-engineering-system-1es/1es-docs/secure-supply-chain/project-artemis/central-feed-services-cfs). The tooling they make for validating feeds being used in a repo rely on a `nuget.config` file at the root. I've moved our feeds from being declared in `RestoreSources` in Packages.targets to a nuget.config at the root. Currently, public feeds in Azure DevOps do not support upstream sources, so this won't be consolidated to a single feed yet.

### Official Build
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6542168&view=results

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8395)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,6 +12,7 @@
 # Set these files to XML (coloring) in PRs
 *.vsmanproj linguist-language=XML
 *.swixproj linguist-language=XML
+*.config linguist-language=XML
 
 ###############################################################################
 # Set default behavior for command prompt diff.
--- a/eng/imports/Packages.targets
+++ b/eng/imports/Packages.targets
@@ -1,40 +1,6 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information. -->
 <Project>
 
-  <PropertyGroup>
-    <RestoreSources>
-      $(RestoreSources);
-      <!-- CPS components including: Microsoft.VisualStudio.ProjectSystem -->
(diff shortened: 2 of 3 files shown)


.gitattributes
```diff
@@ -12,6 +12,7 @@
 # Set these files to XML (coloring) in PRs
 *.vsmanproj linguist-language=XML
 *.swixproj linguist-language=XML
+*.config linguist-language=XML
```
MiYanni, line 15: I added this so it would color the `nuget.config` file in the PR. However, it doesn't seem to be working... odd. 🤔

tmeschter (approved):

.gitattributes
ocallesp: is this working for other extensions like  A.swixproj ?

MiYanni: It is. See: https://github.com/dotnet/project-system/pull/8285/files#diff-1c7cad0a70f92aa5eb19332a8fd5fd7f973bd080fa1569265c7eed4eefa47a26

MiYanni: There might be something special about `.config` I'm not aware of. For now, I'll keep this there. I'll likely need to edit this file again in the future for CFS. I'll investigate at that time.

PR #8396, opened 2022-08-11
Delete a bunch of unused code

tmeschter:
Delete a bunch of unused code that dates back to when we were exposing the _active_ launch profile via the new Project Property Pages UI, before we had a separate Launch Profiles UI.

The `ActiveLaunchProfileNameValueProvider` exposes the _name_ of the active profile to the CPS properties system; the `LaunchTargetPropertyPageValueProvider` exposed the name of the `Rule` associated with the active profile (so we knew which properties to show in the UI), and the `ActiveLaunchProfileCommonValueProvider`, `ActiveLaunchProfileEnvironmentVariableValueProvider`, and `ActiveLaunchProfileExtensionValueProvider` were responsible for exposing the various properties we cared about in our launch profiles. Note the last three derive from the public `LaunchSettingsValueProviderBase` which I have _not_ removed, as I cannot guarantee it is no longer in use. It is still a reasonable base class for anyone who might want to expose some part of the active launch profile to the project property pages.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8396)


ocallesp (approved):

melytc (approved):
This sparks joy, Marie Kondo would approve 😄

MiYanni (approved):
So happy to see code deleted. ♥

PR #8400, opened 2022-08-16
Enable HighDpiMode property in the Project Properties UI.

melytc:
Relates to #8283, where we hid this property as we were not ready with the code generation part.

This PR consists of two commits:
- In the first one, we make the property visible in the UI. Additionally, we simplify the way the enum values are stored: instead of translating from integer to string and back again, we only keep the integer that represents the value. Later in the code generation, we will translate to the proper string, as that's the point in time where we actually need it.
- In the second one, we make the proper changes to read/write from the xml file (App.myapp file), save it into a property and later use its value to generate the necessary code for the Application.Designer.vb file. Note that we had to implement an override function for the `AddFieldAssignment`: we are not in .NET Core, and thus we do not have access to the complete `ApplicationService` namespace ([see documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualbasic.applicationservices?view=windowsdesktop-6.0)); since we cannot access the `EnumType` for `HighDpiMode` property, we opted to override the function to accept a `String` instead of a `Type`, as the CodeDom already works with strings and it would be too much of an effort to support a request to the Designer OOP server-side.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8400)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
+++ b/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
@@ -31,7 +31,22 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
         Private _codeDomProvider As CodeDomProvider
         Private _serviceProvider As ServiceProvider
 
+        ' Constants for properties to be generated.
         Private Const MyNamespaceName As String = "My"
+        Private Const _mainFormFieldName As String = "MainForm"
+        Private Const _singleInstanceFieldName As String = "IsSingleInstance"
+        Private Const _shutdownModeFieldName As String = "ShutDownStyle"
+        Private Const _enableVisualStylesFieldName As String = "EnableVisualStyles"
+        Private Const _saveMySettingsOnExitFieldName As String = "SaveMySettingsOnExit"
+        Private Const _splashScreenFieldName As String = "SplashScreen"
+        Private Const _highDpiModeFieldName As String = "HighDpiMode"
+
+        Private Const _highDpiMode_DpiUnaware = "DpiUnaware"
+        Private Const _highDpiMode_SystemAware = "SystemAware"
+        Private Const _highDpiMode_PerMonitor = "PerMonitor"
+        Private Const _highDpiMode_PerMonitorV2 = "PerMonitorV2"
+        Private Const _highDpiMode_DpiUnawareGdiScaled = "DpiUnawareGdiScaled"
+
         Friend Const SingleFileGeneratorName As String = "MyApplicationCodeGenerator"
 
         ''' <summary>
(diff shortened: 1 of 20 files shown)


src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -31,7 +31,22 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
         Private _codeDomProvider As CodeDomProvider
         Private _serviceProvider As ServiceProvider
 
+        ' Constants for properties to be generated.
         Private Const MyNamespaceName As String = "My"
+        Private Const _mainFormFieldName As String = "MainForm"
```
KlausLoeffelmann, line 36: In the WinForms repo, we're defining Consts starting with a capital letter. 
Should this be done here as well?

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -196,24 +211,42 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+                    AddFieldAssignment(Constructor, _shutdownModeFieldName, EnumType, "AfterAllFormsClose")
                 ElseIf MyApplication.ShutdownMode = ApplicationServices.ShutdownMode.AfterMainFormCloses Then
-                    AddFieldAssignment(Constructor, "ShutDownStyle", EnumType, "AfterMainFormCloses")
+                    AddFieldAssignment(Constructor, _shutdownModeFieldName, EnumType, "AfterMainFormCloses")
                 Else
                     Debug.Fail("Unexpected MyApplication.ShutdownMode")
                 End If
 
+                '    Me.HighDpiMode = HighDpiMode.xxx
+                ' TODO: Add constants
```
KlausLoeffelmann: Can this comment be deleted?

jacdavis (approved):
:shipit:

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -196,24 +211,42 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+                Dim HighDpiValue As String
+                Select Case MyApplication.HighDpiMode
```
KlausLoeffelmann, line 233: Again, I don't know what's best practice here for the repo. 😄 
I would create a `Shared Dictionary(Of Integer, String)` with the values, and just index the respective strings out of it.
But it depends on what your team agrees on readability. In terms of performance, it certainly doesn't matter, since this is code which is only called once in a while. So, this is just an "in case suggestions".

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -196,24 +211,42 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+                    Case 4
+                        HighDpiValue = _highDpiMode_DpiUnawareGdiScaled
+                    Case Else
+                        HighDpiValue = ""
+                End Select
+                AddFieldAssignment(Constructor, _highDpiModeFieldName, _highDpiModeFieldName, HighDpiValue)
+
                 GeneratedType.Members.Add(Constructor)
 
                 If MyApplication.MainFormNoRootNS <> "" Then
```
KlausLoeffelmann, line 251:
```suggestion
                If MyApplication.MainFormNoRootNS <> String.Empty Then
```

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -361,6 +394,25 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
             Method.Statements.Add(Statement)
         End Sub
 
+        ' Overload of AddFieldAssignment to accept a String instead of a Type for EnumType '
+        ' Background: For the case of HighDpiMode, the actual Type is only available in the Designer OOP server-side.
+        ' So to use types, we would need to generate this also in the server-side.
```
KlausLoeffelmann, line 398:
```suggestion
        ' So, to use types, we would need to generate the whole code also server-side.
```

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -361,6 +394,25 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+        ' We don 't think it's necessary though, and consider it is too much of an effort,
```
KlausLoeffelmann, line 399:
```suggestion
        ' We don't think it's necessary though, and consider it is too much of an effort,
```

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -361,6 +393,25 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+        ' so we decided to use a String for the type name client side.
+        Private Shared Sub AddFieldAssignment(Method As CodeMemberMethod, FieldName As String, EnumType As String, EnumFieldName As String)
+            Dim Statement As CodeAssignStatement
+            Statement = New CodeAssignStatement With {
```
KlausLoeffelmann, line 403: So, just for my own education:
How do you use capitalization of variable names in this (VB) repo. In the VB AppFramework, we still use small caps for parameters and local variables, and then also pretty much the same style as in C# for consts and member variables/backing fields.

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationProperties.vb
```diff
@@ -593,6 +596,34 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
+                    Case _
+                    0, ' There is no ApplicationService.HighDpiMode available to do it more clean.
+                    1,
+                    2,
+                    3,
+                    4
+                        ' Valid - continue
+                    Case Else
+                        Throw New ArgumentOutOfRangeException(NameOf(Value))
+                End Select
```
KlausLoeffelmann, line 614: I am not sure what the/your intention here is.
So, if you - and I am just assuming! 😄 - just wanted to test of correct range, you can just test like this:

```suggestion
               If Value < 0 Or Value > 4
                        Throw New ArgumentOutOfRangeException(NameOf(Value))
                End If
```

You could also use `To` as a range operator in VB:

```suggestion
                Select Case Value
                    Case 0 to 4 ' Valid: continue
                    Case Else
                        Throw New ArgumentOutOfRangeException(NameOf(Value))
                End Select
```

KlausLoeffelmann (commented):
Couple of suggestions, looks good to me!

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
melytc: Seems like the code convention for our VB files is a mix of pascal case and camel case for local variables, camel case for parameters, and again a mix of both plus all-caps for global variables. I might open another PR to address this, and try to have a more cohesive convention; it's not that frequent that we add/edit these files, but it would help in readability and code maintenance.

src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
melytc: Yes! Adding in next commit.

PR #8402, opened 2022-08-17
Put all feature flags definitions in one place.

ocallesp:
Code cleanup.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8402)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
@@ -1,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS;
+
+/// <summary>
+///     Provides feature flag names used throughout this project.
+/// </summary>
+internal static class FeatureFlagNames
+{
+    // Use this prefix name for all feature flags where the feature implementations belongs to dotnet project system.
+    public const string Prefix = "ManagedProjectSystem";
+
+    /// <summary>
+    /// Enables NuGet restore to detect cycles of type A -> B -> A to avoid loops.
+    /// </summary>
+    public const string EnableNuGetRestoreCycleDetection = Prefix + ".EnableNuGetRestoreCycleDetection";
+
+    /// <summary>
+    /// Enables logs in incremental builds.
+    /// </summary>
+    public const string EnableIncrementalBuildFailureOutputLogging = Prefix + ".EnableIncrementalBuildFailureOutputLogging";
+
+    /// <summary>
+    /// Enables incremental build to report build failures.
+    /// </summary>
+    public const string EnableIncrementalBuildFailureTelemetry = Prefix + ".EnableIncrementalBuildFailureTelemetry";
+}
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS;
+
+internal class FeatureFlagNames
+{
+    // Use this prefix name for all feature flags where the feature implementations belongs to dotnet project system.
+    private const string BaseName = "ManagedProjectSystem.";
+
+    internal const string EnableNuGetRestoreCycleDetection = BaseName + "EnableNuGetRestoreCycleDetection";
```
MiYanni, line 10: nit: When the class is internal, you can mark the members as public, since they'll be internal based on the class access modifier. This is a normal convention.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
drewnoakes: When this has come up in the past, the approach I've seen which makes the most sense to me is to decide whether the member would be public or internal if the class was made public, and mark it accordingly.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS;
+
+/// <summary>
+///     Provides feature flag names used throughout this project.
+/// </summary>
+internal static class FeatureFlagNames
+{
+    // Use this prefix name for all feature flags where the feature implementations belongs to dotnet project system.
+    public const string Prefix = "ManagedProjectSystem";
```
drewnoakes, line 11: I would avoid pulling this prefix out. It means we cannot search for the full feature flag string to find it's definition in source.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
MiYanni: That totally makes sense. However, I haven't really seen that as an adopted convention. Code gets weird if you try to read into intent that may or may not be there, especially code that is years or decades old.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/FeatureFlagNames.cs
MiYanni: Any chance the feature flag names could change to have a delimiter? Like `"ManagedProjectSystem_"`? Then, it is a bit more natural to think to search for a partial name.

drewnoakes: Fixed in https://github.com/dotnet/project-system/pull/8482

PR #8413, opened 2022-08-22
Add xaml rule for Resource item type

ocallesp:
Fixes [AB#1591205](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1591205)

Mismatching between items schema and rules.

**The bug:**
When a user opens the Property Panel to change the 'Build Action' to Resource,
the Solution Explorer removes the file.

While debugging I found that the project was not able to handle this item type
causing exceptions in the solution explorer when trying to get the icon monikers.

**The fix:**
Add xaml rules for 'Resource' item type.
`Resource.xaml `    		- This will allow the project to recognize this item type.
`Resource.FSharp.xaml`	- This will allow Fsharp projects to recognize this item type.
`Resource.ObjectBrowse.xaml` - This will display the information in the Property Panel

--------------------------------
**Repro**
Repro1591205

**The fix**
Fix1591205
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8413)


davkean:
Custom rules are normally unneeded for normal items; they should be falling back to the None item based on the AvailableItemName handling that automatically fakes up a rule for them. That looks broken for some reason and is handled by CPS.

ocallesp:
> 

Yep. This is an old bug. The recommendation from CPS is to add the xaml rules.

lifengl (approved):
your PR looks good to me.

kvenkatrajan (approved):

PR #8416, opened 2022-08-23
Mark AbstractFindByNameUnderAppDesignerSpecialFileProvider as an 'abstract' class.

ocallesp:
It is confusing this is not an abstract class

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8416)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractFindByNameUnderAppDesignerSpecialFileProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractFindByNameUnderAppDesignerSpecialFileProvider.cs
@@ -7,7 +7,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
     ///     that find their special file by file name under the AppDesigner folder, falling back
     ///     to the root folder if it doesn't exist.
     /// </summary>
-    internal class AbstractFindByNameUnderAppDesignerSpecialFileProvider : AbstractFindByNameSpecialFileProvider
+    internal abstract class AbstractFindByNameUnderAppDesignerSpecialFileProvider : AbstractFindByNameSpecialFileProvider
     {
         private readonly ISpecialFilesManager _specialFilesManager;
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/SpecialFileProviders/AbstractFindByNameUnderAppDesignerSpecialFileProvider.cs
```diff
@@ -7,11 +7,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.SpecialFileProviders
     ///     that find their special file by file name under the AppDesigner folder, falling back
     ///     to the root folder if it doesn't exist.
     /// </summary>
-    internal class AbstractFindByNameUnderAppDesignerSpecialFileProvider : AbstractFindByNameSpecialFileProvider
```
davkean, line 10: It looks like its supposed to be abstract, so looks like the fix should be mark it as such.

ocallesp: Good point. We should mark this as abstract if this is not going to be instantiated. 

Should this class keep using the prefix Abstract instead of Base even if there are no abstracts methods, and the description says this is a base class ?

davkean: Abstract and Base have the same meaning; a base class designed to be inherited from. "abstract" terminology came from the Roslyn code base which this project started from.

ocallesp: Thanks for the explanation. 

It seems that "base" is more used in cps.

adamint (approved):

PR #8429, opened 2022-08-26
Add a solution filter for editors only

ocallesp:
Editors are kind of independent project from dotnet project system, and we can treat them independently.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8429)

### DIFF
--- a/Editors.slnf
+++ b/Editors.slnf
@@ -1,0 +1,10 @@
+{
+  "solution": {
+    "path": "ProjectSystem.sln",
+    "projects": [
+      "setup\\VisualStudioEditorsSetup\\VisualStudioEditorsSetup.csproj",
+      "src\\Microsoft.VisualStudio.AppDesigner\\Microsoft.VisualStudio.AppDesigner.vbproj",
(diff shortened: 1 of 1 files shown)


MiYanni (commented):
Hmm. I don't know if this makes sense as a solution filter. The example in [the documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/solution-filters?view=vs-2022) isn't really representative of their purpose (from my understanding). They are usually for *very large* solutions and you want to filter parts of them down. Here, we basically could just have a solution file for each scenario: Editors, ProjectSystem, and Full (all). Also, I'd likely put it in the `src` folder and not root, even though it uses stuff in `setup`. Only reason is that it will be nice when we can eventually run `dotnet build` at repo root. 🤩

ocallesp:
I'll bring this topic to the design meeting.

drewnoakes:
One plus for this PR is dogfooding the solution filters feature.

drewnoakes (approved):

PR #8420, opened 2022-08-24
AzDO pipeline consolidation

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/7911 \
Fixes: https://github.com/dotnet/project-system/issues/7914 \
Fixes: https://github.com/dotnet/project-system/issues/7915 \
Fixes: https://github.com/dotnet/project-system/issues/7921 \
Fixes: https://github.com/dotnet/project-system/issues/7981

Related: https://github.com/dotnet/project-system/issues/7935 \
Related: https://github.com/dotnet/project-system/issues/8421 \
Related: https://github.com/dotnet/project-system/issues/8422 \
Related: https://github.com/dotnet/project-system/issues/8423 \
Related: https://github.com/dotnet/project-system/issues/8425

## Description

This PR is a wholesale cleanup to our Azure DevOps pipeline infrastructure. Also, it includes necessary changes for continuing to run OptProf after 8/25/2022, as the old means of creating a VM has been deprecated. These changes remove all use of Azure Release Pipelines and merge multiple YAML pipelines into 2 total pipelines: *official* and *pull-request*. Lastly, it has several ancillary changes that were simplistic to resolve after merging/moving pipelines.

## High-Level Changes

- Redesigned *official.yml* entirely, using stages and multiple jobs
  - Certain stages, such as *Insertion* and *Optimization*, can be included via pipeline parameters (in the Azure Pipelines UI)
  - Removed file copying tasks by having jobs filter what artifacts they download to run
  - **Hyper-documented** all pipelines/templates within the `.yml` files
  - Added 'task group'-like headers to the job templates for easier readability in the Azure Pipelines UI
- Renamed and updated the GitHub PR validation pipeline (now *pull-request.yml*)
  - Updated to the latest image for the pool (using `Windows.VS2022Preview.Amd64.Open`), which was required to be done by end-of-September
- Removed separate, single-purpose pipelines
  - *compliance.yml*: Tasks moved into a few templates, ran as stage in *official.yml*
  - *one-loc-build.yml*: Tasks moved into template, ran as stage in *official.yml*
  - *richnav.yml*: Tasks moved into template, ran as stage in *official.yml* and *pull-request.yml*
- Moved Azure Release Pipelines to YAML Pipelines
  - [Project System - OptProf](https://devdiv.visualstudio.com/DevDiv/_release?_a=releases&view=mine&definitionId=3197) is now the *Optimization* stage in *official.yml*.
[...]

PR #8430, opened 2022-08-26
Pipeline Changes: Followup

MiYanni:
A follow-up to the changes in https://github.com/dotnet/project-system/pull/8420.

- Fixing schedule start time for `official.yml` to be 4am PST instead of UTC.
- Set RichNav to always target production, since the other values are not for normal consumers. This was causing it to fail for PR builds.
- Upped the APIScan timeout to 90 minutes. It normally should only run for about 25-30 mins, but timed out last night. So, testing to see how long it could run until a real failure occurs.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8430)


eng/pipelines/official.yml
```diff
@@ -29,11 +29,11 @@ trigger:
...
 
 schedules:
-# Runs the pipelines at 4am and creates a VS insertion PR.
+# Runs the pipelines at 4am PST (11am UTC) and creates a VS insertion PR.
 # YAML reference: https://docs.microsoft.com/azure/devops/pipelines/yaml-schema/schedules-cron?view=azure-pipelines
 # CRON syntax: https://docs.microsoft.com/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#cron-syntax
-- cron: '0 4 * * *'
-  displayName: Daily 4am Build (with Insertion)
+- cron: '0 11 * * *'
+  displayName: Daily 4am Build with Insertion
```
kvenkatrajan: Please update displayName to : Daily 4am PT Build with Insertion

kvenkatrajan (approved):

eng/pipelines/official.yml
MiYanni: Done.

PR #8440, opened 2022-08-29
Update property page documentation

### LINKED ISSUE #8418: Incorrect document causing problems in rule system
The document in https://github.com/dotnet/project-system/blob/main/docs/repo/property-pages/property-specification.md leads people to create project rules in unnamed context group. We have similar document in this repo, which is doing the right thing:

```xml
<PropertyPageSchema Include="path\to\MyPropertyPage.xaml">
  <Context>Project</Context>
</PropertyPageSchema>
```

Currently, adding a rule without specifying context would inject the rule into all valid catalogs. This will add the rule into the "project" one and make the page to work correctly as a project page.  However, the rule will also be injected into 'browseObject' which leads the bowser properties of the project to be wiped out (because only one rule is allowed to describe project's browse object.

Based on this problem, we will kill this behavior, so all rules without specifying context will only stay as a rule in the empty context catalog, and those pages will no longer be shown in the next update.

Please fix the incorrect document, and potentially work with other people who may create pages by following the incorrect document, as those pages will stop working.

tmeschter:
Fixes #8418.

Update our property page documentation to reflect that a `PropertyPageSchema` item must include the appropriate `Context` metadata. If the `Context` is not supplied the `Rule` may be pulled into _all_ contexts; if it is not set to "Project" the item will not be found by the property pages. Either will cause bugs.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8440)

### DIFF
--- a/docs/repo/property-pages/back-end-architecture.md
+++ b/docs/repo/property-pages/back-end-architecture.md
@@ -69,4 +69,4 @@ At the very bottom of the implementation stack are XAML Rule files, so-called be
 
 These Rule files are where the declarative, metadata-driven aspect of the new property pages comes into play. These files declare page, the page's metadata, its properties, the storage mechanisms to use (i.e. which `IProjectPropertiesProvider` will handle a particular property), and a bunch of other property metadata: documentation links, search terms, descriptions, the property type, whether the value can vary across project configurations, etc.
 
-Adding a new property is largely a matter of defining it in the appropriate XAML file. Adding a new page is a matter of creating a new XAML file and including it in the project as a `PropertyPageSchema` item.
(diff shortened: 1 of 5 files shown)


ocallesp (approved):

melytc (approved):

tmeschter:
Bypassing checks and merging as this is a doc-only change.

MiYanni:
> Bypassing checks and merging as this is a doc-only change.

Hmm. That's interesting. We have that pipeline marked as `required` in GitHub, but the pipeline itself has filters so that it doesn't run for doc-only changes. I guess a skipped pipeline is not marked as 'passing' for some reason... I'll figure out how best to make this hassle-free.

PR #8453, opened 2022-09-01
Fix wrong behavior in Symbolic Renaming prompt

### LINKED ISSUE #8441: "Enable symbolic renaming when renaming files" options not persisted properly
## Visual Studio Version

17.4.0 Preview 2.0 [32829.42.main]

## Summary

The options around renaming a type when the containing file is renamed are not persisted properly, which leads to types being renamed even when the user has told us not to do that.

## Steps to Reproduce

1. Open _Tools | Options | Projects and Solutions | General_ and make sure the following two options are enabled:
image
2. Click "OK" to close the Options dialog.
3. Create a new C# class library targeting .NET 6.
4. In Solution Explorer, right-click on Class1.cs, select "Rename" from the menu, and rename it to "Class2.cs".
5. When this dialog comes up:
image
Check "Don't show again" and click "No".
6. Verify that the type inside Class2.cs is still named "Class1".
7. Add a new class named "Class3" to the project.
8. In Solution Explorer, right-click on Class3.cs, select "Rename" from the menu, and rename it to "Class4.cs".

## Expected Behavior

- The file is renamed to Class4.cs.
- The dialog asking if you want to rename the type does _not_ appear.
- The type in the file is _not_ renamed; it is still Class3.

## Actual Behavior

The dialog asking if you want to rename the type does not appear, **but the Class3 type is automatically renamed to Class4.**

Going back to _Tools | Options | Projects and Solutions | General_, we see the settings are now:

image

Which is incorrect.

## User Impact

We will now _always_ rename types to match file names, which is the exact opposite of what the user asked for.

ocallesp:
Fixes #8441 

Fix wrong behavior in Enable symbolic renaming

if "Don't show again" is checked,
--  then persist UNcheck "prompt for symbolic renaming when renaming files"
-----   if "Yes", then persist check "enable symbolic renaming when renaming files"
-----   if "No", then persist UNcheck "enable symbolic renaming when renaming files"

If "Don't show again" is UNchecked, don't make changes to persisted values.

fixGithubBug8453

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
@@ -20,11 +20,6 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
     [AppliesTo(ProjectCapability.CSharpOrVisualBasicLanguageService)]
     internal partial class RenamerProjectTreeActionHandler : ProjectTreeActionHandlerBase
     {
-        private const string SymbolicRenameName = "SolutionNavigator.EnableSymbolicRename";
-        private const string RenameSymbolCategory = "Environment";
-        private const string RenameSymbolPage = "ProjectsAndSolution";
-        private const string RenameSymbolOption = "PromptForRenameSymbol";
-
         private readonly IEnvironmentOptions _environmentOptions;
         private readonly IUnconfiguredProjectVsServices _projectVsServices;
         private readonly IProjectThreadingService _threadingService;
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
```diff
@@ -247,11 +242,15 @@ private async Task<bool> CheckUserConfirmationAsync(string oldFileName)
 
             string renamePromptMessage = string.Format(CultureInfo.CurrentCulture, VSResources.RenameSymbolPrompt, oldFileName);
 
-            bool userSelection = _userNotificationServices.Confirm(renamePromptMessage, out bool disablePromptMessage);
+            bool yesSelected = _userNotificationServices.Confirm(renamePromptMessage, out bool disablePromptMessage);
```
tmeschter, line 245: Suggestion: `shouldRename` instead of `yesSelected`.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
```diff
@@ -247,11 +242,15 @@ private async Task<bool> CheckUserConfirmationAsync(string oldFileName)
...
 
-            _environmentOptions.SetOption(RenameSymbolCategory, RenameSymbolPage, RenameSymbolOption, !disablePromptMessage);
+            if (disablePromptMessage)
+            {
+                await settings.SetValueAsync(VsToolsOptions.OptionEnableSymbolicRename, yesSelected, true);
```
tmeschter, line 249: Suggestion: Add the argument name for the `true` to help someone reading the code understand what it means.

tmeschter (approved):
Approved, with a couple of suggestions.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/RenamerProjectTreeActionHandler.cs
ocallesp: done

PR #8454, opened 2022-09-01
Myapp and Settings file generator fixes

adamint:
This includes two fixes, as noted in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1591002/

1: HighDpiMode is not a valid property name for some netframework versions. It needs to be removed.
2: The generated My.Settings() property (shown below) was missing the "My" identifier if generated in a VB project. This has now been added
image

This has been tested for:
WPF netframework VB: yes
WinForms netframework VB:
WPF netcore VB: yes
WinForms netcore VB: yes

WPF netframework C#: yes
WinForms netframework C#: yes
WPF netcore C#: yes
WinForms netcore C#: yes

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
+++ b/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
@@ -47,7 +47,7 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
         Private Const HighDpiMode_PerMonitor = "PerMonitor"
         Private Const HighDpiMode_PerMonitorV2 = "PerMonitorV2"
         Private Const HighDpiMode_DpiUnawareGdiScaled = "DpiUnawareGdiScaled"
-        
+
         Friend Const SingleFileGeneratorName As String = "MyApplicationCodeGenerator"
 
         ''' <summary>
--- a/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
+++ b/src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
@@ -10,6 +10,7 @@ Imports System.Runtime.InteropServices
 Imports EnvDTE
 
 Imports Microsoft.VisualStudio.Designer.Interfaces
+Imports Microsoft.VisualStudio.Editors.Common
 Imports Microsoft.VisualStudio.Editors.Interop
 Imports Microsoft.VisualStudio.OLE.Interop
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Editors/MyApplication/MyApplicationCodeGenerator.vb
```diff
@@ -227,25 +219,7 @@ Namespace Microsoft.VisualStudio.Editors.MyApplication
...
-                    Case 2
-                        HighDpiValue = HighDpiMode_PerMonitor
-                    Case 3
-                        HighDpiValue = HighDpiMode_PerMonitorV2
-                    Case 4
-                        HighDpiValue = HighDpiMode_DpiUnawareGdiScaled
-                    Case Else
-                        HighDpiValue = String.Empty
-                End Select
-                AddFieldAssignment(Constructor, HighDpiModeFieldName, HighDpiModeFieldName, HighDpiValue)
```
melytc, line 247: Removing the code generation doesn't sound like a proper fix: we introduced it for .NET Core by surfacing it in the project properties UI; removing this would mean that changes in the UI would not be reflected in the actual project.

Could you instead check what target framework is using to decide if we should show the property or generate its code? 

cc: @KlausLoeffelmann

KlausLoeffelmann: I wasn't aware that we are targeting .NET Framework with the new templates?
We won't be running into an issue with .NET Core 3.1, since it was not supported for the VB Application Framework to begin with.
For .NET 5, yes, we might be running into problems. But also, .NET 5 is out of support, and since we only had a very rudimentary AppFramework support solution via templates for it at the time to begin with, I am not sure, if this is something, which _can_ be properly addressed.

adamint: Apologies @melytc, should've checked on this property before removing it. @melytc and @KlausLoeffelmann, I've updated to only write to the myapp file in netcore projects, and now the property only appears in netcore 6+ projects

adamint: Can you confirm this is now the correct behavior?

KlausLoeffelmann: This should be the right thing to do.
We introduced the property in .NET 6.

KlausLoeffelmann (approved):
Looks good to me!

MiYanni (approved):

PR #8456, opened 2022-09-06
Fix language service blocking load in command line mode

drewnoakes:
Fixes [AB#1600327](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1600327)

Previously we checked whether we were running in command line mode only during the initialisation of `LanguageServiceHost`.

Another code path that requires this check is the `ProjectAutoLoad` method. Without that check, a task is registered via `PrioritizedProjectLoadedInHostAsync` that will not complete, blocking project load indefinitely.

This change adds such a check. It also adds validation on other public methods to ensure they are not called while in command line mode. These checks convert bugs from hangs to exceptions, making them easier to diagnose and fix.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8456)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -95,9 +95,9 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
 
     protected override async Task InitializeCoreAsync(CancellationToken cancellationToken)
     {
-        if (_languageServiceHostEnvironment is not null && !await _languageServiceHostEnvironment.IsEnabledAsync(cancellationToken))
+        if (await IsDisabledAsync(cancellationToken))
         {
-            // Our hosting environment is telling us we're not enabled.
+            // We are not enabled, so don't perform any initialization.
             return;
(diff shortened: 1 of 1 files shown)


davkean:
I would consider an approach in the future where you define a capability that causes these services to not even start so you don't need procedural checks for this state.

davkean (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -297,4 +309,18 @@ protected override Task DisposeCoreUnderLockAsync(bool initialized)
...
         return Task.CompletedTask;
     }
+
+    private async Task<bool> IsDisabledAsync(CancellationToken cancellationToken)
+    {
+        // Check whether our hosting environment is telling us we're enabled.
+        return _languageServiceHostEnvironment is not null && !await _languageServiceHostEnvironment.IsEnabledAsync(cancellationToken);
+    }
+
+    private async Task EnsureEnabledAsync(CancellationToken cancellationToken)
```
drewnoakes, line 319: I'll rename this to `ValidateEnabledAsync`.

PR #8462, opened 2022-09-07
Workaround for combobox not announcing value changed notifications in legacy prop pages

adamint:
This works by using reflection to call RaiseAutomationNotification if we're on >=net4.8, invoking with known enum values. This occurs by overriding the OnSelectedIndexChanged event in wrapping ComboBox with our own AccessibleComboBox. ComboBox is then replaced with AccessibleComboBox in all instances where it is created in the property pages.

This is a workaround for https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1576845/


adamint:
For whoever reviews, checkout this branch and verify with NVDA that legacy comboboxes announce value change.

ocallesp:
I think it would be helpful to see an animated gif of this change.

adamint:
> I think it would be helpful to see an animated gif of this change.

Good idea! I'll add one.

adamint:
arrow keys bug blank winforms

src/Microsoft.VisualStudio.Editors/PropPages/AccessibleComboBox.vb
```diff
@@ -0,0 +1,15 @@
...
+Imports System.Windows.Forms
+
+Namespace PropPages
+
+    Public Class AccessibleComboBox
+        Inherits ComboBox
+
+        Protected Overrides Sub OnSelectedIndexChanged(e As EventArgs)
+            MyBase.OnSelectedIndexChanged(e)
+            AccessibilityObject.GetType().GetMethod("RaiseAutomationNotification")?.Invoke(AccessibilityObject, new object() {2,4,  Items.Item(SelectedIndex)})
```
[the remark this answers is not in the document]
adamint: @MiYanni this is VB so points 2&4 don't work here :(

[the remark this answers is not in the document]
adamint: added comments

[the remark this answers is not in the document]
MiYanni: OHHHH. Right, VB. So, the casing is what threw my brain off. In VB, it would normally be like:

`New Object() { 2, 4, Items.Item(SelectedIndex) }`

It is a convention thing for the language [as seen here](http://makble.com/how-to-create-object-array-in-vbnet).
BUT! [According to this](https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/collection-initializers/), you can just use:

`{ 2, 4, Items.Item(SelectedIndex) }` 

without `New Object()`. You'd have to check that it compiles.

src/Microsoft.VisualStudio.Editors/PropPages/AccessibleComboBox.vb
```diff
@@ -0,0 +1,22 @@
...
+        Inherits ComboBox
+
+        ' workaround for https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1576845/
+        ' raises automation notification on net4.8-enabled VS instances, as *something* (likely in winforms) is swallowing these 
+        ' change events
+        Protected Overrides Sub OnSelectedIndexChanged(e As EventArgs)
+            MyBase.OnSelectedIndexChanged(e)
+            ' invokes RaiseAutomationNotification adding a notification of type ActionCompleted (2) with notification processing order
+            ' CurrentThenMostRecent (4)
+            AccessibilityObject.GetType().GetMethod("RaiseAutomationNotification")?.Invoke(AccessibilityObject, new object() { 2, 4, Items.Item(SelectedIndex) })
```
davkean, line 19: Pass in the exact argument types of RaiseAutomationNotification to avoid future versions of .NET from crashing this code if they add an overload.

melytc (approved):
Other than David's comment, LGTM

src/Microsoft.VisualStudio.Editors/PropPages/AccessibleComboBox.vb
adamint: I cannot, as these types are also not available in this version, and are otherwise difficult to get via reflection

drewnoakes: You should be able to resolve the `Type` objects via reflection using namespace and name as strings.

Failing that, can you use `GetMethods` and filter for the one with the number of arguments you expect?

drewnoakes, line 19: Consider calling `GetMethod` once in a `Shared` constructor, and caching the result in a static field. Reflection can be slow, so it's good to avoid doing it per-call. Especially if this method is called for every combo box in the UI as the UI is instantiated.

src/Microsoft.VisualStudio.Editors/PropPages/AccessibleComboBox.vb
```diff
@@ -0,0 +1,22 @@
+﻿' Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+Imports System.Windows.Forms
+
+Namespace PropPages
+
+    ' This class wraps ComboBox, overriding the OnSelectedIndexChanged event to enable change notifications for comboboxes in legacy pages.
+    ' Must be used instead of ComboBox for legacy prop pages
+    Public Class AccessibleComboBox
```
drewnoakes, line 9: Does this have to be public to work with the designers?

drewnoakes (approved):

src/Microsoft.VisualStudio.Editors/PublicAPI.Shipped.txt
```diff
@@ -1084,4 +1084,7 @@ Shared My.Resources.GeneralOptionPageResources.General_PreferSingleTargetBuildsO
 Shared My.Resources.GeneralOptionPageResources.General_MultiTargetingSettings_Title() -> String
 Shared My.Resources.GeneralOptionPageResources.General_PreferSingleTargetBuildsOnLaunchDescription() -> String
 Microsoft.VisualStudio.Editors.MyApplication.IVsMyApplicationProperties.HighDpiMode() -> Integer
-Microsoft.VisualStudio.Editors.MyApplication.IVsMyApplicationProperties.HighDpiMode(Value As Integer) -> Void
\ No newline at end of file
+Microsoft.VisualStudio.Editors.MyApplication.IVsMyApplicationProperties.HighDpiMode(Value As Integer) -> Void
+Overrides PropPages.AccessibleComboBox.OnSelectedIndexChanged(e As System.EventArgs) -> Void
+PropPages.AccessibleComboBox
+PropPages.AccessibleComboBox.New() -> Void
```
tmeschter, line 1090: Do these need to be public?

tmeschter (approved):
Approved, though see my comment about whether or not the new type really needs to be accessible.

src/Microsoft.VisualStudio.Editors/PropPages/AccessibleComboBox.vb
jacdavis: Remove the reflection and pinvoke directly into UiautomationCore. This is just a thin wrapper and reflection is unnecessary.

src/Microsoft.VisualStudio.Editors/PublicAPI.Shipped.txt
adamint: Otherwise, we get analyzer complaints

adamint:
Abandoning because underlying bugs are being marked external

PR #8464, opened 2022-09-08
Add SourceLink to NuGet packages

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/5619

Adds the SourceLink functionality to our NuGet packages. It also makes our build marked with the deterministic flag.

### Verification
I used the [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) to verify the packages. When they started, they looked like this:
image

I enabled SourceLink but it was still showing issues. Seems I needed the `EmbedUntrackedSources` setting.
image

After adding `EmbedUntrackedSources`, the build was broken. After some investigation, I found an issue that was reported for the problem, but no workaround: https://github.com/dotnet/sourcelink/issues/492 \
I looked into how the generated `.g.cs` files wanted to reference the `.xaml`. Then, I determined how to output them to that path. I added those targets to the affected projects. After that, we had SourceLink working.
image

I noticed the *Deterministic* flag was still red. Found that I needed to add `ContinuousIntegrationBuild` for it to add the `/deterministic` compiler flag. After that, everything was green.
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8464)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -55,6 +55,8 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <NoWarn>$(NoWarn);NU5125</NoWarn>
+    <!-- Sets deterministic compilation (/deterministic compiler flag) -->
+    <Deterministic>true</Deterministic>
 
     <!-- SourceLink properties -->
     <!-- https://github.com/dotnet/sourcelink#using-source-link-in-net-projects -->
@@ -66,9 +68,13 @@
     <!-- Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <!-- Marks the assembly be built in a determinstic way. -->
-    <!-- See: https://github.com/dotnet/sdk/issues/16325#issuecomment-981818164 -->
-    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+    <!-- Controls both the compiler determinism (same as Deterministic) and turns on normalized PDB paths, which may break local debugging scenarios. -->
+    <!-- We only want this on for CIBuild, so PDB files point to the local source locations on disk, making finding local files easy during debugging. -->
+    <!-- See:
+      - https://github.com/dotnet/sdk/issues/16325#issuecomment-981818164 
+      - https://github.com/dotnet/project-system/pull/8464#discussion_r965499524
+    -->
+    <ContinuousIntegrationBuild Condition="'$(CIBuild)' == 'true'">true</ContinuousIntegrationBuild>
 
     <!-- Test project properties -->
     <IsUnitTestProject Condition="'$(IsUnitTestProject)' == ''">false</IsUnitTestProject>
@@ -114,6 +120,7 @@
     <PackageLicenseUrl>http://go.microsoft.com/fwlink/?LinkId=529443</PackageLicenseUrl>
     <RepositoryUrl>https://github.com/dotnet/project-system</RepositoryUrl>
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
+    <RepositoryType>git</RepositoryType>
     <!-- By default, do not build a NuGet package for a project. -->
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -54,9 +54,21 @@
     <ForceGenerationOfBindingRedirects>true</ForceGenerationOfBindingRedirects>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
-    <!-- TODO: Might be related to: https://github.com/dotnet/project-system/issues/5619 -->
-    <EnableSourceLink Condition="'$(CIBuild)' == 'true'">true</EnableSourceLink>
     <NoWarn>$(NoWarn);NU5125</NoWarn>
+
+    <!-- SourceLink properties -->
+    <!-- https://github.com/dotnet/sourcelink#using-source-link-in-net-projects -->
+    <EnableSourceLink>true</EnableSourceLink>
+    <!-- Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- Embed source files that are not tracked by the source control manager in the PDB -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
+    <IncludeSymbols>true</IncludeSymbols>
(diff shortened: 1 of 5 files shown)


Directory.Build.props
```diff
@@ -54,10 +54,22 @@
...
+    <!-- Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- Embed source files that are not tracked by the source control manager in the PDB -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <!-- Marks the assembly be built in a determinstic way. -->
+    <!-- See: https://github.com/dotnet/sdk/issues/16325#issuecomment-981818164 -->
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
```
MiYanni, line 71: I didn't make any of these conditional on `CIBuild` as there didn't seem to be a reason to not enable this functionality for local builds too.

drewnoakes (approved):
This looks similar to what I've had to do to get this working in my own libraries (minus the XAML file bit). Thanks for the clear PR description too.

In one of my libs I also added these. Do you think they'd be useful here?

```xml
<Deterministic>true</Deterministic>
<IncludeSource>true</IncludeSource>
```

Also, do we specify `PackageProjectUrl`, `PackageLicenseExpression`, `RepositoryType` and `RepositoryUrl` anywhere?

Directory.Build.props
```diff
@@ -54,10 +54,22 @@
...
+    <!-- https://github.com/dotnet/sourcelink#using-source-link-in-net-projects -->
+    <EnableSourceLink>true</EnableSourceLink>
+    <!-- Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- Embed source files that are not tracked by the source control manager in the PDB -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <!-- Marks the assembly be built in a determinstic way. -->
```
KirillOsenkov, line 69: `determinstic` is misspelled

Directory.Build.props
KirillOsenkov: When you build locally, it is convenient that the .pdbs point to your local source locations on disk, so finding local files is easy during debugging. My understanding is that setting these locally will normalize the paths to source files to not contain your local paths, so VS debugger might find it difficult to find local sources.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
```diff
@@ -77,8 +77,8 @@
 
   <ItemGroup>
     <Page Include="ProjectSystem\VS\PropertyPages\DebugPageControl.xaml">
-      <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
+      <SubType>Designer</SubType>
```
KirillOsenkov, line 81: I'm not sure, but I think the <SubType> metadata is never really useful and can be omitted?

KirillOsenkov:
I think `Deterministic` property is already true by default in SDK-style projects.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
davkean: @MiYanni  Did Visual Studio move this? If so, its a bug.

MiYanni: Nope. I moved it. The 2 `Page` items below it have this ordering. It didn't make sense to have 1 with different ordering, and AFAIK, the ordering doesn't matter. Just satisfies my OCD.

Directory.Build.props
MiYanni: Alright. I'll change it and put that exact information there. You know how difficult it is to find what a single MSBuild property does? MSBuild desperately needs a documentation (IntelliSense-style) system built-in for these things. There hundreds and hundreds of properties in a binlog. Trying to figure out what each does is madness. I'm commenting on these things as much as possible at this point because I won't remember 6 months from now.

Directory.Build.props
MiYanni: Derp! Missed an `i` in there.

MiYanni:
> This looks similar to what I've had to do to get this working in my own libraries (minus the XAML file bit). Thanks for the clear PR description too.
> 
> In one of my libs I also added these. Do you think they'd be useful here?
> 
> ```
> <Deterministic>true</Deterministic>
> <IncludeSource>true</IncludeSource>
> ```
> 
> Also, do we specify `PackageProjectUrl`, `PackageLicenseExpression`, `RepositoryType` and `RepositoryUrl` anywhere?

I'll look into what `Deterministic` and `IncludeSource` actually do. My assumption is `ContinuousIntegrationBuild` sets `Deterministic` to true.

Package properties are here: https://github.com/dotnet/project-system/blob/770722611f07192b9f501fe73932687e7a320d0e/Directory.Build.props#L102-L104

We use `PackageLicenseUrl` instead of `PackageLicenseExpression`, and we don't seem to have `RepositoryType` set. I can add that.

MiYanni:
> I think `Deterministic` property is already true by default in SDK-style projects.

According to what NuGet Package Explorer was showing, the assemblies were not deterministic until I set `ContinuousIntegrationBuild` to true. See the description for this PR.

KirillOsenkov:
Yes, NuGet Package Explorer's definition of deterministic includes the deterministic compilation as well as normalized paths to source files in the Pdb. The `Deterministic` property only controls the compiler determinism, whereas `ContinuousIntegrationBuild` also turns on normalized Pdb paths (which may break local debugging scenarios as I've outlined above). I think ideally set ContinuousIntegrationBuild to true on CI only.

Directory.Build.props
KirillOsenkov: I usually search the preprocessed project text for the property name (or use Find in Files feature). It is hard indeed, so good job commenting it.

Directory.Build.props
MiYanni: My extensions in VSCode somehow all disabled themselves. They all say "Reload Required" beside them. Explains why I didn't see the spelling error. 🤔

MiYanni:
@drewnoakes I'm adding the `Deterministic` property but not the `IncludeSource` property. It seems to not function with the `SymbolPackageFormat` I'm adding. https://github.com/NuGet/Home/issues/8589

PR #8472, opened 2022-09-08
Fix stack overflow in dependencies tree search

drewnoakes:
Fixes [AB#1613079](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1613079)

That feedback ticket showed a stack overflow when populating the parent items of a given search hit. While I do not yet have a repro for the bug, it appears this would only have been possible if an item was its own parent. I do not know how this might occur in practice, but we should defend against it in theory.

The fix here is to ensure that the `ContainedByCollection` property is non-null earlier, as we test this property on the parent to see whether we need to recur. If the parent and the child were the same, previously we would have seen a null and invoked the recursion. With this change, we set the value to non-null before walking parents, which prevents this error.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8472)


jacdavis (approved):
:shipit:

MiYanni (approved):

PR #8475, opened 2022-09-09
Ensure the first workspace update is from evaluation

drewnoakes:
Fixes [AB#1586247](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1586247)

In the language service host, we need to process evaluation data before design-time build data. We need an evaluation to create the Roslyn workspace object, and we need that workspace to apply build data.

In practice, build data can arrive first. This may be due to caching, for example. We see this unexpected ordering more on perf lab machines too, interestingly.

The previous code didn't handle the out-of-order case correctly. If build data arrived first, it would block waiting for evaluation data. However the upstream TPL action block serialises callbacks, so any future update that might unblock the build data handler cannot run, leading to deadlock.

The fix here uses a `TransformManyBlock` to allow any initial build updates to be queued until evaluation data arrives. The downstream handler (`Workspace.OnWorkspaceUpdateAsync`) will therefore always receive an evaluation update first.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8475)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
@@ -43,9 +43,6 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
     private readonly CancellationToken _unloadCancellationToken;
     private readonly string _baseDirectory;
 
-    /// <summary>Completes when the workspace has integrated evaluation data.</summary>
-    private readonly TaskCompletionSource _hasEvaluationData = new(TaskCreationOptions.RunContinuationsAsynchronously);
-
     /// <summary>Completes when the workspace has integrated build data.</summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -43,9 +43,6 @@ internal enum WorkspaceState
     private readonly CancellationToken _unloadCancellationToken;
     private readonly string _baseDirectory;
 
-    /// <summary>Completes when the workspace has integrated evaluation data.</summary>
-    private readonly TaskCompletionSource _hasEvaluationData = new(TaskCreationOptions.RunContinuationsAsynchronously);
```
drewnoakes, line 47: This class no longer has to track whether it has seen evaluation data. The responsibility of ensuring the first update contains evaluation data has moved to the caller. This class now throws if that expectation is not met.

PR #8480, opened 2022-09-12
Use full-form literals for telemetry const strings

drewnoakes:
Previous code used string concatenation and helper functions to create name strings for telemetry events and properties.

This meant that it wasn't possible to search for the full event name and find the source code that produced it.

With this change, telemetry event names and their property names appear as full string literals in the source code and are easily found during search.

This also removes the `BuildPropertyName` method, so that there's no runtime computation of these strings when the type loads during VS start.

I also broke the telemetry property names class up into smaller nested static classes to provide some kind of namespacing on these property names.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8480)


drewnoakes:
Similar to #8482

ocallesp (approved):

MiYanni (approved):

PR #8481, opened 2022-09-12
Fix potential exception handling solution events

drewnoakes:
There are currently two places we might attempt to set the results on our TaskCompletionSource. I've observed, in a debugger, an exception when we set this state twice.

This change uses `TrySetResult` rather than `SetResult`, as the former will not throw an exception if a result has already been set (or the result cancelled/faulted).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8481)


jacdavis (approved):
:shipit:

MiYanni (approved):

PR #8485, opened 2022-09-13
Fix broken links in issue template

drewnoakes:
The `..` syntax does not render correctly on GitHub as a way to reference source files in the repo.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8485)


drewnoakes:
@MiYanni do you know a better way to address this?

ocallesp (approved):

MiYanni (approved):
Oh duh! That's totally my bad. The relative links here would work **from the file itself** if you looked at it in the repo on GitHub. But you want the links to work in the issue that this template creates.

I think my logic was that you're updating the files to create a PR for creating the branch. Making the links use full paths mean they include the branch name (aka `main`), and you're obviously not updating `main` directly. But since this file (the template) is not looked at directly in the repo itself, this is the best option (using a full path).

Issue #6453, opened 2020-07-30
Tag the commits that were inserted into a particular VS version


dmonroym:
Use https://github.com/dotnet/roslyn-tools/tree/master/src/CreateTagsForVSRelease

MiYanni:
We would not use a roslyn-tools for this action. There is likely a way to do this via the automatic insertion process, to be done in: https://github.com/dotnet/project-system/issues/7911

MiYanni:
The original solution in https://github.com/dotnet/project-system/pull/8517 did not work. The multi-repo triggering mechanism requires that the repo containing the YAML file **also** be within the same AzDO org. This was not mentioned in the documentation linked within the PR. To make a solution for this using the same concept, the current idea is to use an internal mirror of this repo. There will need to be changes to the `tagger.yml` to account for this. That PR can resolve this issue.

kvenkatrajan:
@MiYanni, can this be closed as : Closed not planned at the moment.

MiYanni:
@kvenkatrajan The only reason this got put on-hold is a red-tape issue which I won't go into details here. However, I can workaround that in the near term (since the problem wasn't resolved yet). I've added you to an email thread related to it.

MiYanni:
This was *nearly* completed prior to my team move. Everything is in-place for this to work but there were some tweaks needed:
- The [tagger pipeline](https://github.com/dotnet/project-system/blob/main/eng/pipelines/tagger.yml) needs to be modified to run properly internally. This pipeline needs to run based off of our `-Trusted` repo (`main` branch since that is cloned). It cannot be connected to the GitHub repo as the trigger will not function properly.
- The scripts related to the process need to explicitly target the public GitHub repo. Right now, they target the repo they run on, which would mean that it would create tags in `-Trusted` (which we don't want).
- After the automatic tagging is set up and working, follow the information here to tag all previous releases: https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/34672/Release-Tagging-(script)

smitpatel:
Is idea to create tag for VS releases or every VS insertion? Tagger.yaml is working to tag every insertion.

MiYanni:
> Is idea to create tag for VS releases or every VS insertion? Tagger.yaml is working to tag every insertion.

@smitpatel The tagger pipeline has a specific trigger mechanism that only activates when VS creates a tag following a certain format. It combines these two features:
[...]

PR #8486, opened 2022-09-14
Add short commit ID to VS Insertion payload

MiYanni:
Related: https://github.com/dotnet/project-system/issues/6453

Added task to the VS insertion job to create a short commit ID (~~7~~ 10 characters). Then, added this commit ID to the insertion payload name. The commit ID will be parsed out of the insertion PR title and used by a separate pipeline to tag the repo after insertion PRs are completed.

I investigated several ideas to pass the commit ID (of our repo) to the insertion PR. This was (by far) the simplest way to achieve passing the data, since we can read the commit title (the commit a VS insertion creates) directly out of the Git log for the VS repo. Other options would involve REST calls to AzDO APIs, which seems unnecessary when I only need ~~7~~ 10 characters.

More details about how this works will be added to the YAML/script in subsequent PRs when the tagger pipeline consumes this commit ID.

### Test Build
~~https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6684041&view=results~~
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6687793&view=results
> Insert DotNet-Project-System (UpdateInsertionPayloadName:20220914.10:a5fea54503) Payload into main

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8486)


ocallesp:
Perhaps I am missing some context. Why this change is needed ? or what is the issue it is trying to fix ?

MiYanni:
@ocallesp This allows us to create a Git tag in the repo based on an insertion PR being merged. Creating these tags will allow us to create the detailed insertion descriptions we used to have. There's more details about it in our Teams channel.

eng/pipelines/templates/generate-insertion.yml
```diff
@@ -27,6 +27,11 @@ jobs:
       # When this isn't provided, the tool sets the --AadAuth flag instead of using --PatAuthEnvVar. The AadAuth doesn't seem to work for this situation.
       AccessToken: $(System.AccessToken)
 
+  # Sets the ShortCommitId variable for the job.
+  # https://docs.microsoft.com/azure/devops/pipelines/process/set-variables-scripts?view=azure-devops&tabs=powershell#set-variable-properties
+  - powershell: Write-Host "##vso[task.setvariable variable=ShortCommitId;isoutput=true]$('$(Build.SourceVersion)'.Substring(0,6))"
```
drewnoakes, line 32: 6 characters might not be unique. I've seen collisions on 6 characters of the SHA1. Better to include more here.

See https://twitter.com/drewnoakes/status/830190508727205889 for a 7 character collision I hit in real life. (I think the maths in that tweet is wrong. The chance is 1 in 16^7, or 1 in 268,435,456, or 0.00000037%.)

drewnoakes (approved):

ocallesp (approved):

eng/pipelines/templates/generate-insertion.yml
MiYanni: Oy. I meant for this to be 7 characters, but I forgot the parameters are *index*, *length*. 😅 The default for Git is 7 so that's what I was going for. I saw this [response here](https://stackoverflow.com/a/21015031/294804) about length and probabilities and such. I can run `git rev-parse --short` in this small script instead of just truncating the full commit ID (which is `Build.SourceVersion` in an AzDO pipeline). One interesting thing about the idea of using `git rev-parse --short` is that it'll only give an ID that is unique in the *present*; whereas I need to use this ID in the *future* to give a diff over time. Technically, a commit between now and then could conflict with this ID but that is an incredibly low chance of happening.

I think I'll still use the Git command here.

MiYanni: The test build resulted in this title, so I still need to fix it anyway. 😂
> Insert DotNet-Project-System (UpdateInsertionPayloadName:20220914.1:$(ShortCommitId)) Payload into main

drewnoakes: If you can use `git rev-parse --short` then that'd be better. Note that it'd only be unique up until that point. Beyond that, it could conflict. I think just using 10 characters would suffice, no? The odds of a conflict at that point are astronomically low.

MiYanni: Latest build with the change to 10 digits worked appropriately.
> Insert DotNet-Project-System (UpdateInsertionPayloadName:20220914.10:a5fea54503) Payload into main

melytc (approved):

PR #8487, opened 2022-09-14
Re-enable APIScan and allow timeouts (for certain jobs) to not cause build failures

MiYanni:
This PR turns APIScan back on. It sets the timeout to 3 hours. So far, I've seen APIScan go above 2 hours, and we're working with a dev to investigate why it is taking so long.

We also had RichNav fail recently because of timing out, but that seems to be a separate issue. To avoid non-critical jobs from causing the entire build to fail, I've marked the following jobs with `continueOnError: true` which allows the build pipeline to continue, even when these jobs time out:
- `analyze-api.yml` (APIScan)
- `analyze-vulnerabilities.yml` (CodeQL, but we don't run this yet)
- `publish-richnav.yml` (RichNav)

I tested that this works by setting APIScan temporarily for a 5 minute timeout. You can see here that APIScan stopped at 5 mins but the next stage (VS Insertion) that requires the Compliance stage to pass ran successfully.
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8487)

### DIFF
--- a/eng/pipelines/templates/analyze-api.yml
+++ b/eng/pipelines/templates/analyze-api.yml
@@ -3,7 +3,7 @@
 jobs:
 - job: AnalyzeAPI
   displayName: Analyze API
-  timeoutInMinutes: 5
+  timeoutInMinutes: 180
   # This is a non-critical job, so don't fail the build if it fails.
   continueOnError: true
   steps:
--- a/eng/pipelines/templates/analyze-api.yml
+++ b/eng/pipelines/templates/analyze-api.yml
@@ -3,7 +3,9 @@
 jobs:
 - job: AnalyzeAPI
   displayName: Analyze API
-  timeoutInMinutes: 360
+  timeoutInMinutes: 5
+  # This is a non-critical job, so don't fail the build if it fails.
+  continueOnError: true
   steps:
 
   ###################################################################################################################################################################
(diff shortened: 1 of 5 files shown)


eng/pipelines/templates/analyze-api.yml
```diff
@@ -51,17 +54,14 @@ jobs:
       noCopyBinaries: true
       noDecompress: true
       isLargeApp: true
+      verbosityLevel: minimal
+      preserveTempFiles: true
```
MiYanni, line 58: These were requested by the dev that is analyzing the APIScan run duration issues. They simply add more verbosity to the logs and keep additional files in our build artifacts.

eng/pipelines/templates/publish-assets-and-packages.yml
```diff
@@ -143,7 +143,7 @@ jobs:
       allowPackageConflicts: true
 
   # Publishes the NuGet packages to DevDiv/VS (https://dev.azure.com/DevDiv/DevDiv/_artifacts/feed/VS)
-  # The RoslynInsertionTool will republish these packages to DevDiv/VS-CoreXtFeeds (https://dev.azure.com/DevDiv/DevDiv/_artifacts/feed/VS-CoreXtFeeds)
+  # A separate process will republish these packages to DevDiv/VS-CoreXtFeeds (https://dev.azure.com/DevDiv/DevDiv/_artifacts/feed/VS-CoreXtFeeds)
```
MiYanni, line 146: This is accurate as I don't know the process that is involved, but it is still functioning, independent of us (no longer) using RoslynInsertionTool.

drewnoakes (approved):

PR #8492, opened 2022-09-15
Avoid deadlock between CodeModel API and build data

drewnoakes:
Fixes [AB#1600170](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1600170)

When project system components outside the language service host want to write to the Roslyn project context, we block those operations until the `Workspace` object is initialised with both evaluation and build data.

This state is signalled via a `TaskCompletionSource`, unblocked by the arrival of the first build data for that project slice.

Previously we used `JTF.RunAsync` in an attempt to join work between components that acquire a lock over the `Workspace`. That does not correctly join the two operations. What we need to do is temporarily join the `JoinableTaskCollection` used within the dataflow handling logic. This allows dataflow to keep pumping on the main thread, even if other code is waiting for a write lock.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8492)


drewnoakes:
Thanks to @lifengl and @adrianvmsft for help debugging this issue, and teaching me a few nice tricks along the way.

kvenkatrajan (approved):

PR #8494, opened 2022-09-15
Add property to view .user file command line arguments

ocallesp:
Fixes [AB#1274456](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1274456)

MoveUserFileCommandArgs

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8494)

### DIFF
--- a/docs/repo/property-pages/property-specification.md
+++ b/docs/repo/property-pages/property-specification.md
@@ -71,7 +71,9 @@ The `DataSource` specified here will be applied to all properties, however prope
 - `Persistence` may have several values:
   - `ProjectFile` means that the value will be read and written from the project file directly.
   - `ProjectFileWithInterception` means that a MEF part exists that will handle read/write operations for the property (see below).
+  - `UserFile` means that the value will be read and written from the `.user` file directly.
   - `UserFileWithInterception` is the same as `ProjectFileWithInterception` except we write changes to the project's `.user` file.
+  - `LaunchProfile` means that the value will be read and written from the `launchSettings.json` file directly.
 - `HasConfigurationCondition` controls whether the property is intended to be varied by project configuration (e.g. Debug/Release, platform, target framework...). Setting this to true allows varying property values by configuration dimensions.
 
 ### Categories
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
@@ -1,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Xml;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+using Microsoft.VisualStudio.ProjectSystem.Utilities;
+using Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.Designer;
+using Microsoft.VisualStudio.Threading;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+[Export(typeof(ILinkActionHandler))]
+[ExportMetadata("CommandName", "RemoveStartArguments")]
+internal class RemoveStartArgumentsActionHandler : ILinkActionHandler
+{
+    private const string PropertyCommandArgsInUserFile = "StartArguments";
+    private const string UserSuffix = ".user";
+
+    private string? _xmlFile;
+    private List<string>? userFileCommandArgs;
+
+    public async Task HandleAsync(UnconfiguredProject project, IReadOnlyDictionary<string, string> editorMetadata)
+    {
+        _xmlFile = project.FullPath + UserSuffix;
+
+        await RemoveElementAsync();
+
+        await CopyUserArgsToLaunchProfileAsync(project);
+    }
+
+    private async Task RemoveElementAsync()
+    {
+        await TaskScheduler.Default;
+
+        XmlDocument doc = new() { XmlResolver = null };
+
+        try
+        {
+#pragma warning disable CA3075 // Insecure DTD processing in XML
+            doc.Load(_xmlFile);
+#pragma warning restore CA3075 // Insecure DTD processing in XML
+        }
+        catch (Exception ex)
+        {
+            TraceUtilities.TraceException("Failed to load .user file", ex);
+        }
+
+        XmlNodeList xmlNodeList = doc.GetElementsByTagName(PropertyCommandArgsInUserFile);
+
+        for (int i = xmlNodeList.Count - 1; i >= 0; --i)
+        {
+            userFileCommandArgs ??= new();
+            userFileCommandArgs.Add(xmlNodeList[i].InnerText);
+            xmlNodeList[i].ParentNode.RemoveChild(xmlNodeList[i]);
+        }
+
+        try
+        {
+            doc.Save(_xmlFile);
+        }
+        catch (Exception ex)
+        {
+            TraceUtilities.TraceException("Failed to save .user file", ex);
+        }
+
+        return;
+    }
+
+    private async Task CopyUserArgsToLaunchProfileAsync(UnconfiguredProject project)
+    {
+        ILaunchSettingsProvider? launchSettingsProvider = project.Services.ExportProvider.GetExportedValue<ILaunchSettingsProvider>();
+
+        ILaunchSettings launchSettings = await launchSettingsProvider.WaitForFirstSnapshot();
+
+        IWritableLaunchSettings writableLaunchSettings = launchSettings.ToWritableLaunchSettings();
+
+        if (SetPropertyValue(writableLaunchSettings))
+        {
+            await launchSettingsProvider.UpdateAndSaveSettingsAsync(writableLaunchSettings.ToLaunchSettings());
+        }
+    }
+
+    private bool SetPropertyValue(IWritableLaunchSettings writableLaunchSettings)
+    {
+        if (writableLaunchSettings is null || writableLaunchSettings.ActiveProfile is null || userFileCommandArgs is null)
+        {
+            return false;
+        }
+
+        foreach (string? commandArg in userFileCommandArgs)
+        {
+            writableLaunchSettings.ActiveProfile.CommandLineArgs += " " + commandArg;
+        }
+
+        return true;
+    }
+}
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
@@ -36,6 +36,47 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <StringProperty Name="StartArguments"
+                  DisplayName="Command line arguments (from .user file)"
+                  Description="This project's .user file has a value in the StartArguments property. The use of StartArguments is deprecated, and we recommend removing it via the link below.">
+    <StringProperty.DataSource>
+      <DataSource HasConfigurationCondition="False"
+                  SourceOfDefaultValue="AfterContext"
+                  Persistence="UserFile"/>
+    </StringProperty.DataSource>
+    <StringProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(ne "" (unevaluated "Project" "StartArguments"))</NameValuePair.Value>
+      </NameValuePair>
+      <NameValuePair Name="DependsOn" Value="Project::StartArguments" />
+    </StringProperty.Metadata>
+    <StringProperty.ValueEditors>
+      <ValueEditor EditorType="MultiLineString">
+        <ValueEditor.Metadata>
+          <NameValuePair Name="UseMonospaceFont" Value="True" />
+        </ValueEditor.Metadata>
+      </ValueEditor>
+    </StringProperty.ValueEditors>
+  </StringProperty>
+
+  <StringProperty Name="RemoveStartArguments"
+                  DisplayName="Move the command line args from the .user file to launchSettings.json">
+    <StringProperty.Metadata>
+      <NameValuePair Name="VisibilityCondition">
+        <NameValuePair.Value>(ne "" (unevaluated "Project" "StartArguments"))</NameValuePair.Value>
+      </NameValuePair>
+      <NameValuePair Name="DependsOn" Value="Project::StartArguments" />
+    </StringProperty.Metadata>
+    <StringProperty.ValueEditors>
+      <ValueEditor EditorType="LinkAction">
+        <ValueEditor.Metadata>
+          <NameValuePair Name="Action" Value="Command" />
+          <NameValuePair Name="Command" Value="RemoveStartArguments" />
+        </ValueEditor.Metadata>
+      </ValueEditor>
+    </StringProperty.ValueEditors>
+  </StringProperty>
+  
   <StringProperty Name="WorkingDirectory"
                   DisplayName="Working directory"
                   Description="Path to the working directory where the process will be started."
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ProjectDebugPropertyPage.xaml.cs.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ProjectDebugPropertyPage.xaml.cs.xlf
@@ -102,6 +102,21 @@
         <target state="translated">Název vzdáleného počítače</target>
         <note />
       </trans-unit>
+      <trans-unit id="StringProperty|RemoveStartArguments|DisplayName">
+        <source>Move the command line args from the .user file to launchSettings.json</source>
+        <target state="new">Move the command line args from the .user file to launchSettings.json</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="StringProperty|StartArguments|Description">
+        <source>This project's .user file has a value in the StartArguments property. The use of StartArguments is deprecated, and we recommend removing it via the link below.</source>
+        <target state="new">This project's .user file has a value in the StartArguments property. The use of StartArguments is deprecated, and we recommend removing it via the link below.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="StringProperty|StartArguments|DisplayName">
+        <source>Command line arguments (from .user file)</source>
+        <target state="new">Command line arguments (from .user file)</target>
+        <note />
+      </trans-unit>
       <trans-unit id="StringProperty|WorkingDirectory|Description">
         <source>Path to the working directory where the process will be started.</source>
         <target state="translated">Cesta k pracovnímu adresáři, ve kterém se spustí proces</target>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ProjectDebugPropertyPage.xaml.de.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ProjectDebugPropertyPage.xaml.de.xlf
@@ -102,6 +102,21 @@
         <target state="translated">Name des Remotecomputers</target>
         <note />
       </trans-unit>
+      <trans-unit id="StringProperty|RemoveStartArguments|DisplayName">
+        <source>Move the command line args from the .user file to launchSettings.json</source>
+        <target state="new">Move the command line args from the .user file to launchSettings.json</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="StringProperty|StartArguments|Description">
+        <source>This project's .user file has a value in the StartArguments property. The use of StartArguments is deprecated, and we recommend removing it via the link below.</source>
+        <target state="new">This project's .user file has a value in the StartArguments property. The use of StartArguments is deprecated, and we recommend removing it via the link below.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="StringProperty|StartArguments|DisplayName">
+        <source>Command line arguments (from .user file)</source>
+        <target state="new">Command line arguments (from .user file)</target>
+        <note />
+      </trans-unit>
       <trans-unit id="StringProperty|WorkingDirectory|Description">
         <source>Path to the working directory where the process will be started.</source>
         <target state="translated">Pfad zum Arbeitsverzeichnis, in dem der Prozess gestartet wird.</target>
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ProjectDebugPropertyPage.xaml.es.xlf
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/xlf/ProjectDebugPropertyPage.xaml.es.xlf
@@ -102,6 +102,21 @@
         <target state="translated">Nombre de la máquina remota</target>
         <note />
       </trans-unit>
+      <trans-unit id="StringProperty|RemoveStartArguments|DisplayName">
+        <source>Move the command line args from the .user file to launchSettings.json</source>
+        <target state="new">Move the command line args from the .user file to launchSettings.json</target>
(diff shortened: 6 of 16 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
```diff
@@ -36,6 +36,48 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <StringProperty Name="StartArguments"
+                  DisplayName="ReadOnly .user file command line arguments"
+                  Description="Deprecated. This is only visible if there are command line arguments in the .user file to pass to the executable."
+                  ReadOnly="False">
```
drewnoakes, line 42: Why not `ReadOnly="True"` here?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
```diff
@@ -36,6 +36,48 @@
     </StringProperty.ValueEditors>
   </StringProperty>
 
+  <StringProperty Name="StartArguments"
+                  DisplayName="ReadOnly .user file command line arguments"
+                  Description="Deprecated. This is only visible if there are command line arguments in the .user file to pass to the executable."
```
drewnoakes, line 41:
```suggestion
                  DisplayName="Command line arguments (from .user file)"
                  Description="This project's .user file has a value in the StartArguments property. The use of StartArguments is deprecated, and we recommend removing it via the link below."
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using System.Xml;
+using Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.Designer;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+[Export(typeof(ILinkActionHandler))]
+[ExportMetadata("CommandName", "RemoveStartArguments")]
+internal class RemoveStartArgumentsActionHandler : ILinkActionHandler
```
drewnoakes, line 10: It would be good if this action handler could move or append the value it is removing to the argument list in the `launchSettings.json` too.

docs/repo/property-pages/property-specification.md
```diff
@@ -71,6 +71,7 @@ The `DataSource` specified here will be applied to all properties, however prope
 - `Persistence` may have several values:
   - `ProjectFile` means that the value will be read and written from the project file directly.
   - `ProjectFileWithInterception` means that a MEF part exists that will handle read/write operations for the property (see below).
+  - `UserFile` means that the value will be read and written from  the `.user` file directly.
   - `UserFileWithInterception` is the same as `ProjectFileWithInterception` except we write changes to the project's `.user` file.
```
drewnoakes, line 75: I understand why you asked that question on Teams now :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
ocallesp: changed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
ocallesp: removed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        string xmlFile = project.FullPath + UserSuffix;
+
+        XmlDocument doc = new XmlDocument();
+
+#pragma warning disable CA3075 // Insecure DTD processing in XML
+        doc.Load(xmlFile);
+#pragma warning restore CA3075 // Insecure DTD processing in XML
+
+        XmlNodeList xmlNodeLIst = doc.GetElementsByTagName(PropertyNameToDelete);
```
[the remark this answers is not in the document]
ocallesp: changed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+[Export(typeof(ILinkActionHandler))]
+[ExportMetadata("CommandName", "RemoveStartArguments")]
+internal class RemoveStartArgumentsActionHandler : ILinkActionHandler
+{
+    private const string PropertyNameToDelete = "StartArguments";
+    private const string UserSuffix = ".user";
+
+    public RemoveStartArgumentsActionHandler()
+    {
+    }
```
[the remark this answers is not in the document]
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        XmlDocument doc = new XmlDocument();
+
+#pragma warning disable CA3075 // Insecure DTD processing in XML
+        doc.Load(xmlFile);
+#pragma warning restore CA3075 // Insecure DTD processing in XML
+
+        XmlNodeList xmlNodeLIst = doc.GetElementsByTagName(PropertyNameToDelete);
+
+        XmlNode node = xmlNodeLIst.Item(0);
```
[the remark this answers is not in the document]
ocallesp: done in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+    private const string PropertyNameToDelete = "StartArguments";
+    private const string UserSuffix = ".user";
+
+    public RemoveStartArgumentsActionHandler()
+    {
+    }
+
+    public Task HandleAsync(UnconfiguredProject project, IReadOnlyDictionary<string, string> editorMetadata)
+    {
+        string xmlFile = project.FullPath + UserSuffix;
```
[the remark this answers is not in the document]
ocallesp: fixed in last commit

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        XmlDocument doc = new XmlDocument();
+
+#pragma warning disable CA3075 // Insecure DTD processing in XML
```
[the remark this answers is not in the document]
ocallesp: Fixed. Setting `XmlResolver `to null as suggested in https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca3075#solution-6

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: I'll wait for feedback from users on this proposal.
Users can still manually copy and paste it at the front or at the back.

drewnoakes: Yes, but once they click the hyperlink the value will disappear from screen and there is no undo feature. I would find that very frustrating.

Would it be hard to fix?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    private const string PropertyNameToDelete = "StartArguments";
+    private const string UserSuffix = ".user";
+    private string? _xmlFile;
+
+    public Task HandleAsync(UnconfiguredProject project, IReadOnlyDictionary<string, string> editorMetadata)
+    {
+        _xmlFile = project.FullPath + UserSuffix;
+
+        _ = project.Services.ThreadingPolicy.JoinableTaskFactory.RunAsync(RemoveElementAsync);
```
drewnoakes, line 20: This does not fix the threading issue.

The previous code awaited the work's completion, and did all that work on the UI thread.

This code will still do that work on the UI thread, only it won't await the completion. This means that the user could make changes that have a race condition with the below logic. For example, they could double-click the hyperlink, in which case you might run `RemoveElementAsync` twice.

Instead, I think you should await `RemoveElementAsync` as before, but inside that method switch to a background thread via `await TaskScheduler.Default`.

You shouldn't need `JTF.RunAsync` here as there's no work to join. The caller will have done that for you anyway.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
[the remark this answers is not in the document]
drewnoakes: Why is this `#pragma` still in the code? If you've fixed the issue, can you remove it?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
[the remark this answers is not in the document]
drewnoakes: > In fact we don't even have to wait for that async work to complete. We could just _ = JTF.RunAsync(RemoveElementAsync); and not await that task.

Looking at your change, it's possible I gave you bad advice here. It might be worth making the user wait for this operation to complete, to prevent race conditions with subsequent user actions.

I still think we should switch to the thread pool. See my more recent comment.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
drewnoakes: I don't think this should be removed. I think it should be added.

Currently, this property will be editable. Yet I don't think that's what we want, is it? We want to display the value here, and allow the user to migrate the value from .user to .json.

docs/repo/property-pages/property-specification.md
```diff
@@ -71,7 +71,9 @@ The `DataSource` specified here will be applied to all properties, however prope
 - `Persistence` may have several values:
   - `ProjectFile` means that the value will be read and written from the project file directly.
   - `ProjectFileWithInterception` means that a MEF part exists that will handle read/write operations for the property (see below).
+  - `UserFile` means that the value will be read and written from the `.user` file directly.
   - `UserFileWithInterception` is the same as `ProjectFileWithInterception` except we write changes to the project's `.user` file.
+  - `LaunchProfile` means that value will be read and written from the `launchSettings.json` file directly.
```
drewnoakes, line 76:
```suggestion
  - `LaunchProfile` means that the value will be read and written from the `launchSettings.json` file directly.
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        return Task.CompletedTask;
+    }
+
+    private Task RemoveElementAsync()
+    {
+        XmlDocument doc = new XmlDocument() { XmlResolver = null };
+
+#pragma warning disable CA3075 // Insecure DTD processing in XML
+        doc.Load(_xmlFile);
```
drewnoakes, line 30: File IO and parsing can throw exceptions. For example, if the file is missing (you can never assume it's present until you've successfully opened it), locked by some other process, or if it contains invalid XML.

You should handle exceptions here. Ditto below on `Save`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+        _xmlFile = project.FullPath + UserSuffix;
+
+        _ = project.Services.ThreadingPolicy.JoinableTaskFactory.RunAsync(RemoveElementAsync);
+
+        return Task.CompletedTask;
+    }
+
+    private Task RemoveElementAsync()
+    {
+        XmlDocument doc = new XmlDocument() { XmlResolver = null };
```
drewnoakes, line 27: You can avoid the repeated type name here if you like:

```suggestion
        XmlDocument doc = new() { XmlResolver = null };
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+[ExportMetadata("CommandName", "RemoveStartArguments")]
+internal class RemoveStartArgumentsActionHandler : ILinkActionHandler
+{
+    private const string PropertyNameToDelete = "StartArguments";
+    private const string UserSuffix = ".user";
+    private string? _xmlFile;
+
+    public Task HandleAsync(UnconfiguredProject project, IReadOnlyDictionary<string, string> editorMetadata)
+    {
+        _xmlFile = project.FullPath + UserSuffix;
```
drewnoakes, line 18: Please check with the CPS team about whether it would be better to use CPS APIs to modify the .user file, rather than doing direct XML-to-disk changes here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: done

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: fixed with TaskScheduler.Default

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: fixed

docs/repo/property-pages/property-specification.md
ocallesp: fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/ProjectDebugPropertyPage.xaml
ocallesp: Even if we set this ReadOnly="True", cps will mark it as read only because it thinks this is a common property.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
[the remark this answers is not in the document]
ocallesp: For some reason the static analysis still believes this is unsafe.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: The cps apis marked this property as read only and doesn't allow to make changes to it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: exception handled

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
tmeschter: The .user file is part of the loaded project; in general we shouldn't modify directly.

Do we understand why "StartArguments" is being treated as read-only?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            TraceUtilities.TraceException("Failed to save .user file", ex);
+        }
+
+        return;
+    }
+
+    private async Task CopyUserArgsToLaunchProfileAsync(UnconfiguredProject project)
+    {
+        ILaunchSettingsProvider? launchSettingsProvider = project.Services.ExportProvider.GetExportedValue<ILaunchSettingsProvider>();
```
tmeschter, line 70: Any reason to do this rather than import the `ILaunchSettingsProvider` in a constructor?

tmeschter: Among other things it would make it easier to write unit tests for this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+
+[Export(typeof(ILinkActionHandler))]
+[ExportMetadata("CommandName", "RemoveStartArguments")]
+internal class RemoveStartArgumentsActionHandler : ILinkActionHandler
+{
+    private const string PropertyCommandArgsInUserFile = "StartArguments";
+    private const string UserSuffix = ".user";
+
+    private string? _xmlFile;
+    private List<string>? userFileCommandArgs;
```
tmeschter, line 19: `_userFileCommandArgs`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+    private bool SetPropertyValue(IWritableLaunchSettings writableLaunchSettings)
+    {
+        if (writableLaunchSettings is null || writableLaunchSettings.ActiveProfile is null || userFileCommandArgs is null)
+        {
+            return false;
+        }
+
+        foreach (string? commandArg in userFileCommandArgs)
+        {
+            writableLaunchSettings.ActiveProfile.CommandLineArgs += " " + commandArg;
```
tmeschter, line 91: You need to account for:
1. `CommandLineArgs` may be `null`.
2. `CommandLineArgs` may be the empty string.

tmeschter, line 91: Also, this is going to move the `StartArguments` to the command line arguments of the _active_ profile. I see two _potential_ problems:
1. The user may not be editing the active profile. In that case the arguments will just seem to disappear when they click the link.
2. `StartArguments` potentially affect launch/debug regardless of the launch profile, but this will move them to one specific launch profile.

I'm not sure either of those is really a problem, but we should think about those scenarios.

tmeschter (changes requested):
1. We need to understand why the property is being treated as read-only.
2. Don't modify the .user file directly.
3. Please add unit tests.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/DebugPropertyPage/RemoveStartArgumentsActionHandler.cs
ocallesp: I already tried importing `ILaunchSettingsProvider` in the constructor, but I got a lot of composition errors.

ocallesp:
Converting this pr to draft since there is still unclear why `StartArguments` is treated as read-only in cps. Perhaps there is a bug in cps.

PR #8498, opened 2022-09-16
Don't delay workspace writers as long

drewnoakes:
In the prior language service host implementation (replaced a few weeks ago) attempts to lock the IWorkspaceProjectContext would block only on the creation of the context object, not on its population with build data.

Build data can arrive quite a bit later. We do not need to wait that long.

This change moves the signal that triggers this condition earlier, matching the previous behavior.

This fixes a potential UI delay for code paths that access certain APIs such as CodeModel or TempPE.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8498)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -529,19 +529,19 @@ public async Task<T> WriteAsync<T>(Func<IWorkspace, Task<T>> action, Cancellatio
...
-    private async Task WhenInitialized(CancellationToken cancellationToken)
+    private async Task WhenContextCreated(CancellationToken cancellationToken)
     {
         Verify.NotDisposed(this);
 
         await _joinableTaskFactory.RunAsync(async () =>
         {
-            // We only have build data if we also have evaluation data, so this implies both have been integrated.
-            await _hasBuildData.Task.WithCancellation(cancellationToken);
+            // Ensure we have received enough data to create the context.
```
davkean, line 543: This will hang if the data that will eventually cause this TaskCompletionSource to be set needs the UI thread.

drewnoakes: That was fixed in https://github.com/dotnet/project-system/pull/8492

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
drewnoakes: Another similar fix https://github.com/dotnet/project-system/pull/8501

PR #8499, opened 2022-09-17
Use newer Roslyn APIs in the language service

drewnoakes:
@tmat and I designed a new API for the interaction between the .NET Project System and Roslyn when creating instances of `IWorkspaceProjectContext`.

This new API:

- Allows Roslyn to (generally) request additional properties from the project system, without requiring a corresponding change in the project system.

- Avoids having to create a wasteful batched update immediately after constructing the workspace.

This PR updates the API we use.

Bumping that package pulled the thread on a whole bunch of other package upgrades, including analyzer changes that lead to `.editorconfig` changes, some naming guideline changes, and some suppressions.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8499)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/FileMoveNotificationListener.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/FileMoveNotificationListener.cs
@@ -137,7 +137,9 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Rename
 
                     // This is a file item to another directory, it should only detect this a Update Namespace action.
                     // TODO Upgrade this api to get rid of the exclamation sign
+#pragma warning disable CS0618 // Type or member is obsolete https://github.com/dotnet/project-system/issues/8591
                     Renamer.RenameDocumentActionSet documentAction = await Renamer.RenameDocumentAsync(oldDocument, null!, documentFolders);
+#pragma warning restore CS0618 // Type or member is obsolete
 
                     if (documentAction.ApplicableActions.IsEmpty ||
                         documentAction.ApplicableActions.Any(a => !a.GetErrors().IsEmpty))
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
@@ -174,12 +174,24 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
                     _disposableBag.Add(_buildProgressRegistration);
                 }
 
-                await (update.Value switch
+                try
                 {
-                    { EvaluationUpdate: not null } => OnEvaluationUpdateAsync(update.Derive(u => u.EvaluationUpdate!)),
-                    { BuildUpdate: not null } => OnBuildUpdateAsync(update.Derive(u => u.BuildUpdate!)),
-                    _ => throw Assumes.NotReachable()
-                });
+                    await (update.Value switch
+                    {
+                        { EvaluationUpdate: not null } => OnEvaluationUpdateAsync(update.Derive(u => u.EvaluationUpdate!)),
+                        { BuildUpdate: not null } => OnBuildUpdateAsync(update.Derive(u => u.BuildUpdate!)),
+                        _ => throw Assumes.NotReachable()
+                    });
+                }
+                catch
+                {
+                    // Tear down on any exception
+                    await DisposeAsync();
+
+                    // Exceptions here are product errors, so let the exception escape in order
+                    // to produce an upstream NFE.
+                    throw;
+                }
             });
     }
 
@@ -217,24 +229,23 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
             IProjectRuleSnapshot snapshot = evaluationUpdate.Value.EvaluationRuleUpdate.CurrentState[ConfigurationGeneral.SchemaName];
 
             snapshot.Properties.TryGetValue(ConfigurationGeneral.LanguageServiceNameProperty, out string? languageName);
-            snapshot.Properties.TryGetValue(ConfigurationGeneral.TargetPathProperty, out string? binOutputPath);
             snapshot.Properties.TryGetValue(ConfigurationGeneral.MSBuildProjectFullPathProperty, out string? projectFilePath);
-            snapshot.Properties.TryGetValue(ConfigurationGeneral.AssemblyNameProperty, out string? assemblyName);
-            snapshot.Properties.TryGetValue(ConfigurationGeneral.CommandLineArgsForDesignTimeEvaluationProperty, out string? commandLineArgsForDesignTimeEvaluation);
 
-            if (string.IsNullOrEmpty(languageName) || string.IsNullOrEmpty(binOutputPath) || string.IsNullOrEmpty(projectFilePath))
+            if (string.IsNullOrEmpty(languageName) || string.IsNullOrEmpty(projectFilePath))
             {
                 // It's reasonable for a project to exist that doesn't contain enough data to initialize the language
                 // service. For example, the ".ilproj" project is not uncommon and hits this code path.
                 // In such cases, we don't want to throw here as that would fault the project. Instead, we
                 // dispose the workspace tears down the dataflow subscription and clean everything up gracefully.
-                Dispose();
+                await DisposeAsync();
 
                 return;
             }
 
             try
             {
+                EvaluationDataAdapter evaluationData = new(snapshot.Properties);
+
                 _contextId = GetWorkspaceProjectContextId(projectFilePath, _projectGuid, _slice);
 
                 _disposableBag.Add(_activeEditorContextTracker.RegisterContext(_contextId));
@@ -243,36 +254,16 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
 
                 // Call into Roslyn to initialize language service for this project
                 _context = await _workspaceProjectContextFactory.Value.CreateProjectContextAsync(
-                    languageName,
-                    _contextId,
-                    projectFilePath,
                     _projectGuid,
+                    _contextId,
+                    languageName,
+                    evaluationData,
                     hostObject,
-                    binOutputPath,
-                    assemblyName,
                     cancellationToken);
 
-                _disposableBag.Add(_context);
-
-                // Update additional properties within a batch to avoid thread pool starvation.
-                // https://github.com/dotnet/project-system/issues/8027
-                _context.StartBatch();
-
-                try
-                {
-                    _context.LastDesignTimeBuildSucceeded = false; // By default, turn off diagnostics until the first design time build succeeds for this project.
-
-                    // Pass along any early approximation we have of the command line options
-#pragma warning disable CS0618 // This was obsoleted in favor of the one that takes an array, but here just the string is easier; we'll un-Obsolete this API
-                    _context.SetOptions(commandLineArgsForDesignTimeEvaluation ?? "");
-#pragma warning restore CS0618 // Type or member is obsolete
-                }
-                finally
-                {
-                    await _context.EndBatchAsync();
-                }
-
                 _contextCreated.TrySetResult();
+
+                _disposableBag.Add(_context);
             }
             catch (Exception ex)
             {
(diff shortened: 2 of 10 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -211,12 +211,9 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
             IProjectRuleSnapshot snapshot = evaluationUpdate.Value.EvaluationRuleUpdate.CurrentState[ConfigurationGeneral.SchemaName];
 
             snapshot.Properties.TryGetValue(ConfigurationGeneral.LanguageServiceNameProperty, out string? languageName);
-            snapshot.Properties.TryGetValue(ConfigurationGeneral.TargetPathProperty, out string? binOutputPath);
             snapshot.Properties.TryGetValue(ConfigurationGeneral.MSBuildProjectFullPathProperty, out string? projectFilePath);
-            snapshot.Properties.TryGetValue(ConfigurationGeneral.AssemblyNameProperty, out string? assemblyName);
-            snapshot.Properties.TryGetValue(ConfigurationGeneral.CommandLineArgsForDesignTimeEvaluationProperty, out string? commandLineArgsForDesignTimeEvaluation);
 
-            if (string.IsNullOrEmpty(languageName) || string.IsNullOrEmpty(binOutputPath) || string.IsNullOrEmpty(projectFilePath))
+            if (string.IsNullOrEmpty(languageName) || string.IsNullOrEmpty(projectFilePath))
```
drewnoakes, line 216: We no longer pull these properties out on our side. Roslyn will obtain them via an object we provide.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -226,6 +223,8 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
 
             try
             {
+                EvaluationDataAdapter evaluationData = new(snapshot.Properties);
```
drewnoakes, line 226: This object allows Roslyn to call us back for any property that's in ConfigurationGeneral. In future we may also allow requesting arbitrary project properties.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -234,34 +233,14 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
...
-                    _contextId,
-                    projectFilePath,
                     _projectGuid,
+                    _contextId,
+                    languageName,
+                    evaluationData,
                     hostObject,
-                    binOutputPath,
-                    assemblyName,
                     cancellationToken);
```
drewnoakes, line 241: Calling the new overload.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -234,34 +233,14 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
...
-
-                    // Pass along any early approximation we have of the command line options
-#pragma warning disable CS0618 // This was obsoleted in favor of the one that takes an array, but here just the string is easier; we'll un-Obsolete this API
-                    _context.SetOptions(commandLineArgsForDesignTimeEvaluation ?? "");
-#pragma warning restore CS0618 // Type or member is obsolete
-                }
-                finally
-                {
-                    await _context.EndBatchAsync();
-                }
```
drewnoakes, line 264: This whole post-construction update is no longer needed.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/Editors/ImplicitUsingsEditor/CSharpImplicitUsingsControl.cs
```diff
@@ -55,7 +55,7 @@ internal sealed class CSharpImplicitUsingsControl : Control
             new PropertyMetadata(true, (o, e) => ((CSharpImplicitUsingsControl)o).OnStringListOrSupportedValuesPropertyChanged()));
         
         // Used to suppress event handling during our own updates, breaking infinite loops.
-        private bool updating;
+        private bool _updating;
```
drewnoakes, line 58: Code that was recently moved from CPS uses a different naming format. One of the analyzer updates started issuing warnings for these, so they've been updated as part of this PR.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/FileMoveNotificationListener.cs
```diff
@@ -137,7 +137,9 @@ async Task<List<(Renamer.RenameDocumentActionSet, string)>> GetNamespaceUpdateAc
 
                     // This is a file item to another directory, it should only detect this a Update Namespace action.
                     // TODO Upgrade this api to get rid of the exclamation sign
+#pragma warning disable CS0618 // Type or member is obsolete
                     Renamer.RenameDocumentActionSet documentAction = await Renamer.RenameDocumentAsync(oldDocument, null!, documentFolders);
+#pragma warning restore CS0618 // Type or member is obsolete
```
drewnoakes, line 142: Roslyn has deprecated a bunch of these rename APIs.

@ocallesp do we have a work item tracking the update of this? If not, could you file one please?

drewnoakes:
I struggled making the package references and versions work here. It builds on my machine, though I half expect it to fail in CI. I'm not convinced the approach here is right, and it seems like some of the packages we need are incorrectly authored or missing from the feeds we source them from. I will revisit this next soon.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -234,34 +233,14 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
...
             }
             catch (Exception ex)
```
tmat, line 245: >Exception [](http://example.com/codeflow?start=19&length=9)

Roslyn throws `InvalidProjectPropertyValueException` if a property that is required isn't present or has invalid value, in case you want to handle it in a different way.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/Editors/PropertyEditorBase.cs
```diff
@@ -13,20 +13,20 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Implementation.PropertyPages.D
     /// </summary>
     internal abstract class PropertyEditorBase : IPropertyEditor
     {
-        private static readonly Lazy<ResourceDictionary> LazyResources;
-        private static readonly Lazy<DataTemplate> LazyGenericPropertyTemplate;
+        private static readonly Lazy<ResourceDictionary> s_lazyResources;
```
jacdavis, line 16: >s_lazyResources [](http://example.com/codeflow?start=57&length=15)

This is all just naming convention changes in this file, correct?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -552,4 +545,28 @@ private async Task WhenInitialized(CancellationToken cancellationToken)
...
+        {
+            _properties = properties;
+        }
+
+        public override string GetPropertyValue(string name)
+        {
+            _properties.TryGetValue(name, out string? value);
+
+            // Return the empty string rather than null.
+            value ??= "";
```
jacdavis, line 567: >"" [](http://example.com/codeflow?start=22&length=2)

Unless something has changed in the compiler, it is preferable to use String.Empty here to avoid allocating another string object for "".

jacdavis (approved):
:shipit:

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PropertyPages/Editors/PropertyEditorBase.cs
drewnoakes: Yes, just naming conventions.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
drewnoakes: Thankfully the compiler hasn't done that for a long time.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -279,6 +249,29 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
...
+                snapshot.Properties.TryGetValue(ConfigurationGeneral.LanguageServiceNameProperty, out string? languageName);
+                snapshot.Properties.TryGetValue(ConfigurationGeneral.MSBuildProjectFullPathProperty, out string? projectFilePath);
+
+                return (string.IsNullOrEmpty(languageName), string.IsNullOrEmpty(projectFilePath)) switch
+                {
+                    (true, false) => throw Fail("Missing language name."),
+                    (false, true) => throw Fail("Missing project file path."),
+                    (true, true) => throw Fail("Missing language name and project file path"),
+                    (false, false) => (languageName, projectFilePath)
+                };
```
MiYanni, line 264: That's the most interesting way I've seen that done. Works well for 2 entities to check. If it were 3 or more, it would start to get a bit goofy. 😝

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -279,6 +249,29 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
...
+
+                Exception Fail(string reason)
+                {
+                    // Insufficient data to initialize the language service.
+                    _state = WorkspaceState.Failed;
+
+                    // Throw, so we fault and tear down the dataflow.
+                    throw new($"Insufficient project data to initialize the language service. {reason}");
```
MiYanni, line 272: Did you mean to throw the exception from here? This is throwing and not returning the exception to where you are also throwing in the switch expression above. So, the throw in the switch expression is only making the compiler happy since it'll never be hit at runtime.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
drewnoakes: Yes this is intentional, but a good catch all the same! The reason is that each fork of the switch expression has to return a value, or throw. 

It might be possible to annotate this method with `[DoesNotReturn]` though I'm not sure whether that'll appease the compiler's control flow analysis here.

Regardless, the pattern used here is not uncommon. It's used in vs-validation for example, such as [here](https://github.com/microsoft/vs-validation/blob/3caa8c90b5940929b73862ac49dd8a514bdab96f/src/Microsoft.VisualStudio.Validation/Verify.cs#L80-L93) and [here](https://github.com/microsoft/vs-validation/blob/3caa8c90b5940929b73862ac49dd8a514bdab96f/src/Microsoft.VisualStudio.Validation/Requires.cs#L489-L520).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Rename/FileMoveNotificationListener.cs
drewnoakes: I filed https://github.com/dotnet/project-system/issues/8591

Issue #8423, opened 2022-08-25
Add VS insertion detailed description


MiYanni:
As per the changes in [this PR](https://github.com/dotnet/project-system/pull/8420), we moved to using MicroBuild to create our VS insertion PRs. MicroBuild does not provide detailed descriptions for their VS insertion PRs. To provide the same convince that we had prior, we'll need to add this mechanism to our repo.
- Isolate the code from [Roslyn Insertion Tool](https://github.com/dotnet/roslyn-tools/tree/main/src/RoslynInsertionTool) that created this description
- Create a small, single purpose tool in our repo that simply spits out this description string
- Assign description string to a variable in the pipeline
- Assign the variable to the `InsertionDescription` value in the MicroBuild Insertion task

JoeRobich:
Here is the script I ran prior to migrating the code into RIT - https://gist.github.com/JoeRobich/5b02dbe9b9d60b4699df546036040888

MiYanni:
@JoeRobich That's really useful as it uses the local repo on-disk to get the commit information. RIT does a REST call to GitHub to get this information, which is completely unnecessary in our situation. I'll be combining bits of each to make something that produces a similar result.

Issue #8431, opened 2022-08-27
Update pipeline artifact naming/build numbering for official.yml


MiYanni:
### Description/Concerns
Currently, we use `$(Build.BuildNumber)` (for the primary build artifact) which is the value of that particular pipeline run and doesn't relate to our build (assemblies/packages) directly. You can [see here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/run-number?view=azure-devops&tabs=yaml) that the default format is: `$(Date:yyyyMMdd).$(Rev:r)`. For example, you'd get a value like `20220826.7`. Also note that you can change this value by setting `name:` on the pipeline. Our previous reliance on RoslynInsertionTool required it to [use the default `Build.BuildNumber` as the artifact name](https://github.com/dotnet/roslyn-tools/blob/1f64f56b4ea9a0c119bbe840c39b935acc60e4b8/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs#L174). Since we no longer use that tool, we have some freedom in what we name the artifact.

I've previously tried changing `Build.BuildNumber` to `GitBuildVersion` as part of the [mechanism in NerdBank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/cloudbuild.md#automatically-match-cloud-build-numbers-to-to-your-git-version). Since this was problematic with RoslynInsertionTool, I removed that change. We can consider changing this again, but my next point below would need a custom solution *specifically* for the artifact name.

There is another problem where we cannot currently retry certain jobs in the build. Why? Because the artifacts once published [are immutable](https://developercommunity.visualstudio.com/t/publish-artifact-fails-when-retry-failed-job/774908). This means we cannot republish the artifacts with the same name in the same pipeline run. The reason we even publish the artifacts on a failed job is so that someone can debug the failure (download the artifacts and investigate). So, if we want to allow rerunning failed jobs (which saves time over rerunning the pipeline entirely), we would need a *rolling name for all artifacts*. This name needs to be accessible by all the jobs in the pipeline, or else jobs cannot access artifacts from previous jobs. For example, the artifacts from the Build job are downloaded by the Publish jobs. Some solutions [like this one](https://github.com/microsoft/azure-pipelines-tasks/issues/12854#issuecomment-645707122) may be used but needs to be tested. When retrying occurs currently, we get an error such as:
> ##[error]Artifact 20220812.10 already exists for build 6553163.

[...]

PR #8517, opened 2022-09-23
Create detailed VS insertion description

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/8423
Fixes: https://github.com/dotnet/project-system/issues/6453

## Summary
Prior to [this PR](https://github.com/dotnet/project-system/pull/8420), we were using RIT (Roslyn Insertion Tool) to create VS insertion PRs. This tool provided a detailed description in the VS insertion PR that let us (primarily) know what PRs were merged from our repo. Example: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/417674
image

After removing the use of RIT, we no longer have detailed descriptions. Example: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/422741
image

This PR reintroduces detailed PR descriptions with similar(-ish) level of detail.

## Details
While this description content is seemingly straightforward on the surface, in practice, it is not. The way this worked previously was that RIT was creating the entire VS insertion PR, so it had access to the changes that were taking place. Meaning, it knew which VS files were changing. It would check out these files from the VS repo and read a specific value out of `components.json`. Example:
image

As you can see, only a **particular number** is changing. This number was pivotal to how RIT created the detailed description. You can read more about that number here: https://github.com/dotnet/project-system/issues/8431 This number correlates to our BuildNumber in our `official.yml` pipeline. [RIT used this number](https://github.com/dotnet/roslyn-tools/blob/1f64f56b4ea9a0c119bbe840c39b935acc60e4b8/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs#L174) to read information about the build (using Azure Pipelines APIs). From the build, it would gather the commit ID from our repo that relates to that build. Therefore, it has access to the commit ID in our repo that relates to our last insertion. Visually,
```mermaid
graph LR;
    RIT("RIT (Roslyn Insertion Tool)")-->CJSON;
    CJSON(VS repo `components.json`)-->BN;
    BN(`official.yml` BuildNumber)-->AP;
    AP(Azure Pipelines API)-->CI(project-system CommitID);
```

This worked fine considering RIT had all the tools needed to operate in this way. Our current insertion process, however, is handled by a [MicroBuild insertion task](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/629/Automated-VS-Insertion). This task provides the basic mechanisms and options for creating a VS insertion PR. Our repo and tooling has no involvement in this process other than providing the parameters to the task.

In our `official.yml` pipeline, we don't access the VS repo in any way. We also don't have access to the service principal used to perform the creation of the VS insertion PR, since that it handled by the MicroBuild insertion task. The only option we have is providing a parameter named `InsertionDescription` to the MicroBuild insertion task. Therefore, I need to create the description prior to starting insertion PR creation.

At that point in the build pipeline, I have access to the current commit ID via `Build.SourceVersion`. However, I needed a way to get the commit ID of the previous VS insertion we did. The list of PRs that occurred between those 2 commits can be (somewhat) easily determined once you have those 2 values. Since this pipeline checks out our code, we can get access to the Git log of commits that happened between those 2 commit IDs.

### Previous Insertion Commit ID Design
This is where the fun begins. I went through various ideas on how to get this value. I decided that I wanted to *use what was available to me* at that point in the pipeline. I didn't want to have to go through the extra work of managing a service principal and doing queries to the AzDO APIs. The best answer I came up with was using [Git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging).

Git tags are basically a name applied to a certain commit and they are branch independent. There was an issue (see top of this description) that related to having tags for releases of our repo. In our workflow, a 'release' closely resembles when we merge our code into the VS repo; so if I could tag when VS insertions occur, I'd effectively tag when we 'release' and be able to use this for creating the VS insertion PR description.

After doing some research, I determined there is a way to [trigger a pipeline based on another repository](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops#triggers). Therefore, I could create a set of criteria for this trigger to activate when we *merge an insertion PR into the VS repo*. This turned out to be a bit more... odd than I was envisioning.

To make this work, I'd need a **new pipeline** that would specifically do the action of tagging our repo. The trigger for this pipeline, named `tagger.yml`, is based on *specific files being updated in a commit to VS repo's `main`*. I originally wanted to trigger based on a pattern in the branch name (which the MicroBuild insertion task follows), however, the triggers only work on the *destination branch name*, not the source.

The files selected use the `path:` filter mechanism for the pipeline trigger, and the trigger is within a `repository:` node within the pipeline. This node links to the VS repo, and the path is set to these values:
- `src/ConfigData/Packages/ProjectSystem/Managed.props`
- `src/ConfigData/Packages/Wizard/packages.props`

These files are where we update the VSIX package versions for Managed/Managed.VS (former) and AppDesigner/Editors (latter). We are the only source of changes to these files as they only contain references to our packages. Therefore, any changes to these files will come from us.

Unfortunately, I have not been able to fully test this trigger system as it requires the `self` repo (ours) to have this pipeline in the **default branch** according to [the documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops#triggers):
> If a change to any other repository resource triggers the pipeline, then the latest version of YAML from the **default branch** of `self` repository is used.

Because of this limitation, I have yet to test to see that the trigger system works (but I'm hopeful! 🤞). With this mechanism in place, I needed a way to get the *commit ID* from the VS insertion commit. This `tagger.yml` pipeline is aware of which commit ID it was triggered from. Since the VS repo is triggering this pipeline, the commit ID would be stored in `Build.SourceVersion` according to [the documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops#triggers):
> When an update to one of the repositories triggers a pipeline, then the following variables are set based on triggering repository:
> - `Build.Repository.ID`
> - `Build.Repository.Name`
> - `Build.Repository.Provider`
> - `Build.Repository.Uri`
> - `Build.SourceBranch`
> - `Build.SourceBranchName`
> - `Build.SourceVersion`
> - `Build.SourceVersionMessage`

However, having the VS commit ID alone would not let me know what commit ID in *our repo* was contained within that insertion PR. I needed the VS insertion PR's commit to contain the commit ID from our repo. The solution was to put a short (10 characters) commit ID into the VS insertion PR title.

The PR title is the *only piece of data* we can set in the MicroBuild insertion task that will be available in the commit itself. Having this data in the commit itself means that we don't need to do any API calls to AzDO to gather information, since we have access to the commit via Git checkout in the `tagger.yml` pipeline.

Here is how the VS insertion PR title (left) becomes the VS commit title (right) in Azure DevOps:
```mermaid
graph LR;
    PR["{PRTitle}"]-->CI("Merged PR {PRNumber}: {PRTitle}");
```

Therefore, anything we put in the PR title will be present in the commit title. In the MicroBuild insertion task, we cannot set the entire PR title. However, we can set *a portion of it* via `InsertionPayloadName`. That value is inserted into the title as such:
> Insert `{InsertionPayloadName}` Payload into `{VSBranchName}`

I updated the `InsertionPayloadName` to contain this information:
> DotNet-Project-System (`{PSBranchName}`:`{PSBuildPipelineNumber}`:`{PSShortCommitId}`)

With this information, I can extract the `ShortCommitId` out of the VS commit title (using regex). In the `tagger.yml` pipeline, `CreateTagFromVSCommitTitle.ps1` takes this commit ID and then creates a tag in our repo using the format `insertion/*` where `*` is either `PR-{PRNumber}` or `{VSCommitId}` (in the rare case the commit is not a merged PR).

Now that we're able to tag insertions to commits in our repo, we come back to the actual `official.yml` pipeline creating the insertion description. When Azure Pipelines does a checkout, it includes *all the tags* from the repository. In the `GetInsertionDescription.ps1` script, we check for the latest tag that starts with `insertion/*` and use that as our prior insertion commit ID. Thus, we have 2 commit IDs and use the Git Log command to get all the commits between those 2 IDs. We do some parsing/cleanup of the log statements and then filter the log statements based on only those that are PRs to our repo. PRs have the title format: `Merge pull request #{PRNumber} from`. Using only these commits, we construct a single string to send to `InsertionDescription` for the MicroBuild insertion task.

Constructing a single string for this purpose is complicated. The original RIT description was able to send multi-line text as the insertion description. The MicroBuild insertion task is [limited to a single line string](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/powershell/VstsTaskSdk/ToolFunctions.ps1#L84-L86) because of the library it uses to interact with AzDO. The reason we want multi-line text is because the text is rendered as Markdown. Many Markdown elements (lists, horizontal rules, etc.) require newlines to render. However, we cannot use newlines.

The only option (I tested several) was to use `<br>` to put information onto multiple lines. This ends up creating readable descriptions, but most of it isn't rendered as Markdown. Fortunately, links do not require newlines, so all the links in the description render properly. See the [Examples](#Examples) section to see how it is rendered. The description string is passed to the MicroBuild insertion task via the `InsertionDescription` variable, which completes the design.

```mermaid
flowchart LR
    MI[MicroBuild Insertion Task] -- Creates PR<br>CommitId via PR Title --> IPR[VS Insertion PR]
    IPR -- PR Merged<br>CommitId via Commit Title --> VSC[VS Commit]
```

> `tagger.yml`
```mermaid
flowchart LR
    VSC[VS Commit] -- Triggers Pipeline<br>Specific Files in Commit --> TP[Tagger Pipeline]
    TP -- Creates Tag<br>CommitId via Commit Title --> PSR[Project-System Tag: insertion/*]
```

> Current insertion: `official.yml`
```mermaid
flowchart LR
    DPS[Description Script] -- Finds Latest Tag: insertion/*<br>Checkout Contains All Tags --> DS[Description String]
    DS -- Contains Information from...<br>PRs between Tag and Current --> MI
    MI[MicroBuild Insertion Task] -- Creates PR<br>Description String via Variable --> IPR[VS Insertion PR]
```

### Lost Functionality
The only thing that could not be carried over with this design were the links of the builds (of `official.yml`) of the previous and current commits.
image

These links require the `BuildId` of those pipeline runs. In RIT, it used the `BuildNumber` (the value being shown, such as `20220806.7`) to query AzDO for the `BuildId` (a single integer value). In that example, the `BuildId` is `6525200` which is a query parameter in the link.

The only way to reserve this same looking and functioning link with the new design would have been to pass both the `BuildId` and `BuildNumber` via the VS Insertion PR title. At that point, a majority of the title is simply used for passing data, and I wanted the title to at least (mostly) remain human readable. 😝

One restriction to keep in mind is this process only **fully** works for insertions into VS `main`. Multiple insertions into other VS branches will not display a correct PR list since the tags will not be updated on in our repo based on those insertions.

## Examples
- Example where the previous insertion tag was found: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/424200?_a=overview
image

- Example where the previous insertion tag was **not** found: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/424166?_a=overview
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8517)


eng/pipelines/tagger.yml
```diff
@@ -0,0 +1,90 @@
...
+# See: https://docs.microsoft.com/azure/devops/pipelines/yaml-schema/pr?view=azure-pipelines#examples
+pr: none
+
+resources:
+  # https://stackoverflow.com/a/63270937/294804
+  repositories:
+  - repository: self
+    name: project-system
+    type: git
+    ref: feature/InsertionPRDescriptionAndTagging
```
MiYanni, line 26: I need to remove this section as it was something I was trying to test with the trigger mechanism.

eng/pipelines/tagger.yml
```diff
@@ -0,0 +1,90 @@
...
+  # - https://docs.opensource.microsoft.com/tools/nuget_security_analysis/nuget_security_analysis/
+  # - https://docs.opensource.microsoft.com/tools/cg/how-to/nuget-multifeed-configuration/
+  # - https://onebranch.visualstudio.com/OneBranch/_wiki/wikis/OneBranch.wiki/5205/TSG-Build-Broken-Due-to-Using-Multiple-Feeds?anchor=setting-nugetsecurityanalysiswarninglevel-in-cdp
+  NugetSecurityAnalysisWarningLevel: none
+
+parameters:
+- name: VsCommitId
+  displayName: VS Commit ID
+  type: string
+  default: $(Build.SourceVersion)
```
MiYanni, line 65: This might not work and needs tested.

eng/pipelines/tagger.yml
```diff
@@ -0,0 +1,90 @@
...
+  # Changes Build.SourcesDirectory from $(Pipeline.Workspace)/s to $(Pipeline.Workspace)/project-system
+  path: project-system
+  # Hard-coded assumption that the commit referenced by the VS insertion is within the last 100 commits in the repo.
+  fetchDepth: 100
+# Create the VS folder and clone the VS repo into it, excluding the files themselves (only .git history is required).
+# Doing this manual checkout over the checkout: VS task saves around 3-4 minutes.
+# The System.AccessToken is required for the clone to occur. See: https://stackoverflow.com/a/56734304/294804
+- powershell: |
+    $null = New-Item -Path '$(Pipeline.Workspace)' -Name VS -ItemType Directory
+    git clone --no-checkout --depth ${{ parameters.VsCheckoutDepth }} https://$(System.AccessToken)@dev.azure.com/devdiv/DevDiv/_git/VS $(Pipeline.Workspace)/VS
```
MiYanni, line 87: Need to determine better way to check out if depth 1 does not contain the commit that triggered this pipeline (multiple simultaneous insertion PR merges). Can use Build.SourceVersion in some way here.

eng/pipelines/tagger.yml
MiYanni: This has been removed.

eng/pipelines/tagger.yml
MiYanni: Replaced this mechanism with a 'Default' value that conditionally sets a variable to `Build.SourceVersion`. There is no reason for me to believe this won't work when triggered by a VS insertion commit.

eng/pipelines/tagger.yml
MiYanni: Fixed this by only doing a `git fetch` of the specific commit, which also won't checkout the code. Then, making sure the vsCommitId is used in the script itself, as it is required since there is no branch associated with this folder.

melytc (approved):
LGTM! Thanks for the detailed description ❤️

PR #8520, opened 2022-09-23
Possible fix for another manifestation of #5418.

walterpg:
Details in the [#5418 comment](https://github.com/dotnet/project-system/issues/5418#issuecomment-1256656158).

As implied in that post, this may not be a "proper fix" in the larger context of project-system, etc., but it fixes a problem we are seeing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8520)


MiYanni:
@drewnoakes Seems like you'd have the most context on this as you created the issue 3 years ago. 😝

walterpg:
@drewnoakes We have a reliable repro, just not one that we can share yet.

A few days ago, we found that our extension's involvement with the Design Time Build induces the behavior requiring the code mod, as follows:

We have an MSBuild SDK which may generate code, even in DTB, based on something like `BeforeTargets="Compile"`.  If our MSBuild tasks are unavailable due to misconfiguration, we see the "empty" project update and tree UI warning.  The extension should be warning the user of the failure (thus "resolving" the ambiguous UI), but atm we're not sure how to do that from the DTB.  Obviously, a load-time check would help for that scenario.  However, even properly configured tasks may fail silently in the DTB and produce the behavior.

walterpg:
We are closing this for now. It appears we have solutions for both DTB scenarios after all.  Thanks for the help and sorry to have bothered you.

drewnoakes:
@walterpg thanks. If you find a scenario that repros any issue here I'll be happy to investigate.

PR #8521, opened 2022-09-23
Creates Name, Value List Editor for DefineConstants Property in the VB Property Pages

### LINKED ISSUE #8387: Change VB DefineConstants property editor to be NameValuePairListEditor
For the VB DefineConstants property, the format is Name1="Value1",Name2="Value2",Name3="Value3", which is a _perfect_ candidate for @drewnoakes' NameValuePairListEditor so that users don't have to directly edit this string

haileymck:
Adds the Key, Value Editor for the DefineConstants property in the VB property pages. Goes from single text box input to a list of text boxes with two columns: one key column, one value column. Formats so that the values read in are surrounded in quotes.

Before:
before_PR

After:
after_PR

Resolves #8387 

Creates the following files:
- `src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs`: creates an encoding such that a key's corresponding value is surrounded by quotes, example: key="value". Similar tor the `KeyValuePairListEncoding`, but here the quotes are present. The key quotes are required for the `DefineConstants `property.
- `src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/AbstractProjectConfigurationDefineConstantsEncoding.cs` : Uses the `KeyQuotedValuePairListEncoding `outlined above for the `DefineConstants` property. Encodes strings in the format 'key1=value1,key2=value2...' into 'key1="value1",key2="value2"...'      Note that the `KeyValueListPair `Editor gives the 'key1=value1,key2=value2...' format. 
- `src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsValueProvider.cs `: creates a value provider for the `DefineConstants `property. The input value from  KeyValueListPair Editor is in 'key1=value1,key2=value2...' format so calls the quoted encoding before setting. To get, formats using the encoding described above as well. 

Updates the following files:
- `src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs` : adds the respective gets and sets for the `DefineConstants ` property 
- ` src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml` : edits to add key, value editor field for `DefineConstants `in the property page.  Modifies the instructional text to remove the formatting requirements before the key, value editor. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8521)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=8521) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=8521) before we can accept your contribution.<br/><br/>:x: haileymck [sign now](https://cla.dotnetfoundation.org/dotnet/project-system?pullRequest=8521)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/project-system?pullRequest=8521) it.</sub>

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsValueProvider.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties
+{
+    /// <summary>
+    /// Supports the setting of constants in the property page, corresponding to the "DefineConstants" property
+    /// </summary>
+    [ExportInterceptingPropertyValueProvider("DefineConstants", ExportInterceptingPropertyValueProviderFile.ProjectFile)]
```
adamint, line 10: Could you use the constant you define `DefineConstantsPropertyName`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsValueProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+            return OnGetPropertyValueAsync(evaluatedPropertyValue);
+        }
+
+        public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties)
+        {
+            return OnGetPropertyValueAsync(unevaluatedPropertyValue);
+        }
+
+        private static Task<string> OnGetPropertyValueAsync(string propertyValue)
+        {
```
adamint, line 32: OnGetPropertyValueAsync doesn't need to return a Task, since no parallelizable/threaded work is happening here

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -416,7 +416,10 @@
                   DisplayName="Custom constants"
                   HelpUrl="https://go.microsoft.com/fwlink/?linkid=2196264"
                   Category="Advanced">
-    <StringProperty.Description>Enter any custom constants for your application in this text box. Entries should be delimited by commas, using this form: Name1="Value1",Name2="Value2",Name3="Value3".</StringProperty.Description>
+    <StringProperty.Description>Enter any custom constants for your application in this text box.</StringProperty.Description>
```
adamint, line 419: Formatting seems a little off. I'd also recommend a description change to something like
`Use this editor to modify custom constants for your application.`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/AbstractProjectConfigurationDefineConstantsEncoding.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    ///  Converts the single string representation to multiple key, value pairs used
+    ///  for defining constants.
+    /// </summary>
+    internal static class AbstractProjectConfigurationDefineConstantsEncoding
```
adamint, line 12: This class can be non-static, as everything inside can be instance fields and just instantialize in the value provider constructor

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs
```diff
@@ -149,7 +150,25 @@ public bool RunCodeAnalysis
         public bool DebugSymbols { get => throw new System.NotImplementedException(); set => throw new System.NotImplementedException(); }
         public bool DefineDebug { get => throw new System.NotImplementedException(); set => throw new System.NotImplementedException(); }
         public bool DefineTrace { get => throw new System.NotImplementedException(); set => throw new System.NotImplementedException(); }
-        public string DefineConstants { get => throw new System.NotImplementedException(); set => throw new System.NotImplementedException(); }
+        public string DefineConstants
```
adamint, line 153: is this used anywhere? if not, you can revert :)

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+    /// <summary>
+    ///  Encoding corresponding a format of keys eith a respective value that
+    ///  is surrounded in quotes, ex. key1="value1"
+    /// </summary>
+    internal sealed class KeyQuotedValuePairListEncoding
```
adamint, line 9: you can add a new interface, IEncoding, that KeyQuotedValuePairListEncoding and KeyValuePairListEncoding both implement, so that we might be able to eventually export these encodings and make them available

adamint (changes requested):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/AbstractProjectConfigurationProperties.cs
adamint: there are also a couple unused usings that are failing build https://dev.azure.com/dnceng-public/public/_build/results?buildId=28178&view=logs&j=343adee7-d814-546f-dc3f-19ccb8aa9260&t=8dea7d69-d7ce-5872-c9d8-dd4fee7f95a7&l=9991

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsValueProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties
+{
+    /// <summary>
+    /// Supports the setting of constants in the property page, corresponding to the "DefineConstants" property
+    /// </summary>
+    [ExportInterceptingPropertyValueProvider("DefineConstants", ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+    internal sealed class DefineConstantsValueProvider : InterceptingPropertyValueProviderBase
```
adamint, line 11: I forgot to add, you'll need to add an [AppliesTo(ProjectCapabilities.VisualBasic)] here, because we have a _different_ property with the same name for non-VB projects

adamint: so this should also be renamed 😄

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs
tmeschter: I would hold off on doing that until we have an explicit need. Once we make an interface or implementation `public` it becomes much harder to change later.

tmeschter (changes requested):
You'll need to address the issue that `DefineConstants` is used by VB and C#. Beyond that a couple of other suggestions:

1. It's helpful to include screenshots when making a change that affects UI; that helps reviewers evaluate the new experience.
2. I'd like to see unit tests for the interceptor. We have a number of unit tests for other interceptors that you can use as samples.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs
adamint: @tmeschter it would be internal

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsValueProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public override Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties)
+        {
+            return Task.FromResult<string>(OnGetPropertyValue(unevaluatedPropertyValue));
+        }
+
+        private static string OnGetPropertyValue(string propertyValue)
+        {
+            Dictionary<string, string> constantsDict = AbstractProjectConfigurationDefineConstantsEncoding.ParseIntoDictionary(propertyValue);
+            return AbstractProjectConfigurationDefineConstantsEncoding.DisplayFormat(constantsDict);
```
drewnoakes, line 35: There are two similar snippets in this file, both calling `ParseIntoDictionary` then `*Format`.

There's no other consumer of that intermediary dictionary. In fact, I don't think it's needed.

If you use the singleton idea from earlier, you could simplify this case into:

```suggestion
            return KeyValuePairListEncoding.Instance.Santize(propertyValue);
```

Both `KeyValuePairListEncoding` and `KeyQuotedValuePairListEncoding` would define that method. To do that, you could make `AbstractProjectConfigurationDefineConstantsEncoding` an actual abstract class, and have both encoding types derive from it. The base class could define `Format` as `protected abstract` and delegate to it from its own `Sanitize` method.

I'm happy to chat about this if you like.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -416,7 +416,10 @@
                   DisplayName="Custom constants"
                   HelpUrl="https://go.microsoft.com/fwlink/?linkid=2196264"
                   Category="Advanced">
-    <StringProperty.Description>Enter any custom constants for your application in this text box. Entries should be delimited by commas, using this form: Name1="Value1",Name2="Value2",Name3="Value3".</StringProperty.Description>
+    <StringProperty.Description>Use this editor to modify custom constants for your application.</StringProperty.Description>
```
drewnoakes, line 419: Only minor, but this previously used a `<StringProperty.Description>` element because the text contained quotes (`"`). Now that it no longer does, you can use the `Description="..."` attribute form instead, to be more consistent with sibling properties.

drewnoakes, line 419: Could be made grammatically more consistent with other descriptions via something like:

```suggestion
    <StringProperty.Description>Specifies custom conditional compilation constants for your application.</StringProperty.Description>
```

https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/conditional-compilation

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/AbstractProjectConfigurationDefineConstantsEncoding.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return KeyValuePairListEncoding.Instance.Format(EnumerateConstantsDictionary(constantsDictionary));
+    }
+
+    private static IEnumerable<(string key, string value)> EnumerateConstantsDictionary(Dictionary<string, string> constantsDictionary)
+    {
+        return constantsDictionary switch
+        {
+            null or  { Count: 0 } => Enumerable.Empty<(string key, string value)>(),
+            _ => constantsDictionary.OrderBy(kvp => kvp.Key).Select(kvp => (kvp.Key, kvp.Value))
```
drewnoakes, line 39: Why does this have to be ordered?

I feel like we could avoid the use of `Dictionary<,>` here an use `IEnumerable<(string Key, string Value)>` instead.

The benefit of that change would be that duplicate values are not discarded.

Without this I am concerned that a user who enters a duplicate name would see their value disappear, or the other disappear.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    public IEnumerable<(string Name, string Value)> Parse(string input)
+    {
+        if (string.IsNullOrWhiteSpace(input))
+        {
+            yield break;
+        }
+
+        input = input.Replace("\"", "") + ",";
+        var regex = new Regex(@"(\S+?\s*)\=(\S+?\s*),");
```
drewnoakes, line 23: This supports empty names and values. Assuming that's intentional, the regular expression can be simplified a little:

```suggestion
        var regex = new Regex(@"(\S*\s*)\=(\S*\s*),");
```

This is because `+` matches one or more, and `?` makes that optional. In combination, they are the same as `*` which match zero or more.

---

The use of `(\S*\s*)` means that trailing white space is captured, but not leading white space. Is that intentional?

---

I'm concerned that this won't work when the name or value has either the `=`, `,` or `"` characters within.

---

Named capture groups can make regular expressions easier to understand:

```suggestion
        var regex = new Regex(@"(?<name>\S*\s*)\=(<value>\S*\s*),");
```

Then later you write:

```c#
            string name = match.Groups["name"].Value;
            string value = match.Groups["value"].Value;
```

---

Note that I replaced "key" with "name" on the assumption that it's valid for a user to have duplicate names in the data. Doing so would not be invalid within the project file, and we should respect that.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,51 @@
...
+    public static KeyQuotedValuePairListEncoding Instance { get; } = new();
+
+    public IEnumerable<(string Name, string Value)> Parse(string input)
+    {
+        if (string.IsNullOrWhiteSpace(input))
+        {
+            yield break;
+        }
+
+        input = input.Replace("\"", "") + ",";
```
drewnoakes, line 22: I don't think this would work when the input is `name="A \"valid\" value"`.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/DefineConstantsPropertyPageProviderTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            }
+
+            if (formattedInput.Length > 0)
+            {
+                return formattedInput.Remove(formattedInput.Length - 1);           //remove the trailing comma
+            }
+            else
+            {
+                return formattedInput;
+            }
```
drewnoakes, line 191: This would be more efficient if it used a `StringBuilder`.

In the `foreach` loop, each iteration creates a new temporary string of increasing size. Consider pairs of `[A,1],[B,2],[C,3]`. In the current code we'd allocate all these strings:

- `"A=1,"`
- `"B=2,"`
- `"A=1,B=2,"`
- `"C=3,"`
- `"A=1,B=2,C=3,"`
- `"A=1,B=2,C=3"`

It is possible to allocate only the last one using `StringBuilder`. `StringBuilder` internally creates a buffer to write data into which can avoid a lot of allocations.

Something like:

```suggestion
            StringBuilder formattedInput = new();
            foreach (var kvp in uniqueKeyValuePairs)
            {
                formattedInput.Append(kvp.Key).Append('=').Append(kvp.Value).Append(',');
            }

            if (formattedInput.Length > 0)
            {
                return formattedInput.ToString(0, formattedInput.Length - 1);           //remove the trailing comma
            }
            else
            {
                return formattedInput.ToString();
            }
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/DefineConstantsPropertyPageProviderTests.cs
```diff
@@ -0,0 +1,195 @@
...
+                encoded += outputFormat ? key + "=\"" + value + "\"," : key + "=" + value + ",";
+            }
+            return encoded.TrimEnd(',');
+        }
+
+        //parses repeated keys inputted key value pairs
+        private string getDefinedRepeat(string input)
+        {
+            input += ",";
+            List<KeyValuePair<string, string>> uniqueKeyValuePairs = new List<KeyValuePair<string, string>>();
```
drewnoakes, line 161: Looking at how this collection is used, I think you want a `Dictionary<string, string>` instead. Each call to `FindIndex` in the loop below is O(N), and the loop is O(N), so combining them you get O(N^2), which would be bad for large data sets.

adamint:
This is blocked pending #8559, which is blocked pending https://devdiv.visualstudio.com/DevDiv/_queries/edit/1638100/?triage=true

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/DefineConstantsPropertyPageProviderTests.cs
```diff
@@ -175,19 +175,19 @@ private string getDefinedRepeat(string input)
...
 
             if (formattedInput.Length > 0)
             {
-                return formattedInput.Remove(formattedInput.Length - 1);           //remove the trailing comma
+                return formattedInput.ToString(0, formattedInput.Length - 1);           //remove the trailing comma
             }
             else
             {
-                return formattedInput;
+                return formattedInput.ToString();
```
drewnoakes, line 265:
```suggestion
                return "";
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/AbstractProjectConfigurationDefineConstantsEncoding.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            if (!string.IsNullOrEmpty(key))
+            {
+                constantsDictionary[key] = value;
+            }
+        }
+        return constantsDictionary;
+    }
+
+    public static string Format(string propertyValue)
```
adamint, line 24: Why are these methods static? Can they be turned into instance methods

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+using System.Text.RegularExpressions;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug;
+
+/// <summary>
+///  Encoding corresponding a format of keys eith a respective value that
+///  is surrounded in quotes, ex. key1="value1"
+/// </summary>
+internal sealed class KeyQuotedValuePairListEncoding
```
adamint, line 10: This should inherit the abstract class.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/KeyValuePairListEncoding.cs
```diff
@@ -2,11 +2,10 @@
 
 namespace Microsoft.VisualStudio.ProjectSystem.Debug;
 
- using System.Collections.Generic;
- using System.Linq;
-
 internal sealed class KeyValuePairListEncoding
 {
+    public static KeyValuePairListEncoding Instance { get; } = new();
```
adamint, line 7: Is there a reason this isn't a private instance field?

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/DefineConstantsPropertyPageProviderTests.cs
```diff
@@ -0,0 +1,270 @@
...
+            var defaultProperties = Mock.Of<IProjectProperties>();
+
+            var setResultValue = await provider.OnSetPropertyValueAsync(DefineConstantsValueProvider.DefineConstantsPropertyName, input, defaultProperties);
+
+            Assert.NotNull(setResultValue);
+            Assert.True(ValidEncoding(setResultValue));
+
+            Assert.Equal(expected: EncodeOutputFormat(input, true), actual: setResultValue);
+
+            var getResultValue = await provider.OnGetUnevaluatedPropertyValueAsync(DefineConstantsValueProvider.DefineConstantsPropertyName, input, defaultProperties);
```
adamint, line 144: Is there a test with an escaped comma/equals to ensure those are allowed?

adamint (approved):
Some minor feedback, but nothing too drastic

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/DefineConstantsPropertyPageProviderTests.cs
adamint: This is a test though, so I'm not sure it matters much?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/AbstractProjectConfigurationDefineConstantsEncoding.cs
```diff
@@ -0,0 +1,36 @@
...
+/// </summary>
+internal abstract class StringListEncoding
+{
+    public abstract IEnumerable<(string Name, string Value)> Parse(string input);
+    public abstract string Format(IEnumerable<(string Name, string Value)> pairs);
+
+    public static Dictionary<string, string> ParseIntoDictionary(string inputValue)
+    {
+        Dictionary<string, string> dictionary = new Dictionary<string, string>();
+        foreach ((string key, string value) in KeyQuotedValuePairListEncoding.Instance.Parse(inputValue))
```
adamint, line 17: This doesn't seem right. This should call this class' abstract Parse

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsCSharpValueProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+/// <summary>
+/// Supports the setting of constants in the property page, corresponding to the "DefineConstants" property
+/// </summary>
+[ExportInterceptingPropertyValueProvider(DefineConstantsPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+[AppliesTo(ProjectCapability.VisualBasic)]
+internal sealed class DefineConstantsCSharpValueProvider : InterceptingPropertyValueProviderBase
```
adamint, line 12: rename to DefineConstantsVBValueProvider

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+/// <summary>
+/// Supports the setting of constants in the property page, corresponding to the "DefineConstants" property
+/// </summary>
+[ExportInterceptingPropertyValueProvider(DefineConstantsPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+[AppliesTo(ProjectCapability.VisualBasic)]
```
adamint, line 11: Both of these files apply to VisualBasic capability. the C# one needs to apply to `! VB & CPS` OR needs to apply to CSharpOrFSharp like it did before. I don't know which is which now

adamint (changes requested):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+internal sealed class DefineConstantsVBValueProvider : InterceptingPropertyValueProviderBase
+{
+    internal const string DefineConstantsPropertyName = "DefineConstants";
+
+    public override Task<string?> OnSetPropertyValueAsync(string propertyName, string unevaluatedPropertyValue, IProjectProperties defaultProperties, IReadOnlyDictionary<string, string>? dimensionalConditions = null)
+    {
+        Dictionary<string, string> constantsDictionary = StringListEncoding.ParseIntoDictionary(unevaluatedPropertyValue);
+        return Task.FromResult<string?>(KeyQuotedValuePairListEncoding.Instance.Format(StringListEncoding.EnumerateDictionary(constantsDictionary)));
```
adamint, line 19: As mentioned yesterday, looks nicer as a private field instead of singleton, but up to you

tmeschter (commented):
What's the difference between the `DefineConstantsCSharpValueProvider` and `DefineConstantsVBValueProvider`?

drewnoakes:
I took an in-depth look at this today and believe we can make this a fair bit simpler by making a change on the CSP side. In fact, there's a TODO in the CPS code about this.

This PR adds a `NameValueList` editor for the custom define property.

The default `NameValueList` encoding doesn't use quotes around the value (e.g. `A=1`). However the `<DefineConstants>` MSBuild property does (e.g. `A="1"`).

This PR uses property interception to read the value encoded by the `NameValueList` editor, then re-encode it to use quotes.

It would be simpler overall if the `NameValueList` editor allowed custom encodings. We could achieve this via MEF exports, where the export has a contract name that is also provided in the editor metadata of the property's rule.

In that way, we could export the quoted value encoding, reference its contract name in the rule, and have the UI editor produce the expected value in the first place. That way we no longer need to do any property interception, and avoid having to re-encode values on the fly.

Once we're able to share encodings between CPS and the .NET Project System, we can remove our `KeyValuePairListEncoding` and `LaunchProfileEnvironmentVariableEncoding` classes, delegating to CPS instead.

drewnoakes:
CPS PR that adds an extension point for custom `NameValueList` editor encodings: https://dev.azure.com/devdiv/DevDiv/_git/CPS/pullrequest/435839

Issue #8436, opened 2022-08-29
Settings of a VB project cannot be open in Visual Studio 2022 17.3 when a NuGet package is added


vsfeedback:
_This issue has a corresponding [ticket on Developer Community](https://developercommunity.visualstudio.com/t/Settings-of-a-VB-project-cannot-be-open-/10128349). Please vote and comment there to make sure your voice is heard._

---
\[severity:I’m unable to use this version\] \[regression\]  
Hello,

My name is Vladislav Yurgens. I’m a software engineer representing the WinForms team at DevExpress.

Recently, multiple users reported an issue related to the inability to open settings of a VB .NET 6 project in the Visual Studio 2022 v17.3. The issue occurs when users add one of DevExpress NuGet packages to the project.

Please review the [**VBProject\_VS17\_3Problem.zip**](https://devexpress-my.sharepoint.com/:u:/p/vladislav_yurgens/EeQ9GSBF7XdNn-NVemrX6gQBmSteCSX2BLBnTOODKqL8jA?e=gWehZe) archive, which has the project with the mentioned issue. It will be sufficient to run the sample and open the settings of the project to reproduce the problem.

Let me describe the steps that lead to the issue in greater detail:

**Repro Steps**

1.  Extract the attached archive ([**VBProject\_VS17\_3Problem.zip**](https://devexpress-my.sharepoint.com/:u:/p/vladislav_yurgens/EeQ9GSBF7XdNn-NVemrX6gQBmSteCSX2BLBnTOODKqL8jA?e=gWehZe)) with a solution. We added the required DevExpress package (**DevExpress.Data.Desktop**) to the solution.
2.  Please make sure that you have Visual Studio 2022 v_17.3_ and open **VBProject\_VS17\_3Problem.sln** in your Visual Studio. Invoke the menu in the current project and select the Properties item to open the settings of the project:  
    ![image.png](https://aka.ms/dc/image?name=Bc2f8db6128c342f6aea804759c516cff637968566781458811_20220823-170437-image.png&tid=c2f8db6128c342f6aea804759c516cff637968566781458811)
3.  An error is displayed instead of the settings page:  
    ![image.png](https://aka.ms/dc/image?name=Bd2a2e5dde8ef402ca20a62309cbcd624637968567052843782_20220823-170504-image.png&tid=d2a2e5dde8ef402ca20a62309cbcd624637968567052843782)

Here is the error log:

```  
[...]

PR #8529, opened 2022-09-26
Ensure a WPF project does not have UseWindowsForms set to true.

melytc:
Addresses #8436 and [AzDO#1597634](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1597634).

This PR checks for `UseWindowsForms` property before assuming a project is a WPF one. This avoids the scenario where both `UseWPF` and `UseWindowsForms` are set to true and an error message appear when opening the project properties UI, as reported in the linked issues.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8529)


MiYanni (approved):
Any reason we have duplicate constants between these 2 classes? Maybe have a shared constants static class?

davkean:
These should probably be using the capabilities specified here; https://github.com/dotnet/project-system/blob/a3f824aacdeaee9abb7903ac124eb413908597c1/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L133. These properties are implementation details of how these capabilities are set.

melytc:
> These should probably be using the capabilities specified here;

I've added the capabilities so when #8170 is ready, we filter by them. Thanks for the observation, David.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -6,8 +6,8 @@
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Properties;
 
 // The AppliesTo metadata has no effect given the limitations described in https://github.com/dotnet/project-system/issues/8170.
-[ExportInterceptingPropertyValueProvider(StartupObjectProperty, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
-[AppliesTo(ProjectCapability.VisualBasic)]
+[ExportInterceptingPropertyValueProvider(InterceptedStartupObjectProperty, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+[AppliesTo(ProjectCapability.WPF + "&" + ProjectCapability.WindowsForms)]
```
tmeschter, line 10: This means the interceptor would only apply to projects with _both_ the WPF and WindowsForms capabilities; is that what you want? If you want it to apply to projects that have _either_ WPF or WindowsForms, you need to do this instead:

`[AppliesTo(ProjectCapability.WPF + "|" + ProjectCapability.WindowsForms)]`

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/PropertyNameProvider.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+    /// <summary>
+    /// Contains constants representing the property names used in the ApplicationPropertyPage folder.
+    /// </summary>
+    internal class PropertyNameProvider
```
tmeschter, line 8: I suggest calling this `PropertyNames` instead of `PropertyNameProvider`. A "provider" in CPS and dotnet/project-system tends to mean a factory that produces instances of some other type; e.g. an `IProjectPropertiesProvider` produces instances of `IProjectProperties`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -8,41 +8,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
...
+        PropertyNameProvider.SaveMySettingsOnExitProperty,
+        PropertyNameProvider.HighDpiModeProperty,
+        PropertyNameProvider.AuthenticationModeProperty,
+        PropertyNameProvider.ShutdownModeProperty,
+        PropertyNameProvider.SplashScreenProperty,
+        PropertyNameProvider.MinimumSplashScreenDisplayTimeProperty
     },
     ExportInterceptingPropertyValueProviderFile.ProjectFile)]
-    [AppliesTo(ProjectCapability.VisualBasic)]
+    [AppliesTo(ProjectCapability.WPF + "&" + ProjectCapability.WindowsForms)]
```
tmeschter, line 22: Same as previous comment on AppliesTo.

tmeschter (changes requested):
Double-check the `AppliesTo` attributes, and also remember that they aren't being honored yet due to #8170.

Also, would it be possible to add unit tests covering this scenario?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/ApplicationFrameworkValueProvider.cs
```diff
@@ -8,41 +8,27 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
...
-        MinimumSplashScreenDisplayTimeProperty
+        PropertyNameProvider.ApplicationFrameworkProperty,
+        PropertyNameProvider.EnableVisualStylesProperty,
+        PropertyNameProvider.SingleInstanceProperty,
+        PropertyNameProvider.SaveMySettingsOnExitProperty,
+        PropertyNameProvider.HighDpiModeProperty,
+        PropertyNameProvider.AuthenticationModeProperty,
+        PropertyNameProvider.ShutdownModeProperty,
+        PropertyNameProvider.SplashScreenProperty,
+        PropertyNameProvider.MinimumSplashScreenDisplayTimeProperty
```
MiYanni, line 19: Since you moved the constants to another class, you can use a [using static](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#static-modifier) statement at the top of these files, so you don't need to specify `PropertyNameProvider.` before all these constants.

melytc: Oh, that's a great idea, thanks Yanni!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/PropertyNameProvider.cs
melytc: Sounds good, adding the suggestion on my next commit.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
melytc: Oh, my bad, that was what I was thinking. Thanks for catching it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/PropertyNames.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+    /// <summary>
+    /// Contains constants representing the property names used in the ApplicationPropertyPage folder.
+    /// </summary>
+    internal class PropertyNames
```
drewnoakes, line 8: A few points about this type:

- It can be a `static class`
- The `Property` suffix can be removed from all field names as that is clear from their containing type
- Some of the property names don't match their values, such as `UseApplicationFramework`, `VBAuthenticationMode`, and all of the MSBuild properties (how do they differ from other project properties?). To fix this, consider using `nameof(...)` to ensure they're the same on both sides and prevent confusion.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/ApplicationPropertyPage/WPFValueProvider.cs
```diff
@@ -69,13 +71,16 @@ private async Task<string> GetPropertyValueAsync(string propertyName, IProjectPr
         return string.Empty;
     }
 
-    private static async Task<bool> IsWPFApplicationAsync(IProjectProperties defaultProperties)
+    private async Task<bool> IsWPFApplicationAsync(IProjectProperties defaultProperties)
```
drewnoakes, line 74: The method name is now a bit misleading. Perhaps "IsWpfAndNotWinFormsApplicationAsync" or something like that.

drewnoakes: Also it would be helpful to include a comment explaining why it is useful to know when it is WPF but not WinForms.

melytc:
I rebased to origin/main by mistake; I'll cherry-pick the actual changes into a new branch that's up-to-date with main.

melytc:
Closing in favor of https://github.com/dotnet/project-system/pull/8661

PR #8530, opened 2022-09-27
Fix launchSettings parser to read date as string in environment variables

ocallesp:
Fixes [AB#1599326](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1599326)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8530)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
@@ -356,10 +356,20 @@ internal static class LaunchSettingsJsonEncoding
             ReadObject(property =>
             {
                 builder ??= ImmutableArray.CreateBuilder<(string Name, string Value)>();
-                builder.Add((property, ReadString()));
+                builder.Add((property, ReadEnvironmentVariableString()));
             });
 
             return builder?.ToImmutable() ?? ImmutableArray<(string Name, string Value)>.Empty;
+        }
+
+        string ReadEnvironmentVariableString()
+        {
+            if (!reader.Read())
+            {
+                throw new JsonReaderException($"Cannot read string at {reader.LineNumber}.");
+            }
+
+            return reader.Value!.ToString();
         }
 
         string ReadString()


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
```diff
@@ -356,12 +356,27 @@ LaunchProfile ReadProfile(string name)
...
+            if (!reader.Read())
+            {
+                throw new JsonReaderException($"Cannot read string at {reader.LineNumber}.");
+            }
+
+            return reader.TokenType switch
+            {
+                JsonToken.String => (string)reader.Value!,
+                JsonToken.Date   => reader.Value!.ToString(),
+                _ => throw new JsonReaderException($"Expected string at {reader.LineNumber}:{reader.LinePosition} but got {reader.TokenType}.")
```
MiYanni, line 376: I think you're just intending to read the raw string. It doesn't look like you actually care about the type of the original value. According to [this blog post](https://deliverystack.net/2021/05/15/migrating-c-from-newtonsoft-json-to-system-text-json-for-net-5/), the equivalent of [GetRawText()](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getrawtext?view=net-6.0) from System.Text.Json is simply doing [ToString](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_ToString.htm) on the value.

Meaning, you don't need to parse the `TokenType`. Just do `reader.Value.ToString()` for anything.

ocallesp: It seems that for environment variables converting everything to string makes sense.

Changes as suggested.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
```diff
@@ -356,12 +356,22 @@ LaunchProfile ReadProfile(string name)
...
         }
 
+        string ReadEnvironmentVariableString()
+        {
+            if (!reader.Read())
+            {
+                throw new JsonReaderException($"Cannot read string at {reader.LineNumber}.");
+            }
+
+            return reader.Value!.ToString();
```
drewnoakes, line 372: This will fail if the value is null in the JSON file. You cannot suppress this here. The only reason we can suppress it in the other cases is because we check `TokenType`, which isn't happening here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
```diff
@@ -356,12 +356,22 @@ LaunchProfile ReadProfile(string name)
...
             });
 
             return builder?.ToImmutable() ?? ImmutableArray<(string Name, string Value)>.Empty;
         }
 
+        string ReadEnvironmentVariableString()
+        {
+            if (!reader.Read())
+            {
+                throw new JsonReaderException($"Cannot read string at {reader.LineNumber}.");
```
drewnoakes, line 369: "string" -> "value"

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
```diff
@@ -356,12 +356,22 @@ LaunchProfile ReadProfile(string name)
...
             {
                 builder ??= ImmutableArray.CreateBuilder<(string Name, string Value)>();
-                builder.Add((property, ReadString()));
+                builder.Add((property, ReadEnvironmentVariableString()));
             });
 
             return builder?.ToImmutable() ?? ImmutableArray<(string Name, string Value)>.Empty;
         }
 
+        string ReadEnvironmentVariableString()
```
drewnoakes, line 365: Why this special handling only for environment variables? This bug could happen on any string we read.

drewnoakes:
The original bug here is because JSON.NET handles strings that contain an ISO date in a special way.

That problem could occur on any string we read, so this fix is incomplete.

I suspect a better approach would just be to disable JSON.NET's handling of date strings.

Something like:

```c#
new JsonSerializer { DateParseHandling = DateParseHandling.None }
```

drewnoakes:
I'd like to see unit tests here.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
MiYanni: You're right. He'd have to check in his guard clause that `reader.TokenType != JsonToken.Null`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
MiYanni: I'd assume because the bug was specifically filed against environment variables. So, without going through and determining what should/shouldn't be handled this way, the specific case is enough to handle the issue at hand. I'd argue that serialization is really easy to mess up when doing it by-hand like this; so, minimizing impact it a good strategy. It might not be the 'right' strategy but makes sense if you're not highly familiar with serialization utilities.

PR #8531, opened 2022-09-27
Condition and notification table shouldn't shrink to zero height.

bekir-ozturk:
Partially fixes the accessibility issue here: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1614439

If the screen resolution is low and/or window scaling is high, condition and notification table in "Compile" property page can get too small: enough to disappear completely.

This PR makes sure that the table is at least 200 pixels high.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8531)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/compileproppage2.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/compileproppage2.resx
@@ -725,6 +725,9 @@
   </data>
   <data name="WarningsGridView.Location" type="System.Drawing.Point, System.Drawing">
     <value>3, 115</value>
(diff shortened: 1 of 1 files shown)


tmeschter (approved):
Looks good to me. Could you include a screenshot of the affected property page?

bekir-ozturk:
Sure thing!

Before:
accessibility-before

After:
accessibility-after

drewnoakes (approved):

drewnoakes:
Thanks Bekir. Looks great.

PR #8541, opened 2022-09-28
Disable date deserialisation in launchSettings.json

drewnoakes:
This is an alternative to #8530 that modifies the default JSON.NET behaviour of deserialising a string containing an ISO format date as a `System.DateTime` rather than `System.String`.

Also adds a unit test.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8541)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/LaunchSettingsJsonEncoding.cs
```diff
@@ -414,6 +404,7 @@ private sealed class CommentSkippingJsonTextReader : JsonTextReader
     {
         public CommentSkippingJsonTextReader(TextReader reader) : base(reader)
         {
+            DateParseHandling = DateParseHandling.None;
```
MiYanni, line 407: Nice. 😎

MiYanni (approved):

Issue #7288, opened 2021-06-01
NuGet restore may happen multiple times for multi-target projects


lifengl:
Repro:
1, Open Roslyn solution, after everything settled, close VS.

2, Restart VS, open Roslyn solution again, notice it reaches intellisense ready state quickly after the solution explorer shows the tree

3, close the solution (keep VS open)

4, reopen Roslyn solution in the same VS session
 Notice: it takes much longer times (minutes) to reach intellinsense ready state. Check asset files in the obj folder.  In about 40 projects, they were modified. It doesn't happen when the solution is opened in a fresh booted VS. The content of those files are not changed.

The reason it happens is that for multiple target projects, NuGet restore happens for one of the target framework once. That removes all references for other target frameworks in those projects (and projects in their dependency closure). The rest target frameworks will be added later, and projects will eventually return to the correct state. But that leads a big perf hit, including extra number of design time builds.  Also, if projects are not changed within the session, it will eliminate possibility of optimizations like skipping updating build cache, and break incremental build as well.

+ @drewnoakes  

+ @panopticoncentral

ocallesp:
Hi @drewnoakes,

I see that this still reproes when VS opens the Roslyn project.

To reduce the repro steps I wanted to create a more simple project.
I created a C# ConsoleApp project targeting net472, and added a project reference to a class library targeting net472;net48 with the nuGet package Newtonsoft, but I couldn't repro it in here.

Any suggestion on what project configuration can repro this ?

ocallesp:
I noticed that when using Roslyn project, the cache file are not modified the first time you restart and open the solution. I see that sometimes it reproes after many tries.

drewnoakes:
@ocallesp it would be great to narrow down to a single project. Does it repro with the smaller Compilers.sln for example? You could try unloading half the projects at a time, and seeing if it still reproduces.

ocallesp:
Does not repro always.

Using this simple project https://github.com/ocallesp/Repro7288 I found that during solution load some cache files are modified (we can detect this with git status). 
When the solution load finishes, the cache files appear as nothing has changed because the final modification are the same as it was originally.

	modified:   ClassLibrary.csproj.nuget.dgspec.json
	modified:   project.assets.json
	modified:   project.nuget.cache

ocallesp:
@drewnoakes,

[...]

PR #8543, opened 2022-09-28
Fix nuget restore perf regression

ocallesp:
#7288

Avoid doing restore when receiving parts of the target frameworks. We can fix this by waiting for all target frameworks to get loaded and then publish.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8543)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
@@ -76,6 +76,8 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
 
         private PackageRestoreUnconfiguredInput MergeRestoreInputs(RestoreValues restoreValues)
         {
+            System.Diagnostics.Debug.Assert(restoreValues.Item1.Count() == restoreValues.Item2.Count());
+
             IReadOnlyList<PackageRestoreConfiguredInput> inputs = restoreValues.Item2;
 
             // If there are no updates, we have no active configurations
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
@@ -4,7 +4,12 @@ using System.Globalization;
 using System.Threading.Tasks.Dataflow;
 using Microsoft.VisualStudio.ProjectSystem.Utilities;
 using NuGet.SolutionRestoreManager;
+
 using RestoreInfo = Microsoft.VisualStudio.ProjectSystem.IProjectVersionedValue<Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore.PackageRestoreUnconfiguredInput>;
+
+using RestoreValues = System.ValueTuple<
+    System.Collections.Generic.IEnumerable<Microsoft.VisualStudio.ProjectSystem.ConfiguredProject>,
+    System.Collections.Generic.IReadOnlyList<Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore.PackageRestoreConfiguredInput>>;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
 {
@@ -14,13 +19,15 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
     {
         private readonly UnconfiguredProject _project;
         private readonly IActiveConfigurationGroupService _activeConfigurationGroupService;
+        private readonly ILoadedActiveConfiguredProjectDataSource _loadedActiveConfiguredProjectDataSource;
 
         [ImportingConstructor]
-        public PackageRestoreUnconfiguredInputDataSource(UnconfiguredProject project, IActiveConfigurationGroupService activeConfigurationGroupService)
+        public PackageRestoreUnconfiguredInputDataSource(UnconfiguredProject project, IActiveConfigurationGroupService activeConfigurationGroupService, ILoadedActiveConfiguredProjectDataSource loadedActiveConfiguredProjectDataSource)
             : base(project, synchronousDisposal: false, registerDataSource: false)
         {
             _project = project;
             _activeConfigurationGroupService = activeConfigurationGroupService;
+            _loadedActiveConfiguredProjectDataSource = loadedActiveConfiguredProjectDataSource;
         }
 
         protected override UnconfiguredProject ContainingProject
@@ -40,13 +47,22 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.PackageRestore
                 JoinableFactory,
                 _project);
 
-            // Transform all restore data -> combined restore data
-            DisposableValue<ISourceBlock<RestoreInfo>> mergeBlock = joinBlock.TransformWithNoDelta(update => update.Derive(MergeRestoreInputs));
+            IPropagatorBlock<IProjectVersionedValue<RestoreValues>, RestoreInfo> transformBlock
+                = DataflowBlockSlim.CreateTransformBlock<IProjectVersionedValue<RestoreValues>, RestoreInfo>(update => update.Derive(MergeRestoreInputs));
+
+            // Publish when all target frameworks get loaded
+            var mergeBlock = ProjectDataSources.SyncLinkTo(
+                _loadedActiveConfiguredProjectDataSource.SourceBlock.SyncLinkOptions(),
+                joinBlock.SyncLinkOptions(),
+                target: transformBlock,
+                linkOptions: DataflowOption.PropagateCompletion,
+                cancellationToken: CancellationToken.None
+                );
 
             JoinUpstreamDataSources(_activeConfigurationGroupService.ActiveConfiguredProjectGroupSource);
 
             // Set the link up so that we publish changes to target block
-            mergeBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+            transformBlock.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
 
             return new DisposableBag
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
```diff
@@ -40,13 +47,22 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreInfo> targe
...
+                );
 
-            JoinUpstreamDataSources(_activeConfigurationGroupService.ActiveConfiguredProjectGroupSource);
+            JoinUpstreamDataSources(_activeConfigurationGroupService.ActiveConfiguredProjectGroupSource, _loadedActiveConfiguredProjectDataSource);
 
             // Set the link up so that we publish changes to target block
-            mergeBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+            transformBlock.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
 
             return new DisposableBag
```
drewnoakes, line 67: Do you have to dispose `mergeBlock` and `transformBlock` here too?

tmeschter: I think Drew means that the `LinkTo` method returns an `IDisposable` that will break the link between the blocks when disposed.

ocallesp: done. Added the disposable object returned from LinkTo

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LoadedActiveConfiguredProjectsDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    protected override IDisposable? LinkExternalInput(ITargetBlock<IProjectVersionedValue<IEnumerable<ConfiguredProject>>> targetBlock)
+    {
+        IReceivableSourceBlock<IProjectVersionedValue<IConfigurationGroup<ProjectConfiguration>>> sourceBlock = _activeConfigurationGroupService.ActiveConfigurationGroupSource.SourceBlock;
+
+        DisposableValue<ISourceBlock<IProjectVersionedValue<IEnumerable<ConfiguredProject>>>>? transformBlock = sourceBlock.TransformWithNoDelta(TransformAsync);
+
+        transformBlock.Value.LinkTo(targetBlock, DataflowOption.PropagateCompletion);
+
+        return transformBlock;
```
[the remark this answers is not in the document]
ocallesp: fixed in the commit "code cleanup"

ocallesp:
it seems that `WriteAsync_CompletesWhenDataReceived` is failing here, but not when using Test Explorer locally.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LoadedActiveConfiguredProjectsDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        List<ConfiguredProject> generatedResult = new List<ConfiguredProject>();
+
+        foreach (ProjectConfiguration configuration in projectVersionedValue.Value)
+        {
+            // Make sure we aren't currently unloading, or we don't unload while we load the configuration
+            var loadedConfiguredProject = await _tasksService.LoadedProjectAsync(() =>
+            {
+                return _project.LoadConfiguredProjectAsync(configuration);
+            });
```
MiYanni, line 55: nit: This can be simplified. Single-statements in a lambda don't need to be enclosed (it can be used as an expression).
```suggestion
            var loadedConfiguredProject = await _tasksService.LoadedProjectAsync(() => _project.LoadConfiguredProjectAsync(configuration));
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsLoader.cs
```diff
@@ -57,43 +55,14 @@ protected override void Dispose(bool disposing)
 
         private async Task OnActiveConfigurationsChangedAsync(IProjectVersionedValue<IConfigurationGroup<ProjectConfiguration>> e)
         {
-            await GetLoadedProjectsAsync(e);
-        }
-
-        protected override IDisposable? LinkExternalInput(ITargetBlock<IProjectVersionedValue<IEnumerable<ConfiguredProject>>> targetBlock)
-        {
-            IReceivableSourceBlock<IProjectVersionedValue<IConfigurationGroup<ProjectConfiguration>>> sourceBlock = _activeConfigurationGroupService.ActiveConfigurationGroupSource.SourceBlock;
```
lifengl, line 65: it need JoinUpstream to the ActiveConfigurationGroupSource

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsLoader.cs
```diff
@@ -57,43 +55,14 @@ protected override void Dispose(bool disposing)
...
-            List<ConfiguredProject> generatedResult = await GetLoadedProjectsAsync(projectVersionedValue);
-
-            return new ProjectVersionedValue<IEnumerable<ConfiguredProject>>(generatedResult, projectVersionedValue.DataSourceVersions);
-        }
-
-        private async Task<List<ConfiguredProject>> GetLoadedProjectsAsync(IProjectVersionedValue<IConfigurationGroup<ProjectConfiguration>> projectVersionedValue)
-        {
-            List<ConfiguredProject> generatedResult = new List<ConfiguredProject>();
-
-            foreach (ProjectConfiguration configuration in projectVersionedValue.Value)
```
lifengl, line 85: i think this function need wrap the inner code into this.JoinableTaskFactory.RunAsync to prevent deadlocks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LoadedActiveConfiguredProjectsDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Threading.Tasks.Dataflow;
+
+namespace Microsoft.VisualStudio.ProjectSystem;
+
+/// <summary>
+///     Force loads the active <see cref="ConfiguredProject"/> objects so that any configured project-level
+///     services, such as evaluation and build services, are started.
+/// </summary>
+[Export(typeof(ILoadedActiveConfiguredProjectDataSource))]
+internal class LoadedActiveConfiguredProjectsDataSource : ChainedProjectValueDataSourceBase<IEnumerable<ConfiguredProject>>, ILoadedActiveConfiguredProjectDataSource
```
davkean, line 12: I don';t understand the difference between ActiveConfiguredProjectsLoader and this?

lifengl: I agree. We should not have double copied the code. What is the reason we revert the change and create a new class? What didn't work in the original change? I think we need make code change based on understanding the problem, instead of trying with slightly different code.  Basically, this code has similar problem like last change in the original place, it didn't follow dataflow patterns correctly. It need JoinUpstream blocks, and because the transform block has a complex function like this, it need run it inside a JTF.RunAsync to prevent deadlocks.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LoadedActiveConfiguredProjectsDataSource.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private async Task<List<ConfiguredProject>> GetLoadedProjectsAsync(IProjectVersionedValue<IConfigurationGroup<ProjectConfiguration>> projectVersionedValue)
+    {
+        List<ConfiguredProject> generatedResult = new List<ConfiguredProject>();
+
+        foreach (ProjectConfiguration configuration in projectVersionedValue.Value)
+        {
+            // Make sure we aren't currently unloading, or we don't unload while we load the configuration
+            var loadedConfiguredProject = await _tasksService.LoadedProjectAsync(() =>
```
lifengl, line 52: i feel we can wrap the entire loop in one LoadedProjectAsync, and check UnloadingCancellationToken in the loop?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore/PackageRestoreUnconfiguredInputDataSource.cs
```diff
@@ -40,25 +47,37 @@ protected override IDisposable LinkExternalInput(ITargetBlock<RestoreInfo> targe
...
                 _activeConfigurationGroupService.ActiveConfiguredProjectGroupSource.SourceBlock.LinkTo(joinBlock, DataflowOption.PropagateCompletion),
             };
         }
 
-        private PackageRestoreUnconfiguredInput MergeRestoreInputs(IReadOnlyCollection<PackageRestoreConfiguredInput> inputs)
+        private PackageRestoreUnconfiguredInput MergeRestoreInputs(RestoreValues restoreValues)
         {
+            IReadOnlyList<PackageRestoreConfiguredInput> inputs = restoreValues.Item2;
+
             // If there are no updates, we have no active configurations
```
lifengl, line 81: It looks right to me. You might want to assert the two inputs are consistent -- means the number of configured projects from loaded flow and the number of configurations from the joinBlock should match. If they don't match, it would indicate a bug somewhere.

ocallesp: done. Assert added

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsLoader.cs
ocallesp: done.

Loop executed inside JoinableTaskFactory

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ActiveConfiguredProjectsLoader.cs
ocallesp: done.

ActiveConfigurationGroupSource added to JoinUpstream

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LoadedActiveConfiguredProjectsDataSource.cs
ocallesp: done.

ocallesp:
This pr was not merged because I was not able to reproduce this bug.

PR #8550, opened 2022-09-29
Language service tidyup

drewnoakes:
I recommend reviewing this one commit at a time.

Key changes:

- Remove duplicate validation and `JTC.Join` calls
- Ensure validation occurs within joined block to prevent deadlocks
- Remove a potentially redundant UI thread switch
- Add some API docs
- Give a dataflow action block a name to make debugging easier

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -132,7 +132,8 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
                 target: DataflowBlockFactory.CreateActionBlock<IProjectVersionedValue<(ConfiguredProject ActiveConfiguredProject, ConfigurationSubscriptionSources Sources)>>(
                     async update => await ExecuteUnderLockAsync(cancellationToken => OnSlicesChanged(update, cancellationToken)),
                     _unconfiguredProject,
-                    ProjectFaultSeverity.LimitedFunctionality),
+                    ProjectFaultSeverity.LimitedFunctionality,
+                    "LanguageServiceHostSlices {0}"),
                 linkOptions: DataflowOption.PropagateCompletion,
                 cancellationToken: cancellationToken),
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -298,22 +298,22 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
         {
             await ValidateEnabledAsync(cancellationToken);
 
-            await WhenProjectLoaded(cancellationToken);
+            await WhenProjectLoaded();
         }
 
         return _primaryWorkspace ?? throw Assumes.Fail("Primary workspace unknown.");
-    }
 
-    private async Task WhenProjectLoaded(CancellationToken cancellationToken)
-    {
-        // The active configuration can change multiple times during initialization in cases where we've incorrectly
-        // guessed the configuration via our IProjectConfigurationDimensionsProvider3 implementation.
-        // Wait until that has been determined before we publish the wrong configuration.
-        await _tasksService.PrioritizedProjectLoadedInHost.WithCancellation(cancellationToken);
+        async Task WhenProjectLoaded()
+        {
+            // The active configuration can change multiple times during initialization in cases where we've incorrectly
+            // guessed the configuration via our IProjectConfigurationDimensionsProvider3 implementation.
+            // Wait until that has been determined before we publish the wrong configuration.
+            await _tasksService.PrioritizedProjectLoadedInHost.WithCancellation(cancellationToken);
 
-        // We block project load on initialization of the primary workspace.
-        // Therefore by this point we must have set the primary workspace.
-        System.Diagnostics.Debug.Assert(_firstPrimaryWorkspaceSet.Task is { IsCompleted: true, IsFaulted: false });
+            // We block project load on initialization of the primary workspace.
+            // Therefore by this point we must have set the primary workspace.
+            System.Diagnostics.Debug.Assert(_firstPrimaryWorkspaceSet.Task is { IsCompleted: true, IsFaulted: false });
+        }
     }
 
     #endregion
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -296,7 +296,10 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
     {
         await ValidateEnabledAsync(cancellationToken);
 
-        await WhenProjectLoaded(cancellationToken);
+        using (_joinableTaskCollection.Join())
+        {
+            await WhenProjectLoaded(cancellationToken);
+        }
 
         return _primaryWorkspace ?? throw Assumes.Fail("Primary workspace unknown.");
     }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -333,10 +333,7 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
         // Ensure the project is not considered loaded until our first publication.
         Task result = _tasksService.PrioritizedProjectLoadedInHostAsync(async () =>
         {
-            using (_joinableTaskCollection.Join())
-            {
-                await WhenInitialized(_tasksService.UnloadCancellationToken);
-            }
+            await WhenInitialized(_tasksService.UnloadCancellationToken);
         });
 
         // While we want make sure it's loaded before PrioritizedProjectLoadedInHost,
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -278,8 +278,6 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
     {
         token = _tasksService.LinkUnload(token);
 
-        await ValidateEnabledAsync(token);
-
         Workspace workspace = await GetPrimaryWorkspaceAsync(token);
 
         await workspace.WriteAsync(action, token);
@@ -288,8 +286,6 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
     public async Task<T> WriteAsync<T>(Func<IWorkspace, Task<T>> action, CancellationToken token)
     {
         token = _tasksService.LinkUnload(token);
-
-        await ValidateEnabledAsync(token);
 
         Workspace workspace = await GetPrimaryWorkspaceAsync(token);
 
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -266,10 +266,10 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
 
     public async Task WhenInitialized(CancellationToken token)
     {
-        await ValidateEnabledAsync(token);
-
         using (_joinableTaskCollection.Join())
         {
+            await ValidateEnabledAsync(token);
+
             await _firstPrimaryWorkspaceSet.Task.WithCancellation(token);
         }
     }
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -265,10 +266,10 @@ public Task<bool> IsEnabledAsync(CancellationToken cancellationToken)
 
     public async Task WhenInitialized(CancellationToken token)
     {
-        await ValidateEnabledAsync(token);
-
         using (_joinableTaskCollection.Join())
         {
+            await ValidateEnabledAsync(token);
```
davkean, line 271: This commit description for this change does not look correct. Join is about connecting the [JoinableTask created here](http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Managed&leftSymbol=mmqz8avx2ktv&file=ProjectSystem%5CLanguageServices%5CWorkspace.cs&line=152) with the caller of this method. Anyone synchronously blocking on the UI thread on this method will happily let ValidateEnabledAsync switch to the UI thread if they are following JTF rules (ie calling JTF.Run).

davkean: ie Join solely exists because there's not a traditional async connection between _firstPrimaryWorkspaceSet and that code I pointed to.

adrianvmsft (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -132,7 +132,8 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
                 target: DataflowBlockFactory.CreateActionBlock<IProjectVersionedValue<(ConfiguredProject ActiveConfiguredProject, ConfigurationSubscriptionSources Sources)>>(
                     async update => await ExecuteUnderLockAsync(cancellationToken => OnSlicesChanged(update, cancellationToken)),
```
lifengl, line 133: we can remove both async/await here to reduce extra async state machines.

also, I wonder whether you still need ExecuteUnderLockAsync here, because action block never processes two pieces of input at the same time, and the disposing code is to dispose links, and doesn't have any race condition between the two.  The async semaphore is essentially another queue to add more overhead to this.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -333,10 +333,7 @@ public async Task AfterLoadInitialConfigurationAsync()
         // Ensure the project is not considered loaded until our first publication.
         Task result = _tasksService.PrioritizedProjectLoadedInHostAsync(async () =>
         {
-            using (_joinableTaskCollection.Join())
-            {
-                await WhenInitialized(_tasksService.UnloadCancellationToken);
-            }
+            await WhenInitialized(_tasksService.UnloadCancellationToken);
```
lifengl, line 336: we can remove both async/await here, and just return the result of WhenInitialized.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -265,10 +266,10 @@ public Task<bool> IsEnabledAsync(CancellationToken cancellationToken)
 
     public async Task WhenInitialized(CancellationToken token)
     {
-        await ValidateEnabledAsync(token);
-
         using (_joinableTaskCollection.Join())
```
lifengl, line 269: i still have questions around JTF chains in this code. One thing interesting is that we create joinableTaskCollection/Factory here instead of reusing ones created in its base class (OnceInitializeOnceDisposeAsync exposes them through protect members, so joins the collection will join initialization tasks as well (of course, it doesn't matter much, because we ensure the initialization in the LoadAsync method.

this code doesn't seem to JoinUpstream for two blocks linked here including _activeConfiguredProjectProvider and _activeConfigurationGroupSubscriptionService. They should be covered.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
lifengl: BTW, if we do have other code path to call this ExecuteUnderLockAsync, to follow the JTF rule, you must call this code inside a JTF.RunAsync:

async update => await _joinableTaskFactory.RunAsync(() => ExecuteUnderLockAsync(...)),

missing this JTF.RunAsync means that we will not join correctly for the async semaphore JTF chain, which can lead into deadlocks.

lifengl: The code in Workspace.cs looks good. I am not sure why we need retain the ExecuteUnderLock/AsyncSemaphore there as well. I cannot find any usage of WriteAsync methods. Those seem to be a part of old design, and can be removed?

lifengl: BTW, we should consider chaining the fault completion of the action block to abort the task completion source of _firstPrimaryWorkspaceSet. Basically, if we run into any exception inside OnSliceChanged, it will send NFE, but we don't want to hang the solution loading.  It just adds a safe net for this code.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: > This commit description for this change does not look correct. Join is about connecting the [JoinableTask created here](http://index/?leftProject=Microsoft.VisualStudio.ProjectSystem.Managed&leftSymbol=mmqz8avx2ktv&file=ProjectSystem%5CLanguageServices%5CWorkspace.cs&line=152) with the caller of this method. Anyone synchronously blocking on the UI thread on this method will happily let ValidateEnabledAsync switch to the UI thread if they are following JTF rules (ie calling JTF.Run).

This JTC has all upstream data sources joined to it. The `Join` here allows dataflow to access the main thread, even if a caller of this method is on the main thread.

> the information whether it is running under command line is known upfront, the implementation could be further simplified and eliminate async - by getting this information using ICriticalPackageService

`ICriticalPackageService` is not available outside of CPS. I don't know a better way to initialize these variables early enough during solution load to ensure they're populated before we query them.

> moving ValidateEnabledAsync inside the collection join is unnecessary, because the AsyncLazy has its own JTF chain managed inside. it would not break things either.

If the join here was removed and the caller was on the main thread, and they trigger the lazy initialisation of the "is command line mode" state, which also require the main thread, but dataflow is running and is blocking the main thread, wouldn't it be able to deadlock?

adrianvmsft: > ICriticalPackageService is not available outside of CPS. I don't know a better way to initialize these variables early enough during solution load to ensure they're populated before we query them.

Oh, I didn't realize that ICriticalPackageService is not public.
Another option would be to expose IVsShellServices.IsInCommandLineMode from CPS as public but that needs some additional thinking.

davkean: >If the join here was removed and the caller was on the main thread, and they trigger the lazy initialisation of the "is command line mode" state, which also require the main thread, but dataflow is running and is blocking the main thread, wouldn't it be able to deadlock?

Join says "let any tasks added to the JoinableTaskCollection" access the main thread if the caller is blocking it. ValidateEnabledAsync does not add any tasks to the collection, and therefore putting it in the join has no affect.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: > I wonder whether you still need ExecuteUnderLockAsync here

Thanks, indeed `ExecuteUnderLockAsync` here seems redundant. Removed in b63e4434025bd4a9889024b6c806bac69607bc3a.

> The code in Workspace.cs looks good. I am not sure why we need retain the ExecuteUnderLock/AsyncSemaphore there as well. I cannot find any usage of WriteAsync methods. Those seem to be a part of old design, and can be removed?

They're used via `IWorkspaceWriter.WriteAsync` in TempPE, error list, properties providers and code model code. So we cannot remove them, and we must keep the `ExecuteUnderLockAsync` calls to prevent callers of `WriteAsync` overlapping with our own updates.

Does your advice around `_jtf.RunAsync(() => ExecuteUnderLockAsync(...` apply in `Workspace.cs` too then?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: I kept this intentionally. Without it, my concern was that a hang related to this code path would be harder to identify as the call to `WhenInitialized` wouldn't be associated at all with the `ProjectAutoLoad` method. Does that make sense? The overhead in this instance is minimal (one additional `Task` per unconfigured project during solution load).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: > this code doesn't seem to JoinUpstream for two blocks linked here including _activeConfiguredProjectProvider and _activeConfigurationGroupSubscriptionService. They should be covered.

Thanks. Fixed in a4f25885a0706f7404a7b67e238a553508f9f64c. Note that I'm joining to my custom JTC/JTF rather than `OnceInitializedOnceDisposedAsync.JoinableFactory` from the base. We use the custom JTC to join all dataflow work here with callers to `IWorkspaceWriter.WriteAsync`.

> i still have questions around JTF chains in this code. One thing interesting is that we create joinableTaskCollection/Factory here instead of reusing ones created in its base class

When you and I looked at a hang here a while back, I thought we concluded that we'd need our own JTC to sync all this work. I assume the one in `OnceInitializedOnceDisposedAsync.JoinableFactory` uses a global collection.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: > ValidateEnabledAsync does not add any tasks to the collection

I thought tasks in the same collection are implicitly joined. The goal of having `ValidateEnabledAsync` inside the `Join` here is so that when it creates its own `JoinableTask` as part of the `AsyncLazy` (which uses the default JTF via MEF) that the `JoinableTask` instances associated with separate collections are considered as joined for the duration of that scope.

davkean: Awaiting an AsyncLazy<T> implicitly joins the caller's context with the _Func<Task<T>>_ its executing, so if it needs UI thread, it will get it. All in all, there's probably no real side affect of doing what you are doing (other than allowing data to access UI thread earlier), but its unneeded and the lack of doing this will not cause a hang.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -288,32 +287,33 @@ public async Task<T> WriteAsync<T>(Func<IWorkspace, Task<T>> action, Cancellatio
...
+        async Task WhenProjectLoaded()
+        {
+            // The active configuration can change multiple times during initialization in cases where we've incorrectly
+            // guessed the configuration via our IProjectConfigurationDimensionsProvider3 implementation.
+            // Wait until that has been determined before we publish the wrong configuration.
+            await _tasksService.PrioritizedProjectLoadedInHost.WithCancellation(cancellationToken);
+
+            // We block project load on initialization of the primary workspace.
+            // Therefore by this point we must have set the primary workspace.
+            System.Diagnostics.Debug.Assert(_firstPrimaryWorkspaceSet.Task is { IsCompleted: true, IsFaulted: false });
```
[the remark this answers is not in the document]
drewnoakes: We have a `Microsoft.VisualStudio.ProjectSystem.Debug` namespace.

We could alias it, but I don't find the full form offensive. There are 12 instances of this throughout the repo.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
adrianvmsft: `OnceInitializedOnceDisposedAsync.JoinableFactory` uses its own collection - which is exposed as a property

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: Moved the removal of locking in `LanguageServiceHost` to a separate PR: https://github.com/dotnet/project-system/pull/8618

MiYanni:
@drewnoakes Howdy. Just checking the status on this PR.

PR #8555, opened 2022-09-30
Build events property page buttons shouldn't overlap

bekir-ozturk:
**Problem:** On small screens, build events property page won't fit the window. As a result:
- Some parts of the page are clipped,
- Some of the content gets behind the "OK" and "Cancel" buttons, causing an overlap.

**Cause:** The issue is caused by the owner window [PropPageUserControlPage](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.AppDesigner/PropPages/PropPageUserControlBase.vb#L2880), scaling the minimum size of the property page up based on the font:
-> [SetDialogFont(ScaleDialog As Boolean)](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.AppDesigner/PropPages/PropPageUserControlBase.vb#L3541)
-> [MinimumSize = ScaleSize(MinimumSize, dx, dy)](https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.AppDesigner/PropPages/PropPageUserControlBase.vb#L3603)

 Thinking that they have a larger space to lay on, the controls are enlarging themselves to the point they don't fit the window.

**Fix:** The window that displays our property page is responsible for displaying other windows as well. I'm not sure why a font-based scaling is being done here, but I think it is risky to make a change in the owner and potentially break other property pages that use it. Therefore, the fix only handles the case for the Build Events property page.
    Build Events page listens to size change event and makes sure that its minimum size is always the same as the minimum size of its child control.

Screenshots:
**Before**
d1

**After**
d2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8555)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/BuildEventsPropPage.resx
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/BuildEventsPropPage.resx
@@ -208,7 +208,7 @@
     <value>0, 3, 0, 3</value>
   </data>
   <data name="btnPreBuildBuilder.Padding" type="System.Windows.Forms.Padding, System.Windows.Forms">
-    <value>10, 0, 10, 0</value>
+    <value>5, 0, 5, 0</value>
   </data>
   <data name="btnPreBuildBuilder.Size" type="System.Drawing.Size, System.Drawing">
     <value>111, 23</value>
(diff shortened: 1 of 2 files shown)


MiYanni (approved):
LGTM!

melytc (approved):

drewnoakes (approved):

PR #8559, opened 2022-10-03
Add MultiStringSelector edit to DefineConstants property for C#/F# projects

### LINKED ISSUE #6931: Modifying conditional compilation symbols produces unexpected evaluated values. (2)
While testing #6899 I noticed the following strange behaviour.

2. Open new Project Properties UI
3. Modify _Conditional compilation symbols_ by appending a `2` so it reads `$(DefineConstants)TRACE2`
4. Press <kbd>Enter</kbd> to commit the value
5. Notice the evaluated values look wrong

Before

image

After

image

### LINKED ISSUE #8170: Support AppliesTo filtering for property interceptors

adamint:
Adds a MultiStringSelector editor to the DefineConstants property for non-VB projects. The purpose of this is to abstract away the recursive part of the property definition to make it less confusing for users (evidenced by the amount of feedback tickets we get).

Additionally, the following feature is implemented: #8170, in order to prevent a collision with the value provider that will be introduced in #8521.
- Filtering is done in InterceptedProjectProperties.cs in _lazy_ style. Capabilities for a specific propertyName's interceptors are checked when an interceptor is requested. More than one returned interceptor throws an exception.

 An example of this property (with evaluated value preview shown):
image

Resolves #8170 and resolves #6931

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8559)


tmeschter:
Should this be a draft PR?

adamint:
> Should this be a draft PR?

I got sidetracked and forgot to update the title... it's ready to go in

docs/repo/property-pages/property-specification.md
```diff
@@ -233,18 +233,20 @@ The property's string value should be encoded with format resembling `A=1,B=2`,
...
 
+To show the evaluated value of the property below the multi-string selector, set the `ShouldDisplayEvaluatedPreview` metadata value to True.
+
 ```xml
   <DynamicEnumProperty Name="ImportedNamespaces"
                        DisplayName="Import Namespaces"
                        Description="Manage which namespaces to import in your application."
                        Category="General"
-                       EnumProvider="DotNetImportsEnumProvider">
+                       EnumProvider="DotNetImportsEnumProvider"> <!-- use a StringProperty in your application -->
```
drewnoakes, line 249: Why not change the property type in this example code?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsValueProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        _projectAccessor = projectAccessor;
+        _project = project;
+    }
+
+    internal static IEnumerable<string> ParseDefinedConstantsFromUnevaluatedValue(string unevaluatedValue)
+    {
+        return unevaluatedValue.Length <= DefineConstantsRecursivePrefix.Length  || !unevaluatedValue.StartsWith(DefineConstantsRecursivePrefix)
+            ? Array.Empty<string>() 
+            : unevaluatedValue.Substring(DefineConstantsRecursivePrefix.Length).Split(';').Where(x => x.Length > 0);
```
drewnoakes, line 29: Can values contain `;` characters escaped somehow?

adamint: No, they cannot.

adamint:
Likely blocked pending https://devdiv.visualstudio.com/DevDiv/_queries/edit/1638100/?triage=true

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptedProjectProperties.cs
```diff
@@ -27,10 +31,14 @@ public InterceptedProjectProperties(ImmutableArray<Lazy<IInterceptingPropertyVal
...
-
-                    builder.Add(propertyName, valueProvider);
+                    if (!builder.ContainsKey(propertyName))
+                    {
+                        builder.Add(propertyName, (new List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> { valueProvider }, false));
+                    }
+                    else
+                    {
+                        builder[propertyName].Providers.Add(valueProvider);
+                    }
```
drewnoakes, line 41: This does a double lookup.  How about:

```suggestion
                    if (!builder.TryGetValue(propertyName, out var entry))
                    {
                        entry = (new List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> { valueProvider }, false);
                        builder.Add(propertyName, entry);
                    }

                    entry.Providers.Add(valueProvider);
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptedProjectProperties.cs
```diff
@@ -11,14 +11,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.Properties
...
+        
+        public InterceptedProjectProperties(ImmutableArray<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> valueProviders, IProjectProperties defaultProperties, UnconfiguredProject project)
             : base(defaultProperties)
         {
+            _project = project;
             Requires.NotNullOrEmpty(valueProviders, nameof(valueProviders));
 
-            ImmutableDictionary<string, Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>>.Builder builder = ImmutableDictionary.CreateBuilder<string, Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>>(StringComparers.PropertyNames);
+            ImmutableDictionary<string, (List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> Providers, bool Filtered)>.Builder builder = 
+                ImmutableDictionary.CreateBuilder<string, (List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> Providers, bool Filtered)>(StringComparers.PropertyNames);
```
drewnoakes, line 24: Symbols having boolean values are usually prefixed with "is" in .NET. Consider renaming the value tuple member to `IsFiltered`.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptedProjectProperties.cs
```diff
@@ -98,5 +110,34 @@ public void SetRuleContext(Rule rule)
             var ruleAwareProperties = DelegatedProperties as IRuleAwareProjectProperties;
             ruleAwareProperties?.SetRuleContext(rule);
         }
+
+        private IInterceptingPropertyValueProvider? GetFilteredProvider(string propertyName, (List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> Providers, bool Filtered) valueTuple)
```
drewnoakes: Throughout this code in this class, the type `(List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> Providers, bool Filtered)` is used heavily, and makes the code quite hard to read.

Consider promoting the value tuple to a dedicated nested type, and moving the `GetFilteredProvider` method to that new type.

```c#
private readonly struct Providers
{
    public List<Lazy<IInterceptingPropertyValueProvider, IInterceptingPropertyValueProviderMetadata>> Exports { get; }

    public bool Filtered { get; }

    public IInterceptingPropertyValueProvider? GetFilteredProvider(string propertyName) { ... }
}
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptingPropertyValueProviderBase.cs
```diff
@@ -27,13 +25,13 @@ public virtual Task<string> OnGetUnevaluatedPropertyValueAsync(string propertyNa
 
         public virtual async Task<bool> IsValueDefinedInContextAsync(string propertyName, IProjectProperties defaultProperties)
         {
+            var t = await defaultProperties.IsValueInheritedAsync(propertyName);
             return !await defaultProperties.IsValueInheritedAsync(propertyName);
         }
 
-        private protected static async Task<bool> IsValueDefinedInContextMSBuildPropertiesAsync(IProjectProperties defaultProperties, string[] msBuildPropertyNames)
+        internal static async Task<bool> IsValueDefinedInContextMsBuildPropertiesAsync(IProjectProperties defaultProperties, string[] msBuildPropertyNames)
```
drewnoakes, line 32: Super nit: .NET conventions have two letter acronyms capitalized, while longer ones are not. So this name should be `MSBuild`, not `MsBuild`.

This allows you to read something like "MSUI" as two words, while "MSSoapXml" is clearly three words.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptedProjectProperties.cs
```diff
@@ -98,5 +110,34 @@ public void SetRuleContext(Rule rule)
...
+        {
+            if (valueTuple.Filtered)
+            {
+                return ReturnProviderAfterFiltering();
+            }
+
+            valueTuple.Providers.RemoveAll(lazyProvider =>
+            {
+                string? appliesToExpression = lazyProvider.Value.GetType().GetCustomAttributes(true).OfType<AppliesToAttribute>().FirstOrDefault()?.AppliesTo;
```
drewnoakes, line 126:
```suggestion
                string? appliesToExpression = lazyProvider.Value.GetType().GetCustomAttributes(typeof(AppliesToAttribute), inherit: true).OfType<AppliesToAttribute>().FirstOrDefault()?.AppliesTo;
```

- named argument
- specify type of interest to reduce work under the covers

This line's getting quite long so you may like to wrap it.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/InterceptedProjectProperties.cs
```diff
@@ -98,5 +110,34 @@ public void SetRuleContext(Rule rule)
...
+
+                return appliesToExpression is not null && !_project.Capabilities.AppliesTo(appliesToExpression);
+            });
+
+            valueTuple.Filtered = true;
```
drewnoakes, line 131: I don't think this will work. Value tuples are passed by value, so modifying the value here will not stick. It's only modifying a copy, and the original source remains unchanged.

My earlier comment about pulling this out to its own type stands, only that type would become a `class` and the property made read/write.

MiYanni (approved):

PR #8565, opened 2022-10-04
Startup Object fix

melytc:
Addresses this [feedback ticket](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1631593).

When I introduced the changes in https://github.com/dotnet/project-system/pull/8419, I tried to avoid the namespace duplication in the code generated by the myapp file to the Application.Designer.vb file; I did this by modifying the EnumProvider to get the entry points without the root namespace in their name.

As the interceptor's filtering (`AppliesTo(ProjectCapability.VisualBasic)`) is not currently doing the actual filtering, C# users were affected by this and have discovered this produced a build to fail.

This PR reverts the modification I made to the EnumProvider; instead, we will remove the namespace when setting the property and identifying is a VB WinForms project.

I have been doing manual testing for the following scenarios:

| | Console | WinForms| 
| ---            | :--        | :--       | 
| C#            | ✅      | 
| VB            | ✅      | ✅

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8565)


tmeschter:
If this is the same as #8566 you could just merge that change and then wait for dev17.4.x to merge to main.

PR #8566, opened 2022-10-04
[17.4] Startup Object fix

melytc:
Addresses this [feedback ticket](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1631593).

When I introduced the changes in https://github.com/dotnet/project-system/pull/8419, I tried to avoid the namespace duplication in the code generated by the myapp file to the Application.Designer.vb file; I did this by modifying the EnumProvider to get the entry points without the root namespace in their name.

As the interceptor's filtering (`AppliesTo(ProjectCapability.VisualBasic)`) is not currently doing the actual filtering, C# users were affected by this and have discovered this produced a build to fail.

This PR reverts the modification I made to the EnumProvider; instead, we will remove the namespace when setting the property and identifying it is a VB WinForms project.

I have been doing manual testing for the following scenarios:

| | Console | WinForms| 
| ---            | :--        | :--       | 
| C#            | ✅      | ✅
| VB            | ✅      | ✅

Unit tests will follow this PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8566)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
@@ -33,56 +33,65 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
     {
         IProjectCapabilitiesScope capabilities = _project.Capabilities;
         bool isWindowsForms = capabilities.Contains(ProjectCapability.WindowsForms);
+        bool isVB = capabilities.Contains(ProjectCapability.VisualBasic);
         string rootNameSpace = await defaultProperties.GetEvaluatedPropertyValueAsync(RootNamespaceProperty);
 
-        if (isWindowsForms)
+        if (isVB)
         {
-            string applicationFrameworkValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkProperty);
-
-            if (string.Compare(applicationFrameworkValue, EnabledValue) == 0)
+            if (isWindowsForms)
             {
-                // Set the startup object in the myapp file.
-                if (unevaluatedPropertyValue.StartsWith(rootNameSpace + ".", StringComparison.OrdinalIgnoreCase))
-                {
-                    unevaluatedPropertyValue = unevaluatedPropertyValue.Substring(rootNameSpace.Length + 1);
-                }
-                await _myAppXmlFileAccessor.SetMainFormAsync(unevaluatedPropertyValue);
+                string applicationFrameworkValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkProperty);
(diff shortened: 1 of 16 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -17,27 +17,36 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
...
+        IProjectCapabilitiesScope capabilities = _project.Capabilities;
+        bool isWindowsForms = capabilities.Contains(ProjectCapability.WindowsForms);
         string rootNameSpace = await defaultProperties.GetEvaluatedPropertyValueAsync(RootNamespaceProperty);
 
-        if (bool.TryParse(isWindowsFormsProject, out bool b) && b)
+        if (isWindowsForms)
         {
             string applicationFrameworkValue = await defaultProperties.GetEvaluatedPropertyValueAsync(ApplicationFrameworkProperty);
 
             if (Equals(applicationFrameworkValue, EnabledValue))
```
drewnoakes, line 43: This should use an explicit string comparer (I know this pre-dates this PR).

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
```diff
@@ -17,27 +17,36 @@ internal class StartupObjectValueProvider : InterceptingPropertyValueProviderBas
...
             {
                 // Set the startup object in the myapp file.
+                if(unevaluatedPropertyValue.StartsWith(rootNameSpace + ".", StringComparison.OrdinalIgnoreCase))
+                {
+                    unevaluatedPropertyValue = unevaluatedPropertyValue.Substring((rootNameSpace + ".").Length);
```
drewnoakes, line 48: This string allocation could be avoided. You could either re-use the value computed in the `if` condition, or just `rootNameSpace.Length + 1`.

drewnoakes (commented):
Looks good! Just some super minor stuff to consider.

drewnoakes (approved):
Oops, meant to approve this.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Properties/InterceptedProjectProperties/StartupObjectValueProvider.cs
melytc: Do you mean for the `isWindowsForms`? I'm a little confused, I changed it into a bool that checks the presence of the `ProjectCapability`, my sense was that it doesn't need it anymore.

PR #8579, opened 2022-10-07
Eliminate some allocations

drewnoakes:
1. **Avoid allocating task on each call.**

    Reuse our singleton instance.

2. **Avoid allocating a copy of the collection.**

   The runtime type returned by `GetDirectPropertyNamesAsync` is actually `List<string>`. On my machine it typically has 108 values. Rather than calling `ToArray` to copy it, we can sniff for `IReadOnlyCollection<string>` and use that interface directly.

   If the concrete type changes in future, we will fall back to calling `ToList`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8579)


MiYanni (approved):

drewnoakes:
Build test due to known flakey `WriteAsync_CompletesWhenDataReceived` test.

PR #8582, opened 2022-10-10
[17.4] No fault when insufficient data to create workspace

### LINKED ISSUE #8343: Investigate change in language service behaviour when required project data is not available
#8321 changes how we initialize language service workspaces.

When a project does not contain enough information to create an `IWorkspaceProjectContext` we throw an exception that will tear down the slice's dataflow.

The previous code handled this differently. We should investigate this difference.

(Prompted by @davkean in https://github.com/dotnet/project-system/pull/8340#discussion_r928487412)

drewnoakes:
Fixes #8343

There may be projects that do not provide enough information for us to create the workspace. An example of this is `.ilproj`. We should not fault the project in such cases by throwing. Instead, we dispose the workspace which tears down its dataflow and cleans up gracefully.

This restores the behaviour we had prior to the changes made to the language service code to improve performance of configuration switches.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8582)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
@@ -193,6 +193,12 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
             _seenEvaluation = true;
 
             await ProcessInitialEvaluationDataAsync(_unloadCancellationToken);
+
+            // If initialisation failed gracefully, we will have been disposed and should just return here.
+            if (IsDisposed)
+            {
+                return;
+            }
         }
 
         await OnProjectChangedAsync(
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -218,10 +224,13 @@ async Task ProcessInitialEvaluationDataAsync(CancellationToken cancellationToken
 
             if (string.IsNullOrEmpty(languageName) || string.IsNullOrEmpty(binOutputPath) || string.IsNullOrEmpty(projectFilePath))
             {
-                // Insufficient data to initialize the language service.
-                _state = WorkspaceState.Failed;
+                // It's reasonable for a project to exist that doesn't contain enough data to initialize the language
+                // service. For example, the ".ilproj" project is not uncommon and hits this code path.
+                // In such cases, we don't want to throw here as that would fault the project. Instead, we
+                // dispose the workspace tears down the dataflow subscription and clean everything up gracefully.
+                Dispose();
```
tmeschter, line 231: Would it be possible to log a message about this somewhere? I worry about misconfigured .csproj/.vbproj/.fsproj projects; previously we would have thrown an exception, but now everything might appear to work and you just don't get Intellisense.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
drewnoakes: > previously we would have thrown an exception

In 17.3 (prior to the perf work on the language service integration) there was no exception and no log message, so this PR just restores the old behaviour. The exception was a regression in the 17.4 previews until now.

Logging a message might make sense in some cases, however in other cases it might not. The customer who discovered this bug in 17.4 preview 2.1 was using an `.ilproj` which doesn't fire up Roslyn. Logging a message for such projects could be noisy, depending upon how we do it.

Can you think of a good way to balance these?

PR #8585, opened 2022-10-10
Obsolete ILaunchSettingsUIProvider

tmeschter:
`ILaunchSettingsUIProvider` used to be the way custom UI could be injected into the "Debug" property page, e.g., if you needed to add controls specific to a particular launch profile command name. Starting in 17.0 we no longer use that "Debug" page, as all launch settings go through the new Launch Profiles UI.

This change marks the interface as `Obsolete`, deletes our own implementations, deletes everything that directly depended on it, and deletes a bunch of code that is now dead.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8585)


drewnoakes (approved):
Nice!

Issue #8548, opened 2022-09-29
Project Query API changes


kvenkatrajan:
These include the following changes:
- [ ] Namespace changes
- [ ] Type name changes from I[ModelType] to I[ModelType]Snapshot
- [ ] Rename from SourceFile to File
- [ ] StartQueryAsync() -> QueryAsync()

PR #8589, opened 2022-10-11
Update CPS query version, use updated query namings, update comments

adamint:
These are the complete set of code and comment changes needed to address CPS' new namings. In reviewing, if you'd like to try out some manual scenarios, please go here: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/427756?_a=files - check out the PR branch and build.

I've tested property pages, launch profiles, running, and dependency node for C# projects.

#8548

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8589)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/CategoryDataProvider.cs
```diff
@@ -1,20 +1,20 @@
...
+using Microsoft.VisualStudio.ProjectSystem.Query.Framework;
+using Microsoft.VisualStudio.ProjectSystem.Query.Metadata;
 using Microsoft.VisualStudio.ProjectSystem.Query.Providers;
-using Microsoft.VisualStudio.ProjectSystem.Query.QueryExecution;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Query
 {
     /// <summary>
-    /// Creates <see cref="IQueryDataProducer{TRequest, TResult}"/> instances that retrieve category information (<see
+    /// Creates <see cref="IQueryDataProducer{TResult}"/> instances that retrieve category information (<see
```
tmeschter, line 13: The deletion of `TRequest` seems wrong.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/CategoryFromRuleDataProducer.cs
```diff
@@ -1,8 +1,10 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using Microsoft.VisualStudio.OLE.Interop;
```
tmeschter, line 3: Where did this come from?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageFromProjectDataProducer.cs
```diff
@@ -1,13 +1,13 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using Microsoft.VisualStudio.OLE.Interop;
```
tmeschter, line 3: This is another questionable namespace. If this is actually used, then some type name in the file is no longer binding to the same type.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/UIEditorMetadataDataProvider.cs
```diff
@@ -1,20 +1,20 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using Microsoft.CodeAnalysis;
```
tmeschter, line 3: That also seems wrong; that's a Roslyn namespace.

tmeschter (changes requested):
There are now a bunch of `using Microsoft.VisualStudio.OLE.Interop;` directives which I believe are wrong; we don't want types to bind to anything in that namespace.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/PropertyPageFromProjectDataProducer.cs
adamint: Yep, not used. These were all incorrect comment imports.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/PropertyPages/CategoryDataProvider.cs
adamint: updated comment imports. thanks for catching! not sure why this was deleted

adamint:
investigating test failures

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
```diff
@@ -25,6 +25,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudio.Setup.Configuration.Interop" />
```
drewnoakes, line 28: Why do we need to reference this?

adamint: Unclear, but it seems like it's no longer a transitive dependency.

eng/imports/HostAgnostic.props
```diff
@@ -36,12 +36,16 @@
 
     <!-- CPS -->
     <PackageReference Include="Microsoft.VisualStudio.ProjectSystem" />
+    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Implementation" />
```
adrianvmsft, line 39: The implementation package should only be used by unit tests projects, not by product projects. The risk is to accidentally take dependencies on Implementation code.

eng/imports/HostAgnostic.props
```diff
@@ -36,12 +36,16 @@
...
+    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.SDK" />
```
adrianvmsft, line 40: If you use the SDK metapackage, you should be able to remove the package reference to Microsoft.VisualStudio.ProjectSystem (as the SDK metapackage links to both)

eng/imports/Packages.targets
```diff
@@ -18,49 +18,51 @@
...
+    <PackageReference Include="StreamJsonRpc"                                                         Version="2.13.21-alpha" />
 
     <!-- CPS -->
     <!-- Find versions at https://dev.azure.com/azure-public/vside/_artifacts/feed/vs-impl -->
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem"                                   Version="17.4.38-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Analyzers"                         Version="17.4.38-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.Query"                             Version="17.4.38-pre" />
-    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem.SDK.Tools"                         Version="17.4.38-pre" />
-
+    <PackageReference Update="Microsoft.VisualStudio.ProjectSystem"                                   Version="17.5.202-pre-g89e17c9f72" />
```
adrianvmsft, line 59: Same comments as above - is the Implementation package reference really needed.
If using SDK, you can remove the reference to MS.VS.PS

eng/imports/HostAgnostic.props
```diff
@@ -35,13 +35,15 @@
...
-    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem" />
+    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.SDK" />
     <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Query" />
     <PackageReference Include="Microsoft.VisualStudio.Validation" />
     <PackageReference Include="Microsoft.VisualStudio.Threading" />
     <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.SDK.Tools" />
     <PackageReference Include="System.Threading.Tasks.Dataflow" />
-
+    
+    <PackageReference Include="Microsoft.Internal.Performance.CodeMarkers.DesignTime" />
```
tmeschter, line 45: Why is this `PackageReference` needed?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
tmeschter, line 28: That also seems odd.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/IProjectLaunchProfileHandler.cs
```diff
@@ -1,9 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
```
tmeschter, line 4: There are two `ILaunchProfile` interfaces: the one specific to the Query API (defined by CPS) and the one coming from our code which represents a launch profile internally. I suspect `ILaunchProfile` used to bind to the former, but with this using statement probably now binds to the latter.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Query/LaunchProfiles/LaunchProfileActionProvider.cs
```diff
@@ -1,17 +1,17 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using Microsoft.VisualStudio.ProjectSystem.Query;
-using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModel.Implementation;
-using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModel.Metadata;
-using Microsoft.VisualStudio.ProjectSystem.Query.ProjectModelMethods.Actions;
+using Microsoft.VisualStudio.ProjectSystem.Debug;
```
tmeschter, line 3: Same issue with `ILaunchProfile` as above.

tmeschter (commented):
I think there are still some `using <Foo>;` directives that are incorrect. I called out a couple, but in general any _new_ using directives that start with something other than "Microsoft.VisualStudio.ProjectSystem.Query" are probably causing type names to bind to something different now.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
adamint: This was required after updating packages.

eng/imports/HostAgnostic.props
adamint: The reference was required after updating the CPS/dependent packages. I don't know which package changed to require us to add this

drewnoakes: What error message do you get if you remove it?

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IProjectThreadingServiceFactory.ProjectThreadingService.cs
```diff
@@ -12,7 +13,9 @@ private class ProjectThreadingService : IProjectThreadingService
 
             public ProjectThreadingService(bool verifyOnUIThread = true) => _verifyOnUIThread = verifyOnUIThread;
 
+#pragma warning disable VSSDK005
```
drewnoakes, line 16:
```suggestion
#pragma warning disable VSSDK005 // Use the JoinableTaskContext singleton
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IProjectThreadingServiceFactory.ProjectThreadingService.cs
```diff
@@ -12,7 +13,9 @@ private class ProjectThreadingService : IProjectThreadingService
...
             public JoinableTaskContextNode JoinableTaskContext { get; } = new JoinableTaskContextNode(new JoinableTaskContext());
+#pragma warning restore VSSDK005
```
drewnoakes, line 18:
```suggestion
#pragma warning disable VSSDK005 // Use the JoinableTaskContext singleton
            public JoinableTaskContextNode JoinableTaskContext { get; } = new JoinableTaskContextNode(new JoinableTaskContext());
#pragma warning restore VSSDK005 // Use the JoinableTaskContext singleton
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Mocks/IProjectThreadingServiceFactory.ProjectThreadingService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
 using Microsoft.VisualStudio.Threading;
+using Task = System.Threading.Tasks.Task;
```
drewnoakes, line 4: I'm curious why this is needed here (and in other places I think).

drewnoakes:
There is some urgency on getting this in.

@adamint and I looked through the pending review feedback on this PR and concluded that none of it impacts the shipping project.

We're going to file an issue to track the pending work from the review comments and merge this in to get an insertion running and unblock other teams.

Issue #8592, opened 2022-10-12
WriteAsync_CompletesWhenDataReceived fails intermittently


tmeschter:
The WriteAsync_CompletesWhenDataReceived unit test is failing intermittently. We've see this in both PR and CI builds.

I'm going to disable this test, so this issue track investigating the failure, fixing, and re-enabling the test.

drewnoakes:
@MiYanni found examples in CI of this failing with both `true` and `false` arguments.

drewnoakes:
I've been playing with this for a bit.

If I run this locally I think I've just found a repro in the debugger.

The `writeTask` here has an exception:

> JoinableTaskContextException: An attempt to switch to the main thread failed to reach the expected thread. Was the JoinableTaskContext initialized on the wrong thread or with a SynchronizationContext whose Post method does not execute its delegate on the main thread?

Abbreviated stack trace:

```
Microsoft.VisualStudio.Threading.JoinableTaskContextException: An attempt to switch to the main thread failed to reach the expected thread. Was the JoinableTaskContext initialized on the wrong thread or with a SynchronizationContext whose Post method does not execute its delegate on the main thread?
   at Microsoft.VisualStudio.Threading.JoinableTaskFactory.MainThreadAwaiter.GetResult()
   at Microsoft.VisualStudio.Threading.ReentrantSemaphore.StackSemaphore.<>c__DisplayClass3_0.<<ExecuteAsync>b__0>d.MoveNext()
   at Microsoft.VisualStudio.Threading.ReentrantSemaphore.StackSemaphore.<ExecuteAsync>d__3.MoveNext()
   at Microsoft.VisualStudio.ProjectSystem.OnceInitializedOnceDisposedUnderLockAsync.<ExecuteUnderLockCoreAsync>d__6`1.MoveNext() in D:\repos\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\OnceInitializedOnceDisposedUnderLockAsync.cs:line 127
   at Microsoft.VisualStudio.ProjectSystem.LanguageServices.Workspace.<WriteAsync>d__45`1.MoveNext() in D:\repos\project-system\src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\LanguageServices\Workspace.cs:line 538<---
```

Full stack trace:

```
[...]

PR #8593, opened 2022-10-12
Skip WriteAsync_CompletesWhenDataReceived

tmeschter:
Related to #8592.

This test fails intermittently causing PR and CI builds to fail. Here we disable the test until it can be investigated and fixed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8593)


MiYanni (approved):
I was literally just about to do this right now. It is causing too many builds to fail. Thank you!

melytc (approved):

Issue #8576, opened 2022-10-06
Refactor ProjectSystem/VS/Tree


tmeschter:
Refactor the types in [ProjectSystem/VS/Tree](https://github.com/dotnet/project-system/tree/main/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree) to isolate the VS dependencies behind a new interface, and then move the implementation down to the Microsoft.VisualStudio.ProjectSystem.Managed project.

ocallesp:
The remaining code are not using API that depends on VS, but the code only applies to the behavior of Solution Explorer and Commands in VS.

PR #8594, opened 2022-10-12
Refactor ProjectSystem/VS/Tree

ocallesp:
Refactor `FileIconProvider` type to isolate dependencies with Vs
Move `FileIconProvider` to `Microsoft.VisualStudio.ProjectSystem.Managed` project.

This is part of https://github.com/dotnet/project-system/issues/8576

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8594)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/UIImageService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/UIImageService.cs
@@ -3,6 +3,7 @@
 using Microsoft.VisualStudio.Imaging.Interop;
 using Microsoft.VisualStudio.ProjectSystem.Tree;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
 
 namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree;
 
@@ -13,15 +14,19 @@ namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree;
 internal class UIImageService : IUIImageService
 {
     private readonly IVsUIService<SVsImageService, IVsImageService2> _vsImageService;
+    private readonly JoinableTaskContext _context;
 
     [ImportingConstructor]
-    public UIImageService(IVsUIService<SVsImageService, IVsImageService2> vsImageService)
+    public UIImageService(IVsUIService<SVsImageService, IVsImageService2> vsImageService, JoinableTaskContext context)
     {
         _vsImageService = vsImageService;
+        _context = context;
     }
 
     public ImageMoniker GetImageMonikerForFile(string filename)
     {
+        _context.VerifyIsOnMainThread();
+
         return _vsImageService.Value.GetImageMonikerForFile(filename);
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/UIImageService.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/UIImageService.cs
@@ -1,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Imaging.Interop;
+using Microsoft.VisualStudio.ProjectSystem.Tree;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree;
+
+/// <summary>
+/// Wrapper around the VS Service that provides ImageMonikers
+/// </summary>
+[Export(typeof(IUIImageService))]
+internal class UIImageService : IUIImageService
+{
+    private readonly IVsUIService<SVsImageService, IVsImageService2> _vsImageService;
+
+    [ImportingConstructor]
+    public UIImageService(IVsUIService<SVsImageService, IVsImageService2> vsImageService)
+    {
+        _vsImageService = vsImageService;
+    }
+
+    public ImageMoniker GetImageMonikerForFile(string filename)
+    {
+        return _vsImageService.Value.GetImageMonikerForFile(filename);
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/IUIImageService.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.Tree;
+
+internal interface IUIImageService : IVsUIService<SVsImageService, IVsImageService2>
```
ocallesp, line 7: @tmeschter " isolate the VS dependencies behind a new interface"

do we want to just put this behind a new interface 
or 
do we want to create a new mef exported service to complete isolate the vs service and only expose specific methods ?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Shipped.txt
```diff
@@ -333,8 +333,6 @@ Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.Re
 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.RelatableItemBase.RelatableItemBase(string! name) -> void
 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.RelationBase<TParent, TChild>
 Microsoft.VisualStudio.ProjectSystem.VS.Tree.Dependencies.AttachedCollections.RelationBase<TParent, TChild>.RelationBase() -> void
-Microsoft.VisualStudio.ProjectSystem.VS.Tree.IFileIconProvider
-Microsoft.VisualStudio.ProjectSystem.VS.Tree.IFileIconProvider.GetFileExtensionImageMoniker(string! path) -> Microsoft.VisualStudio.Imaging.Interop.ImageMoniker
```
tmeschter, line 337: Hmm. I hadn't realized there were public types that we might need to move. This would be a breaking change for anyone depending on the `IFileIconProvider` interface.

There are a couple of possibilities:
1. If we can prove that no one else uses or implements this interface we can just move it down to Managed.csproj as you have done here. Proving that might be difficult.
2. If not, we can move it and add a [`TypeForwardedTo` attribute](https://learn.microsoft.com/en-us/dotnet/standard/assembly/type-forwarding) to tell the runtime where to find the new version of the type. That requires that we keep the name of the type the same, including the namespace.

tmeschter: @drewnoakes Any suggestions for where we might look to find implementations or uses of `IFileIconProvider` outside of our repo?

We could also play it safe and just assume we need to go with the `TypeForwardedTo` attribute.

tmeschter:
This looks good to me but we need to figure out what to do about the public `IFileIconProvider` before we can merge it.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/PublicAPI.Shipped.txt
ocallesp: @tmeschter  I don't know if we can trust on Index, but it shows that no one else is using `IFileIconProvider`

ocallesp: Added `TypeForwardedTo` attribute

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/FileIconProvider.cs
```diff
@@ -2,21 +2,21 @@
...
+        private readonly IUIImageService _imageService;
 
         private ImmutableDictionary<string, ImageMoniker> _imageMonikerByExtensions = ImmutableDictionary.Create<string, ImageMoniker>(StringComparers.Paths);
 
         [ImportingConstructor]
-        public FileIconProvider(IVsUIService<SVsImageService, IVsImageService2> vsImageService)
+        public FileIconProvider(IUIImageService vsImageService)
         {
-            _vsImageService = vsImageService;
+            _imageService = vsImageService;
```
drewnoakes, line 19:
```suggestion
            _imageService = imageService;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/UIImageService.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    [ImportingConstructor]
+    public UIImageService(IVsUIService<SVsImageService, IVsImageService2> vsImageService)
+    {
+        _vsImageService = vsImageService;
+    }
+
+    public ImageMoniker GetImageMonikerForFile(string filename)
+    {
+        return _vsImageService.Value.GetImageMonikerForFile(filename);
```
drewnoakes, line 25: This requires the main thread, which should be asserted and documented in the interface.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/IUIImageService.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.Composition;
+using Microsoft.VisualStudio.Imaging.Interop;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Tree;
+
+[ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne, Provider = ProjectSystemContractProvider.Private)]
```
drewnoakes, line 8: It's the cardinality actually "zero or one"? Can we assume the VS later always provides an export?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/FileIconProvider.cs
ocallesp: fixed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/UIImageService.cs
ocallesp: done.

The implementation now verifies that it is executing on the main thread, the interface has a remarks saying that is required the main thread.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/IUIImageService.cs
ocallesp: What can change in the behavior if cardinality is "zero or one" ?

My understanding is that `_imageService` can be null and throw a null pointer dereference in `GetFileExtensionImageMoniker`. 
This might help to spot an issue quickly rather than using mef logs.

Is this what we want ?

tmeschter (approved):

Issue #8573, opened 2022-10-06
Refactor ProjectSystem/VS/LanguageServices


tmeschter:
Refactor the types in [ProjectSystem/VS/LanguageServices](https://github.com/dotnet/project-system/tree/main/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices) to isolate the VS dependencies behind a new interface, and then move the implementation down to the Microsoft.VisualStudio.ProjectSystem.Managed project.

PR #8595, opened 2022-10-12
Refactor ActiveEditorContextTracker

MiYanni:
Fixes: https://github.com/dotnet/project-system/issues/8573

After reviewing the files with Tom and Osvaldo, the only class that had non-VS specific logic was `ActiveEditorContextTracker`. I've split this class into two, putting the non-VS specific logic into *ActiveEditorContextTracker.cs* and VS-specific logic into ***Vs****ActiveEditorContextTracker.cs*. No logic was changed. The VS-specific one acquires the non-VS one via MEF.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8595)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/VsActiveEditorContextTracker.cs
```diff
@@ -0,0 +1,51 @@
...
+            pbstrProjectName = ActiveIntellisenseProjectContext;
+
+            return HResult.OK;
+        }
+
+        // Pass-through for unit testing.
+        internal bool IsActiveEditorContext(string contextId) => _activeEditorContextTracker.IsActiveEditorContext(contextId);
+
+        // Pass-through for unit testing.
+        internal IDisposable RegisterContext(string contextId) => _activeEditorContextTracker.RegisterContext(contextId);
```
MiYanni, line 49: These are internal in an internal class because they never are intended to be public (only used for unit testing).

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IActiveEditorContextTracker.cs
```diff
@@ -22,6 +22,11 @@ namespace Microsoft.VisualStudio.ProjectSystem.LanguageServices
     [ProjectSystemContract(ProjectSystemContractScope.UnconfiguredProject, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne)]
     internal interface IActiveEditorContextTracker
     {
+        /// <summary>
+        ///     Gets or sets the current shared item context intellisense project name. Can be null.
+        /// </summary>
+        string? ActiveIntellisenseProjectContext { get; set; }
```
MiYanni, line 28: This was added since it is needed for the non-VS specific implementation. We pass this through to the VS-specific interface.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/VsActiveEditorContextTracker.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS.LanguageServices
+{
+    /// <summary>
+    ///     Tracks the "active" context for the editor by handling the VSHPROPID_ActiveIntellisenseProjectContext hierarchy property.
+    /// </summary>
+    [Export(typeof(IActiveIntellisenseProjectProvider))]
+    [ExportProjectNodeComService(typeof(IVsContainedLanguageProjectNameProvider))]
+    [AppliesTo(ProjectCapability.DotNetLanguageService)]
+    internal class VsActiveEditorContextTracker : IActiveIntellisenseProjectProvider, IVsContainedLanguageProjectNameProvider
```
MiYanni, line 15: These 2 interfaces are VS-specific.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/LanguageServices/VsActiveEditorContextTracker.cs
MiYanni: I'm a bit confused by what a mock would be providing here. Right now, the unit tests create actual instances of both `VsActiveEditorContextTracker` and `ActiveEditorContextTracker`. Both instances are accessible to the tests. I did it this way to reduce the changes necessary to the tests. Instead, I could update the tests to access both class instances directly. My original changes here were actually to expose the composed `ActiveEditorContextTracker` within the class as internal, but that requires adding `._activeEditorContextTracker` to every call of these methods in the tests.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/IActiveEditorContextTracker.cs
MiYanni: > Do you know if this is a full path to the project on disk?

Based on the error message, it is likely just some kind of ID (like a GUID). I didn't run/debug a build with these changes.

> And how does it related to the `contextId` parameter pass to `IsActiveEditorContext`?

They are one in the same. The class stores a collection of context IDs, and `IsActiveEditorContext` checks to see if the `contextId` passed in equals the one stored in `ActiveIntellisenseProjectContext`.

tmeschter (approved):

PR #8596, opened 2022-10-13
Fix AssemblyVersions.tt not updating properly

MiYanni:
Related to: https://github.com/dotnet/project-system/issues/7911

When we updated the repo to `17.5`, the process for updating the `AssemblyVersions.tt` was not in place. This is my fault, as it was a task I needed to do when changing out our insertion process. I simply forgot 😅. To my credit, I did have a `TODO:` comment to do this work.

In this PR, we add the `AssemblyVersion` value and the appropriate identifiers to the insertion process; which will update the `AssemblyVersions.tt` file in the VS repo when we update the version number. I remove the legacy process (based on RoslynInsertionTool), which required a script that created a CSV file, which is no longer needed. I also update comments related to RoslynInsertionTool.

### Prior Insertion Updating `AssemblyVersions.tt`
You can see the 3 values being updated here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/409144?_a=files

### Docs
You can look at how `InsertVersionsValues` is [used here](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/635/Overview?anchor=**build-pipeline**).
(That link is already provided above the task in the YAML)
image

### Test Insertion Build
- ~~https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6818224&view=results~~
  - First build failed with: `Skipping update of D:\a\_work\1\s\VS\src\ProductData\AssemblyVersions.tt or any related files: No values supplied.`
- https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6821807&view=results
  - Success: 
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8596)


drewnoakes (approved):

eng/pipelines/official.yml
```diff
@@ -251,6 +251,8 @@ stages:
     variables:
       # Gets the PackageVersion variable produced by the Build pipeline.
       PackageVersion: $[ stageDependencies.Build.BuildOfficialRelease.outputs['SetPackageVersion.PackageVersion'] ]
+      # Gets the AssemblyVersion variable produced by the Build pipeline.
+      AssemblyVersion: $[ stageDependencies.Build.BuildOfficialRelease.outputs['SetAssemblyVersion.AssemblyVersion'] ]
```
tmeschter, line 255: Just so I'm sure I understand how this works, `SetAssemblyVersion` here refers to the name of the task that set the `AssemblyVersion` variable?

tmeschter (approved):

eng/pipelines/official.yml
MiYanni: Correct. The syntax actually becomes more confusing based on the scenario (meaning, where and how the variable needs to be accessed). But, this syntax is what works in `stage` variables.

PR #8598, opened 2022-10-13
Bump compiler and analyzer versions

drewnoakes:
The newer compiler and analyzers produced new diagnostics on our code, which have also been addressed to remain free of warnings.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8598)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ExternalContracts.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ExternalContracts.cs
@@ -9,6 +9,8 @@ using Microsoft.VisualStudio.Shell.Interop;
 using NuGet.SolutionRestoreManager;
 using NuGet.VisualStudio;
 
+#pragma warning disable RS0030 // Do not used banned APIs
+
 [assembly: ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne, ContractType = typeof(SVsServiceProvider))]
 [assembly: ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne, ContractType = typeof(SAsyncServiceProvider))]
 [assembly: ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.Private, Cardinality = ImportCardinality.ExactlyOne, ContractType = typeof(VisualStudioWorkspace))]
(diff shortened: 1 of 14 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ExternalContracts.cs
```diff
@@ -9,6 +9,8 @@
 using NuGet.SolutionRestoreManager;
 using NuGet.VisualStudio;
 
+#pragma warning disable RS0030 // Do not used banned APIs
```
tmeschter, line 12: Which banned API was it complaining about?

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/WPF/Converters/Converters.LambdaConverter.cs
```diff
@@ -15,15 +15,15 @@ internal static partial class Converters
...
+            private readonly bool _convertFromNull;
 
             public LambdaConverter(Func<TFrom, TTo> convert, Func<TTo, TFrom>? convertBack = null, bool convertFromNull = true)
             {
-                this.convert = convert;
-                this.convertBack = convertBack;
-                this.convertFromNull = convertFromNull;
+                this._convert = convert;
+                this._convertBack = convertBack;
+                this._convertFromNull = convertFromNull;
```
tmeschter, line 26: Nit: still has `this`.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ExternalContracts.cs
drewnoakes: Three different ones. I didn't think it was useful to highlight them in this file.

tmeschter, line 12: Should there be a `#pragma warning restore RS0030` somewhere? I assume it doesn't really matter in this file.

drewnoakes: I don't think it's necessary in this file. So long as we only ever model external contracts here.

This file's just adding metadata to external contracts, which is still useful, even if we have banned their use.

Most of these banned contracts are actually used indirectly, where we wrap them in an unbanned API that gives some guarantee we want to have, and suppress RS0030 on that usage.

PR #8600, opened 2022-10-13
Move ISafeProjectGuidService implementation

tmeschter:
Move `VsSafeProjectGuidService`, the implementation of `ISafeProjectGuidService`, up to the Microsoft.VisualStudio.ProjectSystem.Managed.VS project. It uses the `IProjectGuidService` interface which comes from the VS layer of CPS; as such we should avoid referencing it in our lower layer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8600)


ocallesp (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ISafeProjectGuidService.cs
```diff
@@ -1,21 +1,18 @@
...
     /// </summary>
     /// <remarks>
-    ///     <see cref="IProjectGuidService"/> and <see cref="IProjectGuidService2"/> will retrieve the project GUID of
-    ///     the project *at the time* that it is called. During project initialization, the GUID may be changed by the
-    ///     solution in reaction to a clash with another project. <see cref="ISafeProjectGuidService"/> will wait until
-    ///     it is safe to retrieve the project GUID before returning it.
+    ///     IProjectGuidService and IProjectGuidService2 will retrieve the project GUID of the project *at the time*
+    ///     that it is called. During project initialization, the GUID may be changed by the solution in reaction to a
+    ///     clash with another project. <see cref="ISafeProjectGuidService"/> will wait until it is safe to retrieve
+    ///     the project GUID before returning it.
```
MiYanni, line 15: Since `IProjectGuidService` and `IProjectGuidService2` are VS layer specific, shouldn't these comments exist in where `ISafeProjectGuidService` is used and not on this interface?

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/ISafeProjectGuidService.cs
tmeschter: I think this is a better place to explain **why** `ISafeProjectGuidService` exists. But the issues with layering do make this a little funky, yes.

PR #8605, opened 2022-10-14
Refactor - move interfaces to ProjectSystem.Managed project

ocallesp:
These interfaces don't have dependencies with Vs libraries

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8605)


MiYanni (commented):
The issue with these changes is these are public interfaces that we're moving around, which is a breaking change for anyone that depends on our libraries. @tmeschter Are we allowing breaking API changes for the refactor effort?

tmeschter:
There are two types of breaks we need to worry about:

1. Build breaks: Other projects with a reference to Microsoft.VisualStudio.ProjectSystem.Managed.VS.dll _could_ experience a build break after we move these types because they will no longer be in the same location. However, anyone with a reference to MS.VS.PS.Managed.VS.dll probably also has a reference to MS.VS.PS.Managed.dll--especially as they are likely to be getting these through a NuGet package of some sort. So long as we don't change the full name of the type (that is, namespace + name) their build will just pick it up from the new location.
2. Run-time breaks: If another DLL references these types from MS.VS.PS.Managed.VS.dll, but then runs against the newer builds where they exist in MS.VS.PS.Managed.dll, you could get a runtime error. But this is what the `TypeForwardedTo` attribute takes care of. Again, so long as the full name of the type is the same the runtime will redirect to the new location. Note you can only move types to another assembly referenced by the original location.

So we should be OK so long as we keep the full type names the same, we put appropriate type forwarding in place, and we only move public types down from MS.VS.PS.Managed.VS.dll to MS.VS.PS.Managed.dll. But we should de

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/AttachedCollections/IAggregateRelationCollection.cs
```diff
@@ -32,8 +31,8 @@ public interface IAggregateRelationCollection : IList
         /// Raised whenever <see cref="HasItems"/> changes.
         /// </summary>
         /// <remarks>
-        /// Used by <see cref="AggregateRelationCollectionSource"/> to trigger changes to its
-        /// <see cref="IAttachedCollectionSource.HasItems"/> property.
+        /// Used by AggregateRelationCollectionSource to trigger changes to its
+        /// IAttachedCollectionSource.HasItems/> property.
```
tmeschter, line 35: Looks like the "/>" is a typo.

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Tree/Dependencies/AttachedCollections/IAggregateRelationCollection.cs
ocallesp: fixed

PR #8611, opened 2022-10-17
Use ISafeProjectGuidService

tmeschter:
Replace a few uses of the `UnconfiguredProject.GetProjectGuidAsync()` extension method with the `ISafeProjectGuidService` instead. The extension method comes from the VS layer of CPS and so we really shouldn't be using it in these locations anyway, and the `ISafeProjectGuidService` ensures that the project has loaded to the point where the GUID isn't going to change after we get it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8611)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
@@ -1,7 +1,5 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using Microsoft.VisualStudio.ProjectSystem.VS;
-
 namespace Microsoft.VisualStudio.ProjectSystem
 {
     [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectDynamicLoadComponent))]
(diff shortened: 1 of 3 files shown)


MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
```diff
@@ -1,7 +1,5 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-using Microsoft.VisualStudio.ProjectSystem.VS;
```
drewnoakes, line 3: Can we somehow ban use of CPS's VS layer in our non-VS layer?

https://github.com/dotnet/roslyn-analyzers/pull/2161 seems like it might have helped here.

drewnoakes: At the least we could ban that extension method in our VS layer.

drewnoakes (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Runtimes/MissingSdkRuntimeDetector.cs
tmeschter: I believe the CPS NuGet packages are going to be split up such that we won't automatically get a reference to the assemblies in each layer. That will make it easier to prevent this sort of thing--we just won't reference the VS layer from our non-VS layer.

drewnoakes:
This PR is causing a hang when opening projects. I'm going to revert it so the nightly insertion doesn't pick it up.

PR #8617, opened 2022-10-18
Add dataflow block name and some API docs

drewnoakes:
Split out of #8550 to simplify review.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8617)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
@@ -132,7 +132,8 @@ internal sealed class LanguageServiceHost : OnceInitializedOnceDisposedUnderLock
                 target: DataflowBlockFactory.CreateActionBlock<IProjectVersionedValue<(ConfiguredProject ActiveConfiguredProject, ConfigurationSubscriptionSources Sources)>>(
                     async update => await ExecuteUnderLockAsync(cancellationToken => OnSlicesChanged(update, cancellationToken)),
                     _unconfiguredProject,
-                    ProjectFaultSeverity.LimitedFunctionality),
+                    ProjectFaultSeverity.LimitedFunctionality,
+                    "LanguageServiceHostSlices {0}"),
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
```diff
@@ -132,7 +132,8 @@ protected override async Task InitializeCoreAsync(CancellationToken cancellation
                 target: DataflowBlockFactory.CreateActionBlock<IProjectVersionedValue<(ConfiguredProject ActiveConfiguredProject, ConfigurationSubscriptionSources Sources)>>(
                     async update => await ExecuteUnderLockAsync(cancellationToken => OnSlicesChanged(update, cancellationToken)),
                     _unconfiguredProject,
-                    ProjectFaultSeverity.LimitedFunctionality),
+                    ProjectFaultSeverity.LimitedFunctionality,
+                    "LanguageServiceHostSlices {0}"),
```
tmeschter, line 136: What fills in the place holder here, and what value will it have?

tmeschter (approved):

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
drewnoakes: I've only ever seen it used in WinDBG, to help identify the purpose behind each dataflow block in the graph. I'm not sure what value goes in the placeholder actually, or why it's needed, but all `nameFormat` values have that.

drewnoakes: Digging into this, the `nameFormat` is used on the dataflow block's `ToString` method, and is passed two placeholder values:
- `{0}` &mdash; the block's type name
- `{1}` &mdash; the hash code of the block

PR #8619, opened 2022-10-18
LinkLabel on Settings PropPage should be accessible by tab key

bekir-ozturk:
DataGrid in Settings property page traps the keyboard and won't allow selecting the link label using tab key.

The fix repurposes the tab key to be used by navigating between other controls in the page while leaving the job of navigating within the data grid to arrow keys.

Before:
Tab key will get you into the data grid and stops functioning after all the cells in the table are traversed.
accessibility_before

After:
Tab key will first visit the link label, then go into the table. Pressing the tab key will then move back to the label. To navigate within the table, user will need to use the arrow keys instead of the tab key.
accessibility_after

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8619)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsDesignerView.vb
+++ b/src/Microsoft.VisualStudio.Editors/SettingsDesigner/SettingsDesignerView.vb
@@ -9,16 +9,15 @@ Imports System.Web.ClientServices.Providers
 Imports System.Windows.Forms
 Imports System.Windows.Forms.Design
 Imports System.Xml
-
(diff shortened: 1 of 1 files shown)


melytc (approved):

MiYanni (approved):

bekir-ozturk:
Thanks @melytc & @MiYanni .
I'm not authorized to merge. Feel free to do it at your convenience 🙂

MiYanni:
@bekir-ozturk Merged it for ya. Thanks! :shipit:

PR #8622, opened 2022-10-19
Allow Roslyn to request arbitrary project properties

drewnoakes:
Historically, the set of project properties we passed to Roslyn when initialising an `IWorkspaceProjectContext` was hard-coded in the project system. We changed this recently in #8499, and introduced a means for Roslyn to request whatever properties they wanted, allowing Roslyn to make changes without requiring a synchronised change in the project system.

With that change, we serviced property value requests using `ProjectRuleSource` data. This limited Roslyn to only those properties present in our `ConfigurationGeneral` XAML rule file.

This commit uses a `ProjectInstance` to service such requests instead. This means that Roslyn can add new projects in their own props/targets, and not need the project system to make any changes in order to provide those values.

Motivated by a bug exposed through https://github.com/dotnet/roslyn/pull/64749 which added required property `OutputAssemblyForDesignTimeEvaluation` which did not exist in our rules, leading to runtime errors.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8622)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
@@ -285,6 +285,8 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
                     hostObject,
                     cancellationToken);
 
+                evaluationData.Release();
+
                 _contextCreated.TrySetResult();
 
                 _disposableBag.Add(_context);
@@ -584,7 +586,7 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
     /// </summary>
     private sealed class EvaluationDataAdapter : EvaluationData
     {
-        private readonly ProjectInstance _projectInstance;
+        private ProjectInstance? _projectInstance;
 
         public EvaluationDataAdapter(ProjectInstance projectInstance)
         {
@@ -593,12 +595,27 @@ internal sealed class Workspace : OnceInitializedOnceDisposedUnderLockAsync, IWo
 
         public override string GetPropertyValue(string name)
         {
+            if (_projectInstance is null)
+            {
+                throw new ObjectDisposedException(nameof(EvaluationDataAdapter));
+            }
+
             string? value = _projectInstance.GetProperty(name)?.EvaluatedValue;
 
             // Return the empty string rather than null.
             value ??= "";
 
             return value;
+        }
+
+        /// <summary>
+        /// Ensures the inner reference to <see cref="ProjectInstance"/> is cleared,
+        /// so Roslyn cannot accidentally retain a rooted reference to it. It's a large
+        /// object and must be collected as soon as possible.
+        /// </summary>
+        internal void Release()
+        {
+            _projectInstance = null;
         }
     }
 }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
+using Microsoft.Build.Execution;
 using Microsoft.VisualStudio.LanguageServices.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.OperationProgress;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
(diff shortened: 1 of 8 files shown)


adamint (approved):

tmeschter:
Could there be an alternative approach where Roslyn provides the `Rule` specifying the properties they care about? That way they don't have to call us back, and we don't need to hold on to the project snapshot.

davkean:
How does this play with the old support I added for this via https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/LanguageService.xaml?

drewnoakes:
> How does this play with the old support I added for this via https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/LanguageService.xaml?

`LanguageService.xaml` still defines properties that are passed to `IWorkspaceProjectContext.SetProperty(string, string)` throughout the lifetime of the project.

The code in this PR only comes into play during `IWorkspaceProjectContext` _construction_, during which time Roslyn can request whatever they need from the project in order to initialize things as quickly as possible.

drewnoakes:
> Could there be an alternative approach where Roslyn provides the `Rule` specifying the properties they care about? That way they don't have to call us back, and we don't need to hold on to the project snapshot.

That could work, but would require coordinating another wave of changes between components, and I'm not sure it's necessary.

The API we devised allows Roslyn to specify the set of property names they're interested in on the factory itself, and we could use that to pull specific values into a new data structure here, however in the end I didn't see any value in using that facility. It'd just be a bunch more data copying, when using the `ProjectInstance` directly seemed adequate and avoided copies.

With regards to "hold on to the project snapshot", I don't see any reference being held here. The synclink block holds the reference, but that value is already held by the upstream data source. As soon as a new value is available, it is released. Roslyn's factory could theoretically hold on to that reference, but [does not](https://github.com/dotnet/roslyn/blob/120ba9e0e9430703db99b7634d5a0a7b4f275e54/src/VisualStudio/Core/Impl/ProjectSystem/CPS/CPSProjectFactory.cs#L51-L62). Any change in future that did retain a reference would be flagged in nearly every DDRITs scenario involving SDK-style projects.

I've filed #8630 as a less contentious alternative to this PR. I'd be interested in @lifengl's perspective on this PR in next week's meeting with the CPS team, as this approach decouples the components nicely.

drewnoakes:
We discussed this in the meeting with CPS today and concluded that the approach here is sound.

I've rebased and pushed two updates to this PR:

1. Reverts #8630 as this change replaces it.
2. Ensures we null out the `ProjectInstance` reference when the factory returns, so Roslyn cannot inadvertently retain a reference to it in future.

adamint (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/Workspace.cs
```diff
@@ -581,21 +586,36 @@ private async Task WhenContextCreated(CancellationToken cancellationToken)
...
+            {
+                throw new ObjectDisposedException(nameof(EvaluationDataAdapter));
+            }
+
+            string? value = _projectInstance.GetProperty(name)?.EvaluatedValue;
 
             // Return the empty string rather than null.
             value ??= "";
 
             return value;
```
MiYanni, line 608: nit: Reassigning the local value provides no purpose. Potentially remove the comment too, since it says nothing other than what the code itself says.
```suggestion
            // Return the empty string rather than null.
            return value ?? "";
```
Or even more terse would be just `return _projectInstance.GetProperty(name)?.EvaluatedValue ?? "";` since the variable isn't used for anything other than returning.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/LanguageServices/WorkspaceFactoryTests.cs
```diff
@@ -75,7 +75,8 @@ private static IProjectVersionedValue<WorkspaceUpdate> EvaluationUpdate()
         ConfiguredProject configuredProject = ConfiguredProjectFactory.Create();
         IProjectSubscriptionUpdate evaluationRuleUpdate = IProjectSubscriptionUpdateFactory.CreateEmpty();
         IProjectSubscriptionUpdate sourceItemsUpdate = IProjectSubscriptionUpdateFactory.CreateEmpty();
-        var workspaceUpdate = WorkspaceUpdate.FromEvaluation((configuredProject, evaluationRuleUpdate, sourceItemsUpdate));
+        IProjectSnapshot projectSnapshot = IProjectSnapshot2Factory.Create();
+        var workspaceUpdate = WorkspaceUpdate.FromEvaluation((configuredProject, projectSnapshot, evaluationRuleUpdate, sourceItemsUpdate));
```
MiYanni, line 79: *joke* Double parentheses!! WHAT DOES IT MEAN?!?! 🌈🌈

ValueTuples are weird. 🙃

MiYanni (approved):

tmeschter (approved):

PR #8634, opened 2022-10-21
Create VS insertion description without using tags

MiYanni:
Continuation of: https://github.com/dotnet/project-system/pull/8517

## Description
The original solution in the linked PR above did not actually work. The multi-repo trigger mechanism for tagging required that the YAML file also be within the same AzDO org (this is not documented). After explaining this situation to the team, I realized the logic used for the tagging mechanism I made in the previous PR can simply be directly added to the insertion job. Thus, we can create the insertion description without the use of tagging.

The reason I didn't arrive at this solution originally was that checking out the VS repo is quite large and time consuming. Not only that, but we need to checkout up to a certain commit depth so we can find the previous insertion. (*For the sake of this PR, I've set the commit depth to 100. *) I originally dismissed this idea as I thought it would add a lot of overhead to producing insertions. Additionally, the original mechanism for insertions checked out the repo, inspected the files, and parsed data out of them. At that time, I hadn't learned how to filter Git commit history or really do anything as advanced at this with manipulating Git metadata. 

During the creation of the tagger, I realized how long checkout takes for the VS repo and decided to make my own checkout implementation. I tested various Git commands and found a way to only fetch the history and not actually populate the files on-disk. Additionally, I needed to filter the tags for creating the insertion description, and I knew how to parse (RegEx) the commit title information. Using those bits of script, I simply moved them from the tagging mechanism and simplified them so that I could:
- Fetch the VS repo at depth 100
- Find the latest commit with title matching `Insert DotNet-Project-System`
- Parse the injected short commit ID (I added previously) from the commit title

That short commit ID represents the previous insertion we completed. This solution will now **work for branches too**, since the fetch of the VS repo will fetch from the branch that the insertion is targeting. Doing the fetch (and this additional parsing) only adds ~3 minutes to the insertion process. This solution is much simpler than the original design, but the original design got me to this point; so I can't fault it. Live and learn, and continue to grow. 🌳

## Test Insertion
- Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6855498&view=results
  - Insertion: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/430988?_a=overview
image
- Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6855880&view=results
  - Insertion: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/431005?_a=overview
image

### DIFF
--- a/eng/pipelines/templates/generate-insertion.yml
+++ b/eng/pipelines/templates/generate-insertion.yml
@@ -26,9 +26,7 @@ jobs:
   # Create the VS folder and fetch the VS repo Git information into it, excluding the files themselves (only .git history is required).
   # Doing this manual process over the checkout: VS task saves around 3 minutes. See: https://stackoverflow.com/a/43136160/294804
   # The System.AccessToken is required for the clone to occur. See: https://stackoverflow.com/a/56734304/294804
-  # The Git config change is for disabling a warning message because we are fetching a specific commit instead of a branch. See: https://stackoverflow.com/a/36794768/294804
   - powershell: |
-      # git config --global advice.detachedHead false
       $null = New-Item -Path '$(Pipeline.Workspace)' -Name VS -ItemType Directory
       Set-Location '$(Pipeline.Workspace)/VS'
       git init
--- a/eng/pipelines/templates/generate-insertion.yml
+++ b/eng/pipelines/templates/generate-insertion.yml
@@ -3,8 +3,38 @@
 jobs:
 - job: GenerateInsertion
   displayName: Generate Insertion
-  timeoutInMinutes: 30
+  timeoutInMinutes: 45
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/generate-insertion.yml
```diff
@@ -3,9 +3,37 @@
 jobs:
 - job: GenerateInsertion
   displayName: Generate Insertion
-  timeoutInMinutes: 30
+  timeoutInMinutes: 45
```
MiYanni, line 6: Added some extra time for VS repo fetch.

drewnoakes (approved):
Great to have this working again. Thanks so much for fixing this.

Issue #8574, opened 2022-10-06
Refactor ProjectSystem/VS/PackageRestore


tmeschter:
Refactor the types in [ProjectSystem/VS/PackageRestore](https://github.com/dotnet/project-system/tree/main/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/PackageRestore) to isolate the VS dependencies behind a new interface, and then move the implementation down to the Microsoft.VisualStudio.ProjectSystem.Managed project.

ocallesp:
checking this...

PR #8640, opened 2022-10-26
Refactor vs package restore hide vsinterfaces

ocallesp:
This is part of https://github.com/dotnet/project-system/issues/8574

Interfaces coming from the NuGet Client (e.g. [IVsProjectRestoreInfo2](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsProjectRestoreInfo2.cs)) are using the attribute `ComImport`, but this attribute is irrelevant since these interfaces are not pointing to any COM object anymore. This makes the interfaces independent of Vs and the implementations can be moved to `ProjectSystem.Managed` project

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8640)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
@@ -171,15 +171,6 @@ In order to debug this project, add an executable project to this solution which
   <data name="ImportsFromTargetCannotBeDeleted" xml:space="preserve">
     <value>{0} - Import comes fom target file. This import cannot be removed.</value>
   </data>
-  <data name="Restore_PropertyWithInconsistentValues" xml:space="preserve">
-    <value>The value of the '{0}' property is inconsistent between target frameworks. This property must be identical for NuGet restore to function correctly. The value '{1}' from the '{2}' configuration will be used, other target frameworks may fail to pick NuGet assets.</value>
-  </data>
-  <data name="Restore_DuplicateToolReferenceItems" xml:space="preserve">
-    <value>The metadata on 'DotNetCliToolReference' item '{0}' is inconsistent between target frameworks. Only the first one will be restored.</value>
-  </data>
-  <data name="Restore_EmptyTargetFrameworkMoniker" xml:space="preserve">
-    <value>The value of the 'TargetFrameworkMoniker' and 'NuGetTargetMoniker' properties in the '{0}' configuration are both empty. This configuration will not contribute to NuGet restore, which may result in restore and build errors.</value>
-  </data>
   <data name="XprojNotSupported" xml:space="preserve">
     <value>Xproj project files are no longer supported. For details on migrating to csproj see https://docs.microsoft.com/dotnet/core/migration/.</value>
   </data>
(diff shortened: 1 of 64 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx
```diff
@@ -171,15 +171,6 @@ In order to debug this project, add an executable project to this solution which
   <data name="ImportsFromTargetCannotBeDeleted" xml:space="preserve">
     <value>{0} - Import comes fom target file. This import cannot be removed.</value>
   </data>
-  <data name="Restore_PropertyWithInconsistentValues" xml:space="preserve">
```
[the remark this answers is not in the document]
ocallesp: These string resources were moved from `src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/VSResources.resx`  to `src/Microsoft.VisualStudio.ProjectSystem.Managed/Resources.resx` because the code that is using this string resource was moved to ProjectSystem.Managed project.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
```diff
@@ -76,6 +76,9 @@
 
   <ItemGroup>
     <None Include="**\*.xlf" />
+    <Compile Remove="ProjectSystem\VS\PackageRestore\Snapshots\**" />
+    <EmbeddedResource Remove="ProjectSystem\VS\PackageRestore\Snapshots\**" />
+    <None Remove="ProjectSystem\VS\PackageRestore\Snapshots\**" />
```
drewnoakes, line 81: Looks like these paths don't exist any more. Are these lines needed?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Resources.resx
```diff
@@ -127,7 +127,7 @@
   <data name="JsonErrorsNeedToBeCorrected_WithErrorMessage_2" xml:space="preserve">
     <value>Errors in '{0}' need to be corrected before applying changes. Error message:
 {1}</value>
-   <comment>{0} is the path to the malformed .json file. {1} is an error message containing more details about the problems in the file.</comment>
+    <comment>{0} is the path to the malformed .json file. {1} is an error message containing more details about the problems in the file.</comment>
```
drewnoakes, line 130: This seems to be an unrelated change.

drewnoakes:
@nkolev92 the premise of this PR is that type `IVsProjectRestoreInfo2` (and therefore `IVsTargetFrameworks2` and `IVsReferenceItems`) are not tied to VS, despite their names. Does that sound right to you?

src/Microsoft.VisualStudio.ProjectSystem.Managed/Resources.resx
ocallesp: removed

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Microsoft.VisualStudio.ProjectSystem.Managed.VS.csproj
ocallesp: removed

nkolev92:
> @nkolev92 the premise of this PR is that type `IVsProjectRestoreInfo2` (and therefore `IVsTargetFrameworks2` and `IVsReferenceItems`) are not tied to VS, despite their names. Does that sound right to you?

Yep. It's just convention naming that we followed. The API no dependencies, it's only NuGet.

drewnoakes (approved):

Issue #8642, opened 2022-10-27
Fast Up-to-Date Check tests produce different output with `dotnet test`


tmeschter:
I'm looking into moving our unit tests over to use `dotnet test` instead of `xunit.runner.console.x86.exe` and simplifying some of our unit testing infrastructure along the way. This has led to several unit test failures that appear to be due mostly to unstable output:

- IsUpToDateAsync_False_Kinds_InputNewerThanOutput_NoKindIgnored
- IsUpToDateAsync_False_Kinds_InputNewerThanOutput_WithIgnoredKind
- IsUpToDateAsync_True_Kinds_InputNewerThanOutput_NoKindIgnored
- IsUpToDateAsync_True_Kinds_InputNewerThanOutput_WithIgnoredKind

For at least one of these, however, the output has differences beyond ordering issues. For example, in IsUpToDateAsync_False_Kinds_InputNewerThanOutput_NoKindIgnored we see:

```
Expected:
Adding UpToDateCheckOutput outputs:
    C:\Dev\Solution\Project\TaggedOutput
    C:\Dev\Solution\Project\Output
Adding UpToDateCheckBuilt outputs:
    C:\Dev\Solution\Project\TaggedBuilt
    C:\Dev\Solution\Project\Built
Adding project file inputs:
    C:\Dev\Solution\Project\Project.csproj
Adding newest import input:
    C:\Dev\Solution\Project\Project.csproj
Adding UpToDateCheckInput inputs:
    C:\Dev\Solution\Project\TaggedInput
    C:\Dev\Solution\Project\Input
Input UpToDateCheckInput item 'C:\Dev\Solution\Project\Input' is newer (2022-10-26 14:21:34.513) than earliest output 'C:\Dev\Solution\Project\Output' (2022-10-26 14:19:34.513), not up-to-date.

Actual:
Adding UpToDateCheckOutput outputs:
    C:\Dev\Solution\Project\Output
    C:\Dev\Solution\Project\TaggedOutput
Adding UpToDateCheckBuilt outputs:
    C:\Dev\Solution\Project\Built
    C:\Dev\Solution\Project\TaggedBuilt
Adding project file inputs:
    C:\Dev\Solution\Project\Project.csproj
Adding newest import input:
    C:\Dev\Solution\Project\Project.csproj
Adding UpToDateCheckInput inputs:
    C:\Dev\Solution\Project\Input
[...]

PR #8643, opened 2022-10-27
Migrate to dotnet test

tmeschter:
Currently, our unit test projects rely on pulling out xunit.console.x86.exe from the xunit.runner.console NuGet package and then using an `Exec` task to invoke that executable and point it at the output binary. This has worked pretty well for a while now, but it's clear that moving foward xUnit tests are really expected to be invoked using `dotnet test` instead. Here we do some future proofing by making that update.

I also made some simplifications to our unit testing infrastructure. Currently all unit test assemblies and their dependencies use a common output folder, artifacts\Debug\bin\UnitTests. This is achieved by setting `UseCommonOutputDirectory` to "true". This means that individual projects _don't_ copy their dependencies into the output folder in order to avoid multiple projects trying to copy the same file to the same location at the same time. However, in order to run the unit tests from that directory we need to get the dependencies copied in. This is achieved by the DeployTestDependencies project. Each unit test project depends on DeployTestDependencies, which in turn depends on each product assembly. This works, but means that every unit test assembly is effectively dependent on each product assembly, even if they are unrelated. E.g., building Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests.csproj causes us to build Microsoft.VisualStudio.AppDesigner.vbproj. This is a bit strange, and this change just eliminates the use of a common output directory for unit tests and deletes DeployTestDependencies entirely.

Specific changes:

1. In HostAgnostic.props, add a package references to Microsoft.NET.Test.Sdk. This is needed to run tests with "dotnet test". Also update the `IncludeAssets` and `PrivateAssets` to match the recommendations from xUnit, and remove the use of `GeneratePathProperty` as we no longer need it.
2. In UnitTests.targets, call "dotnet test" instead of "xunit.console.x86.exe". I've largely kept the behavior the same as before, but in some cases equivalents were not immediately obvious. For example, I haven't explicitly replicated the "-noshadow" or "-parallel all" switches to xUnit. This may be an area for further investigation.
3. In Build.proj, fix a type that currently prevents us from running tests from VB-based unit test projects.
4. Delete DeployTestDependencies.
5. Update the individual unit test projects to no longer depend on DeployTestDependencies.
6. ~~Disable some tests that inexplicably fail in the new setup; this is tracked by #8642.~~ Update: The underlying issue affecting the tests has been addressed and they no longer need to be disabled.
7. Turn off `UseCommonOutputDirectory` for unit test projects. They will now build to their own output directories.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8643)


adamint:
Will there be any infrastructure changes needed as a result of this?

eng/Build.proj
```diff
@@ -35,7 +35,7 @@
 
     <ItemGroup>
       <BuildProject Include="$(RepoRoot)src/**/*.csproj;$(RepoRoot)src/**/*.vbproj" Condition="'$(SrcProjects)' == 'true'" />
-      <BuildProject Include="$(RepoRoot)tests/**/*.csproj;$(RepoRoot)src/**/*.vbproj" Condition="'$(TestProjects)' == 'true'" />
+      <BuildProject Include="$(RepoRoot)tests/**/*.csproj;$(RepoRoot)tests/**/*.vbproj" Condition="'$(TestProjects)' == 'true'" />
```
MiYanni, line 38: Ohhh. Was this a bug? It is likely something I cause via copy-paste.

tmeschter:
Ah, the Publish Test Results stage will need to be updated to look for .trx files. On it.

eng/imports/UnitTests.targets
```diff
@@ -1,35 +1,35 @@
...
+  <!-- Runs unit tests for a project via 'dotnet test'. -->
   <Target Name="Test">
     <PropertyGroup>
       <TestResultsDirectory>$(ArtifactsTestResultsDir)</TestResultsDirectory>
-      <XUnitExe>$(Pkgxunit_runner_console)\tools\net472\xunit.console.x86.exe</XUnitExe>
+      <TestRunner>dotnet test</TestRunner>
 
       <!-- For Azure DevOps to read as the test results. -->
-      <XmlTestResultsFile>$(TestResultsDirectory)$(TargetName)$(TargetExt).xml</XmlTestResultsFile>
+      <XmlTestResultsFile>$(TestResultsDirectory)$(TargetName)$(TargetExt).trx</XmlTestResultsFile>
```
MiYanni, line 11: nit: It is no longer an `XmlTestResultsFile` but a `TrxTestResultsFile`.

MiYanni (changes requested):
Looks like you need to change the extension in these 2 pipeline templates, or else the test results won't contribute to the build.

https://github.com/dotnet/project-system/blob/fd10ca0af2964fd936e06922b632aa249d977922/eng/pipelines/templates/build-official-release.yml#L190

https://github.com/dotnet/project-system/blob/fd10ca0af2964fd936e06922b632aa249d977922/eng/pipelines/templates/build-pull-request.yml#L42

eng/Build.proj
tmeschter: Yeah. I noticed that one of our unit test projects (Microsoft.VisualStudio.Editors.UnitTests.vbproj) never produced results with build.cmd, and then I found this. :-)

eng/imports/UnitTests.targets
tmeschter: Fixed.

tmeschter:
@MiYanni I updated the .yml files and verified that the PR build was picking up the .trx files.

MiYanni (approved):

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1342,7 +1342,7 @@ public async Task IsUpToDateAsync_True_Sets_OutputOnly()
                 """);
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/dotnet/project-system/issues/8642")]
```
drewnoakes, line 1345: This (and other skips) are not needed with https://github.com/dotnet/project-system/pull/8645.

drewnoakes: That PR has merged, so you can rebase/merge and verify they work again.

drewnoakes:
I'm unclear on whether removing `UseCommonOutputDirectory` actually improves performance here. The FUTDC should skip building those projects in the general case. Using a common output directory avoids a bunch of file copies, which can slow down build.

Could we have the best of both worlds by continuing to use a common output directory, but reducing the set of dependencies to only those needed on a per-test-project basis?

tmeschter:
> I'm unclear on whether removing `UseCommonOutputDirectory` actually improves performance here. The FUTDC should skip building those projects in the general case. Using a common output directory avoids a bunch of file copies, which can slow down build.

The goal was to remove some build complexity that we didn't really need. That probably made build performance *worse* as the complexity was there for a reason, but we just don't have that many projects and so the effect is limited. On my own development system I haven't noticed a difference.

> Could we have the best of both worlds by continuing to use a common output directory, but reducing the set of dependencies to only those needed on a per-test-project basis?

I don't follow. How would that be different from what we had before?

drewnoakes:
>> Could we have the best of both worlds by continuing to use a common output directory, but reducing the set of dependencies to only those needed on a per-test-project basis?
>
> I don't follow. How would that be different from what we had before?

I was referring to this comment in the PR description:

> This works, but means that every unit test assembly is effectively dependent on each product assembly, even if they are unrelated.

I took it as the justification of this change. I wondered if we could add the proper dependencies yet still have the common output directory (best of both worlds).

Unless I missed something 😉

MiYanni:
@tmeschter Howdy. Just checking on if there is anything blocking this PR from being merged. The PR does have some good QoL changes in it.

smitpatel:
This is blocked on code

smitpatel:
Finished as part of merging dev/PortToNET6

PR #8645, opened 2022-10-28
Make order of checked items deterministic

### LINKED ISSUE #8642: Fast Up-to-Date Check tests produce different output with `dotnet test`

drewnoakes:
Fixes #8642

Two facts conspired here to motivate this change:

1. Items were stored in data structures that order based on hash codes.
2. Unit tests validate the verbatim log output from the FUTDC to validate correct behaviour.

This means that in different environments, unit tests would enumerate items in different orders and the log output would change in subtle ways, leading to unit test failures.

This change ensures we enumerate (and therefore log) items in a deterministic order. This ensures our tests don't fail in certain environments. It may also help to compare log output between machines/environments/runs for the same project.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8645)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
@@ -387,18 +387,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                     log.Verbose(nameof(Resources.FUTD_AddingTypedInputs_1), UpToDateCheckInput.SchemaName);
                     log.Indent++;
 
-                    foreach ((string kind, ImmutableArray<string> items) in upToDateCheckInputItems)
+                    foreach (string kind in state.KindNames)
                     {
-                        if (ShouldIgnoreItems(kind, items))
+                        if (upToDateCheckInputItems.TryGetValue(kind, out ImmutableArray<string> items))
                         {
-                            continue;
-                        }
+                            if (ShouldIgnoreItems(kind, items))
+                            {
+                                continue;
+                            }
 
-                        foreach (string path in items)
-                        {
-                            string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
-                            log.VerboseLiteral(absolutePath);
-                            yield return (Path: absolutePath, ItemType: UpToDateCheckInput.SchemaName, IsRequired: true);
+                            foreach (string path in items)
+                            {
+                                string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
+                                log.VerboseLiteral(absolutePath);
+                                yield return (Path: absolutePath, ItemType: UpToDateCheckInput.SchemaName, IsRequired: true);
+                            }
                         }
                     }
 
@@ -413,18 +416,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                     log.Verbose(nameof(Resources.FUTD_AddingTypedOutputs_1), UpToDateCheckOutput.SchemaName);
                     log.Indent++;
 
-                    foreach ((string kind, ImmutableArray<string> items) in upToDateCheckOutputItems)
+                    foreach (string kind in state.KindNames)
                     {
-                        if (ShouldIgnoreItems(kind, items))
+                        if (upToDateCheckOutputItems.TryGetValue(kind, out ImmutableArray<string> items))
                         {
-                            continue;
-                        }
+                            if (ShouldIgnoreItems(kind, items))
+                            {
+                                continue;
+                            }
 
-                        foreach (string path in items)
-                        {
-                            string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
-                            log.VerboseLiteral(absolutePath);
-                            yield return absolutePath;
+                            foreach (string path in items)
+                            {
+                                string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
+                                log.VerboseLiteral(absolutePath);
+                                yield return absolutePath;
+                            }
                         }
                     }
 
@@ -436,18 +442,21 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                     log.Verbose(nameof(Resources.FUTD_AddingTypedOutputs_1), UpToDateCheckBuilt.SchemaName);
                     log.Indent++;
 
-                    foreach ((string kind, ImmutableArray<string> items) in upToDateCheckBuiltItems)
+                    foreach (string kind in state.KindNames)
                     {
-                        if (ShouldIgnoreItems(kind, items))
+                        if (upToDateCheckBuiltItems.TryGetValue(kind, out ImmutableArray<string> items))
                         {
-                            continue;
-                        }
+                            if (ShouldIgnoreItems(kind, items))
+                            {
+                                continue;
+                            }
 
-                        foreach (string path in items)
-                        {
-                            string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
-                            log.VerboseLiteral(absolutePath);
-                            yield return absolutePath;
+                            foreach (string path in items)
+                            {
+                                string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
+                                log.VerboseLiteral(absolutePath);
+                                yield return absolutePath;
+                            }
                         }
                     }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
```diff
@@ -411,7 +445,11 @@ void AddKeys(ImmutableDictionary<string, ImmutableDictionary<string, ImmutableAr
 
                 var itemsByKindBySet = new Dictionary<string, Dictionary<string, HashSet<string>>>(BuildUpToDateCheck.SetNameComparer);
 
-                foreach ((string item, IImmutableDictionary<string, string> metadata) in projectChangeDescription.After.Items)
+                var after = projectChangeDescription.After.Items as IDataWithOriginalSource<KeyValuePair<string, IImmutableDictionary<string, string>>>;
+
+                Assumes.NotNull(after);
```
drewnoakes, line 450: This is the first approach for ensuring stability: The use of `IDataWithOriginalSource<T>` on the snapshot, to preserve order.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
```diff
@@ -220,17 +236,34 @@ private UpToDateCheckImplicitConfiguredInput(ProjectConfiguration projectConfigu
...
+            AddSetNames(upToDateCheckBuiltItemsByKindBySetName);
             setNames.Remove(BuildUpToDateCheck.DefaultSetName);
             SetNames = setNames.OrderBy(n => n, BuildUpToDateCheck.SetNameComparer).ToImmutableArray();
 
-            void AddKeys(ImmutableDictionary<string, ImmutableDictionary<string, ImmutableArray<string>>> dictionary)
+            var kindNames = new HashSet<string>(BuildUpToDateCheck.KindNameComparer);
+            AddKindNames(upToDateCheckInputItemsByKindBySetName);
+            AddKindNames(upToDateCheckOutputItemsByKindBySetName);
+            AddKindNames(upToDateCheckBuiltItemsByKindBySetName);
+            KindNames = kindNames.OrderBy(n => n, BuildUpToDateCheck.KindNameComparer).ToImmutableArray();
```
drewnoakes, line 249: Here we specify the order of kind names as alphabetical (similar to how we treated "set" names previously).

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1419,17 +1419,16 @@ public async Task IsUpToDateAsync_False_Kinds_InputNewerThanOutput_NoKindIgnored
...
-                    C:\Dev\Solution\Project\TaggedBuilt
                     C:\Dev\Solution\Project\Built
+                    C:\Dev\Solution\Project\TaggedBuilt
                 Adding project file inputs:
                     {_projectPath}
                 Adding newest import input:
                     {_projectPath}
                 Adding UpToDateCheckInput inputs:
-                    C:\Dev\Solution\Project\TaggedInput
                     C:\Dev\Solution\Project\Input
```
drewnoakes, line 1432: An example of where the test expectation needed to be updated. The order here now matches that earlier within this test method where they're added to the snapshot.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/UpToDate/BuildUpToDateCheckTests.cs
```diff
@@ -1506,8 +1506,8 @@ public async Task IsUpToDateAsync_True_Kinds_InputNewerThanOutput_NoKindIgnored(
                 lastSuccessfulBuildStartTimeUtc: lastBuildTime,
                 lastItemsChangedAtUtc: itemChangeTime);
 
-            _fileSystem.AddFile(@"C:\Dev\Solution\Project\Input",        inputTime);
-            _fileSystem.AddFile(@"C:\Dev\Solution\Project\TaggedInput",  inputTime);
+            _fileSystem.AddFile(@"C:\Dev\Solution\Project\Input",        input2Time);
+            _fileSystem.AddFile(@"C:\Dev\Solution\Project\TaggedInput",  input1Time);
```
drewnoakes, line 1510: I created different timestamps for these values to break any race condition where values were identical.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -511,18 +526,21 @@ IEnumerable<string> CollectSetOutputs(string setName)
...
-                        {
-                            string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
-                            log.VerboseLiteral(absolutePath);
-                            yield return absolutePath;
+                            foreach (string path in items)
+                            {
+                                string absolutePath = _configuredProject.UnconfiguredProject.MakeRooted(path);
+                                log.VerboseLiteral(absolutePath);
+                                yield return absolutePath;
+                            }
```
MiYanni, line 543: I'd recommend using a method for these blocks of code since it is the same thing 6 times only with a change in input collection and output type. However, the non-tuple output type is just a subset of the tuple (absolutePath), so tuple output would work in all situations. Or you just make a different named method that returns absolutePath only.

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: Ideally the whole loop would be extracted, but as they're generator functions I want to avoid allocating just to do that. There may be a smaller chunk (three lines) that can be extracted at no cost.

MiYanni: Just out of my own interest, what would you be allocating if this was made into the method? Unless this path is hit a hundreds of thousands or millions of times, I'd personally not consider the memory impact to outweigh the cost of maintenance and readability.

*Aside*: Hmm... I'm sure there's likely have an objective measurement of perf vs code simplicity. I wouldn't be surprised if there's tooling already to do those measurements. I'd be interested to see how it compares things, and how different that value would be for the style of design I lean toward.

PR #8438, opened 2022-08-29
Add implicit usings property, and global usings page

adamint:
This PR adds the new global usings property page. It moves the checkbox to enable global usings, and conditions the usings editor on that value being true.

You are allowed to add an implicit using - the "is static" checkbox is disabled if an alias of any length is added.

You are allowed to edit non-readonly implicit usings (currently, usings imported from props/sdk), or delete them. 

Adding validation to this feature is tracked in #8439

image

note: this exposed a bug I introduced in CPS where focus is removed after collectionview refresh. if testing, please clone CPS off this PR: https://devdiv.visualstudio.com/DevDiv/_git/CPS/pullrequest/420346

Closes #7488 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8438)

PR #8651, opened 2022-11-01
Fix implicit usings (and other properties) highlighting when they shouldn't be

adamint:
Fixes #8438 
At some point, I changed IsValueDefinedInContextAsync to always return true.... in some mystery.

Now, the implicit usings control behaves as normal. 

When it's not been set:
image

When it has:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8651)


drewnoakes (commented):
The code change looks good, but there is a bunch of stuff in the commit history of this PR that shouldn't be there. Could you rebase this PR on top of the latest main then force push to make sure there's only one commit here?

drewnoakes:
CI is failing due to a compiler warning. We treat warnings as errors in CI.

> src\Microsoft.VisualStudio.ProjectSystem.Managed\ProjectSystem\Properties\InterceptedProjectProperties\InterceptingPropertyValueProviderBase.cs(3,1): error IDE0005: Using directive is unnecessary.

adamint:
> The code change looks good, but there is a bunch of stuff in the commit history of this PR that shouldn't be there. Could you rebase this PR on top of the latest main then force push to make sure there's only one commit here?

I was planning on squashing

PR #8660, opened 2022-11-08
Fix typos in legacy package property page

drewnoakes:
Fixes a simple typo.

These are `AccessibleName` values, so would be exposed via UIA, for example to screen readers.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8660)


melytc (approved):
Good catch!

PR #8662, opened 2022-11-08
Fix for define constants removing last item

adamint:
Currently, if there is one selected item in C# defineconstants and you try to unselect it, it does not get removed. That is because we do not set the property if there are no user-defined constants. This is fixed by deleting the property.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8662)


haileymck (approved):

drewnoakes:
This PR has a load of unrelated commits in it and should not have been merged as is.

adamint:
@drewnoakes it does not have unrelated commits, the merge commit only contained the single commit.

drewnoakes:
I'm unclear on why GitHub shows 23 commits in this PR.

MiYanni:
@drewnoakes I wish I knew this answer, too. But, I think it involves some investigative work on the difference between the branches. I trust Git to do the right thing, but we generally should make sure that branches are aligned properly for any PR.

PR #8678, opened 2022-11-14
Disable RichNav

MiYanni:
Disabling RichNav due to issue with the task. See: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1659507
Disabling this erroring task will no longer cause the build to 'Partially Succeed' which causes scheduled builds to run multiple times on the same commit.

## Description
I have details about this, but it was all shared on an internal Teams chat. The errors state how they can't find certain DLLs. The issues started on September 24th 2022 and became completely broken on October 13th 2022. The team I spoke with stated that it is only problematic for GitHub repos.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8678)

### DIFF
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -153,7 +153,8 @@ stages:
   jobs:
   - template: templates/publish-assets-and-packages.yml
   - template: templates/publish-symbols.yml
-  - template: templates/publish-richnav.yml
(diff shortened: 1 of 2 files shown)


MiYanni:
Looks like this task no longer puts the build into a partial success state. However, we no longer need it until the linked bug is fixed, as it doesn't succeed anyway (just a waste of build agents and time).

smitpatel (approved):

PR #8681, opened 2022-11-15
Address Properties being null in some cases

adamint:
Add note to docs mentioning that you cannot expect properties to be non-null.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8681)

### DIFF
--- a/docs/repo/property-pages/localization.md
+++ b/docs/repo/property-pages/localization.md
@@ -106,6 +106,9 @@ The `PropertyPageSchema` items also need to be updated accordingly:
 </PropertyPageSchema>
 ```
 
+## Notes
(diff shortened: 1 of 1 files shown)


drewnoakes:
Do we understand why this occurs in different locales? Given we on the engineering team mostly test in English, this sense likely to come up again and again. If we could prevent it, that would be best.

adamint:
No, we don't understand why this occurs, and we will need to do an investigation on this in the future

drewnoakes:
I think we should try and identify why this occurs. Is it that the localised string is not present?

smitpatel:
@drewnoakes - Should we close this? In future when we come across the same error, we can investigate and make a note in doc with details why it occurs.

drewnoakes:
Yes I think this requires some further investigation.

PR #8688, opened 2022-11-17
Add analyser to catch multiple enumeration

drewnoakes:
This is a pattern that comes up in PR review fairly often. It's easy to inadvertently enumerate a lazy sequence more than once in code, which can have undesirable performance characteristics, or produce invalid behaviour.

This change enables an analyzer that catches such issues, and fixes the existing occurrences of it in our code base.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8688)

### DIFF
--- a/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
+++ b/src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
@@ -132,7 +132,7 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
 
                     Dim supportedTargetFrameworksDescriptor As PropertyDescriptor = GetPropertyDescriptor("SupportedTargetFrameworks")
 
-                    Dim supportedFrameworks As IEnumerable(Of TargetFrameworkMoniker) = TargetFrameworkMoniker.GetSupportedTargetFrameworkMonikers(vsFrameworkMultiTargeting, DTEProject, supportedTargetFrameworksDescriptor?.Converter)
+                    Dim supportedFrameworks As IReadOnlyList(Of TargetFrameworkMoniker) = TargetFrameworkMoniker.GetSupportedTargetFrameworkMonikers(vsFrameworkMultiTargeting, DTEProject, supportedTargetFrameworksDescriptor?.Converter)
 
                     'If the list doesn't contain any tfm, it means the project can't retarget.
                     If Not supportedFrameworks.Any() Then
--- a/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Configuration/ProjectConfigurationDimensionProviderTestBase.cs
+++ b/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Configuration/ProjectConfigurationDimensionProviderTestBase.cs
@@ -560,18 +560,18 @@ namespace Microsoft.VisualStudio.ProjectSystem.Configuration
 
             var result = await provider.GetBestGuessDefaultValuesForDimensionsAsync(UnconfiguredProjectFactory.Create());
 
-            Assert.Single(result);
-            Assert.Equal(provider.DimensionName, result.First().Key);
-            Assert.Equal("expected", result.First().Value);
+            (string key, string value) = Assert.Single(result);
+            Assert.Equal(provider.DimensionName, key);
+            Assert.Equal("expected", value);
         }
 
         private static void AssertDefaultOrEmpty(BaseProjectConfigurationDimensionProvider provider, IEnumerable<KeyValuePair<string, string>> result)
         {
             if (provider.DimensionDefaultValue is not null)
             {
-                Assert.Single(result);
-                Assert.Equal(provider.DimensionName, result.First().Key);
-                Assert.Equal(provider.DimensionDefaultValue, result.First().Value);
+                (string key, string value) = Assert.Single(result);
+                Assert.Equal(provider.DimensionName, key);
+                Assert.Equal(provider.DimensionDefaultValue, value);
             }
             else
             {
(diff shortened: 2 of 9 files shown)


drewnoakes:
Thanks to @Cosifine who added this analyzer in https://github.com/dotnet/roslyn-analyzers/pull/5304 ❤️

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Configuration/ProjectConfigurationDimensionProviderTestBase.cs
```diff
@@ -569,9 +569,9 @@ private static void AssertDefaultOrEmpty(BaseProjectConfigurationDimensionProvid
         {
             if (provider.DimensionDefaultValue is not null)
             {
-                Assert.Single(result);
-                Assert.Equal(provider.DimensionName, result.First().Key);
-                Assert.Equal(provider.DimensionDefaultValue, result.First().Value);
+                (string key, string value) = Assert.Single(result);
```
smitpatel, line 572: Do we use var in this case or no?

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Configuration/ProjectConfigurationDimensionProviderTestBase.cs
```diff
@@ -569,9 +569,9 @@ private static void AssertDefaultOrEmpty(BaseProjectConfigurationDimensionProvid
...
+                Assert.Equal(provider.DimensionName, key);
```
smitpatel, line 573: Line 564 has similar thing? At least it can use result of Assert.Single rather than calling First.

smitpatel (commented):
First time review. Feel free to ignore or point me in the right direction if obviously wrong PR comments.

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
```diff
@@ -135,14 +135,18 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
                     Dim supportedFrameworks As IEnumerable(Of TargetFrameworkMoniker) = TargetFrameworkMoniker.GetSupportedTargetFrameworkMonikers(vsFrameworkMultiTargeting, DTEProject, supportedTargetFrameworksDescriptor?.Converter)
```
Youssef1313, line 135: `GetSupportedTargetFrameworkMonikers` returns the result of `GetSupportedTargetFrameworks`

It seems all implementation of `GetSupportedTargetFrameworks` can be changed to return a list (the implementations calls `ToList` internally already). Then, `GetSupportedTargetFrameworkMonikers` can return a list as well.

After that, The code can just say:

```
If supportedFrameworks.Count > 0 Then
    targetFrameworkComboBox.Items.AddRange(supportedFrameworks)
Else
    targetFrameworkSupported = False
    targetFrameworkComboBox.Items.Clear()
End If
```

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
```diff
@@ -135,14 +135,18 @@ Namespace Microsoft.VisualStudio.Editors.PropertyPages
...
+                    Dim hasSupportedFramework = False
+
+                    For Each supportedFramework As TargetFrameworkMoniker In supportedFrameworks
+                        targetFrameworkComboBox.Items.Add(supportedFramework)
+                        hasSupportedFramework = True
+                    Next
+
+                    If Not hasSupportedFramework Then
                         targetFrameworkSupported = False
                         targetFrameworkComboBox.Items.Clear()
```
Youssef1313, line 148: It looks like this can never be useful? `targetFrameworkComboBox.Items.Clear()` was called at the beginning of this function, and no items should have been added in between.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Configuration/ProjectConfigurationDimensionProviderTestBase.cs
MiYanni: `var` is a forbidden keyword in these parts. 🤠😂

smitpatel: In test code too? I noticed we don't use in product code. Also method above uses var.

drewnoakes: We only use var when the type is evident.

Our.editorconfig file is set up to highlight how we use var. You'll see hints in the IDE.

melytc (approved):

tests/Microsoft.VisualStudio.ProjectSystem.Managed.VS.UnitTests/ProjectSystem/VS/Properties/VisualBasic/MapDynamicEnumValuesProviderTests.cs
```diff
@@ -90,10 +90,26 @@ private static void VerifySameValue(IEnumValue? actual, IEnumValue expected, boo
...
+                }
+                else if (actualHasNext != expectedHasNext)
+                {
+                    Assert.Fail("Sequences contain different numbers of entries.");
+                }
+                else
+                {
+                    break;
+                }
             }
```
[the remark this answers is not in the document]
drewnoakes: Updated.

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
drewnoakes: I think you're right, but I'm really reluctant to change this code as we don't have any tests for it and it's used infrequently. Given your other suggestion, I'm happy enough with how this code landed for now.

src/Microsoft.VisualStudio.Editors/PropPages/ApplicationPropPageInternalBase.vb
drewnoakes: Good point. I've changed those methods to return IROL and reverted my changes here.

PR #8689, opened 2022-11-17
Add analyzer to identify methods that could be static

drewnoakes:
This is another relatively common bit of PR review feedback that could be automated by an analyzer.

Marking a member as static makes it clear that it has no dependency on instance state, and is therefore more likely to have other properties such as concurrency safety. It is also slightly faster to call static members at runtime than instance members, primarily because the null check can be elided.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8689)


.editorconfig
```diff
@@ -236,7 +236,7 @@ dotnet_diagnostic.CA2200.severity = none
 dotnet_diagnostic.CA2000.severity = none        # Dispose objects before losing scope (Noisy)
 dotnet_diagnostic.CA1820.severity = warning
 
-# Microsoft.CodeQuality.Analyzers
+# Microsoft.CodeQuality.NetAnalyzers
```
Youssef1313, line 239:
```suggestion
# Microsoft.CodeAnalysis.NetAnalyzers
```

smitpatel (approved):

jacdavis:
Note that I wouldn't take this change. The optimizations are tiny (completely not measurable) and the true meaning of static is not "it has no dependencies on instance state" but "this method applies to the entire class, not the instance. I have many times had to make breaking changes due to someone making a method static because it has not current dependency on this but wasn't designed to be truly static.

drewnoakes:
> I have many times had to make breaking changes due to someone making a method static because it has not current dependency on this but wasn't designed to be truly static and suddenly a bug fix needs to bring in such as dependency.

None of the changes here are to public API. I can see your point in theory, but it's not something I've ever seen in practice.

I'll downgrade the severity from warning (which would fail CI) to suggestion. Unfortunately it's no possible to condition the severity of the analyzer on the visibility of the member or containing type.

Youssef1313:
> Unfortunately it's no possible to condition the severity of the analyzer on the visibility of the member or containing type.

@drewnoakes You can achieve the needed effect like this:

https://github.com/dotnet/aspnetcore/blob/bc9255b509c4b51d9bbbea7b15ddafe482784699/.editorconfig#L112

drewnoakes:
> You can achieve the needed effect like this

That's perfect! Thank you so much @Youssef1313.

PR #8695, opened 2022-11-28
Define Constants Using NameValueList Editor on VB Property Pages

haileymck:
Resolves https://github.com/dotnet/project-system/issues/8387

- Using `Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.Designer.INameValuePairListEncoding` creates custom encoding `NameQuotedValueListPairEncoding` for the format  key="value"
- UI changes will appear as shown below 
Screenshot 2022-11-28 102357
- includes changes to eng/imports/Packages.targets to get the required changes to CPS
- renames the preexisting DefineConstants Value Provider to specify that it is for C# 

Note: I created a new PR for this because the changes made on CPS side, modified code changes a lot

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8695)


adamint:
have you verified that the defineconstants property works for VB still?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+[ExportInterceptingPropertyValueProvider(DefineConstantsPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+[AppliesTo(ProjectCapability.VisualBasic)]
+internal sealed class DefineConstantsVBValueProvider : InterceptingPropertyValueProviderBase
+{
+    internal const string DefineConstantsPropertyName = "DefineConstants";
```
adamint, line 11: this can be taken as a constant from GeneralBrowseObject

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    private readonly NameQuotedValuePairListEncoding _encoding = new();
```
adamint, line 13: nit> instantiate from constructor

adamint (approved):
couple nits, but they aren't required!

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
drewnoakes: Should probably take the property from `ConfigurationGeneral` which is intended for programmatic access, while the browse object is for the UI.

```suggestion
    internal const string DefineConstantsPropertyName = ConfigurationGeneral.DefineConstantsProperty;
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
drewnoakes: I prefer initializing here, especially when there is no constructor. I don't know that we have a standard throughout the code though.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Properties/InterceptedProjectProperties/BuildPropertyPage/DefineConstantsVBValueProvider.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties;
+
+[ExportInterceptingPropertyValueProvider(DefineConstantsPropertyName, ExportInterceptingPropertyValueProviderFile.ProjectFile)]
+[AppliesTo(ProjectCapability.VisualBasic)]
+internal sealed class DefineConstantsVBValueProvider : InterceptingPropertyValueProviderBase
```
drewnoakes, line 9: I can't see why we need to intercept this property at all. The UI editor should now be using the same encoding as the underlying project file, so I don't think we need to touch the value in any way as it travels by.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -416,7 +416,14 @@
...
                   Category="Advanced">
-    <StringProperty.Description>Enter any custom constants for your application in this text box. Entries should be delimited by commas, using this form: Name1="Value1",Name2="Value2",Name3="Value3".</StringProperty.Description>
+    <StringProperty.ValueEditors>
+      <ValueEditor EditorType="NameValueList">
+        <ValueEditor.Metadata>
+          <NameValuePair Name="Encoding" Value="Default" />
+        </ValueEditor.Metadata>
+      </ValueEditor>
+    </StringProperty.ValueEditors>
+    <StringProperty.Description>Specifies custom conditional compilation constants for your application.</StringProperty.Description>
```
drewnoakes, line 426: Now that the description no longer contains `"` characters, it can be moved to an attribute on the parent `<StringProperty Description="..."` instead.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PropertyPages/BuildPropertyPage.VisualBasic.xaml
```diff
@@ -416,7 +416,14 @@
                   DisplayName="Custom constants"
                   HelpUrl="https://go.microsoft.com/fwlink/?linkid=2196264"
                   Category="Advanced">
-    <StringProperty.Description>Enter any custom constants for your application in this text box. Entries should be delimited by commas, using this form: Name1="Value1",Name2="Value2",Name3="Value3".</StringProperty.Description>
+    <StringProperty.ValueEditors>
+      <ValueEditor EditorType="NameValueList">
+        <ValueEditor.Metadata>
+          <NameValuePair Name="Encoding" Value="Default" />
```
drewnoakes, line 422: I would expect your new encoding name to be used here, rather than "Default".

```suggestion
          <NameValuePair Name="Encoding" Value="NameQuotedValuePairListEncoding" />
```

With that specified, I don't think we need interception.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,128 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.ProjectSystem.VS.PropertyPages.Designer;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Debug
+{
+
+    [Export(typeof(INameValuePairListEncoding))]
+    [ExportMetadata("Encoding", "NameQuotedValuePairListEncoding")]
+    internal sealed class NameQuotedValuePairListEncoding : INameValuePairListEncoding
```
drewnoakes, line 10: This encoding still tries to handle two different formats (A & B), when it only needs to handle one (A).

A: `a="1",b="2"`
A: `a=1,b=2`

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/VBDefineConstantsProviderTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        [InlineData("path=/path/to/somewhere/", "path=\"//path//to//somewhere//\"")]
+        [InlineData("file=/path/is/here/", "file=\"//path//is//here//\"")]
+        [InlineData("files=path=/path/to/somewhere/", "files=\"path/=//path//to//somewhere//\"")]
+        [InlineData("a=1,a=2", "a=\"1\",a=\"2\"")]
+        [InlineData("key1  =value1  ", "key1  =\"value1  \"")]
+        [InlineData("  key1=  value1", "  key1=\"  value1\"")]
+        [InlineData(" key1 = value1 ", " key1 =\" value1 \"")]
+        [InlineData(" = ", " =\" \"")]
+        [InlineData("key1=a b c d", "key1=\"a b c d\"")]
+        public void DefineConstantsEncodingFormatTest(string input, string expectedOutput)
```
drewnoakes, line 25: These test cases handle both quoted and unquoted values. This encoding should only accept quoted values.

One way to write this would be:

```c#
        [InlineData("key1=\"a\",key2=\"b\"", new[] { "key1", "a", "key2", "b" })]
        public void DefineConstantsEncodingFormatTest(string input, string[] pairs)
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/VBDefineConstantsProviderTests.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
+
+using Microsoft.VisualStudio.ProjectSystem.Debug;
+
+namespace Microsoft.VisualStudio.ProjectSystem.Properties.InterceptingProjectProperties.BuildPropertyPage
+{
+    public class VBDefineConstantsProviderTests
```
drewnoakes, line 7:
```suggestion
    public class NameQuotedValuePairListEncodingTests
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/VBDefineConstantsProviderTests.cs
drewnoakes: Consider also adding something like:

```c#
[Theory]
[InlineData("InvalidString")]
// ...other invalid strings here
public void ParseThrows(string input)
{
    FormatException exception = Assert.Throws<FormatException>(() => NameQuotedValuePairListEncoding.Parse(input));

    Assert.Equals("Invalid input string.", exception.Message); // Consider better errors, or even a different error per input
}
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,117 @@
...
+                            i--;
+                        }
+                    }
+                    i--;
+                }
+                return value;
+            }
+        }
+
+        private static string StripQuotes(string value)
```
drewnoakes, line 61: nit: `StringQuotes`, `ReadEntries` and `SplitEntry` methods are only called in one place, so could be converted to a local function of that caller, alongside the other local function the caller already has.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            static string DecodeCharacters(string value)
+            {
+                return value.Replace("/=", "=").Replace("/,", ",").Replace("//", "/").Replace("/\"", "\"");
+            }
+        }
+
+        public string Format(IEnumerable<(string Name, string Value)> pairs)
+        {
+            return string.Join(",", pairs.Select(pair => $"{EncodeCharacters(pair.Name)}=\"{EncodeCharacters(pair.Value)}\""));
```
drewnoakes, line 37: If you're interested, this can be made to allocate fewer temporary objects by using a `StringBuilder` and appending to it as you walk through pairs. The `EncodeCharacters` method could also append to that.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/NameQuotedValuePairListEncodingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        [InlineData("  key1=  value1", "  key1=\"  value1\"")]
+        [InlineData(" key1 = value1 ", " key1 =\" value1 \"")]
+        [InlineData(" = ", " =\" \"")]
+        [InlineData("key1=a b c d", "key1=\"a b c d\"")]
+        [InlineData("key1=a=b=c=d", "key1=\"a/=b/=c/=d\"")]
+        [InlineData("key1=a=b\"\"c=d", "key1=\"a/=b/\"/\"c/=d\"")]
+        [InlineData("key1=\"ab\"\"cd", "key1=\"/\"ab/\"/\"cd\"")]
+        [InlineData("key1=\"\"ab\"\"cd\"\"ef", "key1=\"/\"/\"ab/\"/\"cd/\"/\"ef\"")]
+        [InlineData("key1=\"\"ab\"\"cd", "key1=\"/\"/\"ab/\"/\"cd\"")]
+        public void ValidNameQuotedValuePairListEncoding(string input, string expectedOutput)
```
drewnoakes, line 30: I don't see any test cases that cover scaped characters in the key.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/NameQuotedValuePairListEncodingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            NameQuotedValuePairListEncoding _encoding = new();
+            Assert.Equal(expected: expectedOutput, actual: _encoding.Format(_encoding.Parse(input)));
```
drewnoakes, line 33: I feel like this test is too coarse and could miss bugs. It's testing both the parsing and reformatting at once, without really inspecting the intermediate value. It's the intermediate value that the UI will use, so it's quite important that it's correct.

By testing two things at once here it's possible that a bug in the first is undone in the second, and therefore goes unnoticed.

Something like this (written in a browser, so may not compile, but should get the idea across):

```c#
[Theory]
[InlineData("a=\"1\"", new[] { "a", "1" })]
[InlineData("a=\"1\",b=\"2\"", new[] { "a", "1", "b", "2" })]
public void ValidNameQuotedValuePairListEncoding(string encoded, string[] pairs)
{
    NameQuotedValuePairListEncoding _encoding = new();

    // Try parsing
    var actualPairs = _encoding.Parse(encoded);

    Assert.Equal(pairs, actualPairs.SelectMany(pair => new[] { pair.Name, pair.Value })); 

    // Try formatting
    var actualEncoded = _encoding.Format(ToNameValues(pairs));

    Assert.Equal(encoded, actualEncoded);

    static IEnumerable<(string Name, string Value)> ToNameValues(IEnumerable<string> pairs)
    {
        using var e = pairs.GetEnumerator();
        while (e.MoveNext())
        {
            var name = e.Current;
            Assert.True(e.MoveNext());
            var value = e.Current;
            yield return (name, value);
        }
    }
}
```

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/NameQuotedValuePairListEncodingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        [InlineData(",")]
+        [InlineData(",,,")]
+        [InlineData("")]
+        [InlineData("\"")]
+        [InlineData("key1,=abcd")]
+        [InlineData("key1,=,abcd")]
+        public void InvalidNameQuotedValuePairListEncoding(string input)
+        {
+            NameQuotedValuePairListEncoding _encoding = new();
+            Assert.Empty(_encoding.Format(_encoding.Parse(input)));
```
drewnoakes, line 56: Can we make this throw? If we return empty here, we're basically throwing away whatever data the customer had in their project, which is worse than bringing the error to the customer's attention.

adamint (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,124 @@
...
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                yield break;
+            }
+
+            foreach (var entry in ReadEntries(value))
+            {
+                var (entryKey, entryValue) = SplitEntry(entry);
+
+                if (!string.IsNullOrEmpty(entryKey))
```
smitpatel, line 23: return values of `SplitEntry` doesn't return null or empty. This check is not necessary.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,124 @@
...
+                return value.Replace("/\"", "\"").Replace("/=", "=").Replace("/,", ",").Replace("//", "/");
+            }
+
+            static string StripQuotes(string value)
+            {
+                if (value.StartsWith("\"") && value.EndsWith("\"") && value.Length >= 2)
+                {
+                    return value.Substring(1, value.Length - 2);
+                }
+                return value;
```
smitpatel, line 93: Should this encoding allow formatting of value in quote/non-quote both ways or should enforce that it is always quoted value?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        public string Format(IEnumerable<(string Name, string Value)> pairs)
+        {
+            return string.Join(",", pairs.Select(pair => $"{EncodeCharacters(pair.Name)}=\"{EncodeCharacters(pair.Value)}\""));
+
+            static string EncodeCharacters(string value)
+            {
+                int i = value.Length - 1;
+                while( i > -1)
```
smitpatel, line 104:
```suggestion
                while (i > -1)
```

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,124 @@
...
+    [Export(typeof(INameValuePairListEncoding))]
+    [ExportMetadata("Encoding", "NameQuotedValuePairListEncoding")]
+    internal sealed class NameQuotedValuePairListEncoding : INameValuePairListEncoding
+    {
+        public IEnumerable<(string Name, string Value)> Parse(string value)
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                yield break;
+            }
```
drewnoakes, line 17: Consider:

```suggestion
            Requires.NotNullOrWhiteSpace(value, nameof(value));
```

The rationale here is that a `null` value would be an error (the argument is non-nullable), and throwing when we have invalid values helps to discover issues, rather than failing silently. If we did fail silently here, we might return an empty sequence which would then potentially also overwrite customer data (assuming the issue happened between getting the value from the project and passing it to this class).

Equally, an empty or white space string cannot be parsed correctly, so should throw.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            static string DecodeCharacters(string value)
+            {
+                
+                return value.Replace("/\"", "\"").Replace("/=", "=").Replace("/,", ",").Replace("//", "/");
+            }
+
+            static string StripQuotes(string value)
+            {
+                if (value.StartsWith("\"") && value.EndsWith("\"") && value.Length >= 2)
```
drewnoakes, line 89: I'm curious what would happen for input `abc= "123" ` (i.e. space before/after the quotes). My guess is that the current code would give the value as ` "123" `. However this is an invalid input so should probably throw.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -0,0 +1,124 @@
...
+                        escaped = false;
+                    }
+                }
+                throw new FormatException("Expected valid name value pair.");
+            }
+
+            static string DecodeCharacters(string value)
+            {
+                
+                return value.Replace("/\"", "\"").Replace("/=", "=").Replace("/,", ",").Replace("//", "/");
```
drewnoakes, line 84: What happens if a character other than `"=,/` is escaped?

For example, given input `a/bc`, should it:

1. Return `abc` 
2. Return `a/bc`
3. Throw

Whatever you decide, I'd add a test case to cover it.

drewnoakes (approved):
Looks good! Let a few comments.

I'd also like to understand what happens when the project file contains an incorrectly formatted value and the Project Properties UI is opened. What happens to that FormatException? Is it easy for the customer to understand the problem and fix it themselves?

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
drewnoakes: My vote would be to require quoted values.

melytc (approved):

PR #8711, opened 2022-12-07
Replace CancellationTokenSource.CreateLinkedTokenSource with CancellationTokenExtensions.CombineWith

drewnoakes:
Replaces `CancellationTokenSource.CreateLinkedTokenSource` with `CancellationTokenExtensions.CombineWith`.

The latter will not allocate when either token is `CancellationToken.None`.

We also:

- Remove an unused `CancellationTokenExtensions` class that had a conflicting name, and
- Ban `CancellationTokenSource.CreateLinkedTokenSource` to prevent future use.

See https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24027/Threading?anchor=%E2%9C%94%EF%B8%8F-**do**-use-%60cancellationtokenextensions.combinewith%60-over-%60cancellationtokensource.createlinkedtokensource%60-to-combine-cancellation-tokens for further details.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8711)


src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
```diff
@@ -66,7 +66,9 @@ public CancellationSeries(CancellationToken token = default)
         /// <exception cref="ObjectDisposedException">This object has been disposed.</exception>
         public CancellationToken CreateNext(CancellationToken token = default)
         {
+#pragma warning disable RS0030 // Do not used banned APIs (cannot use alternative API here)
```
melytc, line 69: Just out of curiosity, why is it not possible to use `CancellationTokenExtensions.CombineWith` here?

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: It returns a struct and we need a reference type for the interlocked operation.

MiYanni: Could you declare it as a nullable of that struct type, since nullable is a reference type? (Keep in mind, I don't fully understand the context here, so my question might not make any sense.)

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/Threading/Tasks/CancellationSeries.cs
drewnoakes: Nullable structs are still value types, so that wouldn't help here unfortunately. There's probably a way to change the design here but it'd increase the complexity a fair bit I suspect, or introduce blocking synchronization.

MiYanni: Doy. Not sure why I thought `Nullable<T>` was a reference type. My bad.

PR #8715, opened 2022-12-08
Ensure dataflow faults don't hang the language service during initialisation

drewnoakes:
[AB#1644056](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1644056)

In 17.4 we saw an exception in the dataflow chain that would result in a hang. The cause of that exception has been addressed, however it highlighted a potential hang in the workspace if such an exception occurred during workspace initialisation.

This change hooks dataflow faults and passes those exceptions on to any code waiting for initialisation to complete, preventing a hang and allowing those callers to observe the fault.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8715)


MiYanni (approved):
nit: Only thing I notice is the comments switch back-and-forth between initialize being spelled with a 'z' and initialise being spelled with an 's'. 🤓

PR #8720, opened 2022-12-14
FUTDC learns some new metadata on ProjectReference items

### LINKED ISSUE #8716: FUTDC should handle EmbedInteropTypes on project references correctly
After opting CPS into Build Acceleration, we see an additional DLL in one of the output directories:

```
bin/Microsoft.VisualStudio.ProjectSystem.VS.Implementation/Debug/net472/Microsoft.VisualStudio.ProjectSystem.Interop.dll
bin/Microsoft.VisualStudio.ProjectSystem.VS.Implementation/Debug/net472/Microsoft.VisualStudio.ProjectSystem.Interop.xml
```

From the verbose build log:

```
37>FastUpToDate:     Checking copy items from project 'D:\repos\CPS\src\Microsoft.VisualStudio.ProjectSystem.Interop\Microsoft.VisualStudio.ProjectSystem.Interop.csproj'. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:         Checking PreserveNewest item (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:             Source      2022-12-12 15:53:53.593: 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.Interop\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.dll' (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:             Destination 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.dll' does not exist. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:         Remembering the need to copy file 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.Interop\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.dll' to 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.dll'. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:         Checking PreserveNewest item (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:             Source      2022-12-12 15:53:52.186: 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.Interop\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.xml' (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:             Destination 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.xml' does not exist. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:         Remembering the need to copy file 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.Interop\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.xml' to 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.xml'. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
...
37>FastUpToDate: Copying 2 files to accelerate build: (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:     From 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.Interop\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.dll' to 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.dll'. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
37>FastUpToDate:     From 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.Interop\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.xml' to 'D:\repos\CPS\bin\Microsoft.VisualStudio.ProjectSystem.VS.Implementation\Debug\net472\Microsoft.VisualStudio.ProjectSystem.Interop.xml'. (Microsoft.VisualStudio.ProjectSystem.VS.Implementation)
```

Deleting the output directory and performing a new build will not produce these files. We should understand why these files are being identified as needed and fix this.

I compared outputs of all CPS projects and this was the only difference.

### LINKED ISSUE #8719: FUTDC should correctly model CopyToOutputDirectory items across P2P references when Private="false"
## Visual Studio Version

17.5p2

## Summary

The FUTDC models the graph of project dependencies and the items each contributes to the output directory. It needs this in order to track `CopyToOutputDirectory` items all transitively referenced projects.

`ProjectReference` items may have `Private` metadata that controls whether those references should be copied to the output directory or not.

The FUTDC should respect this metadata to prevent two issues:

1. Without handling this correctly, we expect a file to exist in the output directory, yet it will never appear. This leads to overbuild.
1. For projects that opt in to Build Acceleration, we would copy more files than expected. This can place files in the output directory that would not otherwise be there.

drewnoakes:
Fixes #8719
Fixes #8716

`ProjectReference` items can be modified with metadata that modifies how items the project produces are copied to the referencing project's output directory.

This PR improves handling of such metadata, to avoid two problems:

1. Incorrectly expecting certain items in the output directory and scheduling builds (overbuild)
2. Copying too many files to the output directory when using Build Acceleration

Three kinds of metadata are covered here:

1. `Private="False"`
2. `EmbedInteropTypes="True"`
3. `ReferenceOutputAssembly="False"`

Here we filter such projects from the graph.

In reality, the situation is more subtle, as mentioned in the code comment. We will make some additional changes here to improve fidelity in the near future.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8720)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
@@ -669,14 +669,25 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
 
                 static bool IncludeProjectReference(IImmutableDictionary<string, string> metadata)
                 {
-                    // Exclude any project references for which we do not reference the output assembly.
+                    // TODO this filtering is overzealous. In each of these cases, there are subtleties to how
+                    // builds handle the output assembly vs. CopyToOutputDirectory items both of the directly
+                    // referenced project, and of transitively referenced projects. To improve this we need
+                    // more information on the edges of our project reference graph.
+
+                    // Exclude any project reference for which we do not reference the output assembly.
                     if (metadata.GetBoolProperty(ResolvedProjectReference.ReferenceOutputAssemblyProperty) == false)
                     {
                         return false;
                     }
 
-                    // Exclude any project references having Private="false" (aka CopyLocal="No").
+                    // Exclude any project reference having Private="false" (aka CopyLocal="No").
                     if (metadata.GetBoolProperty(ResolvedProjectReference.PrivateProperty) == false)
+                    {
+                        return false;
+                    }
+
+                    // Exclude any project reference having EmbedInteropTypes="true".
+                    if (metadata.GetBoolProperty(ResolvedProjectReference.EmbedInteropTypesProperty) == true)
                     {
                         return false;
                     }
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
@@ -659,15 +659,30 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                             .Select(pair => new CopyItem(path: pair.Key, metadata: pair.Value))
                             .ToImmutableArray();
 
-                        // Exclude any project references for which we do not reference the output assembly.
-                        // This avoids following references to certain "private" project references, such as PrivateP2PCaching.proj in NBGV.
-                        ImmutableArray<string> referenceItems = change2.After.Items.Where(pair => pair.Value.GetBoolProperty(ResolvedProjectReference.ReferenceOutputAssemblyProperty) == true).Select(item => item.Key).ToImmutableArray();
+                        ImmutableArray<string> referenceItems = change2.After.Items.Where(pair => IncludeProjectReference(pair.Value)).Select(item => item.Key).ToImmutableArray();
 
                         return new ProjectCopyData(msBuildProjectFullPath, targetPath, copyItems, referenceItems);
                     }
                 }
 
                 return ProjectCopyData;
+
+                static bool IncludeProjectReference(IImmutableDictionary<string, string> metadata)
+                {
+                    // Exclude any project references for which we do not reference the output assembly.
+                    if (metadata.GetBoolProperty(ResolvedProjectReference.ReferenceOutputAssemblyProperty) == false)
+                    {
+                        return false;
+                    }
+
+                    // Exclude any project references having Private="false" (aka CopyLocal="No").
+                    if (metadata.GetBoolProperty(ResolvedProjectReference.PrivateProperty) == false)
+                    {
+                        return false;
+                    }
+
+                    return true;
+                }
             }
 
             ProjectFileClassifier BuildClassifier()


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
```diff
@@ -659,15 +659,30 @@ ProjectCopyData UpdateCopyData()
...
+                        return false;
+                    }
+
+                    // Exclude any project references having Private="false" (aka CopyLocal="No").
+                    if (metadata.GetBoolProperty(ResolvedProjectReference.PrivateProperty) == false)
+                    {
+                        return false;
+                    }
+
+                    return true;
```
MiYanni, line 684: Was it an intentional change to invert the logic for `ReferenceOutputAssemblyProperty`? Meaning before, only `true` items were included but now `null` and `true` are included.

Additionally, subsequent boolean logic gets harder for me to understand when it is using statement flow to denote truthy/falsey-ness over booleans actually interacting with each other directly. Meaning, these separate statements equate directly to this one expression:
```suggestion
                    // Exclude any project references for which we do not reference the output assembly, and
                    // exclude any project references having Private="false" (aka CopyLocal="No").
                    return !(metadata.GetBoolProperty(ResolvedProjectReference.ReferenceOutputAssemblyProperty) is false ||
                             metadata.GetBoolProperty(ResolvedProjectReference.PrivateProperty) is false);
```

drewnoakes: > Was it an intentional change

Yes, thanks. Good spot. If the value is unspecified (which I haven't seen in practice) then we assume the assembly is referenced.

> gets harder for me to understand

This is one of those style things where we probably won't see eye to eye. I used to compress code a lot, but now find it more readable to spread things out a bit. Especially when I'm reading other people's code. In your suggested code, it takes me a while to parse `!(A || B)`, especially as it is split across multiple lines and the `||` is off screen.

I also like putting the comments directly with the clauses they apply to.

Since your review I've added a third clause.

As mentioned in the newly added comment, this logic will require some further tweaking. This method may go away anyway.

Right now I just want to unblock dogfooders who are hitting these errors.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
```diff
@@ -659,15 +659,41 @@ ProjectCopyData UpdateCopyData()
...
 
                         return new ProjectCopyData(msBuildProjectFullPath, targetPath, copyItems, referenceItems);
                     }
                 }
 
                 return ProjectCopyData;
+
+                static bool IncludeProjectReference(IImmutableDictionary<string, string> metadata)
+                {
+                    // TODO this filtering is overzealous. In each of these cases, there are subtleties to how
```
melytc, line 672: Today I learned a new word: "overzealous" 🙂

melytc (approved):
LGTM!

tmeschter (approved):
Looks good to me. Three thoughts for possible further review or improvement:

1. You mention that the `IncludeProjectReference` check is overzealous. Will this lead to us building where we could have just copied instead, or will it lead to us copying where we should have built?
2. Adding specific comments about why we want to exclude projects with ReferenceOutputAssembly="false", Private="false", or EmbedInteropTypes="true" would be helpful for long-term maintenance.
3. Unit tests to document the behavior would be nice, though I appreciate that we're trying to get this in to improve dogfooding.

drewnoakes:
> Looks good to me. Three thoughts for possible further review or improvement:
> 
> 1. You mention that the `IncludeProjectReference` check is overzealous. Will this lead to us building where we could have just copied instead, or will it lead to us copying where we should have built?
> 2. Adding specific comments about why we want to exclude projects with ReferenceOutputAssembly="false", Private="false", or EmbedInteropTypes="true" would be helpful for long-term maintenance.
> 3. Unit tests to document the behavior would be nice, though I appreciate that we're trying to get this in to improve dogfooding.

1. It leads to two potential outcomes:
   - Not triggering a build when an item that would be copied is changed. Thankfully this is no worse than what we have in 17.4 (or since the FUTDC began), and doesn't appear to be a significant customer pain point.
   - Not copying an item that should be copied during Build Acceleration. This is likely the more significant, but still only impacts people wishing to try the new feature out, and only if they're using these relatively uncommon modifiers on their project references. Even then, the code as it stands in this PR is pretty close to what we need.
2. Agreed. I do intend to add more documentation here as part of the follow up work.
3. I have a test solution that demonstrates expected behaviour across all of these scenarios and more.  The rules have more complexity than I show here, and having unit tests to validate that behaviour will be important. As an example, `ReferenceOutputAssembly` excludes the output assembly but not other copy items. Further, it excludes the output assembly of transitively referenced projects, but again, not of their copy items. The other two each have their own rules.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/UpToDateCheckImplicitConfiguredInput.cs
MiYanni: Yeah. It is one of those preference things. For me, it is easier to understand things as expressions because I internally convert things to expressions. So, if they are already written that way, it is less work for me to understand. 😁

I've been trying to figure this out for a long time, but I think I do this process:
1. Convert code into shapes
2. Identify if there are any patterns in the shapes
3. If the patterns aren't something I immediately understand, I have to start reading line-by-line and construct a simplified metal expression of the code
4. Lastly, I mentally test input/output from the expression to see if it satisfies the code's intent

For me, the shorter (meaning, less tall) the shapes are, the easier it is for me to see the patterns. I think it is a matter of scale, like how you might see the pattern on the face of building's architectural design, but then it may be harder to realize that same pattern exists throughout many buildings in an entire city. And don't ask me why I have to do a 4-stage process to understand code, but it just kinda developed that way. 🙃

PR #8722, opened 2022-12-14
Move language service support

tmeschter:
In https://github.com/dotnet/roslyn/pull/65592 the Roslyn team defined a new brokered service for communicating information from the project system to the language service. In effect, the `IWorkspaceProjectContext` and `IWorkspaceProjectContextFactory` types are being replaced by the very similar `IWorkspaceProject` and `IWorkspaceProjectFactoryService` types, but it isn't happening all at once. In preparation, this commit moves all the code depending on `IWorkspaceProjectContext` and `IWorkspaceProjectContextFactory` out of our host-agnostic layer (Microsoft.VisualStudio.ProjectSystem.Managed) and into our VS-specific layer (Microsoft.VisualStudio.ProjectSystem.Managed.VS). Specific changes:

1. Any files that actually call the old interfaces are moved as-is to the VS layer.
2. Any further files that depend on those files in 1 are moved as well.
3. Files that refer to the old interfaces only in comments or doc comments have been updated to use phrases like "language service project" instead but have otherwise been left in place.
4. The dependency on the Microsoft.VisualStudio.LanguageServices package has moved from the host-agnostic layer to the VS-specific layer.
5. The `DotNetImportsEnumProvider` has also moved to the VS layer, as it depends on the `VisualStudioWorkspace` type. The host-agnostic layer no longer references the assembly with that type.
6. Unit test files have also been moved around to reflect the new layering.

There are a few further dependencies on the Roslyn `Workspace` type in Microsoft.VisualStudio.ProjectSystem.Managed; those will be moved in a future change as direct access to a `Workspace` is very much dependent on VS.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8722)


MiYanni (approved):

tmeschter:
Oops, meant to request @drewnoakes review of this, not assign it to him. Apologies.

drewnoakes (approved):

PR #8725, opened 2022-12-15
Move the VsTelemetryService

tmeschter:
We report all telemetry events through our `ITelemetryService` interface, which is implemented by `VsTelemetryService`. This type is defined in Microsoft.VisualStudio.ProjectSystem.Managed.VS but, despite the name, it is not strongly tied to VS. For example, the `TelemetrySession` type it uses is also available in processes utilizing ServiceHub (depending on how the process is setup).

This commit moves `VsTelemetryService` and the associated unit tests down to our host-agnostic level, and renames it to `ManagedTelemetryService`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8725)


drewnoakes (approved):
If it's not VS-specific, perhaps we rename our class accordingly.

MiYanni (approved):
+1 on Drew's rename suggestion.

tmeschter:
> If it's not VS-specific, perhaps we rename our class accordingly.

Oh. Right. I knew I was planning a second commit, but then I forgot what I was going to do. It was the rename! 🤣

PR #8726, opened 2022-12-16
Removing the FormatException thrown for the empty case in the NameValueListPairEncoding

haileymck:
Fixes [AB#1706419](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1706419)

Removing the Format Exception thrown in the NameValueListPairEncoding as the user cannot enter an input in which this error will be thrown as the intermediate steps feed in such a format where it will not be encountered by a user.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8726)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
@@ -11,12 +11,13 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
     {
         public IEnumerable<(string Name, string Value)> Parse(string value)
         {
-            Requires.NotNullOrWhiteSpace(value, nameof(value));
-
             foreach (var entry in ReadEntries(value))
             {
                 var (entryKey, entryValue) = SplitEntry(entry);
-                yield return (DecodeCharacters(entryKey), StripQuotes(DecodeCharacters(entryValue)));
+                if (!string.IsNullOrEmpty(entryKey))
+                {
+                    yield return (DecodeCharacters(entryKey), StripQuotes(DecodeCharacters(entryValue)));
+                }
             }
 
             static IEnumerable<string> ReadEntries(string rawText)
@@ -46,6 +47,10 @@ namespace Microsoft.VisualStudio.ProjectSystem.Debug
 
             static (string EncodedKey, string EncodedValue) SplitEntry(string entry)
             {
+                if (string.IsNullOrEmpty(entry))
+                {
+                    return (string.Empty, string.Empty);
+                }
                 bool escaped = false;
                 for (int i = 0; i < entry.Length; i++)
                 {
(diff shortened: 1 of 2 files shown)


melytc (approved):

drewnoakes (commented):
This will lead to users losing their data. Do we understand why we are seeing this exception in the first place?

haileymck:
> This will lead to users losing their data. Do we understand why we are seeing this exception in the first place?

Fixed it! It should still throw the exception now, just accounts for the empty case

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/NameQuotedValuePairListEncodingTests.cs
```diff
@@ -90,9 +93,8 @@ static IEnumerable<(string Name, string Value)> ToNameValues(IEnumerable<string>
                 }
             }
         }
-
+        
         [Theory]
-        [InlineData("=key1=")]
```
drewnoakes, line 95: Why remove this case?> It still looks invalid to me.

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Debug/NameQuotedValuePairListEncoding.cs
```diff
@@ -11,12 +11,13 @@ internal sealed class NameQuotedValuePairListEncoding : INameValuePairListEncodi
     {
         public IEnumerable<(string Name, string Value)> Parse(string value)
         {
-            Requires.NotNullOrWhiteSpace(value, nameof(value));
-
             foreach (var entry in ReadEntries(value))
             {
                 var (entryKey, entryValue) = SplitEntry(entry);
-                yield return (DecodeCharacters(entryKey), StripQuotes(DecodeCharacters(entryValue)));
+                if (!string.IsNullOrEmpty(entryKey))
```
drewnoakes, line 17: This will throw away a value having an empty key, which is data loss. Eg. `="value"`.

tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/ProjectSystem/Properties/InterceptingProjectProperties/BuildPropertyPage/NameQuotedValuePairListEncodingTests.cs
```diff
@@ -36,6 +36,9 @@ public class NameQuotedValuePairListEncodingTests
         [InlineData("\"\"ab\"\"cd=     value1     ", "/\"/\"ab/\"/\"cd=\"     value1     \"")]
         [InlineData("\"\"ab\"\"cd=value1     ", "/\"/\"ab/\"/\"cd=\"value1     \"")]
         [InlineData("\"\"ab\"\"cd=     value1", "/\"/\"ab/\"/\"cd=\"     value1\"")]
+        [InlineData("", "")]
+        [InlineData(",", "")]
+        [InlineData(",,,", "")]
```
drewnoakes, line 41: The first of these three cases makes sense to me, but the second two are less clear. They are still potentially invalid strings that should raise format exceptions.

I think the implementation changes could be reduced in this PR to just checking for an empty string in `Parse` and returning and empty enumeration in such cases.

drewnoakes (approved):
Nice!

PR #8731, opened 2022-12-19
Expand rule tests

drewnoakes:
There are two parts to this PR.

## Run rule tests against PropertyPages folder too (3e57d83f26c8a3d0219f726060d1d7f9c993087d)

We have a suite of unit tests that validate certain properties of our XAML rules. These rules were not being executed against files in the PropertyPages folder.

This commit extends the scope of existing rules, and makes a few fixes/suppressions accordingly.

## Ensure properties of embedded rules are not visible (83b21a4adbfbc73e4fbe4099df8561bceed7f707) 

We are not currently able to localize embedded rules. Such rules must not have visible properties, as all visible values must be localized.

This commit adds a unit test to ensure this condition is met, and fixes three violations.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8731)


melytc (approved):
Nice!

MiYanni (approved):

Issue #8733, opened 2022-12-22
Reduce number of file system checks


drewnoakes:
PR #8696 fixed #4665 by gathering copy items across project references and including them in the FUTDC.

In addition to items indicating `CopyToOutputDirectory`, Build Acceleration tracks DLL/PDB/XML files from all transitively referenced projects, for each project. This is necessary in order to know when to copy files.

Checking these extra files has flagged a regression in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1706479 on the number of files being checked.

When not using build acceleration, we shouldn't need to track so many files. We should be able to filter many of these out.

Another option to explore is whether we could use the `CopyMarker` metadata to reduce the number of timestamp checks. If doing this, we would also need to touch those files when copying. This could further speed builds when Build Acceleration is enabled.

We should also review why the file enumerations in the linked ticket occur on the main thread. None of the FUTDC file IO should be on the main thread.

drewnoakes:
All work here is either done or invalidated, except for:

> investigate using `CopyMarker` metadata to reduce the number of timestamp checks

I'm tracking that elsewhere, so will close this issue.

PR #8735, opened 2022-12-24
Optimize copy item checking when build acceleration is disabled

drewnoakes:
Contributes to #8733

Build Acceleration is a new feature that requires opt in.

When adding that feature, we started checking a larger number of potential inputs to projects. This included files like the output assemblies of referenced projects, along with their debug symbols and documentation files.

We need to check such files for the Build Acceleration feature, but when it is disabled there is no need to check them and doing so creates additional overhead that makes the FUTDC slower.

This change adds new metadata to those items, `BuildAccelerationOnly`, so the FUTDC can decide whether to check them or not.

That change filters all items from many projects during builds, leading to confusing messages being logged. Quite a bit of the diff here relates to moving logging code around to produce better quality log messages. In doing so, a more robust approach to controlling nesting of log messages was introduced. The diff is easier to read if white space changes are ignored.

We also remove the duplicate item checking that we saw previously as I never saw this being hit in practice, and tracking these items has a cost that I no longer believe is worth it.

Unit tests were extended to include items both with and without `BuildAccelerationOnly` metadata to validate behaviour. Note that the logging changes also created some diff in the unit tests of unrelated methods, however those changes mean that the tests now reflect the output that is actually seen when running the FUTDC in VS.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8735)

### DIFF
--- a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Log.cs
+++ b/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Log.cs
@@ -37,14 +37,17 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
                 _upToDateCheckConfiguredInput = upToDateCheckConfiguredInput;
                 _ignoreKinds = ignoreKinds;
                 _checkNumber = checkNumber;
+
                 _fileName = Path.GetFileNameWithoutExtension(projectPath);
             }
+
+            public Scope IndentScope() => new(this);
 
             private string Preamble()
             {
                 return Indent switch
                 {
-                    0 => "FastUpToDate: ",
+                    <= 0 => "FastUpToDate: ",
                     1 => "FastUpToDate:     ",
                     2 => "FastUpToDate:         ",
                     3 => "FastUpToDate:             ",
@@ -68,7 +71,7 @@ namespace Microsoft.VisualStudio.ProjectSystem.UpToDate
             {
                 if (level <= Level)
                 {
-                    // These are user visible, so we want them in local times so that 
+                    // These are user visible, so we want them in local times so that
                     // they correspond with dates/times that Explorer, etc shows
                     ConvertToLocalTime(ref arg0);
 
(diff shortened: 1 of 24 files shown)


tmeschter (approved):

melytc (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.Log.cs
```diff
@@ -213,6 +216,22 @@ public void UpToDate()
...
+                {
+                    _log = log;
+                    _log.Indent++;
+                }
+
+                public void Dispose()
+                {
+                    _log.Indent--;
+                }
+            }
```
MiYanni, line 234: We used a similar design to this for our REST API generators. Hmmm, just looked over that codebase and it only exists [as an example now](https://github.com/Azure/autorest.csharp/blob/29cd031d5f89644e997d667181b2452318215fff/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs#L650-L653). The bottom of the file contains [CodeScope](https://github.com/Azure/autorest.csharp/blob/29cd031d5f89644e997d667181b2452318215fff/src/AutoRest.CSharp/LowLevel/Generation/LowLevelExampleComposer.cs#L851-L883).

MiYanni (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/Resources.resx
```diff
@@ -375,13 +375,9 @@ This project was loaded using the wrong project type, likely as a result of rena
     <value>Checking items to copy to the output directory:</value>
   </data>
   <data name="FUTDC_CheckingCopyItemsForProject_1" xml:space="preserve">
-    <value>Checking copy items from project '{0}'.</value>
+    <value>Checking copy items from project '{0}':</value>
```
MiYanni, line 378: @drewnoakes This change is a problem for translations. This caused our translations to get reset because it is considered to be a 'new string'. So, you can [see here](https://github.com/dotnet/project-system/pull/8745) that these will all be reset because this was merged. ETA for getting updates is ~5 days so that'll be outside of the 17.5 snap.

Best solution I can do for now is block the translations PR until after the snap occurs.

drewnoakes: Thanks for catching that. This string only exists in diagnostic logs which we only started translating recently, so I'm not too concerned, though of course it's good to hold off merging the reset until new strings are available.

PR #8737, opened 2022-12-26
Add GenerateAssemblyInfo property

drewnoakes:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8737)


melytc (approved):

MiYanni:
@drewnoakes I don't think this can be part of `main`. The deadline for changes with translations was 12/16. We'll only get translations on our `main` which won't be targeting 17.5 after the snap this week.

PR #8738, opened 2022-12-29
Add Build Acceleration telemetry

drewnoakes:
Captures the status of the Build Acceleration (BA) feature any time the build is up-to-date. May be one of:

- Disabled, not candidate
- Disabled, candidate
- Enabled, not accelerated
- Enabled, accelerated

Builds that are not up-to-date log the reason, some of which could be helped by BA. We don't know for sure in this case however, as the check works differently when it encounters a copyable file depending upon whether BA is enabled or not.

- When enabled and we hit a copyable file, we remember the file and continue checking, then if we get to the end of the file and only have files to copy, we copy them.
- When disabled and we hit a copyable file, we return immediately on the first failure, even if the file was copyable. For that reason, we don't know for sure whether there were other reasons the project might have been built. The results of current telemetry give a bound for the impact BA would have if enabled.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8738)


drewnoakes:
Example of the new values:

image

melytc (approved):

adamint (approved):

PR #8741, opened 2023-01-03
Include Build Acceleration result when not up-to-date

drewnoakes:
PR #8738 added Build Acceleration telemetry to up-to-date checks. This PR includes the BA "reason" on failures (not up-to-date) too, allowing us to track how widely the feature might help by counting occurrences of `DisabledCandidate` results.


src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
```diff
@@ -973,6 +973,8 @@ async Task<(bool, ImmutableArray<ProjectConfiguration>)> CheckAsync(UpToDateChec
 
                 var fileSystemOperations = new FileSystemOperationAggregator(_fileSystem, logger);
 
+                logger.FileSystemOperations = fileSystemOperations;
```
smitpatel, line 976: Can this be in ctor rather than property assignment?

melytc (approved):

tmeschter (approved):

src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/UpToDate/BuildUpToDateCheck.cs
drewnoakes: Unfortunately not. I don't like this either. I'm considering quite a large refactor of the FUTDC which would avoid this issue, among other things.

PR #8772, opened 2023-01-06
Introduce SolutionSource

drewnoakes:
This class wraps solution event subscriptions in a more friendly wrapper that makes it easier to use the synchronous API correctly, hiding details of cookies and thread affinity, and supporting cancellation.

The pattern here is similar to that introduced in #7903.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/project-system/pull/8772)


src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ISolutionSource.cs
```diff
@@ -0,0 +1,30 @@
...
+using IAsyncDisposable = System.IAsyncDisposable;
+
+namespace Microsoft.VisualStudio.ProjectSystem.VS;
+
+/// <summary>
+/// Provides access to the <see cref="IVsSolution"/> object, and allows subscribing to solution
+/// events via the <see cref="IVsSolutionEvents"/> interface.
+/// </summary>
+[ProjectSystemContract(ProjectSystemContractScope.Global, ProjectSystemContractProvider.System)]
+internal interface ISolutionSource
```
MiYanni, line 13: ~~Excuse my ignorance here, but if we make an interface like this that contains access to CPS's version of the solution (I'm assuming that is what `IVsSolution` is), why does this new interface also not include Roslyn's version of the solution (the class is just called `Solution` and you get it from `Workspace`)? Is that not a use-case to have access to both? Or since we're responsible for maintain Roslyn's understanding that it doesn't make sense here?~~

So, I'm crossing out my statement above before I even make it because it is more me thinking aloud. In this context, I understand 'source' to mean the location of input information. If that is the case, then I guess Rolsyn's Solution class would be the 'output'? Would this 'source' concept apply to other things such as making an `IProjectSource`? I guess you'd need an `IActiveProjectSource` instead... or maybe I'm just thinking about this too much. 😵

MiYanni (approved):
It is looking like the point of this is to combine a repeated pattern you've observed into a class/interface that handles this pattern, correct? I'm just wondering how many other things like this exist in the code already.

src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/ISolutionSource.cs
drewnoakes: > It is looking like the point of this is to combine a repeated pattern you've observed into a class/interface that handles this pattern, correct? I'm just wondering how many other things like this exist in the code already.

That's right. I originally wrote this while coding up #7903 but decided to hold it back because it also exposed the `IVsSolution` object, not just the event subscription. At the time I was just thinking about events and wasn't sure how to expose the solution. Last week I was cleaning up some old branches and buffed that into what you see here.

From the linked PR:

> The `IVsRunningDocumentTable` and `IVsSolutionBuildManager` interfaces do not expose idiomatic .NET patterns, making it easy to hit bugs like https://github.com/dotnet/project-system/pull/7891.

>
> This PR creates new internal MEF components for both RDT and SBM events within VS, which wraps their COM-based APIs with more ergonomic patterns.

>
> It also adds async APIs so we can call from any thread.

Regarding your crossed out comments, I'll share that most (all?) `IVs*` interfaces come from old-school VS APIs (think COM). The VS view of a solution models different concerns to the Roslyn one, though there would be some overlap. At the same time Roslyn would try to minimise the dependencies upon VS-isms.

> I understand 'source' to mean the location of input information.

I meant "source" in that it's an object you can acquire the solution from. `*Source` is not a great name. `ISolution` would better match the names from #7903.

Actually we already have `ISolutionService`, so I will merge `ISolutionSource` into that interface.

smitpatel (approved):

